PR #34, opened 2015-02-10
Makes pre_inject_yum_repo plugin match multiline "RUN yum" commands

adamruzicka:


TomasTomecek:
Thanks for PR. The regex seems to work (requires `re.MULTILINE` though). Unfortunately the rest is not. The dockerfile is not changed, it's actually trashed. Will rework it.

TomasTomecek:
Merged from cli.

PR #52, opened 2015-03-20
tweaked prepub_tests_for_image to enable cloning git repo for tests

l-d-j:
enhanced plugin to enable cloning tests from repository.
to see how it works, clone:
https://github.com/l-d-j/wildfly-dock
and execute:
dock --verbose build --method here --json build.json

it will use tests from https://github.com/l-d-j/wildfly-test and execute them


TomasTomecek:
Thanks!

PR #53, opened 2015-03-20
fix handling of json log messages

mmilata:
wait_for_command tries to parse each line as a json. If the line was a
valid json that is not an object (e.g. "string") the code raised an
exception.


TomasTomecek:
Good catch!

PR #54, opened 2015-03-20
Options to allow connecting to insecure registries

mmilata:
It behaves rather weird wrt http/https and valid/invalid certificate - let's discuss this on Monday.


TomasTomecek:
Wow, this is fatter than I expected. Great job!

PR #63, opened 2015-03-28
Add a newline to the end of the LABEL line.

twaugh:


TomasTomecek:
Is this really necessary? Updated dockerfile is written to filesystem with `fp.writelines` which should append newline after every line.

twaugh:
Try it. ;-) I got a docker error about looking for "=" in "uname",  because it was trying to parse this:

```
LABEL "foo"="bar"RUN uname -a && env
```

From the documentation: "writelines() does not add line separators". Note that readlines() gives you a list of \n-terminated strings.

TomasTomecek:
Aha! Didn't know that. Thanks for this fix then.

PR #65, opened 2015-03-30
Escape keys and values for LABEL instruction.

twaugh:
Without escaping keys and values, an invalid Dockerfile might be written.


TomasTomecek:
could you please update line 29 in that file because it would be pretty misleading once this is merged

twaugh:
Python 2's string.maketrans() is an entirely different thing to Python 3's str.maketrans. It is more like 'tr'.

TomasTomecek:
Thanks Tim!

Issue #38, opened 2015-02-20
separate tests


TomasTomecek:
- [ ] create unit tests (mock docker and docker-registry)
- [ ] and integration tests (communicate with actual registry and docker daemon)

use flexmock (or something similar) to mock api

jpopelka:
Closing.

PR #77, opened 2015-04-29
Travis CI (#38)

jpopelka:
I haven't split the tests to unit & integration tests.

By default all docker.Client methods are now mocked (using flexmock),
so you don't need docker to run these tests.
If you want to run 'integration' tests, i.e. with running docker
instance, set NOMOCK environment variable:
$ NOMOCK=1 py.test -v

It needs some more polishing, but seems to work fine so far at least for Python 2.
https://travis-ci.org/jpopelka/dock/builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language: python
+python:
+  - "2.6"
+  - "2.7"
+  - "3.4"
+# command to install dependencies
+install:
+  - "pip install -r requirements.txt"
+  - "pip install -r tests/requirements.txt"
+# command to run tests
+script: py.test -v
+# run in a docker container
+sudo: false
+notifications:
+  email: false
--- a/tests/docker_mock.py
+++ b/tests/docker_mock.py
@@ -1,0 +1,76 @@
+import os
+import docker
+from flexmock import flexmock
+
+from tests.constants import COMMAND
+
+mock_containers = \
+    [{'Created': 1430292310,
+      'Image': 'fedora',
+      'Names': ['/goofy_mayer'],
+      'Command': '/bin/bash',
+      'Id': 'f8ee920b2db5e802da2583a13a4edbf0523ca5fff6b6d6454c1fd6db5f38014d',
+      'Status': 'Up 2 seconds'},
+     {'Created': 1430293290,
+      'Image': 'busybox:latest',
+      'Names': ['/boring_mestorf'],
+      'Id': '105026325ff668ccf4dc2bcf4f009ea35f2c6a933a778993e6fad3c50173aaab',
+      'Command': COMMAND}]
+
+mock_image = \
+    {'Created': 1414577076,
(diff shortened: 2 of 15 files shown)


.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: python
+python:
+  - "2.6"
+  - "2.7"
+  - "3.4"
+# command to install dependencies
+install:
+  - "pip install -r requirements.txt"
+  - "pip install flexmock"
```
TomasTomecek: might be a good idea to create e.g. `test_requirements.txt`

tests/docker_mock.py
```diff
@@ -0,0 +1,77 @@
...
+    flexmock(docker.Client, inspect_image=lambda im_id: inspect_result)
+    flexmock(docker.Client, logs=lambda cid, **kwargs: iter([mock_logs]) if kwargs.get('stream') else mock_logs)
+    flexmock(docker.Client, pull=lambda img, **kwargs: iter([mock_pull_logs]))
+    flexmock(docker.Client, push=lambda iid, **kwargs: mock_push_logs)
+    flexmock(docker.Client, remove_container=lambda cid, **kwargs: None)
+    flexmock(docker.Client, remove_image=lambda iid, **kwargs: None)
+    flexmock(docker.Client, start=lambda cid, **kwargs: None)
+    flexmock(docker.Client, tag=lambda img, rep, **kwargs: True)
+    flexmock(docker.Client, wait=lambda cid: 1)
+    flexmock(os.path, exists=lambda path: True)  # DOCKER_SOCKET_PATH check
```
TomasTomecek: wow; can imagine we could need this

tests/plugins/test_tag_and_push.py
```diff
@@ -16,6 +19,9 @@ class X(object):
 
 
 def test_tag_and_push_plugin(tmpdir):
+    if MOCK:
+        mock_docker()
```
TomasTomecek: out of curiosity, would it be possible to do this via pytest fixture?

TomasTomecek:
Jirko, ready to merge?

tests/docker_mock.py
jpopelka: I don't know how I can restrict the os.path.exists() mock, because something like the following leads to recursion.

flexmock(os.path, exists=lambda path: True if path == DOCKER_SOCKET_PATH else os.path.exists(path))

But so far it's OK as we use os.path.exists() only on one place where we check for DOCKER_SOCKET_PATH.

tests/plugins/test_tag_and_push.py
jpopelka: That doesn't seem to work. I already tried it in setup_function() with no luck.

jpopelka:
Sure, go ahead.

jpopelka:
Fixes #38

PR #78, opened 2015-05-01
New plugin: Add yum repo by url

twaugh:


TomasTomecek:
Tim, awesome work! Will pick a couple of tiny nits inline.

dock/plugins/pre_add_yum_repo_by_url.py
```diff
@@ -0,0 +1,64 @@
...
+}
+
+"""
+from dock.plugin import PreBuildPlugin
+import requests
+try:
+    # py2
+    from StringIO import StringIO
+    # use StringIO, not cStringIO
+    # cStringIO's readlines() gives bad results for unicode objects
```
TomasTomecek: hmm, looks like worth reporting upstream, since cStringIO should be faster version of StringIO

dock/plugins/pre_add_yum_repo_by_url.py
```diff
@@ -0,0 +1,64 @@
...
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param repourls: list of str, URLs to the repo files
+        """
+        # call parent constructor
+        super(AddYumRepoByUrlPlugin, self).__init__(tasker, workflow)
+        self.repourls = repourls
+
+    @staticmethod
+    def _get(url):
```
TomasTomecek: I'm assuming this is here so it can be mocked in tests, right? Since we already use flexmock, it may be worth mocking get of requests directly.

tests/plugins/test_add_yum_repo_by_url.py
```diff
@@ -0,0 +1,132 @@
...
+from dock.plugins.pre_add_yum_repo_by_url import AddYumRepoByUrlPlugin
+from dock.util import ImageName
+
+
+git_url = "https://github.com/TomasTomecek/docker-hello-world.git"
+TEST_IMAGE = "fedora:latest"
+
+
+class MockResponse(object):
+    def raise_for_status(self):
```
TomasTomecek: similar comment as with `_get`; would be great to use flexmock

dock/plugins/pre_add_yum_repo_by_url.py
twaugh: It only gives bad results on Python 2.6. For Python 2.7 it's fine.

twaugh:
I've updated it to use `flexmock` for `requests.get`. Thanks for the hint.

TomasTomecek:
Last time I merged (well, rebased actually) via CLI, github picked it up. Nvm, closing.

Thanks again Tim!

PR #84, opened 2015-05-07
GitPython -> git, clone only single branch

### LINKED ISSUE #68: Use correct refs to clone repo
Currently when you clone a git repository to get the files to build, you get all the refs with it. This makes it very slow, since *in all cases* you are not interested in all the refs, but only in the specified branch (most probably master), so `+refs/heads/master:refs/remotes/origin/master` should be used only.

This bug makes it slower and slower to build our images. It currently takes 1.5 minute (sic!) to checkout the repository, whereas with proper refs this takes 2 seconds.

TomasTomecek:
GitPython uses git command so why we should have such dep when in the
end we use it only to clone a git repo

Fixes #68

### DIFF
--- a/dock.spec
+++ b/dock.spec
@@ -22,6 +22,8 @@ BuildRequires:  python2-devel
 BuildRequires:  python-setuptools
 Requires:       python-dock
 
+Requires:       git >= 1.7.10
+
 %if 0%{?with_python3}
 BuildRequires:  python3-devel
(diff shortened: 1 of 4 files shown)


jpopelka:
Why were the tests failing with the previous commit ? That one was looking better to me.

TomasTomecek:
rebased to slavek's spec changes

PR #86, opened 2015-05-07
Several specfile improvements

bkabrda:
While working on Python 3 migration in Fedora, it proved useful to package applications like dock in a way that this PR proposes.
The advantage over the previous approach is, that there's a clear distinction between "dock" as an application (e.g. other packages would have `Requires: dock`) and dock as a library (e.g. other packages would have `Requires: python[3]-dock`).
This will also allow us to have "dock" as application both on systems with Python 2 as default as with Python 3 as a default - we'll just need to add a little conditional for the `/usr/bin/dock` symlink.

### DIFF
--- a/dock.spec
+++ b/dock.spec
@@ -8,7 +8,7 @@
 
 Name:           dock
 Version:        1.2.0
-Release:        2%{?dist}
+Release:        3%{?dist}
 
 Summary:        Improved builder for Docker images
 Group:          Development/Tools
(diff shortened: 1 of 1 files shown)


jpopelka:
LGTM

TomasTomecek:
Neat! Will check more thoroughly on Monday.

TomasTomecek:
This is really clever!

PR #93, opened 2015-05-12
Add squashing plugin

goldmann:
This commit adds squashing support for in Dock.

This feature requires docker-scripts package to be installed.

Usage:

A json build config file should be created with following content:

```
  "prepublish_plugins": [{
    "name": "squash",
      "args": {
        "tag": "SQUASH_TAG",
        "from_layer": "FROM_LAYER"
      }
    }
  }
```

The `tag` argument specifes the tag under which the new squashed image will
be registered. The `from_layer` argument specifies from which layer we want
to squash.

Of course it's possible to override it at runtime, like this: `--substitute prepublish_plugins.squash.tag=image:squashed
  --substitute prepublish_plugins.squash.from_layer=asdasd2332`.

It requires docker-scripts `0.3.2` or newer.

### DIFF
--- a/dock/plugins/prepub_squash.py
+++ b/dock/plugins/prepub_squash.py
@@ -1,0 +1,54 @@
+from dock.plugin import PrePublishPlugin
+from docker_scripts.squash import Squash
+
+__all__ = ('PrePublishSquashPlugin', )
+
+
+class PrePublishSquashPlugin(PrePublishPlugin):
+
+    """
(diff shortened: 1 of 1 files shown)


TomasTomecek:
Could you please add that nice, long documentation to the module as a docstring? (at some point, I would like to generate some `plugins.md` documentation out of those)
Otherwise LGTM!

TomasTomecek:
Thank you!

PR #98, opened 2015-05-14
Improve yum workflow

### LINKED ISSUE #83: improve yum workflow
When #82 is in place, we finally don't need to wrap `yum` calls with `printf ... >/etc/yum.repos.d/...`. I think we should improve the workflow to:

1. Add instruction after `FROM`/`MAINTAINER`:
 * either `ADD *.repo /etc/yum.repos.d/`
 * or `RUN wget -O /etc/yum.repos.d/ http://.../*.repo`
2. maintain yum cache and metadata during during build -- meaning, not interfering with present `RUN.*yum.*` calls
3. Add instruction at the end of dockerfile which erases all downloaded yum repos

After the final image is squashed, no build artifacts will leak.

twaugh:
Fixes #83.

### DIFF
--- a/dock/plugins/pre_add_yum_repo_by_url.py
+++ b/dock/plugins/pre_add_yum_repo_by_url.py
@@ -20,8 +20,10 @@ Example configuration to add content of repo file at URL:
 
 """
 from dock.plugin import PreBuildPlugin
+import os
 import os.path
 import re
+import requests
 
 try:
     # py2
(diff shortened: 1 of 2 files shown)


twaugh:
@landscape-bot compares pull request head to master, rather than to the branch point of the pull request head from master.

None of the new problems are actually introduced by this PR.

TomasTomecek:
@twaugh I've noticed that too.

Merging and am going to give it a go!

Thanks Tim!

Issue #96, opened 2015-05-13
plugin for storing dockerfile should figure out nvr from labels present in DF


TomasTomecek:
related #79 #91

PR #99, opened 2015-05-14
pre_add_dockerfile plugin can get N-V-R from LABELs (#96)

jpopelka:


TomasTomecek:
Looking at code, shlex takes care of quotes, right? So the key is not `"key"`, instead is `key`. 

Otherwise LGTM, very nice job!

jpopelka:
Yes, that's the reason for using shlex.

I've added one more testcase (test_adddockerfile_nvr_from_labels2) and I'm merging it myself.

Issue #94, opened 2015-05-12
problems combining --substitute and label plugins


jmtd:
You can't use `--substitute` with  `pre_add_labels_in_df.py`, because the value for the substitution has to be a dictionary (`{labelname: labelvalue}`) and the substitution feature only supports string-values.

Sort-of related, labels should be reversed-domain according to [docker upstream](https://docs.docker.com/userguide/labels-custom-metadata/) (`com.example...` etc.), but the `--substitute` command uses the dot character to determine which part of the configuration to substitute within.

jmtd:
Hi, related/masking this, the substitution has to be for a string only, whereas the `pre_add_labels_in_df` configuration needs a dict `{label: value}`

jmtd:
I've worked around this for what I'm doing right now by writing a wrapper-plugin that accepts a flat string as an argument and decodes it into a JSON dictionary at initialisation (before handing the result to `pre_add_labels_in_df`: https://bitbucket.org/jboss-dockerfiles/dock/commits/4d2c2e462cf15ea4364544bd28df08f04d0307b8?at=jdowland-serialised-label-plugin

TomasTomecek:
When this was implemented, we didn't even know about the reverse-domain policy.

`--substitute` option was implemented for you guys; so if you can come up with a better solution to this, I don't have problem merging it/improving it (but I feel like that passing json-encoded string via CLI is a bit rough -- on the other hand, we might not have another option)

jmtd:
Yeah we're very grateful for `--substitute`, thank you! We make extensive use of it: I've only just extended our use to labelling (and we aren't using my extension in production yet, we're still evaluating it).

I considered submitting this

```
-- a/dock/plugin.py
+++ b/dock/plugin.py
@@ -340,7 +340,8 @@ class InputPlugin(Plugin):
                 build_json[key] = value
             else:
                 try:
-                    plugin_type, plugin_name, arg_name = key_fragments
+                    plugin_type, plugin_name = key_fragments[0],key_fragments[1]
+                    arg_name = '.'.join(key_fragments[2:])
                 except ValueError:
                     logger.error("invalid absolute path: it requires exactly three parts: "
                                  "plugin type, plugin name, argument name separated be dot")
```

But it didn't seem right as I wouldn't be running the affected code.

TomasTomecek:
How about something like this:

```
--substitute "prebuild_plugins->add_labels_in_dockerfile->args->labels->com.example.key=value"
```

[...]

PR #102, opened 2015-05-15
Adjust pre_add_labels_in_df to accept JSON string

jmtd:
In order to be used in conjunction with --substitute, allow the
pre_add_labels_in_df to accept a string as well as a dict.  The
string must be a dict encoded into JSON.

Fixes: #94.


TomasTomecek:
thanks!

PR #106, opened 2015-05-19
DockerfileParser class for parsing Dockerfiles (#104)

### LINKED ISSUE #104: Introduce module/class for parsing dockerfiles
We are already parsing dockerfiles all over the place. Would be great to have a single codebase for that.

jpopelka:
I'm not sure about the name of the class - it's used not only to read & parse Dockerfile, but also to save content to Dockerfile, so Parser might not be the best name.

Fixes #104

### DIFF
--- a/dock/build.py
+++ b/dock/build.py
@@ -13,7 +13,7 @@ Logic above these classes has to set the workflow itself.
 import logging
 
 from dock.core import DockerTasker, LastLogger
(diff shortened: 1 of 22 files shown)


TomasTomecek:
I would name it the same. Will try to look at it later.

jpopelka:
It needed a rebase and I also added a support for handling (parsing/writing) non-ascii content.

TomasTomecek:
:+1:

Issue #48, opened 2015-03-18
provide a way to build from path, not using git


TomasTomecek:
* local builds: copy provided path to tmp dir and build from there
 * container builds: copy provided path to tmp dir and bind mount that folder to build container

bkabrda:
Here's my proposal to handle this (I'll take this bottom-up):
- build json schema will change (see the example at the very bottom):
  - build json will not contain any `git_*` attributes as it does now; it will rather contain attribute `content`, which will be mapped to a dictionary with keys `provider`, `uri`, `dockerfile_path` and `provider_params`
  - `provider` will have value either `fs_path` or `git`
  - `uri` will be URI of the target to build, e.g. `file:///tmp/foo/bar` or `https://github.com/foo/bar`
  - `dockerfile_path` will be path of Dockerfile inside the directory referenced by `URI`
  - `provider_params` will hold a dictionary of params related to provider, e.g. `git_branch`, `git_commit`, etc...
- `DockerBuildWorkflow.__init__` will change:
  - instead of accepting params `git_url`, `git_dockerfile_path` and `git_commit`, a single parameter `content` (as described above) will be accepted
- The question is what to do with `build_image_*` functions from `api`, which should be changed in the same way as `DockerBuildWorkflow.__init__`. If we can break the API, then I'd vote to just change them accor
dingly. Otherwise we'd need to do one of these (I don't like any of these options):
  - Introduce pair functions that would operate on paths, e.g. `build_image_here_from_path` etc.
  - Introduce new arguments to the current functions that would operate on paths, but then we'd have to check whether `git_*` arguments were passed and if not, then use `content`. Also, the API would get bloated
.

The advantage of this proposal is that it'd also be able to accomodate arbitrary content providers, e.g. mercurial, etc.
I'm not sure what these changes would have in `osbs`, but I guess we'd have to reflect them there as well.
Note: I'm still not sure about the parameter names that I chose, feel free to propose better ones.
Finally, an example of what this'd look like:

```
{
  "content": {
    "provider": "git",
    "uri": "https://github.com/foo/bar",
    "dockerfile_path": "",
    "provider_params": {
      "git_branch": "develop",
      "git_commit": "abcdefg"
    }
  }
}
```

TomasTomecek:
Very nice proposal! I really like this!

One downside is that it will make the final build json a bit complicated, but if we are planning to be more generic and support more inputs, there's no other way.

[...]

PR #107, opened 2015-05-20
Implement support for different source types (CLI untouched for now)

bkabrda:
This is a partial fix for #48. It changes dock's API, but CLI is left untouched for now - I will file a PR for it separately after this is accepted. Note that there are some subtle changes from my proposal in #48, see `docs/apidocs/top.md` diff for documentation.

Also, I believe that we should create a "json-able" class for the whole build configuration (usually called `build_json` in most parts of the source) to be able to validate it right after it enters API functions and to be able to use objects instead of dicts. The way I implemented this PR, the `Source` classes get into play quite a long way down the call chain, which I would like to avoid in the long run.


bkabrda:
Please don't merge yet, landscape discovered some actual errors that I need to fix first.

TomasTomecek:
will test extensively tomorrow

bkabrda:
Ready for review now!

jpopelka:
The code looks good to me. I haven't tested it manually but the tests look comprehensive and they pass so +1 from me.

TomasTomecek:
Slavek, I have rebased your work against current master, you can find it here https://github.com/DBuildService/dock/tree/107-rebased feel free to pull in my commits, update the PR and let's merge this!

TomasTomecek:
Good job!
Thanks Slavek!
(Merged from command line)

bkabrda:
Thanks!

Issue #108, opened 2015-05-20
split plugin add yum repo by url


TomasTomecek:
Since we are able to squash, there's pretty much no need for injecting yum calls inside dockerfile. Much nicer way will be to add repos with `ADD`, install packages and remove repos (`RUN rm /etc/yum.repos.d/*`) -- exactly what "add yum repo by url" plugin does. Therefore I think we should split the functionality of the plugin to two:

1. **fetch plugin**: downloads repo files from web
2. **add plugin**: hooks repos to build process by adding `ADD` and `RUN rm ...`

I'm not sure about the first one, how exactly should the plugin operate:
1. put repo configuration to `workflow.repos` object
2. **add** plugin will take everything from the `workflow.repos`, creates repo file out of it and adds instructions in dockerfile

or

1. **fetch** plugin downloads repo files, store them in image
2. **add** plugin will use downloaded files and also creates repo files from whatever is in `workflow.repos`

@twaugh what do you think?

twaugh:
I'm not sure of the utility of splitting fetching from adding. I can see that it might be useful to make it into a more generic 'fetch, add, remove' operation, without assuming it needs to add fetched content to `/etc/yum.repos.d/`, though.

Do you mean that several plugins might want to use `fetch` separately, and then `add` would deal with them all at once? I'm trying to think of a use-case.

TomasTomecek:
Right! My use case is that I want to use plugin which

 * will provide same functionality as plugin `inject_yum_repo`
 * can be used together with `koji` plugin

Essentially I want to take output of `koji` plugin and stuff that repo in dockerfile the same way `add_yum_repo_by_url` plugin does.

Does that make sense?

twaugh:
I see, got it. So how about this?:

```
{
    'name': 'fetch',
    'args': {
        'files': {
            '/etc/yum.repos.d/repo1.repo': 'https://example.com/repo1.repo',
            '/etc/yum.repos.d/repo2.repo': 'https://example.com/repo2.repo'
        }
    }
},
{
    'name': 'add'
}
```
where the pre_fetch plugin simply populates `workflow.tmpconfig` with the filename:url dict entries.

TomasTomecek:
so the koji plugin should store the repo file to the `workflow.tmpconfig`?

twaugh:
Yes, that's what I'm thinking.

```python
self.workflow.tmpconfig['/etc/yum.repos.d/koji.repo'] = ...
```

[...]

PR #109, opened 2015-05-20
add dockerfile: snapshot dockerfile

TomasTomecek:
If we edit dockerfile after running "add" plugin, we may expose some
unwanted information.

This patch copies dockerfile when the "add" plugin is run and stores
that snapshot in final image.

CC @jpopelka 
Related #108 

As I think about it, this new behavior should be configurable. What do you think Jirko?

(please don't merge)


jpopelka:
The code looks good to me.

> this new behavior should be configurable.

You mean conditional ? Like having a switch (bool arg) telling us whether to store the snapshot or final Dockerfile ? That might come in handy, yes.
Or how else configurable ?

TomasTomecek:
> Like having a switch (bool arg) telling us whether to store the snapshot or final Dockerfile ?

Exactly.

TomasTomecek:
rebased & updated; ready to merge from my PoV (not sure how this affects other PRs)

TomasTomecek:
@jpopelka 
you're right! thanks for noticing it

PR #113, opened 2015-05-22
split functionality of add_repos_by_url

### LINKED ISSUE #108: split plugin add yum repo by url

TomasTomecek:
Resolves #108

### DIFF
--- a/dock/constants.py
+++ b/dock/constants.py
@@ -22,3 +22,6 @@ CONTAINER_DOCKERFILE_PATH = os.path.join(CONTAINER_SHARE_PATH, 'Dockerfile')
 
 HOST_SECRET_PATH = ''
 
+YUM_REPOS_DIR = '/etc/yum.repos.d/'
+RELATIVE_REPOS_PATH = "dock-repos/"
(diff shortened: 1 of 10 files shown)


TomasTomecek:
unicode/str & py2/py3 is madness; we need `six`

PR #115, opened 2015-05-25
show and store commit ID of HEAD when building from git

### LINKED ISSUE #114: store commit ID of checked out branch when building from git
- [ ] put it to logs
- [ ] store it in OpenShift's annotations

TomasTomecek:
fixes #114


TomasTomecek:
@goldmann 
to elaborate a bit: we've got a bug report that builds are caching "something" (even though we are running everything with `--no-cache`); so to debug the issue, I want to show commit ID of HEAD in logs; this will also be very useful in final metadata

goldmann:
Right, I was wondering, if we could reuse it to set the `VcsRef` label: https://github.com/projectatomic/ContainerApplicationGenericLabels

TomasTomecek:
@goldmann good point! Let me create issue for that (I'm not sure when those will become "mandatory")

PR #121, opened 2015-05-27
squash: add option to store exported squashed tarball

### LINKED ISSUE #119: squash plugin: don't import image, place it on specific path instead
Since we are going to upload to pulp, which takes `tar.gz`. This should be configurable.

TomasTomecek:
Resolves #119

### DIFF
--- a/dock/inner.py
+++ b/dock/inner.py
@@ -11,6 +11,7 @@ Script for building docker image. This is expected to run inside container.
 
 import json
 import logging
+import os
 import shutil
 import tempfile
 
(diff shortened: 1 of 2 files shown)


jpopelka:
LGTM

jpopelka:
Actually the name of the argument, i.e. `dont_push` can be confusing.
I've renamed (as we agreed on recently) it to `dont_load`, because `docker-scripts squash -h` says:
```
--output-path OUTPUT_PATH
                        If not provided, image will be loaded into Docker
                        daemon
```

jpopelka:
Merging, I need it for #122 and it doesn't look like it should break anything.

PR #130, opened 2015-06-01
Make pulling from registry configurable for inside-build

### LINKED ISSUE #126: Make pulling from registry configurable for inside-build
Currently there is no option to disable pulling from registry the base image. In our case this will re-tag the images with older versions which is not what we want.

jpopelka:
Fixes #126 

Docs need to be regenerated.

### DIFF
--- a/dock/api.py
+++ b/dock/api.py
@@ -22,7 +22,7 @@ __all__ = (
 def build_image_in_privileged_container(build_image, source, image,
         parent_registry=None, target_registries=None, push_buildroot_to=None,
         parent_registry_insecure=False, target_registries_insecure=False,
-        **kwargs):
+        dont_pull_base_image=False, **kwargs):
     """
     build image from provided dockerfile (specified by `source`) in privileged image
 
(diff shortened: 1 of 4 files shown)


dock/api.py
```diff
@@ -37,6 +38,7 @@ def build_image_in_privileged_container(build_image, git_url, image,
     :param push_buildroot_to: str, repository where buildroot should be pushed
     :param parent_registry_insecure: bool, allow connecting to parent registry over plain http
     :param target_registries_insecure: bool, allow connecting to target registries over plain http
+    :param pull_base_image: bool, download base image updates from registry
```
TomasTomecek: could we change this? I would vote for something like `pulls or updates base image specified in dockerfile`

dock/inner.py
```diff
@@ -144,6 +144,7 @@ def __init__(self, git_url, image, git_dockerfile_path=None,
         :param plugin_files: list of str, load plugins also from these files
         :param parent_registry_insecure: bool, allow connecting to parent registry over plain http
         :param target_registries_insecure: bool, allow connecting to target registries over plain http
+        :param pull_base_image: bool, download base image updates from registry
```
TomasTomecek: since @goldmann's issue #126 is about `inside-build`, I think the argument should also be added there

Issue #135, opened 2015-06-02
Race condition when two builds pull the same base image


mmilata:
Start two identical builds roughly at the same time:

```
osbs -i test --verbose build --git-url "https://github.com/TomasTomecek/docker-hello-world.git" -u mmilata -c test-image2
osbs -i test --verbose build --git-url "https://github.com/TomasTomecek/docker-hello-world.git" -u mmilata -c test-image3
```

With a bit of luck you'll see for the first instance:

```
2015-06-02 18:31:18,845 - dock.build - INFO - pull base image from registry
2015-06-02 18:31:18,845 - dock.core - INFO - pull image from registry
2015-06-02 18:31:18,845 - dock.core - DEBUG - image = 'fedora:latest', insecure = 'None'
2015-06-02 18:31:21,761 - dock.util - INFO - wait_for_command
2015-06-02 18:31:21,763 - dock.util - DEBUG - Trying to pull repository registry.access.redhat.com/fedora ...
2015-06-02 18:31:29,352 - dock.util - INFO - no more logs
2015-06-02 18:31:29,354 - dock.core - INFO - tag image
2015-06-02 18:31:29,355 - dock.core - DEBUG - image = 'fedora:latest', target_image_name = 'fedora:latest'
2015-06-02 18:31:29,362 - dock.build - DEBUG - image 'fedora:latest' is available
2015-06-02 18:31:29,363 - dock.inner - INFO - running pre-build plugins
```

And for the second:

```
2015-06-02 18:31:21,681 - dock.build - INFO - pull base image from registry
2015-06-02 18:31:21,681 - dock.core - INFO - pull image from registry
2015-06-02 18:31:21,681 - dock.core - DEBUG - image = 'fedora:latest', insecure = 'None'
2015-06-02 18:31:21,708 - dock.util - INFO - wait_for_command
2015-06-02 18:31:22,453 - dock.util - INFO - no more logs
2015-06-02 18:31:22,453 - dock.core - INFO - tag image
2015-06-02 18:31:22,454 - dock.core - DEBUG - image = 'fedora:latest', target_image_name = 'fedora:latest'
Traceback (most recent call last):
  File "/usr/bin/dock", line 9, in <module>
    load_entry_point('dock==1.3.2', 'console_scripts', 'dock')()
  File "/usr/lib/python2.7/site-packages/dock-1.3.2-py2.7.egg/dock/cli/main.py", line 223, in run
    cli.run()
[...]

PR #136, opened 2015-06-02
remove_built_image: force removal of base image

### LINKED ISSUE #125: APIError(HTTPError('409 Client Error: Conflict',),) in remove_built_image plugin
```
2015-05-29 09:27:38,239 - dock.plugin - DEBUG - running plugin 'remove_built_image'
2015-05-29 09:27:38,239 - dock.plugin - INFO - running plugin instance with args: '{}'
2015-05-29 09:27:38,239 - dock.core - INFO - remove image from filesystem
2015-05-29 09:27:38,239 - dock.core - DEBUG - image_id = 'c54f0273f53a617c3936e59ebe4816c4b667c261c1fc930e0e6ac7d15f43f653'
2015-05-29 09:27:38,324 - dock.core - INFO - remove image from filesystem
2015-05-29 09:27:38,324 - dock.core - DEBUG - image_id = 'brew/rhel7:latest'
2015-05-29 09:27:38,349 - dock.plugin - WARNING - Plugin 'remove_built_image' raised an exception: 'APIError(HTTPError('409 Client Error: Conflict',),)'
2015-05-29 09:27:38,381 - dock.plugin - DEBUG - Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/dock/plugin.py", line 211, in run
  File "/usr/lib/python2.7/site-packages/dock/plugins/post_remove_built_image.py", line 43, in run
    self.tasker.remove_image(ImageName.parse(self.workflow.pulled_base_image))
  File "/usr/lib/python2.7/site-packages/dock/core.py", line 427, in remove_image
    self.d.remove_image(image_id, force=force, noprune=noprune)  # returns None
  File "/usr/lib/python2.7/site-packages/docker/client.py", line 789, in remove_image
    self._raise_for_status(res)
  File "/usr/lib/python2.7/site-packages/docker/client.py", line 95, in _raise_for_status
    raise errors.APIError(e, response, explanation=explanation)
APIError: 409 Client Error: Conflict ("Conflict, cannot delete 10acc31def5d because the running container 34ec1869229c is using it, stop it and use -f to force")

2015-05-29 09:27:38,382 - dock.plugin - INFO - Error is not fatal. Continuing...
```

I'm guessing that what happened here is the plugin tried to delete the base image but someone else was using it at the same time. We should probably wrap this in a try...catch block and ignore the error if that is the case.

mmilata:
Fixes #125.

Removal should be safe since we are referring to the image by name which means we just untag it?

This gets rid of the error message in #125 but the base image handling has still problems when it comes to concurrent builds (see e.g. #135).

### DIFF
--- a/dock/build.py
+++ b/dock/build.py
@@ -123,14 +123,14 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
             base_image_with_registry.registry = source_registry
 
         base_image = self.tasker.pull_image(base_image_with_registry, insecure=insecure)
+        pulled_tags = set([base_image])
 
         if not self.base_image.registry:
             response = self.tasker.tag_image(base_image_with_registry, self.base_image, force=True)
-        else:
-            response = base_image
+            pulled_tags.add(response)
 
         logger.debug("image '%s' is available", response)
-        return response
+        return pulled_tags
 
     def build(self):
         """
--- a/dock/inner.py
+++ b/dock/inner.py
@@ -165,7 +165,7 @@ class DockerBuildWorkflow(object):
         self.built_image_inspect = None
 
         self.dont_pull_base_image = dont_pull_base_image
-        self.pulled_base_image = None
+        self.pulled_base_images = set()
 
         # squashed image tarball
         # set by squash plugin
@@ -192,8 +192,8 @@ class DockerBuildWorkflow(object):
         self.builder = InsideBuilder(self.source, self.image)
         try:
             if not self.dont_pull_base_image:
-                self.pulled_base_image = self.builder.pull_base_image(self.parent_registry,
-                                                                      insecure=self.parent_registry_insecure)
+                self.pulled_base_images = self.builder.pull_base_image(self.parent_registry,
+                                                                       insecure=self.parent_registry_insecure)
 
             # time to run pre-build plugins, so they can access cloned repo,
             # base image
--- a/dock/plugins/post_remove_built_image.py
+++ b/dock/plugins/post_remove_built_image.py
@@ -11,6 +11,7 @@ Remove built image (this only makes sense if you store the image in some registr
 from dock.plugin import PostBuildPlugin
 from dock.util import ImageName
 
+from docker.errors import APIError
 
 __all__ = ('GarbageCollectionPlugin', )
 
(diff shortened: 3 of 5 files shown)


TomasTomecek:
Hm, does this really help? My testing in command line:

```
1% docker run -ti d2738f3875c2 bash
[root@96cdbfe23ec4 /]#

# moving to other terminal

2% docker rmi d2738f3875c2
Error response from daemon: Conflict, cannot delete d2738f3875c2 because the running container 96cdbfe23ec4 is using it, stop it and use -f to force
FATA[0000] Error: failed to remove one or more images

2% docker rmi -f d2738f3875c2
Error response from daemon: Conflict, cannot force delete d2738f3875c2 because the running container 96cdbfe23ec4 is using it, stop it and retry
FATA[0000] Error: failed to remove one or more images
```

As I'm thinking about this issue, I would be probably in favor of ignoring the message. Here's what I understand is happening:

1. Start a build with id 1 using base image X
2. Start a build with id 2 using base image X
3. Build one finishes, garbage collection kicks in
 * Build 2 is still running
 * We call `docker rmi X`
 * Docker fails because X is in use

Once build 2 finishes, it will take care of removing image X, so I think it's perfectly fine if X wasn't removed during build 1.

What do you think Martin?

mmilata:
Maybe I was just lucky enough to not reproduce the race after applying the patch.

Also, I think we are removing the base image by name (tag?):
```
2015-06-02 18:31:46,067 - dock.core - INFO - remove image from filesystem
2015-06-02 18:31:46,068 - dock.core - DEBUG - image_id = 'fedora:latest'
```

I thought that in this case docker will only remove the *fedora:latest* tag and the image itself will be removed once it is no longer used. This does not appear to be the case though.

Let me play with this a bit more ...

mmilata:
Ok, force-pushed a branch that ignores & logs the failure to delete the image. I'm afraid we'll need some kind of garbage collector that deletes unused (base) images after some time. Let's talk about this offline.

mmilata:
Oh, it breaks tests. I'll fix it tomorrow ...

mmilata:
Alright, I think this is ready for review. It should fix #125 and #140. @TomasTomecek, can you please look at it when you have time?

TomasTomecek:
Looks fine! Thanks!

PR #138, opened 2015-06-03
Allow specifying various source providers from CLI. Fixes #48

### LINKED ISSUE #48: provide a way to build from path, not using git

bkabrda:
The only issue is that the manpage generator doesn't seem to work with nested subparsers. Any advise on that would be welcome.


TomasTomecek:
Very nice work! Let's merge this and open issue on manpage generator.

PR #147, opened 2015-06-05
Make sure that 'source' isn't repeated twice for path provider

bkabrda:
I believe this regression was introduced with e4480c587f86f59ca5bfa2de7d605fe636155f5b (but it was the right way to do stuff, so I'm just fixing the code around to adapt it to the correct behaviour).


TomasTomecek:
Thanks Slavek!

PR #159, opened 2015-06-10
Store crane registry API in the push configuration.

twaugh:
This ensures it gets stored in the osv3 metadata.

Untested (having trouble testing).

### DIFF
--- a/dock/inner.py
+++ b/dock/inner.py
@@ -138,6 +138,12 @@ class PushConf(object):
         for registry_uri in registry_uris:
             self.add_docker_registry(registry_uri, insecure=insecure)
 
+    def add_pulp_registry(self, name, crane_uri):
+        if crane_uri is None:
+            raise RuntimeError("registry URI cannot be None")
+        r = PulpRegistry(name, crane_uri)
+        self._registries["pulp"].append(r)
+
     @property
     def has_some_docker_registry(self):
         return len(self.docker_registries) > 0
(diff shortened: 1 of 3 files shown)


twaugh:
So, this works in as much as the metadata that gets stored now has e.g.:

```
'repositories': '{"unique": ["mynamespace/myrepo:origin-20150610155943", "https://registry.example.com/mynamespace/myrepo:origin-20150610155943"], "primary": ["docker-hello-world:1.0_1", "docker-hello-world:1.0", "https://registry.example.com/docker-hello-world:1.0_1", "https://registry.example.com/docker-hello-world:1.0"]}'
```

But is that what we need?

TomasTomecek:
I feel like it should be:
```json
{
  "unique": [
    "registry.example.com/mynamespace/myrepo:origin-20150610155943"
  ],
  "primary": [
    "registry.example.com/docker-hello-world:1.0_1",
    "registry.example.com/docker-hello-world:1.0"
  ]
}
```
I guess we need to tune the crane's config and lose the `http[s]://` part.

twaugh:
OK, that might just be my bad local configuration. I was using 'https://...' in the `[registries]` section, when it should just have been the hostname (and maybe port).

But the registry-less parts there e.g. 'docker-hello-world:1.0' come from the docker registries in the push configuration, not the pulp/crane registries.

Perhaps there shouldn't be docker registries set at all, only pulp/crane registries.

twaugh:
OpenShift was setting OUTPUT_REGISTRY= in the environment rather than leaving it unset.

Now I'm getting what you expect.

PR #160, opened 2015-06-11
Revert previous spec change (02a409e7b4) and fix the code.

jpopelka:
It means that it won't be possible to do
`from dock import build_image_here` but only `from dock.api import build_image_here`, but `docs/api.md` also says to do `from dock.api import ...`


TomasTomecek:
okay, no strong feeling for one way or the other :) `from dock import ...` was really meant for convenience

jpopelka:
Ok, merging then, thanks.

Issue #163, opened 2015-06-11
RFE: PrePublishSquashPlugin: make storing metadata possible with dont_load=false


pbabinca:
I believe squash plugin should always store metadata even if it doesn't load it back to docker.

jpopelka:
AFAIK the `docker-scripts squash` by default (if --output-path isn't specified) doesn't create the tarball, so by default we have no file to get metadata about after squashing.

TomasTomecek:
Exactly as Jirka says. When squashed image is loaded, we should use the updated image in engine.

TomasTomecek:
Okay so... When loading image back to engine, we should export it and store it somewhere and of course, process the metadata.

lcarva:
Closing due to lack of interest and activity. If you're still interested feel free to reopen it.

PR #164, opened 2015-06-11
PrePublishSquashPlugin: Always store metadata of a image tarball

pbabinca:
Resolves: #163


mmilata:
Closing this after discussing on IRC.

TomasTomecek:
To elaborate a bit. With current codebase, this can't work because when the image is loaded, there is no way to do the metadata; we would need to

1. export the image and do the metadata
2. or never let squash tool load the image & do the metadata & load it back to engine ourselves

So basically this is a RFE.

PR #167, opened 2015-06-11
Allow using ImageName as dict key in Python 3

bkabrda:
I'd like to use this in https://github.com/bkabrda/atomicapp-builder. Thanks.


jpopelka:
LGTM

Issue #177, opened 2015-06-12
yum_inject_repo: fails if packages are installed using a script


twaugh:
If packages are installed using a script, 'yum' isn't found in the Dockerfile and so the repo file is never added.

twaugh:
@goldmann

twaugh:
The approach I propose is to:

* inspect the base image to understand what user is being inherited from there,
* track changes made to that by `USER` instructions in the Dockerfile
* finally, now we now which user was meant to be set, add `USER root`, `RUN rm ...`, and (if needed) `USER theuser` to restore it.

twaugh:
Fixed via #179.

PR #179, opened 2015-06-12
yum_inject_repo: handle non-root USER with install script (#177).

twaugh:
Inspect the base image to understand what user is being inherited.
Track changes made to that by USER instructions in the Dockerfile.
Set USER to root if needed before 'RUN rm'.
Restore USER as needed.

Please test.

### DIFF
--- a/dock/plugins/pre_inject_yum_repo.py
+++ b/dock/plugins/pre_inject_yum_repo.py
@@ -24,7 +24,7 @@ def alter_yum_commands(df, wrap_str):
     return regex.sub(sub_func, df)
 
 
-def add_yum_repos_to_dockerfile(yumrepos, df):
+def add_yum_repos_to_dockerfile(yumrepos, df, inherited_user):
     df_lines = df.lines
     if len(df_lines) == 0:
         raise RuntimeError("Empty Dockerfile")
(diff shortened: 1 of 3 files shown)


TomasTomecek:
Will look into this on Monday.

twaugh:
Seems to work as intended from testing a prod run here with both (a) base layer leaving us as root and (b) base layer leaving us an non-root.

dock/plugins/pre_inject_yum_repo.py
```diff
@@ -34,12 +34,12 @@ def add_yum_repos_to_dockerfile(yumrepos, df):
     preinsert = None
     structure = df.structure
     for insndesc in structure:
-        insn = insndesc['instruction'].lower()
-        if insn == 'maintainer':
+        insn = insndesc['instruction']
+        if insn == 'MAINTAINER':
```
goldmann: Will it also work with lower case?

twaugh: Yes, `insndesc['instruction']` is changed to upper-case in `Dockerfile.structure()` now.

twaugh:
We had a mixed-case test for that before, but it got lost in the change. I've added one back in.

goldmann:
I've tested this PR and it seems to work.

PR #180, opened 2015-06-15
DockerfileParser: add 'value' key to structure property and make get_* functions use it

jpopelka:
@twaugh Please review


dock/util.py
```diff
@@ -176,19 +176,27 @@ def structure(self):
...
-             "content": "From fedora\n"},
+             "content": "From fedora\n",
+             "value": "fedora"},
 
             {"instruction": "CMD",
              "startline": 1,
              "endline": 2,
-             "content": "CMD yum -y update && \\\n    yum clean all\n"}
+             "content": "CMD yum -y update && \\\n    yum clean all\n",
+             "value": "yum -y update && yum clean all"}
```
twaugh: What would the value be expected to be for `"CMD yum -y update &&     \\\n  yum clean all"`? Currently leading whitespace is stripped while whitespace up to `\` is preserved. Is that what docker does? What about `"RUN ls&&\\\nusb\n"`?

twaugh:
Oh, right, there's a `test_util.py`, hadn't thought of looking for it. I've got a test-case for `.structure` to add to that after this PR.

twaugh:
Looks really good. I don't know enough about how docker handles whitespace...

twaugh:
OK, the answer is that docker strips whitespace between the instruction and the first character of the value, and after the last character of the value, but not between continuation lines.

So I think the `lstrip()` should only happen for the first line of a multiline set, and `rstrip()` should only happen for the last.

twaugh:
FWIW, here's the test-case I've written: d88e1e21.

twaugh:
Actually, better one: 92afe022.

twaugh:
Looks ready for merge. (Sorry about the typo in the test...)

jpopelka:
@twaugh could you review also the last commit (I haven't touched the others). It fixes #181.

I tested that docker parse it the same via putting into Dockerfile:
LABEL label11 11
LABEL label12 1 2
LABEL "label13" 1 3
LABEL label14 "1" 4
LABEL label15 1 '5'
then `docker build` & `docker inspect` results in:
        "Labels": {
            "label11": "11",
            "label12": "1 2",
            "label13": "1 3",
            "label14": "1 4",
            "label15": "1 5"
        }

twaugh:
:+1:

PR #182, opened 2015-06-16
Fix traceback when FROM image tag includes registry

twaugh:
Noticed by @goldmann.


jpopelka:
OK, I'm fine with this and Tim's replies to my comments.

PR #183, opened 2015-06-17
Clarify description for add_yum_repo plugins (#172)

### LINKED ISSUE #172: Fix docs for add_yum_repo_by_url plugin
Currently it's necessary to add also `inject_yum_repo` to make the `add_yum_repo_by_url` plugin work, but it's not mentioned anywhere. Moreover, the plugin description suggests that to add the repo file you need to only enable this plugin.

In fact the name of the plugin is wrong too. It does not add the repo, but it only fetches that file.

/cc @bkabrda

twaugh:
Fixes #172.


TomasTomecek:
:+1:

PR #184, opened 2015-06-18
tag_conf: create only single repository

TomasTomecek:
...so we don't have two repositories in pulp


mmilata:
Looks good, merged.

PR #188, opened 2015-06-22
Rename to atomic-reactor. Fixes #56.

### LINKED ISSUE #56: Rename project
The current name of the project is very bad. In most cases when you talk about it you need to explain if you are talking about Dock indeed or Docker but you just typed it wrongly. It is very confusing to use such similar name.

bkabrda:
Note: this PR makes py.test skip two tests, since they try to checkout "DBuildService/atomic-reactor" (I already renamed that in the code), which doesn't exist yet. I'll un-skip these when we actually rename the repo. We may want to move the repo under projectatomic before we do that, so that I can also fix the org name in one commit.
Thanks for reviewing!


TomasTomecek:
Hm, we also need to create new review in fedora.

After a brief look, this looks fine. I would like to merge #184 first, do last dock release (because we need that patch on prod) and then merge this and do the renaming.

bkabrda:
Ok, I'll wait for #184 to be merged, rebase this patch and then hopefully wait for this to be accepted :)

Yeah, we'll need a new Fedora review, but that shouldn't be such a big problem. I'm volunteering as either maintainer or reviewer/comaintainer, if noone else wants to take this.

bkabrda:
Rebased and ready to be pulled.

PR #192, opened 2015-06-23
Implement '--input auto'. Fixes #189

### LINKED ISSUE #189: Automatically detect input for inside builds
It should be possible to detect input automatically for inside builds. I think we can create an input plugin that would be triggered on `--input auto` (and will also be the default), that will decide which input to use. Users will still be able to explicitly select different input method if they need to do that for some reason.

bkabrda:


TomasTomecek:
Looks great! I'm fine with merging this.

PR #193, opened 2015-06-23
fix unique repositories in build metadata

TomasTomecek:
since we changed the names of built image in #184, we also need to
update unique repositories in build metadata, therefore this is a fixup
to #184 

Not sure if I should rename the `add_image` method. This may be confusing a bit.

### DIFF
--- a/tests/plugins/test_tag_by_labels.py
+++ b/tests/plugins/test_tag_by_labels.py
@@ -68,8 +68,14 @@ def test_tag_by_labels_plugin(tmpdir):
     assert TagByLabelsPlugin.key in output.keys()
     assert len(workflow.tag_conf.images) == 4
     images = [i.to_str() for i in workflow.tag_conf.images]
+    primary_images = [i.to_str() for i in workflow.tag_conf.primary_images]
+    unique_images = [i.to_str() for i in workflow.tag_conf.unique_images]
     assert ("%s:%s" % (TEST_IMAGE, "unique_tag_123")) in images
     assert ("%s:%s_%s" % (TEST_IMAGE, version, release)) in images
     assert ("%s:%s" % (TEST_IMAGE, version)) in images
     assert ("%s:latest" % (TEST_IMAGE, )) in images
+    assert ("%s:%s" % (TEST_IMAGE, "unique_tag_123")) in unique_images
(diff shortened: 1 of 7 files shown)


jpopelka:
> Not sure if I should rename the add_image method.

I'd rename:
add_image() -> add_primary_image()
add_images() -> add_primary_images()

> This may be confusing a bit.

What, old name or new name ?

TomasTomecek:
> What, old name or new name ?

Old names.

`add_primary_image(s)` sounds better to me

TomasTomecek:
Fixed, updated docstrings, rebased.

TomasTomecek:
Can someone review and merge please?

twaugh:
Looks like coveralls is complaining about [`atomic_reactor/inner.py:86`](https://coveralls.io/builds/2948788/source?filename=atomic_reactor%2Finner.py#L86). Do we test `TagConf.primary_images`?

tests/plugins/test_tag_by_labels.py
```diff
@@ -68,8 +68,14 @@ def test_tag_by_labels_plugin(tmpdir):
     assert TagByLabelsPlugin.key in output.keys()
     assert len(workflow.tag_conf.images) == 4
     images = [i.to_str() for i in workflow.tag_conf.images]
+    primary_images = [i.to_str() for i in workflow.tag_conf._primary_images]
```
twaugh: Should this be `workflow.tag_conf.primary_images`?

TomasTomecek:
AFAIK there are no `TagConf` tests. Will fix.

tests/plugins/test_tag_by_labels.py
TomasTomecek: you are totally right!

twaugh: So, really this was found by coveralls, but it took quite a bit of digging. Although it tells you the relative increase/decrease for coverage, it won't tell you which lines cause the change! You have to compare with the base version side-by-side.

That gets you as far as finding out which bit isn't tested, so then it's a case of reading through the code to find out why it isn't, which lead me here.

jpopelka:
LGTM

TomasTomecek:
From my PoV, yes.

TomasTomecek:
Thanks Jirko!

PR #194, opened 2015-06-26
Improve logging

bkabrda:
- unify log messages to start lowercase
- provide more info on what is going on

Hopefully this will make logs more useful (and also it should make them look a bit better :))

### DIFF
--- a/atomic_reactor/build.py
+++ b/atomic_reactor/build.py
@@ -30,6 +30,7 @@ class ImageNotBuilt(Exception):
 class BuilderStateMachine(object):
     def __init__(self):
         self.is_built = False
+        self.image = None
 
     def _ensure_is_built(self):
         """
@@ -38,7 +39,7 @@ class BuilderStateMachine(object):
         :return: None
         """
         if not self.is_built:
-            logger.error("Image is not built yet!")
+            logger.error("Image '%s' is not built yet!", self.image)
             raise ImageNotBuilt()
 
     def _ensure_not_built(self):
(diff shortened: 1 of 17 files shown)


atomic_reactor/build.py
```diff
@@ -38,7 +39,7 @@ def _ensure_is_built(self):
         :return: None
         """
         if not self.is_built:
-            logger.error("Image is not built yet!")
+            logger.error("Image '%s' is not built yet!", self.image)
```
twaugh: Looks like there are still a few places that have capital letters at the start.

atomic_reactor/build.py
```diff
@@ -48,7 +49,7 @@ def _ensure_not_built(self):
         :return: None
         """
         if self.is_built:
-            logger.error("Image is already built!")
+            logger.error("Image '%s' is already built!", self.image)
```
twaugh: Leading capital

atomic_reactor/buildimage.py
```diff
@@ -55,7 +55,7 @@ def create_image(self, df_dir_path, image, use_cache=False):
         :param image:
         :return:
         """
-        logger.debug("df_dir_path = '%s', image = '%s'", df_dir_path, image)
+        logger.debug("Creating build image: df_dir_path = '%s', image = '%s'", df_dir_path, image)
```
twaugh: Leading capital

bkabrda:
Thanks for the catches, Tim. I fixed these in the second commit.

bkabrda:
In some cases, yes. The thing is, the new info messages don't contain all the information that's also debug-logged for all the functions (it wouldn't make sense to do so). So I'd rather log all arguments for all core functions separately, even if they're logged twice on debug.

jpopelka:
OK, feel free to rebase and merge.

TomasTomecek:
lower case messages; Slavek, you are my hero! Anyway, this will be really handy.

PR #195, opened 2015-06-26
pulp_push: only publish repos we pushed to (suggested by jluza)

twaugh:
Calling `crane()` with no arguments publishes all the repositories in pulp, spawning a task for each one and waiting for it before proceeding to the next one.

This is O(n) for the number of repositories in the registry.

What we actually want to do is only publish the repositories we just pushed to, bringing the cost down to O(1) -- this takes seconds, instead of minutes or hours.


jpopelka:
LGTM

TomasTomecek:
lol, this would be real fun after having 100 pulp repos

PR #197, opened 2015-06-29
plugin,metadata: avoid specifying API version

mmilata:
Even though osbs uses v1beta3 now the plugin had hardcoded configuration
for v1beta1.

### DIFF
--- a/atomic_reactor/plugins/post_store_metadata_in_osv3.py
+++ b/atomic_reactor/plugins/post_store_metadata_in_osv3.py
@@ -9,7 +9,6 @@ from __future__ import unicode_literals
 
 import json
 import os
-from osbs.core import Openshift
 
 try:
     # py2
@@ -17,6 +16,9 @@ try:
 except Exception:
     # py3
     from urllib.parse import urljoin
+
+from osbs.api import OSBS
+from osbs.conf import Configuration
 
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.plugins.pre_return_dockerfile import CpDockerfilePlugin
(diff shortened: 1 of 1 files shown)


atomic_reactor/plugins/post_store_metadata_in_osv3.py
```diff
@@ -55,13 +58,23 @@ def run(self):
...
         self.log.info("build id = %s", build_id)
 
-        api_url = urljoin(self.url, "/osapi/v1beta1/")
-        oauth_url = urljoin(self.url, "/oauth/authorize")  # MUST NOT END WITH SLASH
-
         # initial setup will use host based auth: apache will be set to accept everything
         # from specific IP and will set specific X-Remote-User for such requests
-        # FIXME: use OSBS here
-        o = Openshift(api_url, oauth_url, None, use_auth=self.use_auth, verify_ssl=self.verify_ssl)
+        with tempfile.NamedTemporaryFile(mode='w+') as f:
```
TomasTomecek: I think we don't need to store anything on disk to make this work, something like this should be possible:
```
Configuration(conf_file=None, conf_section=None, cli_args=None, openshift_uri="...", use_auth=..., ver...)
```
If this won't work, I'm in favor of patching osbs.

atomic_reactor/plugins/post_store_metadata_in_osv3.py
```diff
@@ -111,4 +124,4 @@ def run(self):
                 "sha256sum": tar_sha256sum,
                 "filename": os.path.basename(tar_path),
             })
-        o.set_annotations_on_build(build_id, labels)
+        osbs.os.set_annotations_on_build(build_id, labels)
```
TomasTomecek: hm, looks like we should make api method for annotations

mmilata:
I'll fix OSBS to not require a config file (conf_file=None doesn't seem to work) and to provide annotations API method, then update this PR.

mmilata:
I've committed the required osbs changed so this is ready for review.

TomasTomecek:
Works fine. Merging.

PR #198, opened 2015-06-29
Update build_json docs (#169)

### LINKED ISSUE #169: Update build_json docs
https://github.com/DBuildService/dock/blob/master/docs/build_json.md is outdated and it doesn't take the new `source` param into account.

twaugh:
This updates `docs/build_json.md` to reflect how the JSON actually looks. Fixes #169.

### DIFF
--- a/docs/build_json.md
+++ b/docs/build_json.md
@@ -4,10 +4,15 @@ If you want to take advantage of _inner_ part logic of Atomic Reactor, you can d
 
 ```json
 {
-    "git_url": "http://...",
+    "source": {
+        "provider": "git",
+        "uri": "http://...",
+        "dockerfile_path": "django/",
+        "provider_params": {
+            "git_commit": "devel"
+        }
+    },
     "image": "my-test-image",
-    "git_dockerfile_path": "django/",
-    "git_commit": "devel",
     "parent_registry": "registry.example.com:5000",
     "target_registries": ["registry.example2.com:5000"],
     "prebuild_plugins": [{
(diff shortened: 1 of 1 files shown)


bkabrda:
There's a small issue here - the `git_commit` param isn't at top level, but in `provider_params`. So the example json should look like this:

```
"source": {
    "provider": "git",
    "uri": "http://...",
    "dockerfile_path": "django/",
    "provider_params: {
        "git_commit": "devel"
    }
},
```

It mirrors the structure passed to API functions as described in https://github.com/DBuildService/atomic-reactor/blob/master/docs/api.md#source

TomasTomecek:
I got really stuck on this yesterday when I tried to create build json myself.

twaugh:
@bkabrda Thanks, fixed.

bkabrda:
I'd probably consider explaining what `provider_params` is for (extra parameters that may be different across providers), but it's not a blocker for me, so I'm +1 to merge this even without doing that.

TomasTomecek:
Also, should this part of the code be documented?

https://github.com/DBuildService/atomic-reactor/blob/master/atomic_reactor/cli/main.py#L78

Is it used in substitutions and/or `atomic-reactor build json ...`?

(I know it's related to CLI and not build json itself; just wanna have the conversation)

bkabrda:
@TomasTomecek well this is actually not part of the build json, but something that build json is constructed from. So I think this is outside the scope of this bug. I guess it should be documented as part of finishing https://github.com/DBuildService/atomic-reactor/issues/48#issuecomment-109168847

TomasTomecek:
@bkabrda Agreed.

bkabrda:
Merged. Thanks, Tim!

PR #199, opened 2015-06-30
spec: define macros for RHEL-6

jpopelka:
Same problems as in [python-dockerfile-parse review](https://bugzilla.redhat.com/show_bug.cgi?id=1236151#c3)


bkabrda:
Looks good, I'm merging this.

PR #201, opened 2015-06-30
Remove internal DockerfileParser and use dockerfile-parse package.

### LINKED ISSUE #149: Pull out DockerfileParser to separate python package
I'd like to parse Dockerfile but I don't want to pull in all the dependencies which dock has. I'm looking at this class: https://github.com/DBuildService/dock/blob/master/dock/util.py#L98.

jpopelka:
Fixes #149


jpopelka:
I think the coverage decreased because I removed the code that's been covered quite well.

twaugh:
Commit message is missing the 'le' of 'Dockerfile'.

The coverage decrease looks incorrect to me. I think it's talking about one of the 3 'missed' lines in `atomic_reactor/plugins/pre_add_labels_in_df.py` but I can't see that this change would have any effect on that.

twaugh:
LGTM

PR #202, opened 2015-06-30
Fix pulp test

twaugh:
The pulp test-case is skipped in Travis CI because of the `dockpulp` dependency. Running it locally, it fails, simply because the mocked dockpulp interface doesn't accept the argument we now pass to `dockpulp.Pulp.crane()`.

I've also added some testing for error cases.

### DIFF
--- a/atomic_reactor/plugins/post_push_to_pulp.py
+++ b/atomic_reactor/plugins/post_push_to_pulp.py
@@ -135,13 +135,13 @@ class PulpUploader(object):
 
 def compress(filename, ifp):
     _chunk_size = 1024*1024  # 1Mb buffer for writing file
-    with gzip.open(filename, "wb", compresslevel=6) as wfp:
-        while True:
-            data = ifp.read(_chunk_size)
-            if data == b'':
-                break
+    wfp = gzip.open(filename, "wb", compresslevel=6)
+    while True:
+        data = ifp.read(_chunk_size)
+        if data == b'':
+            break
 
-            wfp.write(data)
+        wfp.write(data)
 
 
 class PulpPushPlugin(PostBuildPlugin):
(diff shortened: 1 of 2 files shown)


TomasTomecek:
maybe we could `pip`install dockpulp while CI testing

twaugh:
Yes, once dockpulp is on PyPI.

TomasTomecek:
Well, you can install directly from git. This is what I do e.g.:

```
pip install git+https://github.com/TomasTomecek/dockpulp
```

twaugh:
@TomasTomecek I can't seem to get that to work.
https://travis-ci.org/DBuildService/atomic-reactor/jobs/69119235

.travis.yml
```diff
@@ -8,10 +8,10 @@ install:
   - "pip install -r requirements.txt"
   - "pip install -r tests/requirements.txt"
   - "pip install pytest-cov coveralls"
+  - "sudo pip install git+https://github.com/TomasTomecek/dockpulp"
```
TomasTomecek: is `sudo` needed here? when `pip` calls above are without it

twaugh: This is why I added `sudo`: https://travis-ci.org/DBuildService/atomic-reactor/jobs/69115173

TomasTomecek:
Do you guys think it would be possible do something like `setup.py install --wihout-installing-config-file`. Not sure if python packaging is capable of doing something like that.

twaugh:
I've filed #209 to track the issue preventing us from running the pulp test automatically. It needs a `dockpulp` change to its `setup.py`.

This PR is ready for reviewing now.

twaugh:
See above for proof that the test case now runs. :-)

PR #219, opened 2015-07-10
inject labels: allow inheriting labels from base

### LINKED ISSUE #212: RFE: read Architecture LABEL from image defined in FROM and define that LABEL for current one too
Base image has Architecture LABEL which can be reused in child layered images. This request is to inherit Architecture LABELs.

Proposed changes to atomic-reactor:
1. in the part where atomic-reactor rewrites FROM directive to ID of the image read LABEL of the image and save it for future use
2. in the part where Dockerfile is read check if LABEL is defined. If it isn't write new LABEL to the Dockerfile with the value from parent. If value of the Architecture differs fail the build (at the moment. unless we find some proper justification for having two different values as valid use case).
3. after the build return Architecture LABEL as part of the response

If LABEL wasn't present in the Dockerfile and was added later it should be possible to query this label in any child so the process above would work transparently for them too.

Related issue: https://github.com/release-engineering/koji-containerbuild/issues/3

Snippet of docker inspect on base image:
```
    [{
        "Architecture": "amd64",
        "Config": {
            "Labels": {
                "Architecture": "x86_64",
…
            },
```

TomasTomecek:
+ deny overwriting

Fixes #212

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -259,6 +259,7 @@ class DockerBuildWorkflow(object):
         self.builder = None
         self.build_logs = None
         self.built_image_inspect = None
+        self.base_image_inspect = None
 
         self.dont_pull_base_image = dont_pull_base_image
         self.pulled_base_images = set()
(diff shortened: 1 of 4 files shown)


jpopelka:
LGTM, I've just changed the new test to use pytest.parametrize (recent change https://github.com/DBuildService/atomic-reactor/commit/8517f39d21ec0263fb154d58e15ccbe4668f18ef)

TomasTomecek:
Thanks! Wasn't sure how to parametrize it.

TomasTomecek:
Tested and works great. Ready for merge from my PoV.

Issue #85, opened 2015-05-07
Proposal: redesign atomic-reactor


TomasTomecek:
For quite some time, I think that dock's CLI, code and architecture is getting a bit messy. This is my proposal how we could improve it. We'll have a single command for building images, which will be `build`.
### Build Types

`build` will have subcommands called **build types** (there will be a default build type). We'll probably start with just one build type which will precisely match the current behaviour. Build types will be very similar to what osbs-client has. Build types will be predefined recipes how to build an image (and what plugins should be used) — json templates. Every build type will have a set of its own arguments. Most build types will have similar arguments e.g. `--image`.

We could possibly even improve the json to something like:

``` json
{
  "input": {
    "type": "path",
    "args": {
        "path": "/here/is/my/dockerfile"
    },
    "source_repository": "..."
  },
  "output": {
    "tag": {
      "value": "...",
      "repositories": [...]
    }
  },
  "plugins": ...
```
### Methods

Methods will work the same:
1. `--method privileged`
2. `--method docker-host` (every occurrence of `host.?docker` will be changed)
3. `--method local` (`dock local` could be symlink to `dock build --method local`)
### Input

When you specify build json on your own, build type will be ignored. You may be still able to override values in your json from CLI.

`--input` will be direct mapping for input plugins (getting build json from env var, path,...). You'll be able to override this via CLI.

If you don't care about code, you can stop reading.

In code, I would like to have a lot of changes too:
- start getting rid of `DockerBuildWorkflow` and moving logic to plugins
- `InsideBuilder` should be as light as possible
- I would like to see more classes rather than having one class accepting 15 arguments
- we need to figure what will be the best interface to configure plugins from CLI (`--substitute` might be enough)

CC @goldmann @mmilata @twaugh @jpopelka @l-d-j @adamruzicka

I really would like to hear your feedback on this.

adamruzicka:
Will dock still have ```create-build-image``` and ```inside-build``` comands or will they be converted to build types?

Methods and Input looks good to me.
+1 for moving logic to plugins, making ```InsideBuilder``` lighter and splitting to more classes

twaugh:
Looks good in general. How would plugins communicate with one another? Currently they store state in the `DockerBuildWorkflow` instance, don't they?

TomasTomecek:
@adamruzicka 
[...]

PR #220, opened 2015-07-12
Move pulling base image into a plugin

### LINKED ISSUE #218: Move pulling base image into a plugin
@pbabinca's proposal for scratch builds suggests that if there is `FROM xyz` in the Dockerfile then we should first try pulling `scratch-xyz` as a base image.

Moving the pulling logic into a plugin would make the implementation less messy since we need to know if the build is a scratch build, which can be easily passed as a parameter to the plugin.

mmilata:
For backwards compatibility the plugin is automatically enabled when not explicitly listed - we can remove that once we're sure everyone's osbs-client is updated.

Closes #218.


jpopelka:
@mmilata your change LGTM, please review mine addition, thanks

TomasTomecek:
This is definitely something we should aim for. The only problem I have with current implementation is the code in `inner.py`. I think this is where we should probably introduce build types. Because I think that configuration of pulling base image should be as simple as possible. We should not force users to write json.

mmilata:
@jpopelka Thank you! I'll try running the tests with `NOMOCK=1` next time.

@TomasTomecek Right, I didn't realize that when the code that automatically enables the plugin is removed it becomes impossible to start a build from API without providing a build json. Does this mean we need to solve #85 before moving the base image code to a plugin?

mmilata:
I've fixed the issue with cli/api builds based on offline discussion with @TomasTomecek. I also removed the `change_source_registry` plugin which doesn't do anything when pulling base image happens in a plugin.

twaugh:
Would be nice to see a couple of tests in `tests/plugins/test_pull_base_image.py` for the new plugin.

twaugh:
Oh, ignore me, I missed them before.

Looks good to me.

mmilata:
Found a little bug, pushing once again ...

mmilata:
Rebased.

TomasTomecek:
LGTM, nice work!

PR #222, opened 2015-07-14
Set up .travis.yml to auto-release to PyPI

bkabrda:
This should release every tagged version to PyPI. I see no way to test this, so I guess we just have to hope that I did it right :)


jpopelka:
LGTM, we'll see with next tag :-)

Issue #110, opened 2015-05-21
introduce new plugin argument: `run_when_build_fails`


TomasTomecek:
Suggested by @goldmann

goldmann:
I think the name of the variable should be different, maybe `run_always`?

twaugh:
This ties in with the ability of a plugin to know whether the build has failed or not.

In #229 I've suggested a new class of plugin, 'exit', which will be run after everything else and even in the event of failure. Its constructor takes a bool indicating build success/failure.

What do you think of this approach? Should I redo it as `run_always`? (Harder to see how to implement that.)

jpopelka:
Looking at https://github.com/projectatomic/atomic-reactor/pull/229#issuecomment-125263105 I think this should have already been closed, right ?

twaugh:
Yes, thanks for spotting it.

PR #229, opened 2015-07-16
Ability to run plugin even on failure

twaugh:
Now plugins can determine whether the build has already been marked as failed.

This will be needed when we have BuildConfigs to prune.

### DIFF
--- a/atomic_reactor/plugin.py
+++ b/atomic_reactor/plugin.py
@@ -277,6 +277,27 @@ class PostBuildPluginsRunner(BuildPluginsRunner):
         self.plugins_results = workflow.postbuild_results
         super(PostBuildPluginsRunner, self).__init__(dt, workflow, 'PostBuildPlugin', plugins_conf, *args, **kwargs)
 
+    def create_instance_from_plugin(self, plugin_class, plugin_conf):
+        instance = super(PostBuildPluginsRunner, self).create_instance_from_plugin(plugin_class, plugin_conf)
+        if isinstance(instance, ExitPlugin):
+            logger.error("running exit plugin '%s' as post-build plugin", plugin_class.key)
+
+        return instance
+
+
+class ExitPlugin(PostBuildPlugin):
+    """
+    Plugin base class for plugins which should be run just before
+    exit. It is flavored with DockerTasker and DockerBuildWorkflow instances.
+    """
+
+
+class ExitPluginsRunner(BuildPluginsRunner):
+    def __init__(self, dt, workflow, plugins_conf, *args, **kwargs):
+        logger.info("initializing runner of exit plugins")
+        super(ExitPluginsRunner, self).__init__(dt, workflow, 'ExitPlugin',
+                                                plugins_conf, *args, **kwargs)
+
 
 class InputPlugin(Plugin):
 
--- a/atomic_reactor/plugin.py
+++ b/atomic_reactor/plugin.py
@@ -146,6 +146,9 @@ class PluginsRunner(object):
         plugin_instance = plugin_class(**plugin_conf)
         return plugin_instance
 
+    def on_plugin_failed(self):
+        pass
+
     def run(self):
         """
         run all requested plugins
@@ -184,6 +187,7 @@ class PluginsRunner(object):
                 logger.debug(traceback.format_exc())
                 if not plugin_can_fail:
                     failed_msgs.append(msg)
+                    self.on_plugin_failed()
                 else:
                     logger.info("error is not fatal, continuing...")
                 plugin_response = ex
@@ -209,6 +213,9 @@ class BuildPluginsRunner(PluginsRunner):
         self.dt = dt
         self.workflow = workflow
         super(BuildPluginsRunner, self).__init__(plugin_class_name, plugins_conf, *args, **kwargs)
+
+    def on_plugin_failed(self):
+        self.workflow.plugin_failed = True
 
     def _translate_special_values(self, obj_to_translate):
         """
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -253,6 +253,8 @@ class DockerBuildWorkflow(object):
         self.exit_plugins_conf = exit_plugins
         self.prebuild_results = {}
         self.postbuild_results = {}
+        self.build_failed = False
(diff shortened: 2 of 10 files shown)


TomasTomecek:
Shouldn't there be a code in `inner.py` which sets the variable appropriately (once the build has finished)? Also, wouldn't it make more sense to name it `build_has_failed`?

Maybe I just didn't get the patch.

twaugh:
Yes, `build_has_failed` probably makes more sense.

The point of this change is to allow a post-build plugin to decide how to behave based on whether this build has failed. That's why we have to set the variable as soon as a plugin fails that is not allowed to fail (https://github.com/projectatomic/atomic-reactor/commit/701bf3abdd20a8b11811697a6d53626c7f0ed7cf#diff-6efbaa8f7bea401e03542bd227209afbR218).

Setting it after all the plugins have run would mean no plugin gets to use it.

For an example of this in action, see https://github.com/projectatomic/atomic-reactor/pull/230.

TomasTomecek:
I'm sorry I still don't get it, let me show you how would I implement it.

TomasTomecek:
This is what I would expect this PR to contain: https://github.com/TomasTomecek/atomic-reactor/commit/990fcfb0c9144fee0dbe1e216b22e77518d78ff1

twaugh:
OK, I see what you mean now. Yes, it needs to do that.

However, looking more at that code I see we also need to have a type of plugin that *always* runs, even if (for example) a pre-build plugin failed, or the Docker build itself failed. Maybe it could be called `CleanupPlugin`.

I'll look at adding this.

TomasTomecek:
https://github.com/projectatomic/atomic-reactor/issues/110

twaugh:
Oh, didn't see that.

atomic_reactor/plugin.py
```diff
@@ -210,6 +214,9 @@ def __init__(self, dt, workflow, plugin_class_name, plugins_conf, *args, **kwarg
         self.workflow = workflow
         super(BuildPluginsRunner, self).__init__(plugin_class_name, plugins_conf, *args, **kwargs)
 
+    def on_plugin_failed(self):
+        self.workflow.build_has_failed = True
```
TomasTomecek: what does this mean?

TomasTomecek:
Once we merge this, we should then transform these plugins to be exit plugins:
 * `remove_built_image` (removing base image)
 * `store_metadata_in_osv3`
 * `store_logs_to_file`

atomic_reactor/plugin.py
twaugh: It means that `build_inside()` will raise an exception.

twaugh:
We need to decide which approach is best: exit plugins, or `run_always` plugin arguments.

twaugh:
> Once we merge this, we should then transform these plugins to be exit plugins:

Oh, we'll need to give exit plugins access to the workflow instance then.

twaugh:
I've changed the ExitPlugin argument from a `bool` to a `DockerBuildWorkflow` instance and squashed some commits to make the history nicer. I also changed those plugins to be exit plugins.

twaugh:
Oh, remove_built_image needs the tasker as well.

twaugh:
I've added a test case to show how exit plugins work. It also has machinery for checking the case where the first of two prebuild plugins fails, but it's commented out (last line). Currently the second plugin will still run... is that what we want?

twaugh:
(coveralls failure is due to pulp test no longer getting run, see https://github.com/release-engineering/dockpulp/pull/16)

atomic_reactor/plugin.py
TomasTomecek: I'm sorry I still don't follow. I thought that setting `build_has_failed` makes only sense in `inner.py` where we figure out if the build successful.

twaugh: This is about whether atomic-reactor succeeded or failed, not just whether the docker build did.

Some plugins can cause the entire run to be a failure, and there needs to be a hook to allow different behaviour in that case.

twaugh: Let me explain another way. When we set up automated rebuilds, we only want to rebuild from Builds (OpenShift Build objects) that built successfully in the first place.

That's what this instance variable is intended to be: will this Build get marked by OpenShift as completed or failed?

TomasTomecek: Right. Now I finally get it.

So the variable will reflect two states:
1. `docker build` has failed
2. one of the plugins has failed

Wouldn't it make sense rather use two variables for that?

twaugh: From the point of view of an exit plugin wanting to know whether everything has gone to plan so far, though, it would be nice for it to have a single thing to check.

twaugh:
I've changed the plugin failure hook so that it sets `plugin_failed` instead of `build_failed`, and added a `build_process_failed` property. CI failures are my fault but elsewhere :-(

TomasTomecek:
Very nice! Ready for merge?

twaugh:
Actually, no, I think the conversion of some plugins to exit plugins isn't complete. I'll fix that up.

twaugh:
Previously, an older osbs-client would try running what are now exit plugins as postbuild plugins, and they would not run. Now ExitPlugin inherits from PostBuildPlugin, and the plugins runner logs an error if an exit plugin is being run as a post-build plugin.

The upshot is that we now provide more compatibility for older osbs-client versions. Note: we still need to update osbs-client to actually run the exit plugins as exit plugins not post-build plugins.

This is ready for merge, with one open question: is this test valid? https://github.com/projectatomic/atomic-reactor/commit/0f218ecb8b3d2d98880dc1085406326a3e09a5d2#diff-1d4c7d01e22fd1091edb1a829562521cR261

twaugh:
Let's leave the issue of running all plugins of a certain type even when one fails until after this PR is merged. For now, I've added a test to verify the current behaviour.

Ready for merge.

twaugh:
Fixes #110.

TomasTomecek:
I thought this is already merged. Can you rebase please? (I should have merged this first, sorry about that)

twaugh:
Rebased and fixed a conflict caused by the movement of `test_rpmqa_plugin()`.

PR #230, opened 2015-07-17
[do not merge yet] Exit plugin for removing BuildConfig on failure

twaugh:
Do not merge yet.

This plugin will be used when osbs-client creates BuildConfigs instead of Builds. It uses anticipated support in osbs-client for deleting BuildConfigs. As osbs-client doesn't have API for BuildConfigs yet, some parts of this plugin are subject to change.

### DIFF
--- a/atomic_reactor/plugins/exit_remove_buildconfig.py
+++ b/atomic_reactor/plugins/exit_remove_buildconfig.py
@@ -1,0 +1,84 @@
+"""
+Copyright (c) 2015 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import unicode_literals
+
+import json
+import os
+
+from osbs.api import OSBS
+from osbs.conf import Configuration
+
+from atomic_reactor.plugin import ExitPlugin
+
+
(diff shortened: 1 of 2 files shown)


jpopelka:
LGTM

TomasTomecek:
I'm not sure I follow. Why do we want to remove buildconfigs when builds, created from those buildconfigs, fail?

twaugh:
I was thinking we'd only want to have automatic rebuilds from builds that we know succeeded.

twaugh:
In other words, all BuildConfigs will be either for Builds currently in progress or else poised ready to start automated Builds for images we built previously.

bkabrda:
I'm not sure we should remove BuildConfigs. Maybe the rebuild failed because of a problem in base image and we want to rebuild automatically when the base image is fixed...

twaugh:
That's true.

But maybe we should delete them if the build failed and this is not an automated rebuild but a build submitted by a user?

bkabrda:
I have to say I'd rather keep all BuildConfigs rather then remove some. It's just a gut feeling, really... but *if* the only price for keeping all BuildConfigs is some failed builds during autorebuild, then I think it's ok.

twaugh:
OK, agreed.

PR #234, opened 2015-07-22
More specfile improvements

bkabrda:
- make atomic-reactor package depend on python3-atomic-reactor when building with py3
- make atomic-reactor binaries link to py3 versions when building with py3


jpopelka:
LGTM

PR #236, opened 2015-07-23
Create post_compress plugin, fixes #191

### LINKED ISSUE #191: gzip/xz container image tarball
Save disk space and speedup data transfers by gzipping container image tarball.

[Koji buildsystem uses xz](https://git.fedorahosted.org/cgit/koji/tree/builder/kojid#n3201) to compress base (container) images which it creates If atomic reactor is used together with koji it would make a sense to use same format for layered tarballs too.

I *believe* it does make a sense to compress this by atomic-reactor as we will gain most advantages as described in first paragraph there.

bkabrda:
Don't merge yet.

There are still modifications that need to be done in other plugins, like `pulp_push`, `cp_built_image_to_nfs` or `store_metadata_in_osv3`. One thing I'm struggling with is whether all of these plugins should first check for presence of compressed image and then for presence of plain exported image if compressed image is not found. This seems to be pretty fragile.
More general way of solving this would be introducing a list (in `DockerBuildWorkflow`) that holds metadata dicts that correspond to all steps done to an exported image. Every plugin that would do something with the exported image would just add a dict with metadata to the end of this list. Right now, it could contain metadata for `exported_squashed_image` and `exported_compressed_image`. This way, plugins like `pulp_push` would just work with the last metadata structure and wouldn't have to care how many steps have actually changed the exported image. Also, 3rd party plugins wouldn't need to touch `DockerBuildWorkflow`, they'd just append a new dict to the end of that list, so everything would be quite dynamic.

Does this sound ok? If so, I'll create another commit that implements it and add it to this PR. If someone thinks of something better, feel free to propose your idea!


bkabrda:
So I added another commit that does what I proposed in my previous comment. This is now ready for review, but I'm still testing to see whether everything works now.

bkabrda:
Also, we may want to rename `PulpPushPlugin` constructor argument `load_squashed_image` to more general `load_exported_image`, since generally the exported image can be processed by various plugins, not just squashed; but I'm thinking that this may be an unnecessary breakage for users... Not sure.

twaugh:
Maybe the `PulpPushPlugin` constructor could accept both `load_squashed_image` and `load_exported_image` for the time being, but issue a warning if `load_squashed_image` is used.

This approach, with the metadata list, looks good to me.

bkabrda:
@twaugh good point. I'll do that and add another commit.

twaugh:
LGTM

bkabrda:
I just added a simple test case that tests the basic functionality of the plugin. Should be ok to merge now.

bkabrda:
One thing to note: this PR removes the auto compression done by `pulp_push` plugin, which may not be what we want. I'll try to investigate if there's a sane way to automatically invoke `compress` plugin before `pulp_push`, so that we always push compressed tarball without changing the default build jsons.
So don't merge yet, more commits coming on Monday (probably) :)

mmilata:
Looks good - why store all the exported image metadata, though? For debugging or auditing purposes? The current code is always working with the last one.

bkabrda:
> Looks good - why store all the exported image metadata, though? For debugging or auditing purposes? The current code is always working with the last one.

Yes, debugging and auditing was my primary reason for not throwing away these.

bkabrda:
The more I think about this, I'm in favor of changing the default build jsons. It seems to me that it's better to be explicit about the compression (and explicitly select compression type). Making plugins use and depend on each other this way feels counter-intuitive and may be pretty hard to maintain in the long run. Thoughts @TomasTomecek @twaugh @mmilata?

twaugh:
Something to keep in mind when we address #85.

I'm happy with having the default build JSON be explicit about compression.

mmilata:
@bkabrda agreed.

My PR #220 has similar problem which I tried to solve by introducing code that automatically inserts the plugin, and which can be removed after some time when clients have been upgraded: https://github.com/projectatomic/atomic-reactor/blob/pull-base-image-plugin/atomic_reactor/plugins/input_osv3.py#L45 Haven't got any feedback so far, so not sure if it's a good idea.

twaugh:
Oh, *that*'s where I saw it.

bkabrda:
So I think this is good to merge. I'll open a bug for updating default build json when this is closed.

PR #237, opened 2015-07-23
Allow koji-related testing when koji module not installed

twaugh:


twaugh:
This is what I meant earlier @pbabinca - now the `tests/plugins/test_koji.py` test runs.

TomasTomecek:
Could `tests/koji/__init__.py` be mocked with flexmock? If not, I'm fine with merging this. @bkabrda can you take a look please?

bkabrda:
AFAIK there's no way to improve this with flexmock, I would do it like this as well. So LGTM.

twaugh:
It seems like you can't `flexmock()` anything that doesn't already exist in some form. The idea here is to make it exist in the most minimal form, and `flexmock()` it to use what we want while testing.

TomasTomecek:
LGTM then

PR #238, opened 2015-07-24
[do not merge] Add support for scratch builds

mmilata:
Adds `scratch_build` parameter to `push_to_pulp` and `tag_by_labels` plugins. When set to true, `tag_by_labels` prefixes the tags with *scratch-*.

The `push_to_pulp` doesn't really do anything now due to release-engineering/dockpulp#18 - once/if that is fixed it will change the pulp repo prefix to *scratch-*.

Another missing piece is the use of *scrach-base* base image (if it's present in parent docker repository) when *base* is specified as the base image in Dockerfile. This depends on #220.

@pbabinca, does this look alright to you?

### DIFF
--- a/atomic_reactor/plugins/post_push_to_pulp.py
+++ b/atomic_reactor/plugins/post_push_to_pulp.py
@@ -30,7 +30,7 @@ class PulpUploader(object):
     KEY = 'pulp.key'
 
     def __init__(self, workflow, pulp_instance, filename, log, pulp_secret_path=None, username=None,
-                 password=None):
+                 password=None, scratch_build=False):
         self.workflow = workflow
         self.pulp_instance = pulp_instance
         self.filename = filename
@@ -39,6 +39,7 @@ class PulpUploader(object):
         # U/N & password has bigger prio than secret cert
         self.username = username
         self.password = password
+        self.scratch_build = scratch_build
 
     def _check_file(self):
         # Sanity-check image
@@ -112,7 +113,9 @@ class PulpUploader(object):
             repos_tags_mapping[repo].setdefault("tags", [])
             repos_tags_mapping[repo]["tags"].append(image.tag)
         self.log.info("repo_tags_mapping = %s", repos_tags_mapping)
-        task_ids = p.push_tar_to_pulp(repos_tags_mapping, self.filename)
+
+        kwargs = {'repo_prefix': 'scratch-'} if self.scratch_build else {}
+        task_ids = p.push_tar_to_pulp(repos_tags_mapping, self.filename, **kwargs)
 
         self.log.info("waiting for repos to be published to crane")
         p.watch_tasks(task_ids)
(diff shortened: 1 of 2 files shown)


pbabinca:
I'm not familiar with the atomic-reactor code. Description of the issue seems to correspond with the code though.

atomic_reactor/plugins/post_push_to_pulp.py
```diff
@@ -112,7 +113,9 @@ def push_tarball_to_pulp(self, image_names):
             repos_tags_mapping[repo].setdefault("tags", [])
             repos_tags_mapping[repo]["tags"].append(image.tag)
         self.log.info("repo_tags_mapping = %s", repos_tags_mapping)
-        task_ids = p.push_tar_to_pulp(repos_tags_mapping, self.filename)
+
+        prefix = "scratch-" if self.scratch_build else "redhat-"
```
twaugh: Should this be hard-coded like that?

mmilata: The `redhat-` prefix is hardcoded in dockpulp and is used when no `repo_prefix` is given. You've got a point though - perhaps the plugin can accept the prefix as a parameter instead of the boolean `scratch_build`.

twaugh: Oh, I see. Then maybe we could do something like:
```
kwargs = {'repo_prefix': 'scratch-'} if self.scratch_build else {}
task_ids = p.push_tar_to_pulp(repos_tags_mapping, self.filename, **kwargs)
```

mmilata:
@twaugh good idea, updated.

twaugh:
Looks good so far. During rebase you'll need to account for `post_push_to_pulp.py` being renamed to `exit_push_to_pulp.py` (sorry).

jpopelka:
LGTM

atomic_reactor/plugins/post_push_to_pulp.py
```diff
@@ -112,7 +113,9 @@ def push_tarball_to_pulp(self, image_names):
             repos_tags_mapping[repo].setdefault("tags", [])
             repos_tags_mapping[repo]["tags"].append(image.tag)
         self.log.info("repo_tags_mapping = %s", repos_tags_mapping)
-        task_ids = p.push_tar_to_pulp(repos_tags_mapping, self.filename)
+
+        kwargs = {'repo_prefix': 'scratch-'} if self.scratch_build else {}
```
TomasTomecek: Am thinking if this shouldn't be part of configuration in osbs-client.

TomasTomecek:
I think we can revisit this once scratch builds will be on roadmap.

gbraad:
scratch builds are quite common when using containers for Go. Any estimation when this will be possible?

vrutkovs:
@gbraad this part has been moved to [osbs-client](https://github.com/projectatomic/osbs-client/pull/452) - if the configuration section has 'scratch=True' BuildConfig won't be updated and a new Build object will be created instead

Issue #239, opened 2015-07-27
rpm-packages sometimes contains string with exception


pbabinca:
When plugin `PostBuildRPMqaPlugin` doesn't finish successfully (raise an exception) `rpm-packages` label contain string describing this exception. Because both proper output of that plugin and this error are both strings it is hard to determine if the content of the label is valid.

Please provide a way to determine if the run was successful. E.g. content of rpm-packages would be either `None` or empty string and the reason why the plugin failed would be available somewhere else.

The content of rpm-packages was something like this (it was reconstructed from some `split()` which ate spaces):
```
plugin 'all_rpm_packages' raised an exception: 'ReadTimeout(ReadTimeoutError("UnixHTTPConnectionPool(host='localhost', port=None): Readtimedout.(readtimeout=60)"))'
```

Note: I'm using `BuildResponse().get_rpm_packages()` call of `osbs-client` to get this key.

twaugh:
So, the `all_rpm_packages` plugin raised an exception, the plugins runner stored a failure message in `plugin_results`, and the `store_metadata_in_osv3` dutifully copied that message into the metadata.

I think `PluginsRunner.run()` needs to store plugin results more carefully. Perhaps it could store the actual exception rather than a description of it. Then the metadata plugin can skip it.

TomasTomecek:
Should we increase the timeout? Do you know why it took docker to run the command so long?

pbabinca:
I don't have more data about this.

TomasTomecek:
Since #240 was merged, can we close this?

twaugh:
Yes.

PR #240, opened 2015-07-27
Don't store plugin exceptions in metadata (#239)

twaugh:
With this change, we store the actual exception a plugin raised, rather than a description of it. This allows the `store_metadata_in_osv3` plugin to handle that differently.

I've moved the `rpmqa` test-case out into a separate module and added a failure test-case as well. I also added a failure test-case for the metadata plugin.


twaugh:
Fixes #239.

PR #243, opened 2015-07-29
New check_rebuild pre-build plugin

twaugh:
This plugin will be needed to direct other plugins to behave differently depending on whether this is the first build or an automated rebuild.


jpopelka:
LGTM

PR #251, opened 2015-08-03
update spec file to reflect post_store/exit_store changes and tests

maxamillion:


atomic-reactor.spec
```diff
@@ -202,7 +202,8 @@ cp -a docs/manpage/atomic-reactor.1 %{buildroot}%{_mandir}/man1/
 %{python2_sitelib}/atomic_reactor/cli
 %{python2_sitelib}/atomic_reactor/plugins
 %exclude %{python2_sitelib}/atomic_reactor/plugins/pre_koji.py*
-%exclude %{python2_sitelib}/atomic_reactor/plugins/post_store_metadata_in_osv3.py*
+%exclude %{python2_sitelib}/atomic_reactor/plugins/exit_store_metadata_in_osv3.py*
+%exclude %{python2_sitelib}/tests
```
TomasTomecek: hmmm, I didn't know we ship tests

TomasTomecek:
Thank you!

atomic-reactor.spec
bkabrda: Yeah, the problem is on this line: https://github.com/projectatomic/atomic-reactor/blob/master/setup.py#L54

Any time we include a subdirectory with python files under `tests/`, we also need to add it to `setup.py`, otherwise it will be included in the tarball. Currently, it's `tests/koji` which is causing this.

On my own projects, I usually do this differently - I just explicitly list packages to include without using `find_packages` and then include the tests in tarball by listing them in `MANIFEST.in`. This way, the tests are included in the tarball, which is usually desirable, but they won't get installed in sitelib.

TomasTomecek: sounds really good! let's do that

PR #252, opened 2015-08-03
add python-backports-lzma for buildroot backports-lzma build inside container

maxamillion:


bkabrda:
Why not `yum install python-backports-lzma`?

maxamillion:
That makes a lot more sense. Lets do that instead.

bkabrda:
LGTM

Issue #249, opened 2015-08-03
Changes for CheckRebuildPlugin to satisfy osbs-client#181


bkabrda:
The `CheckRebuildPlugin` should:
- be renamed to `CheckAndSetRebuildPlugin`
- check if this is autorebuild (received label value is `"false"` (string, since Openshift can't use booleans as label values); if so, it should <do some stuff>
- if not, then set the received label to received value (which is always going to be `"true"` (again, string), when invoked through osbs-client).

Note: it's necessary to check `build_json["metadata"]["labels"]`, not just `build_json["metadata"]` for the labels!

twaugh:
Now we need `OSBS.set_label_on_buildconfig()` :-)

PR #253, opened 2015-08-03
Check and set rebuild

twaugh:
This PR implements the changes for #249. Note that it depends on API in OSBS which does not exist yet: `set_labels_on_build_config()`. See https://github.com/projectatomic/osbs-client/pull/189 for that.


bkabrda:
Not sure if I understand this correctly, but can't you use http://has207.github.io/flexmock/user-guide.html#automatically-checked-expectations instead of the `Collect` class? Flexmock will check if the method has been called with proper arguments for you.

twaugh:
I couldn't get that to work to start with. I see now that it was because I was calling `flexmock` twice for the same method, once in the `(..., function=....)` style.

bkabrda:
LGTM now. Let's merge this after projectatomic/osbs-client#189 is merged.

PR #266, opened 2015-08-12
New bump_release pre-build plugin

twaugh:
Here is a plugin for incrementing the Release label on automatic rebuild.

The push URL can be configured separately to the fetch URL used for the build. The `GitSource` provider's repository is modified in-place, the push URL set, and 'git push' run.

For authentication, currently this is achieved with Kerberos using the kernel keyring's TGT. This works by making sure the gssproxy package is installed and having git call out to an ssh process. I haven't managed to get it to work using the pygit2 `remote.push()` method.

For other authentication mechanisms, we would need another secret to be provided. I think that requires having a service account configured, which is out of scope for this pull request.

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -70,8 +70,6 @@ class BumpReleasePlugin(PreBuildPlugin):
     key = "bump_release"
     can_fail = False  # We really want to stop the process
 
-    commit_message = "Bumped release for automated rebuild"
-
     def __init__(self, tasker, workflow,
                  git_ref,
                  author_name, author_email,
@@ -97,16 +95,11 @@ class BumpReleasePlugin(PreBuildPlugin):
         self.git_ref = git_ref
         self.author_name = author_name
         self.author_email = author_email
-        self.committer_name = committer_name
-        self.commiter_email = committer_email
-        if committer_name is None:
-            self.committer_name = author_name
-        if committer_email is None:
-            self.committer_email = author_email
-
+        self.committer_name = committer_name or author_name
+        self.committer_email = committer_email or author_email
         self.push_url = push_url
-        if commit_message is not None:
-            self.commit_message = commit_message
+        self.commit_message = (commit_message or
+                               "Bumped release for automated rebuild")
 
     def bump_local_file(self, parser, label_key, next_release):
         """
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -1,0 +1,270 @@
+"""
+Copyright (c) 2015 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import unicode_literals
+
+import shlex
+try:
+    # py3
+    from shlex import quote
+except ImportError:
+    from pipes import quote
+
+import os
+from pygit2 import init_repository, Signature
+import subprocess
+
+from atomic_reactor.plugin import PreBuildPlugin
+from atomic_reactor.source import GitSource
+from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
+from atomic_reactor.constants import PY2
+from dockerfile_parse import DockerfileParser
+
+
+def shlex_splits(value):
+    if PY2 and isinstance(value, unicode):
+        value = value.encode('utf-8')
+        splits = shlex.split(value)
+        return [split.decode('utf-8') for split in splits]
+    else:
+        return shlex.split(value)
+
+
+class BumpReleasePlugin(PreBuildPlugin):
+    """Git branch management plugin
+
+    For rebuilds, push a new commit incrementing the Release label.
+
+    For initial builds, verify the branch is at the specified commit
+    hash.
+
+    When this plugin is configured by osbs-client, the Build's source
+    git ref is actually the branch (from --git-branch), not the
+    original SHA-1. The SHA-1 specified by --git-commit is stored in
+    the configuration for this plugin.
+
+    Example configuration:
+
+    {
+      "name": "bump_release",
+      "args": {
+        "git_ref": "12345678....",
+        "author_name": "OSBS Build System",
+        "author_email": "root@example.com"
+      }
+    }
+
+    Additional optional arguments:
+    - committer_name
+    - committer_email
+    - commit_message
+    - push_url
+
+    """
+
+    key = "bump_release"
+    can_fail = False  # We really want to stop the process
+
+    commit_message = "Bumped release for automated rebuild"
+
+    def __init__(self, tasker, workflow,
+                 git_ref,
+                 author_name, author_email,
+                 committer_name=None, committer_email=None,
+                 commit_message=None,
+                 push_url=None):
+        """
+        constructor
+
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param git_ref: str, commit hash expected on first build
+        :param author_name: str, name to use for git commits
+        :param author_email: str, email address for git commits
+        :param committer_name: str, name to use for git commits (else author's)
+        :param committer_email: str, email address for git commits (else
+                                     author's)
+        :param commit_message: str, git commit message
+        :param push_url: str, URL for push
+        """
+        # call parent constructor
+        super(BumpReleasePlugin, self).__init__(tasker, workflow)
+        self.git_ref = git_ref
+        self.author_name = author_name
+        self.author_email = author_email
+        self.committer_name = committer_name
+        self.commiter_email = committer_email
+        if committer_name is None:
+            self.committer_name = author_name
+        if committer_email is None:
+            self.committer_email = author_email
+
+        self.push_url = push_url
+        if commit_message is not None:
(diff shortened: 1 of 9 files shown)


atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -0,0 +1,263 @@
...
+        if committer_name is None:
+            self.committer_name = author_name
+        if committer_email is None:
+            self.committer_email = author_email
+
+        self.push_url = push_url
+        if commit_message is not None:
+            self.commit_message = commit_message
+
+    def bump_local_file(self, parser, label_key, next_release):
```
bkabrda: Wouldn't it be better to rewrite this function using dockerfile_parser? You could load the Dockerfile using that, then just traverse the loaded structure to find the proper label, change it and serialize the file again (note: it even covers cases when people use `Label` instead of `LABEL` etc).

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -0,0 +1,263 @@
...
+    def bump(self, repo, branch):
+        # Look in the git repository
+        remote = repo.remotes['origin']
+        if self.push_url:
+            repo.config['push.default'] = 'simple'
+            remote.push_url = self.push_url
+            remote.save()
+
+        # Bump the Release label
+        label_key = 'Release'
```
bkabrda: Are we sure that this will always be `Release`? What if users provide `release` or `RELEASE`?

requirements.txt
```diff
@@ -1,2 +1,3 @@
 docker-py
 dockerfile-parse
+#pygit2
```
bkabrda: Why is this commented out? It seems that the plugin does import pygit2.

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -0,0 +1,263 @@
...
+        cmd = ['/usr/bin/git', 'push', 'origin']
+        try:
+            output = subprocess.check_output(cmd, stderr=subprocess.STDOUT,
+                                             cwd=repo.workdir)
+            self.log.debug("git (success): %s", output)
+        except subprocess.CalledProcessError as exc:
+            self.log.error("git (%d): %s", exc.returncode, exc.output)
+            raise
+        except AttributeError:
+            # py2.6
```
bkabrda: So why not just use the code that is in the except clause for all Python versions? IMO it works on all Python versions we care about, so I don't see the need for the try clause above.

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -0,0 +1,263 @@
...
+
+        if self.workflow.build_process_failed:
+            self.log.info("Build already failed, not incrementing release")
+            return
+
+        # Ensure we can use the git repository already checked out for us
+        source = self.workflow.source
+        assert isinstance(source, GitSource)
+        repo = init_repository(source.get())
+        branch = repo.lookup_branch(source.git_commit)
```
bkabrda: What if user passed an actual commit instead of a branch name as `git_commit`? That can theoretically happen, AFAICS. (Isn't that what Koji will do? https://github.com/release-engineering/koji-containerbuild/blob/master/koji_containerbuild/plugins/builder_containerbuild.py#L335)

atomic_reactor/plugins/pre_bump_release.py
twaugh: It's using DockerfileParser already. But that doesn't let us adjust a label and write it out again, so we have to reconstruct it ourselves... and we have to deal with both 'LABEL' forms.

The `'LABEL'` is used for comparing against `insn['instruction']`, which is guaranteed to be upper-case.

bkabrda: Don't know how I missed this. I guess this is ok, then.

atomic_reactor/plugins/pre_bump_release.py
twaugh: Good point. So, the `assert_labels` plugin defaults to requiring `Release` but is configurable: https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/pre_assert_labels.py.

I should probably make the name of the label configurable here.

atomic_reactor/plugins/pre_bump_release.py
twaugh: You missed it because the code is much larger than it looks like it ought to be when it's getting so much help. :-)

Perhaps some of this could be moved to `dockerfile_parse`.

requirements.txt
twaugh: There's a story here, and I forgot to tell it in the description.

Yes, in theory, we just uncomment that line. The plugin requires pygit2. However, if we do that, Travis CI will try to install it with `pip install`, and that will require building it with gcc which will fail. It will fail even if we install the dev packages it requires.

Of course, we could avoid that by simply installing the Ubuntu packaged version in the `.travis.yml` file... except pygit2 isn't packaged for Ubuntu(!) so that isn't possible after all.

So in short it's commented out because Travis CI's docker container images are based on Ubuntu.

atomic_reactor/plugins/pre_bump_release.py
twaugh: Fair point, will re-code.

atomic_reactor/plugins/pre_bump_release.py
twaugh: Yes, that's what will happen: the osbs client will be run with `--git-commit (commit-hash) --git-branch (branch-name)`. However, the client will cunningly configure the BuildConfig so that its source git ref will be the *branch* name, not the commit hash, and the `bump_release` plugin will be told the original commit hash instead.

To see how that works, take a look at the osbs-client branch tracking those changes: https://github.com/projectatomic/osbs-client/compare/master...enable-rebuild-triggers
It's in `ProductionBuild.render()`.

atomic_reactor/plugins/pre_bump_release.py
twaugh: Filed as https://github.com/DBuildService/dockerfile-parse/issues/2.

twaugh:
Updated with suggestions from comments.

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -0,0 +1,254 @@
...
+        os.environ['GIT_SSH_COMMAND'] = ssh_command
+
+        # This doesn't seem to work:
+        #   remote.push([branch.name])
+        # because it uses libssh rather than /usr/bin/ssh and so krb5
+        # auth fails.
+        # Instead, run the git command to do it
+        cmd = ['/usr/bin/git', 'push', 'origin']
+        with open('/dev/null', 'r+') as devnull:
+            p = subprocess.Popen(cmd,
```
twaugh: I should check whether this tries closing all open file descriptors.

twaugh: UPDATE: verified that it does not try to close all file descriptors, so it's fine.

atomic_reactor/plugins/pre_bump_release.py
bkabrda: Great, thanks! I'm fine with keeping the code here until dockerfile-parse can help us handle it.

atomic_reactor/plugins/pre_bump_release.py
bkabrda: Thinking about this further, we should really expect just `Release`. Label names are case sensitive, so `Release` is a different label than `RELEASE` or `rELELASE` (just tried that).
So if someone doesn't use `Release` (precisely this way), I'd call it an error in the provided Dockerfile.

twaugh: OK, I'll remove the configuration option then.

requirements.txt
bkabrda: So I did some googling and it seems someone has solved this problem: https://github.com/uniphil/venvgit2 - it's not a very clean solution, but for tests in travis, it should be ok to use it.
I think if you add `pip install venvgit2` to `.travis.yml` before running `pip install -r requirements.txt` (with `pygit2` uncommented), it should work. Worth a try :)

twaugh: Great find! Unfortunately it fails on Python 2.6: https://travis-ci.org/projectatomic/atomic-reactor/jobs/75398810

atomic_reactor/plugins/pre_bump_release.py
bkabrda: Cunning! :)
This probably deserves a comment in code and we should also document it in osbs-client docs about build process (I added a comment to that PR that asks for this).
So IIUC, during the user-triggered build (the first one), the plugin checks that last commit in the `--git-branch` is the user-given `--git-commit`? Very nice!

twaugh: Yes, that's exactly it.

requirements.txt
bkabrda: I just filed a PR for Python 2.6 compat, so let's wait and see if upstream accepts it: https://github.com/uniphil/venvgit2/pull/12

twaugh:
Updated with better init code.

twaugh:
I've taken out the venvgit2 business for the time being and filed #269 for it. We can hope the upstream Python 2.6 support comes in soon.

In the mean time, this PR now passes tests.

bkabrda:
LGTM

PR #275, opened 2015-08-21
Stop autorebuild if disabled in config file

bkabrda:
This is a multiproject effort, that will also require:
- change to osbs-client documentation about build process
- changes to pyrpkg to provide commands to get/set this value


twaugh:
Looks pretty good to me, just a few comments above.

bkabrda:
Thanks for the comments Tim. I fixed everything in the second commit.

twaugh:
LGTM

PR #276, opened 2015-08-24
Fail as soon as a required plugin fails (fixes #261)

### LINKED ISSUE #261: Plugin failure should fail the whole build
Plugins should fail the build and should "fail fast". If atomic-reactor can't import a plugin that user requested via `build.json` or that plugin fails, then the whole build should fail. I also think it should fail immediately, except perhaps exit plugins that get run either way.
The dangers of current behavior include
- delivering a different result that user wanted
- user not noticing that he got a different result, since atomic-reactor logs are pretty verbose and hard to read
Let's discuss :) @TomasTomecek @twaugh @mmilata @jpopelka

twaugh:
The only plugins that do not terminate the build are those with `can_fail=True` set, and exit plugins.

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -352,7 +352,7 @@ class DockerBuildWorkflow(object):
                                             self.exit_plugins_conf,
                                             plugin_files=self.plugin_files)
             try:
-                exit_runner.run()
+                exit_runner.run(keep_going=True)
             except PluginFailedException as ex:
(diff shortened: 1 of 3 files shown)


twaugh:
Oh, please hold on this one.

twaugh:
Re-based and adjusted so that a missing configured plugin causes a failure.

TomasTomecek:
LGTM

twaugh:
Clarifications added to comments.

jpopelka:
LGTM

PR #278, opened 2015-08-25
Make source accessible to exit plugins

bkabrda:
- also fix a test issue, where self was passed twice to constructors due to usage of 'super'


twaugh:
LGTM

jpopelka:
LGTM

PR #280, opened 2015-08-26
Store openshift build selflink in build json

bkabrda:
- also fix setting default value for prebuild plugins in openshift v3 input

(I hope it's ok to fix the two not-so-related issues in one PR; they're small enough for me to think it is OK)


jpopelka:
LGTM

twaugh:
LGTM, mind if I rebase before merging?

Issue #270, opened 2015-08-13
pulp_push: should use secret from service account, not sourceSecret


twaugh:
The secret used by the `pulp_push` plugin is not really what `sourceSecret` is for, i.e. the secret needed for fetching the source.

Instead, we should configure the `builder` service account to have access to a named secret, and name that secret in the configuration for `pulp_push`.

Support for this will be in Origin 1.0.6.

The way it should be done is to create a secret and add it to the `builder` service account:
```
$ oc secrets new pulp ./pulp.key ./pulp.cer
secrets/pulp
$ oc secrets add serviceaccount/builder secret/pulp --for=mount
```

Then in the BuildConfig for atomic-reactor, specify the secret in the strategy's `secrets` array, specifying a mount path:
```
"secrets": [{
  "secretSource": {
    "name": "pulp"
  },
  "mountPath": "/var/run/secrets/pulp"
}]
```

In the configuration for the `pulp_push` plugin, specify the same path for `pulp_secret_path`:
```
"pulp_push": {
  "pulp_registry_name": ...,
  ...
  "pulp_secret_path": "/var/run/secrets/pulp"
}
```

The point of doing it this way rather than using `sourceSecret` is that we only get one `sourceSecret` per build, but we can have multiple service account secrets. This way, other plugins get to use secrets too.

PR #282, opened 2015-08-27
Create exit_sendmail plugin

bkabrda:
This PR creates `exit_sendmail` plugin - a plugin that can be used to send notifications on various conditions. There are several todo items for this PR before I'll call it ready for merge:

- [x] Document how to call the plugin
- [x] Figure out how to securely authenticate against PDC (i.e. where to get PDC auth token from)
- [x] Add a `pdc_verify_cert` argument to specify whether or not the plugin should verify PDC SSL certificate
- [x] Make the `contact_role` to send notification to configurable (now it's hardcoded as `Build_Owner` and I'm not even sure that's the correct role)
- [x] Write tests
- [x] Figure out how to get name of user who submitted the build, assuming we're notifying after a user-submitted build (low-prio, this will most likely be done by Koji when deployed in OSBS => low prio, probably not a blocker for merging)

When this is done, we'll need to implement setting the PDC source secret in osbs-client. This can be implemented along with enabling the sendmail plugin.

### DIFF
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -7,6 +7,7 @@ of the BSD license. See the LICENSE file for details.
 """
 
 from email.mime.text import MIMEText
+import os
 import smtplib
 import socket
 try:
@@ -51,9 +52,11 @@ class SendMailPlugin(ExitPlugin):
 
     allowed_states = set([MANUAL_SUCCESS, MANUAL_FAIL, AUTO_SUCCESS, AUTO_FAIL, AUTO_CANCELED])
 
+    PDC_TOKEN_FILE = 'pdc.token'
+
     def __init__(self, tasker, workflow, send_on=None, url=None, pdc_url=None,
-                 pdc_verify_cert=True, pdc_component_df_label=None, smtp_url=None,
-                 from_address=None, error_addresses=None):
+                 pdc_verify_cert=True, pdc_component_df_label=None, pdc_secret_path=None,
+                 smtp_url=None, from_address=None, error_addresses=None):
         """
         constructor
 
@@ -64,6 +67,7 @@ class SendMailPlugin(ExitPlugin):
         :param pdc_url: URL of PDC to query for contact information
         :param pdc_verify_cert: whether or not to verify SSL cert of PDC (defaults to True)
         :param pdc_component_df_label: name of Dockerfile label to use as PDC global_component
+        :param pdc_secret_path: path to pdc.token file; $SOURCE_SECRET_PATH otherwise
         :param smtp_url: URL of SMTP server to use to send the message (e.g. "foo.com:25")
         :param from_address: the "From" of the notification email
         :param error_addresses: list of email addresses where to send an email if there's an error
@@ -75,6 +79,7 @@ class SendMailPlugin(ExitPlugin):
         self.pdc_url = pdc_url
         self.pdc_verify_cert = pdc_verify_cert
         self.pdc_component_df_label = pdc_component_df_label
+        self.pdc_secret_path = pdc_secret_path
         self.smtp_url = smtp_url
         self.from_address = from_address
         self.error_addresses = error_addresses
@@ -123,8 +128,15 @@ class SendMailPlugin(ExitPlugin):
         return (subject_template % formatting_dict, body_template % formatting_dict)
 
     def _get_pdc_token(self):
-        # TODO: get token from file mounted inside the build container?
-        return ''
+        if self.pdc_secret_path is not None:
+            token_file = os.path.join(self.pdc_secret_path, self.PDC_TOKEN_FILE)
+        else:
+            token_file = os.path.join(os.environ['SOURCE_SECRET_PATH'], self.PDC_TOKEN_FILE)
+
+        self.log.debug('getting PDC token from file %s', token_file)
+
+        with open(token_file, 'r') as f:
+            return f.read().strip()
 
     def _get_component_label(self):
         """Get value of Dockerfile label that is to be used as `global_component` to query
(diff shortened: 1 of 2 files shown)


bkabrda:
Also, this depends on #280

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -0,0 +1,207 @@
...
+
+from atomic_reactor.plugin import ExitPlugin, PluginFailedException
+from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
+from atomic_reactor.source import GitSource
+
+
+class SendMailPlugin(ExitPlugin):
+    key = "sendmail"
+    allowed_states = set(['manual_success', 'manual_fail', 'auto_success', 'auto_fail',
+                          'auto_canceled'])
```
twaugh: It would be nice to have symbolic constants for these strings, like

```
MANUAL_SUCCESS = 'manual_success'
...
allowed_states = set([self.MANUAL_SUCCESS, ...])
```

That way typos will be caught more quickly.

bkabrda: Good idea, will do.

bkabrda:
So it seems that we could have a file with PDC token on the host machine and then use the approach described at https://github.com/projectatomic/atomic-reactor/issues/270 to mount it in, read it and use it for authentication. At least that's the way I understand it, can you please confirm, @twaugh?

twaugh:
Yes, that's correct.

The osbs-client specifies the name of the secret and the path it will be mounted at, so the plugin will need to accept a path in its configuration so it knows where to look.

bkabrda:
This PR is now ready for review. I'm working on osbs-client related changes right now.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -0,0 +1,263 @@
...
+                 pdc_verify_cert=True, pdc_component_df_label="BZComponent", pdc_secret_path=None,
+                 pdc_contact_role="Build_Owner", smtp_url=None, from_address=None,
+                 error_addresses=None):
+        """
+        constructor
+
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param url: URL to OSv3 instance where the build logs are stored
+        :param send_on: list of build states when a notification should be sent
```
twaugh: Would be nice to list `url` and `send_on` in the order they appear in the keyword list.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -0,0 +1,263 @@
...
+        self.pdc_verify_cert = pdc_verify_cert
+        self.pdc_component_df_label = pdc_component_df_label
+        self.pdc_secret_path = pdc_secret_path
+        self.pdc_contact_role = pdc_contact_role
+        self.smtp_url = smtp_url
+        self.from_address = from_address
+        self.error_addresses = error_addresses
+
+    def _should_send(self, rebuild, success, canceled):
+        """Return True if mail should be sent under given conditions and `self.send_on`."""
```
twaugh: Should it be "in `self.send_on`"?

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -0,0 +1,263 @@
...
+            'user': '<autorebuild>' if rebuild else 'TODO user',
+            'logs': url
+        }
+        return (subject_template % formatting_dict, body_template % formatting_dict)
+
+    def _get_pdc_token(self):
+        if self.pdc_secret_path is not None:
+            token_file = os.path.join(self.pdc_secret_path, self.PDC_TOKEN_FILE)
+        else:
+            token_file = os.path.join(os.environ['SOURCE_SECRET_PATH'], self.PDC_TOKEN_FILE)
```
twaugh: The `sourceSecret` slot is already taken by pulp, so it's `pdc_secret_path` or nothing.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -0,0 +1,263 @@
...
+                body = '\n'.join([
+                    'Failed to get contact for %s, error: %s' % (str(self.workflow.image), str(e)),
+                    'Since your address is in "error_addresses", this email was sent to you to '
+                    'take action on this.',
+                    'Wanted to send following mail:',
+                    '',
+                    body
+                ])
+                receivers = self.error_addresses
+            self.log.info('sending notification to %s ...' % receivers)
```
twaugh: `receivers` should be its own argument instead of using '%' here.

twaugh:
Looks good aside from the comments above and the TODOs already noted.

jpopelka:
LGTM

atomic_reactor/plugins/exit_sendmail.py
bkabrda: Good catch, will do :)

atomic_reactor/plugins/exit_sendmail.py
bkabrda: Ugh, the docstring doesn't make much sense. I'll rewrite it.

atomic_reactor/plugins/exit_sendmail.py
bkabrda: And I always took pride in not doing C&P mistakes ;)

atomic_reactor/plugins/exit_sendmail.py
bkabrda: Thanks, I missed that.

bkabrda:
I fixed the issues discovered and also added a `submitter` (user who created the build) argument to the plugin; I think this is the best way to go - osbs-client will re-set the submitter on every submitted build and autorebuilds have special value `<autorebuild>`.

bkabrda:
Everything is now fixed and implemented, ready to be pulled.

twaugh:
LGTM

PR #284, opened 2015-08-27
BumpReleasePlugin: DockerfileParser now supports modifying of labels

jpopelka:


twaugh:
We also want to adjust `requirements.txt` so it requires `dockerfile-parse >= 0.0.4` (or else it will silently fail to do anything to the Release label), as well as the equivalent changes in `osbs-client.spec`.

jpopelka:
`requirements.txt` and `atomic-reactor.spec` changes added, but why `osbs-client.spec` ?

twaugh:
Sorry, I meant `atomic-reactor.spec`. :-)

LGTM

PR #291, opened 2015-08-31
plugin,pulp: silence ssl warnings

### LINKED ISSUE #245: get rid of requests' ssl warnings
```
/usr/lib/python2.7/site-packages/requests-2.7.0-py2.7.egg/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
```

TomasTomecek:
Fixes #245

### DIFF
--- a/atomic_reactor/plugins/post_push_to_pulp.py
+++ b/atomic_reactor/plugins/post_push_to_pulp.py
@@ -24,6 +24,13 @@ import re
 import tempfile
 
 
+# let's silence warnings from dockpulp: there is one warning for every request
+# which may result in tenths of messages: very annoying
+# with "module", it just prints one warning -- this should balance security and UX
+import warnings
+warnings.filterwarnings("module")
+
+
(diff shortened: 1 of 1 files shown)


mmilata:
Even though it's probably little security risk in this case, I feel that just silencing the warning is bad form. Can the root cause be fixed in dockpulp by configuring SSL properly? What about at least setting the [warning filter](https://docs.python.org/2/library/warnings.html#the-warnings-filter) to `module` or `once` so that the warning is shown only once instead of filing the screen?

TomasTomecek:
You're right. Silencing everything is actually pretty bad practice. I was just so excited that I figured out how to get rid of the warnings.

jpopelka:
LGTM, merging

Issue #288, opened 2015-08-28
squashing: provide id of base image to the tool


TomasTomecek:
so it will squash just the built image, not multiple layers from base image

jpopelka:
Isn't this rather a osbs-client issue ?
I mean, the squash plugin already knows how to squash from particular layer/image id.
It's just that 'something' doesn't specify the id when using the squash plugin.
And the 'something' is osbs-client, isn't it ?

jpopelka:
So I think this should be added to the squash plugin:
if the id (`from_layer` arg) hasn't been specified, squash from base_image layer

TomasTomecek:
The thing is that most of the time it is undesirable to figure out id of base image to provide to `from_layer`. Therefore I would be in favor of adding new option `from_base` and set it to `True` by default.

jpopelka:
And what if `from_base` and `from_layer` are specified at the same time ? Should that be an error ?

jpopelka:
And if it's undesirable *most of the time* then it should probably be set to `False` by default ?

jpopelka:
We can discuss it further in #292

PR #292, opened 2015-08-31
PrePublishSquashPlugin: new from_base argument (#288)

jpopelka:


TomasTomecek:
This is exactly what I had in mind. Except... (will inline comment)

atomic_reactor/plugins/prepub_squash.py
```diff
@@ -55,20 +56,28 @@ class PrePublishSquashPlugin(PrePublishPlugin):
...
         """
         super(PrePublishSquashPlugin, self).__init__(tasker, workflow)
         self.image = self.workflow.builder.image_id
-        self.from_layer = from_layer
         self.tag = tag or str(self.workflow.builder.image)
+        if from_base and from_layer is not None:
+            raise RuntimeError("Both: 'from_base' and 'from_layer' specified.")
+        self.from_layer = from_layer
+        if from_base:
+            df_path, _ = self.workflow.source.get_dockerfile_path()
```
TomasTomecek: this won't work, because squash tool expects ID

I think we should use [this attribute](https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/inner.py#L297) of `inner.py` so we don't inspect images all over the place.

jpopelka: From what I've seen it takes tags/names too, but your suggestion is definitely better.
Check new commit please.

goldmann: It can take names too. I'm resolving it.

atomic_reactor/plugins/prepub_squash.py
```diff
@@ -55,20 +55,30 @@ class PrePublishSquashPlugin(PrePublishPlugin):
...
         self.image = self.workflow.builder.image_id
-        self.from_layer = from_layer
         self.tag = tag or str(self.workflow.builder.image)
+        if from_base and from_layer is not None:
+            raise RuntimeError("Both: 'from_base' and 'from_layer' specified.")
+        self.from_layer = from_layer
+        if from_base:
+            base_image_id = self.workflow.base_image_inspect.get('Id', None)
+            if base_image_id is None:
+                self.log.error("Id is missing in inspection: '%s'", self.workflow.base_image_inspect)
```
TomasTomecek: I feel like we should `raise` here.

jpopelka: Sure, I've rebased the commit.

TomasTomecek:
Thanks Jirko! LGTM

jpopelka:
Fixes #288

PR #296, opened 2015-09-01
push_pulp: describe using service account secrets (fixes #270)

### LINKED ISSUE #270: pulp_push: should use secret from service account, not sourceSecret

twaugh:
This pull request describes how to use the `pulp_secret_path` parameter and some other build JSON changes to use secrets from the `builder` service account rather than using the `sourceSecret` slot.

It also adds a test case to verify that `pulp_secret_path` works correctly.

### DIFF
--- a/atomic_reactor/plugins/post_push_to_pulp.py
+++ b/atomic_reactor/plugins/post_push_to_pulp.py
@@ -11,6 +11,36 @@ from dockpulp import setup_logger
 
 """
 Push built image to pulp registry
+
+Several authentication schemes are possible, including
+username+password and key/certificate via sourceSecret.
+
+However, the recommended scheme (since Origin 1.0.6) is to store a
+key and certificate in a secret which the builder service account is
+allowed to mount:
+
+$ oc secrets new pulp ./pulp.key ./pulp.cer
+secrets/pulp
+$ oc secrets add serviceaccount/builder secret/pulp --for=mount
+
+In the BuildConfig for atomic-reactor, specify the secret in the
+strategy's 'secrets' array, specifying a mount path:
+
+"secrets": [{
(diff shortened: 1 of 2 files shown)


mmilata:
LGTM, merged. Related question, how is the service account that is used for building determined? Is the `builder` value hardcoded in openshift? I'm using `default` in ansible-osbs so I assume that's wrong now.

twaugh:
I think it's hardcoded, yes:
https://docs.openshift.org/latest/admin_guide/service_accounts.html#managed-service-accounts

PR #298, opened 2015-09-02
'raise' rather than 'return' in plugins' run() to fail early.

jpopelka:
The first commit is unrelated, I just spotted it when checking the plugins.

The second commit (which I'm expecting to be changing according to the comments in PR, that's why it's second) is about failing early rather then silently returning from run(). There's not a big difference if the plugin's can_fail is True (in which case the workflow continues anyway), but better to be consistent.
I'm however not sure in the `prepub_tests_for_image` case - whether it'd be better to fail (after this commit) if there's no image to test or continue (prior to the commit). I'm open to discussion.


twaugh:
Strongly in favour of `raise` in these cases.

TomasTomecek:
LGTM

PR #300, opened 2015-09-03
squash plugin: turn from_base on by default

jpopelka:
See https://github.com/projectatomic/atomic-reactor/issues/288#issuecomment-136359030


TomasTomecek:
Very nice!

PR #304, opened 2015-09-04
Rename can_fail to is_allowed_to_fail (fixes #92)

### LINKED ISSUE #92: Rename can_fail variable in plugins
Currently you use the `can_fail` name for variable that specifies if the plugin can fail the build or not, but there is a problem with the name which makes it **very** confusing, it has two meanings:

1. can fail - it can safely fail, we don't care
2. can fail - watch out, it can fail the build, be prepared!

jpopelka:
If I'm counting correctly we all agreed on renaming `can_fail` variable to `is_allowed_to_fail` in #92.


mmilata:
LGTM, merged. Almost forgot this issue existed:)

PR #306, opened 2015-09-04
Avoid requiring pygit2 (fixes #305)

### LINKED ISSUE #305: bump_release: avoid pygit2 dependency
Availability of the `python-pygit2` package is limited, whereas the `git` binary is everywhere. We should avoid depending on pygit2 and instead use `git` directly.

twaugh:
It's still quite useful for testing the bump_release plugin so is used only for that.

Needs some testing.


twaugh:
Tested now.

twaugh:
I'd like to squash before merge.

jpopelka:
LGTM

twaugh:
Re-based and squashed.

PR #310, opened 2015-09-07
Use autoversion client of docker-py if available (fixes #309)

### LINKED ISSUE #309: use autoversion client of docker-py
- [ ] ensure it's available in version of docker-py in rhel

https://github.com/docker/docker-py/blob/2dfab762241fa5f9d88dc785cf7fcf0a3e2dff8a/docker/client.py#L893

jpopelka:
We don't use the new `AutoVersionClient` class (https://github.com/docker/docker-py/blob/2dfab762241fa5f9d88dc785cf7fcf0a3e2dff8a/docker/client.py#L893), we just check whether it's there and if it is, we know that `Client` understands the `'auto'` value of `'version'`.


jpopelka:
I'm working on the failed tests.

jpopelka:
Done, please review.

TomasTomecek:
LGTM (I would have implemented it differently though)

atomic_reactor/core.py
```diff
@@ -208,12 +208,17 @@ def build_image_privileged_container(self, build_image, json_args_path):
...
         if base_url:
-            self.d = docker.Client(base_url=base_url)
+            client_kwargs['base_url'] = base_url
         elif os.environ.get('DOCKER_CONNECTION'):
-            self.d = docker.Client(base_url=os.environ['DOCKER_CONNECTION'])
-        else:
-            self.d = docker.Client()
+            client_kwargs['base_url'] = os.environ['DOCKER_CONNECTION']
+
+        if 'AutoVersionClient' in docker.__dict__:
```
twaugh: Would `hasattr(docker, 'AutoVersionClient')` be nicer?

twaugh:
If we had a pytest fixture for the DockerTasker, which mocked Docker first, could we eliminate some duplicated code?

atomic_reactor/core.py
jpopelka: Certainly, I didn't realize it works for modules too.

jpopelka:
Rebased with both comments covered.

tests/test_plugin.py
```diff
@@ -29,12 +32,16 @@
 
 
 def test_load_prebuild_plugins():
+    if MOCK:
+        mock_docker()
     runner = PreBuildPluginsRunner(DockerTasker(), DockerBuildWorkflow(SOURCE, ""), None)
```
twaugh: Why not use the `docker_tasker` fixture here, and in `test_load_postbuild_plugins()` below?

jpopelka: Fixed

jpopelka:
But I can't figure out why the fixture doesn't work when used in parametrized function, for example it  doesn't work in parametrized `test_rpmqa_plugin()`, while it's ok in `test_rpmqa_plugin_exception()`.

twaugh:
Yes, I don't know what's going on there. Maybe a fixture wasn't such a good idea, sorry. :-(

PR #313, opened 2015-09-08
rpmqa: ignore gpg-pubkey, not gpg-pubkey-

TomasTomecek:


atomic_reactor/plugins/post_rpmqa.py
```diff
@@ -40,8 +40,8 @@ def run(self):
         # gpg-pubkey- are autogenerated packages by rpm when you import a gpg key
         # these are of course not signed, let's ignore those by default
         if self.ignore_autogenerated_gpg_keys:
-            self.log.debug("ignore rpms 'gpg-pubkey-'")
-            plugin_output = [x for x in plugin_output if not x.startswith("gpg-pubkey-")]
+            self.log.debug("ignore rpms 'gpg-pubkey'")
+            plugin_output = [x for x in plugin_output if not x.startswith("gpg-pubkey")]
```
twaugh: Should it be looking for the delimiting comma as well though?

atomic_reactor/plugins/post_rpmqa.py
```diff
@@ -40,8 +40,8 @@ def run(self):
         # gpg-pubkey- are autogenerated packages by rpm when you import a gpg key
```
twaugh: How about updating this comment?

atomic_reactor/plugins/post_rpmqa.py
TomasTomecek: hmmm, good point

atomic_reactor/plugins/post_rpmqa.py
TomasTomecek: another /facepalm
will address this later, thanks Tim!

PR #316, opened 2015-09-08
Add a testcase for rpmqa plugin

twaugh:
Although we should ignore gpg-pubkey packages, we should include all other packages even if they start with that name.


jpopelka:
LGTM

Issue #307, opened 2015-09-04
Plug-in for submitting build using Content Generator API


twaugh:
We need a plug-in for submitting successful builds to koji using the Content Generator API.

Further reading:
https://fedoraproject.org/wiki/Koji/ContentGenerators

Needs to:
 - [x] find container ID of buildroot (https://github.com/projectatomic/osbs-client/pull/239)
 - [x] find OSBS host (not buildroot container) operating system and arch
 - [x] fill in extra output 'docker' fields correctly
 - [x] call the XMLRPC API method

twaugh:
See https://github.com/projectatomic/atomic-reactor/compare/master...koji_promote for work-in-progress.

twaugh:
Requires a fix for https://github.com/projectatomic/osbs-client/issues/238.

twaugh:
Anyone have any better idea to get the OSBS host operating system and arch than running `cat /host/etc/os-release` in a new container with a `/:/host` volume map? (Hope so...)

twaugh:
Let's see if `docker.Client().info()['OperatingSystem']` is good enough.

bkabrda:
@twaugh how about:

```
import platform
arch = platform.machine()  # arch = 'x86_64'
distro = platform.dist()  # distro = ('fedora', '22', 'Twenty Two')
```

For the most part, the platform module just parses some files like `/etc/os-release`, but at least it does it in a very robust and reliable way (always worked fine for me).

bkabrda:
(...and it's been part of Python's stdlib since 2.3: https://docs.python.org/2/library/platform.html)

twaugh:
@bkabrda: It's no good: we need to know the operating system of the *host* machine, not of the container.

bkabrda:
Ahha, I missed that. In that case I think we're on our own :/ (just a side note - it may be interesting to extend the platform module to allow specifying `root` argument to every function, which would then be prefixed to all the paths currently used)

twaugh:
Fixed e.g. #365.

PR #319, opened 2015-09-09
Plugin for promoting builds to Content Generator API

twaugh:
Here is a plugin for using the Content Generator API to submit successful builds to Koji.

Please do not merge yet as the XMLRPC interface is not yet specified and the metadata format definition is not yet final.

Review comments are still useful, however.


twaugh:
Fixes #307.

twaugh:
Thanks for comments, should all be addressed now.

jpopelka:
Otherwise LooksVeryGoodToMe so far.

twaugh:
I've removed the `constants.VERSION` constant and now use `pkg_resources.get_distribution(PROG).version` for getting the version of atomic-reactor.

bkabrda:
I suggest putting `__version__` to `atomic_reactor/__init__.py`. It's PEP-396 standardized way of doing this.

[1] https://www.python.org/dev/peps/pep-0396/#specification

twaugh:
Sure, but there's a problem to solve there, isn't there? i.e. how do we keep that version in just one place? What method would you suggest?

bkabrda:
From my experience, it's usually best to just keep version in more places (in our case that would be specfile, setup.py and __init__.py). It is unfortunate, but it seems to me that all the other methods I've seen are flawed (like importing __init__.py from setup.py, parsing it, etc...) I'm afraid I don't have a better answer for this.

twaugh:
Example metadata generated by the plugin: https://gist.github.com/twaugh/b2967d37906fc7a010b4

TomasTomecek:
Will `buildroots → components` list every rpm in buildroot?

twaugh:
Yes, everything that isn't named `gpg-pubkey`.

twaugh:
Ready for review.

jpopelka:
LGTM

twaugh:
Sample metadata: https://gist.github.com/twaugh/b2967d37906fc7a010b4
(from `py.test -k promote_success -s`)

twaugh:
I've added support for specifying the Kerberos principal and keytab.

jpopelka:
LGTM

twaugh:
Re-based to solve conflicts.

twaugh:
Sample metadata output updated:
 * v1 image: https://gist.github.com/twaugh/b2967d37906fc7a010b4
 * metadata-only: https://gist.github.com/twaugh/71096604ee6f2862b1ce

twaugh:
Anyone willing to merge this?

mmilata:
Good job, merged!

PR #321, opened 2015-09-10
Don't run any plugins after docker build failure (fixes #320)

### LINKED ISSUE #320: don't run pre-publish and post-build plugins when build fails
...so confusing

twaugh:
This pull request includes more thorough testing for plugin failures, as well as a new test for docker build failure handling.

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -327,7 +327,8 @@ class DockerBuildWorkflow(object):
             self.build_failed = build_result.is_failed()
 
             if build_result.is_failed():
-                # The docker build failed
+                # The docker build failed. Finish here, just run the
+                # exit plugins (from the 'finally:' block below).
                 return build_result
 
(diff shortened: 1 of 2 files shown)


atomic_reactor/inner.py
```diff
@@ -326,8 +326,11 @@ def build_docker_image(self):
 
             self.build_failed = build_result.is_failed()
 
-            if not build_result.is_failed():
-                self.built_image_inspect = self.builder.inspect_built_image()
+            if build_result.is_failed():
+                # The docker build failed
+                return build_result
```
TomasTomecek: just to be sure: this will run exit plugins, right?

twaugh: It will. The `finally` clause is guaranteed to run when we exit the block. I can update the comment.

TomasTomecek:
This is wonderful! The new test cases are awesome! LGTM

PR #323, opened 2015-09-10
rebuilds sub-package (#297)

### LINKED ISSUE #297: Recently added plugins require osbs(-client)
specifically `post_import_image.py` & `pre_check_and_set_rebuild.py`

Not sure in which sub-package are they supposed to be shipped, but the spec file should be amended to either:
a)  add them to `%files -n python[3]-atomic-reactor-metadata` which already requires `osbs`
b)  add `Requires: osbs` to `%package -n python[3]-atomic-reactor`

jpopelka:
fixes #297

### DIFF
--- a/atomic-reactor.spec
+++ b/atomic-reactor.spec
@@ -56,6 +56,25 @@ probably implement if you started hooking Docker into your
 infrastructure.
 
 
+%package -n python-atomic-reactor
+Summary:        Python 2 Atomic Reactor library
+Group:          Development/Tools
+License:        BSD
+Requires:       python-docker-py
+Requires:       python-requests
+Requires:       python-setuptools
+Requires:       python-dockerfile-parse >= 0.0.5
+Requires:       python-backports-lzma
+# Due to CopyBuiltImageToNFSPlugin, might be moved to subpackage later.
+Requires:       nfs-utils
+Provides:       python-dock = %{version}-%{release}
+Obsoletes:      python-dock < %{dock_obsolete_vr}
+
+%description -n python-atomic-reactor
+Simple Python 2 library for building Docker images. It contains
+a lot of helpful functions which you would probably implement if
+you started hooking Docker into your infrastructure.
+
 %package -n python-atomic-reactor-koji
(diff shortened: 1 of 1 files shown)


atomic-reactor.spec
```diff
@@ -84,24 +104,14 @@ Obsoletes:      python-dock-metadata < %{dock_obsolete_vr}
...
-Requires:       python-requests
-Requires:       python-setuptools
-Requires:       python-dockerfile-parse >= 0.0.4
-Requires:       python-backports-lzma
-# Due to CopyBuiltImageToNFSPlugin, might be moved to subpackage later.
-Requires:       nfs-utils
-Provides:       python-dock = %{version}-%{release}
-Obsoletes:      python-dock < %{dock_obsolete_vr}
+Requires:       python-atomic-reactor = %{version}-%{release}
+Requires:       osbs
```
twaugh: This will be 'osbs >= 0.15' I guess, for `get_pod_for_build()` and `update_labels_on_build_config()` and maybe others.

But we don't have a 0.15 release yet so I'm not sure how much sense it makes to put that explicit version requirement in.

twaugh:
Do you think we should update `README.md` (line 35) as well?
```
$ yum install atomic-reactor python-atomic-reactor-koji
```

atomic-reactor.spec
twaugh: Thinking more about this, I think we *should* put "osbs >= 0.15" in here.

TomasTomecek:
I think that chain rebuilds would deserve one document on its own with precise steps how to use them.

bkabrda:
@TomasTomecek most of that is described in https://github.com/projectatomic/osbs-client/blob/master/docs/build_process.md#the-process, but it's still being in works, since some parts of chain rebuilds are still being under construction.

TomasTomecek:
@bkabrda right; I actually meant: "how tu deploy and run chain rebuilds"

jpopelka:
Rebased. I've only added `>= 0.15` to `Requires: osbs` as Tim suggested.

TomasTomecek:
LGTM

PR #324, opened 2015-09-11
display versions of used tools before build

### LINKED ISSUE #131: display version of used software in build logs
* dock
* osbs
* docker
* docker-py
* docker-scripts

Preferably rpms, not python packages.

TomasTomecek:
Fixes #131 

- [x] rewrite to use `__version__`
- [x] make sure `osbs` and `reactor` use `__version__`
- [x] make sure `docker_scripts` use `__version__`
- [x] log `docker info`, `docker version`
- [x] testz

### DIFF
--- a/atomic_reactor/core.py
+++ b/atomic_reactor/core.py
@@ -557,3 +557,19 @@ class DockerTasker(LastLogger):
             response = response is not None
         logger.debug("image exists: %s", response)
         return response
+
+    def get_info(self):
+        """
+        get info about used docker environment
+
+        :return: dict, json output of `docker info`
+        """
+        return self.d.info()
+
+    def get_version(self):
+        """
+        get version of used docker environment
+
+        :return: dict, json output of `docker version`
+        """
+        return self.d.version()
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -19,7 +19,13 @@ import subprocess
 import tempfile
 import logging
 import uuid
-from atomic_reactor.constants import DOCKERFILE_FILENAME
+
+from atomic_reactor.constants import DOCKERFILE_FILENAME, TOOLS_USED
+
+try:
+    from importlib import import_module
+except ImportError:
+    import_module = __import__  # I love python 2.6
 
 
 logger = logging.getLogger(__name__)
@@ -445,3 +451,33 @@ def get_exported_image_metadata(path):
     metadata['sha256sum'] = s.hexdigest()
     logger.debug('sha256sum: %s', metadata['sha256sum'])
     return metadata
+
+
+def get_version_of_tools():
+    """
+    obtain versions of software reactor is using (specified in constants.TOOLS_USED)
+
+    :returns dict, {"tool": "version", ...}
+    """
+    response = {}
+    for tool in TOOLS_USED:
+        try:
+            tool_module = import_module(tool)
+        except ImportError as ex:
+            logger.warning("can't import module %s: %s", tool, repr(ex))
+        else:
+            version = getattr(tool_module, "__version__", None)
+            if version is None:
+                logger.warning("tool %s doesn't have __version__")
+            else:
+                response[tool] = version
+    return response
+
+
+def print_version_of_tools():
+    """
+    print versions of used tools to logger
+    """
+    logger.info("Using these tools:")
+    for tool, version in get_version_of_tools().items():
+        logger.info("%s-%s", tool, version)
--- a/tests/test_tasker.py
+++ b/tests/test_tasker.py
@@ -298,3 +298,21 @@ def test_build_image_from_git(temp_image_name):
     assert response is not None
     assert t.image_exists(temp_image_name)
     t.remove_image(temp_image_name)
+
+
+def test_get_info():
+    if MOCK:
+        mock_docker()
+
+    t = DockerTasker()
+    response = t.get_info()
+    assert isinstance(response, dict)
+
+
+def test_get_version():
+    if MOCK:
+        mock_docker()
+
+    t = DockerTasker()
+    response = t.get_info()
+    assert isinstance(response, dict)
--- a/atomic_reactor/build.py
+++ b/atomic_reactor/build.py
@@ -9,12 +9,12 @@ of the BSD license. See the LICENSE file for details.
 Classes which implement tasks which builder has to be capable of doing.
 Logic above these classes has to set the workflow itself.
 """
+import json
 
 import logging
 from dockerfile_parse import DockerfileParser
 from atomic_reactor.core import DockerTasker, LastLogger
-from atomic_reactor.util import wait_for_command, ImageName
-
+from atomic_reactor.util import wait_for_command, ImageName, print_version_of_tools
 
 logger = logging.getLogger(__name__)
 
@@ -82,7 +82,13 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
         LastLogger.__init__(self)
         BuilderStateMachine.__init__(self)
 
+        print_version_of_tools()
+
         self.tasker = DockerTasker()
+
+        info, version = self.tasker.get_info(), self.tasker.get_version()
+        logger.debug(json.dumps(info, indent=2))
+        logger.info(json.dumps(version, indent=2))
 
         # arguments for build
         self.source = source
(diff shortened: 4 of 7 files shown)


bkabrda:
I'd say that having explicit `import rpm` is a wrong thing to do here, as it won't be available in travis/on debian based platforms, etc... If something, that should be try/except imported.

As for pkg_resources, I think a better approach would be to just see `<pkg>.__version__`, e.g. `rpm.__version__`. As long as we know that our dependencies have these, we should be safe (note that this is PEP 396-standardized way of specifying package version - one that we should adopt as well instead of using `VERSION` in `constants.py`, BTW :))

I do understand that if our dependency comes from RPM, we also want to know NVR to take into account possible downstream patches. The problem is that even if such an RPM is installed, our imported package may be coming from different path, perhaps pip-installed. So the method you implemented is flawed. To make it correct, you'd need to also check that `<pkg>.__path__` is a file owned by that RPM or something like that.

I'd say that, for a start, just seeing `<pkg>.__version__` is the best and most reliable way to detect dependency versions - and should be ok for vast majority of cases.

[1] https://www.python.org/dev/peps/pep-0396/#id25

TomasTomecek:
Okay. Will rewrite then. (this was first implementation anyway, I didn't expect at all it will be merged like this)

TomasTomecek:
Reimplemented. `docker info` will spit tons of data -- hence I changed it to `.debug`. What does other maintainers think about this? I feel like debug info might be really helpful (especially for future bug reports)

cc @projectatomic/build-service-admins-committers 

e.g.
```
{
  "Labels": null,
  "NGoroutines": 29,
  "LoggingDriver": "json-file",
  "HttpProxy": "",
  "DriverStatus": [],
  "OperatingSystem": "Fedora 22 (Twenty Two)",
  "Containers": 2,
  "HttpsProxy": "",
  "BridgeNfIp6tables": true,
  "MemTotal": 8248397824,
  "Driver": "btrfs",
  "IndexServerAddress": "https://index.docker.io/v1/",
  "InitPath": "/usr/libexec/docker/dockerinit",
  "ExecutionDriver": "native-0.2",
  "OomKillDisable": true,
  "SystemTime": "2015-09-15T14:57:56.840887602+02:00",
  "Name": "quahog",
  "RegistryConfig": {
    "InsecureRegistryCIDRs": [
      "127.0.0.0/8"
    ],
    "IndexConfigs": {
      "172.17.0.2:5000": {
        "Official": false,
        "Name": "172.17.0.2:5000",
        "Secure": false,
        "Mirrors": []
      },
      "localhost:5000": {
        "Official": false,
        "Name": "localhost:5000",
        "Secure": false,
        "Mirrors": []
      },
      "10.3.13.125:5000": {
        "Official": false,
        "Name": "10.3.13.125:5000",
        "Secure": false,
        "Mirrors": []
      },
      "10.3.14.233:5000": {
        "Official": false,
        "Name": "10.3.14.233:5000",
        "Secure": false,
        "Mirrors": []
      },
      "docker.io": {
        "Official": true,
        "Name": "docker.io",
        "Secure": true,
        "Mirrors": null
      },
      "172.17.0.1:5000": {
        "Official": false,
        "Name": "172.17.0.1:5000",
        "Secure": false,
        "Mirrors": []
      },
      "172.17.42.1:5000": {
        "Official": false,
        "Name": "172.17.42.1:5000",
        "Secure": false,
        "Mirrors": []
      },
      "10.3.11.31:5000": {
        "Official": false,
        "Name": "10.3.11.31:5000",
        "Secure": false,
        "Mirrors": []
      },
      "172.17.0.3:5000": {
        "Official": false,
        "Name": "172.17.0.3:5000",
        "Secure": false,
        "Mirrors": []
      }
    },
    "Mirrors": null
  },
  "NCPU": 4,
  "NFd": 14,
  "CpuCfsQuota": true,
  "Debug": false,
  "ID": "PD35:TRT5:VF5I:YQBZ:WR5T:DDLP:GAGB:M374:EK7G:ITUG:6PN3:OCCI",
  "IPv4Forwarding": true,
  "KernelVersion": "4.0.7-300.fc22.x86_64",
  "BridgeNfIptables": true,
  "NoProxy": "",
  "InitSha1": "e5c1e560e7a6039966f5a527ceca84075960c3d4",
  "CpuCfsPeriod": true,
  "MemoryLimit": true,
  "SwapLimit": true,
  "Images": 276,
  "DockerRootDir": "/var/lib/docker",
  "NEventsListener": 0,
  "ExperimentalBuild": false
}
```

tests/test_tasker.py
```diff
@@ -298,3 +298,12 @@ def test_build_image_from_git(temp_image_name):
...
     t.remove_image(temp_image_name)
+
+
+def test_get_info():
+    if MOCK:
+        mock_docker()
+
+    t = DockerTasker()
+    response = t.get_info()
+    assert response is not None
```
twaugh: How about `assert isinstance(response, tuple)`, `assert len(response) == 2`, and verifying each is a dict?

atomic_reactor/core.py
```diff
@@ -557,3 +557,11 @@ def image_exists(self, image_id):
             response = response is not None
         logger.debug("image exists: %s", response)
         return response
+
+    def get_info(self):
+        """
+        get info about used docker environment
+
+        :return: tuple of (`docker info`, `docker version`)
```
twaugh: Wouldn't `get_info()` and `get_version()`, each returning one of the parts of that tuple, be less confusing?

atomic_reactor/util.py
```diff
@@ -445,3 +451,21 @@ def get_exported_image_metadata(path):
     metadata['sha256sum'] = s.hexdigest()
     logger.debug('sha256sum: %s', metadata['sha256sum'])
     return metadata
+
+
+def print_version_of_tools():
```
twaugh: This information will also be useful to the `koji_promote` plugin. How about renaming this function to something like `get_version_of_tools()`, returning a dict?

Then calls to `print_version_of_tools()` can be replaced with `for tool, version in versions.items(): logger.info(...)`.

twaugh:
When you add `__version__ = '...'` to `atomic_reactor/__init__.py`, perhaps it could be next to a comment indicating the other two places the version needs to be updated. https://github.com/projectatomic/atomic-reactor/pull/319#issuecomment-140701256

atomic_reactor/core.py
TomasTomecek: True. Will rewrite. (this is actually really good suggestion: I had something in mind before I started to code, then changed it and forgot to update this)

TomasTomecek:
@twaugh I'm actually thinking of writing a test, which will ensure that version in spec, setup.py and `__version__` is the same (and/or maybe create a release script).

@bkabrda Slavek, do you have experience with this?

twaugh:
Excellent plan.

tests/test_tasker.py
TomasTomecek: will test for dicts (I admit, I'm lazy)

bkabrda:
@TomasTomecek that's a good idea, but I have to admit I've never done such a thing. It may be pretty hard to parse the version from setup.py (you can't import it, since that would execute `setup` function); I'd advise putting

```
VERSION = 'xxx'
```

to setup.py and

```
%global version xxx
```

to spec, so that you make it easier for yourself to parse them.

TomasTomecek:
Could we put
```
VERSION = "1.2.3"
```
and use it in setup.py and `__version__`?

bkabrda:
You mean put it in `constants.py`? That's the whole problem, since if you put it there and import it from `setup.py`, then `atomic_reactor/__init__.py` will get executed during the import. That's not what we want, since it may do stuff like setup logging in a way that will screw `setup.py` up, etc. I've had various problems with this in past and I'd recommend against importing anything from the actual package in `setup.py`.

TomasTomecek:
> You mean put it in `constants.py`?

Yes, wrote it in rush and forgot to mention that.


Okay, that's a valid argument. Let's do the test & release script then.

TomasTomecek:
All the comments, can we merge this?
@twaugh @bkabrda

twaugh:
Yes, LGTM now.

bkabrda:
LGTM

TomasTomecek:
Thanks!

PR #327, opened 2015-09-17
add `__version__` to main package, test if versions are the same

TomasTomecek:
Related #324 

This means, that all the releases should be done via PR.


twaugh:
LGTM

PR #328, opened 2015-09-17
improve get_version_of_tools

TomasTomecek:
@twaugh PTAL

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -455,22 +455,23 @@ def get_exported_image_metadata(path):
 
 def get_version_of_tools():
     """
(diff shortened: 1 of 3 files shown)


atomic_reactor/util.py
```diff
@@ -459,18 +459,19 @@ def get_version_of_tools():
 
     :returns dict, {"tool": "version", ...}
```
twaugh: Comment needs updating.

twaugh:
Other than the docstring it looks great!

TomasTomecek:
Good catch, fixed, thanks!

twaugh:
LGTM

PR #331, opened 2015-09-20
Store layering metadata in OSv3 (fixes #329)

### LINKED ISSUE #329: metadata plugin: track layering
add these fields:
 * base image hash
 * base image name
 * built image id (after squashing)

twaugh:
Rather than storing the base image ID and name in a dict in one "base-image" label, a separate label is used for each field, allowing for label-based queries for either value.

I've used hyphens instead of underscores in the names to be consistent with `rpm-packages`, although it is inconsistent with `commit_id`.

This hasn't been tested with a real build yet.

### DIFF
--- a/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
+++ b/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
@@ -108,6 +108,9 @@ class StoreMetadataInOSv3Plugin(ExitPlugin):
             "rpm-packages": "\n".join(self.get_post_result(PostBuildRPMqaPlugin.key)),
             "repositories": json.dumps(repositories),
             "commit_id": commit_id,
+            "base-image-id": self.workflow.base_image_inspect['Id'],
+            "base-image-name": self.workflow.builder.base_image.to_str(),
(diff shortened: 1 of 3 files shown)


atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -108,6 +108,9 @@ def run(self):
             "rpm-packages": "\n".join(self.get_post_result(PostBuildRPMqaPlugin.key)),
             "repositories": json.dumps(repositories),
             "commit_id": commit_id,
+            "base-image-id": self.workflow.builder.base_image_inspect['Id'],
+            "base-image-name": self.workflow.builder.base_image,
```
TomasTomecek: This is instance of `ImageName`, will it serialize to name of the image?

twaugh: Hmm, probably not, no. Will fix.

twaugh:
Should be OK now.

TomasTomecek:
LGTM (I think we can merge and test on a real build later -- that's what I'm planning to do this week anyway)

PR #332, opened 2015-09-22
README.md: add link to atomic-devel list

mmilata:
As @TomasTomecek noticed the project has over 30 stars now. I guess we should have some kind of contact info in the README for people who are considering using atomic-reactor.

Which mailing list is the most suitable?


TomasTomecek:
Yep, that one. I read it daily.

LGTM

bkabrda:
Agreed, I read that list daily as well.

bkabrda:
Oh, and this LGTM :)

PR #334, opened 2015-09-24
Sendmail fixes

bkabrda:
This makes the default contact role "Devel_Owner" (I discussed this with a PDC developer offline and found out that "Build_Owner" is not what we want).
Also, this renames `smtp_url` to `smtp_uri`, which is more appropriate.


jpopelka:
LGTM

TomasTomecek:
LGTM

PR #335, opened 2015-09-24
Fix add_labels plugin for single-line dockerfiles

david-martin:
Fixes this error

```
2015-09-24 08:00:39,293 - atomic_reactor.build - DEBUG - using dockerfile:
LABEL "vcs_ref"="407bf1790a8d119c5d0f55f247a558cade2cb050"
FROM nodejs:0.10

2015-09-24 08:00:39,293 - atomic_reactor.core - INFO - building image 'fh-registry.usersys.redhat.com/fh-mbaas:2.0.3' from path '/tmp/tmp8Sbhps/source'
2015-09-24 08:00:39,419 - atomic_reactor.build - DEBUG - build is submitted, waiting for it to finish
2015-09-24 08:00:39,419 - atomic_reactor.util - INFO - wait_for_command
2015-09-24 08:00:39,420 - atomic_reactor.util - DEBUG - Step 0 : LABEL "vcs_ref" "407bf1790a8d119c5d0f55f247a558cade2cb050"
2015-09-24 08:00:39,420 - atomic_reactor.util - ERROR - {"errorDetail":{"message":"Please provide a source image with `from` prior to commit"},"error":"Please provide a source image with `from` prior to commit"}
```

In this case I am relying on a base image that has `ONBUILD` and `CMD` statements to do all the heavywork, therefore my Dockerfile is minimal i.e.
```Dockerfile
FROM nodejs:0.10
```

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -117,8 +117,13 @@ class AddLabelsPlugin(PreBuildPlugin):
         content = ""
         if labels:
             content = 'LABEL ' + " ".join(labels)
-            # put it before last instruction
-            lines.insert(-1, content + '\n')
+            num_lines = len(lines)
+            if num_lines < 2:
+                # ensure label is after FROM if its the only instruction in the Docker
+                lines.append('\n' + content + '\n')
+            else:
+                # put it before last instruction
+                lines.insert(-1, content + '\n')
 
             dockerfile.lines = lines
 
--- a/tests/plugins/test_add_labels.py
+++ b/tests/plugins/test_add_labels.py
@@ -40,6 +40,8 @@ DF_CONTENT = """\
 FROM fedora
 RUN yum install -y python-django
 CMD blabla"""
+DF_CONTENT_SINGLE_LINE = """\
+FROM fedora"""
 LABELS_CONF_BASE = {"Config": {"Labels": {"label1": "base value"}}}
 LABELS_CONF = OrderedDict({'label1': 'value 1', 'label2': 'long value'})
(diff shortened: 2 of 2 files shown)


atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -117,8 +117,13 @@ def escape(s):
         content = ""
         if labels:
             content = 'LABEL ' + " ".join(labels)
-            # put it before last instruction
-            lines.insert(-1, content + '\n')
+            num_lines = len(lines)
+            if num_lines < 2:
+                # ensure label is after FROM if its the only instruction in the Docker
+                lines.insert(num_lines, '\n' + content + '\n')
```
twaugh: This could be `lines.append()`.

twaugh:
Good catch -- thanks for the fix.

I only had one suggested change, above.

david-martin:
Thanks @twaugh, addressed feedback

TomasTomecek:
LGTM

jpopelka:
LGTM, @TomasTomecek do you remember why we add the labels before last instruction ? Why simply not add it as a last instruction in all cases ?

TomasTomecek:
@jpopelka TBH I'm not sure. Can imagine that `ONBUILD` can have issues with that.

TomasTomecek:
@david-martin Thanks for your contribution.

TomasTomecek:
@jpopelka I can't think of any reason why it should not work, I'll open a PR. `ONBUILD` works just fine b/c of
```
    "Config": {
        "OnBuild": [
            "uname -a"
        ],
```

david-martin:
Thanks @TomasTomecek for merging

PR #336, opened 2015-09-24
Use '%r' % object instead of '%s' % repr(object)

jpopelka:


atomic_reactor/core.py
```diff
@@ -79,8 +79,8 @@ def _check_build_input(self, image, args_path):
             with open(os.path.join(args_path, BUILD_JSON)) as json_args:
                 logger.debug("build input: image = '%s', args = '%s'", image, json_args.read())
         except (IOError, OSError) as ex:
-            logger.error("unable to open json arguments: '%s'", repr(ex))
-            raise RuntimeError("Unable to open json arguments: '%s'" % repr(ex))
+            logger.error("unable to open json arguments: '%r'", ex)
+            raise RuntimeError("Unable to open json arguments: '%r'" % ex)
```
twaugh: We can lose the single-quotes around it as well, otherwise we'll end up with something like `''foo''` or `'u'foo''`.

jpopelka: I think that's case only for representation of strings, not exceptions.
But I'm removing it to be consistent with other occurrences.
```
>>> repr(str())
"''"
>>> repr(Exception())
'Exception()'
```

twaugh:
LGTM

Issue #116, opened 2015-05-25
Support "Standard Container/Application Identifiers"


TomasTomecek:
https://github.com/projectatomic/ContainerApplicationGenericLabels

Related #115

So the standard is set, let's implement.
# Updated Labels
- `Name` → `name`
- `Version` → `version`
- `Release` → `release`
- `Architecture` → `architecture`
- `Vendor` → `vendor`
- `Authoritative_Registry` → `authoritative-source`
- `BZComponent` → `com.redhat.component`
- `Build_Host` → `com.redhat.build-host`
# New Labels
- `build-date` — Date/Time image was built
  - we can populate this automatically
- `vcs-type` — The type of version control used by the container source. Generally one of git, hg, svn, bzr, cvs
  - we can populate this automatically (we'll take this from source type)
- `vcs-ref` — A 'reference' within the version control repository; e.g. a git commit, or a subversion branch
  - we can populate this automatically (we'll put commit ID there, not branch)
- `authoritative-source-url` — The authoritative location in which the image is published
  - **what should we put here?**
  - **is this different than `authoritative-source`?**
# Maintainer Controlled Labels

Not sure how to automate these: we should probably leave the responsibility to maintainers.
- `url` (Optional) — Url with more information on the image
- `summary` (Required) — Short Description of the image
- `description` (Optional) — Detailed description of the image
- `vcs-url` (Optional) — URL of the version control repository
- `distribution-scope` — Intended scope of distribution for image (see below for possible values)
  - **could this be automatically set to `authoritative-source-only`?**
- `changelog-url`
# Assert labels plugin

We should also check, if all required labels are set. **Which are required? Is this list complete?**
- `name`
- `version`
- `release`
- `architecture`
- `vendor`
- `authoritative-source`
- `com.redhat.component`
- `com.redhat.build-host`
- `build-date`
- `vcs-type`
- `vcs-ref`
- `authoritative-source`
- `description`

@pbabinca @aweiteka @riekrh @Zyzyx
Comment please.

riekrh:
Please also see https://github.com/projectatomic/ContainerApplicationGenericLabels/tree/master/vendor/redhat as a reference to what is required.

TomasTomecek:
@riekrh I've read that and it's still not perfectly clear to me.

mmilata:
[...]

PR #337, opened 2015-09-24
Preliminary support for standard labels

mmilata:
See #116 and  https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md

The first commit extends `pre_add_labels_in_df` with couple of labels generated in the buildroot. These have to be explicitly requested in the plugin arguments. I'm not sure if it's sane to assume the docker daemon architecture is the same as the image architecture or if we want to keep it as OSBS API parameter.

Second commit introduces "deprecated labels" for the `pre_assert_labels`. The idea is that OSBS will provide dict of old->new labels and the builder will accept the old forms of required labels with a warning. This should give Dockerfile maintainers some time to start using the new labels. Example: `required_labels=["com.redhat.component"]`, `deprecated_labels={"BZComponent": "com.redhat.component"}`. The builder will allow both forms and if `BZComponent` is used warning will be printed. Question is whether there's a chance that some of the maintiners will see the warning.

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -44,12 +44,14 @@ from __future__ import unicode_literals
 from dockerfile_parse import DockerfileParser
 from atomic_reactor.plugin import PreBuildPlugin
 import json
+import datetime
 
 
 class AddLabelsPlugin(PreBuildPlugin):
     key = "add_labels_in_dockerfile"
 
-    def __init__(self, tasker, workflow, labels, dont_overwrite=("Architecture", )):
+    def __init__(self, tasker, workflow, labels, dont_overwrite=("Architecture", ),
+                 auto_labels=("build-date", "architecture", "vcs-type", "vcs-ref")):
         """
         constructor
 
@@ -57,6 +59,7 @@ class AddLabelsPlugin(PreBuildPlugin):
         :param workflow: DockerBuildWorkflow instance
         :param labels: dict, key value pairs to set as labels; or str, JSON-encoded dict
         :param dont_overwrite: iterable, list of label keys which should not be overwritten
+        :param auto_labels: iterable, list of labels to be determined automatically, if supported
         """
         # call parent constructor
         super(AddLabelsPlugin, self).__init__(tasker, workflow)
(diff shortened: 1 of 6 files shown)


atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -44,12 +44,14 @@
...
 import json
+import datetime
 
 
 class AddLabelsPlugin(PreBuildPlugin):
     key = "add_labels_in_dockerfile"
 
-    def __init__(self, tasker, workflow, labels, dont_overwrite=("Architecture", )):
+    def __init__(self, tasker, workflow, labels, dont_overwrite=("Architecture", ),
+                 auto_labels=()):
```
TomasTomecek: maybe we could do `auto_labels=("build-date", "architecture",...)`

TomasTomecek:
It would be awesome to track deprecated labels within reactor, somehow. Otherwise, I like this.

atomic_reactor/plugins/pre_add_labels_in_df.py
twaugh: It's generally good practice not to use mutable default arguments, but instead have them default to None and describe what they will be set to in the docstring.

mmilata: But tuples are immutable, or aren't they?

twaugh: Oh, thought you could assign to elements. No, you're right, they're immutable. Carry on :-)

mmilata:
Updated. I don't really understand why the tests tried to run against real docker despite the `docker_tasker` fixture calling `mock_docker()`.

I'm planning to submit another PR that updates the default values of `required_labels` and `deprecated_labels` once we have the label updates in OSBS client.

twaugh:
I don't think that fixture works.
https://github.com/projectatomic/atomic-reactor/pull/310#issuecomment-138556238

TomasTomecek:
This PR LGTM

TomasTomecek:
@mmilata could you please fix this? https://landscape.io/github/projectatomic/atomic-reactor/549/modules/atomic_reactor/plugins/pre_assert_labels.py#L4

Otherwise LGTM and ready for merge.

mmilata:
@TomasTomecek Should be fixed now.

TomasTomecek:
@mmilata++

PR #340, opened 2015-10-06
Make squash plugin output both tar and engine image if dont_load is False

bkabrda:
I think we can change the squash plugin this way in preparation for the V2 workflow; outputting the file even on `dont_load=False` may be extraneous in some cases, but it doesn't break API and should be safe to do.

### DIFF
--- a/atomic-reactor.spec
+++ b/atomic-reactor.spec
@@ -21,7 +21,7 @@
 
 Name:           %{project}
 Version:        1.6.0
-Release:        1%{?dist}
+Release:        2%{?dist}
 
 Summary:        Improved builder for Docker images
 Group:          Development/Tools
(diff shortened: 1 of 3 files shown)


jpopelka:
LGTM othewise

bkabrda:
Thanks for the review, both raised points are now fixed (I added the Requires for python-docker-scripts, since we should definitely have that).

jpopelka:
> I added the Requires for python-docker-scripts, since we should definitely have that

Looks good. In that case `requirements.txt` needs to contain it too I think.

bkabrda:
Heh, I knew I forgot something :)

bkabrda:
Hmm, wait a second, 0.4.4 hasn't been released yet, so pip installation of requirements would fail... Let's wait for it to be released, then I'll add it to requirements and then we can merge, ok?

jpopelka:
sure

TomasTomecek:
(looks like I'm really good at commenting on outdated diffs)

TomasTomecek:
We're waiting here for the release of `docker-scripts`, right?

jpopelka:
yes

bkabrda:
Fixed, rebased and ready for review and merging.

jpopelka:
LGTM, merging

PR #341, opened 2015-10-06
plugins,osbs: handle deprecation of openshift_uri -> openshift_url

TomasTomecek:
- [x] update also other plugins which use osbs-client


bkabrda:
LGTM

PR #342, opened 2015-10-06
add_labels_in_df: label rename functionality

mmilata:
This should address problem brought up in https://github.com/projectatomic/osbs-client/pull/266#issuecomment-145532579

When migrating to the new label set, we'll provide a dictionary that maps old labels to new ones. When old label is present but new one is not, the `add_labels_in_df` adds the new label. We'll also change the plugins that use the labels to use the new set. This replaces the *deprecated labels* mechanism that was added to `assert_labels` plugin in #337 / a2a7b09 (am I blind or is the plugin actually unused?).

Please note that the `add_labels_in_df` plugin will have to be moved in front of `bump_release` plugin in `inputs/prod_inner.json`.

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -97,13 +97,13 @@ class AddLabelsPlugin(PreBuildPlugin):
 
         for lbl in auto_labels:
             if lbl in self.labels:
-                self.log.info("label %s is set explicitly, not using generated value", lbl)
+                self.log.info("label %r is set explicitly, not using generated value", lbl)
                 continue
 
             if lbl in generated:
                 self.labels[lbl] = generated[lbl]
             else:
-                self.log.warning("requested automatic label %s is not available", lbl)
+                self.log.warning("requested automatic label %r is not available", lbl)
 
     def run(self):
         """
@@ -149,7 +149,7 @@ class AddLabelsPlugin(PreBuildPlugin):
                         continue
 
             label = '"%s"="%s"' % (escape(key), escape(value))
-            self.log.info("setting label %s", label)
+            self.log.info("setting label %r", label)
             labels.append(label)
 
         content = ""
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -97,21 +102,40 @@ def generate_auto_labels(self, auto_labels):
...
+                self.log.info("label %r is set explicitly, not using generated value", lbl)
                 continue
 
             if lbl in generated:
                 self.labels[lbl] = generated[lbl]
             else:
-                self.log.warning("requested automatic label %s is not available", lbl)
+                self.log.warning("requested automatic label %r is not available", lbl)
+
+    def add_renamed_labels(self, base_labels, df_labels, new_labels):
```
twaugh: Worth pointing out in the docstring that `df_labels` is a mutable view of the labels in the Dockerfile, and that updating it writes out new Dockerfile content?

Also, as the original labels are left in place I does 'rename' really describe it well? They're aliases or copies, really.

bkabrda: I totally agree that "rename" is somewhat misleading and would rather go with alias.

I'd probably also advise emitting debug message if there is alias for a label that doesn't exist. From the docstring in `__init__` method I guess this situation is expected, but it still may be worth logging when it occurs, for easier debugging.

Otherwise this PR looks good.

mmilata:
Pushed new version with your suggestions.

@twaugh Instead of warning in docstring I pass a copy of `dockerfile.labels` to the function, is that OK?

twaugh:
Yes, LGTM.

jpopelka:
The e2c757e comment says:
```
...
{"Authoritative_Registry": "authoritative-scope"}
...
LABEL authoritative-source="http://example.org"
```
Shouldn't there be either `scope` or `source` in both ?

jpopelka:
LGTM

mmilata:
Updated.

PR #345, opened 2015-10-08
Use the new label set

mmilata:
Depends on #342.

### DIFF
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -41,7 +41,7 @@ class SendMailPlugin(ExitPlugin):
                     # optional arguments follow
                     "submitter": "John Smith <jsmith@mycompany.com>",
                     "pdc_verify_cert": true,
-                    "pdc_component_df_label": "BZComponent",
+                    "pdc_component_df_label": "com.redhat.component",
                     "pdc_contact_role": "Devel_Owner"
                 }
         }]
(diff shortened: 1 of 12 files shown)


twaugh:
I think this needs to add support for the new label set without removing support for the old.

mmilata:
Do you mean that we shouldn't add the new tags when only old tags are present (as described in #342) and make other plugins work with both sets instead?

twaugh:
No, I didn't understand #342 properly. I do now, looks fine.

jpopelka:
LGTM once you rename `rename_labels` -> `aliases` in `test_add_dockerfile`

mmilata:
Updated. I should probably read the code I wrote sometimes ...

mmilata:
Added another commit that makes `bump_release` detect the label that it should use. The reasons are outlined in https://github.com/projectatomic/osbs-client/pull/266#issuecomment-146520158. I'll squash it to the previous commit when ready to merge.

jpopelka:
LGTM

mmilata:
Squashed & rebased.

twaugh:
Should this still say '[do not merge]' in the title?

mmilata:
I'm not sure, what do you think @TomasTomecek?

mmilata:
I've added commit that allows plugins to use the old label if the new one is not present. This PR can be merged now without breaking builds initiated by osbs without projectatomic/osbs-client#266 (currently `add_dockerfile` is run before `add_labels`).

Landscape seems to complain about false positives here?

TomasTomecek:
This looks really good. LGTM

PR #349, opened 2015-10-13
Do all image pushing from tag_and_push plugin

mmilata:
Gets rid of the pushing logic in a-r core, `tag_and_push` is now always used instead.

The `input_osv3` plugin is modified to insert the plugin automatically in order to be backwards compatible with older osbs-client. The same is done for python API.

The tags created during pushing to registry (e.g. `localhost:5000/my/image:latest`) are removed in `exit_remove_built_image` plugin.

### DIFF
--- a/atomic_reactor/plugins/input_osv3.py
+++ b/atomic_reactor/plugins/input_osv3.py
@@ -37,37 +37,10 @@ class OSv3InputPlugin(InputPlugin):
         git_ref = os.environ.get('SOURCE_REF', None)
         image = os.environ['OUTPUT_IMAGE']
         target_registry = os.environ.get('OUTPUT_REGISTRY', None)
-        plugins_json = os.environ.get('DOCK_PLUGINS', '{}')
-        plugins_json = json.loads(plugins_json)
+        self.plugins_json = os.environ.get('DOCK_PLUGINS', '{}')
+        self.plugins_json = json.loads(self.plugins_json)
 
-        plugins_json.setdefault('prebuild_plugins', [])
-
-        # XXX: Remove the two try-except blocks after Sep 2015
-        try:
-            pull_plugin = [p for p in plugins_json['prebuild_plugins']
-                           if p.get('name', None) == PullBaseImagePlugin.key][0]
-        except IndexError:
-            self.log.warning("%s is missing in prebuild_plugins - please update your osbs-client!",
-                             PullBaseImagePlugin.key)
-            pull_plugin = { "name": PullBaseImagePlugin.key }
-            plugins_json['prebuild_plugins'].insert(0, pull_plugin)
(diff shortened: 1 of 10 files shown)


atomic_reactor/plugins/input_osv3.py
```diff
@@ -36,55 +37,71 @@ def run(self):
...
+        if not push_plugin and self.target_registry:
+            self.log.warning("%s is missing in postbuild_plugins - please update your osbs-client!",
+                             TagAndPushPlugin.key)
+            push_plugin = { "name": TagAndPushPlugin.key }
+            self.plugins_json['postbuild_plugins'].insert(0, push_plugin)
+
+        if push_plugin and self.target_registry:
+            push_plugin.setdefault('args', {}).setdefault('registries', {})
+            if not push_plugin['args']['registries']:
+                push_plugin['args']['registries'][self.target_registry] = {"insecure": True}
```
twaugh: So we don't add the target registry if any are already defined?

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -20,24 +22,40 @@ class TagAndPushPlugin(PostBuildPlugin):
...
-        self.log.warning("ignoring arguments '%s'", kwargs)
+
+        self.registries = deepcopy(registries)
 
     def run(self):
         pushed_images = []
-        for registry in self.workflow.push_conf.docker_registries:
+
+        if not self.workflow.tag_conf.unique_images:
+            self.workflow.tag_conf.add_unique_image(self.workflow.image)
```
twaugh: So we don't add `self.workflow.image` if there are any unique images configured? (Just checking)

atomic_reactor/plugins/input_osv3.py
mmilata: The reasoning here was that we'll move away from using the target registry (defined in the openshift env variable) eventually, can't really see the use case for using both configuration sources with different registries.

twaugh: OK, sounds good.

atomic_reactor/plugins/post_tag_and_push.py
mmilata: Couldn't think of a reason to have multiple unique images ... though now I see that this deviates from the previous behavior where, with prod.json, we first pushed `self.workflow.image` in `DockerBuildWorkflow` and then the label-based unique tag in the plugin?

twaugh: No, I don't think the label-based tag goes in unique_images.

So I think you're right, there is only one unique image (as the name suggests!).

twaugh:
LGTM

PR #357, opened 2015-10-20
bump_release: support git < 2.3 by using GIT_SSH instead of GIT_SSH_COMMAND

twaugh:
This is required for build roots using e.g. RHEL-7.


jpopelka:
LGTM

mmilata:
LGTM. Just curious, why create temporary directory and then a file in it instead of just creating temporary file?

twaugh:
For one reason only: `ps` will show a useful name for arg 0.

mmilata:
That's good reason. Thanks!

PR #359, opened 2015-10-22
Spec fixes & move exit_sendmail into rebuilds subpackage

jpopelka:
As @TomasTomecek noticed the exit_sendmail plugin requires pre_check_and_set_rebuild from rebuilds subpackage, so it should probably be moved into rebuilds too.


TomasTomecek:
very nice!!

mmilata:
LGTM, merged!

PR #360, opened 2015-10-23
Fix sendmail plugin in reaction to PDC improvements and API changes

bkabrda:


atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -191,30 +193,17 @@ def _get_receivers_list(self):
...
 
-        if len(send_to) == 0:
+        if contacts['count'] == 0:
             self.log.error('no %s role for the component', self.pdc_contact_role)
             raise RuntimeError('no %s role for the component' % self.pdc_contact_role)
 
+        send_to = []
+        for contact in contacts['results']:
+            send_to.append(contact['contact']['email'])
+
```
twaugh: I'd be tempted to put the 'count' check here, and make sure `send_to` isn't empty at the same time.

bkabrda: Not sure what you mean. Line 199 checks that the contact count is non-zero, so unless PDC is intentionally lying (which I hope will never be the case ;)), we can always be sure that at this point there is at least one contact.

twaugh: I guess I'm just used to dealing with CUPS ;-) but I've always found it useful to perform easy checks like that on the client end rather than always trusting the server.

OK, happy with it as it is.

twaugh:
LGTM

PR #362, opened 2015-10-29
New pulp_sync plugin

twaugh:
This is for transferring repository content from a Docker V2 Registry to a Pulp instance.

### DIFF
--- a/atomic_reactor/plugins/post_pulp_sync.py
+++ b/atomic_reactor/plugins/post_pulp_sync.py
@@ -1,0 +1,169 @@
+"""Copyright (c) 2015 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+
+Sync built image to pulp registry using Docker Registry HTTP API V2
+
+Authentication is via a key and certificate in a secret which the
+builder service account is allowed to mount:
+
+$ oc secrets new pulp ./pulp.key ./pulp.cer
+secrets/pulp
+$ oc secrets add serviceaccount/builder secret/pulp --for=mount
+
+In the BuildConfig for atomic-reactor, specify the secret in the
+strategy's 'secrets' array, specifying a mount path:
+
+"secrets": [{
+  "secretSource": {
+    "name": "pulp"
+  },
+  "mountPath": "/var/run/secrets/pulp"
+}]
+
+In the configuration for this plugin, specify the same path for
+pulp_secret_path:
+
+"pulp_sync": {
+  "pulp_registry_name": ...,
+  ...
+  "pulp_secret_path": "/var/run/secrets/pulp"
+}
+
+"""
+
+from __future__ import print_function, unicode_literals
+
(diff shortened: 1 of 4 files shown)


mmilata:
LGTM!

jpopelka:
LGTM

atomic_reactor/plugins/post_pulp_sync.py
```diff
@@ -0,0 +1,169 @@
...
+# let's silence warnings from dockpulp: there is one warning for every
+# request which may result in tenths of messages: very annoying with
+# "module", it just prints one warning -- this should balance security
+# and UX
+from warnings import filterwarnings
+filterwarnings("module")
+
+
+@contextmanager
+def dockpulp_config(docker_registry, **kwargs):
```
TomasTomecek: would be awesome to have support for this in dockpulp

atomic_reactor/plugins/post_pulp_sync.py
```diff
@@ -0,0 +1,169 @@
...
+        self.workflow.push_conf.add_pulp_registry(self.pulp_registry_name,
+                                                  pulp_registry)
+
+        images = []
+        self.log.info("syncing from docker V2 registry %s",
+                      self.docker_registry)
+        with dockpulp_config(docker_registry=self.docker_registry) as config:
+            for image in self.workflow.tag_conf.primary_images():
+                self.log.info("syncing %s", image.pulp_repo)
+                pulp.syncRepo(config.env, image.pulp_repo,
```
TomasTomecek: what does this do precisely? will it sync all tags? if so, we need to push image with all tags to distribution

twaugh: The `tag_and_push` plugin handles that part.

twaugh: Oh, maybe you mean: 'does sync only copy over newer tags or also delete tags not present in the docker registry?'

TomasTomecek: hmmm, I hope it won't delete, just update; b/c when we wipe distribution's repos they will contain only a single build, so pulp would remove references to previous builds

atomic_reactor/plugins/post_pulp_sync.py
twaugh: Support for creating a temporary config file?

atomic_reactor/plugins/post_pulp_sync.py
twaugh: I will find out the answer. I'd been under the impression it was more like an 'import'.

twaugh: Confirmed: it is only additive.

atomic_reactor/plugins/post_pulp_sync.py
TomasTomecek: configuration via API

atomic_reactor/plugins/post_pulp_sync.py
TomasTomecek: cool

PR #364, opened 2015-10-30
Update koji_promote for metadata spec changes

twaugh:
The output.extra.docker section of the metadata now has a string list `repositories` instead of `destination_repo` and `tag` fields.

Some of the metadata values need to be integers: `start_time`, `end_time`, `epoch` for component RPMs (unless it's None).


jpopelka:
LGTM

PR #365, opened 2015-10-30
koji_promote: report all pushed images

twaugh:
When reporting pushed images, don't just look at the results of the `pulp_push` plugin.

For v1: report only `pulp_push` results
For v1+v2: report `pulp_push` and `pulp_sync` results
For v2: report only `pulp_sync` results

As osbs-client will configure pulp_push and pulp_sync according to whether v1 and/or v2 is in use, all we need to do is collect the results from both pulp plugins.

If pulp is not used, report `tag_and_push` results (now modifed to return `ImageName` instances).

### DIFF
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -470,7 +470,7 @@ class KojiPromotePlugin(ExitPlugin):
         release = None
         for image_name in self.workflow.tag_conf.primary_images:
             if '-' in image_name.tag:
-                name = image_name.repo
+                name = image_name.to_str(registry=False, tag=False)
                 version, release = image_name.tag.split('-', 1)
 
         if name is None or version is None or release is None:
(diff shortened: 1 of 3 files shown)


jpopelka:
Shouldn't the `TestKojiPromote` be now skipped also `if PULP_SYNC_KEY is None` ?
i.e. 
```
@pytest.mark.skipif(PULP_PUSH_KEY is None or PULP_SYNC_KEY is None,
                    reason="requires pulp_push and pulp_sync")
class TestKojiPromote(object):
```

twaugh:
Oh, true. In practice they'll either both be or neither, but I'll change it anyway.

jpopelka:
LGTM

TomasTomecek:
what about metadata exit plugin?

twaugh:
Because `pulp_sync` adds to the `PushConf` for the workflow, the metadata plugin picks it up automatically.

...which suggests that's how `koji_promote` ought to be doing it as well.

TomasTomecek:
@twaugh thanks Tim; LGTM then

twaugh:
I've updated the `koji_promote` plugin to use TagConf+PushConf instead of looking through plugin results.

I also added two other fixes in separate commits: I think the metadata "name" field needs to include the image's namespace as well, and the tests did not run correctly in Python 3.

jpopelka:
LGTM

twaugh:
Thanks. Re-basing and squashing fix-fix commits.

TomasTomecek:
LGTM

PR #368, opened 2015-11-06
spec: %check section

jpopelka:


TomasTomecek:
LGTM (except that this won't work in rhel, but there's nothing we can do about it -- it will work in fedora which is a very nice thing)

jpopelka:
rebased & merged

PR #370, opened 2015-11-11
pulp_sync: get certificates from source secret

mmilata:
I haven't tested it because I'm not sure how. If you know please get in touch.

Related osbs PR: projectatomic/osbs-client#306.

### DIFF
--- a/atomic_reactor/plugins/post_pulp_sync.py
+++ b/atomic_reactor/plugins/post_pulp_sync.py
@@ -127,10 +127,17 @@ class PulpSyncPlugin(PostBuildPlugin):
         if self.username and self.password:
             # Use username and password if provided
             pulp.login(self.username, self.password)
-        elif self.pulp_secret_path:
+        elif self.pulp_secret_path or 'SOURCE_SECRET_PATH' in os.environ:
+            if self.pulp_secret_path is not None:
+                path = self.pulp_secret_path
+                self.log.info("using configured path %s for secrets", path)
+            else:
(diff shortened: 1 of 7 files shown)


twaugh:
LGTM. For testing you need to set a secret:

First, log into the pulp server with `dockpulp -s ... login --username=... --password=...`.

After that, create a secret that the builder service account can use for mounting:
```
oc secret new pulp pulp.cer=~/.pulp/pulp.cer pulp.key=~/.pulp/pulp.key
oc secrets add serviceaccount/builder secrets/pulp --for=mount
```

Then you need to configure osbs.conf. Leave out `openshift_required_version` so that it defaults to using `sourceSecret`, and set `pulp_secret = pulp` and `pulp_registry_name = ...` (the value of the `-s ...` parameter you used for `dockpulp login`).

To enable pulp_sync instead of pulp_push, set `registry_api_versions = v2`. (To enable pulp_push as well it would be `v1,v2`).

Make sure the buildroot has your `/etc/dockpulp.conf` installed.

mmilata:
@twaugh Thanks! It seems that you also need two registries with explicit version, e.g. ```registry_uri = localhost:5000/v1,localhost:5000/v2```.

The testing run went fine so this PR is ready for review.

twaugh:
Oh, right, I tripped over that myself when doing `pulp_sync` testing as well, which is the reason for https://github.com/projectatomic/osbs-client/pull/307 :-)

LGTM, merging.

PR #377, opened 2015-11-13
add integration test: dockpulp, push

TomasTomecek:
can't finish b/c dev pulp is timing out


TomasTomecek:
PTAL

mmilata:
LGTM, however my docker-py throws `APIError` (which is not child of `DockerException`) when the image is missing. If the commit looks good to you I'll merge this.

TomasTomecek:
Looks good to me, thanks for the fix!

PR #379, opened 2015-11-15
Separate checksums counting code into a function and test it.

jpopelka:
The checksums in test should be ok:
```
$ echo -n "abc" | md5sum 
900150983cd24fb0d6963f7d28e17f72
$ echo -n "abc" | sha256sum
ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad
```


twaugh:
LGTM

PR #381, opened 2015-11-16
Revert "koji_promote: name value should include namespace"

twaugh:
This reverts commit 404836e23ab16c20d2d9e0d2cb08665f4fa6f1dd and fixes up the tests.

We should not be including the namespace in this field (spotted by @pbabinca).


jpopelka:
LGTM, merging

PR #382, opened 2015-11-17
Fix label alias overrides

twaugh:
When the base image has the new name for a label and the Dockerfile has the old name, we must add the new name in as well in order to override the inherited value.

If we don't do this, built images will have names corresponding to their parents.


twaugh:
Test coverage goes down because the 'should probably have the same value' path is now only taken when the Dockerfile explicitly sets both old and new labels for an alias to different values, which isn't tested.

mmilata:
Good catch! Merging ...

PR #390, opened 2015-11-25
squash: don't load result if dont_load==True

mmilata:
docker-scripts default is to load the image: https://github.com/goldmann/docker-scripts/blob/0.4.4/docker_scripts/squash.py#L38


jpopelka:
Agree, LGTM.

TomasTomecek:
Very nice! LGTM

TomasTomecek:
I missclicked -,-

twaugh:
Oh, *that*'s what causes it!

PR #391, opened 2015-11-25
docs: update garbage collection document

mmilata:
- remove_built_image is now an exit plugin
- mention squash plugin parameters


mmilata:
@jpopelka, sorry, I overwrote your comments when force-pushing fix for the first one ...

I think that if you call the plugin with `dont_load=True` and `remove_former_image=False` then the original image gets deleted in the `remove_built_image` plugin (because it is stored in `self.workflow.builder.image_id`).

jpopelka:
OK, thanks. LGTM

TomasTomecek:
LGTM

PR #393, opened 2015-11-26
core,push: use wait_for_command for parsing logs

TomasTomecek:
I've just realized this slightly conflicts with @twaugh's patch: https://github.com/projectatomic/atomic-reactor/commit/a770fdc20efda4b46ff8a9baf2031dd415994fff

### DIFF
--- a/atomic_reactor/core.py
+++ b/atomic_reactor/core.py
@@ -457,15 +457,11 @@ class DockerTasker(LastLogger):
             # because changing api is fun
             logs = self.d.push(image.to_str(tag=False), tag=image.tag, stream=True)
 
-        parsed_logs = []
-        for line in logs:
-            try:
-                j = json.loads(line.decode('utf-8'))
-            except (ValueError, UnicodeDecodeError):
-                logger.error("cannot parse push logs line: %r", line)
-            else:
-                parsed_logs.append(j)
-        return parsed_logs
+        command_result = wait_for_command(logs)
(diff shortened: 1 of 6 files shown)


twaugh:
Maybe `push_image` could raise an exception if `command_result.is_failed()`? Then the plugin change I made could be backed out.

TomasTomecek:
great idea, will incorporate

TomasTomecek:
PTAL

twaugh:
Now `DockerTasker.push_image()` returns unparsed logs. It used to return parsed logs. The `tag_and_push` plugin needs parsed logs in order to find out the manifest digest.

For the tests to pass we'd need to change `tests/plugins/test_tag_and_push.py` so that it uses `iter(logs)` instead of `logs` when mocking docker-py's push method, and `wait_for_command` so that it adds `parsed_item` instead of `item` to the logs list.

TomasTomecek:
Thanks Tim, will update.

TomasTomecek:
Done.

(expected this to be a small patch)

will add more tests to make coveralls happy

twaugh:
With this change we could also remove the explicit error checking from post_tag_and_push (see last commit).

TomasTomecek:
didn't realize; very nice Tim!

TomasTomecek:
can we merge this now?

twaugh:
LGTM

TomasTomecek:
Thanks Tim!

PR #394, opened 2015-11-30
plugin: fix log level when running exit plugins

TomasTomecek:


twaugh:
This message is intended to warn about a client expecting a postbuild plugin when it's really an exit plugin. Maybe it should be at level info but be clearer about what's up?

PR #395, opened 2015-12-01
koji CG metadata: include arch in filename

twaugh:


jpopelka:
LGTM

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -401,7 +404,7 @@ def add_log_type(output):
         repositories = [image.to_str() for image in output_images
                         if image.tag != 'latest']
         arch = os.uname()[4]
```
TomasTomecek: I hope brew is fine with this and we won't have to do some translation (in future it would be for the best if brew provided the arch)

TomasTomecek:
LGTM

atomic_reactor/plugins/exit_koji_promote.py
twaugh: Oh, right, that's true. And it already does provide the arch to osbs-client so it can fill in the architecture label.

mmilata:
LGTM, merged.

PR #389, opened 2015-11-24
add_labels_in_df: comprehensive tests, don't use aliases by default

mmilata:
Fixes #388.

PR #396, opened 2015-12-11
Revert to using old label set in plugins

mmilata:
Since we don't use the new labels and label aliases are turned off (#389), prod builds end with following error:

```
2015-12-10 21:32:54,469 - atomic_reactor.plugin - DEBUG - Traceback
(most recent call last):
  File "/usr/lib/python2.7/site-packages/atomic_reactor-1.6.0-py2.7.egg/atomic_reactor/plugin.py",
line 203, in run
  File "/usr/lib/python2.7/site-packages/atomic_reactor-1.6.0-py2.7.egg/atomic_reactor/plugins/post_tag_by_labels.py", line 49, in run
    name = get_label("name")
  File "/usr/lib/python2.7/site-packages/atomic_reactor-1.6.0-py2.7.egg/atomic_reactor/plugins/post_tag_by_labels.py", line 47, in get_label
    raise RuntimeError("Missing label '%s'." % label_name)
RuntimeError: Missing label 'name'.
```

This reverts commit 605314d. The pull request that introduced it (#349) also contains another commit (5d20f7b) which uses the new labels only if available, thus I'm leaving it unreverted (good idea?).

### DIFF
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -41,7 +41,7 @@ class SendMailPlugin(ExitPlugin):
                     # optional arguments follow
                     "submitter": "John Smith <jsmith@mycompany.com>",
                     "pdc_verify_cert": true,
-                    "pdc_component_df_label": "com.redhat.component",
+                    "pdc_component_df_label": "BZComponent",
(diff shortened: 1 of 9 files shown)


twaugh:
Wow, these labels are hard.

LGTM

PR #399, opened 2015-12-14
import_image: retry up to 3 times

twaugh:
Retries only occur with OSBS support, see projectatomic/osbs-client#322.


mmilata:
LGTM, merged. I like how it doesn't create a dependency on newer version of OSBS.

twaugh:
Thanks! Yes, nice to avoid extra dependencies where possible.

PR #400, opened 2015-12-14
koji_promote: include v2 digests in metadata

mmilata:
@twaugh, in your original patch there is some logic to avoid duplicates in the list of repositories - can this happen? Do we expect to push to one registry multiple times?

### DIFF
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -358,7 +358,27 @@ class KojiPromotePlugin(ExitPlugin):
 
         return metadata, output
 
-    def get_output_images(self):
+    def get_digests(self):
+        """
+        Returns a map of repositories to digests
+        """
+
+        digests = {}  # repository -> digest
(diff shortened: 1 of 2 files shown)


twaugh:
Nice!

But yes, when we're using both V1 and V2 we can end up generating this metadata:
```
                "docker": {
                    "id": "123456imageid",
                    "parent_id": "123456parent-id",
                    "repositories": [
                        "pulp.example.com/ns/name:1.0-1",
                        "pulp.example.com/ns/name@sha256:0000000000000000000000000000000d",
                        "pulp.example.com/ns/name:1.0-1",
                        "pulp.example.com/ns/name@sha256:0000000000000000000000000000000d"
                    ]
                },
```

(try `py.test-3 -k promote_success -v tests -s`)

twaugh:
We should maybe add a check for duplicates in the test-case, come to think of it.

mmilata:
Added a commit to avoid duplicates. Can squash it into the previous one if you want.

twaugh:
LGTM. Happy for you to squash and merge.

PR #401, opened 2015-12-22
Fix a couple spec file items to fix builds for Fedora

maxamillion:
add git and python-backports-lzma BuildRequires for tests
exclude python sitelib integration-tests


twaugh:
LGTM

PR #406, opened 2016-01-19
plugins: don't pass namespace arg to OSBS methods

mmilata:
Blocked by:
* projectatomic/osbs-client#334
* bkabrda/flexmock#14 (need to wait for new release so that we may use it on Travis)

### DIFF
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -24,7 +24,7 @@ from atomic_reactor.source import GitSource
 from atomic_reactor.plugins.post_rpmqa import PostBuildRPMqaPlugin
 from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
 from atomic_reactor.constants import PROG
-from atomic_reactor.util import get_version_of_tools, get_checksums
+from atomic_reactor.util import get_version_of_tools, get_checksums, get_build_json
 from osbs.conf import Configuration
 from osbs.api import OSBS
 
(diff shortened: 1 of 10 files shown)


mmilata:
Ready to merge.

twaugh:
I think there are two useful things to do:

 * pass `namespace` to the `Configuration` constructor
 * stop passing `namespace` to the various API functions

The first one makes sense to do now, whlie the second one introduces incompatible behaviour (i.e. requires newer osbs-client in order to work the same).

mmilata:
You're right. Will update the patch.

mmilata:
I've updated the patch to only pass `namespace` to `Configuration` constructor.

When/how we should do the other thing?

twaugh:
LGTM

So now atomic-reactor will work correctly with osbs-client versions since https://github.com/projectatomic/osbs-client/pull/334 when using a non-default namespace.

I guess we don't want to stop passing `namespace` keyword arguments to API calls until there's no chance of using an osbs-client version older than when the namespace change was pulled in. Before that change, the `namespace` configuration parameter was ignored. :-(

TomasTomecek:
lgtm, very nice work Martin!

twaugh:
I think we should merge this and open a new issue to track the removal of `namespace` keyword arguments for API functions.

twaugh:
So, this PR is no longer blocked by the osbs-client change. What about flexmock?

mmilata:
Travis has the new flexmock so the test here pass. The test also pass on my machine with the old flexmock, which is sorta weird ... whoever is going to review it, can you perhaps run the tests first (under Python 3) and let me know if they don't pass?

twaugh:
Running `py.test-3 tests` with python-flexmock-0.9.7-2.fc23 here gives failures:

```
tests/plugins/test_check_and_set_rebuild.py FFFFFF
tests/plugins/test_import_image.py ....FFF.F...
tests/plugins/test_store_metadata.py FF
```

twaugh:
LGTM, merging.

mmilata:
Thanks! The tests should pass with `python-flexmock-0.10.2-1.fc23`, soon to be in updates-testing.

PR #407, opened 2016-01-19
Retry 'FROM' layer pull with library namespace on failure

twaugh:
The library/ prefix is only special for Docker Hub. Stripping it out actually breaks V2 pulls for base images that are published as library/repo.


jpopelka:
I agree, AFAIR it's been there to work-around docker/docker#11955

twaugh:
I've enhanced this PR so that atomic-reactor will now try pulling registry/library/repository if pulling registry/repository fails.

mmilata:
Looks reasonable, merging.

PR #409, opened 2016-01-25
store_metadata: if pulp was used, only report pulp repositories

twaugh:
Otherwise the temporary Distribution pushes are also reported, which we don't want.

cf. koji_import

### DIFF
--- a/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
+++ b/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
@@ -49,6 +49,21 @@ class StoreMetadataInOSv3Plugin(ExitPlugin):
     def get_post_result(self, key):
         return self.get_result(self.workflow.postbuild_results.get(key, ''))
 
+    def get_digests(self):
+        """
+        Returns a map of repositories to digests
+        """
+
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -72,19 +87,26 @@ def get_repositories(self):
             "unique": unique_repositories,
         }
 
-    def get_digests(self):
+    def get_pullspecs(self, digests):
+        if self.workflow.push_conf.pulp_registries:
```
TomasTomecek: This means, that it won't be possible to configure deployment when using distribution and pulp, both as final registries. I think it's totally fine.

TomasTomecek:
LGTM, very nice catch Tim!

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
twaugh: Right. If someone needs this in the future we can make it smarter and match up the dockpulp environment name in config.

mmilata:
LGTM!

PR #412, opened 2016-01-29
Avoid reporting duplicate pulp registries

twaugh:
This can happen when using both v1 and v2 Docker Registry APIs.
Previously the `koji_promote` plugin was explicitly avoiding duplicate registries. This logic has now been moved to `PushConf`.

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -183,7 +183,7 @@ class PushConf(object):
     def __init__(self):
         self._registries = {
             "docker": [],
-            "pulp": [],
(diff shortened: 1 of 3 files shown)


twaugh:
travis-ci/push check only failed for Python 3.5, and that was because it couldn't install Python 3.5 due to some sort of Travis environment issue. The same testing succeeded for travis-ci/pr.

twaugh:
Re-basing for cleaner history and to re-run the Travis tests.

TomasTomecek:
nice! lgtm

mmilata:
LGTM, merging!

PR #414, opened 2016-02-02
pulp_push: omit redirect-url when creating repositories on demand

twaugh:
When creating a Pulp repository on demand, do not set a redirect URL. By leaving as the default value, redirects will work correctly for both V1 and V2 with the default Pulp configuration (i.e. not CDN).

Note that `dockpulp.conf` will need an additional section to allow this, something like:

```
[redirect]
myenv = no
```

(see https://github.com/release-engineering/dockpulp/commit/da3260229c738355f9e32790de331f939b92ae1c)


twaugh:
During testing of this I saw a dockpulp traceback, will update when I have more info.

twaugh:
Re-tested, this time using the correct version of dockpulp. :-)

Works fine for me.

mmilata:
Does what it promised, merging.

PR #421, opened 2016-03-01
plugins: remove namespace keyword argument after a change in https://github.com/projectatomic/osbs-client/pull/334

### LINKED ISSUE #413: osbs-client warns about namespace keyword argument in API functions
The `namespace` keyword argument for OSBS API functions is deprecated. Instead, `namespace` should be specified in the `Configuration`.

Unfortunately, until osbs-client-0.16, the configuration-specified `namespace` was not honoured.

When we think atomic-reactor will only be used with osbs-client >= 0.16, we should remove the `namespace` keyword arguments from OSBS API calls.

See also #406, which introduced the change specifying `namespace` for the configuration.

vrutkovs:
Fixes #413


TomasTomecek:
lgtm once we are sure that osbs-client >= 0.16 is available in f23+ and internally

twaugh:
LGTM!

PR #426, opened 2016-03-07
compress: log compression ratio

### LINKED ISSUE #423: compress: log original file size or compression ratio
It's hard to know how well the compression is performing without seeing the compression ratio achieved.

mmilata:
Closes #423.
Or do we want the uncompressed size in metadata as well?

### DIFF
--- a/atomic_reactor/plugins/post_compress.py
+++ b/atomic_reactor/plugins/post_compress.py
@@ -17,7 +17,7 @@ import os
 
 from atomic_reactor.constants import EXPORTED_COMPRESSED_IMAGE_NAME_TEMPLATE
 from atomic_reactor.plugin import PostBuildPlugin
-from atomic_reactor.util import get_exported_image_metadata
+from atomic_reactor.util import get_exported_image_metadata, human_size
 
 
 class CompressPlugin(PostBuildPlugin):
@@ -49,6 +49,7 @@ class CompressPlugin(PostBuildPlugin):
         super(CompressPlugin, self).__init__(tasker, workflow)
         self.load_exported_image = load_exported_image
         self.method = method
+        self.uncompressed_size = 0
 
     def _compress_image_stream(self, stream):
         outfile = os.path.join(self.workflow.source.workdir,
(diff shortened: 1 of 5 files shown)


atomic_reactor/plugins/post_compress.py
```diff
@@ -85,5 +88,12 @@ def run(self):
...
                 outfile = self._compress_image_stream(image_stream)
-        self.workflow.exported_image_sequence.append(get_exported_image_metadata(outfile))
+        metadata = get_exported_image_metadata(outfile)
+        self.workflow.exported_image_sequence.append(metadata)
+
+        compressed_size = metadata['size']
+        if compressed_size != 0:
+            self.log.debug('uncompressed: %s, compressed: %s, ratio: %.2f',
+                human_size(self.uncompressed_size), human_size(compressed_size),
+                self.uncompressed_size/float(compressed_size))
```
twaugh: Nice! Does this ratio calculation agree with what `gzip -l` says though? I think it should be size_decrease / uncompressed_size.

twaugh:
It would be useful to also have that information in the metadata, yes: both compressed and uncompressed size. There might be some small overlap if/when we include full koji metadata in the OSv3 metadata.

atomic_reactor/util.py
```diff
@@ -567,3 +567,12 @@ def get_build_json():
     except KeyError:
         logger.error("No $BUILD env variable. Probably not running in build container")
         raise
+
+# copypasted and slightly modified from
+# http://stackoverflow.com/questions/1094841/reusable-library-to-get-human-readable-version-of-file-size/1094933#1094933
+def human_size(num, suffix='B'):
```
TomasTomecek: don't want to pick the nits so hard but this could use some tests too :)

mmilata:
Sorry, I blindly used the first formula from [wikipedia](https://en.wikipedia.org/wiki/Data_compression_ratio). It looks like compression tools use the second variant (which is more intuitive anyway). Updated the branch to use the other formula, store uncompressed size in metadata, and test the `human_size` function as requested.

twaugh:
LGTM! Merging.

Issue #425, opened 2016-03-05
Store failure reason in metadata


twaugh:
When a build fails, `store_metadata_in_osv3` should store the reason for that failure.

Eg plugin name during which the failure occurred, or whether it was during docker build. Maybe also an error name or description.

TomasTomecek:
This would improve user experience, definitely. Usually it's pretty obvious for us, developers of the tool, who saw them logs thousands of times, but not so obvious for the rest of the world. So +1 on this.

twaugh:
Yes. So in addition to putting the plugin name (e.g. squash) and error name (e.g. SquashFailure), we should also make the final traceback output in the log much better. It isn't clear that this is a repeat of an earlier traceback.

PR #427, opened 2016-03-09
Validate build.json against a simple schema

### LINKED ISSUE #14: validate input json
We should write validator for build.json. Then it will be very simple to precisely describe input format of the build.json (and document it). We could use [json schema](http://json-schema.org/example1.html) for that matter.

vrutkovs:
Fixes #14

TODO:
- [x] Tests
- [x] Check RHEL's package and update .spec file

### DIFF
--- a/atomic_reactor/constants.py
+++ b/atomic_reactor/constants.py
@@ -14,6 +14,77 @@ BUILD_JSON = 'build.json'
 BUILD_JSON_ENV = 'BUILD_JSON'
 RESULTS_JSON = 'results.json'
 
+JSON_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "osbs_client": {"type": "string"},
+        "image": {"type": "string"},
+        "git_url": {"type": "string"},
+        "source": {
+            "type": "object",
+            "properties": {
+                "provider": {"type": "string"},
+                "url": {"type": "string"}
+            },
+            "required": ["provider"]
+        },
+        "prebuild_plugins": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string"},
+                    "is_allowed_to_fail": {"type": "boolean"},
+                    "required": {"type": "boolean"},
+                    "args": {"type": "object"},
+                },
+                "required": ["name"]
+            }
+        },
+        "postbuild_plugins": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string"},
+                    "is_allowed_to_fail": {"type": "boolean"},
+                    "required": {"type": "boolean"},
+                    "args": {"type": "object"},
+                },
+                "required": ["name"]
+            }
+        },
+        "prepublish_plugins": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string"},
+                    "is_allowed_to_fail": {"type": "boolean"},
+                    "required": {"type": "boolean"},
+                    "args": {"type": "object"},
+                },
+                "required": ["name"]
+            }
+        },
+        "exit_plugins": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string"},
+                    "is_allowed_to_fail": {"type": "boolean"},
+                    "required": {"type": "boolean"},
+                    "args": {"type": "object"},
+                },
+                "required": ["name"]
+            }
+        },
+    },
+    "required": ["image"]
+}
+
+
 CONTAINER_SHARE_PATH = '/run/share/'
 CONTAINER_SHARE_SOURCE_SUBDIR = 'source'
 CONTAINER_SECRET_PATH = ''
--- a/atomic_reactor/core.py
+++ b/atomic_reactor/core.py
@@ -32,9 +32,10 @@ import requests
 import docker
 from docker.errors import APIError
 from docker.utils import create_host_config
+from jsonschema import validate
 
 from atomic_reactor.constants import CONTAINER_SHARE_PATH, CONTAINER_SHARE_SOURCE_SUBDIR,\
-        BUILD_JSON, DOCKER_SOCKET_PATH
+        BUILD_JSON, DOCKER_SOCKET_PATH, JSON_SCHEMA
 from atomic_reactor.source import get_source_instance_for
 from atomic_reactor.util import (
     ImageName, wait_for_command, clone_git_repo, figure_out_dockerfile, Dockercfg)
@@ -95,6 +96,8 @@ class BuildContainerFactory(object):
         build_json_path = os.path.join(local_path, BUILD_JSON)
         with open(build_json_path, 'r') as fp:
             build_json = json.load(fp)
+
+        validate(build_json, JSON_SCHEMA)
         source = get_source_instance_for(build_json['source'], tmpdir=local_path)
         if source.provider == 'path':
             logger.debug('copying source from %s to %s', source.schemeless_path, local_path)
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -11,12 +11,16 @@ from __future__ import print_function, unicode_literals
 import logging
 import os
 import sys
+import jsonschema
+from collections import deque
 
 import pytest
 
 from atomic_reactor.buildimage import BuildImageBuilder
 from atomic_reactor.core import DockerTasker
 import atomic_reactor.cli.main
+
+from tempfile import NamedTemporaryFile
 
 from tests.fixtures import is_registry_running, temp_image_name, get_uuid
 from tests.constants import LOCALHOST_REGISTRY, DOCKERFILE_GIT, DOCKERFILE_OK_PATH, FILES, MOCK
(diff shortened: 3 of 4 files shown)


TomasTomecek:
Wow, this is neat!

A couple of points:
 * a test would be greatly appreciated
 * what's the behavior of `validate(build_json, JSON_SCHEMA)`: will it throw an exception if json's not valid? will we be able to figure out easily what's wrong?
 * luckily it seems that `python-jsonschema` is in rhel extras; could you please test this works with latest version there?
 * can you add the dependency to spec also please?

what do other maintainers think?

vrutkovs:
>a test would be greatly appreciated

Right, will add those in the following commit

>what's the behavior of validate(build_json, JSON_SCHEMA): will it throw an exception if json's not valid? will we be able to figure out easily what's wrong?

Yes, it will throw smth like:
```
>>> validate(
...     {"name" : "Eggs", "price" : "Invalid"}, schema
... ) 
    ...
ValidationError: 'Invalid' is not of type 'number'
```
I'll see if we can get an actual line:column from the file.

>Luckily it seems that python-jsonschema is in rhel extras; could you please test this works with latest version there?

I used the one from pypi, I'll check if it works with the one from rhel and add a spec dependency.

atomic_reactor/constants.py
```diff
@@ -14,6 +14,46 @@
...
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string"},
+                    "is_allowed_to_fail": {"type": "boolean"},
+                    "args": {"type": "object"},
+                }
+            }
+        },
```
twaugh: We'll need to add "prepub_plugins" and "exit_plugins" too I guess?

tests/test_cli.py
```diff
@@ -154,21 +156,54 @@ def test_building_from_json_source_provider(self, is_registry_running, temp_imag
...
+        {'valid_json': True, 'valid_schema': False, 'contents':
+            '{ "this": "is", "json": "but", "its": "invalid" }'},
+        {'valid_json': True, 'valid_schema': True, 'contents':
+            '{"image": "image_name", "source": {"provider": "path", "uri": "bar"}}'},
+    ])
+    def test_json_validation(self, is_registry_running, temp_image_name, build_json):
         if MOCK:
             mock_docker()
 
+        fd, temp_json_file_path = tempfile.mkstemp()
```
twaugh: How about `from tempfile import NamedTemporaryFile` and `with NamedTemporaryFile(...) as ...:`?

atomic_reactor/constants.py
vrutkovs: Right, current schema is a simple draft ~~so tests would pass~~, any help making it complete is appreciated

tests/test_cli.py
vrutkovs: Good point, I'll update this one

atomic_reactor/constants.py
twaugh: There are other things too, e.g. "openshift_build_selflink", "provider_params" etc.

How does jsonschema react to extra properties in the JSON than the schema mentions?

vrutkovs: Those are checked for type if present, necessary elements have to be added to ```required```

twaugh:
I like it in general!

I wonder whether it would be useful to put the actual validation in its own plugin. That might make the jsonschema dependency easier to manage for some people. It would also make it easier to figure out what went wrong afterwards, once https://github.com/projectatomic/atomic-reactor/issues/425 is solved.

vrutkovs:
>I wonder whether it would be useful to put the actual validation in its own plugin

Agreed, we can split validation in two parts:
 * Core - e.g. no build triggered if no ```image_name``` specified
 * Plugins - mark plugin as failed if plugin config is missing a key and etc.

This way we can limit this PR to core part

twaugh:
Want to re-base this?

atomic_reactor/constants.py
```diff
@@ -14,6 +14,68 @@
...
+                "url": {"type": "string"}
+            },
+            "required": ["provider"]
+        },
+        "prebuild_plugins": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string"},
```
twaugh: Does anything in the schema mark 'name' as being required?

atomic_reactor/constants.py
```diff
@@ -14,6 +14,68 @@
...
+                    "args": {"type": "object"},
+                }
+            }
+        },
+        "postbuild_plugins": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string"},
```
twaugh: 'name' is required

atomic_reactor/constants.py
```diff
@@ -14,6 +14,68 @@
...
+                    "args": {"type": "object"},
+                }
+            }
+        },
+        "prepublish_plugins": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string"},
```
twaugh: 'name' is required

atomic_reactor/constants.py
```diff
@@ -14,6 +14,68 @@
...
+                    "args": {"type": "object"},
+                }
+            }
+        },
+        "exit_plugins": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string"},
```
twaugh: 'name' is required

atomic_reactor/constants.py
```diff
@@ -14,6 +14,68 @@
 BUILD_JSON_ENV = 'BUILD_JSON'
 RESULTS_JSON = 'results.json'
 
+JSON_SCHEMA = {
+    "type": "object",
+    "properties": {
```
twaugh: `"osbs_client": {"type": "string"},`?

atomic_reactor/constants.py
```diff
@@ -14,6 +14,68 @@
...
+            "required": ["provider"]
+        },
+        "prebuild_plugins": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "name": {"type": "string"},
+                    "is_allowed_to_fail": {"type": "boolean"},
+                    "args": {"type": "object"},
```
twaugh: `"required": {"type": "boolean"},`?

atomic_reactor/core.py
```diff
@@ -95,6 +96,8 @@ def _obtain_source_from_path_if_needed(self, local_path, container_path=CONTAINE
         build_json_path = os.path.join(local_path, BUILD_JSON)
         with open(build_json_path, 'r') as fp:
             build_json = json.load(fp)
+
+        validate(build_json, JSON_SCHEMA)
```
twaugh: Did you come to a conclusion about whether this could be made to give a better pointer to where the problem is?

vrutkovs: Yes, it would be displayed:
```
    def validate(self, *args, **kwargs):
        for error in self.iter_errors(*args, **kwargs):
>           raise error
E           ValidationError: 'provider' is a required property
E           
E           Failed validating 'required' in schema['properties']['source']:
E               {'properties': {'provider': {'type': 'string'},
E                               'url': {'type': 'string'}},
E                'required': ['provider'],
E                'type': 'object'}
E           
E           On instance['source']:
E               {u'uri': u'/atomic-reactor/tests/files/docker-hello-world'}
```

atomic_reactor/constants.py
vrutkovs: Added

atomic_reactor/constants.py
vrutkovs: Yes, outside `properties` objects there's `"required": ["name"]`

atomic_reactor/constants.py
vrutkovs: ditto

atomic_reactor/constants.py
vrutkovs: ditto

atomic_reactor/constants.py
```diff
@@ -14,6 +14,77 @@
 BUILD_JSON_ENV = 'BUILD_JSON'
 RESULTS_JSON = 'results.json'
 
+JSON_SCHEMA = {
```
twaugh: What object is this describing? It doesn't look like a Build, and it doesn't look like the ATOMIC_REACTOR_PLUGINS environment variable either.

FWIW I think it only makes sense to validate ATOMIC_REACTOR_PLUGINS, not the full Build JSON (which OpenShift will already have validated).

vrutkovs: It is validating ATOMIC_REACTOR_PLUGINS env var (its just being called 'build JSON' internally), the one we generate from `prod_inner.json`.

atomic_reactor/core.py
twaugh: Ideally we'd show *all* the problems in the input, not just the first, like this: https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/pre_reactor_config.py#L148

atomic_reactor/constants.py
twaugh: Can we call this plugins.schema and put it in atomic_reactor/schemas? You can retrieve it like this: https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/pre_reactor_config.py#L135

lcarva:
Closing, will open a new PR when we're ready to work on this

PR #430, opened 2016-03-14
Use shallow clone (--depth=1) to save bandswidth 

vrutkovs:
If the command fails (git <=1.9.0), this will fall back to --single-branch


TomasTomecek:
neat improvement

mmilata:
LGTM!

PR #433, opened 2016-03-16
parse_item: use default splitlines() method instead of regexps

vrutkovs:
This eradicates an extra re import and improves performance:
saves about a second during my tests


TomasTomecek:
lgtm

twaugh:
LGTM, nice catch.

PR #442, opened 2016-03-31
Store more metadata to simplify metrics script

### LINKED ISSUE #424: Plugin-level timings stored in OSBS Build metadata
In atomic-reactor, it should note the start time and duration of each plugin, as well as the "docker build" process itself. These figures should be stored as an annotation in the build metadata (by the store_metadata_in_osv3 plugin).

vrutkovs:
Fixes #424, #425

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -12,6 +12,8 @@ Script for building docker image. This is expected to run inside container.
 import json
 import logging
 import tempfile
+import datetime
+import traceback
 
 from atomic_reactor.build import InsideBuilder
 from atomic_reactor.plugin import PostBuildPluginsRunner, PreBuildPluginsRunner, InputPluginsRunner, PrePublishPluginsRunner, \
@@ -255,6 +257,9 @@ class DockerBuildWorkflow(object):
         self.exit_plugins_conf = exit_plugins
         self.prebuild_results = {}
         self.postbuild_results = {}
+        self.plugins_timestamps = {}
+        self.plugins_durations = {}
+        self.plugins_errors = {}
         self.autorebuild_canceled = False
         self.build_failed = False
         self.plugin_failed = False
@@ -325,7 +330,20 @@ class DockerBuildWorkflow(object):
                 self.autorebuild_canceled = True
                 raise
 
+            start_time = datetime.datetime.now()
+            self.plugins_timestamps['dockerbuild'] = start_time.isoformat()
+
             build_result = self.builder.build()
+
+            try:
+                finish_time = datetime.datetime.now()
+                duration = finish_time - start_time
+                seconds = duration.total_seconds()
+                logger.debug("build finished in %ds", seconds)
+                self.plugins_durations['dockerbuild'] = seconds
+            except Exception:
+                logger.debug(traceback.format_exc())
+
             self.build_logs = build_result.logs
 
             self.build_failed = build_result.is_failed()
@@ -333,6 +351,7 @@ class DockerBuildWorkflow(object):
             if build_result.is_failed():
                 # The docker build failed. Finish here, just run the
                 # exit plugins (from the 'finally:' block below).
+                self.plugins_errors['dockerbuild'] = ''
                 return build_result
 
             self.built_image_inspect = self.builder.inspect_built_image()
--- a/atomic_reactor/plugin.py
+++ b/atomic_reactor/plugin.py
@@ -15,6 +15,7 @@ import logging
 import os
 import traceback
 import imp
+import datetime
 
 from atomic_reactor.util import process_substitutions
 
@@ -45,6 +46,8 @@ class Plugin(object):
     key = None
     # by default, if plugin fails (raises exc), execution continues
     is_allowed_to_fail = True
+
+    duration = datetime.timedelta()
 
     def __init__(self, *args, **kwargs):
         """
@@ -156,7 +159,13 @@ class PluginsRunner(object):
         plugin_instance = plugin_class(**plugin_conf)
         return plugin_instance
 
-    def on_plugin_failed(self):
+    def on_plugin_failed(self, plugin=None, exception_repr=None):
+        pass
+
+    def save_plugin_timestamp(self, plugin, timestamp):
+        pass
+
+    def save_plugin_duration(self, plugin, duration):
         pass
 
     def run(self, keep_going=False):
@@ -198,6 +207,8 @@ class PluginsRunner(object):
             logger.debug("running plugin '%s'", plugin_name)
 
             plugin_instance = self.create_instance_from_plugin(plugin_class, plugin_conf)
+            start_time = datetime.datetime.now()
+            self.save_plugin_timestamp(plugin_instance.key, start_time)
 
             try:
                 plugin_response = plugin_instance.run()
@@ -217,11 +228,20 @@ class PluginsRunner(object):
                     if not plugin_is_allowed_to_fail:
                         failed_msgs.append(msg)
                 else:
-                    self.on_plugin_failed()
+                    self.on_plugin_failed(plugin_instance.key, repr(ex))
                     logger.error(msg)
                     raise PluginFailedException(msg)
 
                 plugin_response = ex
+
+            try:
+                finish_time = datetime.datetime.now()
+                duration = finish_time - start_time
+                seconds = duration.total_seconds()
+                logger.debug("plugin '%s' finished in %ds", plugin_name, seconds)
+                self.save_plugin_duration(plugin_instance.key, seconds)
+            except Exception as ex:
+                logger.debug(traceback.format_exc())
 
             self.plugins_results[plugin_instance.key] = plugin_response
         if len(failed_msgs) == 1:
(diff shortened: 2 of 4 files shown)


twaugh:
It's a nice start, but I think we can get all we want without having to parse any logs. I was more thinking that we'd collecting timing data as we're running plugins. Also we already have image size information in the tar_metadata annotation, so we can use it from wherever that comes from.

vrutkovs:
> I was more thinking that we'd collecting timing data as we're running plugins

Yeah, that might be a nice improvement.

> Also we already have image size information in the tar_metadata annotation, so we can use it from wherever that comes from

Right, the string also depends on dockpulp output, which is not a good idea.

vrutkovs:
That should do the trick. Haven't tried it outside of the tests yet though.

@mmilata @twaugh care to take a look?

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -109,6 +110,33 @@ def get_pullspecs(self, digests):
...
+        return '\n'.join(self.workflow.build_logs)
+
+    def _parse_logs(self, regexp):
+        try:
+            return re.compile(regexp).search(self._get_combined_logs()).groups()
+        except:
+            return ['']
+
+    def get_error_message(self):
+        error = self._parse_logs(r'ERROR - .*plugin \'(.*)\' raised an exception: ([^(]*)')
```
vrutkovs: We probably should store exceptions in workflow too

atomic_reactor/plugin.py
```diff
@@ -223,7 +230,12 @@ def run(self, keep_going=False):
 
                 plugin_response = ex
 
+            finish_time = datetime.datetime.now()
+            duration = finish_time - start_time
+            logger.debug("plugin '%s' finished in %ds", plugin_name, duration.seconds)
```
twaugh: I think we need `.total_seconds()` here, not `.seconds`.

atomic_reactor/plugin.py
```diff
@@ -223,7 +230,12 @@ def run(self, keep_going=False):
...
+
             self.plugins_results[plugin_instance.key] = plugin_response
+            self.save_plugin_duration(plugin_instance.key, duration.seconds)
```
twaugh: Same here.

tests/plugins/test_store_metadata.py
```diff
@@ -101,6 +101,11 @@ def test_metadata_plugin(tmpdir):
     workflow.postbuild_results = {
         PostBuildRPMqaPlugin.key: "rpm1\nrpm2",
     }
+    workflow.plugins_durations = {
+        CpDockerfilePlugin.key: "1",
```
twaugh: I guess these would really be floats not strings, e.g. 1.05.

tests/plugins/test_store_metadata.py
```diff
@@ -173,6 +179,11 @@ def test_metadata_plugin_rpmqa_failure(tmpdir):
     workflow.postbuild_results = {
         PostBuildRPMqaPlugin.key: RuntimeError(),
     }
+    workflow.plugins_durations = {
+        CpDockerfilePlugin.key: "1",
```
twaugh: Same here.

atomic_reactor/plugin.py
```diff
@@ -46,6 +47,8 @@ class Plugin(object):
     # by default, if plugin fails (raises exc), execution continues
     is_allowed_to_fail = True
 
+    duration = None
```
twaugh: Is this used?

tests/plugins/test_store_metadata.py
```diff
@@ -94,6 +101,11 @@ def test_metadata_plugin(tmpdir):
     workflow.postbuild_results = {
         PostBuildRPMqaPlugin.key: "rpm1\nrpm2",
     }
+    workflow.plugins_durations = {
+        CpDockerfilePlugin.key: "1",
```
twaugh: I guess these would really be floats not strings, e.g. 1.05.

tests/plugins/test_store_metadata.py
```diff
@@ -153,6 +179,11 @@ def test_metadata_plugin_rpmqa_failure(tmpdir):
     workflow.postbuild_results = {
         PostBuildRPMqaPlugin.key: RuntimeError(),
     }
+    workflow.plugins_durations = {
+        CpDockerfilePlugin.key: "1",
```
twaugh: Same here.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
twaugh: I was thinking we would *only* store the exceptions, e.g. (plugin, exception_name, exception_repr).

twaugh:
Looks good! I added some notes while looking in the 'changes since' view, but it seems if you do that GitHub annotates the wrong lines so I re-did them in the 'full changes' view.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -109,6 +110,33 @@ def get_pullspecs(self, digests):
...
+        else:
+            build_failed = self._parse_logs(r'INFO - build was unsuccess?ful')
+            if build_failed != ['']:
+                return {'failed_plugin': 'build', 'exception': ''}
+            else:
+                return build_failed
+
+    def get_plugin_metadata(self):
+        return {
+            "image": self._parse_logs(r'pulp_push - INFO - image names: \[.*\'([^\']*):latest')[0],
```
vrutkovs: Log parsing should go, we can store all this info in metadata

vrutkovs:
@twaugh all comments addressed, log is not parsed anymore. 

Anything else we want to store there before I start squashing commits?

twaugh:
Something else: we probably also want the start time for each plugin. That allows us to see afterwards which plugins were running at the same time -- e.g. work out how many concurrent 'squash' plugins there were and so on.

vrutkovs:
Yeah, makes sense

atomic_reactor/plugin.py
```diff
@@ -211,6 +225,7 @@ def run(self, keep_going=False):
             except Exception as ex:
                 msg = "plugin '%s' raised an exception: %r" % (plugin_instance.key, ex)
                 logger.debug(traceback.format_exc())
+                self.save_plugin_error(plugin_instance.key, repr(ex))
```
twaugh: The thing I was hoping we could track as far as exceptions go is: "which exception caused the build to fail, and which plugin did it come from?"

What's being stored here is a slightly different thing: "which plugins raised which exceptions?" It doesn't tell us which one actually caused the build to fail -- not all exceptions cause failures.

So I think we should be saving this error in the `on_plugin_failed()` section further down instead.

twaugh:
We don't need to store the pulp image name, as that's stored elsewhere in the annotations ("repositories").

atomic_reactor/plugin.py
vrutkovs: Right, only the critical errors should be recorded

twaugh:
LGTM

atomic_reactor/inner.py
```diff
@@ -325,14 +330,28 @@ def build_docker_image(self):
...
+                logger.debug(traceback.format_exc())
+
             self.build_logs = build_result.logs
 
             self.build_failed = build_result.is_failed()
 
             if build_result.is_failed():
                 # The docker build failed. Finish here, just run the
                 # exit plugins (from the 'finally:' block below).
+                self.plugins_errors['build'] = ''
```
mmilata: Should this be saved under `dockerbuild` as above? Also why empty string when there was an error?

atomic_reactor/plugin.py
```diff
@@ -46,6 +47,8 @@ class Plugin(object):
     # by default, if plugin fails (raises exc), execution continues
     is_allowed_to_fail = True
 
+    duration = datetime.timedelta()
```
mmilata: Is this used anywhere?

atomic_reactor/inner.py
vrutkovs: Agreed, lets make it consistent to ```dockerbuild```

We set it to an empty string as its complicated to get an actual error message without pasting the whole docker log output

atomic_reactor/plugin.py
vrutkovs: This might be important for an empty plugin - it may not have its duration set and we'll try to work with ```None```.

Not sure if its really needed though, its mostly a safety measure

atomic_reactor/inner.py
```diff
@@ -325,14 +330,28 @@ def build_docker_image(self):
                 self.autorebuild_canceled = True
                 raise
 
+            start_time = datetime.datetime.now()
+            self.plugins_timestamps['dockerbuild'] = start_time.isoformat()
```
twaugh: Should this have 'Z' added onto the end?

vrutkovs: That's not really correct - there's no way to know server's tz and properly convert it to UTC and indicate it with 'Z'

twaugh: OK, filed #454 for an instance of that I saw elsewhere in atomic-reactor.

atomic_reactor/plugin.py
twaugh: But where is it referenced? Don't durations get stored directly into workflow.plugins_durations? The `Plugin` base class doesn't use `self.duration` anywhere.

twaugh:
Want to squash any commits together before merge?

vrutkovs:
Done

twaugh:
Could include "(fixes #424, fixes #425)" in the commit message.

atomic_reactor/inner.py
mmilata: Why is there no way to know the TZ - is it some docker-specific issue? The other instance is using `.utcnow()` and I was assuming the result is correct.

vrutkovs:
I usually set 'fixes' in PR's description - this way we'll get less noise when commits are amended and force-pushed. The commit will still have the link to the PR

mmilata:
I would like to continue work on this feature - please review #462. The `duration` member was removed and I also applied a couple of mostly cosmetic fixes.

PR #445, opened 2016-04-01
run koji_promote all the time for CG imports, not just rebuilds

maxamillion:


twaugh:
LGTM

PR #448, opened 2016-04-08
prevent store_metadata_in_osv3 failure when pull_base_image fails

twaugh:
Getting the base_image_inspect property invokes docker.Client().inspect_image(), which can fail.


mmilata:
LGTM, merged.

PR #452, opened 2016-04-11
koji_promote: output file paths and sha256 sum to debug ssl connection

vrutkovs:
This would help us ensure that our instance has correct certs set up


vrutkovs:
Its useless - for some reason sha sums don't match but the certificate works

PR #456, opened 2016-04-11
[WIP] Use koji's cert from secrets or don't check ssl if baseurl has ssl

### LINKED ISSUE #455: koji plugin should also download and install repo's certificate
This would resolve yum's ```Peer's Certificate issuer is not recognized.``` during install

vrutkovs:
Fixes #455

### DIFF
--- a/atomic_reactor/plugins/pre_koji.py
+++ b/atomic_reactor/plugins/pre_koji.py
@@ -11,7 +11,6 @@ Pre build plugin for koji build system
 import os
 import koji
 from atomic_reactor.constants import YUM_REPOS_DIR
-
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.util import render_yum_repo
 
@@ -57,6 +56,16 @@ class KojiPlugin(PreBuildPlugin):
             'enabled': 1,
             'gpgcheck': 0,
         }
+
+        # yum doesn't accept a certificate path in sslcacert - it requires a db with added cert
+        # dnf ignores that option completely
+        # we have to fall back to sslverify=0 everytime we get https repo from brew so we'll surely
+        # be able to pull from it
+
+        if baseurl.startswith("https://"):
+            self.log.info("Ignoring certificates in the repo")
+            repo['sslverify'] = 0
+
         path = os.path.join(YUM_REPOS_DIR, self.target + ".repo")
         self.log.info("yum repo of koji target: '%s'", path)
         self.workflow.files[path] = render_yum_repo(repo, escape_dollars=False)
--- a/tests/plugins/test_koji.py
+++ b/tests/plugins/test_koji.py
@@ -93,20 +93,67 @@ def prepare():
     return tasker, workflow
 
 
-def test_koji_plugin():
-    tasker, workflow = prepare()
-    runner = PreBuildPluginsRunner(tasker, workflow, [{
-        'name': KojiPlugin.key,
-        'args': {
-            "target": KOJI_TARGET,
-            "hub": "",
-            "root": ROOT,
+class TestKoji(object):
+    @pytest.mark.parametrize(('root',
+                              'koji_ssl_certs',
+                              'expected_string',
+                              'expected_file'), [
+        # Plain http repo
+        ('http://example.com',
+         False,
+         None,
+         None),
+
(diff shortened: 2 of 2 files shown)


atomic_reactor/plugins/pre_koji.py
```diff
@@ -57,6 +60,19 @@ def run(self):
...
         }
+        if baseurl.startswith("https://"):
+            # If the repo starts with https we should fetch the certificate and update trusts
+            try:
+                if self.koji_ssl_certs:
+                    self.log.info("Using koji certificates from %s" % self.koji_ssl_certs)
+                    capath = os.path.join(self.koji_ssl_certs, 'ca')
+                    with open(capath, 'rb') as cacert_fd:
+                        repo['sslcacert'] = cacert_fd.read()
+            except Exception:
```
twaugh: We'll never get here if self.koji_ssl_certs is False.

atomic_reactor/plugins/pre_koji.py
```diff
@@ -57,6 +62,27 @@ def run(self):
...
+                        repo['sslcacert'] = cacert_fd.read()
+                else:
+                    self.log.info("Fetching certificate from %s" % self.pathinfo.topdir)
+                    o = urlparse(self.pathinfo.topdir)
+                    cert = ssl.get_server_certificate((o.hostname, 443))
+                    certpath = '/etc/pki/ca-trust/source/anchors/{}.cert'.format(o.hostname)
+                    self.workflow.files[certpath] = cert
+                    repo['sslcacert'] = certpath
+        except Exception:
+            self.log.debug(traceback.format_exc())
```
twaugh: We won't get here if baseurl.startswith("https://") is False.

vrutkovs: That's expected - we don't care about ```sslcerify``` if koji gives a plan HTTP link. Although for testing purpose I'll play with it in the next commit

atomic_reactor/plugins/pre_koji.py
```diff
@@ -57,6 +64,16 @@ def run(self):
             'enabled': 1,
             'gpgcheck': 0,
         }
+
+        # yum doesn't accept a certificate path in sslcacert - it requries a db with added cert
```
twaugh: typo, should be 'requires'

tests/plugins/test_koji.py
```diff
@@ -93,20 +93,67 @@ def prepare():
...
+         None),
+
+        # https with cert available
+        # ('https://example.com',
+        #  False,
+        #  'sslcacert=/etc/yum.repos.d/example.com.cert',
+        #  '/etc/yum.repos.d/example.com.cert'),
+
+    ])
+    def test_koji_plugin_http(self, tmpdir, root, koji_ssl_certs,
```
twaugh: my fault, but this should probably be renamed back to 'test_koji_plugin' since there's nothing especially 'http' about it.

twaugh:
Looks good to me. Want to squash any commits?

atomic_reactor/plugins/pre_koji.py
```diff
@@ -20,7 +25,7 @@ class KojiPlugin(PreBuildPlugin):
     key = "koji"
     is_allowed_to_fail = False
 
-    def __init__(self, tasker, workflow, target, hub, root):
+    def __init__(self, tasker, workflow, target, hub, root, koji_ssl_certs=None):
```
mmilata: Why add the parameter when it's not used?

vrutkovs: The original plan was 'either reuse kojisecret for cert or download the cert from server - and then set ```sslcacert``` in yum repos). It seems that yum doesn't work this way - but we might revisit this.

I'm gonna clean this up and repush

atomic_reactor/plugins/pre_koji.py
```diff
@@ -10,8 +10,13 @@
 """
 import os
 import koji
-from atomic_reactor.constants import YUM_REPOS_DIR
+try:
+    from urlparse import urlparse
+except ImportError:
+    # py3
+    from urllib import parse as urlparse
```
twaugh: We aren't using urlparse now, are we?

mmilata:
LGTM now

PR #458, opened 2016-04-13
push_to_pulp: filter out uploaded layers if possible

vrutkovs:
Needs https://github.com/release-engineering/dockpulp/pull/49

### DIFF
--- a/atomic_reactor/plugins/post_push_to_pulp.py
+++ b/atomic_reactor/plugins/post_push_to_pulp.py
@@ -51,7 +51,9 @@ import dockpulp.imgutils
 import os
 import re
 import tempfile
+import subprocess
 from collections import namedtuple
+from tempfile import NamedTemporaryFile
 
 
 # let's silence warnings from dockpulp: there is one warning for every request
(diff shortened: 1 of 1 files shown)


twaugh:
The way we'd need to implement this properly is to:
1. Run the compress plugin *after* the pulp_push plugin
2. In pulp_push, create a new tarball that omits the unnecessary layers, compress it (gzip), and upload that instead
3. Then copy *all* the layers into the repo and apply tags

twaugh:
Since the running order for plugins is set by osbs-client, we should handle compress being run before pulp_push as well as after. In that case, we'll need to uncompress the tarball in order to create a new smaller one.

atomic_reactor/plugins/post_push_to_pulp.py
```diff
@@ -176,11 +179,39 @@ def push_tarball_to_pulp(self, image_names, repo_prefix="redhat-"):
...
+            elif file_path.endswith('.tar'):
+                p_t.prepend("cat %s" % file_path, ".-")
+            else:
+                raise Exception("Unknown tarball format: %s" % file_path)
+            p_t.append("tar --delete %s" % ' '.join(remove_layers), "--")
+            p_t.append("gzip -", "--")
+            with p_t.open(NamedTemporaryFile('r', prefix='strip_tar_', delete=True), 'r') as f:
+                f.flush()
+                p.upload(f.name)
+        except Exception:
```
twaugh: Any reason to explicitly use `Exception` here?

vrutkovs: Lots of things may go wrong here:
 * failed to pack/unpack
 * unknown archive type
 * failed to write the temp file

atomic_reactor/plugins/post_push_to_pulp.py
```diff
@@ -176,11 +179,39 @@ def push_tarball_to_pulp(self, image_names, repo_prefix="redhat-"):
...
+            self.log.debug(traceback.format_exc())
```
twaugh: I wonder whether to handle the common case of AttributeError separately to avoid tracebacks in the logs, something like:

```
try:
    get_image_ids = getattr(p, 'getImageIdsExist')
except AttributeError:
    self.log.info("uploading full tar")
    p.upload(self.filename)
else:
    try:
        existing_imageids = get_image_ids(layers)
        ....
    except:
        self.log.debug(traceback.format_exc())
        self.log.info("fallback back to full tar upload")
        p.upload(self.filename)
```

atomic_reactor/plugins/post_push_to_pulp.py
twaugh: Could use `except:`?

atomic_reactor/plugins/post_push_to_pulp.py
```diff
@@ -176,11 +179,36 @@ def push_tarball_to_pulp(self, image_names, repo_prefix="redhat-"):
...
+                raise Exception("Unknown tarball format: %s" % self.filename)
+
+            with NamedTemporaryFile(prefix='strip_tar_', delete=True) as outfile:
+                cmd = "{0} {1} | tar --delete {2} | gzip - > {3}".format(
+                    unpacker, self.filename, ' '.join(remove_layers), outfile.name)
+                self.log.debug("running %s" % cmd)
+                subprocess.check_call(cmd, shell=True)
+                self.log.debug("uploading %s" % outfile.name)
+                p.upload(outfile.name)
+        except:
```
mmilata: I'd suggest using `except Exception` if there's really nothing more specific - so that we don't end up catching things like signals.

vrutkovs: We discussed it [here](https://github.com/projectatomic/atomic-reactor/pull/458#discussion_r59534830) - since we don't know what type of exception may occur (some don't even inherit from ```Exception```) we'd rather have as generic catcher as possible - just to be super-sure that it won't fail

atomic_reactor/plugins/post_push_to_pulp.py
```diff
@@ -176,11 +179,36 @@ def push_tarball_to_pulp(self, image_names, repo_prefix="redhat-"):
...
+            self.log.debug(traceback.format_exc())
```
mmilata: Please use something like `self.log.debug("bad thing happened", exc_info=True)` instead. It's python2.7+ but that's not a problem here. Or maybe delete this and use exc_info=True for the self.log.info below?

https://docs.python.org/2/library/logging.html?highlight=logging#logging.debug

vrutkovs: Agreed

atomic_reactor/plugins/post_push_to_pulp.py
mmilata: Ah, I missed that, sorry.

atomic_reactor/plugins/post_push_to_pulp.py
twaugh: Now I've read the section in PEP 8 about it, I see why you might want `except Exception:`.

mmilata:
LGTM, merged.

PR #461, opened 2016-04-14
Support setting proxy for yum repos

vrutkovs:
* [x] Update pre_koji
* [x] Update add_yum_repo_by_url

We also have an option to verify yum repo contents (using ```configparser```) - do we want this being enabled? Currently the proxy won't be injected if the repo file is invalid - but it still will be passed to the buildroot

### DIFF
--- a/atomic_reactor/plugins/pre_add_yum_repo_by_url.py
+++ b/atomic_reactor/plugins/pre_add_yum_repo_by_url.py
@@ -32,9 +32,14 @@ import requests
 try:
     # py2
     from urlparse import unquote, urlsplit
+    import ConfigParser as configparser
+    # We import BytesIO as StringIO as configparser can't properly write
+    from io import BytesIO, BytesIO as StringIO
 except ImportError:
     # py3
     from urllib.parse import unquote, urlsplit
+    import configparser
+    from io import BytesIO, StringIO
 
 
 class YumRepo(object):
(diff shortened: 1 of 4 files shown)


tests/plugins/test_add_yum_repo_by_url.py
```diff
@@ -49,42 +50,51 @@ def prepare():
...
+        repo_content = '%sproxy = %s\n\n' % (repocontent.decode('utf-8'), inject_proxy)
     # next(iter(...)) is for py 2/3 compatibility
     assert next(iter(workflow.files.keys())) == os.path.join(YUM_REPOS_DIR, filename)
-    assert next(iter(workflow.files.values())) == repocontent
+    assert next(iter(workflow.files.values())) == repo_content
     assert len(workflow.files) == 1
 
 
-def test_multiple_repourls():
+@pytest.mark.parametrize('inject_proxy', [None, 'http://proxy.com'])
```
twaugh: Let's use example.com for dummy URIs, e.g. proxy.example.com

tests/plugins/test_koji.py
```diff
@@ -114,8 +123,16 @@ class TestKoji(object):
         ('https://nosuchwebsiteforsure.com',
          False,
          'sslverify=0',
+         None,
          None),
 
+        # https with no cert available
+        ('https://nosuchwebsiteforsure.com',
```
twaugh: Maybe we could flexmock() out the actual certificate fetching, so we aren't hitting the network for this test?

tests/plugins/test_add_yum_repo_by_url.py
vrutkovs: Right, fixed it

tests/plugins/test_koji.py
```diff
@@ -97,13 +97,22 @@ class TestKoji(object):
...
          None,
+         None,
          None),
 
+        # Plain http repo with proxy
+        ('http://example.com',
+         False,
+         None,
+         None,
+         'http://proxy.com'),
```
twaugh: proxy.example com?

tests/plugins/test_koji.py
```diff
@@ -114,8 +123,16 @@ class TestKoji(object):
...
          'sslverify=0',
+         None,
          None),
 
+        # https with no cert available
+        ('https://nosuchwebsiteforsure.com',
+         False,
+         'sslverify=0',
+         None,
+         'http://proxy.com'),
```
twaugh: proxy.example.com?

atomic_reactor/plugins/pre_add_yum_repo_by_url.py
```diff
@@ -57,22 +61,46 @@ def fetch(self):
...
+            buf.close()
+
+    def set_proxy_for_all_repos(self, proxy_name):
+        for section in self.config.sections():
+            self.config.set(section, 'proxy', proxy_name)
+
+        output = StringIO()
+        self.config.write(output)
+        self.content = output.getvalue()
+        output.close()
```
twaugh: Can we use 'with String() as output:' to avoid needing to close it explicitly?

tests/plugins/test_koji.py
vrutkovs: Yep, for now we don't fetch certificate - we'll have to deal with this later

atomic_reactor/plugins/pre_add_yum_repo_by_url.py
vrutkovs: Right, makes sense

mmilata:
LGTM!

Issue #436, opened 2016-03-18
Exit plugins should be able to fail the build


twaugh:
All the configured exit plugins should run, but they should each be able to mark the build as failed.

Eg if store_metadata or koji_promote fails, the build is unsuccessful. But we should still remove the image afterwards.

twaugh:
Closed via #463.

PR #463, opened 2016-04-15
Allowing exit plugins to fail build #436

lcarva:
Also added NoopMixIn because switching from 'or' to 'and'
caused NotImplementedError to occur on dummy plugins.

### DIFF
--- a/tests/test_inner.py
+++ b/tests/test_inner.py
@@ -189,6 +189,16 @@ class ExitRaises(RaisesMixIn, ExitPlugin):
     key = 'exit_raises'
 
 
+class ExitRaisesAllowed(RaisesMixIn, ExitPlugin):
+    """
+    An Exit plugin that should raise an exception.
+    """
+
+    is_allowed_to_fail = True
+
+    key = 'exit_raises_allowed'
+
+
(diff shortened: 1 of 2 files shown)


tests/test_inner.py
```diff
@@ -588,10 +606,9 @@ def test_workflow_plugin_error(fail_at):
     # earlier plugin failures.
     assert watch_post.was_called() == (fail_at not in ('pre', 'prepub', 'post'))
 
-    # But all exit plugins should run, even if one of them also raises
-    # an exception.
-    assert watch_exit.was_called()
-
+    # Because exit_raises is marked as is_allowed_to_fail=False, it prevents
+    # all other exit plugins to continue.
```
twaugh: I don't think this is quite what we need. We want all the exit plugins to run regardless of failure, but if one fails that is "not allowed to fail", the build as a whole should fail - after all the exit plugins have run.

lcarva: Makes sense. Updated the PR.

twaugh:
LGTM

twaugh:
Actually the only nit-pick is that the commit message should reference #436 not #463. :-)

PR #466, opened 2016-04-20
Pass login/password/email to tag_and_push plugin if docker registry requires auth

### LINKED ISSUE #465: [WIP] tag_and_push should support authentication

vrutkovs:
Fixes #465

Docker-py doesn't allow passing a conf file for push (although it does allow it for pull).

TODO:
 * [x] Make sure we can push with login/email/password
 * [x] Use Openshift secrets to store this in a secret
 * [x] Rewrite the code to use secret instead of plain config vars
 * [x] Add tests


mmilata:
LGTM, merged

PR #467, opened 2016-04-21
Add option to not generate archive in squash plugin

goldmann:
Currently when you want to load the image into the daemon
it'll create the archive too, this in't something I want.
I want the flexibility to choose if I want to load image,
or create the archive or do both.

### DIFF
--- a/atomic_reactor/plugins/prepub_squash.py
+++ b/atomic_reactor/plugins/prepub_squash.py
@@ -56,15 +56,17 @@ class PrePublishSquashPlugin(PrePublishPlugin):
     is_allowed_to_fail = False
 
     def __init__(self, tasker, workflow, tag=None, from_base=True, from_layer=None,
-                 dont_load=False):
+                 dont_load=False, save_archive=True):
         """
         :param tasker: DockerTasker instance
         :param workflow: DockerBuildWorkflow instance
         :param from_base: bool, squash from base-image layer, on by default
         :param from_layer: layer from we will squash - if specified, takes precedence over from_base
         :param tag: str, new name of the image - by default use the former one
-        :param dont_load: if `False` (default), squashed image is loaded into Docker *and* saved
-            to `$tmpdir/image.tar`; if `True`, squashed image is only saved as a file
+        :param dont_load: if `False` (default), squashed image is loaded back into Docker daemon;
+            if `True`, squashed image is not loaded back into Docker
+        :param save_archive: if `True` (default), squashed image is saved in an archive on the
+            disk under the image.tar name; if `False`, archive is not generated
         """
         super(PrePublishSquashPlugin, self).__init__(tasker, workflow)
         self.image = self.workflow.builder.image_id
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/prepub_squash.py
```diff
@@ -56,16 +56,18 @@ class PrePublishSquashPlugin(PrePublishPlugin):
...
         :param tasker: DockerTasker instance
         :param workflow: DockerBuildWorkflow instance
         :param from_base: bool, squash from base-image layer, on by default
         :param from_layer: layer from we will squash - if specified, takes precedence over from_base
         :param tag: str, new name of the image - by default use the former one
         :param remove_former_image: bool, remove unsquashed image?
-        :param dont_load: if `False` (default), squashed image is loaded into Docker *and* saved
-            to `$tmpdir/image.tar`; if `True`, squashed image is only saved as a file
+        :param dont_load: if `False` (default), squashed image is not loaded back into Docker; if `True`,
+            squashed image is loaded back into Docker daemon
```
mmilata: If I'm parsing it correctly the comment describes the opposite of the actual behaviour.

mmilata:
LGTM apart from the comment.

atomic_reactor/plugins/prepub_squash.py
goldmann: Good catch, I'll update this tomorrow.

twaugh: This can be merged once the comment is fixed I think.

goldmann: Oh man, I totally forgot about it!

goldmann: Fixed!

twaugh:
We should add some code coverage testing. Seems like there's none at all for this plugin.

goldmann:
Oh, it looks that the PR is not merged. @twaugh - do we want it merged or are you dropping it? I could work on it to make it mergeable again.

vrutkovs:
I think its a good idea to have this

twaugh:
There is no unit test covering this plugin which makes me nervous about changes to it.

goldmann:
Well, my code doesn't have any bugs. Never.

goldmann:
I'll work on it :)

twaugh:
Please rebase... also, there is now a test module for this :-)

goldmann:
@twaugh @vrutkovs Finally, here we go again.

goldmann:
Now looking at the code - we could make the squash plugin even a no-op if `dont_load` is set to `True` and `save_archive` is set to False.

twaugh:
Remind me again what use-case this option is for? ;-)

goldmann:
To burn some CPU cycles without a reason.

lcarva:
@goldmann, any additional changes before this is merged?

goldmann:
No, no, this is good to go :)

lcarva:
Updated release notes draft:
> - `squash` plugin now takes `save_archive` to control creation of squashed image archive.

PR #468, opened 2016-04-21
koji_promote: fix up v1/v2/'docker save'

twaugh:
The `metadata_only` parameter is not really to do with V1 vs V2, but instead relates to whether the output of 'docker save' is uploaded to Koji or not.

CC @maxamillion 
This should probably be adjustable in osbs.conf.


mmilata:
LGTM & merged.

Issue #439, opened 2016-03-18
koji_promote: store CG build ID as label


twaugh:
Users will need to know the ID of the build object created using the Content Generator API.

twaugh:
CC @maxamillion

twaugh:
Seems like the CGImport() call will return a buildinfo map, like get_build() does:
  https://pagure.io/koji/pull-request/71
  https://pagure.io/koji/blob/master/f/hub/kojihub.py#_3278

So we can get the build ID like this:

```
buildinfo = session.CGImport(...)
build_id = buildinfo['id']
```

twaugh:
Closed via #469.

PR #469, opened 2016-04-22
Storing koji build id as label.

lcarva:
[ Issue #439 ]
Modified KojiPromotePlugin to return the ID of the started Koji build.
Added to DockerBuildWorkflow exit_result attribute to store results from exit plugins.
Modified StoreMetadataInOSv3Plugin to set a label on the OS Build with the previously stored Koji build ID.

Also, fixed PrePublishPluginsRunner to use the correct results dict.

### DIFF
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -624,7 +624,14 @@ class KojiPromotePlugin(ExitPlugin):
                 if output.file:
                     output.file.close()
 
-        session.CGImport(koji_metadata, server_dir)
+        build_info = session.CGImport(koji_metadata, server_dir)
+        # Older versions of CGImport do not return a value.
+        build_id = build_info.get("id") if build_info else None
 
         self.log.debug("Submitted with metadata: %s",
                        json.dumps(koji_metadata, sort_keys=True, indent=4))
+        self.log.debug("Build information: %s",
+                       json.dumps(build_info, sort_keys=True, indent=4))
+
+        return build_id
+
--- a/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
+++ b/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
@@ -17,6 +17,7 @@ from osbs.conf import Configuration
 from atomic_reactor.plugin import ExitPlugin
 from atomic_reactor.plugins.pre_return_dockerfile import CpDockerfilePlugin
 from atomic_reactor.plugins.pre_pyrpkg_fetch_artefacts import DistgitFetchArtefactsPlugin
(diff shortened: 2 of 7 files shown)


atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -624,7 +624,13 @@ def run(self):
                 if output.file:
                     output.file.close()
 
-        session.CGImport(koji_metadata, server_dir)
+        build_info = session.CGImport(koji_metadata, server_dir)
+        build_id = build_info["id"]
```
twaugh: Wonder if we should deal with build_info being None here? That will happen when running against a Koji that doesn't have the CGImport() fix. In that case, we should return None as the build_id, but we'll have to make sure we don't try storing a nil value in the annotation because OpenShift will complain.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -113,6 +117,15 @@ def get_plugin_metadata(self):
             "durations": self.workflow.plugins_durations,
         }
 
+    def make_labels(self):
+        labels = {}
+
+        koji_build_id = self.get_exit_result(KojiPromotePlugin.key)
```
twaugh: OK, great, so a None value will be converted to '' to make OpenShift happy.

tests/plugins/test_store_metadata.py
```diff
@@ -239,3 +243,47 @@ def test_metadata_plugin_rpmqa_failure(tmpdir):
     assert "cp_dockerfile" in plugins_metadata["durations"]
     assert "distgit_fetch_artefacts" in plugins_metadata["durations"]
     assert "all_rpm_packages" in plugins_metadata["durations"]
+
+def test_labels_metadata_plugin(tmpdir):
+    workflow = prepare()
+
+    workflow.exit_results = {
+        KojiPromotePlugin.key: "1234",
```
twaugh: Could use a variable to store this string so we're not repeating it.

atomic_reactor/plugins/exit_koji_promote.py
lcarva: Good point. Added check and corresponding unit test.

tests/plugins/test_store_metadata.py
lcarva: Done.

twaugh:
LGTM

twaugh:
Having tested this in an environment where CGImport() returns the build information, I now see that we'll have to change to order that the plugins run. Currently the store_metadata plugin runs before koji_promote, so it doesn't see the build ID result.

I've filed a pull request to fix this: https://github.com/projectatomic/osbs-client/pull/406

PR #471, opened 2016-04-29
koji_promote: tag build when target provided

twaugh:
Depends on https://github.com/projectatomic/atomic-reactor/pull/479.

### DIFF
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -24,7 +24,7 @@ from atomic_reactor.source import GitSource
 from atomic_reactor.plugins.post_rpmqa import PostBuildRPMqaPlugin
 from atomic_reactor.constants import PROG
 from atomic_reactor.util import get_version_of_tools, get_checksums, get_build_json
-from atomic_reactor.koji_util import create_koji_session
+from atomic_reactor.koji_util import create_koji_session, TaskWatcher
 from osbs.conf import Configuration
 from osbs.api import OSBS
 from osbs.exceptions import OsbsException
@@ -66,7 +66,8 @@ class KojiPromotePlugin(ExitPlugin):
                  verify_ssl=True, use_auth=True,
                  koji_ssl_certs=None, koji_proxy_user=None,
                  koji_principal=None, koji_keytab=None,
-                 metadata_only=False, blocksize=None):
+                 metadata_only=False, blocksize=None,
+                 target=None, poll_interval=5):
         """
         constructor
 
@@ -82,6 +83,8 @@ class KojiPromotePlugin(ExitPlugin):
         :param koji_keytab: str, keytab name (must specify principal)
         :param metadata_only: bool, whether to omit the 'docker save' image
         :param blocksize: int, blocksize to use for uploading files
+        :param target: str, koji target
+        :param poll_interval: int, seconds between Koji task status requests
         """
         super(KojiPromotePlugin, self).__init__(tasker, workflow)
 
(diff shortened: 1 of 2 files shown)


twaugh:
Note: not tested against a real Koji hub yet; requires CGImport API fix so we know which build ID to tag.

twaugh:
I should be able to test this in the next few days.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -633,5 +638,30 @@ def run(self):
...
+        # Tag the build
+        if build_id is not None and self.target is not None:
+            self.log.debug("Finding build tag for target %s", self.target)
+            target_info = session.getBuildTarget(self.target)
+            build_tag = target_info['build_tag']
+            self.log.info("Tagging build with %s", build_tag)
+            task_id = session.tagBuild(build_tag, build_id)
+            self.log.debug("Watching task %r", task_id)
+            last = None
+            while last not in ['CLOSED', 'CANCELED', 'FAILED']:
```
lcarva: Might be a good idea to have a `waitForKojiTask` method in utils.py. Something similar to this will be needed for pre_add_filesystem plugin.

twaugh: Good idea. I'll add it in this PR.

lcarva: The koji authorization piece will also be duplicated. Maybe we need a koji_utils.py kind of thing?

twaugh: Yes, I was thinking of adding exactly that, and making sure to package koji_utils.py into the 'koji' sub-package in the spec file.

OK, authorization as well, noted.

twaugh:
New koji_util module added, see what you think.

I still haven't actually tested this against a real koji server.

atomic_reactor/koji_util.py
```diff
@@ -0,0 +1,92 @@
...
+        self.poll_interval = poll_interval
+        self.last_state = False
+
+    def state_changes(self):
+        logger.debug("watching koji task %r", self.task_id)
+        while True:
+            time.sleep(self.poll_interval)
+            task_info = self.session.getTaskInfo(self.task_id, request=True)
+            if task_info:
+                state = task_info.get('state')
```
lcarva: State is an int, you have to convert it using the `TASKS_STATES` enum: https://git.fedorahosted.org/cgit/koji/tree/koji/tasks.py#n526

atomic_reactor/koji_util.py
```diff
@@ -0,0 +1,92 @@
...
+class TaskWatcher(object):
+    def __init__(self, session, task_id, poll_interval=5):
+        self.session = session
+        self.task_id = task_id
+        self.poll_interval = poll_interval
+        self.last_state = False
+
+    def state_changes(self):
+        logger.debug("watching koji task %r", self.task_id)
+        while True:
```
lcarva: Could this be simplified by using `self.session.taskFinished(self.task_id)` ?
The downside is that if the task does not exist the following error occurs:
`GenericError('query returned no rows',)`
Looking at logs, it may not be very clear what this error means.

twaugh:
Re-worked koji_util.

tests/koji/__init__.py
```diff
@@ -6,11 +6,22 @@
 of the BSD license. See the LICENSE file for details.
 """
 
+TASK_STATES = {
```
lcarva: This doesn't seem to be used.

twaugh:
This has now been successfully tested against a real koji hub.

twaugh:
Ready for review/merge.

vrutkovs:
Looks fine to me

mmilata:
LGTM!

PR #476, opened 2016-05-09
Fix logging of 'docker build' output

twaugh:
Previously the JSON representation was being logged instead of the actual log text.

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -169,7 +169,7 @@ class CommandResult(object):
             if l:
                 logger.debug(l)
 
-        self._logs.append(item)
+        self._logs.append(line)
         if parsed_item is not None:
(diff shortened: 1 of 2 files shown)


atomic_reactor/util.py
```diff
@@ -169,7 +169,7 @@ def parse_item(self, item):
             if l:
                 logger.debug(l)
 
-        self._logs.append(item)
+        self._logs.append(line)
```
vrutkovs: Shouldn't we be logging ```l``` here - a stripped line after ```splitlines``` call?

twaugh:
I've added a commit that uses the split stripped lines, instead of the 'stream' value directly.

vrutkovs:
LGTM

mmilata:
LGTM!

PR #481, opened 2016-05-11
Add plugin for creating base images

lcarva:


vrutkovs:
This probably needs a simple test (to make Coveralls happy). Otherwise looks good

lcarva:
@vrutkovs unit tests are on their way. just wanted some feedback as early as possible.

mmilata:
LGTM too

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -0,0 +1,209 @@
...
+        dockerfile = DockerfileParser(self.workflow.builder.df_path)
+
+        if not self.is_kickstart_type(dockerfile.baseimage):
+            self.log.info('Base image not supported: %s', dockerfile.baseimage)
+            return
+
+        self.session = koji.ClientSession(self.koji_hub)
+        # TODO: Remove opts setting workaround
+        self.session.opts = {'krbservice': 'brewhub'}
+        koji_login(self.session, **self.koji_auth_info)
```
twaugh: Should use create_koji_session() now it's merged.

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -0,0 +1,209 @@
...
+        for f in self.session.listTaskOutput(task_id):
+            f = f.strip()
+            match = filesystem_regex.match(f)
+            if match:
+                return task_id, match.group(0)
+
+        # Not found in this task, search sub tasks
+        for sub_task in self.session.getTaskChildren(task_id):
+            found = self.find_filesystem(sub_task['id'], filesystem_regex)
+            if found:
```
twaugh: This will always be evaluated to True because the returned tuple is not empty.

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -0,0 +1,209 @@
...
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+from ConfigParser import ConfigParser
+
+import json
+import re
+import os
+import time
```
twaugh: `time` doesn't seem to be used.

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -0,0 +1,209 @@
...
+    def build_filesystem(self):
+        # Kickstart file should be in the same folder as Dockerfile
+        df_path, df_dir = self.workflow.source.get_dockerfile_path()
+        kickstart = os.path.join(df_dir, 'kickstart.conf')
+        if not os.path.exists(kickstart):
+            raise RuntimeError('Kickstart file not found: {}'
+                               .format(kickstart))
+
+        image_name, args, kwargs = self.parse_image_build_config(kickstart)
+        pattern = ('{}.*(\.tar|\.tar\.gz|\.tgz|\.bzip|\.tar\.xz|\.txz)$'
```
twaugh: `.bzip` seems like an odd-one-out. I would expect this list to be something along the lines of:
* `.tar`
* `.tar.gz`
* `.tar.bz2`
* `.tar.xz`

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -0,0 +1,209 @@
...
+
+        self.log.info('Filesystem downloaded.')
+        return contents
+
+    def import_base_image(self, filesystem):
+        result = self.tasker.d.import_image_from_data(filesystem)
+        # Response not deserialized:
+        #   https://github.com/docker/docker-py/issues/1060
+        result = json.loads(result)
+        return result.get('status')
```
twaugh: If there is no `status` member we'll end up putting the string "None" in the FROM line. Maybe this should just be

```
return result['status']
```

so that we error out immediately?

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -0,0 +1,209 @@
...
+    def download_filesystem(self, task_id, filesystem_regex):
+        task_id, file_name = self.find_filesystem(task_id, filesystem_regex)
+        if not task_id:
+            raise RuntimeError('Filesystem not found as task output: {}'
+                               .format(filesystem_regex.pattern))
+
+        self.log.info('Downloading filesystem: %s from task ID: %s',
+                      file_name, task_id)
+
+        contents = self.session.downloadTaskOutput(task_id, file_name)
```
twaugh: Instead of downloading it, how about using this to work to the URL to it?:

```
pathinfo = koji.PathInfo(topdir=...)
url = os.path.join(pathinfo.imagebuild(buildinfo), file_name)
```

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: Formats taken from docker import doc: https://docs.docker.com/engine/reference/commandline/import/

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: Good point.

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: This requires knowing the base URL for the file storage. We could make this an additional parameter in the config, but IMO that makes usage slightly worse.

twaugh: It's already in osbs.conf as `koji_root`.

atomic_reactor/plugins/pre_add_filesystem.py
twaugh: OK. I don't think I have ever seen a filename with a `.bzip` (rather than eg `.bz2`) extension, I don't think that's a real thing :-)

I can't imagine a scenario in which we will have asked Koji to create a filesystem image, and it wouldn't use the traditional UNIX ".tar.{compression}" scheme for naming the file.

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: Upon further investigation the real problem with this approach is that `import_image_from_url` does not return the image ID in a consumable manner [1], and the ability to pass a flag to ignore identity cert verification errors does not seem to be present.

[1] https://github.com/docker/docker-py/issues/1060

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: Done.

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: Done.

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: Removed.

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: Fixed.

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: Done.

tests/plugins/test_add_filesystem.py
```diff
@@ -0,0 +1,237 @@
...
+from __future__ import print_function, unicode_literals
+from textwrap import dedent
+from flexmock import flexmock
+
+import re
+import json
+import pytest
+import os.path
+
+import koji
```
twaugh: Think this line needs to be removed. We have to catch import error when importing koji in order to know when we need to use the stub.

atomic_reactor/plugins/pre_add_filesystem.py
twaugh: Since we're using `None` as the error case, I think I'd rather change the test to `if found is not None:` but it's not a blocker for merging this.

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -0,0 +1,212 @@
...
+    https://fedoraproject.org/wiki/Koji/BuildingImages#Building_Disk_Images
+
+    Once image build task is complete the tarball is downloaded and
+    it's imported into docker. This creates a new image. The existing
+    FROM instruction value is replaced with the ID of this new image.
+
+    The "FROM" instruction should be in the following format:
+        FROM scratch:<type>
+    Where type is one of the build types supported by this plugin,
+    currently only "kickstart" is supported.
```
twaugh: I don't remember which of us suggested it, it might have been me, but I wonder whether 'kickstart' is descriptive enough. How about this?:
```
FROM scratch:koji-image-build
```

Reasoning: the configuration file is not actually for kickstart directly (that's the .ks file it references), but for Koji's image-build task. Since that's the specific thing we're interacting with to get a base layer made I think we ought to name that rather than kickstart, which would most likely be common to any method of doing this.

What do you think?

atomic_reactor/plugins/pre_add_filesystem.py
twaugh: Agreed. In that case, a note about `downloadTaskOutput`: without additional parameters it downloads the entire file and stores it in memory.

Since this file is likely to be very large, might it be a better approach to call this method in a loop specifying the offset and size for each part? We could either spool it into a file and import that file, or else could we use `import_image_from_stream`?

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: Agreed and done.

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: Modified to use `import_image_from_stream`, works like a charm.

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: Since I had to make some modifications anyways, I took care of this.

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -0,0 +1,227 @@
...
+    Runs as a pre build plugin in order to properly adjust base image.
+    """
+
+    key = 'add_filesystem'
+    is_allowed_to_fail = False
+
+    def __init__(self, tasker, workflow, koji_hub,
+                 koji_proxyuser=None, koji_ssl_certs_dir=None,
+                 koji_krb_principal=None, koji_krb_keytab=None,
+                 from_task_id=None, poll_interval=5, blocksize=102400):
```
twaugh: Let's set the blocksize higher, e.g. 10 * 1024 * 1024 (10Mb).

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def parse_base_image(self, base_image):
+        image_build_conf = ''
+        base_image = base_image.split(':')
+        if len(base_image) > 1:
+            image_build_conf = base_image[1]
+
+        if not image_build_conf:
+            image_build_conf = 'image-build.conf'
+        base_image = base_image[0]
```
twaugh: Can use atomic_reactor.util.ImageName.parse for this, eg:

```
image_name = ImageName.parse(base_image)
if image_name.namespace != 'koji' or image_name.repo != 'image-build': ...
image_build_conf = image_name.tag or 'image-build.conf'

File: atomic_reactor/koji_util.py [hunk 11]
```diff
@@ -96,3 +98,22 @@ def wait(self):
...
         return self.state in ['CANCELED', 'FAILED']
+
+
+def stream_task_output(session, task_id, file_name,
+                       blocksize=DEFAULT_DOWNLOAD_BLOCK_SIZE):
+    """
+    Generator to download file from task without loading the whole
+    file into memory.
+    """
+    logger.debug('Streaming {} from task {}'.format(file_name, task_id))
```
twaugh: Not worth fixing unless something else comes up, but landscape.io suggests using '%s' here and passing file_name and task_id as parameters, e.g. logger.debug("Streaming %s from task %s", file_name, task_id) -- just means it doesn't have to format the string if debugging isn't turned on.

Same a few lines down too.

twaugh:
LGTM!

PR #483, opened 2016-05-12
koji_util: don't import unicode_literals from __future__

twaugh:
Doing so causes koji to be unhappy since it requires strings for parameter values.

Couldn't figure out a way of adding a test case for this.

CC @lcarva


lcarva:
LGTM

PR #485, opened 2016-05-13
Construct NVR string early and fail in pre_koji if build was already submitted

vrutkovs:
TODO:
 * [x] Add tests


twaugh:
This plugin doesn't get run at all if a yum repourl is specified for the build, so the check won't be performed when building with signed packages.

vrutkovs:
Right, we probably don't need it if https://github.com/release-engineering/koji-containerbuild/pull/38 will be accepted, although it might help if osbs cli is used.

vrutkovs:
We probably don't need this one after koji-containerbuild change has landed?

twaugh:
Probably not.

vrutkovs:
Closing this one for now - this is implemented in k-cb now

PR #487, opened 2016-05-15
koji_promote: use component in nvr, not Name label

twaugh:
We should be using component-version-release for the Koji Build, i.e.:
 * 'name' = com.redhat.component / BZComponent
 * 'version' = version / Version
 * 'release' = release / Release


vrutkovs:
LGTM

PR #488, opened 2016-05-16
koji_promote: store originating koji task ID in build metadata

twaugh:
This is the ID of the Koji Task responsible for creating the build configuration.

Note: the 'extra' field in the 'build' map is optional, and is no longer required to have an 'image' member.


lcarva:
LGTM

twaugh:
See also https://github.com/projectatomic/osbs-client/pull/418.

vrutkovs:
Looks good, merging

PR #490, opened 2016-05-16
store_metadata_in_osv3: ensure we don't store null as value for image-id

twaugh:


vrutkovs:
LGTM. Just to be sure we do the same trick for other objects

twaugh:
It looks to me like all the other keys are protected.

lcarva:
LGTM. We should consider creating a generic utility for this.

PR #491, opened 2016-05-16
Updates base image in builder

lcarva:


twaugh:
LGTM. We can also drop the ImageName import now.

PR #492, opened 2016-05-17
koji_promote: rename koji_task_id to container_koji_task_id

twaugh:
This avoids ambiguity when we introduce filesystem_koji_task_id.


lcarva:
LGTM

twaugh:
Not sure what's up with landscape, it doesn't know which base to compare with.

PR #499, opened 2016-05-20
koji_promote: seems like build.extra.image is required

twaugh:
It is not listed in the specification any more, but builds do not display correctly without this.


lcarva:
LGTM

mmilata:
LGTM

PR #503, opened 2016-05-24
squash: removal of old image should not be fatal

twaugh:
This pull request adds a new helper function to the remove_built_image plugin, defer_removal. This helper function can be called from other plugins in order to ask that an image be removed at the end of the build. This removal should not be fatal.

As part of this I added a plugin workspace to the workflow object, which is storage that can be used by helper functions. It is persistent for the duration of the build workflow.

There are probably other member variables in the workflow that were being used for this purpose (eg pull_base_images), and they could probably be moved into the workspace.

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -258,6 +258,7 @@ class DockerBuildWorkflow(object):
         self.postbuild_results = {}
         self.prepub_results = {}
         self.exit_results = {}
+        self.plugin_workspace = {}
         self.plugins_timestamps = {}
         self.plugins_durations = {}
         self.plugins_errors = {}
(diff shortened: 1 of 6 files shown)


vrutkovs:
LGTM. Should we move other plugin data (e.g. plugin durations) in ```plugin_workspace``` or its not worth it?

twaugh:
My idea was that the workspace would be for things that only the plugins know about. It's opaque to the core code.

The durations and timings are used by the core code so they should not be put into the workspace. I fully expect there are other things that only the plugins should know about -- especially once the build phase is moved into a plugin.

twaugh:
Tested.

PR #504, opened 2016-05-24
Use labels from Config section of docker inspect

### LINKED ISSUE #434: plugins: use ContainerConfig to get Labels from docker inspect
New base images are going to store labels in `inspect_result['ContainerConfig']['Labels']`. We need to look there instead of `inspect_result['Config']['Labels']`. The `pre_add_labels_in_df` plugin is mainly affected.

Notes:
* Until now, base images had the labels in `inspect_result['Config']['Labels']` and have `inspect_result['ContainerConfig']['Labels']` empty.
* The result of `docker build` seems to always have both `Config` and `ContainerConfig` populated identically (this means post_ plugins needn't care what they use).
* https://github.com/docker/docker/blob/9842c7a741ee480c71bfd8e71e1f22893d010411/image/image.go#L33
* `docker images --filter "label=KEY=VALUE"` seems to look at `ContainerConfig` and ignore `Config` (docker-1.8.2).

CC @eliskasl

mmilata:
I'm unable to find a definition of the docker inspect output, however I think that Labels should be read from the Config dictionary and not ContainerConfig.

Functionally this commit doesn't change anything because pre_ plugins have been using Config already and the only use of ContainerConfig was in post_ plugins where it doesn't really matter because after "docker build" the two dictionaries are identical. The intent is to make code less ambiguous.

Fixes #434.

### DIFF
--- a/atomic_reactor/constants.py
+++ b/atomic_reactor/constants.py
@@ -32,6 +32,10 @@ DEFAULT_YUM_REPOFILE_NAME = 'atomic-reactor-injected.repo'
 
 SOURCE_DIRECTORY_NAME = "source"
 
+# key in dictionary returned by "docker inspect" that holds the image
+# configuration (such as labels)
+INSPECT_CONFIG = "Config"
+
 # docs constants
 
 DESCRIPTION = "Python library with command line interface for building docker images."
--- a/atomic_reactor/plugins/post_tag_by_labels.py
+++ b/atomic_reactor/plugins/post_tag_by_labels.py
@@ -7,6 +7,7 @@ of the BSD license. See the LICENSE file for details.
 """
 
 from atomic_reactor.plugin import PostBuildPlugin
+from atomic_reactor.constants import INSPECT_CONFIG
 
 
 __all__ = ('TagByLabelsPlugin', )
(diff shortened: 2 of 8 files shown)


vrutkovs:
IIUC ```ContainerConfig``` is the config of the base image (see https://github.com/docker/docker/issues/17780#issuecomment-154614319), so +1 from me

mmilata:
@vrutkovs, good find. And according to the templates in https://github.com/redhat-imaging/imagefactory/blob/master/imagefactory_plugins/Docker/Docker.py#L49 we're probably not going to produce images with ImageConfig.Labels populated (IIRC we're using `docker_json_template_1_7_0` now and are going to switch to `docker_json_template_1_10_1` eventually).

twaugh:
Reading https://github.com/docker/docker/issues/17780#issuecomment-154614319 I think `ContainerConfig` isn't the config of the base image, but something more subtle.

When you run 'docker build', each instruction is run in a container. So the last instruction, let's say it's a RUN instruction, is run in a container and the result committed into the image.

```
Step 7 : RUN make install && make distclean
---> Running in d339f36961d1
```

The `ContainerConfig`, then, is the config for the container that was used for committing to the image, in this case d339f36961d1.

Images are not always created using commit, and that's why `ContainerConfig` can be empty.

twaugh:
LGTM

twaugh:
Tested in OSD.

PR #508, opened 2016-06-02
Add defaults for image-build.conf

lcarva:


atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -10,9 +10,17 @@
 try:
     # py2
     from ConfigParser import ConfigParser
+    # Although io.StringIO is available in python2, it
+    # always expects unicode during write. Since
+    # ConfigParser.ConfigParser does not support
+    # writing unicode, io cannot be used.
+    from StringIO import StringIO
```
vrutkovs: Importing ```BytesIO``` as ```StringIO``` might also work (I did that hack at https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/pre_add_yum_repo_by_url.py#L37)

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -94,8 +124,13 @@ def parse_image_build_config(self, config_file_name):
         opts = {}
 
         config = ConfigParser()
+        config.readfp(self.get_default_image_build_conf())
         config.read(config_file_name)
 
+        config_str = StringIO()
+        config.write(config_str)
```
vrutkovs: We should use context manager here

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: I tried that, but it seems to not work if `from __future__ import unicode literals` is used.

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: `StringIO.StringIO` does not support context manager, only `io.StringIO`:

```
In [1]: import io

In [2]: with io.StringIO() as foo:
   ...:     pass
   ...: 

In [3]: import StringIO

In [4]: with StringIO.StringIO() as foo:
   ...:     pass
   ...: 
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-4c8d5c900858> in <module>()
----> 1 with StringIO.StringIO() as foo:
      2     pass
      3 

AttributeError: StringIO instance has no attribute '__exit__'

```

atomic_reactor/plugins/pre_add_filesystem.py
vrutkovs: I think the trick was to read via ```BytesIO``` and write using ```StringIO``` due to a bug in py3, see ```pre_add_yum_repo_by_url.py``` code.

OTOH this is counter-intuitive, so probably we should keep StringIO

atomic_reactor/plugins/pre_add_filesystem.py
vrutkovs: Right, lets leave it as it is then

twaugh:
LGTM

PR #514, opened 2016-06-09
[WIP] Use Koji to set a release label if one is absent

twaugh:
The bump_release plugin is not used in production anywhere right now so it is safe to re-use it for this purpose. The parameters it takes are now different.

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -8,315 +8,64 @@ of the BSD license. See the LICENSE file for details.
 
 from __future__ import unicode_literals
 
-from contextlib import contextmanager
-from copy import deepcopy
-import os
-import re
-import subprocess
-from shutil import rmtree
-from tempfile import mkdtemp
-
 from atomic_reactor.plugin import PreBuildPlugin
-from atomic_reactor.source import GitSource
-from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
-from atomic_reactor.util import get_preferred_label_key
+from atomic_reactor.util import get_all_label_keys, get_preferred_label_key
 from dockerfile_parse import DockerfileParser
-
-
-class GitRepo(object):
-    """
-    In an ideal world, pygit2 would be everywhere we'd need it to be,
-    and we could just use that. Meanwhile, in this world, here's a
-    wrapper around the git binary.
-    """
-
-    def __init__(self, workdir, log, cmd='/usr/bin/git'):
-        self.workdir = workdir
-        self.log = log
-        self.cmd = cmd
-        self.tmpdir = mkdtemp()
-        self.git_wrapper = os.path.join(self.tmpdir, 'git-wrapper.sh')
-        with open(self.git_wrapper, mode='wt') as gitfp:
-            gitfp.write('#!/bin/sh\n'
-                        'exec /usr/bin/ssh -o StrictHostKeyChecking=no "$@"\n')
-
-        os.chmod(self.git_wrapper, 0o755)
-
-    def __enter__(self):
-        return self
-
-    def __exit__(self, exc, value, tb):
-        rmtree(self.tmpdir)
-
-    def git(self, args):
-        """
-        Run git with arguments.
-
-        :param args: list, argument strings
-        :return: str, output (including stderr)
-        """
-
-        argv = [self.cmd] + args
-        env = deepcopy(os.environ)
-        env['GIT_SSH'] = self.git_wrapper
-        self.log.debug("executing command: %r", argv)
-        try:
-            with open('/dev/null', 'r+') as devnull:
-                output = subprocess.check_output(argv,
-                                                 stdin=devnull,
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -8,315 +8,63 @@
...
-
-        if branch_sha != self.git_ref:
-            self.log.error("Branch '%s' is at commit %s (expected %s)",
-                           branch, branch_sha, self.git_ref)
-            raise RuntimeError("Not at expected commit")
-
-        self.log.info("Branch '%s' is at expected commit (%s)",
-                      branch, self.git_ref)
+        self.target = target
+        self.xmlrpc = koji.ClientSession(hub)
```
vrutkovs: We probably should reuse ```create_koji_session``` from ```atomic_reactor.koji_util``` here

twaugh: We don't need to login for this. As written, `create_koji_session()` will try to use Kerberos if no credentials are supplied.

lcarva: It'll only attempt login if you provide a second parameter: https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/koji_util.py#L77

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -8,315 +8,64 @@
...
+        latest = self.xmlrpc.getLatestBuilds(self.target, package=component)
+        try:
+            next_release = self.xmlrpc.getNextRelease(latest[0])
+        except IndexError:
+            next_release = 1
+
+        next_release = str(next_release)
+        for release_label in release_labels:
+            self.log.info("setting %s=%s", release_label, next_release)
+            dockerfile_labels[release_label] = next_release
```
lcarva: Is this meant to actually write back to the Dockerfile? If so, I think you need an additional step: `parser.labels = dockerfile_labels`. https://github.com/DBuildService/dockerfile-parse/blob/master/dockerfile_parse/parser.py#L332

lcarva:
We probably need changes in koji-containerbuild to properly perform n-v-r check. Maybe if release is not provided, bypass it completely?

atomic_reactor/plugins/pre_bump_release.py
twaugh: `dockerfile_labels` is a `Labels` instance at this point. I should make that clearer.

twaugh:
Yes, koji-containerbuild, osbs-client, and the assert_labels plugin -- and possibly other places -- will need to be able to handle the absence of a Release label.

atomic_reactor/plugins/pre_bump_release.py
lcarva: Ah sweet! Didn't realize that `Label` class actually invokes the `parser.labels` property setter. A comment might help the next guy ;)

atomic_reactor/plugins/pre_bump_release.py
twaugh: Aha.

mmilata:
LGTM; -500 lines total, not bad;)

lcarva:
LGTM

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -8,315 +8,66 @@
...
-        :param author_name: str, name to use for git commits
-        :param author_email: str, email address for git commits
-        :param committer_name: str, name to use for git commits (else author's)
-        :param committer_email: str, email address for git commits (else
-                                     author's)
-        :param commit_message: str, git commit message
-        :param push_url: str, URL for push
+        :param target: string, koji target to use as a source
+        :param hub: string, koji hub (xmlrpc)
+        :param proxy: string, proxy user
```
lcarva: Missing `proxy` parameter?

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -8,315 +8,66 @@
...
+        component_label = get_preferred_label_key(dockerfile_labels,
+                                                  'com.redhat.component')
+        try:
+            component = dockerfile_labels[component_label]
+        except KeyError:
+            raise RuntimeError("missing label: {}".format(component_label))
 
-            # We checked out the right branch
-            assert repo.git(['rev-parse', 'HEAD']) == branch_sha
+        latest = self.xmlrpc.getLatestBuilds(self.target, package=component)
```
lcarva: `getLatestBuilds` uses `tag`, not `target`: https://pagure.io/koji/blob/master/f/hub/kojihub.py#_9111

lcarva: See https://github.com/projectatomic/atomic-reactor/pull/516 for fix.
Testing was successful.

atomic_reactor/plugins/pre_bump_release.py
twaugh: Wrong comment, will fix in a separate PR.

PR #515, opened 2016-06-09
Add tag_from_config plugin

lcarva:


mmilata:
LGTM, is this needed for building base images?

lcarva:
@mmilata It was a feature request. It's not really needed, but it gives users ability to add more tags if desired. I can be used not just for building base images, bur for any image.

atomic_reactor/plugins/post_tag_from_config.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def get_component_name(self):
+        if not self.workflow.built_image_inspect:
+            raise RuntimeError('There is no inspect data for built image. '
+                               'Has the build succeeded?')
+
+        try:
+            name = (self.workflow.built_image_inspect[INSPECT_CONFIG]
+                                                     ['Labels']
+                                                     ['Name'])
```
twaugh: We should take care here. The 'Name' label is from the 'old' set. In my auto-increment-release branch I added a `get_all_label_keys()` function that I think could help, e.g.

```
name_labels = get_all_label_keys('name')
name = None
for name_label in name_labels:
    name = ....get(name_label)
    if name:
        break
else:
    self.log.error(...)
```

atomic_reactor/plugins/post_tag_from_config.py
```diff
@@ -0,0 +1,88 @@
...
+        tags = self.get_tags()
+
+        if tags:
+            name = self.get_component_name()
+            for i, tag_suffix in enumerate(tags):
+                tag = '{}:{}'.format(name, tag_suffix)
+                self.log.debug('Using additional tag: %s', tag)
+                self.workflow.tag_conf.add_primary_image(tag)
+                # Store modified name.
+                tags[i] = tag
```
twaugh: Pretty sure this is safe while we're looping over enumerate().

lcarva: Yes. See unit tests.

atomic_reactor/plugins/post_tag_from_config.py
lcarva: Maybe also add a `get_first_value_from_all_label_keys(key, df)` ?
I know it's a terrible name.

twaugh:
LGTM

atomic_reactor/plugins/post_tag_from_config.py
twaugh: Well, there's already one for getting the 'best' choice from a Dockerfile. But this is from inspect data.

mmilata: Oops, merged it already, sorry. Seeing as this is not critical can you please open new PR if you want to add the function?

lcarva: @mmilata Since Tim's function hasn't been merged yet, I'll wait to do that once it's merged. Thanks.

PR #517, opened 2016-06-17
Reading secrets as files in tag_and_push plugin

lcarva:
Per Kubernets documentation, items defined in data portion of a secret are available as files in the container: http://kubernetes.io/docs/user-guide/secrets/#consuming-secret-values-from-volumes

### DIFF
--- a/atomic_reactor/plugins/post_tag_and_push.py
+++ b/atomic_reactor/plugins/post_tag_and_push.py
@@ -61,13 +61,9 @@ class TagAndPushPlugin(PostBuildPlugin):
             login = password = email = None
             if docker_push_secret:
                 try:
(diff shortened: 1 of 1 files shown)


vrutkovs:
Tests should be adjusted to ensure coverage is preserved

vrutkovs:
Not entirely sure about this - we're using dockercfg-style for storing secrets (as those are automatically created by Kubernetes and used during the pull), which are not files - .dockercfg is one json file and contains creds for all known registries.

lcarva:
I wasn't aware of secrets being used via dockercfg. If reading the Json
file as it was previously done works, let's go with that and close the PR.

PR #519, opened 2016-06-20
some docs fixes

twaugh:
docs/api.md is intended to be generated from docs/apidocs/ but was not entirely in sync for the boilerplate part (top.md).

The bump_release plugin documented a parameter that does not exist.


lcarva:
LGTM

PR #520, opened 2016-06-20
Mark plugin as failed if plugin crashed on init

### LINKED ISSUE #518: Build not marked as failed if plugin init fails
In the case where the plugin instance fails to be created (by providing invalid args, for example) an exception is raised but the build is not marked as failed immediately. This causes erroneous behavior on the exit plugins, since some of them will first check if the build failed prior to proceeding (`koji_promote`, for example).

The issue seems to be in `plugins.py` when `create_instance_from_plugin` is called. The call needs to be wrapped in a `try/except` block and call `self.on_plugin_failed(plugin_class.key, ex)` on any `Exception`. The `plugin_instance.run()` call already does that, so a potential fix is to just move plugin instantiation to that `try` block.

Note that `plugin_class.key` should be used instead of `plugin_instance.key` since the instance may not have been created.

vrutkovs:
Put plugin init in try..catch block so that an exception in plugin init fails the plugin, not the build.

Reproduced this in ```bump_release``` plugin with local Koji, which dissallows connections without a certificate.

Fixes #518

### DIFF
--- a/atomic_reactor/plugin.py
+++ b/atomic_reactor/plugin.py
@@ -204,11 +204,10 @@ class PluginsRunner(object):
 
             logger.debug("running plugin '%s'", plugin_name)
 
-            plugin_instance = self.create_instance_from_plugin(plugin_class, plugin_conf)
-            start_time = datetime.datetime.now()
-            self.save_plugin_timestamp(plugin_instance.key, start_time)
-
             try:
+                plugin_instance = self.create_instance_from_plugin(plugin_class, plugin_conf)
+                start_time = datetime.datetime.now()
+                self.save_plugin_timestamp(plugin_class.key, start_time)
                 plugin_response = plugin_instance.run()
             except AutoRebuildCanceledException as ex:
                 # if auto rebuild is canceled, then just reraise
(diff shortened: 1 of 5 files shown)


atomic_reactor/plugin.py
```diff
@@ -218,30 +217,31 @@ def run(self, keep_going=False):
...
-                self.save_plugin_duration(plugin_instance.key, seconds)
+                if start_time:
+                    finish_time = datetime.datetime.now()
+                    duration = finish_time - start_time
+                    seconds = duration.total_seconds()
+                    logger.debug("plugin '%s' finished in %ds", plugin_name, seconds)
+                    self.save_plugin_duration(plugin_class.key, seconds)
             except Exception:
-                logger.exception("failed to save plugin duration")
+                logger.error("failed to save plugin duration")
```
mmilata: Why?

vrutkovs: Oops, I didn't change it back. Tests use FakeLogger, which didn't have exception call - fixing it would be a better idea

mmilata:
Question inline, otherwise LGTM.

mmilata:
LGTM, merged

PR #522, opened 2016-06-22
Pass dockercfg path straight to docker-py's login

vrutkovs:
This doesn't expose the password in the logs, as docker daemon 
sends used creds as a response to login.

### DIFF
--- a/atomic_reactor/core.py
+++ b/atomic_reactor/core.py
@@ -451,9 +451,10 @@ class DockerTasker(LastLogger):
             raise RuntimeError(
                 "Failed to login to '%s': email = '%s', login = '%s', password = '%s'" %
                 (registry, email, login, password))
-        logger.debug(response)
         if u'Status' in response and response[u'Status'] == u'Login Succeeded':
             logger.info("login succeeded")
+        else:
+            logger.debug("response: %r", response)
 
     def push_image(self, image, insecure=False):
         """
--- a/atomic_reactor/core.py
+++ b/atomic_reactor/core.py
@@ -34,7 +34,8 @@ from docker.errors import APIError
 from atomic_reactor.constants import CONTAINER_SHARE_PATH, CONTAINER_SHARE_SOURCE_SUBDIR,\
         BUILD_JSON, DOCKER_SOCKET_PATH
 from atomic_reactor.source import get_source_instance_for
-from atomic_reactor.util import ImageName, wait_for_command, clone_git_repo, figure_out_dockerfile
+from atomic_reactor.util import (
+    ImageName, wait_for_command, clone_git_repo, figure_out_dockerfile, Dockercfg)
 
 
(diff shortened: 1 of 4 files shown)


atomic_reactor/core.py
```diff
@@ -440,20 +441,30 @@ def tag_image(self, image, target_image, force=False):
...
-            username=login,
-            password=password,
-            email=email,
-            registry=registry)
+    def login(self, registry, docker_secret_path):
+        """
+        login to docker registry
+
+        :param registry: registry name
+        :param dockercfg: path to docker config
```
twaugh: parameter name is wrong, should be docker_secret_path

atomic_reactor/core.py
```diff
@@ -440,20 +441,30 @@ def tag_image(self, image, target_image, force=False):
...
+        """
+        logger.info("logging in: registry '%s', secret path '%s'" % (registry, docker_secret_path))
+        # Docker-py needs username
+        dockercfg = Dockercfg(docker_secret_path)
+        username = dockercfg.get_credentials(registry)['username']
+        logger.info("found username %s for registry %s", username, registry)
+
+        dockercfg_path = os.path.join(docker_secret_path, '.dockercfg')
```
twaugh: maybe Dockercfg should expose the filename it used

atomic_reactor/core.py
vrutkovs: Fixed

atomic_reactor/core.py
vrutkovs: Agreed

atomic_reactor/core.py
```diff
@@ -440,20 +441,30 @@ def tag_image(self, image, target_image, force=False):
...
-            email=email,
-            registry=registry)
+    def login(self, registry, docker_secret_path):
+        """
+        login to docker registry
+
+        :param registry: registry name
+        :param docker_secret_path: path to docker config directory
+        """
+        logger.info("logging in: registry '%s', secret path '%s'" % (registry, docker_secret_path))
```
twaugh: Not a blocker for merge: can use separate args here instead of '%'.

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -8,6 +8,7 @@
 
 from copy import deepcopy
 import re
+import os.path
```
twaugh: Is this needed?

twaugh: Also we no longer need to import Dockercfg in this module.

twaugh:
LGTM

atomic_reactor/plugins/post_tag_and_push.py
vrutkovs: Right, fixed

PR #523, opened 2016-06-22
bump_release: add several checks to avoid incorrect builds

vrutkovs:


atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -60,7 +60,15 @@ def run(self):
 
         target_info = self.xmlrpc.getBuildTarget(self.target)
         self.log.debug('target info: %s', target_info)
+        if not target_info:
+            raise RuntimeError("no such tag: %s" % self.target)
+
         tag_id = target_info['dest_tag']
+        try:
+            assert self.xmlrpc.checkTagPackage(tag_id, component)
```
twaugh: Any reason to do this rather than just `if not self.xmlrpc.checkTagPackage(): raise ...`?

Didn't know about checkTagPackage, nice find.

vrutkovs: I wanted to protect against exception in ```checkTagPackage``` call - in case the user has insufficient permissions or our brew doesn't have it.

I'll probably revert this to a simple ```if``` check

twaugh:
LGTM

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -60,7 +60,13 @@ def run(self):
 
         target_info = self.xmlrpc.getBuildTarget(self.target)
         self.log.debug('target info: %s', target_info)
+        if not target_info:
+            raise RuntimeError("no such tag: %s" % self.target)
```
lcarva: Typo: target, not tag.

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -60,7 +60,13 @@ def run(self):
...
+
         tag_id = target_info['dest_tag']
+        if not self.xmlrpc.checkTagPackage(tag_id, component):
```
lcarva: Nice!

atomic_reactor/plugins/pre_bump_release.py
vrutkovs: Good catch, fixed

twaugh:
LGTM

PR #524, opened 2016-06-22
pulp_sync: don't rely on syncRepo() return value

twaugh:
The dockpulp syncRepo() method no longer returns the prefixed repository ID. As pulp_sync can be told what the prefix is, let's just use that to work out what the prefixed repository ID is.

dockpulp change:
https://github.com/release-engineering/dockpulp/pull/59


twaugh:
Tested.

vrutkovs:
LGTM

vrutkovs:
LGTM

PR #525, opened 2016-06-25
Add support for name and version to bump release

lcarva:
The `bump_release` plugin uses koji's `getLatestBuilds` method which groups results by name only, not by name-version. This produces incorrect behavior when a tag contains multiple versions for same component.

Since we have both name (component) and version already from Dockerfile, just use `getNextRelease`.
This works because name and version have to be unique within koji, checking the target and package seems to be unnecessary.

My only concern is that we're changing that plugin's definition which was released today. Let's consider making target an optional parameter in the mean time?

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -23,18 +23,20 @@ class BumpReleasePlugin(PreBuildPlugin):
     key = "bump_release"
     is_allowed_to_fail = False  # We really want to stop the process
 
-    def __init__(self, tasker, workflow, target, hub):
+    # The target parameter is no longer used by this plugin. It's
+    # left as an optional parameter to allow a graceful transition
+    # in osbs-client.
+    def __init__(self, tasker, workflow, hub, target=None):
         """
         constructor
 
         :param tasker: DockerTasker instance
         :param workflow: DockerBuildWorkflow instance
-        :param target: string, koji target to use as a source
         :param hub: string, koji hub (xmlrpc)
+        :param target: unused - backwards compatibility
         """
         # call parent constructor
         super(BumpReleasePlugin, self).__init__(tasker, workflow)
-        self.target = target
         self.xmlrpc = create_koji_session(hub)
 
     def run(self):
@@ -50,7 +52,6 @@ class BumpReleasePlugin(PreBuildPlugin):
             self.log.debug("release set explicitly so not incrementing")
             return
 
-        # No release labels are set so set them.
         component_label = get_preferred_label_key(dockerfile_labels,
                                                   'com.redhat.component')
         try:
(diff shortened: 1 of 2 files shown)


twaugh:
Great catch!

Yes, we need to keep target as an optional parameter and ignore it, at least for a short time.

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -50,34 +48,26 @@ def run(self):
...
 
-        # No release labels are set so set them.
         component_label = get_preferred_label_key(dockerfile_labels,
                                                   'com.redhat.component')
         try:
             component = dockerfile_labels[component_label]
         except KeyError:
             raise RuntimeError("missing label: {}".format(component_label))
 
-        target_info = self.xmlrpc.getBuildTarget(self.target)
```
vrutkovs: We shouldn't be removing those checks

lcarva: I don't think that logic belongs in this plugin. From this plugin's point of view a target is not required at all.

vrutkovs: Agreed, I'll prepare a PR to move this to ```pre_koji```

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -23,18 +23,24 @@ class BumpReleasePlugin(PreBuildPlugin):
     key = "bump_release"
     is_allowed_to_fail = False  # We really want to stop the process
 
-    def __init__(self, tasker, workflow, target, hub):
+    # The target parameter is no longer used by this plugin. It's
+    # left as an optional parameter to allow a graceful transition
+    # in osbs-client.
+    def __init__(self, tasker, workflow, target=None, hub=None):
```
twaugh: Actually plugin parameters are always passed by keyword, so it's fine to swap target and hub around and only make target optional.

lcarva: Done.

tests/plugins/test_bump_release.py
```diff
@@ -58,64 +61,69 @@ def test_component_missing(self, tmpdir):
...
-        ('comp', 'dest_tag', False),
-        ('comp', 'wrong_tag', True),
-        ('wrong_comp', 'dest_tag', True),
+    @pytest.mark.parametrize('version', [
+        {'version': '7.1'},
+        {'Version': '7.2'},
     ])
-    def test_increment(self, tmpdir, release_label, latest_builds, next_release,
-                       expected, component, tag, throws_exception):
+    @pytest.mark.parametrize('include_target', [
```
twaugh: Nice :)

twaugh:
LGTM

vrutkovs:
LGTM

PR #526, opened 2016-06-27
Move tag and target checks to pre_koji plugin

vrutkovs:


atomic_reactor/plugins/pre_koji.py
```diff
@@ -36,15 +38,33 @@ def __init__(self, tasker, workflow, target, hub, root, proxy=None):
         self.pathinfo = koji.PathInfo(topdir=root)
         self.proxy = proxy
 
+    def _check_target_and_tag(self):
```
lcarva: Do you think this might be better suited for koji-containerbuild?

vrutkovs: Not sure, we're reading 'BZComponent' from Dockerfile here, I don't think k-cb can do that, as its the only way to know the component as far as I know

lcarva: It definitely has a way to pull the component from the Dockerfile as it's currently done to check for existing nvr.
The benefit of having the check in k-cb is that the build will fail earlier.

vrutkovs: Right, in fact we already have a check for that - https://github.com/release-engineering/koji-containerbuild/blob/develop/koji_containerbuild/plugins/builder_containerbuild.py#L529. 

Closing this as its now handled in k-cb

PR #527, opened 2016-06-27
docs: updated functions in DockerTasker class

vrutkovs:
Updated documentation after registry auth support and other changes


twaugh:
I think docs/api.md is meant to be generated using docs/apidocs, but when I try it I get DockerTasker's methods listed as static methods.

twaugh:
These changes look good to me.

PR #529, opened 2016-06-28
pulp_push: get versions from manifest, not metadata

twaugh:
Fixes issue when using docker-1.10.

See https://github.com/release-engineering/dockpulp/issues/34


vrutkovs:
LGTM

vrutkovs:
Verified on local koji instance

PR #533, opened 2016-06-30
Use yum repos for base image creation

lcarva:


atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -137,6 +152,8 @@ def parse_image_build_config(self, config_file_name):
         section = 'image-build'
         for option in ('name', 'version', 'arches', 'target', 'install_tree'):
             value = config.get(section, option)
```
vrutkovs: configparser would throw ```NoOptionError``` if the option is not specified. Either we should catch it or verify with  ```config.has_option```

vrutkovs:
LGTM

twaugh:
LGTM too

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: It still would, but we're building a default config as an input to config parser which is dynamically generated. This check is to ensure that neither the config that was dynamically generated nor the config provided by user has somehow provided an empty value, such as:

```
install_tree =
```

lcarva:
Tested.

PR #535, opened 2016-07-05
koji_promote: use floating tags in extra.docker.repositories

twaugh:
This new extra field lists the primary tags that are set on the image.

### DIFF
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -471,6 +471,7 @@ class KojiPromotePlugin(ExitPlugin):
         digests = self.get_digests()
         repositories = self.get_repositories(digests)
         arch = os.uname()[4]
+        tags = set(image.tag for image in self.workflow.tag_conf.primary_images)
         metadata, output = self.get_image_output(arch)
         metadata.update({
             'arch': arch,
@@ -484,6 +485,7 @@ class KojiPromotePlugin(ExitPlugin):
                     'id': image_id,
                     'parent_id': parent_id,
                     'repositories': repositories,
+                    'tags': list(tags),
                 },
             },
         })
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -432,16 +432,18 @@ class KojiPromotePlugin(ExitPlugin):
 
         output_images = []
         for registry in registries:
-            image = self.nvr_image.copy()
-            image.registry = registry.uri
-            pullspec = image.to_str()
+            for image in self.workflow.tag_conf.primary_images:
+                image.registry = registry.uri
+                pullspec = image.to_str()
 
-            output_images.append(pullspec)
+                output_images.append(pullspec)
 
-            digest = digests.get(image.to_str(registry=False))
-            if digest:
-                digest_pullspec = image.to_str(tag=False) + "@" + digest
-                output_images.append(digest_pullspec)
+                if image == self.nvr_image:
+                    digest = digests.get(image.to_str(registry=False))
+                    if digest:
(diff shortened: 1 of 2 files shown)


twaugh:
I'm pretty sure coveralls is just lying.

twaugh:
pytest-cov seems to be getting confused by the buildimage tests. Investigating.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -471,6 +471,7 @@ def add_log_type(output):
         digests = self.get_digests()
         repositories = self.get_repositories(digests)
         arch = os.uname()[4]
+        tags = [image.tag for image in self.workflow.tag_conf.primary_images]
```
lcarva: Should this be a `set` to avoid possible duplicates? Probably not a problem now, but if we ever use multiple pulp servers, there could be a duplicate for each tag.

tests/plugins/test_koji_promote.py
```diff
@@ -737,6 +741,10 @@ def validate_output(self, output, metadata_only):
                 digest_pullspec = image.to_str(tag=False) + '@' + fake_digest(image)
                 assert digest_pullspec in repositories_digest
 
+            tags = docker['tags']
+            assert isinstance(tags, list)
+            assert all(is_string_type(tag) for tag in tags)
```
lcarva: Maybe check if each string startswith `name:` ?

tests/plugins/test_koji_promote.py
```diff
@@ -829,6 +837,46 @@ def test_koji_promote_filesystem_koji_task_id_missing(self, tmpdir, os_env,
         assert 'filesystem_koji_task_id' not in extra
         assert AddFilesystemPlugin.key in caplog.text()
 
+    @pytest.mark.parametrize('additional_tags', [
+        None,
+        ['3.2'],
+    ])
+    def test_koji_promote_image_tags(self, tmpdir, os_env, additional_tags):
```
lcarva: What's `os_env`? I see other tests also take as a parameter so it must be a fixture. It doesn't seem to be used in some of the tests in this file.

atomic_reactor/plugins/exit_koji_promote.py
twaugh: Not sure if JSON can deal with sets can it?

tests/plugins/test_koji_promote.py
twaugh: Any suggestions?

tests/plugins/test_koji_promote.py
twaugh: It's a fixture, defined further up in that module.

atomic_reactor/plugins/exit_koji_promote.py
lcarva: It cannot, but you could create it as a set and then convert it to a list once modifications are done.

tests/plugins/test_koji_promote.py
lcarva: I got nothing... I was thinking we could extract the tag from each image, but this seems wrong for a test. The expected values should be hard coded and not dynamically generated. Perhaps this validation method can take a list of expected tags? This is very minor as the `test_koji_promote_image_tags` already test the code changes.

twaugh: There is a separate test that actually tests for expected values. This 'validate' method is more about types.

lcarva:
LGTM

atomic_reactor/plugins/exit_koji_promote.py
twaugh: Take a look at the commit I just added. I can squash it before we merge.

lcarva: Perfect

vrutkovs:
LGTM, merging

PR #536, opened 2016-07-06
buildimage tests: disable these tests due to coveralls reporting 0% coverage

twaugh:
See eg https://travis-ci.org/projectatomic/atomic-reactor/jobs/142701266
near the end, where it reports total coverage as 42%, with many files listed as 0%.


lcarva:
Can't we configure coveralls to ignore some files?

twaugh:
Much better idea, yes.

PR #538, opened 2016-07-07
tag_from_config: ignore tags containing hyphens

twaugh:
If the additional-tags file is allowed to specify a tag containing a hyphen, there is a possibility of overwriting an n-v-r tag from a previous build.

The rationale for allowing additional tags in the first place is to allow a floating tag that refers to a version 'path', e.g. '3.2' which would float across versions 3.2.1, 3.2.2, and so on. Disallowing hyphens does not prevent this use-case.


vrutkovs:
LGTM

lcarva:
LGTM

PR #539, opened 2016-07-07
Revert "koji_promote: use floating tags in extra.docker.repositories"

lcarva:
This reverts commit 88a3dda1301da1e393a5d507c92899cd7e047ff0.

Still waiting on confirmation regarding this being the correct approach.


twaugh:
I believe this is resolved now.

PR #541, opened 2016-07-30
Extract baseurl from repo for base images

lcarva:


atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -113,13 +115,28 @@ def __init__(self, tasker, workflow, koji_hub,
     def is_image_build_type(self, base_image):
         return base_image.strip().lower() == 'koji/image-build'
 
+    def extract_base_url(self, repo_url):
+        response = requests.get(repo_url)
+        response.raise_for_status()
+        repo = ConfigParser()
```
twaugh: Should this be RawConfigParser or perhaps SafeConfigParser? Or maybe there's a reason to use ConfigParser here, but the docs say to prefer the Safe variant.

twaugh:
LGTM, ConfigParser thing needn't block merging.

atomic_reactor/plugins/pre_add_filesystem.py
lcarva: Looks like SafeConfigParser has been deprecated from python 3: https://docs.python.org/3/whatsnew/3.2.html#configparser
And RawConfigParser is referred as legacy in python3 docs, with a note to consider using  ConfigParser.

I think ConfigParser is the most reliable multi-version variant. https://docs.python.org/3/library/configparser.html#rawconfigparser-objects

We're not using interpolation at all in this use case. I don't think ConfigParser would be unsafe to use.

PR #546, opened 2016-08-24
Change to use dnf/clean, and update to use existing images

gbraad:
Uses `dnf` to install and clean: to remove the cached packages that have been installed.
Also changed the names of the images that are used in the build process (buildroot doesn't exist)

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,14 +125,14 @@ $ docker build -t buildroot-hostdocker .
 As soon as our build image is built, we can start building stuff in it:
 
 ```bash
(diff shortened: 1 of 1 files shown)


twaugh:
Thanks!

LGTM. One note though: the commit logs are correct but the description above is not quite correct -- it's not the installed packages that are removed from the container but the cached packages and metadata.

lcarva:
LGTM

gbraad:
@twaugh totally right... semantics...

I removed the cached packages that have been installed. Sorry, my bad...

PR #548, opened 2016-08-25
Fixed FROM version in Dockerfile

gbraad:
This to prevent breaking automated builds when a new Fedora release
gets tagged with 'latest'. This means that subsequent releases need
this file to be modified to reflect the latest Fedora release, when
confirmed nothing breaks.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
-FROM fedora:latest
+FROM fedora:24
 # e2fsprogs -- docker @ F20 wants it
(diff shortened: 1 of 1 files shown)


vrutkovs:
LGTM

lcarva:
LGTM

twaugh:
I'm not so sure we want to do this. Really it is an example Dockerfile for building the builder image. In that role, 'fedora:latest' feels like the better choice.

vrutkovs:
Hmm, [dockerhub](https://hub.docker.com/_/fedora/) says ```fedora:latest``` is actually pointing to 24, not Rawhide. I think we should keep it and check if its buildable on each major version

gbraad:
OK...

PR #549, opened 2016-08-25
Use dnf instead of yum

gbraad:
Prevents deprecation warnings...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,5 @@
 FROM fedora:latest
-# e2fsprogs -- docker @ F20 wants it
-RUN yum -y install docker git python-docker-py python-setuptools e2fsprogs koji python-pip python-backports-lzma osbs gssproxy
+RUN dnf -y install docker git python-docker-py python-setuptools e2fsprogs koji python-pip python-backports-lzma osbs gssproxy && dnf clean all
 RUN mkdir /tmp/atomic-reactor
 ADD . /tmp/atomic-reactor
 RUN cd /tmp/atomic-reactor && python setup.py install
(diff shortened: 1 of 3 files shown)


vrutkovs:
Looks good.

I think we should add ```dnf clean all``` to that RUN also

gbraad:
agreed

Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM fedora:latest
 # e2fsprogs -- docker @ F20 wants it
```
lcarva: Since we're going with `dnf` which was adopted on Fedora 22 can we remove the e2fsprogs comment as well? It seems to suggest it's only needed for Fedora 20.

gbraad: I was wondering about this statement and thought it was left in with a 'purpose'... :-P... removing...

gbraad: Q: Just the comment or also the installation of e2fsprogs?

gbraad:
Changes were pushed...

Dockerfile
```diff
@@ -1,6 +1,5 @@
 FROM fedora:latest
-# e2fsprogs -- docker @ F20 wants it
-RUN yum -y install docker git python-docker-py python-setuptools e2fsprogs koji python-pip python-backports-lzma osbs gssproxy
+RUN dnf -y install docker git python-docker-py python-setuptools e2fsprogs koji python-pip python-backports-lzma osbs gssproxy; dnf clean all
```
twaugh: Maybe `&&` instead of `;`? Not sure if it makes any difference in this context.

vrutkovs: Good catch! Its important: if the packages will fail to install it would run ```clean```, it would succeed and move on

gbraad: Exactly, it would just `dnf clean` and succeed. Changed and pushed...

vrutkovs:
Looks good. @twaugh @lcarva, please take a look

twaugh:
LGTM

PR #550, opened 2016-08-30
koji_promote: task IDs in metadata should be ints

twaugh:


twaugh:
Needs integration testing.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -512,17 +512,22 @@ def get_build(self, metadata):
         if koji_task_id is not None:
             self.log.info("build configuration created by Koji Task ID %s",
```
lcarva: This is very minor, but it might be a good idea to use %r, instead of %s. The side effect is that this prints the representation of `koji_task_id` instead of its value:

```
In [22]: logger.warning('Task ID %r', '1234')
WARNING:tcpserver:Task ID '1234'

In [23]: logger.warning('Task ID %s', '1234')
WARNING:tcpserver:Task ID 1234
```

twaugh: I changed the code below to use that on error if needed. At this point, koji_task_id is expected to be a string since it comes from the build JSON.

twaugh:
Integration testing passed.

lcarva:
LGTM

PR #551, opened 2016-08-30
rpmqa: also report package signatures

twaugh:


vrutkovs:
LGTM

twaugh:
Needs integration testing.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -370,8 +370,14 @@ def get_image_components(self):
                            PostBuildRPMqaPlugin.key)
             return []
 
+        try:
+            sep = PostBuildRPMqaPlugin.sep
+        except AttributeError:
+            # sep instance variable added in Aug 2016
```
lcarva: This implies plugins are from different atomic-reactor releases. Is that a valid scenario?

twaugh: It's defensive. I can't imagine why someone would want to do that right now, but I don't want to prevent them from doing it.

twaugh:
Integration testing passed.

lcarva:
LGTM

PR #554, opened 2016-09-09
Handle unique_tag_only flag

lcarva:
When unique_tag_only parameter is used in osbs-client, atomic-reactor's tag_by_labels will not tag image with transient tags: N:V, N:V-R, latest. Instead only the unique timestamp tag is used.

Due to this, other plugins also had to be modified to allow the unique timestamp tag to be used in lieu of any primary tags.

### DIFF
--- a/tests/plugins/test_tag_by_labels.py
+++ b/tests/plugins/test_tag_by_labels.py
@@ -20,6 +20,8 @@ from tests.constants import LOCALHOST_REGISTRY, TEST_IMAGE, INPUT_IMAGE, MOCK
 if MOCK:
     from tests.docker_mock import mock_docker
 
+import pytest
+
 
 class Y(object):
     pass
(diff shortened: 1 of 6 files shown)


tests/plugins/test_tag_by_labels.py
```diff
@@ -63,20 +69,33 @@ def test_tag_by_labels_plugin(tmpdir):
...
+        assert len(workflow.tag_conf.images) == 1
+        assert len(primary_images) == 0
+
+    else:
+        assert len(workflow.tag_conf.images) == 4
+        assert len(primary_images) == 3
+
+        assert ("%s:%s-%s" % (TEST_IMAGE, version, release)) in images
+        assert ("%s:%s" % (TEST_IMAGE, version)) in images
+        assert ("%s:latest" % (TEST_IMAGE, )) in images
```
vrutkovs: Can we also assert there is a datestamp tag here (or its not being created in the test)?

lcarva: assertions done on lines 99 and 100 are always performed. It's not a timestamp, but it's a unique tag which in this case serves the same purpose.

twaugh:
LGTM

PR #555, opened 2016-09-16
New style labels

vrutkovs:
Make sure we supply both old-style and new-style tags
    
The only exception is when old-style tag is in base image - we should set new-style only in this case


tests/plugins/test_bump_release.py
```diff
@@ -126,4 +126,5 @@ def getNextRelease(self, build_info):
         parser = DockerfileParser(plugin.workflow.builder.df_path)
         # Only the new-style spellings of the release labels should be set
         assert parser.labels['release'] == next_release
-        assert 'Release' not in parser.labels
+        if 'BZComponent' not in parser.labels.keys():
+            assert 'Release' not in parser.labels
```
twaugh: Can we add the `assert parser.labels['Release'] == next_release` assertion for the other case?

tests/plugins/test_bump_release.py
```diff
@@ -126,4 +126,5 @@ def getNextRelease(self, build_info):
         parser = DockerfileParser(plugin.workflow.builder.df_path)
         # Only the new-style spellings of the release labels should be set
```
twaugh: Needs updating?

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -132,20 +132,36 @@ def add_aliases(self, base_labels, df_labels, plugin_labels):
         applied_alias = False
         not_applied = []
 
-        for old, new in self.aliases.items():
-            if old in all_labels:
-                if new in new_labels:
-                    if all_labels[old] != all_labels[new]:
```
twaugh: Was it not any good to just set `all_labels[old] = all_labels[new]` at this point?

tests/plugins/test_bump_release.py
vrutkovs: Right, will do

atomic_reactor/plugins/pre_add_labels_in_df.py
vrutkovs: We have to make sure that the old labels are not inherited - otherwise we should be replacing those with new-style tags, right? This is what happens on line 138

vrutkovs:
@twaugh, right, rewrote the commit message, I hope it makes sense now

vrutkovs:
Renamed the var to 'is_old_inherited'

twaugh:
LGTM, thanks.

twaugh:
@lcarva, want to take a look before this gets merged?

twaugh:
Oh, thought of something else we should do before merging.

twaugh:
Any places in atomic-reactor that assume the labels 'Name', 'Version', 'Release', 'BZComponent' etc are present should be changed to look for 'name', 'version', 'release', 'com.redhat.component' etc.

vrutkovs:
Right, lets include this here, working on it

twaugh:
- [ ] tag_by_labels
- [ ] tag_from_config
- [ ] add_dockerfile (comments and error message only I think)
- [ ] assert_labels
- [ ] docs/plugins.md

atomic_reactor/plugins/post_tag_from_config.py
```diff
@@ -69,9 +69,8 @@ def get_component_name(self):
                                'Has the build succeeded?')
 
         try:
-            name = (self.workflow.built_image_inspect[INSPECT_CONFIG]
-                                                     ['Labels']
-                                                     ['name'])
+            labels = self.workflow.built_image_inspect[INSPECT_CONFIG]['Labels']
+            name = labels.get("name") or labels['name']
```
twaugh: How about using util.get_preferred_label('name')?

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -81,7 +81,8 @@ def __init__(self, tasker, workflow, labels,
             raise RuntimeError("labels have to be dict")
         self.labels = labels
         self.dont_overwrite = dont_overwrite
-        self.aliases = aliases or {}
+        default_aliases = {k: v for k, v in _PREFERRED_LABELS}
```
twaugh: This looks the wrong way round.

Also we need to make sure all these are covered:
https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md

as well as any we use from here:
https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/README.md

atomic_reactor/plugins/post_tag_from_config.py
vrutkovs: Right, implemented via `get_preferred_label_key`

atomic_reactor/plugins/post_tag_from_config.py
```diff
@@ -69,9 +70,9 @@ def get_component_name(self):
                                'Has the build succeeded?')
 
         try:
-            name = (self.workflow.built_image_inspect[INSPECT_CONFIG]
-                                                     ['Labels']
-                                                     ['name'])
+            labels = self.workflow.built_image_inspect[INSPECT_CONFIG]['Labels']
+            name_label = str(get_preferred_label_key(labels, "name"))
+            name = labels[name_label]
```
vrutkovs: No, not really. `get_preferred_label` will mask the `KeyError` if the label is missing

twaugh: I see, thanks.

atomic_reactor/plugins/pre_add_labels_in_df.py
vrutkovs: Okay, not sure if I should update `_PREFERRED_LABELS` or there is a better place for it

twaugh: Personally I think it's fine for pre_add_labels_in_df to have its own copy as it will include labels that are not inspected by atomic-reactor.

twaugh: Actually it looks like `_PREFERRED_LABELS` has all the labels required, in order (new, old).

Just need to make sure they are mapped the correct way round in `self.aliases`, i.e. `{old: new for new, old in _PREFERRED_LABELS}`.

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,14 +45,41 @@
...
 class AddLabelsPlugin(PreBuildPlugin):
     key = "add_labels_in_dockerfile"
 
-    def __init__(self, tasker, workflow, labels, dont_overwrite=("Architecture", ),
+    DEFAULT_ALIASES = {
+        'Name': 'name',
+        'Version': 'version',
+        'Release': 'release',
+        'Architecture': 'architecture',
+        'BuildDate': 'build-date',
```
twaugh: I don't think we've ever used BuildDate have we?

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,14 +45,41 @@
...
+        'Vendor': 'vendor',
+        'URL': 'url',
```
twaugh: Pretty sure we've never used URL.

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,14 +45,41 @@
...
+        'Summary': 'summary',
```
twaugh: Summary has never been used.

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,14 +45,41 @@
...
+        'Description': 'description',
```
twaugh: Description has never been used.

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,14 +45,41 @@
...
+        'VcsType': 'vcs-type',
```
twaugh: VcsType has never been used.

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,14 +45,41 @@
...
+        'VcsUrl': 'vcs-url',
```
twaugh: VcsUrl has never been used.

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,14 +45,41 @@
...
+        'VcsRef': 'ref',
```
twaugh: VcsRef has never been used.

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,14 +45,41 @@
...
+        'RUN': 'run',
```
twaugh: RUN has never been used that I know of.

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,14 +45,41 @@
...
+        'HELP': 'help',
```
twaugh: HELP has never been used that I know of.

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,14 +45,41 @@
...
+        'INSTALL': 'install',
```
twaugh: INSTALL has never been used that I know of.

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,14 +45,41 @@
...
+        'UNINSTALL': 'uninstall',
```
twaugh: UNINSTALL has never been used that I know of.

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,14 +45,41 @@
...
+        'STOP': 'stop',
```
twaugh: STOP has never been used (I'd never even heard of it until now :) ).

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,14 +45,41 @@
...
+        'DEBUG': 'debug',
```
twaugh: DEBUG has never been used that I know of.

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,14 +45,41 @@
...
+        'Authoritative_Registry': 'authoritative-source',
+        'BZComponent': 'com.redhat.component',
+        'Build_Host': 'com.redhat.build-host',
+        'BuildHost': 'com.redhat.build-host',
```
twaugh: Has BuildHost ever been used?

twaugh:
This just in -- we should use 'authoritative-source-url' and not 'authoritative-source'. It seems 'authoritative-source' may have been a mistake in the specification.
See https://github.com/projectatomic/ContainerApplicationGenericLabels/pull/74

atomic_reactor/plugins/pre_add_labels_in_df.py
vrutkovs: I don't think we did, although I don't think there is a way to know that for sure.

Anyway, it was added in https://github.com/projectatomic/ContainerApplicationGenericLabels/commit/b291b5f3f4636904ea2f01a2c5759898351c8545.

vrutkovs:
>we should use 'authoritative-source-url' and not 'authoritative-source'

Right, that's a typo, replaced it

atomic_reactor/plugins/pre_add_labels_in_df.py
vrutkovs: RUN was definitely used: `rsyslog-docker` for instance

atomic_reactor/plugins/pre_add_labels_in_df.py
vrutkovs: Removed the labels we haven't seen. I suppose we'll be adding those on request

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -45,15 +45,29 @@
...
-    def __init__(self, tasker, workflow, labels, 
+    DEFAULT_ALIASES = {
+        'Name': 'name',
+        'Version': 'version',
+        'Release': 'release',
+        'Architecture': 'architecture',
+        'Vendor': 'vendor',
+        'RUN': 'run',
+        'INSTALL': 'install',
+        'UNINSTALL': 'uninstall',
```
twaugh: OK, these atomic CLI labels are accepted as upper-case but preferred as lower-case, so this what you have here is correct.
https://github.com/projectatomic/ContainerApplicationGenericLabels/issues/8

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -69,8 +69,17 @@ def run(self):
         self.log.debug('getting next release from build info: %s', build_info)
         next_release = self.xmlrpc.getNextRelease(build_info)
 
+        # Always set preferred release label - other will be set if old-style 
+        # label is present
+        preferred_release_label = get_preferred_label_key(dockerfile_labels, 
+                                                         'release')
+        old_style_label = get_all_label_keys('com.redhat.component')[1]
```
lcarva: Shouldn't this be fetching `Release` ? Not sure why component is relevant here.

vrutkovs: We're checking if an old-style alternative for 'com.redhat.component' ('BZComponent') is present. If it is - we should set old-style release label here as well

lcarva: Ah ok. I finally understand now. Not sure if it's worth the effort, but we can probably put this logic in a helper method in utils. It might make it easier to read the code.

PR #558, opened 2016-09-23
Work around a brew bug when failed builds were imported

vrutkovs:
Koji seems to be importing failed builds due to a bug. We should check that the 
build with expected release doesn't exist and increment `next_release` if we find one

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -69,6 +69,19 @@ class BumpReleasePlugin(PreBuildPlugin):
         self.log.debug('getting next release from build info: %s', build_info)
         next_release = self.xmlrpc.getNextRelease(build_info)
 
+        # getNextRelease will return the release of the last successful build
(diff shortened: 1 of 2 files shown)


twaugh:
Can we test it?

vrutkovs:
Updated `test_bump_release`

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -69,6 +69,17 @@ def run(self):
         self.log.debug('getting next release from build info: %s', build_info)
         next_release = self.xmlrpc.getNextRelease(build_info)
 
+        # getNextRelease will return the release of the last successful build
+        # but next_release might be a failed build
```
twaugh: We could also mention that `CGImport` doesn't allow importing a build whose nvr matches that of a failed build, since that's the reason we have to do all this.

vrutkovs: Agreed, rewriting a comment

PR #559, opened 2016-09-29
Use v1 manifest digest for koji-promote

lcarva:
~~Still needs unit testing.~~ Added unit tests.

Was able to perform integration test and verified build behaved properly.

- Verified koji metadata is set
- Verified OpenShift build annotations is set
- Verified images are deleted from intermediate docker registry

~~However, I don't have access to a builder with docker >= 1.10, so both v1 and v2 manifests digests are the same.~~ Tested with docker 1.10, expected results were seen.

### DIFF
--- a/atomic_reactor/plugins/exit_delete_from_registry.py
+++ b/atomic_reactor/plugins/exit_delete_from_registry.py
@@ -42,6 +42,8 @@ class DeleteFromRegistryPlugin(ExitPlugin):
         self.registries = deepcopy(registries)
 
     def run(self):
+        deleted_digests = set()
+
         for registry, registry_conf in self.registries.items():
             if not registry.startswith('http://') and not registry.startswith('https://'):
                 registry = 'https://' + registry
@@ -71,7 +73,14 @@ class DeleteFromRegistryPlugin(ExitPlugin):
                                  registry_noschema)
                 continue
 
-            for tag, digest in push_conf_registry.digests.items():
+            for tag, digests in push_conf_registry.digests.items():
+                digest = digests.default
+                if digest in deleted_digests:
+                    # Manifest schema version 2 uses the same digest
+                    # for all tags
+                    self.log.info('digest already deleted %s', digest)
+                    continue
+
                 repo = tag.split(':')[0]
                 url = registry + "/v2/" + repo + "/manifests/" + digest
                 insecure = push_conf_registry.insecure
@@ -79,6 +88,7 @@ class DeleteFromRegistryPlugin(ExitPlugin):
 
                 if response.status_code == requests.codes.ACCEPTED:
                     self.log.info("deleted manifest %s/%s@%s", registry_noschema, repo, digest)
+                    deleted_digests.add(digest)
                 elif response.status_code == requests.codes.NOT_FOUND:
                     self.log.warning("cannot delete %s/%s@%s: not found",
                                      registry_noschema, repo, digest)
@@ -90,3 +100,5 @@ class DeleteFromRegistryPlugin(ExitPlugin):
                                                              response.reason)
                     self.log.error("%s\n%s", msg, response.text)
                     raise PluginFailedException(msg)
+
+        return deleted_digests
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -416,7 +416,11 @@ class KojiPromotePlugin(ExitPlugin):
             for image in self.workflow.tag_conf.images:
                 image_str = image.to_str()
                 if image_str in registry.digests:
-                    digest = registry.digests[image_str]
+                    # pulp/crane supports only manifest schema v1
+                    if self.workflow.push_conf.pulp_registries:
+                        digest = registry.digests[image_str].v1
+                    else:
+                        digest = registry.digests[image_str].default
                     digests[image.to_str(registry=False)] = digest
 
         return digests
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -11,7 +11,9 @@ from __future__ import print_function, unicode_literals
 import hashlib
 import json
 import os
+import re
 from pipes import quote
+import requests
 import shutil
 import subprocess
 import tempfile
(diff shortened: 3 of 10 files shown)


atomic_reactor/util.py
```diff
@@ -626,3 +627,44 @@ def get_credentials(self, docker_registry):
...
+    digests = {}
+    for version in versions:
+        headers = {
+            'Accept': 'application/vnd.docker.distribution.manifest.{}+json'.format(version)
+        }
+        response = requests.head(url, verify=not insecure, headers=headers,
+                                 auth=auth)
+        response.raise_for_status()
+
+        digests[version] = response.headers['Docker-Content-Digest']
```
twaugh: We don't know for sure which schema version we're getting. Maybe check the Content-Type header?

atomic_reactor/util.py
```diff
@@ -626,3 +627,44 @@ def get_credentials(self, docker_registry):
...
+
+
+def get_manifest_digests(image, insecure=False, dockercfg_path=None,
+                         versions=('v1', 'v2')):
+    """Return manifest digest for image.
+
+    :param image: ImageName, the remote image to inspect
+    :param insecure: bool, when True registry's cert is not verified
+    :param dockercfg_path: str, dirname of .dockercfg location
+    :param versions: tuple, which versions to fetch digest
```
twaugh: Let's make it clear we're talking about 'schema versions'.

atomic_reactor/plugins/exit_delete_from_registry.py
```diff
@@ -71,7 +71,8 @@ def run(self):
                                  registry_noschema)
                 continue
 
-            for tag, digest in push_conf_registry.digests.items():
+            for tag, digests in push_conf_registry.digests.items():
+                digest = digests['v2']
```
twaugh: Can we put a comment in there explaining why we're using the schema 2 digest?

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -416,7 +416,7 @@ def get_digests(self):
             for image in self.workflow.tag_conf.images:
                 image_str = image.to_str()
                 if image_str in registry.digests:
-                    digest = registry.digests[image_str]
+                    digest = registry.digests[image_str]['v1']
```
twaugh: Let's explain why we're using the schema 1 digest.

lcarva: Done

atomic_reactor/plugins/exit_delete_from_registry.py
lcarva: Done

atomic_reactor/util.py
lcarva: Done

atomic_reactor/util.py
lcarva: Done

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -416,7 +416,10 @@ def get_digests(self):
             for image in self.workflow.tag_conf.images:
                 image_str = image.to_str()
                 if image_str in registry.digests:
-                    digest = registry.digests[image_str]
+                    # Must use schema v1 digest because the content will
+                    # be served by pulp/crane, which does not support schema
+                    # v2 digest
+                    digest = registry.digests[image_str]['v1']
```
twaugh: Thinking about this -- won't Fedora be using Koji but not Pulp? So actually we probably need to select the correct schema version here.

lcarva: Done.

lcarva:
PTAL @twaugh @vrutkovs

atomic_reactor/util.py
```diff
@@ -626,3 +627,54 @@ def get_credentials(self, docker_registry):
...
+        dockercfg = Dockercfg(dockercfg_path).get_credentials(image.registry)
+
+        username = dockercfg.get('username')
+        password = dockercfg.get('password')
+        if username and password:
+            auth = requests.auth.HTTPBasicAuth(username, password)
+
+    context = '/'.join([x for x in [image.namespace, image.repo] if x])
+    tag = image.tag or 'latest'
+    url = 'https://{}/v2/{}/manifests/{}'.format(image.registry, context, tag)
```
twaugh: Are we happy about hard-coding the `https` scheme? (Just checking)

atomic_reactor/util.py
```diff
@@ -626,3 +627,54 @@ def get_credentials(self, docker_registry):
...
+
+        # Only compare prefix as response may use +prettyjws suffix
+        # which is the case for signed manifest
+        response_h_prefix = response.headers['Content-Type'].rsplit('+', 1)[0]
+        request_h_prefix = media_type.rsplit('+', 1)[0]
+        if response_h_prefix != request_h_prefix:
+            logger.debug('request headers: %s', headers)
+            logger.debug('response headers: %s', response.headers)
+            raise RuntimeError('Unable to fetch digest for {}'
+                               .format(media_type))
```
twaugh: I'm not sure we want this to error out here. Instead, we could either just set the digest for the requested version, or leave it unset and require that calling code wanting e.g. v2 must first try v2 and fall back to v1.

The reason raising an error might be bad is that someone might want to run this code using a Docker version earlier than 1.10, or against a v2 registry that does not support schema 2 (e.g. docker-distribution < 2.3.0), and we still want that set-up to work.

In that situation, we want e.g. Koji to use whatever digest the registry is storing the manifest for.

atomic_reactor/util.py
lcarva: The delete_from_registry plugin seems to check if the registry starts with either http or https, then falls back to https. I'm not sure what that check is all about, or even if it's ever true.

twaugh: delete_from_registry is looking at the configuration it was given which is meant to include the URI scheme to use for the registry. The pulp_sync plugin also needs to know the URI scheme to use for the registry in order to tell it to Pulp.

Here, 'image.registry' is the registry name for use by Docker, and Docker generally guesses (tries https and then http).

So either we should do the same try-https-then-http thing, or we assume https and don't allow it to be overridden, or else we'll need to be told the scheme somehow.

If we assume https and don't allow configuration for it, we should probably note in the documentation that v2 registries must be accessible via https.

lcarva: How about we enhance this method to also take the registry url from the plugin, that way we can implement the same logic: if a URI scheme is not defined, fallback to https, otherwise use whatever is already there. Thoughts?

twaugh: That sounds good.

lcarva: Done

atomic_reactor/util.py
lcarva: Done

atomic_reactor/util.py
```diff
@@ -626,3 +628,78 @@ def get_credentials(self, docker_registry):
...
+            logger.debug('request headers: %s', headers)
+            logger.debug('response headers: %s', response.headers)
+            logger.warning('Unable to fetch digest for %s', media_type)
+            continue
+
+        digests[version] = response.headers['Docker-Content-Digest']
+        logger.debug('Image %s:%s has %s manifest digest: %s',
+                     context, tag, version, digests[version])
+
+    return ManifestDigest(**digests)
```
twaugh: Should it raise an exception if none could be found?

lcarva: Yes, it should. Added check.

atomic_reactor/util.py
```diff
@@ -626,3 +628,81 @@ def get_credentials(self, docker_registry):
...
+
+        # Only compare prefix as response may use +prettyjws suffix
+        # which is the case for signed manifest
+        response_h_prefix = response.headers['Content-Type'].rsplit('+', 1)[0]
+        request_h_prefix = media_type.rsplit('+', 1)[0]
+        if response_h_prefix != request_h_prefix:
+            logger.debug('request headers: %s', headers)
+            logger.debug('response headers: %s', response.headers)
+            logger.warning('Unable to fetch digest for %s', media_type)
+            continue
```
twaugh: I don't think the lines leading up to `continue` have test coverage. Is that hard to add?

lcarva: Good catch! Added test for this specific scenario.

twaugh:
LGTM

PR #561, opened 2016-10-06
koji_promote: store resulting image config

vrutkovs:
This change will make sure brew/koji extra info has container info from the registry. After the image was pushed registry is queried for container config using v2 schema 2 digest

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -174,6 +174,8 @@ class DockerRegistry(Registry):
         """
         super(DockerRegistry, self).__init__(uri, insecure=insecure)
         self.digests = {}  # maps tags (str) to their digest, if available
+        self.config = None  # stores image config from the registry, 
+        # media type of the config is application/vnd.docker.container.image.v1+json
 
 
 class PushConf(object):
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -16,6 +16,7 @@ from string import ascii_letters
 import subprocess
 from tempfile import NamedTemporaryFile
 import time
+import copy
 
 from atomic_reactor import __version__ as atomic_reactor_version
 from atomic_reactor import start_time as atomic_reactor_start_time
@@ -476,6 +477,14 @@ class KojiPromotePlugin(ExitPlugin):
         # Parent of squashed built image is base image
         image_id = self.workflow.builder.image_id
         parent_id = self.workflow.base_image_inspect['Id']
+
+        # Read config from the registry using v2 schema 2 digest
+        config = copy.deepcopy(self.workflow.push_conf.docker_registries[0].config)
+
+        # We don't need container_config section
+        if config and 'container_config' in config:
+            del config['container_config']
+
         digests = self.get_digests()
         repositories = self.get_repositories(digests)
         arch = os.uname()[4]
@@ -494,9 +503,13 @@ class KojiPromotePlugin(ExitPlugin):
                     'parent_id': parent_id,
                     'repositories': repositories,
                     'tags': list(tags),
+                    'config': config
                 },
             },
         })
+
+        if not config:
+            del metadata['extra']['docker']['config']
 
         # Add the 'docker save' image to the output
         image = add_buildroot_id(output)
--- a/atomic_reactor/plugins/post_tag_and_push.py
+++ b/atomic_reactor/plugins/post_tag_and_push.py
@@ -7,11 +7,10 @@ of the BSD license. See the LICENSE file for details.
 """
 
 from copy import deepcopy
-import re
 
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.plugins.exit_remove_built_image import defer_removal
-from atomic_reactor.util import get_manifest_digests
+from atomic_reactor.util import get_manifest_digests, get_config_from_registry
 
 
 __all__ = ('TagAndPushPlugin', )
@@ -50,6 +49,8 @@ class TagAndPushPlugin(PostBuildPlugin):
         if not self.workflow.tag_conf.unique_images:
             self.workflow.tag_conf.add_unique_image(self.workflow.image)
 
+        first_v2_digest = None
+        first_registry_image = None
         for registry, registry_conf in self.registries.items():
             insecure = registry_conf.get('insecure', False)
             push_conf_registry = \
@@ -73,11 +74,19 @@ class TagAndPushPlugin(PostBuildPlugin):
 
                 digests = get_manifest_digests(registry_image, registry,
                                                insecure, docker_push_secret)
-
                 tag = registry_image.to_str(registry=False)
                 push_conf_registry.digests[tag] = digests
 
+                if not first_v2_digest and digests.v2:
+                    first_v2_digest = digests.v2
+                    first_registry_image = registry_image
+
+            if first_v2_digest:
+                push_conf_registry.config = get_config_from_registry(
+                    first_registry_image, registry, first_v2_digest, insecure,
+                    docker_push_secret, 'v2')
+            else:
+                self.log.info("V2 schema 2 digest is not available")
 
         self.log.info("All images were tagged and pushed")
         return pushed_images
-
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -649,8 +649,12 @@ class ManifestDigest(object):
         return self.v2 or self.v1
 
 
-def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
-                         versions=('v1', 'v2')):
+def get_manifest_media_type(version):
+    return 'application/vnd.docker.distribution.manifest.{}+json'.format(version)
+
+
+def query_registry(image, registry, digest=None, insecure=False, dockercfg_path=None,
+                   version='v1'):
     """Return manifest digest for image.
 
     :param image: ImageName, the remote image to inspect
@@ -658,9 +662,9 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
                           https:// will be used
     :param insecure: bool, when True registry's cert is not verified
     :param dockercfg_path: str, dirname of .dockercfg location
-    :param versions: tuple, which manifest schema versions to fetch digest
+    :param versions: str, which manifest schema version to fetch digest
 
-    :return: dict, versions mapped to their digest
+    :return: requests.Response object
     """
     auth = None
     if dockercfg_path:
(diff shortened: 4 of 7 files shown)


atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -26,7 +26,8 @@
 from atomic_reactor.constants import PROG
 from atomic_reactor.util import (get_version_of_tools, get_checksums,
                                  get_build_json, get_preferred_label,
-                                 get_docker_architecture)
+                                 get_docker_architecture,
```
lcarva: None of the added imports are not used.

atomic_reactor/util.py
```diff
@@ -649,18 +649,19 @@ def default(self):
         return self.v2 or self.v1
 
 
-def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
-                         versions=('v1', 'v2')):
+def query_registry(image, registry, digest=None, insecure=False, dockercfg_path=None,
```
lcarva: What if we just query the registry once? We could enhance `ManifestDigest` class to also store config contents from manifest.

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -73,11 +73,12 @@ def run(self):
 
                 digests = get_manifest_digests(registry_image, registry,
                                                insecure, docker_push_secret)
-
                 tag = registry_image.to_str(registry=False)
                 push_conf_registry.digests[tag] = digests
 
+                push_conf_registry.config = get_config_from_registry(
```
lcarva: Adding a dynamic attribute like this will increase the difficulty of understanding the code base. I would've expected `DockerRegistry` to at least define it.

tests/plugins/test_tag_and_push.py
```diff
@@ -152,5 +157,5 @@ def test_tag_and_push_plugin(tmpdir, monkeypatch, image_name, logs, should_raise
         if MOCK:
             # we only test this when mocking docker because we don't expect
             # running actual docker against v2 registry
-            assert workflow.push_conf.docker_registries[0].digests[image_name]['v1'] == DIGEST_V1
-            assert workflow.push_conf.docker_registries[0].digests[image_name]['v2'] == DIGEST_V2
+            assert workflow.push_conf.docker_registries[0].digests[image_name].v1 == DIGEST_V1
```
lcarva: Thanks for fixing these :)

tests/plugins/test_tag_and_push.py
```diff
@@ -121,7 +121,12 @@ def test_tag_and_push_plugin(tmpdir, monkeypatch, image_name, logs, should_raise
 
     (flexmock(atomic_reactor.util)
         .should_receive('get_manifest_digests')
-        .and_return({'v1': DIGEST_V1, 'v2': DIGEST_V2})
+        .and_return(ManifestDigest(v1=DIGEST_V1, v2=DIGEST_V2))
+    )
+
+    (flexmock(atomic_reactor.util)
+        .should_receive('get_config_from_registry')
+        .and_return({'id': '123', 'container_config': {}, 'config': {}, 'docker_version': '1.9.1'})
```
lcarva: Please add assertion that config is actually set.

atomic_reactor/util.py
```diff
@@ -706,3 +737,25 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
 
     return ManifestDigest(**digests)
 
+
+def get_config_from_registry(image, registry, digest, insecure=False,
```
lcarva: Missing tests in `test_util.py`

atomic_reactor/plugins/exit_koji_promote.py
vrutkovs: Fixed

atomic_reactor/util.py
lcarva: Won't a `get` also return the digest? or just `head` ?

atomic_reactor/plugins/post_tag_and_push.py
vrutkovs: It is defined, in `atomic_reactor/inner.py`

tests/plugins/test_tag_and_push.py
vrutkovs: Agreed

atomic_reactor/plugins/post_tag_and_push.py
lcarva: Ah right, missed it,

atomic_reactor/util.py
twaugh: `get` also will.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -476,6 +476,16 @@ def add_log_type(output):
...
         parent_id = self.workflow.base_image_inspect['Id']
+
+        # Read config from the registry using v2 schema 2 digest
+        config = self.workflow.push_conf.docker_registries[0].config
+
+        # We don't need container_config section
+        for k in config.keys():
+            if k.lower() == 'container_config':
+                del config[k]
+                break
```
twaugh: You can just
```
if 'container_config' in config:
    del config['container_config']
```

atomic_reactor/inner.py
```diff
@@ -174,6 +174,7 @@ def __init__(self, uri, insecure=False):
         """
         super(DockerRegistry, self).__init__(uri, insecure=insecure)
         self.digests = {}  # maps tags (str) to their digest, if available
+        self.config = None  # stores image config from the registry
```
twaugh: Media type for this is `application/vnd.docker.container.image.v1+json` -- can we include that in the comment so we know exactly what this is intended to be?

atomic_reactor/plugins/exit_koji_promote.py
twaugh: Stupid observation, but does this modify the object that is stored in `self.workflow.push_conf.docker_registries[0]`?

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -73,11 +73,13 @@ def run(self):
 
                 digests = get_manifest_digests(registry_image, registry,
                                                insecure, docker_push_secret)
-
                 tag = registry_image.to_str(registry=False)
                 push_conf_registry.digests[tag] = digests
 
+            first_registry_image = self.workflow.tag_conf.images[0]
+            push_conf_registry.config = get_config_from_registry(
+                first_registry_image, registry, digests.v2, insecure,
```
twaugh: What happens when `digests.v2` is `None` here? i.e. when running this using Docker 1.9?

atomic_reactor/inner.py
vrutkovs: Right, will do

atomic_reactor/plugins/post_tag_and_push.py
twaugh: No, v2 schema 1 is no good. The config object is not available that way.

If the schema 2 manifest is not available, no `config` key should be added to the extra metadata.

vrutkovs: Fixed - if `digests.v2` is None we print an info message

atomic_reactor/util.py
vrutkovs: Currently we make 3 requests to the registry - 2 for digests and 1 for config. I'll try to come with a way to optimize this, but I think this is acceptable for now

atomic_reactor/plugins/exit_koji_promote.py
vrutkovs: Ah, indeed. Removed the key from a deepcopy of the registry object

atomic_reactor/inner.py
twaugh: I think you added this to the metadata... really I just wanted it mentioned in the above comment. I don't think we want it in the metadata.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -494,6 +502,7 @@ def add_log_type(output):
                     'parent_id': parent_id,
                     'repositories': repositories,
                     'tags': list(tags),
+                    'config': config
```
twaugh: Could add a trailing "," here so that future appending is easier.

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -7,11 +7,11 @@
 """
 
 from copy import deepcopy
-import re
 
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.plugins.exit_remove_built_image import defer_removal
-from atomic_reactor.util import get_manifest_digests
+from atomic_reactor.util import (
+    get_manifest_digests, get_config_from_registry, get_media_type)
```
twaugh: `get_media_type` not used

atomic_reactor/util.py
```diff
@@ -649,18 +649,22 @@ def default(self):
...
+                   version='v1'):
     """Return manifest digest for image.
 
     :param image: ImageName, the remote image to inspect
     :param registry: str, URI for registry, if URI schema is not provided,
                           https:// will be used
     :param insecure: bool, when True registry's cert is not verified
     :param dockercfg_path: str, dirname of .dockercfg location
-    :param versions: tuple, which manifest schema versions to fetch digest
+    :param versions: int, which manifest schema version to fetch digest
```
twaugh: This doesn't seem to  describe the parameter correctly (int vs str).

atomic_reactor/util.py
```diff
@@ -649,18 +649,22 @@ def default(self):
         return self.v2 or self.v1
 
 
-def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
-                         versions=('v1', 'v2')):
+def get_media_type(version):
+    return 'application/vnd.docker.distribution.manifest.{}+json'.format(version)
```
twaugh: Is this module-level function name descriptive enough? Wondered if it should be `get_manifest_media_type`.

atomic_reactor/inner.py
vrutkovs: Oops, I've misunderstood that. Removed

atomic_reactor/util.py
vrutkovs: Fixed

atomic_reactor/plugins/post_tag_and_push.py
vrutkovs: Right, removed it

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -494,6 +506,7 @@ def add_log_type(output):
                     'parent_id': parent_id,
                     'repositories': repositories,
                     'tags': list(tags),
+                    'config': config,
```
twaugh: If we don't have a v2 schema 2 config object we use use to fill this entry in, it should be omitted altogether (as currently).

In other words it should be present and populated, or not present at all.

atomic_reactor/util.py
```diff
@@ -675,17 +679,39 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
         registry = 'https://{}'.format(registry)
 
     context = '/'.join([x for x in [image.namespace, image.repo] if x])
-    tag = image.tag or 'latest'
+    tag = digest or image.tag or 'latest'
```
twaugh: Nit-pick (won't block PR): we could call this `reference` or `ref`, seeing as it could be either a tag or a digest.
cf. https://docs.docker.com/registry/spec/api/#pulling-an-image-manifest

atomic_reactor/util.py
```diff
@@ -706,3 +734,25 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
...
+    :param dockercfg_path: str, dirname of .dockercfg location
+    :param versions: tuple, which manifest schema versions to fetch digest
+
+    :return: dict, versions mapped to their digest
+    """
+    response = query_registry(
+        image, registry, digest=digest, insecure=insecure,
+        dockercfg_path=dockercfg_path, version=version)
+    response_json = response.json()
+    history = response_json['history']
```
twaugh: This is wrong. We aren't looking for response.history[0].v1Compatibility -- we are looking for response, plain and simple, just parsed as JSON.

v1Compatibility is a v2 schema 1 thing. We are not using that for fetching config (and never will).

tests/plugins/test_koji_promote.py
```diff
@@ -731,6 +735,7 @@ def validate_output(self, output, metadata_only):
                 'id',
                 'repositories',
                 'tags',
+                'config'
```
twaugh: The test should be: either config is present and truthy, or it is not present.

atomic_reactor/util.py
vrutkovs: Agreed, fixed

tests/plugins/test_koji_promote.py
vrutkovs: `config` is always present, its being set explicitly in `mock_environment`

twaugh: Maybe we need to add a test for what happens when it is not present then?

vrutkovs: Yeah, that's a good idea. I'm gonna add one after some testing on actual docker 1.9 / 1.10 environments

atomic_reactor/util.py
vrutkovs: Fixed and tested on docker 1.10 - a correct JSON has been returned

atomic_reactor/plugins/exit_koji_promote.py
vrutkovs: Fixed and tested on docker 1.9 - this key is removed if config cannot be fetched

tests/plugins/test_tag_and_push.py
twaugh: In addition this would mean we'd get coverage here:
https://coveralls.io/builds/8279004/source?filename=atomic_reactor%2Fplugins%2Fpost_tag_and_push.py#L89

vrutkovs: Implemented, this line is now covered

twaugh:
Looks great!

PR #562, opened 2016-10-12
koji_promote: store correct config

vrutkovs:
Store a correct config in koji's metadata: its a config of the blob, identified by a digest in the response from the registry

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -654,15 +654,17 @@ def get_manifest_media_type(version):
 
 
 def query_registry(image, registry, digest=None, insecure=False, dockercfg_path=None,
-                   version='v1'):
+                   version='v1', is_blob=False):
     """Return manifest digest for image.
 
     :param image: ImageName, the remote image to inspect
     :param registry: str, URI for registry, if URI schema is not provided,
                           https:// will be used
+    :param digest: str, digest of the image manifest
     :param insecure: bool, when True registry's cert is not verified
     :param dockercfg_path: str, dirname of .dockercfg location
-    :param versions: str, which manifest schema version to fetch digest
+    :param version: str, which manifest schema version to fetch digest
+    :param is_blob: bool, read blob config if set to True
 
     :return: requests.Response object
     """
@@ -680,7 +682,10 @@ def query_registry(image, registry, digest=None, insecure=False, dockercfg_path=
 
     context = '/'.join([x for x in [image.namespace, image.repo] if x])
     reference = digest or image.tag or 'latest'
-    url = '{}/v2/{}/manifests/{}'.format(registry, context, reference)
+    object_type = 'manifests'
+    if is_blob:
+        object_type = 'blobs'
+    url = '{}/v2/{}/{}/{}'.format(registry, context, object_type, reference)
     logger.debug("url: {}".format(url))
 
     headers = {'Accept': (get_manifest_media_type(version))}
@@ -742,9 +747,10 @@ def get_config_from_registry(image, registry, digest, insecure=False,
     :param image: ImageName, the remote image to inspect
     :param registry: str, URI for registry, if URI schema is not provided,
                           https:// will be used
+    :param digest: str, digest of the image manifest
     :param insecure: bool, when True registry's cert is not verified
     :param dockercfg_path: str, dirname of .dockercfg location
-    :param versions: tuple, which manifest schema versions to fetch digest
+    :param version: str, which manifest schema versions to fetch digest
 
     :return: dict, versions mapped to their digest
     """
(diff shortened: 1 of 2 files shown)


atomic_reactor/util.py
```diff
@@ -752,10 +752,16 @@ def get_config_from_registry(image, registry, digest, insecure=False,
         image, registry, digest=digest, insecure=insecure,
         dockercfg_path=dockercfg_path, version=version)
     response.raise_for_status()
+    config_digest = response.json()['config']['digest']
+
+    config_response = query_registry(
+        image, registry, digest=config_digest, insecure=insecure,
+        dockercfg_path=dockercfg_path, version=version)
```
twaugh: I just realised -- this will try to fetch /manifests/$config_digest but that's not what we want.

The config.digest in the manifest is a *blob* digest.

i.e. we want to do the equivalent of this:
```
curl -H "Accept: application/vnd.docker.distribution.manifest.v2+json" \
  $REGISTRY/v2/$REPO/manifests/$MANIFEST_DIGEST | \
  jq -r .config.digest | \
  xargs -i curl $REGISTRY/v2/$REPO/blobs/\{} | \
  jq .
```

vrutkovs: Yes, it does that, a new argument `is_blob` (probably a bad name) is setting URL to use `blobs` instead of `manifests`. This is fixed in the latest comment.

atomic_reactor/util.py
```diff
@@ -681,7 +682,10 @@ def query_registry(image, registry, digest=None, insecure=False, dockercfg_path=
 
     context = '/'.join([x for x in [image.namespace, image.repo] if x])
     reference = digest or image.tag or 'latest'
-    url = '{}/v2/{}/manifests/{}'.format(registry, context, reference)
+    config_type = 'manifests'
```
twaugh: Maybe `object_type`?

atomic_reactor/util.py
twaugh: Ah, ok. Would you mind interactively re-basing to merge that final commit together with the first one so the commit history adds logical steps in order?

tests/plugins/test_tag_and_push.py
```diff
@@ -130,32 +130,88 @@ def test_tag_and_push_plugin(
...
-        'layers': [
-            {'mediaType': 'application/vnd.docker.image.rootfs.diff.tar.gzip',
-             'digest': 'sha256:16dc1f96e3a1bb628be2e00518fec2bb97bd5933859de592a00e2eb7774b6ecf',
-             'size': 71907148},
-            {'mediaType': 'application/vnd.docker.image.rootfs.diff.tar.gzip',
-             'digest': 'sha256:024c492078957644297e3a76e58ff97558398e8b7ea7702a7f9ac8885fe52626',
-             'size': 3945714}
-        ],
-        'schemaVersion': 2,
+    MANIFEST_DIGEST = 'sha256:2c782e3a93d34d89ea4cf54052768be117caed54803263dd1f3798ce42aac14e'
```
vrutkovs: Agreed

atomic_reactor/util.py
vrutkovs: Sounds good

atomic_reactor/util.py
vrutkovs: Done

atomic_reactor/util.py
```diff
@@ -742,20 +748,29 @@ def get_config_from_registry(image, registry, digest, insecure=False,
...
         dockercfg_path=dockercfg_path, version=version)
     response.raise_for_status()
+    logger.debug("response json: {}".format(response.json()))
+    config_digest = response.json()['config']['digest']
+
+    config_response = query_registry(
+        image, registry, digest=config_digest, insecure=insecure,
+        dockercfg_path=dockercfg_path, version=version, is_blob=True)
+    config_response.raise_for_status()
+    logger.debug("response json: {}".format(config_response.json()))
```
twaugh: landscape suggests using `logging.debug("response json: %s". config_response.json())`. Could even do `config = response.json()` since we are using the result immediately.

atomic_reactor/util.py
```diff
@@ -742,20 +748,29 @@ def get_config_from_registry(image, registry, digest, insecure=False,
...
-    :param versions: tuple, which manifest schema versions to fetch digest
+    :param version: str, which manifest schema versions to fetch digest
 
     :return: dict, versions mapped to their digest
     """
     response = query_registry(
         image, registry, digest=digest, insecure=insecure,
         dockercfg_path=dockercfg_path, version=version)
     response.raise_for_status()
+    logger.debug("response json: {}".format(response.json()))
```
vrutkovs: Agreed, fixed

atomic_reactor/util.py
vrutkovs: Fixed

vrutkovs:
Updated the tests (instead of mocking `get_manifest_digests` function we're mocking request.Response now) and trying out on docker 1.9/1.10

PR #564, opened 2016-10-13
Warn on failure to remove container

lcarva:


vrutkovs:
Looks good, but maybe using `defer_removal` would be a better idea?

lcarva:
`defer_removal` is for images, not containers. I can add the same functionality for containers in the form of a new plugin (similar to `exit_remove_built_image.py`), but I think this is overkill. This container is created, started by the `post_rpmqa.py` plugin (and only used by it), I think it makes perfect sense to remove it here.

vrutkovs:
>defer_removal is for images, not containers.

Ah, right

> I think it makes perfect sense to remove it here.

Agreed

PR #565, opened 2016-10-18
new distribution_scope plugin for applying scoping rules

twaugh:


atomic_reactor/plugins/pre_distribution_scope.py
```diff
@@ -0,0 +1,107 @@
...
+    If the distribution-scope for this image is set to a less
+    restrictive value than the distribution-scope label inherited from
+    the parent image, fail the build.
+    """
+
+    # Name of this plugin
+    key = 'distribution_scope'
+
+    # The label and its possible values.
+    # https://github.com/projectatomic/ContainerApplicationGenericLabels
```
lcarva: Would it be better for this to be in the docstring where the values are listed out?

atomic_reactor/plugins/pre_distribution_scope.py
```diff
@@ -0,0 +1,107 @@
...
+    def get_scope(self, which, labels):
+        """
+        :param which: str, description of the image this belongs to
+        :param labels: dict, labels on the image
+        """
+
+        try:
+            scope_choice = labels[self.SCOPE_LABEL]
+        except KeyError:
+            self.log.debug("no distribution scope set for %s image", which)
```
lcarva: Maybe fallback to the least permissive value if one is not found?

atomic_reactor/plugins/pre_distribution_scope.py
```diff
@@ -0,0 +1,107 @@
...
+            parent_scope = self.get_scope('parent', parent_labels)
+        except NothingToCheck:
+            self.log.debug("no checks performed")
+            return
+        else:
+            if scope > parent_scope:
+                self.log.error("%s=%s set but parent has %s=%s",
+                               self.SCOPE_LABEL, self.SCOPE_NAME[scope],
+                               self.SCOPE_LABEL, self.SCOPE_NAME[parent_scope])
+                raise DisallowedDistributionScope
```
lcarva: If we use the error message to initialize the exception, it will show up in the results section in koji making it easier to debug.

atomic_reactor/plugins/pre_distribution_scope.py
```diff
@@ -0,0 +1,107 @@
...
+        "public",
+    ]
+
+    def __init__(self, tasker, workflow):
+        """
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        """
+
+        self.workflow = workflow
```
twaugh: I'll remove it.

atomic_reactor/plugins/pre_distribution_scope.py
twaugh: That's fine for the parent image, but we probably want the default value for the current image to be 'inherit from parent', which is technically what will happen once we build it.

I figured it was easiest to explicitly not perform checks when that label is not present.

atomic_reactor/plugins/pre_distribution_scope.py
```diff
@@ -0,0 +1,107 @@
...
+            labels = DockerfileParser(self.workflow.builder.df_path).labels
+            scope = self.get_scope('current', labels)
+
+            # Find out the parent's intended scope
+            parent_labels = self.workflow.builder.base_image_inspect['Labels']
+            parent_scope = self.get_scope('parent', parent_labels)
+        except NothingToCheck:
+            self.log.debug("no checks performed")
+            return
+        else:
```
twaugh: Will remove.

atomic_reactor/plugins/pre_distribution_scope.py
lcarva: Seems reasonable.

atomic_reactor/plugins/pre_distribution_scope.py
twaugh: Fixed.

atomic_reactor/plugins/pre_distribution_scope.py
```diff
@@ -0,0 +1,104 @@
...
+        except NothingToCheck:
+            self.log.debug("no checks performed")
+            return
+
+        if scope > parent_scope:
+            error = ("{label}={scope} but parent has {label}={parent_scope}"
+                     .format(label=self.SCOPE_LABEL,
+                             scope=self.SCOPE_NAME[scope],
+                             parent_scope=self.SCOPE_NAME[parent_scope]))
+            self.log.error("%s", error)
```
lcarva: Why not just `self.log.error(error)`  ?

twaugh: I'm glad you asked. ;) I want to avoid unexpected behaviour when someone uses e.g. `LABEL distribution-scope="%r"`.

atomic_reactor/plugins/pre_distribution_scope.py
twaugh: Yes, I'll change that.

twaugh:
Re-based.

atomic_reactor/plugins/pre_distribution_scope.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def run(self):
+        try:
+            # Find out the intended scope for this image
+            labels = DockerfileParser(self.workflow.builder.df_path).labels
+            scope = self.get_scope('current', labels)
+
+            # Find out the parent's intended scope
+            parent_labels = self.workflow.base_image_inspect[INSPECT_CONFIG]['Labels']
+            parent_scope = self.get_scope('parent', parent_labels)
```
lcarva: This fails for base image builds:

```
2016-10-19 15:00:33,186 - atomic_reactor.plugin - DEBUG - Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/atomic_reactor/plugin.py", line 221, in run
  File "/usr/lib/python2.7/site-packages/atomic_reactor/plugins/pre_distribution_scope.py", line 94, in run
    parent_scope = self.get_scope('parent', parent_labels)
  File "/usr/lib/python2.7/site-packages/atomic_reactor/plugins/pre_distribution_scope.py", line 72, in get_scope
    scope_choice = labels[self.SCOPE_LABEL]
TypeError: 'NoneType' object has no attribute '__getitem__'
```

twaugh: Nice catch. https://github.com/projectatomic/atomic-reactor/pull/569

PR #569, opened 2016-10-19
distribution_scope: handle imported images

twaugh:
These have {'Labels': None} in their inspection data. They occur during creation of base images (pre_add_filesystem).


lcarva:
Verified it fixes the issue.

PR #570, opened 2016-10-20
docs: describe 'required' in plugins.md

twaugh:


docs/plugins.md
```diff
@@ -23,12 +23,15 @@ Build plugins are requested and configured via input json: key `prebuild_plugins
...
         "args1": "value"
-    }
+    },
+    "required": true
 }
 ```
 
 Order is important, because plugins are executed in the order as they are specified (one plugin can use input from another plugin). `args` are directly passed to a plugin in constructor. Any plugin with `is_allowed_to_fail` set to `false` that raises an exception causes the build process to proceed directly to the stage of running exit plugins.
 
+The optional `required` key, which defaults to `true`, specifies whether this plugin is required for a successful build. If the plugin is not available and `required` is set to `false`, the build will not fail. However if the plugin is available and that plugin sets `is_allowed_to_fail` to `true`, the plugin can still cause the build to fail. This is useful for validation plugins not present in older builder images.
```
csomh: Are the plugins setting the `is_allowed_to_fail` flags? Isn't this a configuration for the plugins?

twaugh: It's a property of the plugin implementation.
https://github.com/projectatomic/atomic-reactor/blob/1.6.17/atomic_reactor/plugin.py#L47

csomh: oh, ok, now it's clear, no more comments :)

twaugh: Ah, but I've got the sense of it the wrong way round. Will fix.

PR #571, opened 2016-10-20
Limit release to 128 chars

vrutkovs:
See https://github.com/docker/docker/issues/8445

PTAL @twaugh @lcarva

### DIFF
--- a/atomic_reactor/plugins/post_tag_by_labels.py
+++ b/atomic_reactor/plugins/post_tag_by_labels.py
@@ -66,6 +66,15 @@ class TagByLabelsPlugin(PostBuildPlugin):
         version = get_label("version")
         release = get_label("release")
 
+        # docker tags cannot be longer than 128 chars
+        # see https://github.com/docker/docker/pull/8447
+        longest_tag = "%s-%s" % (version, release)
+        if len(longest_tag) > 128:
+            self.log.warn('Release is too long, docker only supports 128 char tags')
+            # Make sure that release is short enough - remove the length of version
+            # Remove one char for a dash and one as we count from zero
+            release = release[:128 - len(version) - 2]
+
         nvr = "%s:%s-%s" % (name, version, release)
         nv = "%s:%s" % (name, version)
         n = "%s:latest" % name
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/post_tag_by_labels.py
```diff
@@ -66,6 +66,11 @@ def get_label(label_name):
         version = get_label("version")
         release = get_label("release")
 
+        # docker tags cannot be longer than 128 chars
+        # see https://github.com/docker/docker/pull/8447
+        if len(release) > 128:
+            release = release[:127]
```
vrutkovs: Fixed. Do we want to print the resulting tag? It would be visible after pulp push anyway, but it might make sense to print it early in the log

atomic_reactor/plugins/post_tag_by_labels.py
```diff
@@ -66,6 +66,12 @@ def get_label(label_name):
         version = get_label("version")
         release = get_label("release")
 
+        # docker tags cannot be longer than 128 chars
+        # see https://github.com/docker/docker/pull/8447
+        if len(release) > 128:
+            self.log.warn('Release is too long, docker only supports 128 char tags')
+            release = release[:127]
```
twaugh: Actually the longest tag would be `"%s-%s" % (version, release)"`, so maybe that's what we should be checking. And thinking about it more, maybe we should just fail if we get here. I can't think of any circumstance where truncating would result in something the user wanted.

vrutkovs: Ah, good catch! Updated the code and the test.

>we should just fail if we get here

Its probably worth doing in osbs-client or k-cb then, but this should be confirmed. This happened during throughput testing with incorrect parameters and didn't actually occur on prod or any real image

twaugh: osbs-client is probably the right place, yes. That's also where we handle limitations on BuildConfig name lengths.

vrutkovs:
This should be done in osbs-client

PR #572, opened 2016-10-20
push_to_pulp: only publish if pulp_sync not configured

twaugh:


atomic_reactor/plugins/post_push_to_pulp.py
```diff
@@ -285,6 +294,17 @@ def __init__(self, tasker, workflow, pulp_registry_name, load_squashed_image=Non
         self.username = username
         self.password = password
 
+        found_self = False
+        self.publish = True
+        for plugin in self.workflow.postbuild_plugins_conf or []:
+            if plugin['name'] == self.key:
+                found_self = True
```
lcarva: Besides embracing a personal journey, why is this plugin finding itself?

lcarva: Ah to ensure `PulpSyncPlugin` comes *after* to perform publish

twaugh:
Log message added for when publishing is deferred.

PR #574, opened 2016-10-21
pulp_pull: new plugin

twaugh:
When adding content to Pulp, the conversion to v2 schema 1 causes the pulled image ID to be different than the one we re-imported from the squash plugin.

This plugin pulls the image from Pulp's Crane instance in order to discover the image ID.

### DIFF
--- a/atomic_reactor/plugins/post_pulp_pull.py
+++ b/atomic_reactor/plugins/post_pulp_pull.py
@@ -47,16 +47,12 @@ class PulpPullPlugin(PostBuildPlugin):
         start = time()
 
         # Work out the name of the image to pull
-        image = None
-        for unique_image in self.workflow.tag_conf.unique_images:
-            image = unique_image
-            break
-
-        if not image:
-            raise RuntimeError('Unable to determine unique image name')
+        assert self.workflow.tag_conf.unique_images  # must be set
+        image = self.workflow.tag_conf.unique_images[0]
 
         assert self.workflow.push_conf.pulp_registries  # must be configured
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -0,0 +1,88 @@
...
+        # call parent constructor
+        super(PulpPullPlugin, self).__init__(tasker, workflow)
+        self.timeout = timeout
+        self.retry_delay = retry_delay
+
+    def run(self):
+        start = time()
+
+        # Work out the name of the image to pull
+        image = None
```
vrutkovs: Maybe
```
try:
    image = self.workflow.tag_conf.unique_images[0]
except IndexError:
    raise RuntimeError('Unable to determine unique image name')
```
would be simplier?

atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -0,0 +1,84 @@
...
+        while True:
+            # Pull the image from Crane
+            name = self.tasker.pull_image(pullspec)
+
+            # Inspect it
+            try:
+                metadata = self.tasker.inspect_image(name)
+            except NotFound:
+                if time() - start > self.timeout:
+                    raise CraneTimeoutError("{}s exceeded".format(self.timeout))
```
lcarva: `'600s exceeded'` ?

twaugh: Suggestions for something better are welcome. Maybe "Time exceeded: {}s"?

PR #577, opened 2016-10-26
rpmqa: log exception info if remove fails

twaugh:
For some exceptions repr() actually gives less information than str().

Since this exception is not a business-as-usual occurrence and may need investigation, guard against this by using the built-in exception display functionality of the logging module.


twaugh:
@lcarva

PR #578, opened 2016-10-31
Update docs to prefer secrets over sourceSecret

lcarva:


lcarva:
> We should also remove cli/secret.py.

Isn't that still relevant for newer versions of OpenShift?
`cli/secret.py` provides `pulpsecret-gen` which seems to still be useful in creating `Secret` objects in OpenShift (although it could be updated to use `apiVersion` `'v1'` instead of `'v1beta3'`).

twaugh:
It doesn't do anything `oc secrets new` can't do.

PR #581, opened 2016-11-02
koji_promote: encode logs to UTF-8

vrutkovs:
Make sure logs are converted to UTF-8 correctly

### DIFF
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -339,8 +339,8 @@ class KojiPromotePlugin(ExitPlugin):
             # Deleted once closed
             logfile = NamedTemporaryFile(prefix=self.build_id,
                                          suffix=".log",
-                                         mode='w')
-            logfile.write(logs)
+                                         mode='wb')
+            logfile.write(logs.encode('UTF-8'))
             logfile.flush()
             metadata = self.get_output_metadata(logfile.name,
                                                 "openshift-final.log")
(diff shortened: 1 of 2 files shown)


tests/plugins/test_koji_promote.py
```diff
@@ -223,15 +223,15 @@ def mock_environment(tmpdir, session=None, name=None,
     (flexmock(OSBS)
         .should_receive('get_build_logs')
         .with_args(BUILD_ID)
-        .and_return('build logs'))
+        .and_return(u'build logs - \u2018 \u2017 \u2019'))
```
lcarva: do you need the `u''` prefix here? This file uses `unicode_literals`.

vrutkovs: Right, I don't really need it

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -339,7 +339,8 @@ def get_logs(self):
             # Deleted once closed
             logfile = NamedTemporaryFile(prefix=self.build_id,
                                          suffix=".log",
-                                         mode='w')
+                                         mode='w',
+                                         encoding='UTF-8')
```
lcarva: Only available in python3. In python 2, there's codec.open, but not sure about temp file.

vrutkovs:
Tested with rsyslog and previously failing cfme build - worked on both.

Not sure why coverage doesn't like this change though

lcarva:
I think the coverage drop is because it's using the 3.X build result, not the 2.7 build result. Maybe it just picks a random build? Anyways, I rebuilt it and now coveralls is happy.

PR #585, opened 2016-11-15
Run exit plugins if the build has been cancelled

vrutkovs:
Docker sends SIGTERM to a container when its stopped (OSE build is cancelled).

This PR traps the signal and throws an  exception in the process if the build or pre-/post-build plugins are running. The signal handler will ignore the exception if exit plugins are running.

Default docker timeout is 30 secs, so exit plugins have plenty of time to finish.

TODO:
 - [x] Add tests

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -13,10 +13,12 @@ import json
 import logging
 import tempfile
 import datetime
+import signal
 
 from atomic_reactor.build import InsideBuilder
-from atomic_reactor.plugin import PostBuildPluginsRunner, PreBuildPluginsRunner, InputPluginsRunner, PrePublishPluginsRunner, \
-    ExitPluginsRunner, PluginFailedException, AutoRebuildCanceledException
+from atomic_reactor.plugin import (
+    PostBuildPluginsRunner, PreBuildPluginsRunner, InputPluginsRunner, PrePublishPluginsRunner,
+    ExitPluginsRunner, PluginFailedException, AutoRebuildCanceledException, BuildCanceledException)
 from atomic_reactor.source import get_source_instance_for
 from atomic_reactor.util import ImageName
 
@@ -312,6 +314,9 @@ class DockerBuildWorkflow(object):
             self._base_image_inspect = self.builder.tasker.inspect_image(self.builder.base_image)
         return self._base_image_inspect
 
+    def throw_canceled_build_exception(self, *args, **kwargs):
+        raise BuildCanceledException("Build was canceled")
+
     def build_docker_image(self):
         """
         build docker image
@@ -320,6 +325,7 @@ class DockerBuildWorkflow(object):
         """
         self.builder = InsideBuilder(self.source, self.image)
         try:
+            signal.signal(signal.SIGTERM, self.throw_canceled_build_exception)
             # time to run pre-build plugins, so they can access cloned repo
             logger.info("running pre-build plugins")
             prebuild_runner = PreBuildPluginsRunner(self.builder.tasker, self, self.prebuild_plugins_conf,
@@ -379,6 +385,8 @@ class DockerBuildWorkflow(object):
 
             return build_result
         finally:
+            # We need to make sure all exit plugins are executed
+            signal.signal(signal.SIGTERM, lambda *args: None)
             exit_runner = ExitPluginsRunner(self.builder.tasker, self,
                                             self.exit_plugins_conf,
                                             plugin_files=self.plugin_files)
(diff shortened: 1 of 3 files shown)


twaugh:
Looks like the right approach to me.

atomic_reactor/inner.py
```diff
@@ -337,6 +341,7 @@ def build_docker_image(self):
             start_time = datetime.datetime.now()
             self.plugins_timestamps['dockerbuild'] = start_time.isoformat()
 
+            signal.signal(signal.SIGTERM, self.throw_cancelled_build_exception)
```
lcarva: Do you think this should be prior to prebuild plugins? I think for now there's nothing in there that needs to be cleaned up, but not sure if it'll always be the case.

twaugh: +1

vrutkovs: Agreed, moving this up, as it would affect chain rebuilds in the future - `autorebuild_canceled` should be set in this case. Fixed

atomic_reactor/inner.py
```diff
@@ -312,6 +313,9 @@ def base_image_inspect(self):
             self._base_image_inspect = self.builder.tasker.inspect_image(self.builder.base_image)
         return self._base_image_inspect
 
+    def throw_cancelled_build_exception(self, *args, **kwargs):
+        raise RuntimeError("Build was cancelled")
```
twaugh: Can we define and use a specific exception for this? Perhaps `CancelError`? That way plugins can perform any specific cancel operations they need to by catching it.

vrutkovs: Good point, the signal handler now raises `BuildCanceledException`

tests/test_inner.py
```diff
@@ -848,3 +867,59 @@ def test_workflow_plugin_results():
...
+    Verifies that exit plugins are executed when the build is cancelled
+    """
+    this_file = inspect.getfile(PreRaises)
+    mock_docker()
+    fake_builder = MockInsideBuilder()
+    flexmock(InsideBuilder).new_instances(fake_builder)
+
+    watch_pre = Watcher()
+    watch_prepub = Watcher()
+    watch_post = WatcherWithPause()
```
twaugh: Would be nice if we could parametrize this to have the pause in each of these three phases and check that all phases up to the one we sent the signal to completed, that the one we sent the signal to did not complete, and that the exit phase did complete.

I guess this doesn't test what happens when the actual build step receives the signal but that will be easier once it is its own plugin phase.

tests/test_inner.py
```diff
@@ -848,3 +867,59 @@ def test_workflow_plugin_results():
...
+    workflow = DockerBuildWorkflow(MOCK_SOURCE, 'test-image',
+                                   prebuild_plugins=[{'name': 'pre_watched',
+                                                      'args': {
+                                                          'watcher': watch_pre
+                                                      }}],
+                                   prepublish_plugins=[{'name': 'prepub_watched',
+                                                        'args': {
+                                                            'watcher': watch_prepub,
+                                                        }}],
+                                   postbuild_plugins=[{'name': 'post_watched_raises',
```
twaugh: Why 'raises'? Wouldn't 'post_watched' do the job?

tests/test_inner.py
```diff
@@ -848,3 +867,59 @@ def test_workflow_plugin_results():
...
+        target=lambda: (
+            sleep(2),
+            os.kill(pid, signal.SIGTERM)))
+    thread.start()
+
+    workflow.build_docker_image()
+
+    assert ("plugin 'post_watched_raises' raised an exception:" +
+            " BuildCanceledException('Build was cancelled',)",) in fake_logger.warnings
+    assert watch_exit.was_called()
```
twaugh: We should assert whether each phase was called.

tests/test_inner.py
vrutkovs: Ah, its a remainder from the first attempt - I tried to make it fail. Replacing with `post_watched`

tests/test_inner.py
vrutkovs: Fixed

tests/test_inner.py
vrutkovs: Correct, I'll parametrize the test to make sure the signal is called during each phase

atomic_reactor/inner.py
```diff
@@ -312,6 +314,9 @@ def base_image_inspect(self):
             self._base_image_inspect = self.builder.tasker.inspect_image(self.builder.base_image)
         return self._base_image_inspect
 
+    def throw_cancelled_build_exception(self, *args, **kwargs):
```
lcarva: **NITPICK ALERT** I know it's silly, but I can never remember how to spell canceled (1 L vs 2 LLs), can we at least make `throw_cancelled_build_exception` and `BuildCanceledException` consistent ? :)

vrutkovs: Ah, agree

PR #586, opened 2016-11-16
add parent_env handling

maxamillion:

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -196,7 +196,13 @@ class AddLabelsPlugin(PreBuildPlugin):
         else:
             base_image_labels = config["Labels"] or {}
 
-        dockerfile = DockerfileParser(self.workflow.builder.df_path)
+            try:
+                base_image_env = config["Env"]
+            except KeyError:
+                self.log.debug("Parent Environment not found, not applied to Dockerfile")
+                base_image_env = {}
+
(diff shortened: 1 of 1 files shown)


maxamillion:
https://github.com/DBuildService/dockerfile-parse/issues/20

maxamillion:
I'm pretty sure the tests are failing because this PR to dockerfile-parse is needed: https://github.com/DBuildService/dockerfile-parse/pull/21

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -196,7 +196,13 @@ def run(self):
...
             base_image_labels = config["Labels"] or {}
 
-        dockerfile = DockerfileParser(self.workflow.builder.df_path)
+            try:
+                base_image_env = config["Env"]
+            except KeyError:
+                self.log.debug("Parent Environment not found, not applied to Dockerfile")
+                base_image_env = {}
+
+        dockerfile = DockerfileParser(self.workflow.builder.df_path, parent_env=base_image_env)
```
twaugh: Could we wrap this in a `try: ... except:` block to catch `TypeError` and warn about using an old parser if `base_image_env` is non-empty?

lcarva: Maybe this can be handled by requiring the minimum version in requirements.txt?

twaugh: Doing it this way allows us to release dockerfile-parse async, if that's worth anything.

PR #587, opened 2016-11-21
use df_parser wrapper of DockerfileParser for consistent parent_env handling

maxamillion:

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -822,9 +822,9 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
                 tmp_env = parent_config["Env"]
                 logger.debug("Parent Config ENV: %s" % tmp_env)
 
-                if type(tmp_env) is dict:
+                if isinstance(tmp_env, dict):
                     p_env = tmp_env
-                elif type(tmp_env) is list:
+                elif isinstance(tmp_env, list):
                     try:
                         for key_val in tmp_env:
                             key, val = key_val.split("=")
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -20,7 +20,9 @@ import tempfile
 import logging
 import uuid
 
-from atomic_reactor.constants import DOCKERFILE_FILENAME, TOOLS_USED
+from atomic_reactor.constants import DOCKERFILE_FILENAME, TOOLS_USED, INSPECT_CONFIG
+
+from dockerfile_parse import DockerfileParser
 
 try:
     from importlib import import_module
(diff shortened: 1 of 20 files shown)


atomic_reactor/util.py
```diff
@@ -785,3 +787,71 @@ def get_config_from_registry(image, registry, digest, insecure=False,
     logger.debug('Image %s:%s has config:\n%s', context, tag, blob_config)
 
     return blob_config
+
+def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, parent_env=None):
```
lcarva: This function gives a lot of flexibility. But most of it is not yet used. Consider having a wrapper function for the most common use case which can be done by simply taking the workflow object.

maxamillion: I don't follow, it only adds one option that the DockerfileParser itself does not already offer. How is that too many?

lcarva: `DockerfileParser` is a library that takes into account different use cases. 
We only use a subset of it. Since this is a util function in `atomic-reactor`, it can be made more specific which would lessen complexity.

To make this as easy to use as possible, I think we should only have to pass workflow object.

maxamillion: @lcarva `DockerfileParser` is used through out `atomic-reactor` in places where workflow objects don't exist. Should I just leave those `DockerfileParser` and only use `df_parser` wrapper where necessary? My main motivation for going "all-in" was an attempt to not introduce ambiguity.

lcarva: You are correct. An oversight on my end. Could you add tests?

maxamillion: @lcarva absolutely, I should have to begin with. Tests coming up!

maxamillion: @lcarva I looked at writing a test where I pass in the Workflow and extract the `parent_env` as well as passing the `parent_env` by hand but I realized that the assertions on those tests will fail or pass depending on the version of `dockerfile-parse` at least until the version with support for `parent_env` is officially released.

twaugh:
What would be really great would be to get coverage on all those paths in its tests. Coveralls says there are still 11 lines that don't get executed.

maxamillion:
Agreed, but like I commented to Lui, we can't test those yet until there's a new release of dockerfile-parse. Otherwise the assertions of "what should happen" won't be accurate.

twaugh:
Oh yes, got it.

atomic_reactor/util.py
```diff
@@ -785,3 +787,71 @@ def get_config_from_registry(image, registry, digest, insecure=False,
...
+        try:
+            parent_config = workflow.base_image_inspect[INSPECT_CONFIG]
+        except (AttributeError, TypeError, KeyError):
+            logger.debug("base image unable to be inspected")
+        else:
+            try:
+                tmp_env = parent_config["Env"]
+                logger.debug("Parent Config ENV: %s" % tmp_env)
+
+                if type(tmp_env) is dict:
```
twaugh: `isinstance()`?

atomic_reactor/util.py
```diff
@@ -785,3 +787,71 @@ def get_config_from_registry(image, registry, digest, insecure=False,
...
+                    p_env = tmp_env
+                elif type(tmp_env) is list:
```
twaugh: `isinstance()`?

maxamillion:
@twaugh fixed the `isinstance` type checks

PR #590, opened 2016-11-28
Make sure build container is removed if build was unsuccessful

### LINKED ISSUE #575: atomic-reactor should remove a container if dockerbuild stage has failed
Atomic-reactor leaves a container in 'Created' state if dockerbuild has failed, which clutters the nodes. AR should be able to remove those automatically

vrutkovs:
Fixes #575 

Docker-py's `forcerm` option makes sure that all intermediate containers are removed ii the build is unsuccessful


twaugh:
Does it work?

vrutkovs:
Yes, tested on osbs-dev - the container was removed when the build has failed. Since the magic happens in `docker-py` we can't unfortunately test it reliably

PR #591, opened 2016-11-29
Disable pulp_pull plugin functionality

lcarva:


twaugh:
This will switch from using v2 schema 1 image IDs (from pulling from Crane's v2 schema 1 manifest) to using v2 schema 2 image IDs (created by docker-squash). That doesn't sound like what you want, as the v2 schema 2 image IDs will not correspond to anything in Koji when Pulp is used.

(In any case, can this change not be done using an osbs-client change alone?)

lcarva:
We discovered that pub relies on the image ID in the koji build archive metadata to be the v1 image ID.
The ID reported by squash is also not the correct one. Instead we need to inspect the tarball for the correct value.

v1 image ID, as in pulp's docker_image content unit.

Without this change pub cannot publish images as it fails to verify the docker_image content units were properly copied.

twaugh:
Are you sure this change will get you the v1 image ID? It seems unlikely.

lcarva:
We have verified that with these changes the image_id reported by OSBS matches the value reported by  the registry via:
```
curl -sLk http://<registry>/v1/repositories/<repo>/tags
```

We have also verified that pub works properly with these changes.

Additionally, search for the docker_image content unit in pulp with the reported image_id returns the correct data.

PR #592, opened 2016-12-02
Fix koji krbV auth handling

maxamillion:


twaugh:
LGTM

puiterwijk:
Do note that in some setups krb_rdns:false might not be what you want.
But I'd say that when you get setups that need it to be true, they can file a PR to get that fixed as another config option.

twaugh:
Agreed.

PR #593, opened 2016-12-06
add python{2,3}-pytest-runner to Requires: in spec when checks are run

maxamillion:

### DIFF
--- a/atomic-reactor.spec
+++ b/atomic-reactor.spec
@@ -50,6 +50,7 @@ BuildRequires:  python2-devel
 BuildRequires:  python-setuptools
 %if 0%{?with_check}
 BuildRequires:  pytest
+BuildRequires:  python2-pytest-runner
 BuildRequires:  python-pytest-capturelog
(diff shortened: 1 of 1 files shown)


maxamillion:
@twaugh Either way :)

lcarva:
Adding those dependencies would break stuff for rhel, as Tim mentioned.
Two possible alternate approaches:

1. Add `sed` in spec file to nuke `setup_requires` dependencies:
```
%prep
%setup -qn %{name}-%{commit}

sed -i '/setup_requires/d' setup.py
```

2. Modify `setup.py` to read `setup_requires` items from a file (like requirements.txt), and modify `spec` to clear that file.

twaugh:
We should drop the pytest-runner dependency entirely and just document how to run tests, as in https://github.com/projectatomic/osbs-client/pull/503.

lcarva:
@twaugh should this be dropped then?

twaugh:
Yes.

Issue #596, opened 2016-12-14
Not compatible with Docker 1.12


pkgw:
When I try to build a test image on my machine running Docker 1.12, I get an error:

```
$ atomic-reactor build git --method hostdocker --build-image buildroot-fedora --image test-image --uri "https://github.com/TomasTomecek/docker-hello-world.git"
2016-12-14 17:23:59,226 - atomic_reactor.outer - INFO - building image 'test-image'
2016-12-14 17:23:59,227 - atomic_reactor.core - INFO - building image 'buildroot-fedora' in container using docker from host
2016-12-14 17:23:59,227 - atomic_reactor.core - INFO - checking whether image 'buildroot-fedora' exists
2016-12-14 17:23:59,230 - atomic_reactor.core - INFO - creating container from image 'buildroot-fedora' and running it
2016-12-14 17:23:59,646 - atomic_reactor - ERROR - exception caught: APIError(HTTPError('400 Client Error: Bad Request for url: http+docker://localunixsocket/v1.24/containers/ef5c8f24671ec5329024d7b75fafdb599898692278f0e87450171e86d3166f5b/start',),)
```

From the Docker server logs, it appears that a deprecated argument is being passed to the Docker API call that starts the container:

```
Dec 14 17:23:59 lambda.localdomain dockerd-current[1201]: time="2016-12-14T17:23:59.644762301-05:00" level=error msg="Handler for POST /v1.24/containers/ef5c8f24671ec5329024d7b75fafdb599898692278f0e87450171e86d3166f5b/start returned error: starting container with HostConfig was deprecated since v1.10 and removed in v1.12"
```

It's not exactly clear to me which parameter is the illegal one. I'm running atomic-reactor version 1.6.17.

twaugh:
Seems related: https://github.com/docker/docker-py/issues/1267

PR #598, opened 2016-12-14
Fix deprecated API usage that was breaking us on Docker 1.12 (#596)

pkgw:
This is my stab at fixing issue #596. This change presumably breaks the client for too-old versions of Docker, but I haven't looked into which versions count as "too old".


atomic_reactor/core.py
```diff
@@ -158,9 +159,12 @@ def build_image_dockerhost(self, build_image, json_args_path):
                 open(os.path.join(json_args_path, BUILD_JSON)).read())
         container_id = self.tasker.run(
             ImageName.parse(build_image),
-            create_kwargs={'volumes': [DOCKER_SOCKET_PATH, json_args_path]},
-            start_kwargs={'binds': volume_bindings,
-                          'privileged': True}
+            create_kwargs={'volumes': [DOCKER_SOCKET_PATH, json_args_path],
+                           'host_config': create_host_config (
+                               binds = volume_bindings,
+                               privileged =True
```
twaugh: Minor nit: I don't think we want spaces before '(' or around '=' in these three lines, and likewise for the other change.

twaugh:
Thanks for this! Added review.

twaugh:
Investigating this some more it looks like the ability to set HostConfig has been around since earlier than this commit: https://github.com/docker/docker/commit/16e649c589ce60ba56812a8fffb98aee34fceb94, and that went into v1.8.0:

```
$ git describe --contains 16e649c589ce60ba56812a8fffb98aee34fceb94
v1.8.0-rc1~549^2
```

atomic_reactor/core.py
pkgw: Ah, sorry, totally forgot about checking for style consistency. I'll force-push a tweaked commit momentarily.

PR #599, opened 2016-12-15
Change yum to dnf for fedora based distro

trishnaguha:


twaugh:
Thanks!

I wonder if we should really use `-y` though. I'd rather let it prompt so users know what will be installed.

trishnaguha:
@twaugh Sure.

PR #603, opened 2016-12-20
Add plugin update_parent_image_stream_tag

lcarva:


.travis.yml
```diff
@@ -8,7 +8,7 @@ install:
   - "pip install -r requirements.txt"
   - "pip install -r tests/requirements.txt"
   - "pip install pytest-cov coveralls"
-  - "pip install git+https://github.com/projectatomic/osbs-client"
+  - "pip install git+https://github.com/projectatomic/osbs-client@ensure-image-stream-tag"
```
lcarva: This needs to be reverted before merge. It was added simply to show that unit tests are working as expected until `ensure-image-stream-tag` is merged into `osbs-client`.

lcarva:
Requires https://github.com/projectatomic/osbs-client/pull/492

tests/plugins/test_update_parent_image_stream_tag.py
```diff
@@ -0,0 +1,125 @@
...
+            assert (workflow
+                    .prebuild_results[UpdateParentImageStreamTagPlugin.key] is
+                    changed)
+
+        elif status_code == 404:
+            runner.run()
+            # ImageStreamTag is never changed if ImageStream is missing
+            assert (workflow
+                    .prebuild_results[UpdateParentImageStreamTagPlugin.key] is
+                    None)
```
twaugh: Shouldn't it return `False` in this case?

atomic_reactor/plugins/pre_update_parent_image_stream_tag.py
```diff
@@ -0,0 +1,89 @@
...
+
+        image_stream_id, tag_name = self.image_stream_tag.split(':')
+        try:
+            image_stream = osbs.get_image_stream(image_stream_id).json()
+        except OsbsResponseException as exc:
+            if exc.status_code == 404:
+                self.log.warning(
+                    'Parent image stream, %s, not found, skipping',
+                    image_stream_id)
+                return
```
twaugh: Should this return `False` instead of implicitly `None`?

lcarva: I was trying to use `None` vs `False` to indicate different scenarios. I do realize now that's just confusing. I'll change it to return False instead.

tests/plugins/test_update_parent_image_stream_tag.py
```diff
@@ -0,0 +1,125 @@
...
+
+            (_osbs.should_receive('ensure_image_stream_tag')
+             .with_args(image_stream, image_stream_tag_name, bool(scheduled))
+             .once()
+             .and_return(changed))
+
+        else:
+            (get_image_stream
+             .and_raise(OsbsResponseException('error', status_code))
+             .and_return(None))
```
twaugh: I think that says 'raise an exception the first time it is called, then return None subsequently'.

atomic_reactor/plugins/pre_update_parent_image_stream_tag.py
```diff
@@ -0,0 +1,91 @@
...
+        self.image_stream_tag = image_stream_tag
+        self.openshift_url = openshift_url
+        self.verify_ssl = verify_ssl
+        self.use_auth = use_auth
+        self.scheduled = scheduled
+
+    def _get_osbs(self):
+        """Create and return an instance of OSBS."""
+        namespace = get_build_json().get("metadata", {}).get('namespace')
+        osbs_conf = Configuration(conf_file=None,
```
lcarva: This needs to also take into account the build_json_dir, like post_import_image does.

lcarva: Done.

PR #605, opened 2016-12-21
don't overwrite distribution-scope label in dockerfile

rcerven:
allow users to define distribution-scope label in dockerfile, it will work unless you specify wider scope than in parent image

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -67,7 +67,7 @@ class AddLabelsPlugin(PreBuildPlugin):
     }
 
     def __init__(self, tasker, workflow, labels,
-                 dont_overwrite=("Architecture", "architecture", ),
+                 dont_overwrite=("distribution-scope", ),
                  auto_labels=("build-date",
                               "architecture",
                               "vcs-type",
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -67,13 +67,14 @@ class AddLabelsPlugin(PreBuildPlugin):
     }
 
     def __init__(self, tasker, workflow, labels,
-                 dont_overwrite=dont_overwrite=("Architecture", "architecture"),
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -67,24 +67,28 @@ class AddLabelsPlugin(PreBuildPlugin):
...
+                 aliases=None,
+                 dont_overwrite_if_in_dockerfile=("distribution-scope",)):
         """
         constructor
 
         :param tasker: DockerTasker instance
         :param workflow: DockerBuildWorkflow instance
         :param labels: dict, key value pairs to set as labels; or str, JSON-encoded dict
         :param dont_overwrite: iterable, list of label keys which should not be overwritten
+                               IF they are present in baseimage!
```
twaugh: Let's call it the "parent image" in the docstring, despite what the variables are called in the implementation.

tests/plugins/test_add_labels.py
```diff
@@ -295,3 +295,34 @@ def test_add_labels_aliases(tmpdir, docker_tasker, caplog,
...
+    setattr(workflow, 'builder', X)
+    flexmock(workflow, base_image_inspect=labels_conf_base)
+    setattr(workflow.builder, 'df_path', df.dockerfile_path)
+
+    runner = PreBuildPluginsRunner(
+        docker_tasker,
+        workflow,
+        [{
+            'name': AddLabelsPlugin.key,
+            'args': {'labels': {"distribution-scope": "restricted"}, "dont_overwrite_if_in_dockerfile": ["distribution-scope"], "auto_labels": [],
```
twaugh: I wonder if we should test that this is the default behaviour, i.e. not provide a `dont_overwrite_if_in_dockerfile` parameter, as well.

twaugh:
Are there any docs updates needed? I think maybe the existing docs already cover it (docs/plugins.md says "Labels that are specified in the builder configuration, such as the vendor name, distribution scope, and authoritative registry, are added to the Dockerfile using LABEL").

PR #606, opened 2016-12-21
new labels fix and new labels alias dictionary

rcerven:
fix name for 1 new label name

added new labels alias dictionary

### DIFF
--- a/atomic_reactor/constants.py
+++ b/atomic_reactor/constants.py
@@ -62,12 +62,12 @@ TOOLS_USED = (
 DEFAULT_DOWNLOAD_BLOCK_SIZE = 10 * 1024 * 1024 # 10Mb
 
 LABEL_ALIASES = {
-    name : ('name', 'Name'),
-    version : ('version', 'Version'),
-    release : ('release', 'Release'),
-    arch : ('architecture', 'Architecture'),
-    vendor : (('vendor', 'Vendor'),
-    source : ('authoritative-source-url', 'Authoritative_Registry'),
-    component : ('com.redhat.component', 'BZComponent'),
-    host : ('com.redhat.build-host', 'Build_Host')
+    "name" : ('name', 'Name'),
+    "version" : ('version', 'Version'),
+    "release" : ('release', 'Release'),
+    "arch" : ('architecture', 'Architecture'),
+    "vendor" : (('vendor', 'Vendor'),
+    "source" : ('authoritative-source-url', 'Authoritative_Registry'),
+    "component" : ('com.redhat.component', 'BZComponent'),
+    "host" : ('com.redhat.build-host', 'Build_Host')
 }
--- a/atomic_reactor/constants.py
+++ b/atomic_reactor/constants.py
@@ -60,14 +60,3 @@ TOOLS_USED = (
 )
 
 DEFAULT_DOWNLOAD_BLOCK_SIZE = 10 * 1024 * 1024 # 10Mb
-
-LABEL_ALIASES = {
-    "name" : ('name', 'Name'),
-    "version" : ('version', 'Version'),
-    "release" : ('release', 'Release'),
-    "arch" : ('architecture', 'Architecture'),
-    "vendor" : ('vendor', 'Vendor'),
-    "source" : ('authoritative-source-url', 'Authoritative_Registry'),
-    "component" : ('com.redhat.component', 'BZComponent'),
-    "host" : ('com.redhat.build-host', 'Build_Host')
-}
--- a/atomic_reactor/constants.py
+++ b/atomic_reactor/constants.py
@@ -66,7 +66,7 @@ LABEL_ALIASES = {
     "version" : ('version', 'Version'),
     "release" : ('release', 'Release'),
     "arch" : ('architecture', 'Architecture'),
-    "vendor" : (('vendor', 'Vendor'),
+    "vendor" : ('vendor', 'Vendor'),
     "source" : ('authoritative-source-url', 'Authoritative_Registry'),
     "component" : ('com.redhat.component', 'BZComponent'),
     "host" : ('com.redhat.build-host', 'Build_Host')
--- a/atomic_reactor/constants.py
+++ b/atomic_reactor/constants.py
@@ -60,3 +60,14 @@ TOOLS_USED = (
 )
 
 DEFAULT_DOWNLOAD_BLOCK_SIZE = 10 * 1024 * 1024 # 10Mb
+
+LABEL_ALIASES = {
+    name : ('name', 'Name'),
+    version : ('version', 'Version'),
+    release : ('release', 'Release'),
+    arch : ('architecture', 'Architecture'),
+    vendor : (('vendor', 'Vendor'),
+    source : ('authoritative-source-url', 'Authoritative_Registry'),
+    component : ('com.redhat.component', 'BZComponent'),
+    host : ('com.redhat.build-host', 'Build_Host')
+}
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -854,37 +854,36 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
 
 
 class Labels(object):
+    LABEL_TYPE_NAME = object()
+    LABEL_TYPE_VERSION = object()
+    LABEL_TYPE_RELEASE = object()
+    LABEL_TYPE_ARCH = object()
+    LABEL_TYPE_VENDOR = object()
+    LABEL_TYPE_SOURCE = object()
+    LABEL_TYPE_COMPONENT = object()
+    LABEL_TYPE_HOST = object()
+    LABEL_NAMES = {
+        LABEL_TYPE_NAME : ('name', 'Name'),
+        LABEL_TYPE_VERSION : ('version', 'Version'),
+        LABEL_TYPE_RELEASE : ('release', 'Release'),
+        LABEL_TYPE_ARCH : ('architecture', 'Architecture'),
+        LABEL_TYPE_VENDOR : ('vendor', 'Vendor'),
+        LABEL_TYPE_SOURCE : ('authoritative-source-url', 'Authoritative_Registry'),
+        LABEL_TYPE_COMPONENT : ('com.redhat.component', 'BZComponent'),
+        LABEL_TYPE_HOST : ('com.redhat.build-host', 'Build_Host')
+    }
+
     def __init__(self, df_labels):
         """
         Create a new Labels object
         providing access to actual newest labels as well as old ones
         """
-        self.LABEL_TYPE_NAME = 1
-        self.LABEL_TYPE_VERSION = 2
-        self.LABEL_TYPE_RELEASE = 3
-        self.LABEL_TYPE_ARCH = 4
-        self.LABEL_TYPE_VENDOR = 5
-        self.LABEL_TYPE_SOURCE = 6
-        self.LABEL_TYPE_COMPONENT = 7
-        self.LABEL_TYPE_HOST = 8
-
-        self._label_names = {
-            self.LABEL_TYPE_NAME : ('name', 'Name'),
-            self.LABEL_TYPE_VERSION : ('version', 'Version'),
-            self.LABEL_TYPE_RELEASE : ('release', 'Release'),
-            self.LABEL_TYPE_ARCH : ('architecture', 'Architecture'),
-            self.LABEL_TYPE_VENDOR : ('vendor', 'Vendor'),
-            self.LABEL_TYPE_SOURCE : ('authoritative-source-url', 'Authoritative_Registry'),
-            self.LABEL_TYPE_COMPONENT : ('com.redhat.component', 'BZComponent'),
-            self.LABEL_TYPE_HOST : ('com.redhat.build-host', 'Build_Host')
-        }
-
         self._df_labels = df_labels
         self._label_values = {}
-        for label_type in self._label_names:
-            for lbl_name in self._label_names[label_type]:
+        for label_type in Labels.LABEL_NAMES:
+            for lbl_name in Labels.LABEL_NAMES[label_type]:
                 if lbl_name in df_labels:
-                    self._label_values[label_type] = df_labels[lbl_name]
+                    self._label_values[label_type] = (lbl_name, df_labels[lbl_name])
                     break
 
     def get_label_only(self, label_type):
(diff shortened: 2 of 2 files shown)


atomic_reactor/constants.py
```diff
@@ -60,3 +60,14 @@
 )
 
 DEFAULT_DOWNLOAD_BLOCK_SIZE = 10 * 1024 * 1024 # 10Mb
+
+LABEL_ALIASES = {
+    "name" : ('name', 'Name'),
```
twaugh: I'd rather these were symbolic names in some way, to make it clear while reading code that this is not the real label.

atomic_reactor/util.py
```diff
@@ -542,7 +542,7 @@ def print_version_of_tools():
     ('release', 'Release'),
     ('architecture', 'Architecture'),
     ('vendor', 'Vendor'),
-    ('authoritative-source', 'Authoritative_Registry'),
+    ('authoritative-source-url', 'Authoritative_Registry'),
```
twaugh: Is this actually used in the code? If not maybe it should be removed altogether.

atomic_reactor/constants.py
rcerven: well this is symbolic name :) but how else I should name the name ;) I can change it so something like : alias_name?

atomic_reactor/util.py
rcerven: well that constant is used to for get_preferred_label, so it should have correct names

atomic_reactor/constants.py
twaugh: I mean that code using it ought not to be using a string literal. Eg
```
labels = Labels(df.labels)
name = labels.get_name()
```

or something similar.

atomic_reactor/util.py
twaugh: Is "authoritative-source" used in the code anywhere?

atomic_reactor/constants.py
rcerven: so you mean make class from it?

atomic_reactor/util.py
twaugh: Do we even need to describe this label here at all? Where is it used?

atomic_reactor/constants.py
twaugh: If this is being changed, let's just make sure code using it is very clear to read.

With your dict for instance, how should code wanting the label for the repository name look?

twaugh: (My preference would be to have a class that knows about which label names to use, yes - similarly to how it is done for eg BuildResponse).

rcerven: ok I am also now thinking more about class, as I can put there  get_new_label, get_old_labels etc methods

atomic_reactor/util.py
rcerven: well it is used for get_preffered_label method, which is used in different modules, explicitly is never called with authoritative-source-url, but in one case it is using some variable and I didn't look on that module yet "post_tag_by_labels.py", but I think this should be complete list of new & old labels

atomic_reactor/util.py
```diff
@@ -852,4 +852,145 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
     return dfparser
 
 
+class Labels(object):
```
twaugh: How about making some class variables such as `LABEL_NAME=1`, `LABEL_COMPONENT=2`, etc, and having a get_preferred_label_value() method which knows about iterating through the preferred list of label names and checking for each one in turn.

Then get_name() etc could use it. In other words make it DRY - don't repeat yourself.

The common case is that we just want to know the value for a particular field and don't care which label provides it.

However we also need a method which provides access to the full list of labels relevant to a particular field. (I don't remember where that is required.) So that could be get_preferred_label_names().

rcerven: I am not sure that I understand, but this class should provide labels not values, we need to know name of new label, name of old names, name of preferred from given dockerfile

rcerven: basically this class should remove need of any dictionaries for new/old labels which are used in whole atomic reactor

vrutkovs: +1 to Tim's suggestion - we should be able to quickly add more methods, using metaprogramming:
```
class Labels(object):
    self.methods = {
        "name": ["Name"],
        "version": ["Version"],
    }

def add_methods(cls, name, values):
    def inner_get(self):
        return values[0]

    inner_get.__name__ = "get_%s" % name
    setattr(cls, inner_get.__name__, inner_get)

    def inner_old(self):
        return values[1:]

    inner_old.__name__ = "get_%s_old" % name
    setattr(cls, inner_old.__name__, inner_old)

    def inner_all(self, labels):
        return values

    inner_all.__name__ = "get_%s_all" % name
    setattr(cls, inner_all.__name__, inner_all)

    def inner_pref(self):
        for lbl in values:
                if lbl in labels:
                    return lbl
        return values[0]
        
    inner_pref.__name__ = "get_%s_pref" % name
    setattr(cls, inner_pref.__name__, inner_pref)

for k, v in self.methods:
    self.add_methods(Labels, k, v)
```

twaugh:
I don't really like this way of doing it. I don't find it very readable to have modifications to __class__ if they can be avoided.

But also this class doesn't really work the way I would hope.

Being able to inspect a dictionary of old labels and new labels is of limited use. Only a couple of plugins care about that.

Most plugins *do* care about these sorts of questions:
 * 'what is the Bugzilla component for this Dockerfile?'
 * 'what is the repository name this should be pushed to?'
 * 'does this Dockerfile set a release number?'

To answer these questions requires knowing which labels are present in the Dockerfile, which labels to look for and in which order.

If we're going to the trouble of having a class, we can allow plugins to express these questions nicely, like this:

```
parser = df_parser(...)
labels = Labels(parser.labels)
try:
    release_label, release_value = labels.get_release()
except KeyError:
    # no release set
    release_label = labels.get_release_label()
    release_value = 1
```

The `Labels` class could have a constructor that takes a dictionary of {label:value}, and the class could define a method like this:
```
def get_label_and_value(self, label_type):
    """Return tuple of (label name, label value)

    Find the label name from the provided labels
    corresponding to the required label type.
    Return that label, and its value.
    Raise KeyError if there is no label for that type.
    """
    ...

def get_label_only(self, label_type):
    """Return label name in provided labels for given label type

    If there is no such label name, return the label name that
    should be added.
    """
    ...
```

The class could define label types as instance variables:
```
    LABEL_TYPE_NAME = 1
    LABEL_TYPE_BUILD_HOST = 2
    LABEL_TYPE_RELEASE = 3
```

The values aren't really important as long as they are distinct.

Finally, the class can define the methods that would be called:
```
def get_release(self):
    return self.get_label_and_value(self.LABEL_TYPE_RELEASE)
```
etc

Finally, for the benefit of the add_labels plugin only, it should be able to provide a dictionary of {old:new} label names.
```
def get_new_names_by_old(self):
    """Return dictionary, new label name indexed by old label name."""
```

atomic_reactor/util.py
```diff
@@ -852,4 +853,226 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
...
+        self._label_values = {}
+        for label_type in self._label_names:
+            for lbl_name in self._label_names[label_type]:
+                if lbl_name in df_labels:
+                    self._label_values[label_type] = df_labels[lbl_name]
+                    break
+
+    def get_label_only(self, label_type):
+        """returns newest label name"""
+        return self._label_names[label_type][0]
```
twaugh: This isn't implemented correctly. It should return the label name that should be used for *this Dockerfile*. If the Dockerfile uses the old label, it should return that.

atomic_reactor/util.py
```diff
@@ -852,4 +853,226 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
...
+
+    def get_label_component(self):
+        return self.get_label_only(self.LABEL_TYPE_COMPONENT)
+
+    def get_label_host(self):
+        return self.get_label_only(self.LABEL_TYPE_HOST)
+
+    def get_old_labels(self, label_type):
+        """returns tuple with all old label names"""
+        return self._label_names[label_type][1:]
```
twaugh: What's this for? Will that ever be used? I can't see any use for this.

atomic_reactor/util.py
```diff
@@ -852,4 +853,226 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
...
+
+    def get_old_labels_component(self):
+        return self.get_old_labels(self.LABEL_TYPE_COMPONENT)
+
+    def get_old_labels_host(self):
+        return self.get_old_labels(self.LABEL_TYPE_HOST)
+
+    def get_all_labels(self, label_type):
+        """returns tuple with all label names"""
+        return self._label_names[label_type]
```
twaugh: What's this for?

atomic_reactor/util.py
```diff
@@ -852,4 +853,226 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
...
+            for oldname in self._label_names[label_type][1:]:
+                newdict[oldname] = self._label_names[label_type][0]
+        return newdict
+
+    def get_value_only(self, label_type):
+        """
+        returns label value
+        Raises KeyError if label doesn't exist
+        """
+        return self._label_values[label_type]
```
twaugh: What's this for? We don't need this if we have a method that returns both the label name and its value.

atomic_reactor/util.py
rcerven: that is implemented in get_preferred_label

this returns only newest label, which was I think needed somewhere, like check if they are using newest labels, and print some warning

atomic_reactor/util.py
rcerven: warning about deprecated labels?

atomic_reactor/util.py
rcerven: this is only to get value, instead of tuple

atomic_reactor/util.py
rcerven: all available/valid names for label

atomic_reactor/util.py
twaugh: Why wouldn't a caller simply do this if they don't care what the label's name is?:
```
_, value = labels.get_version()
```

atomic_reactor/util.py
twaugh: Isn't that what the 'return the mapping' method was for?

atomic_reactor/util.py
twaugh: Do we do that? Do we want to do that?

atomic_reactor/util.py
twaugh: That's confusing, because `get_label_only()` is the name of the method I described.

Where is this needed?

rcerven: https://github.com/projectatomic/osbs-client/blob/master/osbs/api.py#L376
here is needed the newest label

atomic_reactor/util.py
vrutkovs: I don't think we do. The plugin used to add necessary labels without additional warning.

Anyway, it would be best to have all places in code replaced first so we could evaluate which functions are needed. This would save us some time. Thoughts?

atomic_reactor/util.py
vrutkovs: I don't see where this can be used

atomic_reactor/util.py
vrutkovs: +1 on Tim's comment - we should limit the amount of functions here

atomic_reactor/util.py
```diff
@@ -852,4 +853,226 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
     return dfparser
 
 
+class Labels(object):
+    def __init__(self, df_labels):
+        """
+        Create a new Labels object
+        providing access to actual newest labels as well as old ones
+        """
+        self.LABEL_TYPE_NAME = 1
```
lcarva: These should be class attributes.

atomic_reactor/util.py
```diff
@@ -852,4 +853,226 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
...
+        self.LABEL_TYPE_VERSION = 2
+        self.LABEL_TYPE_RELEASE = 3
+        self.LABEL_TYPE_ARCH = 4
+        self.LABEL_TYPE_VENDOR = 5
+        self.LABEL_TYPE_SOURCE = 6
+        self.LABEL_TYPE_COMPONENT = 7
+        self.LABEL_TYPE_HOST = 8
+
+        self._label_names = {
```
lcarva: This should also be a class attribute.

atomic_reactor/util.py
```diff
@@ -852,4 +853,226 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
...
+            self.LABEL_TYPE_ARCH : ('architecture', 'Architecture'),
+            self.LABEL_TYPE_VENDOR : ('vendor', 'Vendor'),
+            self.LABEL_TYPE_SOURCE : ('authoritative-source-url', 'Authoritative_Registry'),
+            self.LABEL_TYPE_COMPONENT : ('com.redhat.component', 'BZComponent'),
+            self.LABEL_TYPE_HOST : ('com.redhat.build-host', 'Build_Host')
+        }
+
+        self._df_labels = df_labels
+        self._label_values = {}
+        for label_type in self._label_names:
```
lcarva: Nitpick, but a more pythonic way of doing this loop is to use `items` to avoid indexing on the next `for-loop`:

```python
for label_type, lbl_names in self._label_names.items():
  for lbl_name in lbl_names:
```

atomic_reactor/util.py
```diff
@@ -852,4 +853,226 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
...
+            for lbl_name in self._label_names[label_type]:
+                if lbl_name in df_labels:
+                    self._label_values[label_type] = df_labels[lbl_name]
+                    break
+
+    def get_label_only(self, label_type):
+        """returns newest label name"""
+        return self._label_names[label_type][0]
+
+    def get_label_name(self):
```
lcarva: I think all of these helper methods will make this class hard to maintain in the long run. I don't think it adds much benefit to the usage.

```python
l = Labels(df_labels)
l.get_label_version()
# vs
l.get_label_only(l.LABEL_TYPE_VERSION)

File: atomic_reactor/util.py (hunk 8 above)
Comment (lcarva) [inline atomic_reactor/util.py]: The usage of `int`s imply a sequence, which is not the case here (which is correct). Consider using something like this instead to avoid confusion in the future:

```python
LABEL_TYPE_NAME = object()
LABEL_TYPE_VERSION = object()
...
```

Each call to `object()` returns a unique hashable object.

atomic_reactor/util.py
twaugh: Wouldn't this want to call the method that returns the most appropriate label to use for this Dockerfile for a particular type of value? If the label is missing, that method will return the new-style label, right?

atomic_reactor/util.py
twaugh: That's what `get_new_names_by_old()` is for isn't it? For the add_labels_in_dockerfile plugin to use because it needs to know the complete mapping.

atomic_reactor/util.py
rcerven: yeah in this case it would be same

atomic_reactor/util.py
twaugh: Oh, that's nice. :)

atomic_reactor/util.py
```diff
@@ -852,4 +853,111 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
...
+            self.LABEL_TYPE_HOST : ('com.redhat.build-host', 'Build_Host')
+        }
+
+        self._df_labels = df_labels
+        self._label_values = {}
+        for label_type in self._label_names:
+            for lbl_name in self._label_names[label_type]:
+                if lbl_name in df_labels:
+                    self._label_values[label_type] = df_labels[lbl_name]
+                    break
```
twaugh: Since we're looping here, why not also take a note of the label name? Then we won't have to recompute it in `get_label_only()` and `get_label_and_value()`. e.g.
```
self._labels_names_and_values[label_type] = (lbl_name, df_labels[lbl_name])
```

atomic_reactor/util.py
```diff
@@ -852,4 +853,62 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
...
+    def get_label_only(self, label_type):
+        """
+        returns the most preferred label name
+        if there isn't any correct name in the list
+        it will return newest label name
+        """
+        for lbl in Labels.LABEL_NAMES[label_type]:
+            if lbl in self._df_labels:
+                return lbl
+        return Labels.LABEL_NAMES[label_type][0]
```
twaugh: The constructor already calculated what this needs to return so there's no need to loop here.

atomic_reactor/util.py
```diff
@@ -852,4 +853,62 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
...
+            for oldname in label_names[1:]:
+                newdict[oldname] = Labels.LABEL_NAMES[label_type][0]
+        return newdict
+
+    def get_label_and_value(self, label_type):
+        """
+        Return tuple of (label name, label value)
+        Raises KeyError if label doesn't exist
+        """
+        return (self._label_values[label_type][0], self._label_values[label_type][1])
```
twaugh: Since we know it is a 2-tuple, isn't `(_label_values[label_type][0], self._label_values[label_type][1])` the same as just `_label_values[label_type]`?

PR #607, opened 2017-01-03
pre_koji: avoid crash if repo or tag is misconfigured

vrutkovs:
Don't try to add the repo if target returns an unconfigured tag or repo. I've 
seen this happening on internal testing Koji - and so did Adam on staging Fedora 
koji, although we're not sure why the tags are not configured properly

### DIFF
--- a/atomic_reactor/plugins/pre_koji.py
+++ b/atomic_reactor/plugins/pre_koji.py
@@ -45,7 +45,17 @@ class KojiPlugin(PreBuildPlugin):
             self.log.error("provided target '%s' doesn't exist", self.target)
             raise RuntimeError("Provided target '%s' doesn't exist!" % self.target)
         tag_info = self.xmlrpc.getTag(target_info['build_tag_name'])
+
+        if not tag_info or 'name' not in tag_info:
+            self.log.warning("No tag info was retrieved")
+            return
+
         repo_info = self.xmlrpc.getRepo(tag_info['id'])
+
+        if not repo_info or 'id' not in repo_info:
+            self.log.warning("No repo info was retrieved")
+            return
+
         # to use urljoin, we would have to append '/', so let's append everything
         baseurl = self.pathinfo.repo(repo_info['id'], tag_info['name']) + "/$basearch"
 
--- a/tests/plugins/test_koji.py
+++ b/tests/plugins/test_koji.py
@@ -41,11 +41,20 @@ class X(object):
 
 
 KOJI_TARGET = "target"
-GET_TARGET_RESPONSE = {"build_tag_name": "asd"}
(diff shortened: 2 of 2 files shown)


atomic_reactor/plugins/pre_koji.py
```diff
@@ -45,7 +45,17 @@ def run(self):
             self.log.error("provided target '%s' doesn't exist", self.target)
             raise RuntimeError("Provided target '%s' doesn't exist!" % self.target)
         tag_info = self.xmlrpc.getTag(target_info['build_tag_name'])
+
+        if not tag_info or 'name' not in tag_info.keys():
```
lcarva: No need to actually call `keys()`. The in operator will work as expected by checking the keys in a `dict`.

atomic_reactor/plugins/pre_koji.py
```diff
@@ -45,7 +45,17 @@ def run(self):
...
+            self.log.warning("No tag info was retrieved")
+            return
+
         repo_info = self.xmlrpc.getRepo(tag_info['id'])
+
+        if not repo_info or 'id' not in repo_info.keys():
```
lcarva: Is there a case where a `getRepo` returns a dict with some keys but without the `'id'` key? I'd expect it to return either None or an empty dict, which will both be checked by the first part conditional.

vrutkovs: I'm not sure, but that's unlikely - the [source code](https://github.com/koji-project/koji/blob/master/koji/tasks.py#L350) shows that its either `None` or a correct dict

atomic_reactor/plugins/pre_koji.py
vrutkovs: Ah, good idea. Updated the commit

atomic_reactor/plugins/pre_koji.py
```diff
@@ -45,7 +45,17 @@ def run(self):
             self.log.error("provided target '%s' doesn't exist", self.target)
             raise RuntimeError("Provided target '%s' doesn't exist!" % self.target)
         tag_info = self.xmlrpc.getTag(target_info['build_tag_name'])
+
+        if not tag_info or 'name' not in tag_info:
```
twaugh: Do we think 'name' is a required attribute of the response? If we're checking for it here, are we testing it in the unit test?

vrutkovs: [Koji source](https://github.com/koji-project/koji/blob/master/util/kojira#L59) suggests it is. I experimented a bit on stage server and it seems it either returns `None` or a correct dict - this is what the tests are checking too.

PR #610, opened 2017-01-05
Render client version if build json has information about it

vrutkovs:
See https://github.com/projectatomic/osbs-client/issues/477

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -176,7 +176,7 @@ class DockerRegistry(Registry):
         """
         super(DockerRegistry, self).__init__(uri, insecure=insecure)
         self.digests = {}  # maps tags (str) to their digest, if available
-        self.config = None  # stores image config from the registry, 
+        self.config = None  # stores image config from the registry,
         # media type of the config is application/vnd.docker.container.image.v1+json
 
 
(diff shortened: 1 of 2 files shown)


atomic_reactor/inner.py
```diff
@@ -297,6 +297,9 @@ def __init__(self, source, image, prebuild_plugins=None, prepublish_plugins=None
 
         self.openshift_build_selflink = openshift_build_selflink
 
+        if client_version:
+            logger.debug("build json was build by osbs-client %s", client_version)
```
twaugh: Actually this should probably say 'built by'

vrutkovs: Oh, right. Fixed

tests/test_inner.py
```diff
@@ -940,3 +941,40 @@ def restore_logger():
...
+        'exit_plugins': []
+    }
+    if has_version:
+        params['client_version'] = VERSION
+
+    workflow = DockerBuildWorkflow(MOCK_SOURCE, 'test-image', **params)
+    workflow.build_docker_image()
+
+    if has_version:
+        assert ("build json was built by osbs-client %s", VERSION) in fake_logger.debugs
```
lcarva: Maybe to catch both cases:

```python
assert (("build json was built by osbs-client %s", VERSION) in fake_logger.debugs) == has_version
```

vrutkovs: Ah, that's a good idea. Fixed

PR #611, opened 2017-01-05
rpm_qa: remove dangling volumes after the container is removed

### LINKED ISSUE #583: rpmqa plugin should remove volumes
When creating a container to run rpm in, docker will automatically create any volumes declared by the image.

After removing the container the plugin should attempt to remove any volumes that had been in use by the container.

Details:
The DockerTasker class should have a new method added, `remove_volume()`. Before removing the container, the post_rpmqa plugin should inspect the container to find which volumes it was using (like `docker inspect $CONTAINER | jq '.[0].Mounts[] | select(.Source != "") | .Name'`). After removing the container, it should call `remove_volume()` for each volume that was used, ignoring `APIError` exceptions whose `.response.status_code` is `requests.codes.conflict`.

vrutkovs:
Fixes #583

### DIFF
--- a/atomic_reactor/core.py
+++ b/atomic_reactor/core.py
@@ -27,6 +27,7 @@ import shutil
 import logging
 import tempfile
 import json
+import requests
 
 import docker
 from docker.errors import APIError
@@ -620,3 +621,39 @@ class DockerTasker(LastLogger):
         :return: dict, json output of `docker version`
         """
         return self.d.version()
+
+    def get_volumes_for_container(self, container_id, skip_empty_source=True):
+        """
+        get a list of volumes mounter in a container
+
+        :param container_id: str
+        :param skip_empty_source: bool, don't list volumes which are not created on FS
+        :return: list, a list of volume names
+        """
+        logger.info("listing volumes for container '%s'", container_id)
+        inspect_output = self.d.inspect_container(container_id)
+        volumes = inspect_output['Mounts'] or []
+        volume_names = [x['Name'] for x in volumes]
+
+        if skip_empty_source:
+            # Don't show volumes which are not on the filesystem
+            volume_names = [x['Name'] for x in volumes if x['Source'] != ""]
+
+        logger.debug("volumes = %s", volume_names)
+        return volume_names
+
+    def remove_volume(self, volume_name):
+        """
+        remove a volume by its name
+
+        :param volume_name: str
+        :return None
+        """
+        logger.info("removing volume '%s'", volume_name)
+        try:
+            self.d.remove_volume(volume_name)
+        except APIError as ex:
+            if ex.response.status_code == requests.codes.CONFLICT:
+                logger.debug("ignoring a conflict when removing volume %s", volume_name)
+            else:
+                raise ex
--- a/atomic_reactor/plugins/post_rpmqa.py
+++ b/atomic_reactor/plugins/post_rpmqa.py
@@ -59,10 +59,18 @@ class PostBuildRPMqaPlugin(PostBuildPlugin):
             self.log.debug("ignore rpms 'gpg-pubkey'")
             plugin_output = [x for x in plugin_output if not x.startswith("gpg-pubkey" + self.sep)]
 
+        volumes = self.tasker.get_volumes_for_container(container_id)
+
         try:
             self.tasker.remove_container(container_id)
         except APIError:
             self.log.warning("error removing container (ignored):",
                              exc_info=True)
+
+        for volume_name in volumes:
+            try:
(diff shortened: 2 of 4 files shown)


atomic_reactor/core.py
```diff
@@ -620,3 +621,32 @@ def get_version(self):
...
+        get a list of volumes mounter in a container
+
+        :param container_id: str
+        :return: list, a list of volume ids
+        """
+        logger.info("listing volumes for container '%s'", container_id)
+        inspect_output = self.d.inspect_container(container_id)
+        volumes = inspect_output['Mounts'] or []
+
+        volume_names = [x['Name'] for x in volumes if x['Source'] != ""]
```
lcarva: Perhaps some comment explaining what `Source` is and why ignore if empty?

atomic_reactor/core.py
```diff
@@ -620,3 +621,32 @@ def get_version(self):
...
+        logger.debug("volumes = '%s'", volume_names)
```
lcarva: Nitpick, but I think wrapping the list in single quotes is cumbersome, consider just removing it.

atomic_reactor/core.py
```diff
@@ -620,3 +621,32 @@ def get_version(self):
...
+        remove a volume by its id
+
+        :param container_id: str
+        :return None
+        """
+        logger.info("removing volume '%s'", volume_id)
+        try:
+            self.d.remove_volume(volume_id)
+        except APIError as ex:
+            if ex.response.status_code == requests.codes.CONFLICT:
```
lcarva: This will quietly swallow the exception if it's not due to `CONFLICT`, do we want to re-raise the exception? or maybe display a generic debug message?

atomic_reactor/core.py
```diff
@@ -620,3 +621,32 @@ def get_version(self):
...
+        inspect_output = self.d.inspect_container(container_id)
+        volumes = inspect_output['Mounts'] or []
+
+        volume_names = [x['Name'] for x in volumes if x['Source'] != ""]
+        logger.debug("volumes = '%s'", volume_names)
+        return volume_names
+
+    def remove_volume(self, volume_id):
+        """
+        remove a volume by its id
```
lcarva: Maybe a note indicating that ID is actually the volume Name?

atomic_reactor/core.py
vrutkovs: Ah, yes, other exceptions should be raised. Fixed

atomic_reactor/core.py
vrutkovs: Agreed, I've renamed all 'id' to 'name', as volumes are identified by names anyway

atomic_reactor/core.py
vrutkovs: Ah, right, I was thinking this would help to show an empty list, but its not really required. Fixed

atomic_reactor/core.py
vrutkovs: Added an option to `get_volumes_for_container` (enabled by default) and a comment with short explanation

atomic_reactor/plugins/post_rpmqa.py
```diff
@@ -6,6 +6,8 @@
 of the BSD license. See the LICENSE file for details.
 """
 
+import requests
```
twaugh: Not used?

tests/docker_mock.py
```diff
@@ -229,13 +268,13 @@ def mock_docker(build_should_fail=False,
     :param build_should_fail: True == build() log will contain error
     :param inspect_should_fail: True == inspect_image() will return None
     :param wait_should_fail: True == wait() will return 1 instead of 0
-    :param provided_image_repotags: images() will contain provided image
+    :param provided_image_repotags: images() will contain provolume_nameed image
```
twaugh: Something went wrong here?

tests/test_inner.py
```diff
@@ -401,7 +401,7 @@ class Exit(ExitPlugin):
      True,  # is fatal
      True,  # logs error
     ),
-     
+
```
twaugh: Could even remove this line maybe.

atomic_reactor/plugins/post_rpmqa.py
vrutkovs: Fixed

tests/docker_mock.py
vrutkovs: Right, I shouldn't have replaced all `vid` instances to `volume_name` blindly. Fixed

tests/test_inner.py
vrutkovs: Done

PR #613, opened 2017-01-06
new plugin: add_help

vrutkovs:
This plugin converts a markdown file, stored in the same repo as Dockerfile, to
an 'atomic help'-compatible man page.

### DIFF
--- a/atomic_reactor/plugins/pre_add_help.py
+++ b/atomic_reactor/plugins/pre_add_help.py
@@ -1,0 +1,81 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+
+
+Convert a help markdown file a man page and store it to /help.1 in the image
+so that 'atomic help' could display it.
+This is accomplished by appending an ADD command to it.
+
+Example configuration:
+{
+    'name': 'add_help',
+    'args': {'help_file': 'help.md'}
+}
+"""
+
+import os
+from subprocess import check_output, CalledProcessError, STDOUT
+from atomic_reactor.plugin import PreBuildPlugin
+from atomic_reactor.util import df_parser
+
+
+class AddHelpPlugin(PreBuildPlugin):
+    key = "add_help"
+    man_filename = "help.1"
+
+    def __init__(self, tasker, workflow, help_file="help.md"):
+        """
+        constructor
+
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param help_file: filename of the markdown help file
+        """
+        # call parent constructor
+        super(AddHelpPlugin, self).__init__(tasker, workflow)
+        self.help_file = help_file
+
+    def run(self):
+        """
+        run the plugin
+        """
+
+        help_path = os.path.join(self.workflow.builder.df_dir, self.help_file)
+
+        if not os.path.exists(help_path):
+            self.log.info("File %s not found", help_path)
+            return
+
+        man_path = os.path.join(self.workflow.builder.df_dir, self.man_filename)
+
+        go_md2man_cmd = ['go-md2man', '-in={}'.format(help_path), '-out={}'.format(man_path)]
+
+        try:
+            check_output(go_md2man_cmd, stderr=STDOUT)
+        except CalledProcessError as e:
+            if e.returncode == 127:
+                raise RuntimeError(
+                    "Help file is available, but go-md2man is not present in a buildroot")
+            raise RuntimeError("Error running %s: %r, exit code: %s, output: '%s'" % (
+                e.cmd, e, e.returncode, e.output))
+
+        if not os.path.exists(man_path):
+            raise RuntimeError("go-md2man run complete, but man file is not found")
+
+        # Include the help file in the docker file
+        dockerfile = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        lines = dockerfile.lines
+
+        content = 'ADD {0} /{0}'.format(self.man_filename)
+        # put it before last instruction
+        lines.insert(-1, content + '\n')
+
+        dockerfile.lines = lines
+
+        self.log.info("added %s", man_path)
+
+        return content
--- a/tests/plugins/test_add_help.py
+++ b/tests/plugins/test_add_help.py
@@ -1,0 +1,191 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import unicode_literals
+
+import os
+import subprocess
+import pytest
+
+from atomic_reactor.inner import DockerBuildWorkflow
+from atomic_reactor.plugin import PreBuildPluginsRunner
+from atomic_reactor.plugins.pre_add_help import AddHelpPlugin
+from atomic_reactor.util import ImageName, df_parser
+from tests.constants import MOCK_SOURCE
+from tests.fixtures import docker_tasker
+
+import atomic_reactor
+from tests.test_inner import FakeLogger
+from textwrap import dedent
+from flexmock import flexmock
+
+
+class Y(object):
+    pass
+
+
+class X(object):
+    image_id = "xxx"
+    source = Y()
+    source.dockerfile_path = None
+    source.path = None
+    base_image = ImageName(repo="qwe", tag="asd")
+
+
+def generate_a_file(destpath, contents):
+    with open(destpath, 'w') as f:
+        f.write(dedent(contents))
+
+
+@pytest.mark.parametrize('filename', ['help.md', 'other_file.md'])
+def test_add_help_plugin(tmpdir, docker_tasker, filename):
+    df_content = dedent("""
+        FROM fedora
+        RUN yum install -y python-django
+        CMD blabla""")
+    df = df_parser(str(tmpdir))
+    df.content = df_content
+
+    workflow = DockerBuildWorkflow(MOCK_SOURCE, 'test-image')
+    workflow.builder = X
+    workflow.builder.df_path = df.dockerfile_path
+    workflow.builder.df_dir = str(tmpdir)
+
+    help_markdown_path = os.path.join(workflow.builder.df_dir, filename)
+    generate_a_file(help_markdown_path, "foo")
+    help_man_path = os.path.join(workflow.builder.df_dir, AddHelpPlugin.man_filename)
+    generate_a_file(help_man_path, "bar")
+
+    cmd = ['go-md2man', '-in={}'.format(help_markdown_path), '-out={}'.format(help_man_path)]
+
+    def check_cmd(received_cmd, stderr):
+        assert received_cmd == cmd
+        assert stderr == subprocess.STDOUT
+
+    (flexmock(subprocess)
+         .should_receive("check_output")
+         .replace_with(check_cmd))
+
+    runner = PreBuildPluginsRunner(
+        docker_tasker,
+        workflow,
+        [{
+            'name': AddHelpPlugin.key,
+            'args': {'help_file': filename}
+        }]
+    )
+    runner.run()
+
+    assert  df.content == dedent("""
+        FROM fedora
+        RUN yum install -y python-django
+        ADD %s /%s
+        CMD blabla""" % (AddHelpPlugin.man_filename, AddHelpPlugin.man_filename))
+
+
+@pytest.mark.parametrize('filename', ['help.md', 'other_file.md'])
+def test_add_help_no_help_file(request, tmpdir, docker_tasker, filename):
+    df_content = "FROM fedora"
+    df = df_parser(str(tmpdir))
+    df.content = df_content
+
+    workflow = DockerBuildWorkflow(MOCK_SOURCE, 'test-image')
+    workflow.builder = X
+    workflow.builder.df_path = df.dockerfile_path
+    workflow.builder.df_dir = str(tmpdir)
+
+    runner = PreBuildPluginsRunner(
+        docker_tasker,
+        workflow,
+        [{
+            'name': AddHelpPlugin.key,
+            'args': {'help_file': filename}
+        }]
+    )
+    # Runner should not crash if no help.md found
+    runner.run()
+
+
+@pytest.mark.parametrize('filename', ['help.md', 'other_file.md'])
+@pytest.mark.parametrize('go_md2man_result', ['binary_missing', 'result_missing', 'fail', 'pass'])
+def test_add_help_md2man_error(request, tmpdir, docker_tasker, filename, go_md2man_result):
+    df_content = "FROM fedora"
+    df = df_parser(str(tmpdir))
+    df.content = df_content
+
+    workflow = DockerBuildWorkflow(MOCK_SOURCE, 'test-image')
+    workflow.builder = X
+    workflow.builder.df_path = df.dockerfile_path
+    workflow.builder.df_dir = str(tmpdir)
+
+    help_markdown_path = os.path.join(workflow.builder.df_dir, filename)
+    generate_a_file(help_markdown_path, "foo")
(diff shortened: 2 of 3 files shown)


atomic_reactor/plugins/pre_add_help.py
```diff
@@ -0,0 +1,86 @@
...
+            raise RuntimeError(
+                "Help file is available, but go-md2man is not present in a buildroot")
+        except CalledProcessError as ex:
+            # Ignore exit code 2, because this is what go-md2man does
+            if ex.returncode != 2:
+                raise RuntimeError(
+                    "There was an error with go-md2man, help returned %s" % ex.returncode)
+
+        man_path = os.path.join(self.workflow.builder.df_dir, self.man_filename)
+        cmd = "go-md2man -in=%s -out=%s" % (help_path, man_path)
```
twaugh: Can we make this a fixed string with no formatting, e.g. '... -in="$HELP_MD" -out="$HELP_1"', and set those environment variables?

tests/plugins/test_add_help.py
```diff
@@ -0,0 +1,181 @@
+"""
+Copyright (c) 2015 Red Hat, Inc
```
twaugh: 2017

.travis.yml
```diff
@@ -3,6 +3,10 @@ python:
   - "2.7"
   - "3.4"
   - "3.5"
+addons:
+  apt:
+    packages:
+    - go-md2man
```
twaugh: Is this used?

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -0,0 +1,86 @@
+"""
+Copyright (c) 2015 Red Hat, Inc
```
vrutkovs: Ah, it changes so fast. Fixed

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -0,0 +1,86 @@
...
+        try:
+            check_output(["go-md2man", "-h"], stderr=STDOUT)
+        except OSError:
+            raise RuntimeError(
+                "Help file is available, but go-md2man is not present in a buildroot")
+        except CalledProcessError as ex:
+            # Ignore exit code 2, because this is what go-md2man does
+            if ex.returncode != 2:
+                raise RuntimeError(
+                    "There was an error with go-md2man, help returned %s" % ex.returncode)
```
vrutkovs: Running it without params gives an error, so `-h` seems to be the simplest test

I'm not sure if we can rely on `which go-md2man` in this case (as `which` might not be installed).

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -0,0 +1,86 @@
...
+        cmd = "go-md2man -in=%s -out=%s" % (help_path, man_path)
+        try:
+            output = check_output(cmd, stderr=STDOUT, shell=True)
+        except CalledProcessError:
+            raise RuntimeError("Error running '%s':\n%s" % (cmd, output))
+
+        if not os.path.exists(man_path):
+            raise RuntimeError("go-md2man run complete, but man file is not found")
+
+        # Inlude the help file in the docker file
```
vrutkovs: Fixed

atomic_reactor/plugins/pre_add_help.py
twaugh: But we run it anyway with `-in ... -out ...`. Can't we use the result of that?

.travis.yml
vrutkovs: I tried to install it for travis tests to pass, but its not available in `precise` repos. Due to that `test_add_help_plugin` fails.

I'll play with faking a binary with a simple script instead

atomic_reactor/plugins/pre_add_help.py
vrutkovs: Yes, we could, but I don't really see the benefit in this. Do we want to set the paths to those in the buildroot via env vars?

twaugh: Security-aware programs should use fixed strings when passing them to the shell, so they know exactly how quoting will work.

atomic_reactor/plugins/pre_add_help.py
vrutkovs: Ah, yes, its probably a good idea, as we can wrap this call `flexmock` and avoid having the binary actually installed.

tests/plugins/test_add_help.py
```diff
@@ -0,0 +1,181 @@
...
+
+    request.addfinalizer(restore_logger)
+    atomic_reactor.plugin.logger = fake_logger
+
+    runner = PreBuildPluginsRunner(
+        docker_tasker,
+        workflow,
+        [{
+            'name': AddHelpPlugin.key,
+            'args': {'help_file': 'help.md'}
```
twaugh: Can we parametrize this to also test that it works the same without having to specify a `help_file` location?

vrutkovs: Sure, fixed

atomic_reactor/plugins/pre_add_help.py
vrutkovs: Agreed, implemented this

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -0,0 +1,81 @@
...
+import os
+from subprocess import check_output, CalledProcessError, STDOUT
+from atomic_reactor.plugin import PreBuildPlugin
+from atomic_reactor.util import df_parser
+
+
+class AddHelpPlugin(PreBuildPlugin):
+    key = "add_help"
+    man_filename = "help.1"
+    go_md2man_cmd = "go-md2man -in=$HELP_MD -out=$HELP_1"
```
twaugh: Can we explicitly double-quote the environment variables to make it safe?
`'go-md2man -in="$HELP_MD" -out="$HELP_1"'`

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -0,0 +1,81 @@
...
+        man_path = os.path.join(self.workflow.builder.df_dir, self.man_filename)
+        cmd_env = os.environ.copy()
+        cmd_env['HELP_MD'] = help_path
+        cmd_env['HELP_1'] = man_path
+        try:
+            check_output(self.go_md2man_cmd, stderr=STDOUT, shell=True, env=cmd_env)
+        except CalledProcessError as e:
+            if e.returncode == 127:
+                raise RuntimeError(
+                    u"Help file is available, but go-md2man is not present in a buildroot")
```
twaugh: Do we need to explicitly make this a unicode object with `u""`?

atomic_reactor/plugins/pre_add_help.py
vrutkovs: Agreed, fixed

atomic_reactor/plugins/pre_add_help.py
vrutkovs: Ah, it was a part of the hack to make py3 tests pass, its not needed now. Fixed

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -0,0 +1,81 @@
...
+import os
+from subprocess import check_output, CalledProcessError, STDOUT
+from atomic_reactor.plugin import PreBuildPlugin
+from atomic_reactor.util import df_parser
+
+
+class AddHelpPlugin(PreBuildPlugin):
+    key = "add_help"
+    man_filename = "help.1"
+    go_md2man_cmd = 'go-md2man -in=$"HELP_MD" -out=$"HELP_1"'
```
twaugh: No, `$"HELP_MD"` asks bash to provide the translation for the text "HELP_MD". We want to make the HELP_MD environment variable not create extra words, so we want `"$HELP_MD"`.

vrutkovs: Ouch, right. Fixed

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -0,0 +1,81 @@
...
+            raise RuntimeError("Error running %s: %s" % (e.cmd, repr(e)))
+
+        if not os.path.exists(man_path):
+            raise RuntimeError("go-md2man run complete, but man file is not found")
+
+        # Include the help file in the docker file
+        dockerfile = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        lines = dockerfile.lines
+
+        content = 'ADD {0} /{0}'.format(self.man_filename, self.man_filename)
```
twaugh: Do we have an extra parameter than we need here?

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -0,0 +1,81 @@
...
+        cmd_env = os.environ.copy()
+        cmd_env['HELP_MD'] = help_path
+        cmd_env['HELP_1'] = man_path
+        try:
+            check_output(self.go_md2man_cmd, stderr=STDOUT, shell=True, env=cmd_env)
+        except CalledProcessError as e:
+            if e.returncode == 127:
+                raise RuntimeError(
+                    "Help file is available, but go-md2man is not present in a buildroot")
+            raise RuntimeError("Error running %s: %s" % (e.cmd, repr(e)))
```
twaugh: Could use `"... %s: %r" % (e.cmd, e)` to avoid explicitly calling repr().

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -0,0 +1,81 @@
...
+
+        if not os.path.exists(man_path):
+            raise RuntimeError("go-md2man run complete, but man file is not found")
```
twaugh: Test coverage doesn't reach this line -- do we care?

vrutkovs: I think this situation is possible so lets keep this error

atomic_reactor/plugins/pre_add_help.py
vrutkovs: Done

atomic_reactor/plugins/pre_add_help.py
vrutkovs: Fixed

vrutkovs:
`docs/plugins.md` updated

twaugh:
LGTM

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -0,0 +1,81 @@
...
+        help_path = os.path.join(self.workflow.builder.df_dir, self.help_file)
+
+        if not os.path.exists(help_path):
+            self.log.info("File %s not found", help_path)
+            return
+
+        man_path = os.path.join(self.workflow.builder.df_dir, self.man_filename)
+        cmd_env = os.environ.copy()
+        cmd_env['HELP_MD'] = help_path
+        cmd_env['HELP_1'] = man_path
```
lcarva: Why use environment variables here? Couldn't you just string format `go_md2man_cmd`?

twaugh: To prevent security issues getting introduced at some future date.

lcarva: You mean this is less secure?

```python
go_md2man_cmd_template = 'go-md2man -in="{{HELP_MD}}" -out="{{HELP_1}}"'

go_md2man_cmd = go_md2man_cmd_template.format(HELP_MD=help_path, HELP_1=man_path)
```

If you mean because someone may craft a special value for `help_path` which may cause multiple commands to be called, I think a better solution is to properly sanitize the string.

Just seems odd to rely on bash for this, and not python.

vrutkovs: This is a security measure, as filenames might be unsafe here.

See [discussion](https://github.com/projectatomic/atomic-reactor/pull/613#discussion_r94952038)

twaugh: Not really odd. The POSIX shell behaviour for variable quoting is pretty well-known, and any string-sanitizing scheme is not.

When passing commands to the shell, IMHO they should always be fixed strings.

I've [ranted about this before](http://cyberelk.net/tim/articles/cmdline/index.html).

atomic_reactor/plugins/pre_add_help.py
lcarva: This wouldn't be a problem if `shell=False` was to be used. But in any case, `shlex.quote()`/`pipes.quote()` can be used to properly sanitize the filename.

https://docs.python.org/2/library/subprocess.html#frequently-used-arguments

twaugh: Either way is fine, and yes, `shell=False` is even more preferable.

lcarva: If we don't have a reason to use `shell=True`, please let's not use it and simply build a list of each param.

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -0,0 +1,81 @@
...
+        try:
+            check_output(self.go_md2man_cmd, stderr=STDOUT, shell=True, env=cmd_env)
```
lcarva: Why not use the preferred form `shell=False` ?

vrutkovs: I'm not sure if it would work with `env`. I'll test this option as well

lcarva: You may not need to use env, as each list item is properly handled as a single param, effectively removing the shell injection vulnerability seen with `shell=True`.

vrutkovs: Didn't work, `OSError: [Errno 2] No such file or directory` was thrown (as `subprocess` treats the cmdline as binary name and can't find it)

twaugh: Is there a reason we wouldn't want to just specify the path /usr/bin/go-md2man?

vrutkovs: In other parts this was not used (e.g. 'git clone' or 'rhpkg sources'). I'll play a bit more with this one

twaugh: Maybe supply a `PATH` in the environment? (e.g. from our own environment)

vrutkovs: `PATH` var are correct (`env` is a copy of existing env via `os.environ.copy()`), but shell substitution doesn't work without `shell=True` (see https://github.com/vrutkovs/atomic-reactor/commit/a38db1086f96f6932ca4ab93ca3c44eba6275a48):
`
RuntimeError: Error running ['go-md2man', '-in', '"$HELP_MD"', '-out', '"$HELP_1"']: CalledProcessError(), exit code: 1, output: 'open "$HELP_MD": no such file or directory
'`

I think its fine to use `shell=True` (btw, other plugins are using it too)

vrutkovs: Fixed

lcarva: Think you need to remove the inner double quotes. Because it's a list, python itself will ensure that each list item is properly quoted to avoid shell injection. And because of that, environment variable approach is also not needed.

This is just a best practices approach. The current implementation, although not as simple, seems fine to me. I'll approve the PR and leave it up to you to decide if you want to further investigate using `shell=False`.

tests/plugins/test_add_help.py
```diff
@@ -0,0 +1,191 @@
...
+    workflow.builder = X
+    workflow.builder.df_path = df.dockerfile_path
+    workflow.builder.df_dir = str(tmpdir)
+
+    help_markdown_path = os.path.join(workflow.builder.df_dir, filename)
+    generate_a_file(help_markdown_path, "foo")
+    help_man_path = os.path.join(workflow.builder.df_dir, AddHelpPlugin.man_filename)
+    generate_a_file(help_man_path, "bar")
+
+    cmd = [u'go-md2man', u'-in={}'.format(help_markdown_path), u'-out={}'.format(help_man_path)]
```
twaugh: With `unicode_literals` imported, do we need to specify `u` here?

Similarly, in the plugin module itself, should we import `unicode_literals`?

vrutkovs: Fixed, forgot to remove this when I was trying to fix tests

PR #615, opened 2017-01-09
Cancel 'build_filesystem' task if build has been canceled

vrutkovs:


atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -277,8 +277,17 @@ def run(self):
...
 
-        task = TaskWatcher(self.session, task_id, self.poll_interval)
-        task.wait()
+        try:
+            task = TaskWatcher(self.session, task_id, self.poll_interval)
+            task.wait()
+        except BuildCanceledException as exc:
+            self.log.info("Build was canceled, canceling task %s", task_id)
+            try:
+                self.session.cancelTaskFull(task_id)
```
lcarva: `cancelTaskFull` requires admin permission. Isn't `cancelTask` sufficient?

vrutkovs: Ah, correct, `cancelTask` should be sufficient, I was confused by 'getTaskChildren' mocking in tests. Fixed

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -277,8 +277,17 @@ def run(self):
 
         task_id, filesystem_regex = self.build_filesystem(image_build_conf)
 
-        task = TaskWatcher(self.session, task_id, self.poll_interval)
-        task.wait()
+        try:
+            task = TaskWatcher(self.session, task_id, self.poll_interval)
+            task.wait()
+        except BuildCanceledException as exc:
```
twaugh: Is `exc` used?

tests/plugins/test_add_filesystem.py
```diff
@@ -293,12 +314,23 @@ def _mockGetTaskResult(task_id):
...
+    with caplog.atLevel(logging.INFO):
+        with pytest.raises(PluginFailedException) as exc:
+            results = runner.run()
+
     assert task_result in str(exc)
     # Also ensure getTaskResult exception message is wrapped properly
     assert 'image task,' in str(exc)
 
+    if build_cancel:
+        assert "Build was canceled, canceling task 1234567" in caplog.records()[-3].message
```
twaugh: This is asserting how many logs there will be afterwards too? Do we need to?

vrutkovs: I think its worth to check the order of the messages here, but probably we'd have to change the tests too often for this.

Rewrote this part to use message order only when its necessary - when cancel has thrown an (ignored) error

atomic_reactor/plugins/pre_add_filesystem.py
vrutkovs: Removed it, in fact it was masked in the `catch` below.

PR #616, opened 2017-01-11
disable_autorebuild_by_default

rcerven:
raise error if autorebuild aren't explicitly set

### DIFF
--- a/atomic_reactor/plugins/pre_stop_autorebuild_if_disabled.py
+++ b/atomic_reactor/plugins/pre_stop_autorebuild_if_disabled.py
@@ -42,7 +42,7 @@ class StopAutorebuildIfDisabledPlugin(PreBuildPlugin):
         source_path = self.workflow.source.get()
         config_path = os.path.join(source_path, self.config_file)
         cfp = configparser.SafeConfigParser()
-        result = True
+        result = False
 
         if os.path.exists(config_path):
             try:
@@ -52,13 +52,13 @@ class StopAutorebuildIfDisabledPlugin(PreBuildPlugin):
                               'enabled' if result else 'disabled',
                               self.config_file)
             except configparser.Error:
-                self.log.error('can\'t parse "%s", assuming autorebuild is enabled',
+                self.log.error('can\'t parse "%s", assuming autorebuild is disabled',
                                self.config_file)
             except ValueError as e:
                 self.log.error('can\'t parse [autorebuild].enabled as bool in "%s", assuming'
-                               'autorebuild is enabled (error: "%s")', self.config_file, e)
+                               'autorebuild is disabled (error: "%s")', self.config_file, e)
         else:
-            self.log.info('no "%s", assuming autorebuild is enabled', self.config_file)
+            self.log.info('no "%s", assuming autorebuild is disabled', self.config_file)
 
         return result
 
--- a/tests/plugins/test_stop_autorebuild_if_disabled.py
+++ b/tests/plugins/test_stop_autorebuild_if_disabled.py
@@ -117,18 +117,17 @@ class TestStopAutorebuildIfDisabledPlugin(object):
         flexmock(configparser.SafeConfigParser).should_receive('read').and_return(None)
         # flexmock(configparser.SafeConfigParser).should_receive('getboolean').\
         #     with_args('autorebuild', 'enabled').and_raise(configparser.Error)
-        # assert this doesn't raise
-        with mocked_configparser_getboolean(configparser.Error):
+        with pytest.raises(AutoRebuildCanceledException):
             self.runner.run()
         self.assert_message_logged(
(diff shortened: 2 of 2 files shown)


atomic_reactor/plugins/pre_stop_autorebuild_if_disabled.py
```diff
@@ -52,13 +52,16 @@ def _is_rebuild_enabled(self):
...
+
             except ValueError as e:
-                self.log.error('can\'t parse [autorebuild].enabled as bool in "%s", assuming'
-                               'autorebuild is enabled (error: "%s")', self.config_file, e)
+                self.log.error('can\'t parse [autorebuild].enabled as bool in "%s", failing'
+                               'plugin (error: "%s")', self.config_file, e)
+                raise AutoRebuildCanceledException(self.key, 'can\'t parse "%s"' % config_file)
         else:
-            self.log.info('no "%s", assuming autorebuild is enabled', self.config_file)
+            self.log.info('no "%s", failing plugin', self.config_file)
```
vrutkovs: Nitpick: this probably should be a `warn` instead

atomic_reactor/plugins/pre_stop_autorebuild_if_disabled.py
```diff
@@ -52,13 +52,16 @@ def _is_rebuild_enabled(self):
                               'enabled' if result else 'disabled',
                               self.config_file)
             except configparser.Error:
-                self.log.error('can\'t parse "%s", assuming autorebuild is enabled',
-                               self.config_file)
+                self.log.error('can\'t parse "%s", failing plugin', self.config_file)
+                raise AutoRebuildCanceledException(self.key, 'can\'t parse "%s"' % config_file)
```
lcarva: This method shouldn't be raising exceptions. It should return a bool to indicate whether or not rebuild is enabled.

atomic_reactor/plugins/pre_stop_autorebuild_if_disabled.py
```diff
@@ -52,13 +52,16 @@ def _is_rebuild_enabled(self):
                               'enabled' if result else 'disabled',
                               self.config_file)
             except configparser.Error:
-                self.log.error('can\'t parse "%s", assuming autorebuild is enabled',
-                               self.config_file)
+                self.log.error('can\'t parse "%s", failing plugin', self.config_file)
```
lcarva: These should be: ... "assuming autorebuild is disabled."

atomic_reactor/plugins/pre_stop_autorebuild_if_disabled.py
```diff
@@ -52,13 +52,16 @@ def _is_rebuild_enabled(self):
                               'enabled' if result else 'disabled',
                               self.config_file)
```
lcarva: If you initialize `result` to `False`, you only need to adjust the log messages and set it to `True` here.

tests/plugins/test_stop_autorebuild_if_disabled.py
```diff
@@ -118,17 +118,18 @@ def test_malformed_config(self, caplog):
         # flexmock(configparser.SafeConfigParser).should_receive('getboolean').\
         #     with_args('autorebuild', 'enabled').and_raise(configparser.Error)
         # assert this doesn't raise
```
twaugh: This comment is no longer correct. While here you could remove the commented-out code if that's no longer relevant.

tests/plugins/test_stop_autorebuild_if_disabled.py
```diff
@@ -118,17 +118,18 @@ def test_malformed_config(self, caplog):
...
         self.assert_message_logged(
-            'can\'t parse ".osbs-repo-config", assuming autorebuild is enabled',
+            'can\'t parse ".osbs-repo-config", assuming autorebuild is disabled',
             caplog)
 
     def test_no_config(self, caplog):
         if MOCK:
             mock_docker()
 
         # assert this doesn't raise
```
twaugh: Comment needs updating.

atomic_reactor/plugins/pre_stop_autorebuild_if_disabled.py
twaugh: Where is result set to `True`? More importantly, if it is never set to `True`, how come the unit testing doesn't catch that?

rcerven:
line 50:  result = cfp.getboolean('autorebuild', 'enabled')

atomic_reactor/plugins/pre_stop_autorebuild_if_disabled.py
twaugh: Ah, line 50, thanks.

lcarva:
@rcerven please squash and fix commit message

rcerven:
squashed and changed commit message

PR #617, opened 2017-01-13
Remove unused 'remove_former_image' parameter in squash

rcerven:
remove_former_image argument isn't used anymore at all

### DIFF
--- a/atomic_reactor/plugins/prepub_squash.py
+++ b/atomic_reactor/plugins/prepub_squash.py
@@ -36,7 +36,6 @@ class PrePublishSquashPlugin(PrePublishPlugin):
           "args": {
             "tag": "SQUASH_TAG",
             "from_layer": "FROM_LAYER",
-            "remove_former_image": false,
             "dont_load": false
           }
         }
@@ -45,8 +44,7 @@ class PrePublishSquashPlugin(PrePublishPlugin):
 
     The `tag` argument specifes the tag under which the new squashed image will
     be registered. The `from_layer` argument specifies from which layer we want
-    to squash. `remove_former_image` is an optional boolean argument which specifies
-    if former, unsquashed image should be removed.
+    to squash.
 
     Of course it's possible to override it at runtime, like this: `--substitute prepublish_plugins.squash.tag=image:squashed
       --substitute prepublish_plugins.squash.from_layer=asdasd2332`.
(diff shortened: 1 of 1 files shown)


atomic_reactor/plugins/prepub_squash.py
```diff
@@ -57,14 +55,13 @@ class PrePublishSquashPlugin(PrePublishPlugin):
     is_allowed_to_fail = False
 
     def __init__(self, tasker, workflow, tag=None, from_base=True, from_layer=None,
-                 remove_former_image=True, dont_load=False):
+                 dont_load=False):
```
twaugh: I don't think we can just remove it from the parameter list -- older osbs-clients will generate build JSON that will no longer work.

rcerven: so should I just leave it in args and in docstring write : not used ? something like that?

twaugh: Yes, I think so.

This is not actually set by default in osbs-client, but we should leave a grace period and mention it in the release notes when we release atomic-reactor next.

rcerven: ok

rcerven:
squashed

twaugh:
Maybe a better commit message would be "Mark squash plugin 'remove_former_image' parameter as not used".

atomic_reactor/plugins/prepub_squash.py
```diff
@@ -64,7 +62,7 @@ def __init__(self, tasker, workflow, tag=None, from_base=True, from_layer=None,
         :param from_base: bool, squash from base-image layer, on by default
         :param from_layer: layer from we will squash - if specified, takes precedence over from_base
         :param tag: str, new name of the image - by default use the former one
-        :param remove_former_image: bool, remove unsquashed image?
+        :param remove_former_image: not used
```
lcarva: Can this actually be removed from the constructor? `atomic-reactor` should ignore unknown args: https://github.com/projectatomic/atomic-reactor/blob/b14b5abd75e2b1cb67be6e39fafcf2b98e5a8bbe/atomic_reactor/plugin.py#L326

twaugh: Oh! I hadn't realised it did that... :(

rcerven: so you mean remove that docstring about that param?

twaugh: You can remove that parameter from the docstring and also from the constructor's signature... like you had before I asked you to change it (sorry).

rcerven: ahhh ok

twaugh:
The commit history could have been cleaned up before this got merged :(

lcarva:
I squashed the commits prior to merging it.

PR #618, opened 2017-01-16
koji_promote: store help markdown filename in the build metadata if available

vrutkovs:


atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -560,6 +561,17 @@ def get_build(self, metadata):
                 else:
                     extra['filesystem_koji_task_id'] = task_id
 
+        help_result = self.workflow.prebuild_results.get(AddHelpPlugin.key)
+        if help_result is not None:
+            try:
+                cgit_url = self.workflow.source.uri.replace('/rpms/', '/cgit/rpms/')
```
lcarva: This seems too specific to our deployment.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -560,6 +561,17 @@ def get_build(self, metadata):
...
+                help_md = self.workflow.prebuild_plugins_conf[AddHelpPlugin.key]['help_file']
+                commit = self.workflow.source.lg.commit_id
+                extra['help'] = "{cgit_url}/tree/{help_md}?id={commit}".format(
```
lcarva: This seems too specific to our deployment.

atomic_reactor/plugins/exit_koji_promote.py
lcarva: I think this can be solved by modifying this plugin, or `AddHelpPlugin??`, to take a new param that specifies the template to be used to construct the URL.

twaugh: I don't know if we even want a cgit url at all. That's not what's discussed in the story.

vrutkovs: This probably should be a part of `GitSource` class - it might try to detect the git frontend

lcarva: A template string should allow users to use whatever hosting solution they'd like. We could format it with what we know: `commit`, `help_file_name`, others?

twaugh: We already have "the git URL" -- I was thinking we'd use that. There is a list of required fields in the story description.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -157,7 +157,7 @@ def run(self):
 
         annotations = {
             "dockerfile": open(self.workflow.builder.df_path).read(),
-            "artefacts": self.get_pre_result(DistgitFetchArtefactsPlugin.key),
+            "artefacts": self.get_pre_result(DistgitFetchArtefactsPlugin.key) or [],
```
lcarva: How about remove this key altogether?

atomic_reactor/plugins/exit_koji_promote.py
vrutkovs: How about formatting it as `{{git_url}}#{{commit}};{{filename}}`? Do we know if other koji plugins are using any scheme?

Sorry for discussing this after implementing, I expected the task to detect a git frontend to be much simplier

twaugh: The requirement is that the metadata can be used to discover the dist-git URL, commit hash, and pathname to the Markdown source file.

It needn't be a single field.

Note that build.source already includes the dist-git URL and commit hash, so I think all that's needed is to include the path within the repository in the build.extra.image map.

vrutkovs: >Note that build.source already includes the dist-git URL and commit hash, so I think all that's needed is to include the path within the repository in the build.extra.image map.

Agreed, implemented

lcarva:
How about rewording the commit message and PR title?

twaugh:
Once https://github.com/projectatomic/atomic-reactor/pull/622 is merged this could update docs/koji.md.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -560,6 +561,13 @@ def get_build(self, metadata):
                 else:
                     extra['filesystem_koji_task_id'] = task_id
 
+        help_result = self.workflow.prebuild_results.get(AddHelpPlugin.key)
+        if help_result is not None:
+            try:
+                extra['help'] = self.workflow.prebuild_plugins_conf[AddHelpPlugin.key]['help_file']
+            except:
+                extra['help'] = DEFAULT_HELP_FILENAME
```
twaugh: Although it appears to be the correct place because we have existing keys there, I think technically it ought to live in extra['image']['help'].

vrutkovs: Fixed

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -560,6 +561,13 @@ def get_build(self, metadata):
                 else:
                     extra['filesystem_koji_task_id'] = task_id
 
+        help_result = self.workflow.prebuild_results.get(AddHelpPlugin.key)
+        if help_result is not None:
+            try:
+                extra['image']['help'] = self.workflow.prebuild_plugins_conf[AddHelpPlugin.key]['help_file']
+            except:
+                extra['image']['help'] = DEFAULT_HELP_FILENAME
```
twaugh: Could set this at line 539, like in https://github.com/projectatomic/atomic-reactor/pull/619?

vrutkovs: Yeah, it would be set there.

Do we want a key with an empty value if help file was not found? I guess not, right?

twaugh: Good question. Possible ways of handling this:

- empty string, so an empty string means there is no help to find
- use `None`, so that a null value means there is no help to find
- omit the key altogether, so the absence of the key means there is no help to find

vrutkovs: Updated the code to have `help` set to `None`

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -560,6 +561,13 @@ def get_build(self, metadata):
                 else:
                     extra['filesystem_koji_task_id'] = task_id
 
+        help_result = self.workflow.prebuild_results.get(AddHelpPlugin.key)
+        if help_result is not None:
+            try:
+                extra['image']['help'] = self.workflow.prebuild_plugins_conf[AddHelpPlugin.key]['help_file']
+            except:
```
twaugh: `except KeyError:`?

vrutkovs: Right, fixed

twaugh:
Should the commit message say "build metadata" instead of "build config"?

docs/koji.md
```diff
@@ -23,6 +23,7 @@ For atomic-reactor container image builds the `image` type is used, and so type-
 Data which is placed here includes:
 
 - `build.extra.image.autorebuild` (boolean): true if this build was triggered automatically; false otherwise
+- `build.extra.image.help` (boolean): filename if this build has a markdown help converted; None otherwise
```
twaugh: Says 'boolean' but should say 'string or null'. Also let's call it 'null' instead of 'None' as this is a JSON object.

vrutkovs: Updated

docs/koji.md
```diff
@@ -23,6 +23,7 @@ For atomic-reactor container image builds the `image` type is used, and so type-
 Data which is placed here includes:
 
 - `build.extra.image.autorebuild` (boolean): true if this build was triggered automatically; false otherwise
+- `build.extra.image.help` (string or null): filename if this build has a markdown help converted; null otherwise
```
twaugh: I feel like we need to make it clear that it is the path to the Markdown file in the git repository (rather than to the converted file in the built image).

vrutkovs: Does "filename of the markdown help file in the repository if this build has a markdown help converted to man page; null otherwise" sound good?

twaugh: Perfect.

vrutkovs: Fixed

docs/koji.md
```diff
@@ -61,6 +62,8 @@ Example type-specific content generator metadata in context:
     "release": "1",
     "extra": {
       "image": {
+        "autorebuild": "False",
+        "help": None,
```
twaugh: Ah, good catch! Yes, we need to update this. Probably should read like this though:
```
{
  "autorebuild": false,
  "help": null
},
```
(no comma after the final entry)

vrutkovs: Fixed

docs/koji.md
```diff
@@ -61,6 +62,8 @@ Example type-specific content generator metadata in context:
     "release": "1",
     "extra": {
       "image": {
+        "autorebuild": "False",
```
twaugh: `false`

vrutkovs: Ah, weird, I though I updated

PR #619, opened 2017-01-16
Users can tell if a brew/koji build was created via automated rebuilds rather than a direct build request

rcerven:


atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -607,6 +608,12 @@ def get_metadata(self):
         buildroot = self.get_buildroot(build_id=self.build_id)
         output_files = self.get_output(buildroot['id'])
 
+        for output in output_files:
+            if output.metadata['buildroot_id'] == buildroot['id'] and \
+               output.metadata['type'] == 'docker-image':
+                buildroot['extra']['docker']['autorebuild'] = is_rebuild(self.workflow)
```
twaugh: This should go in `get_output()` I think. Also, could you make the commit message more descriptive?

rcerven: should get_output() really modify buildroot, which is already generated?

twaugh: We don't want to change anything in `buildroot` at all. It's the `output` object we want to set `extra.docker.autorebuild` on.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -509,7 +509,8 @@ def add_log_type(output):
                     'parent_id': parent_id,
                     'repositories': repositories,
                     'tags': list(tags),
-                    'config': config
+                    'config': config,
+                    'autorebuild': is_rebuild(self.workflow)
```
lcarva: Should this be done in get_build instead? Just thinking if this is a property of the build or the archive.

rcerven: but maybe you want to know that each specific output was rebuild?

lcarva: A build will have multiple archives when multi-arch is in the picture. In this case, each archive is bundled together in a single build. The case of having a mix of archives in a single build that were created via auto rebuild and manually, is not valid.

twaugh: Good point, it should probably be a build property not an output property. What should the key be then?

twaugh: How about `build.extra.autorebuild`?

twaugh: Alternative is `build.extra.image.autorebuild`. I'm not entirely sure what we're free to change.

lcarva: @mikebonnet, @mikem23: any preference as to the key that should be used in content-generator build's metadata to indicate it was automatically rebuilt?

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -568,6 +569,7 @@ def get_build(self, metadata):
             'start_time': start_time,
             'end_time': int(time.time()),
             'extra': extra,
+            'autorebuild': is_rebuild(self.workflow)
```
twaugh: It definitely can't go here. At the very least it must be inside 'extra'. The question is whether it needs to be in `build.extra.autorebuild` (I think this is best) or `build.extra.image.autorebuild`.

twaugh:
Reading through the koji code to try to understand more of what these objects are for, I think we should probably be storing additional things inside the `image` map.

Background: the build's `extra` map should have a `typeinfo` entry, which itself is a map containing a further entry named after the type of content generator being used -- this can be a map containing information specific to this content generator type.

So in theory it would look like this:
```json
{
  "build": {
    "extra": {
      "typeinfo": {
        "image": {
          "autorebuild": true
        }
      }
    }
  }
}
```

However, some types of content generator were defined before the 'typeinfo' concept: `win`, `maven`, and `image` -- which is for OSBS builds. For these content generator types, their type-specific information should live inside the map named after them, effectively skipping the 'typeinfo' entry, e.g.:
```json
{
  "build": {
    "extra": {
      "image": {
        "autorebuild": true
      }
    }
  }
}
```

@mikebonnet, does that look correct to you? I know I've heard you talk about this before.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -560,14 +561,15 @@ def get_build(self, metadata):
                 else:
                     extra['filesystem_koji_task_id'] = task_id
 
+        extra['image']['autorebuild'] = is_rebuild(self.workflow)
```
twaugh: Set this at line 539 instead maybe, where `image` is first assigned?

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -508,7 +509,7 @@ def add_log_type(output):
                     'parent_id': parent_id,
                     'repositories': repositories,
                     'tags': list(tags),
-                    'config': config
+                    'config': config,
```
twaugh: Not strictly needed for this change (although I prefer to have a trailing comma after each entry anyway).

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -567,7 +568,7 @@ def get_build(self, metadata):
             'source': "{0}#{1}".format(source.uri, source.commit_id),
             'start_time': start_time,
             'end_time': int(time.time()),
-            'extra': extra,
+            'extra': extra
```
twaugh: Not needed for this change.

tests/plugins/test_koji_promote.py
```diff
@@ -961,13 +961,15 @@ def test_koji_promote_image_tags(self, tmpdir, os_env, additional_tags):
          None),
 
     ])
-    @pytest.mark.parametrize('has_config', [
-        True,
-        False,
+    @pytest.mark.parametrize(('has_config', 'is_autorebuild'), [
+        (True,
```
vrutkovs: Those testcases are identical

rcerven: mmm what do you mean? I can use same values for is_autorebuild from has_config, but this is more explicit

rcerven: and this way I am testing 2 unrelated scenarios

lcarva: The way `parametrize` is being used here is by providing two sets of values for the `has_config` and `is_autorebuild`. However, both sets have the same value: `(True, False)`. Meaning, in both cases, `has_config` is `True`, and `is_autorebuild` is `False`. Perhaps you meant this:

```python
@pytest.mark.parametrize('has_config', [True, False])
@pytest.mark.parametrize('is_autorebuild', [True, False])
def test_koji_promote_success(...
```

rcerven: as it works now is it provides, true, true & false, false now
as these 2 test cases are independent I don't need full set t,t & t,f & f,f & f,t

I could even use has_config for is_autorebuild, but I wanted to explicitly name it

lcarva: Each tuple in the list is a test case, and they are both the same, meaning it's the same test case twice. Please read the docs: http://doc.pytest.org/en/latest/parametrize.html

rcerven: ahhhh ok I see what you mean mine bad

PR #621, opened 2017-01-16
Make importing istags fail after number of retries

csomh:
When importing new ImageStreamTags in `post_import_image` plug-in,
the code hit an infinite loop, in case the import never
succeeded.

The above was solved by raising a `PluginFailedException` in case
the number of retries was reached.

Optional parameter `retry_attempts` was also added. Defaults to 3,
which was the previous hard-coded value for this function.

Signed-off-by: Hunor Csomortáni <csomh@redhat.com>

### DIFF
--- a/atomic_reactor/plugins/post_import_image.py
+++ b/atomic_reactor/plugins/post_import_image.py
@@ -14,7 +14,7 @@ from osbs.api import OSBS
 from osbs.conf import Configuration
 from osbs.exceptions import OsbsResponseException
 
-from atomic_reactor.plugin import PostBuildPlugin, PluginFailedException
+from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.util import get_build_json
 
 
@@ -29,7 +29,7 @@ class ImportImagePlugin(PostBuildPlugin):
 
     def __init__(self, tasker, workflow, imagestream, docker_image_repo,
                  url, build_json_dir, verify_ssl=True, use_auth=True,
-                 insecure_registry=None, retry_delay=30, retry_attempts=3):
+                 insecure_registry=None, retry_delay=30, import_attempts=3):
         """
         constructor
 
@@ -44,8 +44,8 @@ class ImportImagePlugin(PostBuildPlugin):
         :param insecure_registry: bool, whether the Docker registry uses
                plain HTTP
         :param retry_delay: int, number of seconds to delay before retrying
-        :param retry_attempts: int, number of times it will be retried to
-               import image
+        :param import_attempts: int, number of times it will be retried to
+               import image; at least 1.
         """
         # call parent constructor
         super(ImportImagePlugin, self).__init__(tasker, workflow)
@@ -57,7 +57,12 @@ class ImportImagePlugin(PostBuildPlugin):
         self.use_auth = use_auth
         self.insecure_registry = insecure_registry
         self.retry_delay = retry_delay
-        self.retry_attempts = retry_attempts
+
+        if import_attempts < 1:
+            msg = "import_attempts is %d, should be at least 1"
+            raise ValueError(msg % import_attempts)
+
+        self.import_attempts = import_attempts
 
     def run(self):
         metadata = get_build_json().get("metadata", {})
@@ -88,15 +93,14 @@ class ImportImagePlugin(PostBuildPlugin):
         else:
             self.log.info("Importing new tags for %s", self.imagestream)
 
-            for attempts in range(self.retry_attempts):
-                new_tags_imported = osbs.import_image(self.imagestream)
-                if new_tags_imported:
-                    break
+            attempts = 0
+            while not osbs.import_image(self.imagestream):
+                attempts += 1
+
+                if attempts >= self.import_attempts:
+                    msg = "Failed to import new tags for %s"
+                    raise RuntimeError(msg % self.imagestream)
 
                 self.log.info("no new tags, will retry after %d seconds (%d/%d)",
-                              self.retry_delay, attempts + 1, self.retry_attempts)
+                              self.retry_delay, attempts, self.import_attempts)
                 sleep(self.retry_delay)
-
-            if not new_tags_imported:
-                msg = "Failed to import new tags for %s"
-                raise PluginFailedException(msg % self.imagestream)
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/post_import_image.py
```diff
@@ -83,15 +86,17 @@ def run(self):
...
-                    self.log.info("no new tags, will retry after %d seconds",
-                                  self.retry_delay)
-                    sleep(self.retry_delay)
+                self.log.info("no new tags, will retry after %d seconds (%d/%d)",
+                              self.retry_delay, attempts + 1, self.retry_attempts)
+                sleep(self.retry_delay)
+
+            if not new_tags_imported:
+                msg = "Failed to import new tags for %s"
+                raise PluginFailedException(msg % self.imagestream)
```
lcarva: It would be better to raise a `ValueError`, or `RuntimeError`, the atomic-reactor internals will ensure the exception is properly wrapped in `PluginFailedException`.

atomic_reactor/plugins/post_import_image.py
```diff
@@ -55,6 +57,7 @@ def __init__(self, tasker, workflow, imagestream, docker_image_repo,
         self.use_auth = use_auth
         self.insecure_registry = insecure_registry
         self.retry_delay = retry_delay
+        self.retry_attempts = retry_attempts
```
lcarva: Raise `ValueError` if `retry_attempts < 1`?

atomic_reactor/plugins/post_import_image.py
```diff
@@ -83,15 +86,17 @@ def run(self):
...
-                if retry_attempts > 0:
-                    retry_attempts -= 1
-                    self.log.info("no new tags, will retry after %d seconds",
-                                  self.retry_delay)
-                    sleep(self.retry_delay)
+                self.log.info("no new tags, will retry after %d seconds (%d/%d)",
+                              self.retry_delay, attempts + 1, self.retry_attempts)
+                sleep(self.retry_delay)
+
+            if not new_tags_imported:
```
lcarva: Corner case, but if `retry_attempts` is 0, this will throw a `NameError` as`new_tags_imported` won't be defined.
You could use `else` as a way to detect this scenario, or define/initialize `new_tags_imported` prior to for-loop.

Even if `retry_attempts` is validated to be `> 0`, I still think referencing a variable that was defined in a for-loop block can be confusing in the future.

tests/plugins/test_import_image.py
```diff
@@ -188,6 +189,35 @@ def test_import_image_retry(monkeypatch):
...
+     .once()
+     .with_args(TEST_IMAGESTREAM))
+    (flexmock(OSBS)
+     .should_receive('create_image_stream')
+     .never())
+    (flexmock(OSBS)
+     .should_receive('import_image')
+     .times(3)
+     .with_args(TEST_IMAGESTREAM)
+     .and_return(False)
```
lcarva: Maybe a test with:
```python
.and_return(False)
.and_return(False)
.and_return(True)
```
to ensure retry is attempted and plugin works as expected?

atomic_reactor/plugins/post_import_image.py
csomh: True, added.

atomic_reactor/plugins/post_import_image.py
csomh: My bad, fixed.

atomic_reactor/plugins/post_import_image.py
csomh: Changed to raise `RuntimeError`

tests/plugins/test_import_image.py
csomh: This was already tested only with 2 attempts configured. I've parameterized the test for the retry aspects, and merged them together.

csomh:
Refactored the code and tests a little bit. Changes can be squashed with the 1st commit.
I was not able to quickly find a way to test the exception raised in `__init__()`. Hints or solutions (maintainers can edit this feature branch) are welcome.

atomic_reactor/plugins/post_import_image.py
```diff
@@ -56,6 +58,12 @@ def __init__(self, tasker, workflow, imagestream, docker_image_repo,
         self.insecure_registry = insecure_registry
         self.retry_delay = retry_delay
 
+        if import_attempts < 1:
```
vrutkovs: This code branch is not checked by the tests

lcarva: @csomh could add a test with retry < 1, and ensure that it raises an exception and that exception msg includes "ValueError" and 'should be at least 1'. Not great, but will get the job done.

atomic_reactor/plugins/post_import_image.py
```diff
@@ -83,15 +91,16 @@ def run(self):
...
-
-                if retry_attempts > 0:
-                    retry_attempts -= 1
-                    self.log.info("no new tags, will retry after %d seconds",
-                                  self.retry_delay)
-                    sleep(self.retry_delay)
+            self.log.info("Importing new tags for %s", self.imagestream)
+
+            attempts = 0
+            while not osbs.import_image(self.imagestream):
```
lcarva: More like a style comment, but I think a for-loop is probably cleaner here. Feel free to ignore this comment :)

tests/plugins/test_import_image.py
```diff
@@ -162,12 +163,17 @@ def test_import_image(namespace, monkeypatch):
     runner.run()
 
 
-def test_import_image_retry(monkeypatch):
+@pytest.mark.parametrize(('retry_delay', 'results'), [
+    (1, [False, True]),
```
lcarva: How about using a small number like: `0.01` ? Things like this really slow down the overall unit testing phase.

atomic_reactor/plugins/post_import_image.py
csomh: I would keep like this, mainly because with a for-loop I don't see a way to avoid a conditional block after the loop, to raise the error. The while-loop keeps the functionality (retry n times and fail if not succeeded) in a single block of code.

tests/plugins/test_import_image.py
csomh: Done.

atomic_reactor/plugins/post_import_image.py
csomh: Done.

PR #625, opened 2017-01-17
When specified 'required':False, plugins should not fail the build

### LINKED ISSUE #624: koji_promote ends with INFO - Not promoting failed build to koji
this relates to the change, where dockerfile_content was removed

in plugin.py PluginsRunner

self.on_plugin_failed() is called and sets plugin_failed to True,
even when plugin 
plugin 'dockerfile_content' requested but not available
even when in prod_inner.json the plugin has set required to False

so based on that even when inside container build is success, but this not required plugin sets plugin_failed
koji_promote won't promote anything

rcerven:
fixes #624 

koji_promote ends with INFO - Not promoting failed build to koji #624

this relates to the change, where dockerfile_content was removed

in plugin.py PluginsRunner

self.on_plugin_failed() is called and sets plugin_failed to True,
even when plugin
plugin 'dockerfile_content' requested but not available
even when in prod_inner.json the plugin has set required to False

so based on that even when inside container build is success, but this not required plugin sets plugin_failed
koji_promote won't promote anything

so what this does it that if plugin which doesn't exist, and isn't required, won't set plugins_failed in workflow, therefore koji_promote WILL continue to pushing things to koji

### DIFF
--- a/atomic_reactor/plugin.py
+++ b/atomic_reactor/plugin.py
@@ -195,9 +195,8 @@ class PluginsRunner(object):
             try:
                 plugin_class = self.plugin_classes[plugin_name]
             except KeyError:
-                self.on_plugin_failed()
-
                 if plugin_request.get('required', True):
+                    self.on_plugin_failed()
                     msg = ("no such plugin: '%s', did you set "
                            "the correct plugin type?") %  plugin_name
                     logger.error(msg)
--- a/tests/test_plugin.py
+++ b/tests/test_plugin.py
@@ -17,7 +17,8 @@ import pytest
 from atomic_reactor.inner import DockerBuildWorkflow
 from atomic_reactor.plugin import (BuildPluginsRunner, PreBuildPluginsRunner,
                                    PostBuildPluginsRunner, InputPluginsRunner,
-                                   PluginFailedException)
+                                   PluginFailedException, PrePublishPluginsRunner,
+                                   ExitPluginsRunner)
 from atomic_reactor.plugins.post_rpmqa import PostBuildRPMqaPlugin
 from atomic_reactor.plugins.pre_add_yum_repo_by_url import AddYumRepoByUrlPlugin
 from atomic_reactor.util import ImageName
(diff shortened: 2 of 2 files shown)


twaugh:
What does this do? How does it change the behaviour?

vrutkovs:
Whenever the plugin runner doesn't find a plugin it checks if its required or not. The problem is that it first marks the build as failed.

As a result `koji_promote` will exit with message `Won't promote a failed build` whenever it sees the build is 'failed' (although its perfectly fine)

atomic_reactor/plugin.py
```diff
@@ -192,19 +192,18 @@ def run(self, keep_going=False):
                 raise PluginFailedException(msg)
 
             plugin_conf = plugin_request.get("args", {})
+            run_fail = False
             try:
                 plugin_class = self.plugin_classes[plugin_name]
             except KeyError:
-                self.on_plugin_failed()
-
                 if plugin_request.get('required', True):
```
vrutkovs: Why not simply move `self.on_plugin_failed()` here?

twaugh:
Can we have a unit test that demonstrates this case?

atomic_reactor/plugin.py
rcerven: and what about : is_allowed_to_fail?

rcerven:
yeah we should, I'll make tomorrow

atomic_reactor/plugin.py
twaugh: If we get here, there's no plugin to inspect the `is_allowed_to_fail` property of. :)

twaugh:
Ah, I think I get it. Nice catch!

twaugh:
Error was introduced in https://github.com/projectatomic/atomic-reactor/pull/568.

atomic_reactor/plugin.py
rcerven: yeah I just realized that is_allowed_to_fail is used when plugin is already running

tests/test_plugin.py
```diff
@@ -63,6 +63,23 @@ def test_build_plugin_failure(docker_tasker):
         results = runner.run()
     assert workflow.build_process_failed is True
 
+@pytest.mark.parametrize('required', [
+    True,
+    False,
+])
+def test_not_required_prebuild_plugin_failure(docker_tasker, required):
```
twaugh: Do we need to also test postbuild, prepub, and exit plugin types?

tests/test_plugin.py
```diff
@@ -63,6 +63,23 @@ def test_build_plugin_failure(docker_tasker):
...
+    workflow = DockerBuildWorkflow(SOURCE, "test-image")
+    assert workflow.plugin_failed is False
+    runner = PreBuildPluginsRunner(docker_tasker, workflow,
```
vrutkovs: Lets extend this plugin to other plugin types, e.g. post-build?

tests/test_plugin.py
```diff
@@ -63,6 +63,23 @@ def test_build_plugin_failure(docker_tasker):
...
+                                   [{"name": "dockerfile_content",
```
twaugh: How about using a name that makes it clear the plugin doesn't exist, like `"name": "no_such_plugin"`?

tests/test_plugin.py
rcerven: well that fail setting method on_plugin_failed, is in the class BuildPluginsRunner
and post, pre, exit plugins are subclasses of this and don't override this method
so in the end, is always called just method from super class

tests/test_plugin.py
rcerven: yeah good idea :)

tests/test_plugin.py
rcerven: well that fail setting method on_plugin_failed, is in the class BuildPluginsRunner
and post, pre, exit plugins are subclasses of this and don't override this method
so in the end, is always called just method from super class

vrutkovs: They don't now, but this test should protect from all possible situations.

Probably this is worth doing in next PRs for this test

rcerven: well it is done ;)

PR #626, opened 2017-01-18
Avoid sleeping unnecessarily during tests

twaugh:
This timeout parameter is used (and set) by the cancel tests but not by all the other tests that use this class.


twaugh:
Before: 5m43s per branch
After: 1m24s per branch

PR #627, opened 2017-01-18
(ignore for now) Record errors for exit plugins that must succeed

twaugh:
Previously plugins such as koji_promote, i.e. exit plugins which must succeed, do not have their failures recorded in the workflow's `plugins_errors` dict. This means build errors due to failures in such plugins have no meaningful error message stored in the build annotations.

This PR fixes that.


lcarva:
@twaugh does "(ignore for now)" in PR title mean do not merge?

twaugh:
I was trying to dissuade anyone from reviewing it yet so as not to pull it into this sprint. :)

PR #630, opened 2017-01-19
koji_promote: trap TypeError when checking help plugin config

vrutkovs:
This also fixes `add_help` plugin's incorrect verification of `go-md2man` binary

### DIFF
--- a/tests/plugins/test_koji_promote.py
+++ b/tests/plugins/test_koji_promote.py
@@ -1186,10 +1186,19 @@ class TestKojiPromote(object):
                                             version='1.0',
                                             release='1',
                                             session=session)
-        if expect_result in ['empty_result', 'pass']:
-            workflow.prebuild_plugins_conf[AddHelpPlugin.key] = {
-                'help_file': 'foo.md',
-            }
+
+        if expect_result == 'pass':
+            workflow.prebuild_plugins_conf = [{
+                "name": AddHelpPlugin.key,
+                "args": {
+                  "help_file": 'foo.md',
+                }
+            }]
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_add_help.py
```diff
@@ -58,10 +58,10 @@ def run(self):
 
         try:
             check_output(go_md2man_cmd, stderr=STDOUT)
-        except CalledProcessError as e:
-            if e.returncode == 127:
-                raise RuntimeError(
+        except OSError:
```
lcarva: Seems conflicting with: https://github.com/projectatomic/atomic-reactor/pull/631

vrutkovs: Right, Tim has come up with a much better test for this. Fixed

vrutkovs:
The tests are now correct and are failing :) This PR will be rebased after #631 is merged

tests/plugins/test_koji_promote.py
```diff
@@ -1186,11 +1186,15 @@ def test_koji_promote_add_help(self, tmpdir, os_env, expect_result):
...
                                             release='1',
                                             session=session)
-        if expect_result in ['empty_result', 'pass']:
+
+        if expect_result == 'pass':
             workflow.prebuild_plugins_conf[AddHelpPlugin.key] = {
                 'help_file': 'foo.md',
             }
 
+        if expect_result == 'empty_config':
```
twaugh: Could be `elif` but not really important.

Issue #584, opened 2016-11-14
when running builds in atomic-reactor via OSBS, parent image ENV variables not parsed correctly


maxamillion:
When building a Layered Image that uses `ENV` vars that come from the parent image, these variables are not being properly expanded/evaluated.

Dockerfile
```
FROM registry.stg.fedoraproject.org/fedora:24
MAINTAINER "Adam Miller" <maxamillion@fedoraproject.org>

ENV NAME=cockpit VERSION=0.117 RELEASE=1.11 SUFFIX=fc24
LABEL BZComponent="$NAME" \
	Name="$FGC/$NAME" \
	Version="$VERSION" \
	Release="$RELEASE.$DISTTAG" \
	Architecture="x86_64"

RUN dnf install -y cockpit-ws curl

# And the stuff that starts the container
RUN mkdir -p /container && ln -s /host/proc/1 /container/target-namespace
ADD atomic-install /container/atomic-install
ADD atomic-uninstall /container/atomic-uninstall
ADD atomic-run /container/atomic-run
RUN chmod -v +x /container/atomic-install
RUN chmod -v +x /container/atomic-uninstall
RUN chmod -v +x /container/atomic-run

# Make the container think it's the host OS version
RUN rm -f /etc/os-release /usr/lib/os-release && ln -sv /host/etc/os-release /etc/os-release && ln -sv /host/usr/lib/os-release /usr/lib/os-release

LABEL INSTALL /usr/bin/docker run -ti --rm --privileged -v /:/host IMAGE /container/atomic-install
LABEL UNINSTALL /usr/bin/docker run -ti --rm --privileged -v /:/host IMAGE /container/atomic-uninstall
LABEL RUN /usr/bin/docker run -d --privileged --pid=host -v /:/host IMAGE /container/atomic-run --local-ssh

# Look ma, no EXPOSE

CMD ["/container/atomic-run"]
```

Output from `docker inspect` on the resulting image:
```json
[
    {
        "Id": "sha256:e5e3d99f279c79f705dae40ac55b879060951314f2b1daa0482ca3d0a4455faa",
        "RepoTags": [
            "candidate-registry.stg.fedoraproject.org/cockpit:latest"
        ],
        "RepoDigests": [
            "candidate-registry.stg.fedoraproject.org/cockpit@sha256:6358423bee68e643460a6acae2d0dd9ac220fa5acba41fc1943d1052c4dc62a1"
        ],
        "Parent": "",
[...]

PR #634, opened 2017-01-25
df_parser parent_env unit tests

rcerven:
unit tests for #584

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -16,6 +16,7 @@ import responses
 import six
 
 from tempfile import mkdtemp
+from textwrap import dedent
 from flexmock import flexmock
 
 try:
@@ -24,6 +25,7 @@ except ImportError:
     # Python 2.6
     from ordereddict import OrderedDict
 import docker
+from atomic_reactor.inner import DockerBuildWorkflow
 from atomic_reactor.util import (ImageName, wait_for_command, clone_git_repo,
                                  LazyGit, figure_out_dockerfile,
                                  render_yum_repo, process_substitutions,
@@ -33,7 +35,9 @@ from atomic_reactor.util import (ImageName, wait_for_command, clone_git_repo,
                                  get_manifest_digests, ManifestDigest,
                                  get_build_json, is_scratch_build, df_parser)
 from atomic_reactor import util
-from tests.constants import DOCKERFILE_GIT, INPUT_IMAGE, MOCK, DOCKERFILE_SHA1
+from tests.constants import DOCKERFILE_GIT, INPUT_IMAGE, MOCK, DOCKERFILE_SHA1, MOCK_SOURCE
+from atomic_reactor.constants import INSPECT_CONFIG
+
 from tests.util import requires_internet
 
 if MOCK:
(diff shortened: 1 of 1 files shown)


tests/test_util.py
```diff
@@ -456,3 +459,46 @@ def test_df_parser(tmpdir):
     assert len(df.labels) == 1
     assert df.labels.get('label') == 'foobar barfoo'
 
+def test_df_parser_parent_env_arg(tmpdir):
+    tmpdir_path = str(tmpdir.realpath())
+    p_env = {
+        "test_env": "first"
+    }
+    df_content = """\
+FROM fedora
```
twaugh: Could use textwrap's `dedent()` here (as in other places) to make it easier to read.

tests/test_util.py
```diff
@@ -456,3 +459,46 @@ def test_df_parser(tmpdir):
...
+ENV foo=bar
+LABEL label="foobar $test_env"
+"""
+   
+    with open(tmpdir_path + '/Dockerfile', 'w') as dockerfile:
+        dockerfile.write(df_content)
+        dockerfile.flush()
+
+    df = df_parser(tmpdir_path, parent_env=p_env)
+    assert df.parent_env.get('test_env') == 'first'
```
twaugh: I think `parent_env` isn't really meant to be a public part of the interface for that object.

tests/test_util.py
```diff
@@ -456,3 +459,46 @@ def test_df_parser(tmpdir):
...
+   
+    with open(tmpdir_path + '/Dockerfile', 'w') as dockerfile:
+        dockerfile.write(df_content)
+        dockerfile.flush()
+
+    workflow = DockerBuildWorkflow(MOCK_SOURCE, 'test-image')
+    flexmock(workflow, base_image_inspect=env_conf)
+
+    df = df_parser(tmpdir_path, workflow=workflow)
+    assert df.parent_env.get('test_env') == 'first'
```
twaugh: As above, not sure this line is necessary.

tests/test_util.py
```diff
@@ -456,3 +483,57 @@ def test_df_parser(tmpdir):
...
+
+    def restore_logger():
+        util.logger = existing_logger
+    
+    request.addfinalizer(restore_logger)
+    df = df_parser(tmpdir_path, workflow=workflow)
+
+    if isinstance(env_arg, list) and ('=' not in env_arg[0]):
+        expected_log_message = ("Unable to parse all of Parent Config ENV")
+        assert (expected_log_message, ) in fake_logger.debugs
```
twaugh: You can use `caplog` to do this sort of thing and it's generally easier. Just add 'caplog' to the parameter list for that method and the pytest caplog fixture will be passed in -- there are other examples around in this project.

You probably want something like:
```
with caplog.atLevel(logging.DEBUG):
   ...  # use df_parser()

assert msg in [x.message for x in caplog.records()]
```

tests/test_util.py
```diff
@@ -456,3 +460,49 @@ def test_df_parser(tmpdir):
...
+    p_env = {
+        "test_env": "first"
+    }
+    df_content = dedent("""\
+        FROM fedora
+        ENV foo=bar
+        LABEL label="foobar $test_env"
+        """)
+
+    with open(tmpdir_path + '/Dockerfile', 'w') as dockerfile:
```
vrutkovs: We don't really need that:
```
    df = df_parser(str(tmpdir))
    df.content = df_content
```
should be sufficient

tests/test_util.py
```diff
@@ -456,3 +460,49 @@ def test_df_parser(tmpdir):
...
+
+    df = df_parser(tmpdir_path, parent_env=p_env)
+    assert df.labels.get('label') == 'foobar first'
+
+@pytest.mark.parametrize('env_arg', [
+    {"test_env": "first"},
+    ['test_env=first'],
+    ['test_env_first'],
+])
+def test_df_parser_parent_env_wf(tmpdir, caplog, env_arg):
```
vrutkovs: Those testcases should probably be joined to test 'parent image envs + inspected config'

tests/test_util.py
```diff
@@ -456,3 +460,49 @@ def test_df_parser(tmpdir):
...
+    with open(tmpdir_path + '/Dockerfile', 'w') as dockerfile:
+        dockerfile.write(df_content)
+        dockerfile.flush()
+
+    df = df_parser(tmpdir_path, parent_env=p_env)
+    assert df.labels.get('label') == 'foobar first'
+
+@pytest.mark.parametrize('env_arg', [
+    {"test_env": "first"},
+    ['test_env=first'],
```
vrutkovs: Why would env vars lib contain lists?

Also, in `pre_distribution_scope` `base_image_inspect` contains a dict 'Labels' - is this test correct?

vrutkovs:
Code added in https://github.com/projectatomic/atomic-reactor/pull/586

tests/test_util.py
vrutkovs: See the actual inspect output: it has to be a list of dicts (not a dict like in a first option and not a list of strings like in the last option). 

Instead it should test that 
* an empty list is parsed correctly (no envs in the base image), 
* a list of strings separated by '='

PR #636, opened 2017-01-26
koji_promote: set correct value to 'help' key in metadata, depending on status of add_help plugin

vrutkovs:


twaugh:
Does this set help=None when the add_help plugin runs and finds no help file?

vrutkovs:
Right, I appended the missing commit. Now `koji_promote` would set "help" to None if the plugin is configured, but it returned None (which happens when exception occurred or help file is missing)

twaugh:
Just to be clear: if add_help gets an exception we should not set `help` in the metadata.

vrutkovs:
Added a commit which makes it result more explicit:
 - `None` if exception occurred
 - `NO_HELP_FILE_FOUND` if there is no help file in the repo
 - `HELP_GENERATED` if man page was generated
and reworked `koji_promote` plugin to set `help` key in metadata accordingly

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -84,4 +91,4 @@ def run(self):
 
         self.log.info("added %s", man_path)
 
-        return content
+        return self.HELP_GENERATED
```
twaugh: Could return the path to the help file it used?

vrutkovs: Yes, its possible although I wanted to avoid two things here:
 - return `None` if exception and an empty string if no help file was used
 - duplicating information from the config

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -562,14 +562,19 @@ def get_build(self, metadata):
...
-                extra['image']['help'] = DEFAULT_HELP_FILENAME
+            if help_result == AddHelpPlugin.HELP_GENERATED:
+                try:
+                    extra['image']['help'] = add_help_plugin_config[0]['args']['help_file']
+                except (KeyError, TypeError, IndexError):
+                    extra['image']['help'] = DEFAULT_HELP_FILENAME
+            elif help_result == AddHelpPlugin.NO_HELP_FILE_FOUND:
+                extra['image']['help'] = None
+            else:
+                self.log.error("Unknown result from add_help plugin: %s" % help_result)
```
twaugh: Nitpick: use separate arg instead of %.

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -45,13 +48,17 @@ def __init__(self, tasker, workflow, help_file=DEFAULT_HELP_FILENAME):
     def run(self):
         """
         run the plugin
+
+        The plugin returns None if exception occurred,
```
lcarva: What's the benefit of returning these 3 values? Anything inspecting the result of this plugin likely cares whether or not the help file was generated. IMO using `int`s like this will make it harder to debug plugin results.

If we do care about the error cause, I'd much rather this plugin returned something like this:

```python
{
  'help_file': 'my-super-helpful-file',
  'error': None # or ''
}
```

```python
{
  'help_file': None, # or ''
  'error': 'No help file found'
}
```

Having the plugin return the `help_file` that was generated also encapsulates the default name behavior.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -562,14 +562,19 @@ def get_build(self, metadata):
...
-                extra['image']['help'] = add_help_config['args']['help_file']
-            except (KeyError, TypeError, IndexError):
-                extra['image']['help'] = DEFAULT_HELP_FILENAME
+            if help_result == AddHelpPlugin.HELP_GENERATED:
+                try:
+                    extra['image']['help'] = add_help_plugin_config[0]['args']['help_file']
+                except (KeyError, TypeError, IndexError):
+                    extra['image']['help'] = DEFAULT_HELP_FILENAME
+            elif help_result == AddHelpPlugin.NO_HELP_FILE_FOUND:
+                extra['image']['help'] = None
```
lcarva: Based on this PRs title, this seems to be exactly what we don't want :)

atomic_reactor/plugins/pre_add_help.py
lcarva: > return None if exception and an empty string if no help file was used

Why do we need to distinguish between these two cases?

> duplicating information from the config

It's not really duplicating information in the config. This plugin decides which filename to use; I think it makes perfect sense for it to return the value of the decision.

lcarva:
My review is based on my lack of understanding as to why we need to distinguish between help file not found and plugin didn't run use cases in a Koji build. It seems more natural to me to simply omit "help" key in either case. The logs do provide sufficient information for further analysis of what exactly happened.

atomic_reactor/plugins/pre_add_help.py
vrutkovs: Updated the code: now it returns a dict with `help_file` and a numeric status (so that `koji_promote` can easily detemine the result)

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -45,13 +48,23 @@ def __init__(self, tasker, workflow, help_file=DEFAULT_HELP_FILENAME):
...
+            'help_file': self.help_file,
+            'status': None
+        }
+
         help_path = os.path.join(self.workflow.builder.df_dir, self.help_file)
 
         if not os.path.exists(help_path):
             self.log.info("File %s not found", help_path)
-            return
+            result['result'] = self.NO_HELP_FILE_FOUND
```
lcarva: Shouldn't this say `result['status`]' ?

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -84,4 +97,5 @@ def run(self):
 
         self.log.info("added %s", man_path)
 
-        return content
+        result['result'] = self.HELP_GENERATED
```
lcarva: Shouldn't this say `result['status`]' ?

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -563,13 +563,13 @@ def get_build(self, metadata):
                     extra['filesystem_koji_task_id'] = task_id
 
         help_result = self.workflow.prebuild_results.get(AddHelpPlugin.key)
-        if help_result is not None:
-            try:
-                add_help_config = [x for x in self.workflow.prebuild_plugins_conf
-                                   if x['name'] == AddHelpPlugin.key][0]
-                extra['image']['help'] = add_help_config['args']['help_file']
-            except (KeyError, TypeError, IndexError):
-                extra['image']['help'] = DEFAULT_HELP_FILENAME
```
lcarva: Remove import for `DEFAULT_HELP_FILENAME`?

atomic_reactor/plugins/pre_add_help.py
vrutkovs: Fixed, updated `test_add_help` to catch that earlier

atomic_reactor/plugins/exit_koji_promote.py
vrutkovs: Ah, right, this one is no longer needed. Removed

atomic_reactor/plugins/pre_add_help.py
vrutkovs: Ditto

PR #638, opened 2017-02-03
Tweak tito config

vrutkovs:
This will make a package release as simple as:
```
tito tag
git push origin && git push origin atomic-reactor-<version>-<release>
tito release rhel7
```

Note, that 
 1. tito required those ugly tags to be created
 1. relies on version-release tarballs instead of a particular commit


.gitignore
```diff
@@ -1,4 +1,3 @@
-rel-eng/releasers.conf
```
lcarva: What was this for anyways?

rel-eng/packages/atomic-reactor
```diff
@@ -0,0 +1 @@
+1.6.19-1 ./
```
lcarva: Is this supposed to be latest version-release? If so, it's outdated.

rel-eng/releasers.conf
```diff
@@ -0,0 +1,3 @@
+[rhel7]
+releaser = tito.release.DistGitReleaser
+branches = eng-rhel-7
```
lcarva: Is this internal information?

atomic-reactor.spec
```diff
@@ -38,7 +38,8 @@ Summary:        Improved builder for Docker images
 Group:          Development/Tools
 License:        BSD
 URL:            https://github.com/%{owner}/%{project}
-Source0:        https://github.com/%{owner}/%{project}/archive/%{commit}/%{project}-%{commit}.tar.gz
+Source0:        %{name}-%{version}.tar.gz
+BuildRoot:      %(mktemp -ud %{_tmppath}/%{name}-%{version}-%{release}-XXXXXX)
```
lcarva: koji-containerbuild uses this:

```
%{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
```

The `-XXXXXX` is odd, but honestly I have no idea which one is better.

rel-eng/releasers.conf
vrutkovs: Yes, but I think its safe to have this set (dockpulp has it too)

rel-eng/packages/atomic-reactor
vrutkovs: Updated

.gitignore
vrutkovs: Previously `releasers.conf` was ignored, as it contained information about dist-git and internal COPR. Tito needs this file, so the releasing engineer had to have this file there.

I removed the copr configuration and now `releasers.conf` contains only a distgit releaser and the only internal info there is branch name

atomic-reactor.spec
vrutkovs: The `-XXXXX` version is recommended on a wiki page and it won't require removing it in `%clean`. Not sure which is better to be honest

twaugh: Don't use BuildRoot.
https://fedoraproject.org/wiki/Packaging:Guidelines#Tags_and_Sections

vrutkovs: Removed `BuildRoot`

PR #639, opened 2017-02-03
Ignore 'dockerfile_content' plugin requests

twaugh:
This change provides more compatibility when upgrading. If osbs-client is still requesting 'dockerfile_content' but it has been removed, we will ignore it.

Note that 'required' was broken prior to this release, hence the need for this compatibility work-around.

### DIFF
--- a/atomic_reactor/plugins/input_osv3.py
+++ b/atomic_reactor/plugins/input_osv3.py
@@ -62,6 +62,16 @@ class OSv3InputPlugin(InputPlugin):
 
         self.log.debug("build json: %s", input_json)
 
+        # Compatibility code for dockerfile_content plugin
+        # If this (removed) plugin is requested, ignore it.
+        if 'prebuild_plugins' in self.plugins_json:
+            for index, plugin in enumerate(self.plugins_json['prebuild_plugins']):
+                if plugin['name'] == 'dockerfile_content':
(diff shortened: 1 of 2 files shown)


twaugh:
This is to avoid making atomic-reactor 1.6.20 the minimum version osbs-client will work correctly with.

atomic_reactor/plugins/input_osv3.py
```diff
@@ -62,6 +62,16 @@ def run(self):
 
         self.log.debug("build json: %s", input_json)
 
+        # Compatibility code for dockerfile_content plugin
+        # If this (removed) plugin is requested, ignore it.
+        if 'prebuild_plugins' in self.plugins_json:
+            for index, plugin in enumerate(self.plugins_json['prebuild_plugins']):
+                if plugin['name'] == 'dockerfile_content':
```
vrutkovs: `cp_dockerfile` was deprecated also, should we try to remove it as well?

twaugh: `dockerfile_content` is the only one that prevents using atomic-reactor-1.6.20 with osbs-client-0.33.1. I was thinking we'd remove this work-around before 1.6.21.

vrutkovs: Ah, right, its not worth adding it here as its not used by default

PR #640, opened 2017-02-07
Catch ImportError in koji_promote

twaugh:
If the check_and_set_rebuild plugin is not available (for example, the RPM sub-package containing it is not installed), catch the ImportError and stub is_rebuild().


twaugh:
Not sure if a test-case can be written for this. Any ideas?

lcarva:
You can try patching `is_rebuild` but probably need to force re-import the module which can get messy.

vrutkovs:
Not sure if we really need this - its a packaging issue and doesn't happen if `atomic-reactor` is installed from source (this is why buildroot pipeline didn't catch it)

twaugh:
We either need to do this, or fix the spec file so it doesn't install those plugins separately.

twaugh:
Closing, approach in #642 is better.

PR #642, opened 2017-02-08
spec: make sure pre_check_and_set_rebuild plugin is always installed

vrutkovs:
koji_promote and store_metadata_in_osv3 depends on it

Alternative to #640


lcarva:
Does it make sense to include `check_and_set_rebuild` plugin in both `python-atomic-reactor-koji` and `python-atomic-reactor-metadata`. Not sure if that's allowed in RPM, but it would be neat to have a clearer dependency stated.

vrutkovs:
No, we can't do that - it would conflict if both RPMs are installed

vrutkovs:
>IMHO we should make a function for doing this and put it in util.py, and use that instead of check_and_set_rebuild.

Yes, +1 on that

PR #643, opened 2017-02-08
unit tests for buildsteps

rcerven:


atomic_reactor/inner.py
```diff
@@ -379,16 +379,6 @@ def build_docker_image(self):
...
 
-            self.build_logs = build_result.logs
-
-            self.build_failed = build_result.is_failed()
-
-            if build_result.is_failed():
-                # The docker build failed. Finish here, just run the
-                # exit plugins (from the 'finally:' block below).
-                self.plugins_errors['dockerbuild'] = ''
-                return build_result
```
lcarva: Why don't we want this anymore? Seems pretty important.

tests/test_inner.py
```diff
@@ -755,36 +790,51 @@ def test_workflow_docker_build_error():
...
-    # No subsequent build phases should have run except 'exit'
-    assert not watch_prepub.was_called()
-    assert not watch_post.was_called()
-    assert watch_exit.was_called()
+    with pytest.raises(Exception):
+
+        workflow.build_docker_image()
+        # No subsequent build phases should have run except 'exit'
+        assert watch_pre.was_called()
+        assert watch_build.was_called()
```
lcarva: Because this raises an exception, the assert statements after this are never executed.

tests/test_inner.py
```diff
@@ -755,36 +790,51 @@ def test_workflow_docker_build_error():
...
+                                                       'watcher': watch_post
                                                        }}],
                                    exit_plugins=[{'name': 'exit_watched',
                                                   'args': {
                                                       'watcher': watch_exit
                                                   }}],
                                    plugin_files=[this_file])
 
-    assert workflow.build_docker_image().is_failed()
+    flexmock(watch_build).should_call('was_called').and_raise(Exception)
```
lcarva: Why raise an exception here? It's method that is called from the unit test, not from the code you're testing.

tests/test_inner.py
```diff
@@ -891,21 +943,16 @@ def test_cancel_build(request, fail_at):
...
-    build_timeout = 10 if fail_at == 'build' else 0
-    sigterm_timeout = 2
-    fake_builder = MockInsideBuilder(timeout=build_timeout)
+    fake_builder = MockInsideBuilder()
     flexmock(InsideBuilder).new_instances(fake_builder)
-    watch_pre = WatcherWithSignal(phase_signal['pre'])
-    watch_prepub = WatcherWithSignal(phase_signal['prepub'])
-    watch_post = WatcherWithSignal(phase_signal['post'])
-    watch_exit = WatcherWithSignal(phase_signal['exit'])
+    watch_pre = WatcherWithSignal(signal.SIGTERM)
```
lcarva: The reasoning in this test seems to have changed completely, why?

atomic_reactor/inner.py
rcerven: pluginsrunner, will call on_plugin_failed from buildpluginsrunner
which does collect errors

tests/test_inner.py
rcerven: I know but I need to simulate that build failed and that is by raising exc from build plugins

tests/test_inner.py
rcerven: I see, well have to change how it works

tests/test_inner.py
rcerven: because now we are treating build plugins as any other type of plugins, but it still tests for canceling each phase

atomic_reactor/inner.py
twaugh: @lcarva, what do you think about getting rid of BuildResults altogether? This method would just return self.plugin_failed and build_inside() would use the return value to decide whether the build failed.

tests/test_inner.py
vrutkovs: This doesn't really test the code then - exception should be thrown from `inner.py`

tests/test_inner.py
vrutkovs: `WatcherWithSignal(phase_signal['exit'])` does exactly the same - prefills `phase_signal` with a correct signal to call. This

tests/test_inner.py
```diff
@@ -935,20 +986,9 @@ def restore_logger():
...
-                os.kill(pid, signal.SIGTERM)))
-        thread.start()
-
-        with pytest.raises(BuildCanceledException):
-            workflow.build_docker_image()
-    else:
-        workflow.build_docker_image()
+    workflow.build_docker_image()
 
-    if fail_at not in ['exit', 'build']:
```
vrutkovs: This should be retained - exception should be thrown if build has failed

tests/test_inner.py
```diff
@@ -76,12 +77,21 @@ class X(object):
...
+    def __init__(self, failed=False):
         self.tasker = MockDockerTasker()
         self.base_image = ImageName(repo='Fedora', tag='22')
         self.image_id = 'asd'
+        self.image = 'image'
         self.failed = failed
-        self.timeout = timeout
+        self.df_path = 'some'
+        self.df_dir = 'some'
+        self.tasker.build_image_from_path = flexmock()
```
twaugh: Seems like this isn't needed if you're mocking it on line 94?

tests/test_inner.py
```diff
@@ -93,18 +103,14 @@ def source(self):
...
-        return result
-
     def inspect_built_image(self):
         return None
 
+    def _ensure_not_built(self):
+        pass
+
+#    def get_built_image_info(self):
+#        return {'Id': 'some'}
```
twaugh: Commented-out lines not needed.

tests/test_inner.py
```diff
@@ -240,24 +253,41 @@ def call(self):
...
+class WatcherWithPause(Watcher):
+    def __init__(self, timeout=10):
+        super(WatcherWithPause, self).__init__()
+        self.timeout = timeout
+
+    def call(self):
+        super(WatcherWithPause, self).call()
+        sleep(self.timeout)
+
+
```
twaugh: This class is not used and can be removed.

tests/test_inner.py
```diff
@@ -268,14 +298,15 @@ def test_workflow():
                                                        }}],
                                    exit_plugins=[{'name': 'exit_watched',
                                                   'args': {
-                                                      'watcher': watch_exit
+                                                  'watcher': watch_exit
```
twaugh: This change looks wrong.

tests/test_inner.py
```diff
@@ -755,36 +790,51 @@ def test_workflow_docker_build_error():
...
     watch_exit = Watcher()
 
     workflow = DockerBuildWorkflow(MOCK_SOURCE, 'test-image',
+                                   prebuild_plugins=[{'name': 'pre_watched',
+                                                      'args': {
+                                                          'watcher': watch_pre
+                                                      }}],
+                                   buildstep_plugins=[{'name': 'build_watched',
+                                                       'args': {
+                                                       'watcher': watch_build
```
twaugh: Indentation looks wrong.

atomic_reactor/inner.py
lcarva: That seems reasonable.

tests/test_inner.py
```diff
@@ -755,36 +790,51 @@ def test_workflow_docker_build_error():
...
+                                                       }}],
                                    prepublish_plugins=[{'name': 'prepub_watched',
                                                         'args': {
                                                             'watcher': watch_prepub,
                                                         }}],
                                    postbuild_plugins=[{'name': 'post_watched',
                                                        'args': {
-                                                           'watcher': watch_post
+                                                       'watcher': watch_post
```
twaugh: This looks wrong.

tests/test_inner.py
```diff
@@ -218,18 +232,19 @@ class ExitCompat(WatchedMixIn, ExitPlugin):
 
 
 class Watcher(object):
-    def __init__(self):
+    def __init__(self, rais=False):
```
twaugh: Typo in parameter name, should this be 'raise'? Why not create a new class using the RaisesMixIn?

tests/test_inner.py
```diff
@@ -779,9 +828,13 @@ def test_workflow_docker_build_error():
                                                   }}],
                                    plugin_files=[this_file])
 
-    assert workflow.build_docker_image().is_failed()
-
+    try:
+        workflow.build_docker_image()
+    except Exception as ex:
+        pass
```
twaugh: Do we expect an exception? If so we should test for it.

tests/test_inner.py
rcerven: yeah it did the same but this way is more clear, as there is only 1 signal anyway used, no need to complicate things

tests/test_inner.py
rcerven: that was changed already

tests/test_inner.py
twaugh: But the way you have it is wrong. Every plugin will always raise a signal, rather than only the one we want to test.

tests/test_plugin.py
```diff
@@ -88,6 +110,89 @@ def test_not_required_prebuild_plugin_failure(docker_tasker, runner_type, requir
         results = runner.run()
     assert workflow.plugin_failed is required
 
+@pytest.mark.parametrize('return1,return2', [
```
twaugh: Can these be named more clearly?

tests/test_inner.py
rcerven: ahhh ok now I got it

tests/test_inner.py
rcerven: well I need to declare it, if I mock nonexisting variable, it will complain

tests/test_inner.py
rcerven: yeah I need build plugins to raise exception, so I can check that only exit plugins are run after

tests/test_inner.py
rcerven: well is not typo, but I can't use keyword, I've changed the name

well RaiseMixIn will raise on run, and won't call watcher, I need to call watcher to "count" and after that raise, so raising in the watcher is actually nice

tests/test_inner.py
twaugh: So should this be using `with pytest.raises():`?

rcerven: I had that before, but Lui pointed out that the asserts after won't be run anymore, which is what I need to check

twaugh: You want something like:
```
with pytest.raises(PluginFailedException):
    workflow.build_docker_image()
assert ...
```

i.e. just don't have the assert in the same block as the thing you expect to raise an exception (because it will escape that block).

rcerven: ok that works

tests/test_inner.py
```diff
@@ -93,18 +103,12 @@ def source(self):
...
-        if self.timeout:
-            sleep(self.timeout)
-
-        return result
-
     def inspect_built_image(self):
         return None
 
+    def _ensure_not_built(self):
+        pass
```
twaugh: What is this for?

tests/test_inner.py
```diff
@@ -129,6 +133,14 @@ class PreRaises(RaisesMixIn, PreBuildPlugin):
     key = 'pre_raises'
 
 
+class BuildStepRaises(RaisesMixIn, BuildStepPlugin):
+    """
+    This plugin must run and cause the build to abort.
+    """
+
+    key = 'buildstep_raises'
```
twaugh: Let's be consistent with what we call these objects. The 'watched' one is called BuildWatched with key 'build_watched'. Let's choose either "Build..."+"build_..." or "BuildStep..."+"buildstep_...".

tests/test_inner.py
```diff
@@ -755,16 +798,28 @@ def test_workflow_docker_build_error():
     """
     This is a test for what happens when the docker build fails.
     """
-
+    flexmock(DockerfileParser, content='df_content')
     this_file = inspect.getfile(PreRaises)
     mock_docker()
     fake_builder = MockInsideBuilder(failed=True)
     flexmock(InsideBuilder).new_instances(fake_builder)
+    watch_pre  = Watcher()
```
twaugh: There's an extra space here.

tests/test_inner.py
rcerven: BuildStepPluginRunner is using that

tests/test_inner.py
```diff
@@ -807,6 +867,9 @@ def test_source_not_removed_for_exit_plugins():
                                                   'args': {
                                                       'watcher': watch_exit,
                                                   }}],
+
+                                   buildstep_plugins=[{'name': 'build_watched'
+                                                       }],
```
twaugh: Could make this indentation better -- either all on one line, or with `'name': ...` on its own line.

tests/test_inner.py
twaugh: Ah, got it.

tests/test_plugin.py
```diff
@@ -88,6 +110,89 @@ def test_not_required_prebuild_plugin_failure(docker_tasker, runner_type, requir
...
+])
+@pytest.mark.parametrize('stop', [
+    True,
+    False,
+])
+def test_stop_on_success_build_plugin(caplog, docker_tasker, stop, success1, success2):
+    """
+    test stop on success option
+    if is true, plugin runner should stop after first successful plugin
+    and if is false it should run all plugins
```
twaugh: This is unclear. Which parameter is it talking about?

tests/test_plugin.py
```diff
@@ -88,6 +110,89 @@ def test_not_required_prebuild_plugin_failure(docker_tasker, runner_type, requir
...
+    setattr(workflow.builder.image, 'to_str', lambda: 'image')
+    setattr(workflow.builder.source, 'path', 'path')
+    setattr(workflow.builder, 'base_image', flexmock())
+    setattr(workflow.builder.base_image, 'to_str', lambda: 'base-image')
+
+    flexmock(PluginsRunner, load_plugins=lambda x: {
+                                        DockerApiPlugin.key: DockerApiPlugin,
+                                        MyPlugin.key: MyPlugin})
+
+    class MyPlugin(object):
```
twaugh: Could make this `MyPlugin(BuildStepPlugin)`.

tests/test_plugin.py
```diff
@@ -88,6 +110,89 @@ def test_not_required_prebuild_plugin_failure(docker_tasker, runner_type, requir
...
+@pytest.mark.parametrize('success1,success2', [
+    (True, True),
+    (True, False),
+    (False, True),
+    (False, False),
+])
+@pytest.mark.parametrize('stop', [
+    True,
+    False,
+])
```
twaugh: This test doesn't actually make any assertions when `stop` is False, and other tests cover the False case anyway.

It might help readability to remove this parameter. I don't think there is much value in testing what happens when a plugin raises InappropriateBuildStepError but stop_on_success isn't set, since that's not really a valid situation.

twaugh:
When there are multiple build step plugins configured, and the first one raises some other exception than InappropriateBuildStepError, we should not run further build step plugins. Only exit plugins should be run after that.

Do we have a test case that covers this scenario?

rcerven:
yeah I just wrote that one that

tests/test_inner.py
```diff
@@ -70,18 +71,28 @@ class MockDockerTasker(object):
...
         self.tasker = MockDockerTasker()
         self.base_image = ImageName(repo='Fedora', tag='22')
         self.image_id = 'asd'
+        self.image = 'image'
         self.failed = failed
-        self.timeout = timeout
+        self.df_path = 'some'
+        self.df_dir = 'some'
+        self.get_built_image_info = flexmock()
+        flexmock(self, get_built_image_info={'Id': 'some'})
```
twaugh: This could be a method on this MockInsideBuilder class couldn't it?

tests/test_inner.py
twaugh: This looks fine now, thanks.

tests/test_inner.py
```diff
@@ -723,7 +769,6 @@ def test_workflow_plugin_error(fail_at):
     # an exit plugin that's explicitly allowed to fail.
     if fail_at == 'exit_raises_allowed':
         build_result = workflow.build_docker_image()
-        assert build_result and not build_result.is_failed()
```
twaugh: What does workflow.build_docker_image() return now, nothing? It just raises an exception on error? If that's the case we shouldn't assign to build_result.

atomic_reactor/inner.py
twaugh: So, I think if we remove the lines above we also need to remove BuildResults altogether.

tests/test_plugin.py
```diff
@@ -88,6 +111,174 @@ def test_not_required_prebuild_plugin_failure(docker_tasker, runner_type, requir
...
+    flexmock(PluginsRunner, load_plugins=lambda x: {
+                                        DockerApiPlugin.key: DockerApiPlugin,
+                                        MyPlugin.key: MyPlugin})
+
+    class MyPlugin(BuildStepPlugin):
+        key = 'MyPlugin'
+        def __init__(self, tasker, workflow, spam=None, **kwargs):
+            self.spam = spam
+        def run (self):
+            pass
```
twaugh: Any reason we need to define any methods here at all?

tests/test_plugin.py
```diff
@@ -88,6 +111,174 @@ def test_not_required_prebuild_plugin_failure(docker_tasker, runner_type, requir
...
+                flexmock(MyPlugin).should_receive('run').and_raise(InappropriateBuildStepError).times(1)
+
+        else:
+            flexmock(DockerApiPlugin).should_receive('run').and_raise(Exception).times(1)
+            will_fail = True
+            flexmock(MyPlugin).should_receive('run').times(0)
+
+    if will_fail:
+        with pytest.raises(Exception):
+            results = runner.run(stop_on_success=stop)
```
twaugh: Is `results` used?

tests/test_plugin.py
```diff
@@ -88,6 +111,174 @@ def test_not_required_prebuild_plugin_failure(docker_tasker, runner_type, requir
...
+
+    else:
+        results = runner.run(stop_on_success=stop)
```
twaugh: Is `results` used?

tests/test_plugin.py
twaugh: (and remove the methods)

tests/test_inner.py
rcerven: it depends how build finishes, can be None, Exception, or really results

atomic_reactor/inner.py
rcerven: not really, we still have build results BUT only if buildstep plugin finishes successfully

tests/test_plugin.py
rcerven: not anymore when subclassing from buildstepplugin

atomic_reactor/inner.py
twaugh: OK, I'm going to let this go for this PR since it's largely about getting unit tests in order, but I will remember it when we want to merge into master. :-)

tests/test_plugin.py
twaugh: (It actually tests that the plugins runner does not raise an exception, so that's fine.)

rcerven: yop exactly

tests/test_plugin.py
```diff
@@ -88,6 +111,166 @@ def test_not_required_prebuild_plugin_failure(docker_tasker, runner_type, requir
...
+])
+def test_stop_on_success_build_plugin(caplog, docker_tasker, stop, success1, success2):
+    """
+    test stop on success option
+    if 'stop' is true, plugin runner should stop after first successful plugin
+    and if 'stop' is false it should run all plugins
+    InappropriateBuildStepError exception isn't critical,
+    and won't fail buildstep runner
+    """
+    workflow = DockerBuildWorkflow(SOURCE, 'test-image')
```
lcarva: Can we have this abstracted? Maybe a helper function that returns the workflow?

PR #646, opened 2017-02-13
Clone git repo into local dirname matching repo name

twaugh:
Previously the cloned repository was placed in a directory named 'source'. Some tools running as the sources_command may use the directory name they run in to obtain the git module name, to avoid a dependency on 'git'.

Accommodate such tools by placing the cloned repository in an appropriately-named directory.

### DIFF
--- a/atomic_reactor/source.py
+++ b/atomic_reactor/source.py
@@ -15,9 +15,12 @@ import os
 import shutil
 import tempfile
 import collections
+try:
+    from urlparse import urlparse
+except ImportError:
+    from urllib.parse import urlparse
 
 from atomic_reactor import util
-from atomic_reactor.constants import SOURCE_DIRECTORY_NAME
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


twaugh:
Tests currently fail, not entirely sure how to fix yet.

twaugh:
Tests now pass but that's because I removed some for the non-git source case.

atomic_reactor/source.py
```diff
@@ -37,7 +40,12 @@ def __init__(self, provider, uri, dockerfile_path=None, provider_params=None, tm
         # TODO: do we want to delete tmpdir when destroying the object?
         self.tmpdir = tmpdir or tempfile.mkdtemp()
         logger.debug("workdir is %r", self.tmpdir)
-        self.source_path = os.path.join(self.tmpdir, SOURCE_DIRECTORY_NAME)
+        (_, _, uri_path, _, _, _) = urlparse(uri)
```
lcarva: `urlparse` returns a named tuple. The `path` attribute should get this value.

twaugh: Updated to use `path` attribute.

PR #647, opened 2017-02-13
Abstract build step

rcerven:


atomic_reactor/inner.py
```diff
@@ -363,7 +363,7 @@ def build_docker_image(self):
                                                      self.buildstep_plugins_conf,
                                                      plugin_files=self.plugin_files)
             try:
-                buildstep_runner.run(stop_on_success=True)
+                buildstep_runner.run(buildstep_phase=True)
```
vrutkovs: Wouldn't `is_a_buildstep` be better?

rcerven: what do you think ? @lcarva @twaugh

twaugh: I think `buildstep_phase` is fine.

atomic_reactor/plugins/build_docker_api.py
```diff
@@ -0,0 +1,50 @@
...
+"""
+from __future__ import print_function, unicode_literals
+
+from atomic_reactor.plugin import BuildStepPlugin
+from atomic_reactor.util import wait_for_command
+
+
+class DockerApiPlugin(BuildStepPlugin):
+    """
+    TODO
```
lcarva: Please update.

atomic_reactor/plugins/build_docker_api.py
```diff
@@ -0,0 +1,50 @@
+"""
+Copyright (c) 2016 Red Hat, Inc
```
lcarva: 2017

atomic_reactor/plugin.py
```diff
@@ -258,10 +274,21 @@ def run(self, keep_going=False):
...
+                logger.debug('stopping further execution of plugins '
+                             'after first successful plugin')
+                break
+
         if len(failed_msgs) == 1:
             raise PluginFailedException(failed_msgs[0])
         elif len(failed_msgs) > 1:
             raise PluginFailedException("Multiple plugins raised an exception: " + str(failed_msgs))
+
+        if (plugins_skipped == len(self.plugins_conf)) and buildstep_phase:
```
rcerven: that would work, except for when you would also set is_allowed_to_fail on plugin before last failed,

this way it really counts only when it raises InappropriateBuildStepError,
or when plugin doesn't exist

lcarva: In that case, `plugin_successful` wouldn't be set to `True`, right?

If a build phase plugin fails and is allowed to fail (valid scenario??), then `plugins_skipped` is not incremented. In which case, this method wouldn't raise any exceptions. To me that seems incorrect.

rcerven: well as of now is probably not valid, but maybe in future we could use that for something?

anyway this exception is specific only for nonexisting or raises InappropriateBuildStepError

atomic_reactor/constants.py
```diff
@@ -24,6 +24,7 @@
 HOST_SECRET_PATH = ''
 
 EXPORTED_SQUASHED_IMAGE_NAME = 'image.tar'
+EXPORTED_BUILT_IMAGE_NAME = 'built-image.tar'
```
twaugh: Not used?

atomic_reactor/plugin.py
```diff
@@ -354,6 +381,57 @@ def __init__(self, dt, workflow, plugins_conf, *args, **kwargs):
...
+    def run(self, *args, **kwargs):
+        builder = self.workflow.builder
+
+        logger.info('building image %r inside current environment',
+                    builder.image)
+        builder._ensure_not_built()
+        logger.debug('using dockerfile:\n%s',
+                     DockerfileParser(builder.df_path).content)
+
+        plugins_results = super(BuildStepPluginsRunner, self).run(*args, **kwargs)
```
twaugh: This invocation should supply buildstep_phase=True to the inherited class's run(), instead of having callers of this run() have to supply it.

atomic_reactor/plugin.py
```diff
@@ -171,15 +176,20 @@ def save_plugin_timestamp(self, plugin, timestamp):
...
 
-    def run(self, keep_going=False):
+    def run(self, keep_going=False, buildstep_phase=False):
         """
         run all requested plugins
 
         :param keep_going: bool, whether to keep going after unexpected
                                  failure (only used for exit plugins)
+        :param buildstep_phase: bool, when True remaining plugins will
+                                not be executed after a plugin succeeds
```
twaugh: Instead of 'succeeds', how about 'completes'?

atomic_reactor/plugin.py
```diff
@@ -354,6 +381,57 @@ def __init__(self, dt, workflow, plugins_conf, *args, **kwargs):
...
+        self._build_result = None
+
+        logger.info("initializing runner of build-step plugin")
+        self.plugins_results = workflow.buildstep_result
+
+        if not plugin_conf:
+            plugin_conf = [{'name': 'docker_api'}]
+        else:
+            for i in range(len(plugin_conf)):
+                plugin_conf[i]['required'] = False
```
twaugh: How about a comment explaining why we're doing this (i.e. any build plugin which cannot be instantiated must be skipped without error).

atomic_reactor/plugin.py
```diff
@@ -354,6 +381,57 @@ def __init__(self, dt, workflow, plugins_conf, *args, **kwargs):
         self.plugins_results = workflow.prebuild_results
         super(PreBuildPluginsRunner, self).__init__(dt, workflow, 'PreBuildPlugin', plugins_conf, *args, **kwargs)
 
+
+class BuildStepPlugin(BuildPlugin):
+    pass
```
twaugh: How about setting the default `is_allow_to_fail` to `False` for build-step plugins, and not setting it in docker_api?

atomic_reactor/plugins/build_docker_api.py
```diff
@@ -0,0 +1,51 @@
...
+            BuildResult
+            built_image_info
+            image_id
+        """
+        builder = self.workflow.builder
+
+        logs_gen = self.tasker.build_image_from_path(
+            builder.df_dir,
+            builder.image,
+        )
```
twaugh: Minor, but is there any reason not to put this on two lines?
```
logs_gen = ...(builder.df_dir,
               builder.image)
```

atomic_reactor/plugins/post_push_to_pulp.py
```diff
@@ -340,6 +340,9 @@ def run(self):
             self.log.info("fetching image %s from docker", image)
             with tempfile.NamedTemporaryFile(prefix='docker-image-', suffix='.tar') as image_file:
                 image_file.write(self.tasker.d.get_image(image).data)
+                # This file will be referenced by its filename, not file
+                # discriptor - must ensure contents are written to disk
```
twaugh: Typo, 'descriptor'.
Has this bug been lurking all along? Would it be better in its own commit (and may PR)?

tests/test_inner.py
```diff
@@ -526,6 +609,19 @@ class Exit(ExitPlugin):
...
+                             'required': False},
+                            {'name': 'buildstep_watched',
+                             'args': {
+                                 'watcher': Watcher(),
+                             }
+                            }]},
+     False,  # not fatal
+     False,  # does not log error
+    ),
+
```
twaugh: Could we add a test-case here for when there are multiple buildstep_plugins which would succeed, to verify that only the first one is called?

tests/test_inner.py
```diff
@@ -755,16 +867,28 @@ def test_workflow_docker_build_error():
...
     workflow = DockerBuildWorkflow(MOCK_SOURCE, 'test-image',
+                                   prebuild_plugins=[{'name': 'pre_watched',
+                                                      'args': {
+                                                          'watcher': watch_pre
+                                                      }}],
+                                   buildstep_plugins=[{'name': 'buildstep_watched',
+                                                       'args': {
+                                                           'watcher': watch_buildstep,
+                                                       },
+                                                       'is_allowed_to_fail': False
```
twaugh: See comment elsewhere -- I think this should be the default for BuildStep plugins.

tests/test_plugin.py
```diff
@@ -31,15 +36,47 @@
 TEST_IMAGE = "fedora:latest"
 SOURCE = {"provider": "git", "uri": DOCKERFILE_GIT}
 
+class MyBsPlugin1(BuildStepPlugin):
+    key = 'MyBsPlugin1'
+class MyBsPlugin2(BuildStepPlugin):
```
twaugh: Two blank lines before class definitions.

atomic_reactor/plugins/post_push_to_pulp.py
lcarva: Yes. It's triggered by not using squash plugin which was uncovered when building with `oc dockerbuild`.

atomic_reactor/plugin.py
twaugh: Build plugins should not have `is_allowed_to_fail` set. I don't think that can be valid.

tests/test_inner.py
rcerven: that test is in test_plugin

twaugh: Can we add a test here to make sure the build doesn't get marked as failed in that situation?

atomic_reactor/plugin.py
twaugh: Still doesn't seem to be set here.

atomic_reactor/plugins/post_push_to_pulp.py
```diff
@@ -340,6 +340,9 @@ def run(self):
             self.log.info("fetching image %s from docker", image)
             with tempfile.NamedTemporaryFile(prefix='docker-image-', suffix='.tar') as image_file:
                 image_file.write(self.tasker.d.get_image(image).data)
+                # This file will be referenced by its filename, not file
+                # descriptor - must ensure contents are written to disk
+                image_file.flush()
```
twaugh: I would like this change to be in a separate PR.

tests/test_inner.py
twaugh: OK, not needed.

tests/test_plugin.py
```diff
@@ -70,23 +109,140 @@ def test_build_plugin_failure(docker_tasker):
...
+    flexmock(PluginsRunner, load_plugins=lambda x: {
+                                        MyBsPlugin1.key: MyBsPlugin1,
+                                        MyBsPlugin2.key: MyBsPlugin2,})
+    runner = BuildStepPluginsRunner(docker_tasker, workflow,
+                                   [{"name": MyBsPlugin1.key},
+                                    {"name": MyBsPlugin2.key}])
+
+    will_fail = False
+    if success1:
+        flexmock(MyBsPlugin1).should_receive('run')
```
twaugh: `.once()`?

tests/test_plugin.py
```diff
@@ -70,23 +109,140 @@ def test_build_plugin_failure(docker_tasker):
...
+        flexmock(MyBsPlugin2).should_receive('run').times(0)
```
twaugh: Maybe `.never()` reads better?

tests/test_plugin.py
```diff
@@ -70,23 +109,140 @@ def test_build_plugin_failure(docker_tasker):
...
+    else:
+        flexmock(MyBsPlugin1).should_receive('run').and_raise(InappropriateBuildStepError).times(1)
```
twaugh: Maybe `.once()` reads better?

tests/test_plugin.py
```diff
@@ -70,23 +109,140 @@ def test_build_plugin_failure(docker_tasker):
...
+        assert workflow.plugin_failed is True
+    else:
+        assert workflow.plugin_failed is required
+
+@pytest.mark.parametrize('success1,success2', [
+    (True, True),
+    (True, False),
+    (False, True),
+    (False, False),
+])
```
twaugh: Could write this as
```
@pytest.mark.parametrize('success1', [True, False])
@pytest.mark.parametrize('success2', [True, False])
```

atomic_reactor/plugin.py
rcerven: I've set it together with required at 402

twaugh: Ah, ok, and that overrides anything set in the plugin class, got it.

tests/test_plugin.py
```diff
@@ -70,23 +109,140 @@ def test_build_plugin_failure(docker_tasker):
...
+        expected_log_message = "stopping further execution of plugins after first successful plugin"
+        assert expected_log_message in [l.getMessage() for l in caplog.records()]
+
+@pytest.mark.parametrize('success1,success2', [
+    (True, True),
+    (True, False),
+    (False, True),
+    (False, False),
+])
+def test_buildstep_phase_build_plugin_failing_exception(tmpdir, caplog, docker_tasker, success1, success2):
```
twaugh: This doesn't seem to use success2.

rcerven: yeah not needed

atomic_reactor/plugins/post_push_to_pulp.py
rcerven: https://github.com/projectatomic/atomic-reactor/pull/649

atomic_reactor/plugin.py
```diff
@@ -219,17 +229,21 @@ def run(self, keep_going=False):
...
+                plugin_successful = True
             except AutoRebuildCanceledException as ex:
                 # if auto rebuild is canceled, then just reraise
                 # NOTE: We need to catch and reraise explicitly, so that the below except clause
                 #   doesn't catch this and make PluginFailedException out of it in the end
                 #   (calling methods would then need to parse exception message to see if
                 #   AutoRebuildCanceledException was raised here)
                 raise
+            except InappropriateBuildStepError:
+                logger.debug('Build step %s is not appropriate', plugin_class.key)
```
rcerven: sure

rcerven:
yeah will update doc also

twaugh:
I think the biggest remaining issue relates to the return value from the build-step plugin.

It does make sense to separate the build-step log output (i.e. output from whatever the build-step plugin does) from the rest of the atomic-reactor logs. The koji_promote plugin stores this as a separate log file output in the Koji Build object. Doing this allows the user to quickly find what's wrong with their Dockerfile.

Given that, we *also* need a way of signalling that the build-step failed in addition to providing logs for it.

If the build succeeded, we need to know what the resulting image ID was.

So it looks like the conclusion is that every build-step plugin *must* return a BuildResult object. However, I think that object needs changing a bit. At the moment BuildResult and ExceptionBuildResult are not related. Also BuildResult assumes it will be formed from a CommandResult which is not necessarily the case.

So I'd like to propose that we change BuildResult to be like this:
```
class BuildResult(object):
    def __init__(self, logs=None, fail_reason=None, image_id=None):
        """
        :param logs: iterable of log lines (without newlines)
        :param fail_reason: str, description of failure or None if successful
        :param image_id: str, ID of built container image
        """
        assert fail_reason or image_id  # must provide one of these
        assert not (fail_reason and image_id)  # not both
        self._logs = logs or []
        self._fail_reason = fail_reason
        self._image_id = image_id

    @property
    def logs(self):
        return self._logs

    @property
    def fail_reason(self):
        return self._fail_reason

    def is_failed(self):
        return self._fail_reason is not None

    @property
    def image_id(self):
        return self._image_id
```

Also we should remove ExceptionBuildResult and instead change any uses of it to BuildResult(fail_reason=...).

Then in PluginsRunner.run() I think we should 'assert' that build-step plugins must return a BuildResult instance from their run() methods, and document that in the class docstring and in docs/plugins.md. If the buildresult.is_failed() is False we should call on_plugin_failed(buildresult.fail_reason).

Finally in build_docker_image() we should inspect the image ID returned by the plugin to get built_image_inspect.

What do you think of this idea?

atomic_reactor/inner.py
```diff
@@ -357,16 +379,6 @@ def build_docker_image(self):
             except Exception:
                 logger.exception("failed to save build duration")
 
-            self.build_logs = build_result.logs
```
twaugh: See comment I made in main discussion.

rcerven:
ExceptionBuildResult isn't used anywhere

but the idea of it was good, but putting it all into 1 class is nicer, and it would solve that awkward behavior when exception, as there would be always buildresult

mmm leaving it for build_docker_image() is good idea, because as of now anyway if 1 plugins fails, buildsteppluginsrunner won't continue anyway but build_docker_image will, so it can properly set  buildresult

atomic_reactor/inner.py
```diff
@@ -346,7 +360,20 @@ def build_docker_image(self):
             start_time = datetime.datetime.now()
             self.plugins_timestamps['dockerbuild'] = start_time.isoformat()
```
rcerven: yeah that is useless now, as it is already tracked by pluginsrunner

atomic_reactor/plugin.py
```diff
@@ -258,10 +284,21 @@ def run(self, keep_going=False):
...
+                logger.debug('stopping further execution of plugins '
+                             'after first successful plugin')
+                break
+
         if len(failed_msgs) == 1:
             raise PluginFailedException(failed_msgs[0])
         elif len(failed_msgs) > 1:
             raise PluginFailedException("Multiple plugins raised an exception: " + str(failed_msgs))
+
+        if not plugin_successful and buildstep_phase and not plugin_response:
```
twaugh: Do we need to check plugin_response here? On line 288 we're happy with just checking the other two variables.

atomic_reactor/plugin.py
```diff
@@ -354,6 +391,44 @@ def __init__(self, dt, workflow, plugins_conf, *args, **kwargs):
...
+        logger.info('building image %r inside current environment',
+                    builder.image)
+        builder._ensure_not_built()
+        logger.debug('using dockerfile:\n%s',
+                     DockerfileParser(builder.df_path).content)
+
+        kwargs['buildstep_phase'] = True
+
+        plugins_results = super(BuildStepPluginsRunner, self).run(*args, **kwargs)
+        return plugins_results
```
twaugh: Since this class only ever returns a single result, how about just returning that single result?
```
return plugins_results.values()[0]
```

This would allow build_docker_image() to continue to return a BuildResult instead of changing it to return a list (and have to modify callers to deal with the list).

rcerven: yeah I am actually expecting BuildResult in build_docker_image() already

atomic_reactor/plugin.py
rcerven: 288 has plugin_successful True, so it won't hit here
but if all plugins are skipped OR plugin fails (no raising), I have no other way to distinguish those 2 cases then also check plugin_response

twaugh: Ah, ok.

atomic_reactor/inner.py
```diff
@@ -343,31 +357,31 @@ def build_docker_image(self):
...
-                seconds = duration.total_seconds()
-                logger.debug("build finished in %ds", seconds)
-                self.plugins_durations['dockerbuild'] = seconds
-            except Exception:
-                logger.exception("failed to save build duration")
-
-            self.build_logs = build_result.logs
-
-            self.build_failed = build_result.is_failed()
+                self.build_results = buildstep_runner.run()
```
twaugh: Now that this is a BuildResult instance, maybe name it 'build_result' again?

atomic_reactor/inner.py
```diff
@@ -250,18 +259,23 @@ def __init__(self, source, image, prebuild_plugins=None, prepublish_plugins=None
...
+        self.buildstep_plugins_conf = buildstep_plugins
         self.prepublish_plugins_conf = prepublish_plugins
         self.postbuild_plugins_conf = postbuild_plugins
         self.exit_plugins_conf = exit_plugins
         self.prebuild_results = {}
+        self.buildstep_result = {}
         self.postbuild_results = {}
         self.prepub_results = {}
         self.exit_results = {}
+        self.build_result = None
```
twaugh: Wonder if we really need to add this. Alternatively, if we add this, do we still need self.build_failed, self.build_logs, etc?

docs/plugins.md
```diff
@@ -70,6 +72,16 @@ All you need to do to accomplish this is to pass argument `--substitute` with va
...
 
 
+## Buidstep plugins
+
+
+Unlike other plugins, buildstep plugins have some differences.
+Buildstep plugins have to return BuildResult object.
+If no plugin specified 'docker_api' will be used by default.
+All buildstep plugins will have set `required` and `is_allowed_to_fail` to `false`.
+Buildstep plugins are run in order, and if one plugin successfully completes or fails following plugins won't be run anymore.
```
twaugh: Suggested re-wording:
```
Unlike other plugins, buildstep plugins have some differences. The `required` and `is_allowed_to_fail` properties will be set to `false` at runtime. Buildstep plugins are run in order, and if one plugin successfully completes or fails, the remaining buildstep plugins will not be attempted.

The run() method must return a BuildResult instance, or raise `InappropriateBuildStepError` to indicate the next buildstep plugin in the list should be attempted.
```

atomic_reactor/build.py
```diff
@@ -54,37 +54,31 @@ def _ensure_not_built(self):
...
-    def __init__(self, command_result, image_id=None):
-        """ when build fails, image_id is None """
-        self.command_result = command_result
+    def __init__(self, logs=None, fail_reason=None, image_id=None):
+        """
+        :param logs: iterable of log lines (without newlines)
+        :param fail_reason: str, description of failure or None if successful
+        :param image_id: str, ID of built container image
+        """
+        assert bool(fail_reason) != bool(image_id)  # must provide one, not both
```
vrutkovs: nitpick: add a more informative assert error, smth like 'Either of `fail_reason` or `image_id` should be provided, not both`

atomic_reactor/plugins/build_docker_api.py
```diff
@@ -0,0 +1,51 @@
...
+                                                     builder.image)
+
+        self.log.debug('build is submitted, waiting for it to finish')
+        command_result = wait_for_command(logs_gen)
+
+        if command_result.is_failed():
+            return BuildResult(logs=command_result.logs,
+                               fail_reason=command_result.error_detail)
+        else:
+            image_id = self.tasker.get_built_image_info()['Id']
```
vrutkovs: no such function, should be `builder.get_built_image_info()['Id']` probably

atomic_reactor/build.py
```diff
@@ -198,7 +196,6 @@ def get_built_image_info(self):
         :return dict
         """
         logger.info("getting information about built image '%s'", self.image)
-        self._ensure_is_built()
```
twaugh: Just spotted this -- why is it removed?

atomic_reactor/inner.py
```diff
@@ -250,31 +260,34 @@ def __init__(self, source, image, prebuild_plugins=None, prepublish_plugins=None
...
+        self.buildstep_plugins_conf = buildstep_plugins
         self.prepublish_plugins_conf = prepublish_plugins
         self.postbuild_plugins_conf = postbuild_plugins
         self.exit_plugins_conf = exit_plugins
         self.prebuild_results = {}
+        self.buildstep_result = {}
         self.postbuild_results = {}
         self.prepub_results = {}
         self.exit_results = {}
+        self.build_result = BuildResult(fail_reason="not built")
```
twaugh: This seems reasonable to me, yes.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -231,7 +231,7 @@ def run(self):
                                         '", "'.join(sorted(unknown_states)))
 
         rebuild = is_rebuild(self.workflow)
-        success = not self.workflow.build_failed
+        success = not self.workflow.build_result.is_failed()
```
twaugh: This should probably be `self.workflow.build_process_failed` anyway -- it cares about whether the build as a whole (plugins etc) has failed, rather than specifically the build step.

tests/plugins/test_koji_promote.py
```diff
@@ -232,7 +233,7 @@ def mock_environment(tmpdir, session=None, name=None,
     setattr(workflow, 'source', source)
     setattr(workflow.source, 'lg', X())
     setattr(workflow.source.lg, 'commit_id', '123456')
-    setattr(workflow, 'build_logs', ['docker build log - \u2018 \u2017 \u2019 \n'])
+#    setattr(workflow, 'build_logs', ['docker build log - \u2018 \u2017 \u2019 \n'])
```
twaugh: Delete the line instead of commenting it out.

tests/plugins/test_koji_promote.py
```diff
@@ -259,7 +260,10 @@ def mock_environment(tmpdir, session=None, name=None,
         fp.write('x' * 2**12)
         setattr(workflow, 'exported_image_sequence', [{'path': fp.name}])
 
-    setattr(workflow, 'build_failed', build_process_failed)
+    if build_process_failed:
+        workflow.build_result = BuildResult(logs=["not built"], fail_reason="not built")
+    else:
+        workflow.build_result = BuildResult(logs=["image built"], image_id="id1234")
```
twaugh: Please can we use the test logs we had before which included Unicode characters? I think it's important to test that.

atomic_reactor/plugins/exit_sendmail.py
twaugh: (But it's fine to keep what you have.)

atomic_reactor/build.py
rcerven: get_build_image_info is called from docker_api to get image_id, to create BuildResults,
so I can't get that method asking for something from future :)

tests/plugins/test_koji_promote.py
rcerven: sure

tests/plugins/test_koji_promote.py
rcerven: yop

atomic_reactor/build.py
twaugh: Ah! Got it.

atomic_reactor/plugin.py
```diff
@@ -219,17 +230,36 @@ def run(self, keep_going=False):
...
                 #   doesn't catch this and make PluginFailedException out of it in the end
                 #   (calling methods would then need to parse exception message to see if
                 #   AutoRebuildCanceledException was raised here)
                 raise
+            except InappropriateBuildStepError:
+                logger.debug('Build step %s is not appropriate', plugin_class.key)
+                # don't put None, in results for InappropriateBuildStepError
+                skip_response = True
+                if not buildstep_phase:
+                    raise
```
twaugh: This `raise` doesn't get test coverage, but it looks ok on visual inspection. ;)

rcerven: covered

vrutkovs:
Integration tests passed, this PR needs a rebase and optionally some more tests

atomic_reactor/build.py
```diff
@@ -54,37 +53,33 @@ def _ensure_not_built(self):
...
 
     @property
-    def image_id(self):
-        return None
+    def fail_reason(self):
+        return self._fail_reason
 
     def is_failed(self):
-        return True
+        return self._fail_reason is not None
```
lcarva: If the object is initialized with `fail_reason=False`, and a valid `image_id`, this method will incorrectly return `True`. Suggest modifying this to return `bool(self._fail_reason)`.

Any "falsy" value for `fail_reason` will hit this scenario.

```python
>>> BuildResult(fail_reason=None, image_id='spam').is_failed()
False # Good!

>>> BuildResult(fail_reason=True, image_id='spam').is_failed()
  File "atomic_reactor/build.py", line 64, in __init__
    "Either fail_reason or image_id should be provided, not both"
AssertionError: Either fail_reason or image_id should be provided, not both
Either fail_reason or image_id should be provided, not both
# Good! Error is expected.

>>> BuildResult(fail_reason=False, image_id='spam').is_failed()
True  # Uh oh! This is not right.
```

atomic_reactor/plugins/build_docker_api.py
```diff
@@ -0,0 +1,54 @@
...
+            built_image_info
+            image_id
+        """
+        builder = self.workflow.builder
+
+        logs_gen = self.tasker.build_image_from_path(builder.df_dir,
+                                                     builder.image)
+
+        self.log.debug('build is submitted, waiting for it to finish')
+#        for l in logs_gen:
```
lcarva: Let's remove this commented out code.

atomic_reactor/inner.py
```diff
@@ -343,31 +355,27 @@ def build_docker_image(self):
...
-            except Exception:
-                logger.exception("failed to save build duration")
-
-            self.build_logs = build_result.logs
-
-            self.build_failed = build_result.is_failed()
+                self.build_result = buildstep_runner.run()
+
+                if self.build_result.is_failed():
+                    self.builder.is_built = False
```
lcarva: Isn't setting to `False` here redundant? Once exception is raised the `except` block below will do the same.

atomic_reactor/plugin.py
```diff
@@ -354,6 +396,44 @@ def __init__(self, dt, workflow, plugins_conf, *args, **kwargs):
...
+class BuildStepPluginsRunner(BuildPluginsRunner):
+
+    def __init__(self, dt, workflow, plugin_conf, *args, **kwargs):
+        logger.info("initializing runner of build-step plugin")
+        self.plugins_results = workflow.buildstep_result
+
+        if not plugin_conf:
+            # if none buildstep plugins specified, fallback to docker api plugin
+            plugin_conf = [{'name': 'docker_api', 'is_allowed_to_fail': False}]
+        else:
```
lcarva: Don't think there should be an `else` here. If "falsy", just define `plugin_conf` as `[{'name': 'docker_api'}]` and let it run through normal execution. 

Right now, it won't set `required` to False as it would be if `plugin_conf` was "truthy".

atomic_reactor/plugins/build_docker_api.py
rcerven: ok

atomic_reactor/inner.py
rcerven: yeah in this case is redundant, but it is at least very readable

atomic_reactor/plugin.py
rcerven: logic here is quite different

if no plugins specified we will fallback to docker_api, required isn't specified because it is default in pluginsrunner

else is there for case there are buildstepplugins specified BUT we want to control their required & is_allowed_to_fail regardless what they had set

atomic_reactor/build.py
rcerven: well in docstring is specified, that both has to be str

anyway added another assert

twaugh: Would `assert fail_reason is None or image_id is None, "Provide either fail_reason or image_id, not both"` in the constructor cover what's needed? We don't ever treat either value as bool elsewhere, only decide whether it is None.

atomic_reactor/build.py
```diff
@@ -54,37 +53,35 @@ def _ensure_not_built(self):
...
-    def __init__(self, command_result, image_id=None):
-        """ when build fails, image_id is None """
-        self.command_result = command_result
+    def __init__(self, logs=None, fail_reason=None, image_id=None):
+        """
+        :param logs: iterable of log lines (without newlines)
+        :param fail_reason: str, description of failure or None if successful
+        :param image_id: str, ID of built container image
+        """
+        assert fail_reason == None or bool(fail_reason), \
```
twaugh: `== None` is better written as `is None`, but see https://github.com/projectatomic/atomic-reactor/pull/647/files#r102438121.

atomic_reactor/inner.py
```diff
@@ -343,31 +355,27 @@ def build_docker_image(self):
...
+                    logger.error('buildstep plugin failed: %s',
+                                 self.build_result.fail_reason)
+                    raise PluginFailedException('buildstep plugin failed')
+                else:
```
twaugh: `else:` after `raise` isn't needed.

atomic_reactor/inner.py
twaugh: I think it reduces understanding of the flow -- it makes it seem as though the exception *won't* handle it.

atomic_reactor/build.py
rcerven: >>> BuildResult(fail_reason=None, image_id=None).is_failed()
False

twaugh: OK, fair. Asserts looks fine to me then except for `== None`.

atomic_reactor/inner.py
rcerven: mmmm? that means it was successful

atomic_reactor/build.py
rcerven: done

atomic_reactor/inner.py
rcerven: done

atomic_reactor/inner.py
```diff
@@ -343,31 +355,26 @@ def build_docker_image(self):
...
 
-            start_time = datetime.datetime.now()
-            self.plugins_timestamps['dockerbuild'] = start_time.isoformat()
-
-            build_result = self.builder.build()
-
+            logger.info("running buildstep plugins")
+            buildstep_runner = BuildStepPluginsRunner(self.builder.tasker, self,
+                                                     self.buildstep_plugins_conf,
+                                                     plugin_files=self.plugin_files)
```
twaugh: atomic_reactor/inner.py:360:54: E128 continuation line under-indented for visual indent
atomic_reactor/inner.py:361:54: E128 continuation line under-indented for visual indent

atomic_reactor/plugin.py
```diff
@@ -41,6 +43,9 @@ class PluginFailedException(Exception):
 class BuildCanceledException(Exception):
     """Build was canceled"""
 
+class InappropriateBuildStepError(Exception):
```
twaugh: atomic_reactor/plugin.py:46:1: E302 expected 2 blank lines, found 1

atomic_reactor/plugins/build_docker_api.py
```diff
@@ -0,0 +1,52 @@
...
+        self.log.debug('build is submitted, waiting for it to finish')
+        command_result = wait_for_command(logs_gen)
+
+        if command_result.is_failed():
+            return BuildResult(logs=command_result.logs,
+                               fail_reason=command_result.error_detail)
+        else:
+            image_id = builder.get_built_image_info()['Id']
+            return BuildResult(logs=command_result.logs, image_id=image_id)
+
```
twaugh: atomic_reactor/plugins/build_docker_api.py:52:1: W391 blank line at end of file

tests/plugins/test_docker_api.py
```diff
@@ -0,0 +1,96 @@
...
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import unicode_literals
+
+import json
```
twaugh: tests/plugins/test_docker_api.py:11:1: F401 'json' imported but unused

tests/plugins/test_docker_api.py
```diff
@@ -0,0 +1,96 @@
...
+from dockerfile_parse import DockerfileParser
+
+from atomic_reactor.plugins.build_docker_api import DockerApiPlugin
```
twaugh: tests/plugins/test_docker_api.py:14:1: F401 'atomic_reactor.plugins.build_docker_api.DockerApiPlugin' imported but unused

tests/plugins/test_docker_api.py
```diff
@@ -0,0 +1,96 @@
...
+from atomic_reactor.plugin import PluginFailedException
+from atomic_reactor.build import InsideBuilder, BuildResult
+from atomic_reactor.util import ImageName, CommandResult
+from atomic_reactor.inner import DockerBuildWorkflow
+
+from tests.docker_mock import mock_docker
+from flexmock import flexmock
+import pytest
+from tests.constants import MOCK_SOURCE, SOURCE
```
twaugh: tests/plugins/test_docker_api.py:23:1: F401 'tests.constants.SOURCE' imported but unused

atomic_reactor/plugin.py
rcerven: done

twaugh:
Added a couple of new flake8 warnings.

atomic_reactor/inner.py
rcerven: done

tests/plugins/test_docker_api.py
rcerven: done

tests/plugins/test_docker_api.py
rcerven: done

tests/plugins/test_docker_api.py
rcerven: done

atomic_reactor/plugins/build_docker_api.py
rcerven: done

twaugh:
Re-confirmed my approval.

Probably needs to be smoke-tested in a buildroot before merge.

@lcarva, anything else to change?

atomic_reactor/build.py
lcarva: Verified changes fix the issue. Thanks!

atomic_reactor/inner.py
lcarva: I'm ok with leaving it as is.

atomic_reactor/plugin.py
lcarva: I don't think they're that different of use cases. For instance, if the user defines `plugin_conf` as `[{'name': 'docker_api'}]`, then it'll be modified to `[{'name': 'docker_api', 'is_allowed_to_fail': False, 'required': False}]` by the for-loop below. 

My point is that, there is at least one way that the fallback flow will result in a different plugin_conf for the same plugin. At this time, this doesn't seem to be a problem though. It just would've made the code better future proof.

I'm fine with leaving this is.

PR #650, opened 2017-02-16
Revert "integrate py.test with 'python setup.py test'"

vrutkovs:
This reverts commit 016176a7df37c62389071b0175f393697fc1bf57.

pytest-runner is not available in RHEL, so instead of 'python setup.py test' a 
different command to run tests is required - see travis.yml or soon-to-be-merged contribution guide

Related: #593, #556

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,0 @@
-[aliases]
-test=pytest
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,5 @@ setup(
     packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"]),
     install_requires=_install_requirements(),
     data_files=data_files.items(),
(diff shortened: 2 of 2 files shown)


twaugh:
...there isn't a contribution guide. :-)

integration-tests/test_pulp.py
```diff
@@ -41,7 +41,6 @@
 SOURCE = {"provider": "git", "uri": "git://example.com/test.git"}
 
 
-@pytest.mark.integration
```
twaugh: Why remove this?

pytest.ini
```diff
@@ -1,2 +0,0 @@
-[pytest]
-addopts = -m 'not integration'
```
twaugh: Why remove this?

vrutkovs: Fixed, this part is actually useful

PR #651, opened 2017-02-17
Add new reactor_config pre-build plugin

twaugh:


atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -0,0 +1,363 @@
...
+    """
+    try:
+        workspace = workflow.plugin_workspace[ReactorConfigPlugin.key]
+        return workspace[WORKSPACE_CONF_KEY]
+    except KeyError:
+        # The plugin did not run or was not successful: use defaults
+        return ReactorConfig()
+
+
+class ConfigBuilder(object):
```
lcarva: If we want to validate the configuration against a schema, can we use an existing library for it? Something like: https://pypi.python.org/pypi/pykwalify/

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -0,0 +1,363 @@
...
+    REQUIRED = [
+        # clusters.*.name is required (but 'clusters' is not)
+        ('clusters', None, ['name']),
+    ]
+
+    def __init__(self, config_path=None, log=None):
+        self.log = log or logging.getLogger(__name__)
+
+        if config_path is None:
+            self.conf = {}
```
lcarva: I think for future proofing the code, it might be beneficial to also validate this default conf.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -0,0 +1,363 @@
...
+            for error in errors:
+                self.log.error(error)
+
+            raise RuntimeError('config validation failed')
+
+        self.conf = conf
+
+    def get_enabled_clusters_for_platform(self, platform):
+        clusters = self.conf.get(self.CLUSTERS_KEY, {}).get(platform, [])
+        cluster_configs = [ClusterConfig(**cluster) for cluster in clusters]
```
lcarva: Would it be better to only create these `ClusterConfig` objects once? Seems that the use case is for  `get_enabled_clusters_for_platform` to be called for each platform.

atomic_reactor/plugins/pre_reactor_config.py
twaugh: The reason for this class is two-fold:
- yes, to validate the data object
- but also to convert ambiguous data (such as true/false or numbers) so it matches the expected types (which might be str)

Do you know of any way of doing the second?

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -0,0 +1,363 @@
...
+    Does not fail
+
+    :return: ReactorConfig instance
+    """
+    try:
+        workspace = workflow.plugin_workspace[ReactorConfigPlugin.key]
+        return workspace[WORKSPACE_CONF_KEY]
+    except KeyError:
+        # The plugin did not run or was not successful: use defaults
+        return ReactorConfig()
```
twaugh: Yes, fixed.

atomic_reactor/plugins/pre_reactor_config.py
twaugh: Yes, fixed.

atomic_reactor/plugins/pre_reactor_config.py
twaugh: Yes, fixed.

atomic_reactor/plugins/pre_reactor_config.py
lcarva: Usually that's what Json and YAML are for. If the user creates a config that uses the incorrect type, and validation fails, I think that should be treated as an error. 

IMO, if the config format used allows different types, we should leverage, and enforce, that. This is different, for instance, than INI files via python's ConfigParser which pretty much treats all values as strings.

vrutkovs: Is there a reason to use YAML here instead of JSON? It feels weird to use some part of configuration in JSON and the other one in YAML

twaugh: JSON is great for transferring a data object from one program to another. It's picky in its syntax, though, and doesn't allow comments.

YAML is easier to write and allows comments.

This file is likely to be hand-edited from time to time. It would quite likely benefit from comments.

Other formats also allow comments, for example INI, but I wanted something in common use that provided structured data.

twaugh:
I'm planning to remove ConfigBuilder and instead perform validation by converting the YAML to JSON and validating it against a JSON schema. This is what e.g. docker-compose does.

Then it's up to the user to avoid ambiguity for value types.

atomic_reactor/plugins/pre_reactor_config.py
vrutkovs: >YAML is easier to write and allows comments.

That's a good point, lets stick to yaml then.

Although comments are possible in json (as any other javascript code) python's interpreter is strict and doesn't allow those

twaugh:
Updated. Please take another look.

twaugh:
I hope there isn't anything else left to do.

Open questions:
- Should additional properties be allowed within a cluster configuration? Currently they are not, so if we later to decide to add some new keyword alongside 'name', 'max_concurrent_builds', and 'enabled', this parser will fail to load it.
- Should we have a 'version' keyword at the top-level? If so should it be mandatory or optional?

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -0,0 +1,155 @@
...
+        try:
+            resource = resource_stream('atomic_reactor', 'schemas/config.json')
+            config_schema = codecs.getreader('utf-8')(resource)
+        except (IOError, TypeError):
+            self.log.error("unable to extract JSON schema, cannot validate")
+        else:
+            try:
+                schema = json.load(config_schema)
+            except ValueError:
+                self.log.error("unable to decode JSON schema, cannot validate")
```
lcarva: Should the exception be re-raised here?

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -0,0 +1,155 @@
...
+        self.log.info("reading config from %s", self.config_path)
+        with open(self.config_path) as fp:
+            conf = yaml.safe_load(fp)
+
+        # Validate against JSON schema
+        try:
+            resource = resource_stream('atomic_reactor', 'schemas/config.json')
+            config_schema = codecs.getreader('utf-8')(resource)
+        except (IOError, TypeError):
+            self.log.error("unable to extract JSON schema, cannot validate")
```
lcarva: Should the exception be re-raised here?

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -0,0 +1,155 @@
...
+                schema = json.load(config_schema)
+            except ValueError:
+                self.log.error("unable to decode JSON schema, cannot validate")
+            else:
+                validator = jsonschema.Draft4Validator(schema=schema)
+                try:
+                    jsonschema.Draft4Validator.check_schema(schema)
+                    validator.validate(conf)
+                except jsonschema.SchemaError:
+                    self.log.error("invalid schema, cannot validate")
```
lcarva: Should the exception be re-raised here?

docs/config.md
```diff
@@ -0,0 +1,28 @@
...
+  - name: worker01
+    max_concurrent_builds: 4
+    enabled: true
+  - name: worker02
+    max_concurrent_builds: 8
+    enabled: false
+  - name: worker03
+```
+
+In this example builds for the x86_64 platform can be sent to worker01 if it has fewer than 4 active worker builds, or worker03.
```
lcarva: `worker03` doesn't provide the `max_concurrent_builds`, what's the expected default behavior? Unlimited?

tests/plugins/test_reactor_config.py
```diff
@@ -0,0 +1,224 @@
...
+        same_conf = get_config(workflow)
+        assert conf is same_conf
+
+    def test_no_schema_resource(self, tmpdir, caplog):
+        class FakeProvider(object):
+            def get_resource_stream(self, pkg, rsc):
+                raise IOError
+
+        (flexmock(pkg_resources)
+            .should_receive('get_provider')
```
lcarva: `with_args` ?

atomic-reactor.spec
```diff
@@ -166,8 +168,10 @@ Requires:       python3-requests
 Requires:       python3-setuptools
 Requires:       python3-dockerfile-parse >= 0.0.5
 Requires:       python3-docker-squash >= 1.0.0-0.3
+Requires:       python3-jsonschema
 # Due to CopyBuiltImageToNFSPlugin, might be moved to subpackage later.
 Requires:       nfs-utils
+Requires:	python3-PyYAML
```
twaugh: Fixed.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -0,0 +1,155 @@
...
+        workflow.plugin_workspace[ReactorConfigPlugin.key] = workspace
+        return conf
+
+
+class ClusterConfig(object):
+    """
+    Configuration relating to a particular cluster
+    """
+
+    def __init__(self, name, max_concurrent_builds=None, enabled=True):
```
twaugh: No, must be specified. Fixed.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class ReactorConfig(object):
+    """
+    Class to parse the atomic-reactor configuration file
+    """
+
+    CLUSTERS_KEY = 'clusters'
+
+    def __init__(self, config, log=None):
```
twaugh: I've changed that now. The DEFAULT_CONFIG is now an attribute of the ReactorConfig class.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -0,0 +1,155 @@
...
+        self.conf = config or {}
+        self.log = log or logging.getLogger(__name__)
```
twaugh: It isn't. Removed.

atomic_reactor/plugins/pre_reactor_config.py
twaugh: In this situation, the schema itself could not be retrieved. Yes, I guess that should be fatal, will fix.

twaugh: Fixed.

atomic_reactor/plugins/pre_reactor_config.py
twaugh: Fixed.

atomic_reactor/plugins/pre_reactor_config.py
twaugh: Fixed.

docs/config.md
twaugh: No, it's a required keyword. Fixed.

tests/plugins/test_reactor_config.py
twaugh: It's a class instance that it gets. I could add `.with_args(object)`? This is pkg_resources internals really, but it won't let me flexmock `get_resource_stream()` which is all I care about mocking.

lcarva: Fair enough

lcarva:
> Should additional properties be allowed within a cluster configuration? Currently they are not, so if we later to decide to add some new keyword alongside 'name', 'max_concurrent_builds', and 'enabled', this parser will fail to load it.

The benefit of having this restriction is that if there's a typo in an optional property, it'll be known immediately. 

> Should we have a 'version' keyword at the top-level? If so should it be mandatory or optional?

If we have one, I think it should be mandatory. I can't think of an undesired behavior for versioning. My vote is yes, let's add it :)

twaugh:
New mandatory `version` keyword added, value must be 1.

Note: other unknown keys will still be permitted at the top-level, alongside 'clusters'.

tests/plugins/test_reactor_config.py
twaugh: I've added a comment to make this clearer.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -0,0 +1,157 @@
...
+    VERSION_KEY = 'version'
+    CLUSTERS_KEY = 'clusters'
+
+    DEFAULT_CONFIG = {'version': 1}
+
+    def __init__(self, config=None):
+        self.conf = config or self.DEFAULT_CONFIG
+
+        version = self.conf[self.VERSION_KEY]
+        if version != 1:
```
rcerven: why not use here version from DEFAULT_CONFIG?

twaugh: No particular reason. Note that once we have more than a single version, this code might look more like:
```
if version == 1:
    handle_v1()
elif version == 2:
    handle_v2()
else:
    raise
```

rcerven: ahh, I was thinking that DEFAULT_CONFIG might used as well asthe "latest" version allowed

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -0,0 +1,157 @@
...
+
+class ReactorConfig(object):
+    """
+    Class to parse the atomic-reactor configuration file
+    """
+
+    VERSION_KEY = 'version'
+    CLUSTERS_KEY = 'clusters'
+
+    DEFAULT_CONFIG = {'version': 1}
```
rcerven: maybe use ReactorConfig.VERSION_KEY ?

twaugh: That's a good idea, yes.

twaugh: https://github.com/projectatomic/atomic-reactor/pull/654

PR #653, opened 2017-02-21
Make reactor_config default to reading config.yaml

twaugh:
It is easy for osbs-client to pass the path at which the Kubernetes secret is mounted but harder for it to pass the filename within the secret representing the config.

Add a 'basename' parameter to reactor_config which defaults to 'config.yaml' to handle this.

### DIFF
--- a/atomic_reactor/plugins/pre_reactor_config.py
+++ b/atomic_reactor/plugins/pre_reactor_config.py
@@ -11,6 +11,7 @@ from atomic_reactor.plugin import PreBuildPlugin
 import codecs
 import json
 import jsonschema
+import os
 from pkg_resources import resource_stream
 import yaml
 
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -108,8 +111,9 @@ def run(self):
         Store in workflow workspace for later retrieval.
         """
 
-        self.log.info("reading config from %s", self.config_path)
-        with open(self.config_path) as fp:
+        config_filename = os.path.join(self.config_path, self.basename)
+        with open(config_filename) as fp:
+            self.log.info("reading config from %s", config_filename)
```
lcarva: Log before opening the file? Might facilitate debugging.

twaugh: Fixed.

PR #654, opened 2017-02-22
Put config key names in new ReactorConfigKeys class

twaugh:
This allows the default configuration to be specified using these keys.

### DIFF
--- a/atomic_reactor/plugins/pre_reactor_config.py
+++ b/atomic_reactor/plugins/pre_reactor_config.py
@@ -50,26 +50,41 @@ class ClusterConfig(object):
         self.enabled = enabled
 
 
-class ReactorConfig(object):
+class ReactorConfigKeys(object):
     """
-    Class to parse the atomic-reactor configuration file
+    Symbolic names to use for the key names in the configuration file
+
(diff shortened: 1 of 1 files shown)


twaugh:
cc @csomh

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -50,26 +50,28 @@ def __init__(self, name, max_concurrent_builds, enabled=True):
         self.enabled = enabled
 
 
+class ReactorConfigKeys(object):
```
csomh: I assume the intention is to be able to use this in other/upcoming plugins. In this case, maybe a one-line doc-string would be nice - what is this, and what is the intention (even though at this point this is pretty obvious).

twaugh: Really it's for use by the ReactorConfig class, which is concerned with fetching values for specific keys in the configuration file.

Other modules are expected to use the ReactorConfig instance returned by get_config(), which can provide methods for making queries against the configuration file.

csomh: I see, then it's fine this way, too.

PR #655, opened 2017-02-22
Some improvements related to the buildstep phase

twaugh:
- Only log buildstep failures once
- Improve error logging if inspecting the built image fails
- The docker_api fallback only applies when the buildstep_plugins key is not present at all

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -364,15 +364,14 @@ class DockerBuildWorkflow(object):
 
                 if self.build_result.is_failed():
                     raise PluginFailedException(self.build_result.fail_reason)
-                
-                self.builder.is_built = True
-                self.builder.image_id = self.build_result.image_id
-                self.built_image_inspect = self.builder.inspect_built_image()
-
             except PluginFailedException as ex:
                 self.builder.is_built = False
                 logger.error('buildstep plugin failed: %s', ex)
                 raise
+
+            self.builder.is_built = True
+            self.builder.image_id = self.build_result.image_id
+            self.built_image_inspect = self.builder.inspect_built_image()
 
             # run prepublish plugins
             prepublish_runner = PrePublishPluginsRunner(self.builder.tasker, self, self.prepublish_plugins_conf,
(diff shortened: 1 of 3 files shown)


atomic_reactor/inner.py
```diff
@@ -394,6 +391,9 @@ def build_docker_image(self):
                 raise
 
             return self.build_result
+        except Exception as ex:
+            logger.debug("caught exception (%r) so running exit plugins", ex)
```
lcarva: Is this `except` block only for logging the error? (Not against it, just trying to understand it)

atomic_reactor/plugin.py
```diff
@@ -408,8 +408,8 @@ def __init__(self, dt, workflow, plugin_conf, *args, **kwargs):
         logger.info("initializing runner of build-step plugin")
         self.plugins_results = workflow.buildstep_result
 
-        if not plugin_conf:
-            # if none buildstep plugins specified, fallback to docker api plugin
+        if plugin_conf is None:
```
lcarva: We want this because we may want to explicitly not have any buildstep_plugin executed?

atomic_reactor/inner.py
twaugh: That's right. If an exit plugin later raises its own exception, we'll never find out what the original problem was.

atomic_reactor/plugin.py
rcerven: but what if we provide empty list?? it isn't None, so 'not plugin_conf' covered that

twaugh: Just that if the build JSON explicitly says not to run any buildstep plugins, we shouldn't (and this will fail).

It's minor: in practice it would only happen if there was an osbs-client misconfiguration or something.

twaugh: Acceptance criteria for this say that an empty list causes failure.

rcerven:
+1

lcarva:
@twaugh holding off on the merge until you confirm, or squashes, commits.

twaugh:
They are all separate changes I think, so ought to be in separate commits.

twaugh:
I'll test it out and update with results.

twaugh:
Works fine.

PR #656, opened 2017-02-24
Add new orchestrate_build plugin

lcarva:
Creating this PR for early feedback. Please keep in mind the following known missing pieces:

- ~~Unit tests not yet included~~
- ~~Doc files not updated~~
- ~~Plugin Class docstring not added~~

All pieces are in place at this moment.

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -370,8 +370,9 @@ class DockerBuildWorkflow(object):
                 raise
 
             self.builder.is_built = True
-            self.builder.image_id = self.build_result.image_id
-            self.built_image_inspect = self.builder.inspect_built_image()
+            if self.build_result.is_image_available():
+                self.builder.image_id = self.build_result.image_id
+                self.built_image_inspect = self.builder.inspect_built_image()
 
             # run prepublish plugins
             prepublish_runner = PrePublishPluginsRunner(self.builder.tasker, self, self.prepublish_plugins_conf,
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -1,0 +1,255 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+from __future__ import unicode_literals, division
+
+from collections import namedtuple
+from copy import deepcopy
+from multiprocessing.pool import ThreadPool
+
+import json
+import os
+
+from atomic_reactor.build import BuildResult
+from atomic_reactor.plugin import BuildStepPlugin
+from atomic_reactor.plugins.pre_reactor_config import get_config
+from atomic_reactor.util import get_preferred_label, df_parser
+from osbs.api import OSBS
+from osbs.conf import Configuration
+from osbs.constants import BUILD_FINISHED_STATES
+
+
+ClusterInfo = namedtuple('ClusterInfo', ('cluster', 'platform', 'osbs', 'load'))
+
+
+class WorkerBuildInfo(object):
+
+    def __init__(self, build, cluster_info):
+        self.build = build
+        self.cluster = cluster_info.cluster
+        self.osbs = cluster_info.osbs
+        self.platform = cluster_info.platform
+
+        self.monitor_exception = None
+
+    @property
+    def name(self):
+        return self.build.get_build_name() if self.build else 'N/A'
+
+    def wait_to_finish(self):
+        self.build = self.osbs.wait_for_build_to_finish(self.name)
+        return self.build
+
+    def watch_logs(self, logger):
+        for line in self.osbs.get_build_logs(self.name, follow=True):
+            # TODO: This is a little clunky:
+            # 2017-02-24 14:22:51,472 - atomic_reactor.plugins.orchestrate_build - INFO - x86_64 - 2017-02-24 14:22:51,314 - atomic_reactor.plugin - INFO - <worker build log message>
+            logger.info('%s - %s', self.platform, line)
+
+    def get_annotations(self):
+        build_annotations = self.build.get_annotations() or {}
+        return {
+            'build': {
+                'cluster-url': self.osbs.os_conf.get_openshift_base_uri(),
+                'namespace': self.osbs.os_conf.get_namespace(),
+                'build-name': self.name,
+            },
+            'digests': json.loads(
+                build_annotations.get('digests', '[]')),
+            'plugins-metadata': json.loads(
+                build_annotations.get('plugins-metadata', '{}')),
+        }
+
+    def get_fail_reason(self):
+        if not self.build:
+            return {'general': 'build not started'}
+        build_annotations = self.build.get_annotations() or {}
+        metadata = json.loads(build_annotations.get('plugins-metadata', '{}'))
+        fail_reason = metadata.get('errors', {})
+
+        if self.monitor_exception:
+            fail_reason['general'] = repr(self.monitor_exception)
+
+        return fail_reason
+
+    def cancel_build(self):
+        if self.build and not self.build.is_finished():
+            self.osbs.cancel_build(self.name)
+
+
+class OrchestrateBuildPlugin(BuildStepPlugin):
+    """
+    Start and monitor worker builds for each platform
+
+    This plugin will find the best suited worker cluster to
+    be used for each platform. It does so by calculating the
+    current load of active builds on each cluster and choosing
+    the one with smallest load.
+
+    The list of available worker clusters is retrieved by fetching
+    the result provided by reactor_config plugin.
+
+    If any of the worker builds fail, this plugin will return a
+    failed BuildResult. Although, it does wait for all worker builds
+    to complete in any case.
+
+    If all worker builds succeed, then this plugin returns a
+    successful BuildResult, but with a remote image result. The
+    image is built in the worker builds which is likely a different
+    host than the one running this build. This means that the local
+    docker daemon has no knowledge of the built image.
+    """
+
+    EXCLUDE_FILENAME = 'exclude-platform'
+
+    key = 'orchestrate_build'
+
+    def __init__(self, tasker, workflow, platforms, build_kwargs,
+                 osbs_client_config=None):
+        """
+        constructor
+
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param platforms: list<str>, platforms to build
+        :param build_kwargs: dict, keyword arguments for starting worker builds
+        :param osbs_client_config: str, path to directory containing osbs.conf
+        """
+        super(OrchestrateBuildPlugin, self).__init__(tasker, workflow)
+        self.platforms = set(platforms)
+        self.build_kwargs = build_kwargs
+        self.osbs_client_config = osbs_client_config
+
+        self.worker_builds = []
+
+    def get_excluded_platforms(self):
+        df_dir = self.workflow.source.get_dockerfile_path()[1]
+        exclude_platforms = set()
+        exclude_path = os.path.join(df_dir, self.EXCLUDE_FILENAME)
+        if os.path.exists(exclude_path):
+            with open(exclude_path) as f:
+                for platform in f:
+                    platform = platform.strip()
+                    if not platform:
+                        continue
+                    exclude_platforms.add(platform)
+
+        return exclude_platforms
+
+    def get_platforms(self):
+        return self.platforms - self.get_excluded_platforms()
+
+    def get_current_builds(self, osbs):
+        field_selector = ','.join(['status!={status}'.format(status=status.capitalize())
+                                   for status in BUILD_FINISHED_STATES])
+        return len(osbs.list_builds(field_selector=field_selector))
+
+    def get_cluster_info(self, cluster, platform):
+        kwargs = {'conf_section': cluster.name}
+        if self.osbs_client_config:
+            kwargs['conf_file'] = os.path.join(self.osbs_client_config, 'osbs.conf')
+        conf = Configuration(**kwargs)
+        osbs = OSBS(conf, conf)
+        current_builds = self.get_current_builds(osbs)
+        load = current_builds / cluster.max_concurrent_builds
+        return ClusterInfo(cluster, platform, osbs, load)
+
+    def choose_cluster(self, platform):
+        config = get_config(self.workflow)
+        clusters = [self.get_cluster_info(cluster, platform) for cluster in
+                    config.get_enabled_clusters_for_platform(platform)]
+
+        if not clusters:
+            raise RuntimeError('No clusters found for platform {}!'
+                               .format(platform))
+
+        selected = min(clusters, key=lambda c: c.load)
+        self.log.info('platform %s will use cluster %s',
+                      platform, selected.cluster.name)
+        return selected
+
+    def get_release(self):
+        labels = df_parser(self.workflow.builder.df_path, workflow=self.workflow).labels
+        return get_preferred_label(labels, 'release')
+
+    def get_worker_build_kwargs(self, release, platform):
+        build_kwargs = deepcopy(self.build_kwargs)
+
+        build_kwargs.pop('architecture', None)
+
+        build_kwargs['release'] = release
+        build_kwargs['platform'] = platform
+
+        return build_kwargs
+
+    def do_worker_build(self, release, cluster_info):
+        build = None
+        try:
+            kwargs = self.get_worker_build_kwargs(release, cluster_info.platform)
+            build = cluster_info.osbs.create_worker_build(**kwargs)
+        except Exception:
+            self.log.exception('%s - failed to create worker build',
+                               cluster_info.platform)
+
+        build_info = WorkerBuildInfo(build=build, cluster_info=cluster_info)
+        self.worker_builds.append(build_info)
+
+        if build_info.build:
+            try:
+                self.log.info('%s - created build %s', cluster_info.platform,
+                              build_info.name)
+                build_info.watch_logs(self.log)
+                build_info.wait_to_finish()
+            except Exception as e:
+                build_info.monitor_exception = e
+                self.log.exception('%s - failed to monitor worker build',
+                                   cluster_info.platform)
+
+    def run(self):
+        release = self.get_release()
+        platforms = self.get_platforms()
+
+        thread_pool = ThreadPool(len(platforms))
+        result = thread_pool.map_async(
+            lambda cluster_info: self.do_worker_build(release, cluster_info),
+            [self.choose_cluster(platform) for platform in platforms]
+        )
+
+        try:
+            while not result.ready():
+                # The wait call is a blocking call which prevents signals
+                # from being processed. Wait for short intervals instead
+                # of a single long interval, so build cancellation can
+                # be handled virtually immediately.
+                result.wait(1)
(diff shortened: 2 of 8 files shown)


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -0,0 +1,189 @@
...
+        for cluster_info in clusters:
+            kwargs = self.get_worker_build_kwargs(release, cluster_info.platform)
+            worker_build = cluster_info.osbs.create_worker_build(**kwargs)
+            worker_builds.append(
+                WorkerBuildInfo(build=worker_build, cluster_info=cluster_info))
+
+        for build_info in worker_builds:
+            build_info.wait_to_schedule()
+
+        thread_pool = ThreadPool(len(worker_builds))
```
twaugh: Can we start threads any earlier than this?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -0,0 +1,189 @@
...
+        for build_info in worker_builds:
+            annotations['worker-builds'][build_info.platform] = build_info.get_annotations()
+
+            if build_info.build.is_failed():
+                self.log.error('worker build %s failed on %s',
+                               build_info.name, build_info.cluster.name)
+                fail_reasons[build_info.platform] = build_info.get_fail_reason()
+
+        if fail_reasons:
+            return BuildResult(fail_reason=json.dumps(fail_reasons))
```
twaugh: When we create the BuildResult we should also provide `logs` - this is intended to end up in `build.log` in the Koji build.

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: I was just contemplating that :)

We could have threads for handling the lifetime of each worker build. It might also make it easier to make sure that we wait for all worker builds to complete. Right now, for instance, if the second worker build fails to be started, this plugin will quit, leaving an unfinished worker build.

twaugh: Also worth contemplating: how will cancellation be handled?

lcarva: Modified so that the life cycle of each worker build is handled by its own thread. This has the benefit of avoid an overall failure if one of the builds fails to be monitored properly.

Using `map_async` with short wait intervals to ensure build cancellation is handled immediately. Verified that when atomic_reactor receives SIGTERM, worker builds are canceled properly.

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: So the logs here are the full atomic-reactor logs for each worker build. Do we want that to be the build.log? Not sure if there's a straight forward way of grabbing just the docker build logs.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -0,0 +1,224 @@
...
+                # The wait call is a blocking call which prevents signals
+                # from being processed. Wait for short intervals instead
+                # of a single long interval, so build cancellation can
+                # be handled virtually immediately.
+                result.wait(1)
+        # Always clean up worker builds on any error to avoid
+        # runaway worker builds (includes orchestrator build cancellation)
+        except Exception:
+            self.log.info('build cancelled, cancelling worker builds')
+            for build_info in self.worker_builds:
```
lcarva: Since there's a limited amount of time between receiving SIGTERM and SIGKILL, it's probably a good idea to perform the cancellations in parallel, yes.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -0,0 +1,224 @@
...
+                    exclude_platforms.add(platform)
+
+        return exclude_platforms
+
+    def get_platforms(self):
+        return self.platforms - self.get_excluded_platforms()
+
+    def get_current_builds(self, osbs):
+        field_selector = ','.join(['status!={status}'.format(status=status.capitalize())
+                                   for status in BUILD_FINISHED_STATES])
```
lcarva: We could apply some labels when creating the worker build?

I'm thinking now that this is actually needed for using a cluster for both orchestrator and worker builds. If we count all current builds together, the computed load will, likely, always be higher in the orchestrator cluster, which means it won't be chosen! This, of course, depends on the `max_concurrent_builds` setting for each cluster.

Having said that, it seems beyond the scope of this change. We need to have changes in osbs-client first in order to leverage it in this plugin. How about we focus this PR on adding ability to orchestrate builds, and create a new story for allowing orchestrator cluster to also be used for worker builds?

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: done

atomic_reactor/plugins/build_orchestrate_build.py
twaugh: I'm starting to wonder if the worker builds ought to upload their own buildstep logs (`build-$platform.log`) to Koji, as well as the image archive.

atomic_reactor/plugins/build_orchestrate_build.py
twaugh: Agreed.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -0,0 +1,219 @@
...
+        try:
+            while True:
+                # The wait call is a blocking call which prevents signals
+                # from being processed. Wait for short intervals instead
+                # of a single long interval, so build cancellation can
+                # be handled virtually immediately.
+                result.wait(1)
+        # Always clean up worker builds on any error to avoid
+        # runaway worker builds (includes orchestrator build cancellation)
+        except Exception:
```
twaugh: Question: what actually happens to the worker threads at this point? They're still going?

lcarva: Yes. They're still running in background. Perhaps we should do another `result.wait()` here without timeout to make sure all logs are fetched, and cleanup at worker builds are finished. (I didn't see the need to do this in my testing though, but it's probably still worth doing it)

twaugh: I think it would be nicer to do that, yes. For one thing, we'd like to avoid another build for the same image being submitted while the worker builds are still cleaning up -- not sure if we really get a chance to stop that here though.

lcarva: Done.

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: yes! I think this is the cleanest solution.

lcarva: Currently the worker build will do this. Think we just need to worry about it when multi arch is tackled. Thoughts?

twaugh: Yes, agreed. For "just" orchestration, the code is fine as-is.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -0,0 +1,254 @@
...
+        return build_kwargs
+
+    def do_worker_build(self, release, cluster_info):
+        build = None
+        try:
+            kwargs = self.get_worker_build_kwargs(release, cluster_info.platform)
+            build = cluster_info.osbs.create_worker_build(**kwargs)
+        except Exception:
+            self.log.exception('%s - failed to create worker build',
+                               cluster_info.platform)
```
lcarva: `log.exception` does that.

twaugh:
Is it possible to get test coverage for cancellation and for when a requested platform has no cluster defined?

atomic_reactor/plugins/build_orchestrate_build.py
twaugh: Didn't know that one, nice.

atomic_reactor/inner.py
```diff
@@ -370,8 +370,9 @@ def build_docker_image(self):
                 raise
 
             self.builder.is_built = True
-            self.builder.image_id = self.build_result.image_id
-            self.built_image_inspect = self.builder.inspect_built_image()
+            if self.build_result.is_image_available():
+                self.builder.image_id = self.build_result.image_id
+                self.built_image_inspect = self.builder.inspect_built_image()
```
lcarva: There are quite a few plugins that read from these values, but none of them should actually be used with orchestrate_build plugin. Not sure what's the best way to handle plugin compatibility.

twaugh: OK, that's fine.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -0,0 +1,254 @@
...
+        return selected
+
+    def get_release(self):
+        labels = df_parser(self.workflow.builder.df_path, workflow=self.workflow).labels
+        return get_preferred_label(labels, 'release')
+
+    def get_worker_build_kwargs(self, release, platform):
+        build_kwargs = deepcopy(self.build_kwargs)
+
+        build_kwargs.pop('architecture', None)
```
lcarva: Yes. I figured we could start off worker builds without support for this. Can remove this line if this should be done at a later time instead.

twaugh: Let's leave it as it is, just wanted to be sure it was intended.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -0,0 +1,254 @@
...
+from atomic_reactor.plugin import BuildStepPlugin
+from atomic_reactor.plugins.pre_reactor_config import get_config
+from atomic_reactor.util import get_preferred_label, df_parser
+from osbs.api import OSBS
+from osbs.conf import Configuration
+from osbs.constants import BUILD_FINISHED_STATES
+from osbs.exceptions import OsbsException
+
+import json
+import os
```
lcarva: Moved.

tests/plugins/test_orchestrate_build.py
```diff
@@ -0,0 +1,352 @@
...
+    def __init__(self, tmpdir):
+        tmpdir = str(tmpdir)
+        self.dockerfile_path = os.path.join(tmpdir, 'Dockerfile')
+        self.path = tmpdir
+
+    def get_dockerfile_path(self):
+        return self.dockerfile_path, self.path
+
+    def get_vcs_info(self):
+        return VcsInfo('git', DOCKERFILE_GIT, DOCKERFILE_SHA1)
```
lcarva: Somewhere in my conscious. Removed `get_vcs_info` method as it's not needed.

lcarva:
Additional tests added as requested.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -0,0 +1,254 @@
...
+                # be handled virtually immediately.
+                result.wait(1)
+        # Always clean up worker builds on any error to avoid
+        # runaway worker builds (includes orchestrator build cancellation)
+        except Exception:
+            self.log.info('build cancelled, cancelling worker builds')
+            if self.worker_builds:
+                ThreadPool(len(self.worker_builds)).map(
+                    lambda bi: bi.cancel_build(), self.worker_builds)
+            while not result.ready():
```
twaugh: Is `result` the right thing at this point? We never assigned the result of the map() call to anything.

Since it's map() we're using anyway, I guess we don't need this polling loop in any case?

twaugh: Ah, no, I see, we're waiting for the original threads to finish.

lcarva: There are two thread pools being used here. The first one is for running worker builds, while the second one is for cancelling them.

The first one is asynchronous which returns a result. The second one, blocks until all threads are finished. 

So, what this is doing is cancelling all worker builds, wait for cancellation of each, then wait for the worker builds to end.

twaugh: Something odd seems to be going on here. There's no coverage for this polling loop or for the raise after it.

lcarva: The mock on `AsyncResult.wait` seems to be interfering with the second thread pool. The `BuildCanceledException` is being raised twice. Any suggestions on how to improve this?

Tried something like this, but exception is still raised twice (perhaps a flexmock oddity with threads?) :
```
(flexmock(AsyncResult).should_receive('wait')                               
  .and_raise(BuildCanceledException)                                      
  .and_return(None))
```

twaugh: Each new instance will raise an exception. Maybe new_instances() can help?

lcarva: No luck :(
Tried something like this:
```
(flexmock(AsyncResult).new_instances(MockAsyncResult, AsyncResult))
```
Which actually causes the worker threads to not execute...

twaugh: Try this: https://github.com/projectatomic/atomic-reactor/commit/97a019c00b1f7aa9aaeb64a186db23a9a805897e

This gives me full coverage for build_orchestrate_build.py locally.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -0,0 +1,254 @@
...
+        return self.platforms - self.get_excluded_platforms()
+
+    def get_current_builds(self, osbs):
+        field_selector = ','.join(['status!={status}'.format(status=status.capitalize())
+                                   for status in BUILD_FINISHED_STATES])
+        return len(osbs.list_builds(field_selector=field_selector))
+
+    def get_cluster_info(self, cluster, platform):
+        conf_file = os.path.join(self.osbs_client_config, 'osbs.conf')
+        osbs = OSBS(Configuration(conf_file=conf_file, conf_section=cluster.name),
```
vrutkovs: Nitpick: no need to create configuration twice here

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -0,0 +1,254 @@
...
+            raise RuntimeError('No clusters found for platform {}!'
+                               .format(platform))
+
+        selected = min(clusters, key=lambda c: c.load)
+        self.log.info('platform %s will use cluster %s',
+                      platform, selected.cluster.name)
+        return selected
+
+    def get_release(self):
+        labels = df_parser(self.workflow.builder.df_path, workflow=self.workflow).labels
```
vrutkovs: This would require pulling a base image on orchestrator build, same as in https://github.com/projectatomic/atomic-reactor/pull/657, right?

lcarva: Yes! In this case, we actually care about the value of the release label, including env substitution from parent's env, as used by Fedora. Closed the other PR, as we need to pull the base image even for orchestrator build. Thank you for catching this.

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: Changed.

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: Clever! Cherry-picked commit.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -0,0 +1,254 @@
...
+import os
+
+from atomic_reactor.build import BuildResult
+from atomic_reactor.plugin import BuildStepPlugin
+from atomic_reactor.plugins.pre_reactor_config import get_config
+from atomic_reactor.util import get_preferred_label, df_parser
+from osbs.api import OSBS
+from osbs.conf import Configuration
+from osbs.constants import BUILD_FINISHED_STATES
+from osbs.exceptions import OsbsException
```
twaugh: Not used.

lcarva: Removed.

lcarva:
Made `osbs_client_config` config param optional.

atomic_reactor/plugins/build_orchestrate_build.py
twaugh: Wait, why do we care about environment variable substitution here?

Also if we do want to pull the parent image, we'll also need to remove it afterwards. I'd really like to avoid the need to pull it in the orchestrator build though.

twaugh: Oh, and eventually we'll need to know the real value of that label in order to know how to import the image to Koji. OK.

Is there any other reason to pull the parent image, or is it only to find out what the parent image's environment variables are?

lcarva: That's the only reason I can think of ATM.

twaugh: I just filed https://github.com/containers/image/issues/248. If skopeo could fetch this for us from a remote image we wouldn't have to pull it.

tests/plugins/test_orchestrate_build.py
```diff
@@ -0,0 +1,433 @@
...
+                'platforms': ['x86_64', 'spam'],
+                'build_kwargs': make_worker_build_kwargs(),
+                'osbs_client_config': str(tmpdir),
+            }
+        }]
+    )
+
+    with pytest.raises(PluginFailedException) as exc:
+        runner.run()
+        assert "'No clusters found for platform spam!'"
```
twaugh: What's this line doing?

lcarva: Moved outside of with-block so it actually executes.

tests/plugins/test_orchestrate_build.py
```diff
@@ -0,0 +1,433 @@
...
+                self.exception_raised = True
+                raise BuildCanceledException()
+
+    raise_once = RaiseOnce()
+    (flexmock(AsyncResult).should_receive('wait')
+        .replace_with(raise_once.wait))
+
+    with pytest.raises(PluginFailedException) as exc:
+        build_result = runner.run()
+        assert 'BuildCanceledException' in str(exc)
```
twaugh: Here's another assert-inside-raises.

tests/plugins/test_orchestrate_build.py
```diff
@@ -0,0 +1,433 @@
...
+                'platforms': ['x86_64', 'spam'],
+                'build_kwargs': make_worker_build_kwargs(),
+                'osbs_client_config': str(tmpdir),
+            }
+        }]
+    )
+
+    with pytest.raises(PluginFailedException) as exc:
+        runner.run()
+    assert "'No clusters found for platform spam!'"
```
twaugh: But what's it testing? :-)

lcarva: A platform was requested that is not found in the reactor config.

twaugh: The assert will always succeed: the string is truthy.

tests/plugins/test_orchestrate_build.py
lcarva: Fixed. Think it's the last one.

tests/plugins/test_orchestrate_build.py
lcarva: Clearly, I haven't had coffee yet... Fixed.

twaugh:
Ready for merge?

lcarva:
Yes, I have no pending changes.

PR #657, opened 2017-03-02
Remove base image inspect in bump_release

lcarva:
This change removes the `bump_release` dependency of having the base image pulled locally. This is not needed as this plugin merely checks whether or not, one of, the `release` labels is set.

Also, this removes the need of orchestrator builds having to pull the base image locally.

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -43,7 +43,7 @@ class BumpReleasePlugin(PreBuildPlugin):
         run the plugin
         """
 
-        parser = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        parser = df_parser(self.workflow.builder.df_path)
         release_labels = get_all_label_keys('release')
         dockerfile_labels = parser.labels
         if any(release_label in dockerfile_labels


twaugh:
To be clear about what this means: no `release` label can ever use environment variables.

Note that currently Fedora *does* use environment variables in release labels:
http://pkgs.fedoraproject.org/cgit/docker/etcd.git/tree/Dockerfile

Do we really want to break that?

twaugh:
Oh... but which parent image would you pull?

twaugh:
I guess this isn't a problem: `bump_release` only needs to take action when there is no `release` label present, and for that it does not need the parent image.

If there is a release label present and it uses an environment variable, `bump_release` will leave it alone and the worker builds will pull the parent image and perform environment variable substitution.

Remaining concern then: do we want to adjust a unit test to ensure `bump_release` doesn't get changed in the future to require a parent image to be pulled?

lcarva:
@twaugh yes this wouldn't break Fedora's example you pointed out, as this plugin merely checks for the presence of the label.

However! As @vrutkovs kindly pointed out on #656, there are other cases for actually reading the value of the release label. It appears the base image must be pulled after all even for orchestrator builds.

Although the code changes in this PR are still valid, we don't need it as there's no longer a use case for using `bump_release` plugin without `pull_base_image` plugin. Closing PR for now.

PR #660, opened 2017-03-05
pulp_pull: make pulls insecure

vrutkovs:


atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -59,7 +59,7 @@ def run(self):
 
         while True:
             # Pull the image from Crane
-            name = self.tasker.pull_image(pullspec)
+            name = self.tasker.pull_image(pullspec, insecure=True)
```
lcarva: Would be nice to have this configurable.

vrutkovs: Agree, a config option is a good idea. Which default value should we stick to - I guess `insecure=False`?

lcarva: Yes, I think that would be in line with how the other plugins behave. There's probably also an existing way to determine this value from osbs-client for when configuring this plugin.

vrutkovs:
Test case added

PR #661, opened 2017-03-06
pulp_sync: print 'image available at' in pulp_sync plugin

vrutkovs:
This is essential for v2-only setups. 

The commit also uses correct repository_id (namespace/repo vs. repo) in order to sync from correct location.

TODO:
 * [x] Add a test case
 * [x] Show this message only if it was not shown during `pulp_push`
 * [x] ~~Add a testcase for missing namespace~~ the test already exists

### DIFF
--- a/atomic_reactor/plugins/post_pulp_sync.py
+++ b/atomic_reactor/plugins/post_pulp_sync.py
@@ -119,6 +119,18 @@ class PulpSyncPlugin(PostBuildPlugin):
             self.log.error("will not delete from registry as instructed: "
                            "not implemented")
 
+        found_self = False
+        self.publish = True
+        for plugin in workflow.postbuild_plugins_conf or []:
+            if plugin['name'] == self.key:
+                found_self = True
+                continue
+
+            # A hardcoded name is used to avoid circular imports
+            if found_self and plugin['name'] == 'pulp_push':
+                self.publish = False
+                break
+
     def set_auth(self, pulp):
         path = self.pulp_secret_path
         if path is not None:
@@ -216,10 +228,15 @@ class PulpSyncPlugin(PostBuildPlugin):
                 repos[image.pulp_repo] = repo_id
 
             images.append(ImageName(registry=pulp_registry,
-                                    repo=image.repo))
+                                    repo=image.repo,
+                                    tag=image.tag))
 
         self.log.info("publishing to crane")
         pulp.crane(list(repos.values()), wait=True)
+
+        if self.publish:
+            for image_name in images:
+                self.log.info("image available at %s", str(image_name))
 
         # Return the set of qualified repo names for this image
         return images
--- a/atomic_reactor/plugins/post_push_to_pulp.py
+++ b/atomic_reactor/plugins/post_push_to_pulp.py
@@ -345,7 +345,8 @@ class PulpPushPlugin(PostBuildPlugin):
                 image_file.flush()
                 crane_repos = self.push_tar(image_file.name, image_names)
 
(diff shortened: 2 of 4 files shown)


atomic_reactor/plugins/post_pulp_sync.py
```diff
@@ -208,18 +208,22 @@ def run(self):
             if image.pulp_repo not in repos:
                 repo_id = self.create_repo_if_missing(pulp,
                                                       image.pulp_repo,
-                                                      image.repo)
+                                                      image.to_str(registry=False, tag=False))
```
twaugh: Was this wrong previously?

atomic_reactor/plugins/post_pulp_sync.py
```diff
@@ -208,18 +208,22 @@ def run(self):
...
                 self.log.info("syncing %s", repo_id)
                 pulp.syncRepo(repo=repo_id,
                               feed=self.docker_registry,
                               **kwargs)
                 repos[image.pulp_repo] = repo_id
 
             images.append(ImageName(registry=pulp_registry,
-                                    repo=image.repo))
+                                    repo=image.to_str(registry=False, tag=False),
+                                    tag=image.tag))
```
twaugh: Was this wrong previously?

vrutkovs: It was using just the repo name, omitting namespace, so an image 'registry:5000/namespace/repo' would render in crane as 'crane:8888/repo' (unless a correct redirect link was created previously)

twaugh: Is this value covered by testing?

atomic_reactor/plugins/post_push_to_pulp.py
```diff
@@ -345,7 +345,12 @@ def run(self):
                 image_file.flush()
                 crane_repos = self.push_tar(image_file.name, image_names)
 
-        for image_name in crane_repos:
-            self.log.info("image available at %s", str(image_name))
+        if not hasattr(self.workflow, "image_names_printed") or \
+           not self.workflow.image_names_printed:
```
twaugh: Should use `self.publish` to determine whether to do this.

atomic_reactor/plugins/post_pulp_sync.py
vrutkovs: Yes, it was not passing the namespace param. Fixed

atomic_reactor/plugins/post_push_to_pulp.py
vrutkovs: Implemented this.

I had to hardcode 'pulp_push' plugin key in order to prevent cyclic imports, not sure if there is a better way to do this

atomic_reactor/plugins/post_pulp_sync.py
twaugh: Can we include a test case to catch it?

atomic_reactor/plugins/post_push_to_pulp.py
twaugh: Yes, I don't see any way around it.

atomic_reactor/plugins/post_pulp_sync.py
lcarva: A test to ensure `createRepo` is called with correct args and only at expected scenarios would be a great improvement on the existing tests.

vrutkovs: Turns out this case is checked in `test_create_missing_repo` so this change is not needed

PR #667, opened 2017-03-14
metadata in the workflow

rcerven:

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -249,7 +249,7 @@ class DockerBuildWorkflow(object):
     def __init__(self, source, image, prebuild_plugins=None, prepublish_plugins=None,
                  postbuild_plugins=None, exit_plugins=None, plugin_files=None,
                  openshift_build_selflink=None, client_version=None,
-                 buildstep_plugins=None, **kwargs):
+                 buildstep_plugins=None, metadata=None, **kwargs):
         """
         :param source: dict, where/how to get source code to put in image
         :param image: str, tag for built image ([registry/]image_name[:tag])
@@ -283,6 +283,7 @@ class DockerBuildWorkflow(object):
         self.autorebuild_canceled = False
         self.plugin_failed = False
         self.plugin_files = plugin_files
+        self.metadata = metadata
 
         self.kwargs = kwargs
 
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -597,13 +597,22 @@ def get_build_json():
         raise
 
 
-def is_scratch_build():
-    build_json = get_build_json()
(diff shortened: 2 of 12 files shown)


tests/plugins/test_store_metadata.py
```diff
@@ -432,6 +433,8 @@ def test_filter_repositories(tmpdir, pulp_registries, docker_registries,
         }]
     )
     output = runner.run()
+    print("===============================")
```
vrutkovs: Is this required? How would this help if tests don't pass?

vrutkovs:
Some parts of this PR look like a good idea - e.g. `get_build_json_metadata` - but `__do_init__` isn't. 

What's the reason for this change?

atomic_reactor/util.py
```diff
@@ -597,13 +597,16 @@ def get_build_json():
...
 
 
-def is_scratch_build():
-    build_json = get_build_json()
-    try:
-        return build_json['metadata']['labels'].get('scratch', False)
-    except KeyError:
-        logger.error('metadata.labels not found in build json')
-        raise
+def get_build_json_metadata(workflow):
```
lcarva: The problem here is that `get_build_json` is not reliable because `os.environ["BUILD"]` is not always set, right? 

Why not just modify the input plugins to set this environment variable? Or even better, make `get_build_json` take the workflow so it can inspect the result of the `InputPluginsRunner`?

lcarva: Seems that workflow knows nothing about InputPluginsRunner so it can't currently fetch its result.

atomic_reactor/inner.py
```diff
@@ -245,11 +245,16 @@ class DockerBuildWorkflow(object):
     5. tag it
     6. push it to registries
     """
+    def __init__(self, build_json):
+        """
+        init wrapper, so we can process build_json intact as well
+        """
+        self.__do_init__(build_json, **build_json)
 
-    def __init__(self, source, image, prebuild_plugins=None, prepublish_plugins=None,
```
lcarva: A less intrusive approach would be to add an additional optional param to `__init__`: `raw_build_json`.
Initializing the workflow would simply change to: `DockerBuildWorkflow(raw_build_json=build_json, **build_json)`
This would avoid the need to `__do_init__` hack.

tests/plugins/test_store_metadata.py
rcerven: that was forgotten to delete

atomic_reactor/util.py
rcerven: get_build_json is used by inputplugin to obtain input for workflow so at that time workflow doesn't even exist

PR #669, opened 2017-03-21
if build_image is used, it is also passed to worker build

rcerven:

### DIFF
--- a/tests/plugins/test_orchestrate_build.py
+++ b/tests/plugins/test_orchestrate_build.py
@@ -20,6 +20,7 @@ from dockerfile_parse import DockerfileParser
 from flexmock import flexmock
 from multiprocessing.pool import AsyncResult
 from osbs.api import OSBS
+from osbs.conf import Configuration
 from osbs.build.build_response import BuildResponse
 from osbs.exceptions import OsbsException
 from tests.constants import MOCK_SOURCE, TEST_IMAGE, INPUT_IMAGE, SOURCE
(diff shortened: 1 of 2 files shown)


lcarva:
Looks great! Just needs some unit testing.

vrutkovs:
Looks good, just need tests

tests/plugins/test_orchestrate_build.py
```diff
@@ -183,9 +188,11 @@ def test_orchestrate_build(tmpdir):
                 'platforms': ['x86_64'],
                 'build_kwargs': make_worker_build_kwargs(),
                 'osbs_client_config': str(tmpdir),
+                'worker_build_image': worker_build_image
```
lcarva: Maybe only set `worker_build_image` in `args` if it's not `None`? That way you can verify that it default to `None` when not provided.

rcerven:
build_image is in config, so create_worker_build doesn't know about it yet, as it is parsed during creating build_reqeust in set_params

twaugh:
Checking the `build_image` used by the `Configuration` object (or used to instantiate it) would be a good test.

tests/plugins/test_orchestrate_build.py
```diff
@@ -170,22 +171,43 @@ def make_worker_build_kwargs(**overrides):
...
-                'platforms': ['x86_64'],
-                'build_kwargs': make_worker_build_kwargs(),
-                'osbs_client_config': str(tmpdir),
-            }
+            'args': plugin_args
         }]
     )
+
+    plugin = runner.create_instance_from_plugin(OrchestrateBuildPlugin, plugin_args)
+    assert plugin.worker_build_image == worker_build_image
```
twaugh: This line is testing an aspect of the implementation, which we no longer need to do.

rcerven: removed

PR #670, opened 2017-03-22
Additional tags: check invalid tag names

csomh:
As for now, additional tag names are checked not to contain '-',
if they do, they are ignored.

This causes invalid tag names to be used, and failing the build
during tag_and_push.

The change validates tag names and removes additional tags, which
are not valid or have '-'.

### DIFF
--- a/atomic_reactor/plugins/post_tag_from_config.py
+++ b/atomic_reactor/plugins/post_tag_from_config.py
@@ -7,9 +7,10 @@ of the BSD license. See the LICENSE file for details.
 """
 
 import os
+import re
 
 from atomic_reactor.plugin import PostBuildPlugin
-from atomic_reactor.constants import INSPECT_CONFIG
+from atomic_reactor.constants import INSPECT_CONFIG, TAG_NAME_REGEX
 from atomic_reactor.util import get_preferred_label_key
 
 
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/post_tag_from_config.py
```diff
@@ -56,11 +56,11 @@ def get_tags(self):
         with open(tags_filename) as tags_file:
             for tag in tags_file:
                 tag = tag.strip()
-                if tag:
-                    if '-' in tag:
-                        self.log.warning("tag '%s' includes '-', ignoring", tag)
-                    else:
-                        tags.append(tag)
+                if is_valid_tag_name(tag) and '-' not in tag:
```
vrutkovs: Do we need `'-' not in tag` here? This won't pass `is_valid_tag_name`

atomic_reactor/plugins/post_tag_from_config.py
```diff
@@ -56,11 +56,11 @@ def get_tags(self):
...
+                    tags.append(tag)
+                else:
+                    self.log.warning("tag '%s' does not match "
+                        "'[a-zA-Z0-9_][a-zA-Z0-9_.]{0,127}', ignoring", tag)
```
lcarva: I'd rather not have this regex re-defined here. Can we just make it available publicly from `util` module? Message would have to be written as "does not match regex, or includes dashes".

atomic_reactor/util.py
```diff
@@ -852,4 +852,18 @@ def df_parser(df_path, workflow=None, cache_content=False, env_replace=True, par
...
+    """Check if tag name of an image is valid
 
+    See <https://docs.docker.com/engine/reference/commandline/tag/>
+
+        A tag name must be valid ASCII and may contain lowercase and
+        uppercase letters, digits, underscores, periods and dashes.
+        A tag name may not start with a period or a dash and may contain
+        a maximum of 128 characters.
+    """
+    PATTERN = r'^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$'
```
lcarva: Let's use `\w` here since it's the equivalent to `[a-zA-Z0-9_]`: `r'[\w][\w.-]{0,127}$`

https://docs.python.org/2/library/re.html

atomic_reactor/plugins/post_tag_from_config.py
vrutkovs: +1 for moving this to `util.py`

csomh: Agree with the proposed wording for the message and with avoiding redefinition.

What do you think about moving the regex to `atomic_reactor.constants` instead of `util`?

twaugh: Seems fine.

PR #671, opened 2017-03-22
pulp_sync: include namespace in the generated image name

vrutkovs:
This also makes sure tests are parsing the passed image name and not constructing
a new one without a namespace, as its the way the actual code acts

### DIFF
--- a/atomic_reactor/plugins/post_pulp_sync.py
+++ b/atomic_reactor/plugins/post_pulp_sync.py
@@ -220,7 +220,8 @@ class PulpSyncPlugin(PostBuildPlugin):
             if image.pulp_repo not in repos:
                 repo_id = self.create_repo_if_missing(pulp,
                                                       image.pulp_repo,
-                                                      image.repo)
+                                                      image.to_str(registry=False,
+                                                                   tag=False))
                 self.log.info("syncing %s", repo_id)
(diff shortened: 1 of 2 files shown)


twaugh:
Should [the test-case for image availability logging](https://github.com/vrutkovs/atomic-reactor/blob/71df6919a8d8c2271cb5d4cb81bf42f4530ea3e4/tests/plugins/test_pulp_sync.py#L657) be a bit more explicit about what the line should look like? e.g.:
```
expected_log = 'image available at %s/%s:%s' % (mockpulp.registry, docker_repository, image.tag)
```

i.e. making sure that `docker_repository` (which is 'prod/myrepository') forms part of the message?

vrutkovs:
This string is being constructed in `ImageName.to_str()`. This class could also have `registry=mockpulp.registry` set to make it more explicit (and more similar to actual plugin code) - I'll check if this is a better option

vrutkovs:
That doesn't really work, as pulp address is stored in `MockPulp` and `workflow` image list should be set before pulp is being mocked there. Lets leave it as it is

vrutkovs:
Rewrote the test to make sure we generate less names, relying on `ImageName.to_str()` instead, PTAL

PR #672, opened 2017-03-24
exit_sendmail: rework sendmail plugin to use information from koji

vrutkovs:
PDC part is removed, instead we'll use the following recepients:
 * If `always_notify` param is set it will be used an email address
 * If `to_koji_submitter` is set, the plugin will query Koji for kerberos 
 principal of the user, who initiated the task. The kerberos principal would be 
 used as an email address
 * If `to_koji_pkgowner` is set, the plugin will query Koji for build info 
 and include owners of the package-tags combinations in the notification
 
If no recepients are found, the plugin will send an error notification to people,
specified in `error_addresses`

TODO:
 * [x] Sort out `submitter` param - its best be renamed to `always_notify` or similar
 * [x] Add tests to cover exceptions when fetching koji submitter / package owner
 * [x] Introduce `email_domain` and construct email addresses if Koji instance doesn't use kerberos

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -281,6 +281,7 @@ class DockerBuildWorkflow(object):
         self.plugins_durations = {}
         self.plugins_errors = {}
         self.autorebuild_canceled = False
+        self.build_canceled = False
         self.plugin_failed = False
         self.plugin_files = plugin_files
 
@@ -330,6 +331,7 @@ class DockerBuildWorkflow(object):
         return self._base_image_inspect
 
     def throw_canceled_build_exception(self, *args, **kwargs):
+        self.build_canceled = True
         raise BuildCanceledException("Build was canceled")
 
     def build_docker_image(self):
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -7,7 +7,6 @@ of the BSD license. See the LICENSE file for details.
 """
 
 from email.mime.text import MIMEText
-import os
 import smtplib
 import socket
 try:
@@ -15,12 +14,11 @@ try:
 except ImportError:
     from urllib.parse import urljoin
 
-import requests
-
 from atomic_reactor.plugin import ExitPlugin, PluginFailedException
 from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
-from atomic_reactor.source import GitSource
-from atomic_reactor.util import df_parser
+from atomic_reactor.plugins.exit_koji_promote import KojiPromotePlugin
+from atomic_reactor.koji_util import create_koji_session
+from atomic_reactor.util import get_build_json
 
 
 class SendMailPlugin(ExitPlugin):
@@ -32,17 +30,13 @@ class SendMailPlugin(ExitPlugin):
                 "args": {
                     "send_on": ["auto_canceled", "auto_fail"],
                     "url": "https://openshift-instance.com",
-                    "pdc_url": "https://pdc-instance.com",
-                    # pdc_secret_path is filled in automatically by osbs-client
-                    "pdc_secret_path": "/path/to/file/with/pdc/token",
-                    "smtp_uri": "smtp-server.com",
+                    "smtp_host": "smtp-server.com",
                     "from_address": "osbs@mycompany.com",
-                    "error_addresses": ["admin@mycompany.com"],
-                    # optional arguments follow
-                    "submitter": "John Smith <jsmith@mycompany.com>",
-                    "pdc_verify_cert": true,
-                    "pdc_component_df_label": "com.redhat.component",
-                    "pdc_contact_role": "Devel_Owner"
+                    "error_addresses": ["admin@mycompany.com", "manager@mycompany.com"],
+                    "additional_addresses": ["jsmith@mycompany.com", "user@mycompany.com"],
+                    "email_domain": "example.com",
+                    "to_koji_submitter": True,
+                    "to_koji_pkgowner": True,
                 }
         }]
     """
@@ -51,51 +45,100 @@ class SendMailPlugin(ExitPlugin):
     # symbolic constants for states
     MANUAL_SUCCESS = 'manual_success'
     MANUAL_FAIL = 'manual_fail'
+    MANUAL_CANCELED = 'manual_canceled'
     AUTO_SUCCESS = 'auto_success'
     AUTO_FAIL = 'auto_fail'
     AUTO_CANCELED = 'auto_canceled'
+    DEFAULT_SUBMITTER = 'Unknown'
 
-    allowed_states = set([MANUAL_SUCCESS, MANUAL_FAIL, AUTO_SUCCESS, AUTO_FAIL, AUTO_CANCELED])
+    allowed_states = set([MANUAL_SUCCESS, MANUAL_FAIL, MANUAL_CANCELED,
+                          AUTO_SUCCESS, AUTO_FAIL, AUTO_CANCELED])
 
-    PDC_TOKEN_FILE = 'pdc.token'
-    PDC_CONTACT_ROLE = 'Devel_Owner'
-
-    def __init__(self, tasker, workflow, send_on=None, url=None, submitter='unknown', pdc_url=None,
-                 pdc_verify_cert=True, pdc_component_df_label="com.redhat.component", pdc_secret_path=None,
-                 pdc_contact_role=None, smtp_uri=None, from_address=None,
-                 error_addresses=None):
+    def __init__(self, tasker, workflow,
+                 smtp_host, from_address,
+                 send_on=(AUTO_CANCELED, AUTO_FAIL, MANUAL_SUCCESS, MANUAL_FAIL),
+                 url=None,
+                 error_addresses=(),
+                 additional_addresses=(),
+                 email_domain=None,
+                 koji_hub=None,
+                 koji_root=None,
+                 koji_proxyuser=None,
+                 koji_ssl_certs_dir=None,
+                 koji_krb_principal=None,
+                 koji_krb_keytab=None,
+                 to_koji_submitter=False,
+                 to_koji_pkgowner=False):
         """
         constructor
 
         :param tasker: DockerTasker instance
         :param workflow: DockerBuildWorkflow instance
-        :param send_on: list of build states when a notification should be sent
-        :param url: URL to OSv3 instance where the build logs are stored
-        :param submitter: name of user who submitted a build (plain string)
-        :param pdc_url: URL of PDC to query for contact information
-        :param pdc_verify_cert: whether or not to verify SSL cert of PDC (defaults to True)
-        :param pdc_component_df_label: name of Dockerfile label to use as PDC global_component
-        :param pdc_secret_path: path to pdc.token file; $SOURCE_SECRET_PATH otherwise
-        :param pdc_contact_role: name of PDC role to contact
-        :param smtp_uri: URL of SMTP server to use to send the message (e.g. "foo.com:25")
-        :param from_address: the "From" of the notification email
-        :param error_addresses: list of email addresses where to send an email if there's an error
-            (e.g. if we can't find out who to notify about the failed build)
+        :param send_on: list of str, list of build states when a notification should be sent
+            see 'allowed_states' constant and rules in '_should_send' function
+        :param url: str, URL to OSv3 instance where the build logs are stored
+        :param smtp_host: str, URL of SMTP server to use to send the message (e.g. "foo.com:25")
+        :param from_address: str, the "From" of the notification email
+        :param error_addresses: list of str, list of email addresses where to send an email
+            if an error occurred (e.g. if we can't find out who to notify about the failed build)
+        :param additional_addresses: list of str, always send a message to these email addresses
+        :param email_domain: str, email domain used when email addresses cannot be fetched via
+            kerberos principal
+        :param koji_hub: str, koji hub (xmlrpc)
+        :param koji_root: str, koji root (storage)
+        :param koji_proxyuser: str, proxy user
+        :param koji_ssl_certs_dir: str, path to "cert", "ca", and "serverca"
+        :param koji_krb_principal: str, name of Kerberos principal
+        :param koji_krb_keytab: str, Kerberos keytab
+        :param to_koji_submitter: bool, send a message to the koji submitter
+        :param to_koji_pkgowner: bool, send messages to koji package owners
         """
         super(SendMailPlugin, self).__init__(tasker, workflow)
-        self.send_on = send_on
+        self.send_on = set(send_on)
         self.url = url
-        self.submitter = submitter
-        self.pdc_url = pdc_url
-        self.pdc_verify_cert = pdc_verify_cert
-        self.pdc_component_df_label = pdc_component_df_label
-        self.pdc_secret_path = pdc_secret_path
-        self.pdc_contact_role = pdc_contact_role or self.PDC_CONTACT_ROLE
-        self.smtp_uri = smtp_uri
+        self.additional_addresses = list(additional_addresses)
+        self.smtp_host = smtp_host
         self.from_address = from_address
-        self.error_addresses = error_addresses
+        self.error_addresses = list(error_addresses)
+        self.email_domain = email_domain
+        self.koji_hub = koji_hub
+        self.koji_root = koji_root
+        self.koji_auth_info = {
+            'proxyuser': koji_proxyuser,
+            'ssl_certs_dir': koji_ssl_certs_dir,
+            'krb_principal': koji_krb_principal,
+            'krb_keytab': koji_krb_keytab,
+        }
+        self.to_koji_submitter = to_koji_submitter
+        self.to_koji_pkgowner = to_koji_pkgowner
+        self.submitter = self.DEFAULT_SUBMITTER
 
-    def _should_send(self, rebuild, success, canceled):
+        try:
+            metadata = get_build_json().get("metadata", {})
+            self.koji_task_id = int(metadata['labels']['koji-task-id'])
+        except Exception:
+            self.log.exception("Failed to fetch koji task ID")
+            self.koji_task_id = None
+        else:
+            self.log.info("Koji task ID: %s", self.koji_task_id)
+
+        try:
+            self.koji_build_id = self.workflow.exit_results.get(KojiPromotePlugin.key)
+        except Exception:
+            self.log.exception("Failed to fetch koji build ID")
+            self.koji_build_id = None
+        else:
+            self.log.info("Koji build ID: %s", self.koji_build_id)
+
+        try:
+            self.session = create_koji_session(self.koji_hub, self.koji_auth_info)
+        except Exception:
+            self.log.exception("Failed to connect to koji")
+            self.session = None
+        else:
+            self.log.info("Koji connection established")
+
+    def _should_send(self, rebuild, success, auto_canceled, manual_canceled):
         """Return True if any state in `self.send_on` meets given conditions, thus meaning
         that a notification mail should be sent.
         """
@@ -104,18 +147,19 @@ class SendMailPlugin(ExitPlugin):
         should_send_mapping = {
             self.MANUAL_SUCCESS: not rebuild and success,
             self.MANUAL_FAIL: not rebuild and not success,
+            self.MANUAL_CANCELED: not rebuild and manual_canceled,
             self.AUTO_SUCCESS: rebuild and success,
             self.AUTO_FAIL: rebuild and not success,
-            self.AUTO_CANCELED: rebuild and canceled
+            self.AUTO_CANCELED: rebuild and auto_canceled
         }
 
         for state in self.send_on:
             should_send |= should_send_mapping[state]
         return should_send
 
-    def _render_mail(self, rebuild, success, canceled):
+    def _render_mail(self, rebuild, success, auto_canceled, manual_canceled):
         """Render and return subject and body of the mail to send."""
-        subject_template = 'Image %(image)s; Status %(endstate)s; Submitted by %(user)s'
+        subject_template = '%(endstate)s building image %(image)s'
         body_template = '\n'.join([
             'Image: %(image)s',
             'Status: %(endstate)s',
@@ -124,13 +168,12 @@ class SendMailPlugin(ExitPlugin):
         ])
 
         endstate = None
-        if canceled:
-            endstate = 'canceled'
+        if auto_canceled or manual_canceled:
+            endstate = 'Canceled'
         else:
-            endstate = 'successful' if success else 'failed'
-        url = None
-        if self.url and self.workflow.openshift_build_selflink:
-            url = urljoin(self.url, self.workflow.openshift_build_selflink + '/log')
+            endstate = 'Succeeded' if success else 'Failed'
+
+        url = self._get_logs_url()
 
         formatting_dict = {
             'image': self.workflow.image,
@@ -140,120 +183,131 @@ class SendMailPlugin(ExitPlugin):
         }
         return (subject_template % formatting_dict, body_template % formatting_dict)
 
-    def _get_pdc_token(self):
-        # we want to allow pdc_secret_path to be None in __init__ - I'm assuming that in future
-        #  we'll want different sources of contact info, so we only want to raise when
-        #  the plugin actually tries to authenticate against PDC and doesn't have pdc_secret_path
-        if self.pdc_secret_path is None:
-            raise PluginFailedException('Getting PDC token, but pdc_secret_path is unspecified')
-        token_file = os.path.join(self.pdc_secret_path, self.PDC_TOKEN_FILE)
-
-        self.log.debug('getting PDC token from file %s', token_file)
-
-        with open(token_file, 'r') as f:
-            return f.read().strip()
-
-    def _get_component_label(self):
-        """Get value of Dockerfile label that is to be used as `global_component` to query
-        PDC release-components API endpoint.
-        """
-        labels = df_parser(self.workflow.builder.df_path, workflow=self.workflow).labels
-        if self.pdc_component_df_label not in labels:
-            raise PluginFailedException('No %s label in Dockerfile, can\'t get PDC component',
-                                        self.pdc_component_df_label)
-        return labels[self.pdc_component_df_label]
-
-    def _get_receivers_list(self):
-        """Return list of receivers of the notification.
-
-        :raises RuntimeError: if PDC can't be contacted or doesn't provide sufficient data
-        :raises PluginFailedException: if there's a critical error while getting PDC data
-        """
-
-        # TODO: document what this plugin expects to be in Dockerfile/where it gets info from
-        global_component = self._get_component_label()
-        # this relies on bump_release plugin configuring source.git_commit to actually be
-        #  branch name, not a commit
-        if not isinstance(self.workflow.source, GitSource):
-            raise PluginFailedException('Source is not of type "GitSource", panic!')
-        git_branch = self.workflow.source.git_commit
-        try:
-            r = requests.get(urljoin(self.pdc_url, 'rest_api/v1/release-component-contacts/'),
-                             headers={'Authorization': 'Token %s' % self._get_pdc_token()},
-                             params={'global_component': global_component,
-                                     'dist_git_branch': git_branch,
-                                     'role': self.pdc_contact_role},
-                             verify=self.pdc_verify_cert)
-        except requests.RequestException as e:
-            self.log.error('failed to connect to PDC: %s', str(e))
-            raise RuntimeError(e)
-
-        if r.status_code != 200:
-            self.log.error('PDC returned status code %s, full response: %s',
-                           r.status_code, r.text)
-            raise RuntimeError('PDC returned non-200 status code (%s), see referenced build log' %
-                               r.status_code)
-
-        contacts = r.json()
-
-        if contacts['count'] == 0:
-            self.log.error('no %s role for the component', self.pdc_contact_role)
-            raise RuntimeError('no %s role for the component' % self.pdc_contact_role)
-
-        send_to = []
-        for contact in contacts['results']:
-            send_to.append(contact['contact']['email'])
-
-        return send_to
-
     def _send_mail(self, receivers_list, subject, body):
         """Actually sends the mail with `subject` and `body` to all members of `receivers_list`."""
         msg = MIMEText(body)
         msg['Subject'] = subject
         msg['From'] = self.from_address
-        msg['To'] = ', '.join(receivers_list)
+        msg['To'] = ', '.join([x.strip() for x in receivers_list])
 
         s = None
         try:
-            s = smtplib.SMTP(self.smtp_uri)
+            s = smtplib.SMTP(self.smtp_host)
             s.sendmail(self.from_address, receivers_list, msg.as_string())
-        except (socket.gaierror, smtplib.SMTPException) as e:
-            raise PluginFailedException('Error communicating with SMTP server: %s' % str(e))
+        except (socket.gaierror, smtplib.SMTPException):
+            self.log.error('Error communicating with SMTP server')
+            raise
         finally:
             if s is not None:
                 s.quit()
 
+    def _get_email_from_koji_obj(self, obj):
+        if obj.get('krb_principal'):
+            return obj['krb_principal'].lower()
+        else:
+            if not self.email_domain:
+                raise RuntimeError("Empty email_domain specified")
+            return '@'.join([obj['name'], self.email_domain])
+
+    def _get_koji_submitter(self):
+        koji_task_info = self.session.getTaskInfo(self.koji_task_id)
+        koji_task_owner = self.session.getUser(koji_task_info['owner'])
+        koji_task_owner_email = self._get_email_from_koji_obj(koji_task_owner)
+        self.submitter = koji_task_owner_email
+        return koji_task_owner_email
+
+    def _get_koji_owners(self):
+        result = []
+        koji_build_info = self.session.getBuild(self.koji_build_id)
+
+        koji_tags = self.session.listTags(self.koji_build_id)
+        for koji_tag in koji_tags:
+            koji_tag_id = koji_tag['id']
+            koji_package_id = koji_build_info['package_id']
+            koji_pkg_tag_config = self.session.getPackageConfig(koji_tag_id, koji_package_id)
+            koji_pkg_tag_owner = self.session.getUser(koji_pkg_tag_config['owner_id'])
+
+            result.append(self._get_email_from_koji_obj(koji_pkg_tag_owner))
+
+        return result
+
+    def _get_logs_url(self):
+        url = None
+        try:
+            # We're importing this here in order to trap ImportError
+            from koji import PathInfo
+            pathinfo = PathInfo(topdir=self.koji_root)
+            url = urljoin(pathinfo.work(), pathinfo.taskrelpath(self.koji_task_id))
+        except Exception:
+            self.log.exception("Failed to fetch logs from koji")
+            if self.url and self.workflow.openshift_build_selflink:
+                url = urljoin(self.url, self.workflow.openshift_build_selflink + '/log')
+        return url
+
+    def _get_receivers_list(self):
+        receivers_list = []
+        if self.additional_addresses:
+            receivers_list += self.additional_addresses
+
+        if self.session and (self.to_koji_submitter or self.to_koji_pkgowner):
+            if self.to_koji_submitter:
+                try:
+                    koji_task_owner_email = self._get_koji_submitter()
+                except Exception:
+                    self.log.exception("Failed to include a task submitter")
+                else:
+                    receivers_list.append(koji_task_owner_email)
+
+            if self.to_koji_pkgowner:
+                try:
+                    koji_task_owner_emails = self._get_koji_owners()
+                except Exception:
+                    self.log.exception("Failed to include a package owner")
+                else:
+                    receivers_list += koji_task_owner_emails
+
+        # Remove duplicates
+        receivers_list = list(set(receivers_list))
+
+        if not receivers_list:
+            raise RuntimeError("No recepients found")
+
+        return receivers_list
+
     def run(self):
         # verify that given states are subset of allowed states
-        unknown_states = set(self.send_on) - self.allowed_states
+        unknown_states = self.send_on - self.allowed_states
         if len(unknown_states) > 0:
             raise PluginFailedException('Unknown state(s) "%s" for sendmail plugin' %
                                         '", "'.join(sorted(unknown_states)))
 
         rebuild = is_rebuild(self.workflow)
-        success = not self.workflow.build_result.is_failed()
-        canceled = self.workflow.autorebuild_canceled
+        success = not self.workflow.build_process_failed
+        auto_canceled = self.workflow.autorebuild_canceled
+        manual_canceled = self.workflow.build_canceled
 
         self.log.info('checking conditions for sending notification ...')
-        if self._should_send(rebuild, success, canceled):
+        if self._should_send(rebuild, success, auto_canceled, manual_canceled):
             self.log.info('notification about build result will be sent')
-            subject, body = self._render_mail(rebuild, success, canceled)
             try:
                 self.log.debug('getting list of receivers for this component ...')
                 receivers = self._get_receivers_list()
             except RuntimeError as e:
                 self.log.error('couldn\'t get list of receivers, sending error message ...')
-                # TODO: maybe improve the error message/subject
+                # Render the body although the receivers cannot be fetched for error message
+                _, expected_body = self._render_mail(
+                    rebuild, success, auto_canceled, manual_canceled)
                 body = '\n'.join([
                     'Failed to get contact for %s, error: %s' % (str(self.workflow.image), str(e)),
                     'Since your address is in "error_addresses", this email was sent to you to '
                     'take action on this.',
                     'Wanted to send following mail:',
                     '',
-                    body
+                    expected_body
                 ])
                 receivers = self.error_addresses
             self.log.info('sending notification to %s ...', receivers)
+            subject, body = self._render_mail(rebuild, success, auto_canceled, manual_canceled)
             self._send_mail(receivers, subject, body)
         else:
             self.log.info('conditions for sending notification not met, doing nothing')
--- a/tests/plugins/test_sendmail.py
+++ b/tests/plugins/test_sendmail.py
@@ -1,230 +1,449 @@
 import os
 import smtplib
 
-from dockerfile_parse import DockerfileParser
 from flexmock import flexmock
 import pytest
-import requests
 import six
+import json
+
+try:
+    import koji
+except ImportError:
+    import inspect
+    import sys
+
+    # Find our mocked koji module
+    import tests.koji as koji
+    mock_koji_path = os.path.dirname(inspect.getfile(koji.ClientSession))
+    if mock_koji_path not in sys.path:
+        sys.path.append(os.path.dirname(mock_koji_path))
+
+    # Now load it properly, the same way the plugin will
+    del koji
+    import koji
+
 
 from atomic_reactor.plugin import PluginFailedException
 from atomic_reactor.plugins.pre_check_and_set_rebuild import CheckAndSetRebuildPlugin
 from atomic_reactor.plugins.exit_sendmail import SendMailPlugin
-from atomic_reactor.source import GitSource
-from atomic_reactor.util import ImageName
-from atomic_reactor.build import BuildResult
+from atomic_reactor.plugins.exit_koji_promote import KojiPromotePlugin
+from atomic_reactor import util
+from smtplib import SMTPException
 
 MS, MF = SendMailPlugin.MANUAL_SUCCESS, SendMailPlugin.MANUAL_FAIL
 AS, AF = SendMailPlugin.AUTO_SUCCESS, SendMailPlugin.AUTO_FAIL
-AC = SendMailPlugin.AUTO_CANCELED
+MC, AC = SendMailPlugin.MANUAL_CANCELED, SendMailPlugin.AUTO_CANCELED
+
+MOCK_EMAIL_DOMAIN = "domain.com"
+MOCK_KOJI_TASK_ID = 12345
+MOCK_KOJI_BUILD_ID = 98765
+MOCK_KOJI_PACKAGE_ID = 123
+MOCK_KOJI_TAG_ID = 456
+MOCK_KOJI_OWNER_ID = 789
+MOCK_KOJI_OWNER_NAME = "foo"
+MOCK_KOJI_OWNER_EMAIL = "foo@bar.com"
+MOCK_KOJI_OWNER_GENERATED = "@".join([MOCK_KOJI_OWNER_NAME, MOCK_EMAIL_DOMAIN])
+MOCK_KOJI_SUBMITTER_ID = 123456
+MOCK_KOJI_SUBMITTER_NAME = "baz"
+MOCK_KOJI_SUBMITTER_EMAIL = "baz@bar.com"
+MOCK_KOJI_SUBMITTER_GENERATED = "@".join([MOCK_KOJI_SUBMITTER_NAME, MOCK_EMAIL_DOMAIN])
+MOCK_ADDITIONAL_EMAIL = "spam@bar.com"
+
+
+class MockedClientSession(object):
+    def __init__(self, hub, opts=None, has_kerberos=True):
+        self.has_kerberos = has_kerberos
+
+    def krb_login(self, principal=None, keytab=None, proxyuser=None):
+        raise RuntimeError('No certificates provided')
+
+    def ssl_login(self, cert, ca, serverca, proxyuser=None):
+        return True
+
+    def getBuild(self, build_id):
+        assert build_id == MOCK_KOJI_BUILD_ID
+        return {'package_id': MOCK_KOJI_PACKAGE_ID}
+
+    def listTags(self, build_id):
+        assert build_id == MOCK_KOJI_BUILD_ID
+        return [{"id": MOCK_KOJI_TAG_ID}]
+
+    def getPackageConfig(self, tag_id, package_id):
+        assert tag_id == MOCK_KOJI_TAG_ID
+        assert package_id == MOCK_KOJI_PACKAGE_ID
+        return {"owner_id": MOCK_KOJI_OWNER_ID}
+
+    def getUser(self, user_id):
+        if user_id == MOCK_KOJI_OWNER_ID:
+            if self.has_kerberos:
+                return {"krb_principal": MOCK_KOJI_OWNER_EMAIL}
+            else:
+                return {"krb_principal": "",
+                        "name": MOCK_KOJI_OWNER_NAME}
+
+        elif user_id == MOCK_KOJI_SUBMITTER_ID:
+            if self.has_kerberos:
+                return {"krb_principal": MOCK_KOJI_SUBMITTER_EMAIL}
+            else:
+                return {"krb_principal": "",
+                        "name": MOCK_KOJI_SUBMITTER_NAME}
+
+        else:
+            assert False, "Don't know user with id %s" % user_id
+
+    def getTaskInfo(self, task_id):
+        assert task_id == MOCK_KOJI_TASK_ID
+        return {"owner": MOCK_KOJI_SUBMITTER_ID}
+
+    def listTaskOutput(self, task_id):
+        assert task_id == MOCK_KOJI_TASK_ID
+        return ["openshift-final.log", "build.log"]
+
+
+class MockedPathInfo(object):
+    def __init__(self, topdir=None):
+        self.topdir = topdir
+
+    def work(self):
+        return "https://koji/work/"
+
+    def taskrelpath(self, task_id):
+        assert task_id == MOCK_KOJI_TASK_ID
+        return "tasks/%s" % task_id
 
 
 class TestSendMailPlugin(object):
     def test_fails_with_unknown_states(self):
-        p = SendMailPlugin(None, None, send_on=['unknown_state', MS])
+        p = SendMailPlugin(None, None,
+                           smtp_host='smtp.bar.com', from_address='foo@bar.com',
+                           send_on=['unknown_state', MS])
         with pytest.raises(PluginFailedException) as e:
             p.run()
         assert str(e.value) == 'Unknown state(s) "unknown_state" for sendmail plugin'
 
-    @pytest.mark.parametrize('rebuild, success, canceled, send_on, expected', [
+    @pytest.mark.parametrize('rebuild, success, auto_canceled, manual_canceled, send_on, expected', [
         # make sure that right combinations only succeed for the specific state
-        (False, True, False, [MS], True),
-        (False, True, False, [MF, AS, AF, AC], False),
-        (False, False, False, [MF], True),
-        (False, False, False, [MS, AS, AF, AC], False),
-        (True, True, False, [AS], True),
-        (True, True, False, [MS, MF, AF, AC], False),
-        (True, False, False, [AF], True),
-        (True, False, False, [MS, MF, AS, AC], False),
-        (True, False, True, [AC], True),
+        (False, True, False, False, [MS], True),
+        (False, True, False, True, [MS], True),
+        (False, True, False, False, [MF, AS, AF, AC], False),
+        (False, True, False, True, [MF, AS, AF, AC], False),
+        (False, False, False, False, [MF], True),
+        (False, False, False, True, [MF], True),
+        (False, False, False, False, [MS, AS, AF, AC], False),
+        (False, False, False, True, [MS, AS, AF, AC], False),
+        (False, False, True, True, [MC], True),
+        (False, True, True, True, [MC], True),
+        (False, True, False, True, [MC], True),
+        (False, True, False, False, [MC], False),
+        (True, True, False, False, [AS], True),
+        (True, True, False, False, [MS, MF, AF, AC], False),
+        (True, False, False, False, [AF], True),
+        (True, False, False, False, [MS, MF, AS, AC], False),
+        (True, False, True, True, [AC], True),
         # auto_fail would also give us True in this case
-        (True, False, True, [MS, MF, AS], False),
+        (True, False, True, True, [MS, MF, AS], False),
         # also make sure that a random combination of more plugins works ok
-        (True, False, False, [AF, MS], True)
+        (True, False, False, False, [AF, MS], True)
     ])
-    def test_should_send(self, rebuild, success, canceled, send_on, expected):
-        p = SendMailPlugin(None, None, send_on=send_on)
-        assert p._should_send(rebuild, success, canceled) == expected
+    def test_should_send(self, rebuild, success, auto_canceled, manual_canceled, send_on, expected):
+        p = SendMailPlugin(None, None,
+                           smtp_host='smtp.bar.com', from_address='foo@bar.com',
+                           send_on=send_on)
+        assert p._should_send(rebuild, success, auto_canceled, manual_canceled) == expected
 
-    @pytest.mark.parametrize('autorebuild, submitter', [
-        (True, 'John Smith <jsmith@foobar.com>'),
-        (False, 'John Smith <jsmith@foobar.com>'),
-        (True, None),
-        (False, None),
+    @pytest.mark.parametrize(('autorebuild', 'auto_cancel', 'manual_cancel',
+                              'to_koji_submitter', 'has_koji_logs'), [
+        (True, False, False, True, True),
+        (True, True, False, True, True),
+        (True, False, True, True, True),
+        (True, False, False, True, False),
+        (True, True, False, True, False),
+        (True, False, True, True, False),
+        (False, False, False, True, True),
+        (False, True, False, True, True),
+        (False, False, True, True, True),
+        (False, False, False, True, False),
+        (False, True, False, True, False),
+        (False, False, True, True, False),
+        (True, False, False, False, True),
+        (True, True, False, False, True),
+        (True, False, True, False, True),
+        (True, False, False, False, False),
+        (True, True, False, False, False),
+        (True, False, True, False, False),
+        (False, False, False, False, True),
+        (False, True, False, False, True),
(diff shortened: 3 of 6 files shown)


atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -54,46 +47,59 @@ class SendMailPlugin(ExitPlugin):
     AUTO_SUCCESS = 'auto_success'
     AUTO_FAIL = 'auto_fail'
     AUTO_CANCELED = 'auto_canceled'
+    DEFAULT_SUBMITTER = 'Unknown'
```
lcarva: This should probably just be None.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -54,46 +47,59 @@ class SendMailPlugin(ExitPlugin):
...
+                 to_koji_submitter=False,
+                 to_koji_pkgowner=False):
         """
         constructor
 
         :param tasker: DockerTasker instance
         :param workflow: DockerBuildWorkflow instance
         :param send_on: list of build states when a notification should be sent
         :param url: URL to OSv3 instance where the build logs are stored
-        :param submitter: name of user who submitted a build (plain string)
```
lcarva: This is still a param.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -223,6 +164,51 @@ def _send_mail(self, receivers_list, subject, body):
...
+                session = create_koji_session(self.koji_hub, self.koji_auth_info)
+            except Exception as e:
+                self.log.error("Failed to connect to koji:\n%s", repr(e))
+            else:
+                if self.to_koji_submitter:
+                    try:
+                        koji_task_id = metadata.get('labels', {}).get('koji-task-id')
+                        koji_task_info = session.getTaskInfo(koji_task_id)
+                        koji_task_owner = session.getUser(koji_task_info['owner'])
+                        koji_task_owner_email = koji_task_owner['krb_principal'].lower()
```
lcarva: Think it's probably better to take a param for `email_domain` which can be used to make up the user's email address.
Kerberos shouldn't be required for this.

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Its being used in the body later on: `Submitted by %(user)`

Currently it would show either `<autorebuild>` or `Unknown`, which is incorrect, as data fetched from Koji is not being set there.

I'll update `_get_receivers_list` to make sure `self.submitter` is being reset to the user who submitted the koji build, but I think the `__init__` field should be renamed to smth like `cc` or `always_notify`. Thoughts?

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Oops, reverted this

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Hmm, that's a good point. I think kerberos principal should be used first - the address can be constructed from koji's user name and email address if krb is not available. Does that sound good?

lcarva: Seems reasonable. I wonder how Koji does it.

vrutkovs: Right, I'll check what's happening there

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -223,6 +169,63 @@ def _send_mail(self, receivers_list, subject, body):
...
+                session = create_koji_session(self.koji_hub, self.koji_auth_info)
+            except Exception as e:
+                self.log.error("Failed to connect to koji:\n%s", repr(e))
+            else:
+                if self.to_koji_submitter:
+                    try:
+                        koji_task_id = metadata.get('labels', {}).get('koji-task-id')
+                        koji_task_info = session.getTaskInfo(koji_task_id)
+                        koji_task_owner = session.getUser(koji_task_info['owner'])
+                        if koji_task_owner['krb_principal']:
```
rcerven: I think you can make function for this if & else as it is same code used below for the pkg owner

vrutkovs: Agreed, I'll probably split this into several more functions then

tests/plugins/test_sendmail.py
```diff
@@ -74,154 +182,172 @@ class WF(object):
         if autorebuild:
             exp_subject += '<autorebuild>'
             exp_body[2] += '<autorebuild>'
-        elif submitter:
-            exp_subject += submitter
-            exp_body[2] += submitter
+        elif to_koji_submitter:
```
rcerven: trailing space

vrutkovs: Thanks, fixed

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -54,46 +47,64 @@ class SendMailPlugin(ExitPlugin):
...
-        :param pdc_url: URL of PDC to query for contact information
-        :param pdc_verify_cert: whether or not to verify SSL cert of PDC (defaults to True)
-        :param pdc_component_df_label: name of Dockerfile label to use as PDC global_component
-        :param pdc_secret_path: path to pdc.token file; $SOURCE_SECRET_PATH otherwise
-        :param pdc_contact_role: name of PDC role to contact
         :param smtp_uri: URL of SMTP server to use to send the message (e.g. "foo.com:25")
         :param from_address: the "From" of the notification email
         :param error_addresses: list of email addresses where to send an email if there's an error
             (e.g. if we can't find out who to notify about the failed build)
+        :param always_notify: always send a message to this user (plain string)
```
rcerven: maybe write here also that it has to be email?

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -54,46 +47,64 @@ class SendMailPlugin(ExitPlugin):
...
+        :param email_domain: set email domain to use in constructed email addresses
+        :param koji_hub: str, koji hub (xmlrpc)
+        :param koji_proxyuser: str, proxy user
+        :param koji_ssl_certs_dir: str, path to "cert", "ca", and "serverca"
+        :param koji_krb_principal: str, name of Kerberos principal
+        :param koji_krb_keytab: str, Kerberos keytab
+        :param to_koji_submitter: find the koji submitter and include the user in receivers list
```
rcerven: say it is boolean?

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -54,46 +47,64 @@ class SendMailPlugin(ExitPlugin):
...
+        :param to_koji_pkgowner: find the koji package owner and include the user in receivers list
```
rcerven: say it is boolean?

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Updated the docstring, now all params have type specified

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -223,6 +169,69 @@ def _send_mail(self, receivers_list, subject, body):
             if s is not None:
                 s.quit()
 
+    def _get_email_from_koji_obj(self, obj):
+        if obj.get('krb_principal') and obj['krb_principal'] != '':
+            return obj['krb_principal'].lower()
+        else:
+            return '@'.join([obj['owner_name'], self.email_domain])
```
rcerven: this will raise TypeError if email_domain is '' or None

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -54,46 +47,65 @@ class SendMailPlugin(ExitPlugin):
...
-                 pdc_verify_cert=True, pdc_component_df_label="com.redhat.component", pdc_secret_path=None,
-                 pdc_contact_role=None, smtp_uri=None, from_address=None,
-                 error_addresses=None):
+    def __init__(self, tasker, workflow,
+                 send_on=None,
+                 url=None,
+                 smtp_uri=None,
+                 from_address=None,
+                 error_addresses=None,
+                 always_notify='',
```
rcerven: this might be by default None

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: It doesn't for me:
```
>>> '@'.join(['',''])
'@'
```

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Good catch, reset this to `None`

atomic_reactor/plugins/exit_sendmail.py
rcerven: ahh right sorry, I missed that [],
anyway email_domain should contain at least something "char.char"

rcerven: so if there will be check for at least valid domainname, it may also default to None

vrutkovs: Oh, actually we want to have exception thrown here - in this case the email would be set to `error_addresses`. If osbs-client has incorrectly configured the plugin then there's not much we can do - error would be logged anyway.

Reset the default value to `None` so that this error would be caught earlier

rcerven: on the note of error_address it shouldn't be probably None by default :)

rcerven: for that error_address we could also try obtain email from last commit to that repo?

vrutkovs: `error_address` should be OSBS admins (read: us), as it means the plugin has crashed / misconfigured.

This would be configured in osbs-client

rcerven: ahh ok, it will be passed by osbs-client

rcerven:
+1

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -223,6 +169,71 @@ def _send_mail(self, receivers_list, subject, body):
...
+        koji_task_id = metadata.get('labels', {}).get('koji-task-id')
+        koji_task_info = session.getTaskInfo(koji_task_id)
+        koji_task_owner = session.getUser(koji_task_info['owner'])
+        koji_task_owner_email = self._get_email_from_koji_obj(koji_task_owner)
+        self.submitter = koji_task_owner_email
+        return koji_task_owner_email
+
+    def _get_koji_owners(self, session, metadata):
+        result = []
+        koji_build_id = metadata.get('labels', {}).get('koji-build-id')
```
twaugh: Why `.ge()t` here?

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -223,6 +169,71 @@ def _send_mail(self, receivers_list, subject, body):
...
+        receivers_list = []
+        if self.always_notify:
+            receivers_list.append(self.always_notify)
+
+        if self.to_koji_submitter or self.to_koji_pkgowner:
+            try:
+                metadata = get_build_json().get("metadata", {})
+                session = create_koji_session(self.koji_hub, self.koji_auth_info)
+            except Exception as e:
+                self.log.error("Failed to connect to koji:\n%s", repr(e))
```
twaugh: Could use `.exception` here (nitpick)

vrutkovs: Good catch, fixed

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -223,6 +169,71 @@ def _send_mail(self, receivers_list, subject, body):
...
+    def _get_email_from_koji_obj(self, obj):
+        if obj.get('krb_principal') and obj['krb_principal'] != '':
+            return obj['krb_principal'].lower()
+        else:
+            if not self.email_domain:
+                raise RuntimeError("Empty email_domain specified")
+            return '@'.join([obj['owner_name'], self.email_domain])
+
+    def _get_koji_submitter(self, session, metadata):
+        koji_task_id = metadata.get('labels', {}).get('koji-task-id')
```
vrutkovs: `.get` would not throw exceptions if labels don't have `koji-task-id` key. But since the exceptions are trapped anyway I think this can be removed

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -211,21 +211,21 @@ def _get_receivers_list(self):
                 metadata = get_build_json().get("metadata", {})
                 session = create_koji_session(self.koji_hub, self.koji_auth_info)
             except Exception as e:
-                self.log.error("Failed to connect to koji:\n%s", repr(e))
+                self.log.exception("Failed to connect to koji:\n%s", repr(e))
```
twaugh: I think you want `self.log.exception("Failed to connect to koji")`

vrutkovs: Right, updated

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -223,6 +169,71 @@ def _send_mail(self, receivers_list, subject, body):
...
+    def _get_receivers_list(self):
+        receivers_list = []
+        if self.always_notify:
+            receivers_list.append(self.always_notify)
+
+        if self.to_koji_submitter or self.to_koji_pkgowner:
+            try:
+                metadata = get_build_json().get("metadata", {})
+                session = create_koji_session(self.koji_hub, self.koji_auth_info)
+            except Exception as e:
```
twaugh: Heh, now we don't need `as e` :-)

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -32,17 +29,13 @@ class SendMailPlugin(ExitPlugin):
                 "args": {
                     "send_on": ["auto_canceled", "auto_fail"],
                     "url": "https://openshift-instance.com",
-                    "pdc_url": "https://pdc-instance.com",
-                    # pdc_secret_path is filled in automatically by osbs-client
-                    "pdc_secret_path": "/path/to/file/with/pdc/token",
                     "smtp_uri": "smtp-server.com",
                     "from_address": "osbs@mycompany.com",
                     "error_addresses": ["admin@mycompany.com"],
                     # optional arguments follow
```
lcarva: Missing `email_domain`.
Maybe we should remove this "optional arguments follow" comment, since all params in this plugin are optional.
Or, make the ones above this line, not optional.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -54,46 +47,65 @@ class SendMailPlugin(ExitPlugin):
...
-        :param pdc_verify_cert: whether or not to verify SSL cert of PDC (defaults to True)
-        :param pdc_component_df_label: name of Dockerfile label to use as PDC global_component
-        :param pdc_secret_path: path to pdc.token file; $SOURCE_SECRET_PATH otherwise
-        :param pdc_contact_role: name of PDC role to contact
-        :param smtp_uri: URL of SMTP server to use to send the message (e.g. "foo.com:25")
-        :param from_address: the "From" of the notification email
-        :param error_addresses: list of email addresses where to send an email if there's an error
-            (e.g. if we can't find out who to notify about the failed build)
+        :param send_on: list of str, list of build states when a notification should be sent
+        :param url: str, URL to OSv3 instance where the build logs are stored
```
lcarva: Should we send the users to logs in koji instead?
I understand the build may not have been initiated via koji, but if it was, it may cause confusion to end-user.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -54,46 +47,65 @@ class SendMailPlugin(ExitPlugin):
...
-        :param pdc_url: URL of PDC to query for contact information
-        :param pdc_verify_cert: whether or not to verify SSL cert of PDC (defaults to True)
-        :param pdc_component_df_label: name of Dockerfile label to use as PDC global_component
-        :param pdc_secret_path: path to pdc.token file; $SOURCE_SECRET_PATH otherwise
-        :param pdc_contact_role: name of PDC role to contact
-        :param smtp_uri: URL of SMTP server to use to send the message (e.g. "foo.com:25")
-        :param from_address: the "From" of the notification email
-        :param error_addresses: list of email addresses where to send an email if there's an error
-            (e.g. if we can't find out who to notify about the failed build)
+        :param send_on: list of str, list of build states when a notification should be sent
```
lcarva: Would be great to list, and explain, each build state.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -223,6 +169,71 @@ def _send_mail(self, receivers_list, subject, body):
...
+                        receivers_list += koji_task_owner_emails
+
+        if not receivers_list:
+            raise RuntimeError("No recepients found")
+
+        return receivers_list
+
     def run(self):
         # verify that given states are subset of allowed states
         unknown_states = set(self.send_on) - self.allowed_states
```
lcarva: I can't comment further than this, but we should use: `self.workflow.build_process_failed` instead of `self.workflow.build_result.is_failed()`

By using `build_process_failed` we'll treat other plugin failures (not just docker build failures) as not success state.

atomic_reactor/plugins/exit_sendmail.py
lcarva: We might need to add support for MANUAL_CANCELED too.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -223,6 +169,71 @@ def _send_mail(self, receivers_list, subject, body):
             if s is not None:
                 s.quit()
 
+    def _get_email_from_koji_obj(self, obj):
+        if obj.get('krb_principal') and obj['krb_principal'] != '':
```
lcarva: Can be simplified to: `if obj.get('krb_principal'):`

Empty strings are falsy.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -223,6 +169,71 @@ def _send_mail(self, receivers_list, subject, body):
...
+        return result
+
+    def _get_receivers_list(self):
+        receivers_list = []
+        if self.always_notify:
+            receivers_list.append(self.always_notify)
+
+        if self.to_koji_submitter or self.to_koji_pkgowner:
+            try:
+                metadata = get_build_json().get("metadata", {})
```
lcarva: I'm wondering if we need this. Can't the `koji-task-id` be a plugin parameter and `koji-build-id` be retrieved from `koji_promote` plugin?

If this is not running in OpenShift, `_get_koji_owners` and `_get_koji_submitter` will always raise KeyError exceptions. They will be caught since the calling code uses try-except, but I'm not sure if this should be logging an exception in that case.

Anyways, this is more of a nitpicking comment.

atomic_reactor/plugins/exit_sendmail.py
twaugh: +1

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -54,46 +47,65 @@ class SendMailPlugin(ExitPlugin):
...
-        :param from_address: the "From" of the notification email
-        :param error_addresses: list of email addresses where to send an email if there's an error
-            (e.g. if we can't find out who to notify about the failed build)
+        :param send_on: list of str, list of build states when a notification should be sent
+        :param url: str, URL to OSv3 instance where the build logs are stored
+        :param smtp_uri: str, URL of SMTP server to use to send the message (e.g. "foo.com:25")
+        :param from_address: str, the "From" of the notification email
+        :param error_addresses: list of str, list of email addresses where to send an email if
+            there's an error (e.g. if we can't find out who to notify about the failed build)
+        :param always_notify: str, always send a message to this email address
```
vrutkovs: I don't mind changing it to `cc_address`, but in fact this address would be used in `To:`, so I guess it would still be confusing.

Perhaps the better way would be to use this address in CC if koji submitter/pkg owner are available and use it in `To:` if koji recipients are not set?

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Fixed

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Removed a note about optional argument and added email_domain

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Added `MANUAL_CANCELED`

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Fixed

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: We are retrieving this from Openshift metadata, which gets set by `koji_promote` plugin. If this is expected to work outside of the Openshift pod then `koji_promote` should set this somewhere in `workflow`.

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Updated the code to use `build_process_failed`

atomic_reactor/plugins/exit_sendmail.py
twaugh: Or else just `additional_address`? It's the "always" that makes it confusing I think.

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Added a change to first look for logs in Koji. If those are unavailable a link to OSE would be inserted

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: `additional_address` sounds good, updated the code

atomic_reactor/plugins/exit_sendmail.py
twaugh: Storing it in the workflow (in the plugin's workspace) sounds like a good idea.

vrutkovs: Not sure how this will work with orchestration workflow - current implementation would work with both, as both worker and orchestration builds store metadata in OSE. If we transition to `workflow.plugins` osbs-client should make sure `sendmail` is called in the same build json as `koji_promote`

twaugh: Given this, I'm happy with making this require OpenShift for this.

vrutkovs: Just to clarify - the plugin can work without Openshift - but the logs would say 'None' there.

Koji integration would require openshift (even `koji_promote` would fail without openshift, as there would be no place to store metadata).

twaugh: +1

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -104,6 +133,7 @@ def _should_send(self, rebuild, success, canceled):
         should_send_mapping = {
             self.MANUAL_SUCCESS: not rebuild and success,
             self.MANUAL_FAIL: not rebuild and not success,
+            self.MANUAL_CANCELED: not rebuild and canceled,
```
lcarva: `canceled` is being computed as: `canceled = self.workflow.autorebuild_canceled`, think we need something else here.

atomic_reactor/plugins/exit_sendmail.py
lcarva: Agree.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -215,14 +185,87 @@ def _send_mail(self, receivers_list, subject, body):
...
+                    koji_task_owner_emails = self._get_koji_owners()
+                except Exception:
+                    self.log.exception("Failed to include a package owner")
+                else:
+                    receivers_list += koji_task_owner_emails
+
+        if not receivers_list:
+            raise RuntimeError("No recepients found")
+
+        return list(chain(receivers_list))
```
lcarva: I'm not sure this will do what you want:
```python
>>> list(chain(['spam', ['eggs', 'sausage'], 'bacon']))
['spam', ['eggs', 'sausage'], 'bacon']

# Docs indicate arg must be expanded (use star)
>>> list(chain(*[['eggs', 'sausage'], 'bacon']))
['eggs', 'sausage', 'b', 'a', 'c', 'o', 'n']
```
To work with `chain`, or `chain.from_iterable`, each list element should be a list, which is not our case.
Here's a hackish solution:

```python
from types import GeneratorType
breakfast = [['eggs', 'sausage'], 'bacon']

def force_list(item):
    if type(item) not in (list, tuple, GeneratorType):
        item = [item]
    return item

list(chain.from_iterable(force_list(item) for item in breakfast))
```

vrutkovs: Oh snap. Perhaps it would be best to use `isinstance` solution for `error_address` and `additional_address`

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,49 +45,85 @@ class SendMailPlugin(ExitPlugin):
...
-        :param pdc_contact_role: name of PDC role to contact
-        :param smtp_uri: URL of SMTP server to use to send the message (e.g. "foo.com:25")
-        :param from_address: the "From" of the notification email
-        :param error_addresses: list of email addresses where to send an email if there's an error
-            (e.g. if we can't find out who to notify about the failed build)
+        :param send_on: list of str, list of build states when a notification should be sent
+        :param url: str, URL to OSv3 instance where the build logs are stored
+        :param smtp_host: str, URL of SMTP server to use to send the message (e.g. "foo.com:25")
+        :param from_address: str, the "From" of the notification email
+        :param error_address: list of str, list of email addresses where to send an email if
```
twaugh: Any particular reason for this to be a list of strings instead of just a string?

atomic_reactor/plugins/exit_sendmail.py
twaugh: Is there a reason we don't want to just say "this must be a list" or "this must be a string of comma-separated addresses"?

Alternatively, if we want to handle either case, we need to test both cases.

vrutkovs: Right, I've hit this when I was running integration tests.

I'll rework `error_addresses` and `additional_addresses` to make those strings and mention those can contain multiple recipients separated by a comma

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -32,8 +31,8 @@ class SendMailPlugin(ExitPlugin):
                     "url": "https://openshift-instance.com",
                     "smtp_host": "smtp-server.com",
                     "from_address": "osbs@mycompany.com",
-                    "error_address": ["admin@mycompany.com"],
-                    "additional_address": "John Smith <jsmith@mycompany.com>",
+                    "error_addresses": ["admin@mycompany.com"],
```
twaugh: Docs say this is a str but example shows a list.

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Fixed - now its a string, which can contain multiple recipients separated by a comma

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Right, fixed

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -215,14 +184,87 @@ def _send_mail(self, receivers_list, subject, body):
 
         s = None
         try:
-            s = smtplib.SMTP(self.smtp_uri)
+            s = smtplib.SMTP(self.smtp_host)
             s.sendmail(self.from_address, receivers_list, msg.as_string())
```
twaugh: Is `receivers_list` a string here?

vrutkovs: It is, later on its being joined in a string in `_send_mail`

twaugh: ...but smtplib.SMTP.sendmail() expects a list there... a bare string is treated as a singleton list.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -231,8 +273,8 @@ def run(self):
                                         '", "'.join(sorted(unknown_states)))
 
         rebuild = is_rebuild(self.workflow)
-        success = not self.workflow.build_result.is_failed()
-        canceled = self.workflow.autorebuild_canceled
+        success = not self.workflow.build_process_failed
+        canceled = self.workflow.autorebuild_canceled or self.workflow.build_canceled
```
twaugh: This won't distinguish between auto and manual cancel.

Question though: is 'auto_canceled' a useful event to be notified about?

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Oh, right. `receivers_list` is being joined into a string for `msg['To']` and the list is passed in `smtplib.SMTP.sendmail`. Not sure why we should pass this list twice to make it work

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -215,14 +188,87 @@ def _send_mail(self, receivers_list, subject, body):
...
+        result = []
+        koji_build_id = self.metadata['labels']['koji-build-id']
+        koji_build_info = self.session.getBuild(koji_build_id)
+
+        koji_tags = self.session.listTags(koji_build_id)
+        for koji_tag in koji_tags:
+            koji_tag_id = koji_tag['id']
+            koji_package_id = koji_build_info['package_id']
+            koji_pkg_tag_config = self.session.getPackageConfig(koji_tag_id, koji_package_id)
+            koji_pkg_tag_owner = self.session.getUser(koji_pkg_tag_config['owner'])
```
twaugh: I think this should be `'owner_id'`.

tests/plugins/test_sendmail.py
```diff
@@ -1,22 +1,113 @@
...
+        return {'package_id': MOCK_KOJI_PACKAGE_ID}
+
+    def listTags(self, build_id):
+        assert build_id == MOCK_KOJI_BUILD_ID
+        return [{"id": MOCK_KOJI_TAG_ID}]
+
+    def getPackageConfig(self, tag_id, package_id):
+        assert tag_id == MOCK_KOJI_TAG_ID
+        assert package_id == MOCK_KOJI_PACKAGE_ID
+        return {"owner": MOCK_KOJI_OWNER_ID}
```
twaugh: Should be `'owner_id': ...`.

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: I'm not sure if it is - and it hard to try it out without an integration test. 

Lets keep it and adjust it later on when chained rebuilds are testable

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,49 +44,89 @@ class SendMailPlugin(ExitPlugin):
...
-                 pdc_verify_cert=True, pdc_component_df_label="com.redhat.component", pdc_secret_path=None,
-                 pdc_contact_role=None, smtp_uri=None, from_address=None,
-                 error_addresses=None):
+    DEFAULT_SUBMITTER = 'Unknown'
+
+    allowed_states = set([MANUAL_SUCCESS, MANUAL_FAIL, MANUAL_CANCELED,
+                          AUTO_SUCCESS, AUTO_FAIL, AUTO_CANCELED])
+
+    def __init__(self, tasker, workflow,
+                 send_on=None,
```
twaugh: Why does this default to `None`? Isn't it required?

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,49 +44,89 @@ class SendMailPlugin(ExitPlugin):
...
+                 url=None,
+                 smtp_host=None,
```
twaugh: Why does this default to `None`?

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,49 +44,89 @@ class SendMailPlugin(ExitPlugin):
...
+                 from_address=None,
```
twaugh: Why does this default to `None`?

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,49 +44,89 @@ class SendMailPlugin(ExitPlugin):
...
+                 error_addresses=None,
```
twaugh: This defaults to `None` but is never checked before being split.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,49 +44,89 @@ class SendMailPlugin(ExitPlugin):
...
-        :param smtp_uri: URL of SMTP server to use to send the message (e.g. "foo.com:25")
-        :param from_address: the "From" of the notification email
-        :param error_addresses: list of email addresses where to send an email if there's an error
-            (e.g. if we can't find out who to notify about the failed build)
+        :param send_on: list of str, list of build states when a notification should be sent
+        :param url: str, URL to OSv3 instance where the build logs are stored
+        :param smtp_host: str, URL of SMTP server to use to send the message (e.g. "foo.com:25")
+        :param from_address: str, the "From" of the notification email
+        :param error_addresses: str, comma-separated list of email addresses where to send an email
+            if an error occurred (e.g. if we can't find out who to notify about the failed build)
```
twaugh: Would it be better for this to be a list? We immediately split it before use.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,49 +44,89 @@ class SendMailPlugin(ExitPlugin):
...
+        :param additional_addresses: str, always send a message to this email address
```
twaugh: Would it be better for this to be a list? All we do is split it.

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: I don't have a preference - it can be turned into a list in osbs-client and passed here. Do we have any guidelines or examples in other plugins to decide on this?

twaugh: Just trying to get the types right. If it's a list of things, let's make it a JSON array. (e.g. with `send_on`...)

vrutkovs:
Marking this as WIP as at the moment Koji integration is broken, so I'll be pushing commits to make integration tests pass - and squash them when those are ready for review

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -207,7 +207,7 @@ def run(self):
             self.log.debug("annotations: %r", annotations)
             raise
 
-        labels = self.make_labels()
+        labels = self.make_labels(metadata)
```
twaugh: This doesn't look right to me. What are you trying to do?

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -36,7 +38,13 @@ def __init__(self, tasker, workflow, hub, target=None):
...
         # call parent constructor
         super(BumpReleasePlugin, self).__init__(tasker, workflow)
-        self.xmlrpc = create_koji_session(hub)
+        koji_auth_info = {
+            'proxyuser': koji_proxyuser,
+            'ssl_certs_dir': koji_ssl_certs_dir,
+            'krb_principal': koji_krb_principal,
+            'krb_keytab': koji_krb_keytab,
+        }
+        self.xmlrpc = create_koji_session(hub, koji_auth_info)
```
twaugh: This plugin doesn't do anything that could require authentication.

atomic_reactor/plugins/pre_koji.py
```diff
@@ -32,7 +34,13 @@ def __init__(self, tasker, workflow, target, hub, root, proxy=None):
...
         super(KojiPlugin, self).__init__(tasker, workflow)
         self.target = target
-        self.xmlrpc = koji.ClientSession(hub)
+        koji_auth_info = {
+            'proxyuser': koji_proxyuser,
+            'ssl_certs_dir': koji_ssl_certs_dir,
+            'krb_principal': koji_krb_principal,
+            'krb_keytab': koji_krb_keytab,
+        }
+        self.xmlrpc = create_koji_session(hub, koji_auth_info)
```
twaugh: This plugin doesn't do anything that could require authentication.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -215,14 +192,81 @@ def _send_mail(self, receivers_list, subject, body):
...
+                    koji_task_owner_emails = self._get_koji_owners()
+                except Exception:
+                    self.log.exception("Failed to include a package owner")
+                else:
+                    receivers_list += koji_task_owner_emails
+
+        if not receivers_list:
+            raise RuntimeError("No recepients found")
+
+        return receivers_list
```
twaugh: Let's remove duplicate addresses first.

atomic_reactor/plugins/exit_sendmail.py
twaugh: No, the current behaviour is incorrect: the config can ask for manual cancellation but not automatic cancellation, and the implementation doesn't do that.

Why not just have `auto_canceled = self.workflow.autorebuild_canceled` and `manual_canceled = self.workflow.build_canceled` and pass both through to `_should_send()`?

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Fixed, canceled is now `self.workflow.autorebuild_canceled or self.workflow.build_canceled`

twaugh: Not fixed. Asking for "manual_canceled" but not "auto_canceled" will also give notifications for auto_canceled.

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Fixed, reset the value to osbs-client defaults

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: its required, but there's no universal default for it. OSBS Client would remove a plugin if `smtp_host` or `from_address` is not set.

Should we keep this as a `None` or set some fake host name?

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Same as for `smtp_host`

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Okay, sounds good to me

atomic_reactor/plugins/exit_sendmail.py
twaugh: My preference would be to move this parameter to before the optional ones, i.e. right after tasker and workflow. We have this plugin marked as "not yet enabled" in the docs so it's safe to rearrange parameters (which we're doing anyway in this PR).

Another option: verify the parameter is not set and raise an exception if it is not.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
vrutkovs: `store_metadata` doesn't read existing metadata (e.g. `koji-task-id` label) and returns an incomplete dict of labels - it would only contain a new labels added - `koji-build-id`.

Since there is no easy way to re-read build labels here, the plugin returns a full list of labels

atomic_reactor/plugins/exit_sendmail.py
twaugh: Maybe it just should be a required parameter with no default?

atomic_reactor/plugins/pre_bump_release.py
vrutkovs: If `koji_auth_info` is not used we assume that Koji's SSL certificate is already added in the buildroot - and that might not be the case for any koji instance.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
twaugh: For the Koji task ID, you should use get_build_json() and follow the 'metadata', 'labels', and 'koji-task-id' keys, converting the resulting value to an int, like koji_promote does.

For the Koji build ID, the koji_promote plugin (which has already run) returns that so why not just get it straight from `self.workflow.exit_results[PLUGIN_KEY_KOJI_PROMOTE]`?

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Good idea, fixed

atomic_reactor/plugins/pre_bump_release.py
twaugh: Right... but this plugin doesn't do *anything* that could require authentication. It's done anonymously. Any Koji installation that doesn't allow that is broken.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
vrutkovs: That's possible, although it would complicate `sendmail` plugin a bit more - it would duplicate actions `store_metadata` does to fetch koji task id.

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Agree, reworked this to have two distinct params for that

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: This was reworked to have distinct cancel params for auto and manual cases

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
twaugh: But what you have here complicates the store_metadata plugin. It has no reason to do this.

If you want to avoid duplication, how about making a util.py function to get a named label from the build JSON metadata and have both koji_promote and sendmail use it?

atomic_reactor/plugins/exit_sendmail.py
twaugh: How about adding a brief list of each build state to the description of the send_on parameter?

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,51 +44,90 @@ class SendMailPlugin(ExitPlugin):
...
-                 pdc_contact_role=None, smtp_uri=None, from_address=None,
-                 error_addresses=None):
+    DEFAULT_SUBMITTER = 'Unknown'
+
+    allowed_states = set([MANUAL_SUCCESS, MANUAL_FAIL, MANUAL_CANCELED,
+                          AUTO_SUCCESS, AUTO_FAIL, AUTO_CANCELED])
+
+    def __init__(self, tasker, workflow,
+                 smtp_host, from_address,
+                 send_on=["auto_canceled", "auto_fail", "manual_success", "manual_fail"],
```
twaugh: Nitpick: technically we shouldn't use mutable types such as lists in parameter defaults.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -231,29 +277,32 @@ def run(self):
...
 
         self.log.info('checking conditions for sending notification ...')
-        if self._should_send(rebuild, success, canceled):
+        if self._should_send(rebuild, success, auto_canceled, manual_canceled):
             self.log.info('notification about build result will be sent')
-            subject, body = self._render_mail(rebuild, success, canceled)
             try:
                 self.log.debug('getting list of receivers for this component ...')
                 receivers = self._get_receivers_list()
+                # body should be re-rendered as self.submitter has changed
```
twaugh: Stale comment? I don't think we've rendered at all by this point.

atomic_reactor/plugins/exit_sendmail.py
twaugh: This is resolved now.

atomic_reactor/plugins/exit_sendmail.py
twaugh: This looks resolved now.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
vrutkovs: Technically `store_metadata` should know which labels are set already (not sure if reading `BUILD` env var is sufficient though). But I agree - lets not overcomplicate other plugins.

I'll rewrite the sendmail constructor to fetch task id and build id from two sources

twaugh: It calls update_labels_on_build(), which takes a delta not the full set.

Thanks, I think that's a better approach.

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Ah, correct, removed

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Fixed - default value is a tuple, and `self.send_on` would convert it to a list

atomic_reactor/plugins/pre_bump_release.py
vrutkovs: I'll move those parts in a separate commit - and drop it before the merge. For some reason my internal Koji server started requiring those.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -124,88 +167,31 @@ def _render_mail(self, rebuild, success, canceled):
         ])
 
         endstate = None
-        if canceled:
-            endstate = 'canceled'
+        if auto_canceled or manual_canceled:
+            endstate = 'Canceled'
```
twaugh: This line doesn't get covered in unit tests.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -124,88 +167,31 @@ def _render_mail(self, rebuild, success, canceled):
...
+        try:
+            url = self._get_logs_url_from_build()
+        except Exception:
+            self.log.exception("Failed to fetch logs from koji")
+            if self.url and self.workflow.openshift_build_selflink:
+                url = urljoin(self.url, self.workflow.openshift_build_selflink + '/log')
+
+        image = self.workflow.image
+        if len(self.workflow.tag_conf.unique_images) > 0:
+            image = self.workflow.tag_conf.unique_images[0]
```
twaugh: This line doesn't get covered in unit tests.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -215,14 +201,80 @@ def _send_mail(self, receivers_list, subject, body):
 
         s = None
         try:
-            s = smtplib.SMTP(self.smtp_uri)
+            s = smtplib.SMTP(self.smtp_host)
             s.sendmail(self.from_address, receivers_list, msg.as_string())
         except (socket.gaierror, smtplib.SMTPException) as e:
             raise PluginFailedException('Error communicating with SMTP server: %s' % str(e))
```
twaugh: The above two lines (exception handling for `s.sendmail()`) don't get covered by unit testing.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,51 +45,99 @@ class SendMailPlugin(ExitPlugin):
...
+            self.log.info("Koji task ID: %s" % self.koji_task_id)
 
-    def _should_send(self, rebuild, success, canceled):
+        try:
+            self.koji_build_id = int(self.workflow.exit_results.get(KojiPromotePlugin.key, {}))
+        except Exception:
+            self.log.exception("Failed to fetch koji build ID")
+            self.koji_build_id = None
+        else:
+            self.log.info("Koji build ID: %s" % self.koji_build_id)
```
vrutkovs: It is, later on it would be used in other functions to fetch details on koji build

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,51 +45,99 @@ class SendMailPlugin(ExitPlugin):
...
+
+        try:
+            self.session = create_koji_session(self.koji_hub, self.koji_auth_info)
+        except Exception:
+            self.log.exception("Failed to connect to koji")
+            self.session = None
+        else:
+            self.log.info("Koji connection established")
```
vrutkovs: It is, later on it would be used in other functions to fetch details on koji task

twaugh: Ok.

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: I guess those are pretty much self-descriptive, aren't they? Added a short notice about where those strings are used

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,51 +45,99 @@ class SendMailPlugin(ExitPlugin):
...
-                 pdc_contact_role=None, smtp_uri=None, from_address=None,
-                 error_addresses=None):
+    DEFAULT_SUBMITTER = 'Unknown'
+
+    allowed_states = set([MANUAL_SUCCESS, MANUAL_FAIL, MANUAL_CANCELED,
+                          AUTO_SUCCESS, AUTO_FAIL, AUTO_CANCELED])
+
+    def __init__(self, tasker, workflow,
+                 smtp_host, from_address,
+                 send_on=("auto_canceled", "auto_fail", "manual_success", "manual_fail"),
```
vrutkovs: Agree, fixed

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,51 +45,99 @@ class SendMailPlugin(ExitPlugin):
...
+        self.submitter = self.DEFAULT_SUBMITTER
+
+        try:
+            metadata = get_build_json().get("metadata", {})
+            self.koji_task_id = int(metadata['labels']['koji-task-id'])
+        except Exception:
+            self.log.exception("Failed to fetch koji task ID")
+            self.koji_task_id = None
+        else:
+            self.log.info("Koji task ID: %s" % self.koji_task_id)
```
vrutkovs: Fixed

vrutkovs: Yes, so far. Later on we'll add more details about koji task

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,51 +45,99 @@ class SendMailPlugin(ExitPlugin):
...
 
-    def _should_send(self, rebuild, success, canceled):
+        try:
+            self.koji_build_id = int(self.workflow.exit_results.get(KojiPromotePlugin.key, {}))
```
vrutkovs: This was planned as an extra protection in case koji would return a string ID after `CGImport`, although its probably useless. Removed

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: This doesn't seem to work as expected in integration tests (it outputs the same image as `self.workflow.image`). Removed it for now - it would be introduced in the next PRs

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -215,14 +201,80 @@ def _send_mail(self, receivers_list, subject, body):
...
+
+            result.append(self._get_email_from_koji_obj(koji_pkg_tag_owner).strip())
+
+        return result
+
+    def _get_logs_url_from_build(self):
+        # We're importing this here in order to trap ImportError
+        from koji import PathInfo
+        pathinfo = PathInfo(topdir=self.koji_root)
+        return '%s/%s' % (pathinfo.work(), pathinfo.taskrelpath(self.koji_task_id))
```
vrutkovs: Right, fixed

tests/plugins/test_sendmail.py
```diff
@@ -1,230 +1,418 @@
...
 import six
+import json
+
+try:
+    import koji
+except ImportError:
+    import inspect
+    import sys
+
+    # Find out mocked koji module
```
vrutkovs: Ah, right. Fixed in all places

tests/plugins/test_sendmail.py
```diff
@@ -1,230 +1,418 @@
...
+            'koji_hub': '',
+            'koji_proxyuser': None,
+            'koji_ssl_certs_dir': '/certs',
+            'koji_krb_principal': None,
+            'koji_krb_keytab': None
+        }
+
+        if not has_koji_logs:
+            flexmock(pathinfo) \
+                .should_receive('work') \
```
vrutkovs: Fixed all occurrences of this

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,51 +45,100 @@ class SendMailPlugin(ExitPlugin):
...
+    DEFAULT_SUBMITTER = 'Unknown'
+
+    allowed_states = set([MANUAL_SUCCESS, MANUAL_FAIL, MANUAL_CANCELED,
+                          AUTO_SUCCESS, AUTO_FAIL, AUTO_CANCELED])
+
+    def __init__(self, tasker, workflow,
+                 smtp_host, from_address,
+                 send_on=(AUTO_CANCELED, AUTO_FAIL, MANUAL_SUCCESS, MANUAL_FAIL),
+                 url=None,
+                 error_addresses=[],
```
lcarva: Let's use an empty tuple here since it's immutable.

vrutkovs: +1, fixed

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: This was resolved

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Added more testcases for auto and manual cancel

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -124,13 +168,18 @@ def _render_mail(self, rebuild, success, canceled):
...
+            endstate = 'Canceled'
         else:
-            endstate = 'successful' if success else 'failed'
+            endstate = 'Succeeded' if success else 'Failed'
+
         url = None
-        if self.url and self.workflow.openshift_build_selflink:
-            url = urljoin(self.url, self.workflow.openshift_build_selflink + '/log')
+        try:
+            url = self._get_logs_url_from_build()
```
lcarva: Nitpick: This method actually is fetching the logs URL from a Koji task, not build. It might be a bit cleaner to simply encapsulate that and this try-except block in a single function: `_get_logs_url`.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -231,29 +280,32 @@ def run(self):
...
             self.log.info('notification about build result will be sent')
-            subject, body = self._render_mail(rebuild, success, canceled)
             try:
                 self.log.debug('getting list of receivers for this component ...')
                 receivers = self._get_receivers_list()
             except RuntimeError as e:
                 self.log.error('couldn\'t get list of receivers, sending error message ...')
-                # TODO: maybe improve the error message/subject
+                # Render the body although the receivers canoot be fetched for error message
+                _, expected_body = self._render_mail(
```
lcarva: "`canoot`"

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,51 +45,100 @@ class SendMailPlugin(ExitPlugin):
...
+        :param koji_proxyuser: str, proxy user
+        :param koji_ssl_certs_dir: str, path to "cert", "ca", and "serverca"
+        :param koji_krb_principal: str, name of Kerberos principal
+        :param koji_krb_keytab: str, Kerberos keytab
+        :param to_koji_submitter: bool, send a message to the koji submitter
+        :param to_koji_pkgowner: bool, send messages to koji package owners
         """
         super(SendMailPlugin, self).__init__(tasker, workflow)
-        self.send_on = send_on
+        self.send_on = list(send_on)
```
vrutkovs: Oh, good catch. Fixed

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -51,51 +45,100 @@ class SendMailPlugin(ExitPlugin):
...
+            self.koji_task_id = int(metadata['labels']['koji-task-id'])
+        except Exception:
+            self.log.exception("Failed to fetch koji task ID")
+            self.koji_task_id = None
+        else:
+            self.log.info("Koji task ID: %s", self.koji_task_id)
 
-    def _should_send(self, rebuild, success, canceled):
+        try:
+            self.koji_build_id = self.workflow.exit_results.get(KojiPromotePlugin.key, {})
```
vrutkovs: Correct, fixed

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Renamed this function into `_get_logs_url` and moved try..except part there

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -215,14 +198,80 @@ def _send_mail(self, receivers_list, subject, body):
 
         s = None
         try:
-            s = smtplib.SMTP(self.smtp_uri)
+            s = smtplib.SMTP(self.smtp_host)
             s.sendmail(self.from_address, receivers_list, msg.as_string())
         except (socket.gaierror, smtplib.SMTPException) as e:
             raise PluginFailedException('Error communicating with SMTP server: %s' % str(e))
```
vrutkovs: Hmm, right. Perhaps in this case it would be easier to raise the original exception?

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Fixed

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -215,45 +192,122 @@ def _send_mail(self, receivers_list, subject, body):
 
         s = None
         try:
-            s = smtplib.SMTP(self.smtp_uri)
+            s = smtplib.SMTP(self.smtp_host)
             s.sendmail(self.from_address, receivers_list, msg.as_string())
         except (socket.gaierror, smtplib.SMTPException) as e:
-            raise PluginFailedException('Error communicating with SMTP server: %s' % str(e))
+            self.log.error('Error communicating with SMTP server')
+            raise e
```
twaugh: Don't need `as e` on the except line, or `e` on the raise line.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -215,45 +192,122 @@ def _send_mail(self, receivers_list, subject, body):
...
+        except Exception:
+            self.log.exception("Failed to fetch logs from koji")
+            if self.url and self.workflow.openshift_build_selflink:
+                url = urljoin(self.url, self.workflow.openshift_build_selflink + '/log')
+        return url
+
+    def _get_receivers_list(self):
+        receivers_list = []
+        if self.additional_addresses:
+            receivers_list += [x.strip() for x in self.additional_addresses]
```
twaugh: Slight improvement might be to perform this stripping on `receivers_list` once it's been assembled, rather than in several different places before assembly.

atomic_reactor/inner.py
```diff
@@ -330,6 +331,7 @@ def base_image_inspect(self):
         return self._base_image_inspect
 
     def throw_canceled_build_exception(self, *args, **kwargs):
+        self.build_canceled = True
```
lcarva: Let's add a test to make sure this behavior doesn't accidentally break in the future.

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Fixed, now recipient addresses are being stripped in `__send_mail` only

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Fixed

atomic_reactor/inner.py
vrutkovs: Good catch, updated `test_inner.py`

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -215,45 +192,122 @@ def _send_mail(self, receivers_list, subject, body):
...
+        if obj.get('krb_principal'):
+            return obj['krb_principal'].lower()
+        else:
+            if not self.email_domain:
+                raise RuntimeError("Empty email_domain specified")
+            return '@'.join([obj['name'], self.email_domain])
+
+    def _get_koji_submitter(self):
+        koji_task_info = self.session.getTaskInfo(self.koji_task_id)
+        koji_task_owner = self.session.getUser(koji_task_info['owner'])
```
rcerven: won't this throw exception if koji_task_id is None? from line 121 ?

vrutkovs: The exception would be caught in `_get_receivers_list`

tests/plugins/test_sendmail.py
```diff
@@ -1,230 +1,450 @@
...
-        (True, None),
-        (False, None),
+    def test_should_send(self, rebuild, success, auto_canceled, manual_canceled, send_on, expected):
+        p = SendMailPlugin(None, None,
+                           smtp_host='smtp.bar.com', from_address='foo@bar.com',
+                           send_on=send_on)
+        assert p._should_send(rebuild, success, auto_canceled, manual_canceled) == expected
+
+    @pytest.mark.parametrize('autorebuild, auto_cancel, manual_cancel' +
+                             ', to_koji_submitter, has_koji_logs',
```
twaugh: FWIW you can write this:
```
@pytest.mark.parametrize(('autorebuild', 'auto_cancel', 'manual_cancel',
                          'to_koji_submitter', 'has_koji_logs'),
```

vrutkovs: Oh, nice, updated this

PR #675, opened 2017-03-31
Split koji build tag

lcarva:


atomic_reactor/plugins/exit_koji_tag_build.py
```diff
@@ -0,0 +1,97 @@
...
+
+from __future__ import unicode_literals
+
+from atomic_reactor.constants import PLUGIN_KOJI_TAG_BUILD_KEY
+from atomic_reactor.koji_util import create_koji_session, TaskWatcher
+from atomic_reactor.plugin import ExitPlugin
+from atomic_reactor.plugins.exit_koji_promote import KojiPromotePlugin
+
+
+class KojiTagBuildPlugin(ExitPlugin):
```
rcerven: why not reuse the code from koji_promote ? it is exactly the same code isn't it?

rcerven: that way also if we would need change in that we don't have to modify it in 2 places

lcarva: That's an interesting idea. I'd rather not call directly into another plugin, though. How about moving the tag build logic to koji_utils.py? That way both plugins can use it.

rcerven: we already have plugins using other plugins,
but sure we can move it to koji_utils as long we don't have same code in 2 places

because basically as of now you can just as well call __init__ from koji promote, 
in run, have that check before create_koji_session
call login from koji_promote

and make the rest method in koji_promote and call it :)

vrutkovs: +1 to using this in `koji_utils.py`

lcarva: Done.

atomic_reactor/plugins/exit_koji_tag_build.py
```diff
@@ -0,0 +1,88 @@
...
+            self.log.info('Build failed, skipping koji tagging')
+            return
+
+        build_id = self.workflow.exit_results.get(KojiPromotePlugin.key)
+        if not build_id:
+            self.log.info('No koji build from %s', KojiPromotePlugin.key)
+            return
+
+        session = create_koji_session(self.kojihub, self.koji_auth)
+        build_tag = tag_koji_build(session, build_id, self.target,
```
rcerven: shouldn't it be checking as  well if store_metadata_in_osv3 was successful ?

lcarva: `store_metadata_in_osv3` is not allowed to fail anymore. If it fails to execute, then this will return True `self.workflow.build_process_failed` and this plugin will stop execution.

rcerven:
+1

PR #676, opened 2017-04-03
Make bump_release and koji plugins use koji auth and ssl_dir

vrutkovs:
Buildroot is assumed to have Koji certificate trusted already, which might not
be true for any given koji server.

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -25,7 +25,7 @@ class BumpReleasePlugin(PreBuildPlugin):
     # The target parameter is no longer used by this plugin. It's
     # left as an optional parameter to allow a graceful transition
     # in osbs-client.
-    def __init__(self, tasker, workflow, hub, target=None):
+    def __init__(self, tasker, workflow, hub, target=None, koji_ssl_certs_dir=None):
         """
         constructor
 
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -25,18 +25,24 @@ class BumpReleasePlugin(PreBuildPlugin):
...
-    def __init__(self, tasker, workflow, hub, target=None):
+    def __init__(self, tasker, workflow, hub, target=None, koji_ssl_certs_dir=None):
         """
         constructor
 
         :param tasker: DockerTasker instance
         :param workflow: DockerBuildWorkflow instance
         :param hub: string, koji hub (xmlrpc)
         :param target: unused - backwards compatibility
+        :param koji_ssl_certs_dir: str, path to "cert", "ca", and "serverca"
```
lcarva: Let's call out that these plugins do not need to be authenticated, and this is simply to allow trusting hub's identity cert.

vrutkovs: I was thinking to add all koji params here - just in case this plugin would ever need a proper authentication

lcarva: There's no downside in doing so, I don't hink.

vrutkovs: Added a note about usage of this param

tests/plugins/test_koji.py
```diff
@@ -172,11 +183,11 @@ def test_koji_plugin(self,
...
             'root': root,
-            'proxy': proxy
+            'proxy': proxy,
         }
 
         if koji_ssl_certs:
-            args['koji_ssl_certs'] = str(tmpdir)
+            args['koji_ssl_certs_dir'] = str(tmpdir)
             with open('{}/ca'.format(tmpdir), 'w') as ca_fd:
                 ca_fd.write('ca')
```
twaugh: Is this file ever read during the test? It isn't created for tests for the other plugin.

tests/plugins/test_koji.py
```diff
@@ -83,6 +83,9 @@ def getRepo(self, repo):
             return None
         return GET_REPO_RESPONSE
 
+    def ssl_login(self, cert, ca, serverca, proxyuser=None):
+        return True
```
twaugh: Should this verify the `ca` parameter has the expected value? (The constructor could be told what the expected value is.)

vrutkovs: I'll update the test to make sure `ca`, `cert` and `serverca` are checked, but its pretty complicated, as koji session can't store test vars in `workflow` - and workflow can't reach the particular instance of the plugin

tests/plugins/test_koji.py
vrutkovs: Fixed. Now `ssl_login` saves the received paths in `self` and those are verified in the test

PR #678, opened 2017-04-06
Initialise TaskWatcher state to avoid traceback

twaugh:
Previous behaviour of add_filesystem after cancelling task:
AttributeError: 'TaskWatcher' object has no attribute 'state'

### DIFF
--- a/atomic_reactor/koji_util.py
+++ b/atomic_reactor/koji_util.py
@@ -85,6 +85,7 @@ class TaskWatcher(object):
         self.session = session
         self.task_id = task_id
         self.poll_interval = poll_interval
+        self.state = 'CANCELED'
 
     def wait(self):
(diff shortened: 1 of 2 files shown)


atomic_reactor/koji_util.py
```diff
@@ -85,6 +85,7 @@ def __init__(self, session, task_id, poll_interval=5):
         self.session = session
         self.task_id = task_id
         self.poll_interval = poll_interval
+        self.state = 'CANCELED'
```
lcarva: Why not initialize to None?

twaugh: Because then the `.failed()` method would return False. The point of that method is to know whether we should proceed (really it should probably be named .succeeded()...).

https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/pre_add_filesystem.py#L292

PR #683, opened 2017-04-19
sendmail: return earlier if koji build id or task id are not determined

vrutkovs:
This helps to avoid long tracebacks on downstream Brew instances

### DIFF
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -210,6 +210,9 @@ class SendMailPlugin(ExitPlugin):
             return '@'.join([obj['name'], self.email_domain])
 
     def _get_koji_submitter(self):
+        if not self.koji_task_id:
+            return ""
+
         koji_task_info = self.session.getTaskInfo(self.koji_task_id)
         koji_task_owner = self.session.getUser(koji_task_info['owner'])
         koji_task_owner_email = self._get_email_from_koji_obj(koji_task_owner)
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -210,6 +210,9 @@ def _get_email_from_koji_obj(self, obj):
             return '@'.join([obj['name'], self.email_domain])
 
     def _get_koji_submitter(self):
+        if not self.koji_task_id:
+            return ""
```
twaugh: Won't this add the empty string to the list of addresses?

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -218,6 +221,9 @@ def _get_koji_submitter(self):
 
     def _get_koji_owners(self):
         result = []
+        if not self.koji_build_id:
+            return []
```
twaugh: How about `return result`?

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Ah, yes, indeed. Updated `_get_receivers_list` to remove empty addresses

atomic_reactor/plugins/exit_sendmail.py
vrutkovs: Agree. Added test cases for these code paths

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -269,6 +275,9 @@ def _get_receivers_list(self):
         # Remove duplicates
         receivers_list = list(set(receivers_list))
 
+        # Remove empty and None items
+        receivers_list = [x for x in receivers_list if x]
```
twaugh: Could combine these two as `[x for x in set(receivers_list) if x]` but maybe it's clearer as two lines.

vrutkovs: I 'd rather have this in two separate lines - combining those would produce a hard-to-read mess

PR #684, opened 2017-04-19
URL to info page as docker label

rcerven:

requires osbs-client change https://github.com/projectatomic/osbs-client/pull/553 to work

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -53,9 +53,10 @@ from __future__ import unicode_literals
 from atomic_reactor import start_time as atomic_reactor_start_time
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.constants import INSPECT_CONFIG
-from atomic_reactor.util import get_docker_architecture, _PREFERRED_LABELS, df_parser
+from atomic_reactor.util import get_docker_architecture, df_parser
 import json
 import datetime
+import string
 
 
 class AddLabelsPlugin(PreBuildPlugin):
@@ -83,7 +84,8 @@ class AddLabelsPlugin(PreBuildPlugin):
                               "vcs-ref",
                               "com.redhat.build-host"),
                  aliases=None,
-                 dont_overwrite_if_in_dockerfile=("distribution-scope",)):
+                 dont_overwrite_if_in_dockerfile=("distribution-scope",),
+                 info_url_format=None):
         """
         constructor
 
@@ -98,6 +100,7 @@ class AddLabelsPlugin(PreBuildPlugin):
                         added (with the same value)
         :param dont_overwrite_if_in_dockerfile : iterable, list of label keys which should not be
                                                  overwritten if they are present in dockerfile
+        :param info_url_format : string, format for url dockerfile label
         """
         # call parent constructor
         super(AddLabelsPlugin, self).__init__(tasker, workflow)
@@ -109,6 +112,7 @@ class AddLabelsPlugin(PreBuildPlugin):
         self.dont_overwrite = dont_overwrite
         self.dont_overwrite_if_in_dockerfile = dont_overwrite_if_in_dockerfile
         self.aliases = aliases or self.DEFAULT_ALIASES
+        self.info_url_format = info_url_format
 
         self.generate_auto_labels(auto_labels)
 
@@ -196,6 +200,23 @@ class AddLabelsPlugin(PreBuildPlugin):
             self.log.debug("applied only some aliases, following old labels were not found: %s",
                            ", ".join(not_applied))
 
+    def add_info_url(self, base_labels, df_labels, plugin_labels):
+        all_labels = base_labels.copy()
+        all_labels.update(df_labels)
+        all_labels.update(plugin_labels)
+
+        class MyFormatter(string.Formatter):
+            """
+            using this because str.format can't handle keys with dots and dashes
+            which are included in some of the labels, such as
+            'authoritative-source-url', 'com.redhat.component', etc
+            """
+            def get_field(self, field_name, args, kwargs):
+                return (self.get_value(field_name, args, kwargs), field_name)
+
+        info_url = MyFormatter().vformat(self.info_url_format, [], all_labels)
+        self.labels['url'] = info_url
+
     def run(self):
         """
         run the plugin
@@ -216,6 +237,8 @@ class AddLabelsPlugin(PreBuildPlugin):
         # changing dockerfile.labels writes out modified Dockerfile - err on
         # the safe side and make a copy
         self.add_aliases(base_image_labels.copy(), dockerfile.labels.copy(), self.labels.copy())
+        if self.info_url_format:
+            self.add_info_url(base_image_labels.copy(), dockerfile.labels.copy(), self.labels.copy())
 
         # correct syntax is:
         #   LABEL "key"="value" "key2"="value2"
--- a/tests/plugins/test_add_labels.py
+++ b/tests/plugins/test_add_labels.py
@@ -50,6 +50,14 @@ FROM fedora"""
 DF_CONTENT_LABEL = '''\
 FROM fedora
 LABEL "label2"="df value"'''
+DF_CONTENT_LABELS = '''\
+FROM fedora
+LABEL "label1"="label1_value"
+LABEL "label2"="label2_value"
+LABEL "Authoritative_Registry"="authoritative-source-url_value"
+LABEL "BZComponent"="com.redhat.component_value"
+LABEL "Build_Host"="com.redhat.build-host_value"
+LABEL "Version"="version_value"'''
 LABELS_CONF_BASE = {INSPECT_CONFIG: {"Labels": {"label1": "base value"}}}
 LABELS_CONF_BASE_NONE = {INSPECT_CONFIG: {"Labels": None}}
 LABELS_CONF = OrderedDict({'label1': 'value 1', 'label2': 'long value'})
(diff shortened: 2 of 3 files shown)


atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -81,9 +83,11 @@ def __init__(self, tasker, workflow, labels,
                               "architecture",
                               "vcs-type",
                               "vcs-ref",
-                              "com.redhat.build-host"),
+                              "com.redhat.build-host",
+                              "url"),
                  aliases=None,
-                 dont_overwrite_if_in_dockerfile=("distribution-scope",)):
+                 dont_overwrite_if_in_dockerfile=("distribution-scope",),
+                 info_url_format=None):
```
vrutkovs: Add documentation for this parameter

rcerven:
one concern is that as I am using Labels, which doesn't contains full set of labels, I am allowing substitution only of that limited set of labels

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -126,6 +131,28 @@ def generate_auto_labels(self, auto_labels):
         docker_info = self.tasker.get_info()
         generated['com.redhat.build-host'] = docker_info['Name']
 
+        info_url = self.info_url_format
+        labels = Labels(self.labels)
+        find_subs = re.finditer('{([^}]+)}', info_url)
```
vrutkovs: This looks a lot like a python format string, maybe we could use `self.info_url_format.format(**kwargs)`, where `kwargs` would contain a list of labels?

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -126,6 +131,28 @@ def generate_auto_labels(self, auto_labels):
...
+                if found_label in label_list:
+                    found_type = label_type
+                    break
+
+            if found_type:
+                try:
+                    _, label_value = labels.get_name_and_value(found_type)
+                    info_url = re.sub('{'+ found_label +'}', label_value, info_url)
+                except KeyError:
+                    self.log.warning("label used to construct RHCC url is not defined %s", found_label)
```
vrutkovs: We shouldn't state its an 'RHCC url' - its just a URL with image info, it might be useful for others who don't use RHCC

atomic_reactor/plugins/pre_add_labels_in_df.py
rcerven: well problem is that I don't have such a list, as there could be used mixed (old/new) labels, so I need to use Labels

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -109,6 +113,7 @@ def __init__(self, tasker, workflow, labels,
         self.dont_overwrite = dont_overwrite
         self.dont_overwrite_if_in_dockerfile = dont_overwrite_if_in_dockerfile
         self.aliases = aliases or self.DEFAULT_ALIASES
+        self.info_url_format = info_url_format or DEFAULT_INFO_URL_FORMAT
```
twaugh: I think we should not provide a default for this. If a value is not provided, we should leave the url label alone.

atomic_reactor/plugins/pre_add_labels_in_df.py
rcerven: unless I create directory with every possible key/value combination for new/old labels, using Labels, that way I can blindly pass it to format

twaugh: Yes, that's exactly what I had in mind.

atomic_reactor/plugins/pre_add_labels_in_df.py
rcerven: ok will not use any default

atomic_reactor/plugins/pre_add_labels_in_df.py
vrutkovs: +1 to preparing a list before formatting the string

atomic_reactor/plugins/pre_add_labels_in_df.py
rcerven: where?

vrutkovs: In docstring below, around line 100 (github won't let me create a comment there, as its not part of the diff)

rcerven: ahh right in docstring

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -126,6 +132,29 @@ def generate_auto_labels(self, auto_labels):
         docker_info = self.tasker.get_info()
         generated['com.redhat.build-host'] = docker_info['Name']
 
+        if self.info_url_format:
+            info_url = self.info_url_format
```
lcarva: Nack on implementing our own version of string templating. Why not simply:
```python
info_url = self.info_url_format.format(**labels)
```

atomic_reactor/plugins/pre_add_labels_in_df.py
lcarva: `kwargs` would not be a list of labels, but a dict mapping label name to label value. This is retrieved by `DockerfileParser.labels`. The only reason I can think of for not using that is if we want to support automatic translation of old-style to new-style labels. This to me seems unnecessary as the `info_url_format` is configurable.

```python
>>> from dockerfile_parse import DockerfileParser
>>> parser = DockerfileParser()
>>> parser.labels
{u'foo': u'oof', u'spam': u'maps'}
>>> 'registry.com/{foo}/{spam}'.format(**parser.labels)
'registry.com/oof/maps'
>>> 
```

vrutkovs: +1 to using DockerfileParser, we should move `generate_auto_labels` invocation to `run` then probably

twaugh: Note that if setting 'url' needs to happen in a separate plugin to make this implementation easier, let's do that. We certainly don't have to have *every* label adjustment done in this one plugin -- we already adjust 'release' separately in bump_release, for example.

Let's keep it simple.

atomic_reactor/plugins/pre_add_labels_in_df.py
vrutkovs: This was resolved

tests/plugins/test_add_labels.py
```diff
@@ -354,3 +358,48 @@ def test_dont_overwrite_distribution_scope(tmpdir, docker_tasker, parent_scope,
 
     result = df.labels.get("distribution-scope")
     assert result == result_scope
+
+
+@pytest.mark.parametrize('url_format, info_url, expected_log', [
+    ('url_pre {label1} {label2} url_post', 'url_pre label1_value label2_value url_post', None),
+    ('url_pre url_post', 'url_pre url_post', None),
+    ('url_pre {label1} {label2} {label3_non_existent} url_post', None, "KeyError:"),
```
twaugh: Let's add a test to make sure that a keyword for an alias is substituted correctly, for example "{version}" when the Dockerfile has "Version".

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -216,6 +227,8 @@ def run(self):
         # changing dockerfile.labels writes out modified Dockerfile - err on
         # the safe side and make a copy
         self.add_aliases(base_image_labels.copy(), dockerfile.labels.copy(), self.labels.copy())
+        if self.info_url_format:
+            self.add_info_url(base_image_labels.copy(), dockerfile.labels.copy(), self.labels.copy())
```
vrutkovs: `add_aliases` updates `self.labels`, why can't we use it in `add_info_url`?

tests/plugins/test_add_labels.py
```diff
@@ -354,3 +358,48 @@ def test_dont_overwrite_distribution_scope(tmpdir, docker_tasker, parent_scope,
...
+])
+def test_url_label(tmpdir, docker_tasker, caplog, url_format, info_url, expected_log):
```
vrutkovs: Checking log is not sufficient, we should be checking that exception is not just logged, but actually thrown

atomic_reactor/plugins/pre_add_labels_in_df.py
rcerven: we are using it but we need all labels, which are as well in dockerfile and from base_image

vrutkovs: I'm not sure if we want that. Why would the URL contain a label which doesn't get rendered in this docker image in the end?

tests/plugins/test_add_labels.py
rcerven: plugin is allowed to fail, runner catches exception and no other is thrown

vrutkovs: It is, but currently tests don't check that the exception is thrown

atomic_reactor/plugins/pre_add_labels_in_df.py
rcerven: labels are added to dockerfile, but I need all of them, not just the newly added

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -216,6 +233,8 @@ def run(self):
...
```
lcarva: Shouldn't this happen at a later time, right before `content = ""` ?

rcerven: we are just adding in into self.labels, there still has to be check for don't overwrite

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -196,6 +200,19 @@ def add_as_an_alias(new, old, is_old_inherited):
...
+    def add_info_url(self, base_labels, df_labels, plugin_labels):
+        all_labels = base_labels.copy()
+        all_labels.update(df_labels)
+        all_labels.update(plugin_labels)
+
+        class MyFormatter(string.Formatter):
+            def get_field(self, field_name, args, kwargs):
+                return (self.get_value(field_name, args, kwargs), field_name)
+
+        print(all_labels)
```
twaugh: Don't want this line. :)

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -196,6 +200,19 @@ def add_as_an_alias(new, old, is_old_inherited):
...
                            ", ".join(not_applied))
 
+    def add_info_url(self, base_labels, df_labels, plugin_labels):
+        all_labels = base_labels.copy()
+        all_labels.update(df_labels)
+        all_labels.update(plugin_labels)
+
+        class MyFormatter(string.Formatter):
+            def get_field(self, field_name, args, kwargs):
+                return (self.get_value(field_name, args, kwargs), field_name)
```
twaugh: Could add a comment (maybe a class docstring on that class?) saying why we need to do this rather than just calling str.format().

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -196,6 +200,23 @@ def add_as_an_alias(new, old, is_old_inherited):
...
+        class MyFormatter(string.Formatter):
+            """
+            using this because str.format can't handle keys with dots and dashes
+            which are included in some of the labels, such as
+            'authoritative-source-url', 'com.redhat.component', etc
+            """
+            def get_field(self, field_name, args, kwargs):
+                return (self.get_value(field_name, args, kwargs), field_name)
+
+        info_url = MyFormatter().vformat(self.info_url_format, [], all_labels)
```
lcarva: sleek!

PR #686, opened 2017-04-25
Add fetch_maven_artifacts plugin

lcarva:

### DIFF
--- a/atomic_reactor/plugins/pre_fetch_maven_artifacts.py
+++ b/atomic_reactor/plugins/pre_fetch_maven_artifacts.py
@@ -1,0 +1,228 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+from __future__ import unicode_literals
+
+import fnmatch
+import hashlib
+import os
+import requests
+
+from atomic_reactor import util
+from atomic_reactor.koji_util import create_koji_session
+from atomic_reactor.plugin import PreBuildPlugin
+from collections import namedtuple
+from koji import PathInfo
+
+
+try:
+    from urlparse import urlparse
+except ImportError:
+    from urllib.parse import urlparse
+
+
+class NvrRequest(object):
+
+    def __init__(self, nvr, archives=None):
+        self.nvr = nvr
+        self.archives = archives or []
+
+        for archive in self.archives:
+            archive['matched'] = False
+
+    def match(self, build_archive):
+        if not self.archives:
+            return True
+
+        for archive in self.archives:
+            req_filename = archive.get('filename')
+            req_group_id = archive.get('group_id')
+
+            if req_filename and not fnmatch.filter([build_archive['filename']],
+                                                   req_filename):
+                continue
+
+            if req_group_id and req_group_id != build_archive['group_id']:
+                continue
+
+            archive['matched'] = True
+            return True
+
+        return False
+
+    def match_all(self, build_archives):
+        return filter(self.match, build_archives)
+
+    def unmatched(self):
+        return [archive for archive in self.archives if not archive['matched']]
+
+
+DownloadRequest = namedtuple('DownloadRequest', 'url dest md5')
+
+
+class FetchMavenArtifactsPlugin(PreBuildPlugin):
+
+    key = 'fetch_maven_artifacts'
+    is_allowed_to_fail = False
+
+    NVR_REQUESTS_FILENAME = 'fetch-artifacts-koji.yaml'
+    URL_REQUESTS_FILENAME = 'fetch-artifacts-url.yaml'
+
+    DOWNLOAD_DIR = 'artifacts'
+
+    def __init__(self, tasker, workflow, koji_hub, koji_root,
+                 koji_proxyuser=None, koji_ssl_certs_dir=None,
+                 koji_krb_principal=None, koji_krb_keytab=None,
+                 allowed_domains=None):
+        """
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param koji_hub: str, koji hub (xmlrpc)
+        :param koji_root: str, koji root (storage)
+        :param koji_proxyuser: str, proxy user
+        :param koji_ssl_certs_dir: str, path to "cert", "ca", and "serverca"
+        :param koji_krb_principal: str, name of Kerberos principal
+        :param koji_krb_keytab: str, Kerberos keytab
+        :param allowed_domains: list<str>: list of domains that are
+               allowed to be used when fetching artifacts by URL (case insensitive)
+        """
+        super(FetchMavenArtifactsPlugin, self).__init__(tasker, workflow)
+        koji_auth = {
+            'proxyuser': koji_proxyuser,
+            'ssl_certs_dir': koji_ssl_certs_dir,
+            'krb_principal': koji_krb_principal,
+            'krb_keytab': koji_krb_keytab,
+        }
+        # Remove empty values from auth dict to avoid login when not needed
+        koji_auth = {k: v for k, v in koji_auth.items() if v}
+        self.koji_info = {
+            'hub': koji_hub,
+            'root': koji_root,
+            'auth': koji_auth or None
+        }
+        self.path_info = PathInfo(topdir=self.koji_info['root'])
+        self.allowed_domains = set(domain.lower() for domain in allowed_domains or [])
+        self.workdir = self.workflow.source.get_dockerfile_path()[1]
+        self.session = None
+
+    def read_nvr_requests(self):
+        file_path = os.path.join(self.workdir, self.NVR_REQUESTS_FILENAME)
+        if not os.path.exists(file_path):
+            self.log.debug('%s not found', self.NVR_REQUESTS_FILENAME)
+            return []
+
+        nvr_requests = util.read_yaml(file_path, 'schemas/fetch-artifacts-nvr.json')
+        return [NvrRequest(**nvr_request) for nvr_request in nvr_requests]
+
+    def read_url_requests(self):
+        file_path = os.path.join(self.workdir, self.URL_REQUESTS_FILENAME)
+        if not os.path.exists(file_path):
+            self.log.debug('%s not found', self.URL_REQUESTS_FILENAME)
+            return []
+
+        return util.read_yaml(file_path, 'schemas/fetch-artifacts-url.json')
+
+    def process_by_nvr(self, nvr_requests):
+        download_queue = []
+        errors = []
+
+        for nvr_request in nvr_requests:
+            build_info = self.session.getBuild(nvr_request.nvr)
+            if not build_info:
+                errors.append('Build {} not found.'.format(nvr_request.nvr))
+                continue
+
+            maven_build_path = self.path_info.mavenbuild(build_info)
+            build_archives = self.session.listArchives(buildID=build_info['id'],
+                                                       type='maven')
+            build_archives = nvr_request.match_all(build_archives)
+
+            for build_archive in build_archives:
+                maven_file_path = self.path_info.mavenfile(build_archive)
+                # NOTE: Don't use urljoin here because maven_build_path does
+                # not contain a trailing slash, which causes the last dir to
+                # be dropped.
+                url = maven_build_path + '/' + maven_file_path
+                download_queue.append(DownloadRequest(url, maven_file_path,
+                                                      build_archive['checksum']))
+
+            unmatched_archive_requests = nvr_request.unmatched()
+            if unmatched_archive_requests:
+                errors.append('NVR request for "{}", failed to find archives for: "{}"'
+                              .format(nvr_request.nvr, unmatched_archive_requests))
+                continue
+
+        if errors:
+            raise ValueError('Errors found while processing {}: {}'
+                             .format(self.NVR_REQUESTS_FILENAME, ', '.join(errors)))
+        return download_queue
+
+    def process_by_url(self, url_requests):
+        download_queue = []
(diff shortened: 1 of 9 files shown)


atomic_reactor/plugins/pre_fetch_maven_artifacts.py
```diff
@@ -0,0 +1,216 @@
...
+
+            maven_build_path = self.path_info.mavenbuild(build_info)
+            build_archives = self.session.listArchives(buildID=build_info['id'],
+                                                       type='maven')
+            build_archives = nvr_request.match_all(build_archives)
+
+            for build_archive in build_archives:
+                maven_file_path = self.path_info.mavenfile(build_archive)
+                # NOTE: Don't use urljoin here because maven_build_path does
+                # not contain a traling slash, which causes the last dir to
```
twaugh: Couldn't we urljoin `maven_build_path + '/'` and `maven_file_path`? What if maven_build_path *does* contain a trailing slash in future?

(also, typo: traling)

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
```diff
@@ -0,0 +1,216 @@
...
+                # be dropped.
+                url = maven_build_path + '/' + maven_file_path
+                download_queue.append(DownloadRequest(url, maven_file_path,
+                                                      build_archive['checksum']))
```
twaugh: Would there be any benefit in making this a generator function? (Although, not sure how would ever collect all errors in that case.)

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
```diff
@@ -0,0 +1,216 @@
...
+            download_queue.append(DownloadRequest(url, target, checksum))
+
+        return download_queue
+
+    def download_files(self, downloads):
+        artifacts_path = os.path.join(self.workdir, self.DOWNLOAD_DIR)
+
+        self.log.debug('%d files to download', len(downloads))
+
+        for index, download in enumerate(downloads):
```
twaugh: I think the answer is no, but is there any benefit in using a thread pool here to have multiple downloads in progress at once?

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -128,48 +125,7 @@ def run(self):
...
-                if path.startswith('.'):
-                    path = path[1:]
-
-                self.log.error("validation error (%s): %s",
-                               path or "at top level",
-                               error.message)
-
-            raise
-
+        conf = read_yaml(config_filename, schema='schemas/config.json')
```
twaugh: This could be its own commit, but if that's hard to split out it's probably fine as it is.

tests/plugins/test_fetch_maven_artifacts.py
```diff
@@ -0,0 +1,692 @@
...
+    if not archives:
+        archives = DEFAULT_ARCHIVES
+
+    flexmock(koji, PathInfo=MockedPathInfo)
+
+    session = flexmock()
+
+    (flexmock(koji)
+        .should_receive('ClientSession')
+        # .once()
```
twaugh: Why is this commented out?

tests/plugins/test_fetch_maven_artifacts.py
```diff
@@ -0,0 +1,692 @@
...
+
+    with pytest.raises(PluginFailedException) as e:
+        runner.run()
+
+    assert 'ValidationError' in str(e)
+
+
+@responses.activate
+def test_fetch_maven_artifacts_url_with_target(tmpdir, docker_tasker):
+    """Err when downloaded remote file has unexpected checksum."""
```
twaugh: Is this docstring right?

atomic_reactor/schemas/fetch-artifacts-nvr.json
```diff
@@ -0,0 +1,56 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+
+  "title": "Artifacts to be fetched from Koji builds.",
```
twaugh: We should be consistent (with other files in schemas/) in whether we use a full-stop in the title and descriptions. Currently config.json doesn't use them.

atomic_reactor/schemas/fetch-artifacts-nvr.json
```diff
@@ -0,0 +1,56 @@
...
+
+  "type": "array",
+
+  "items": {
+
+    "type": "object",
+
+		"description": "A specific Koji buid and archives to be fetched.",
```
twaugh: Something up with the indentation here?

Also, type: buid

atomic_reactor/schemas/fetch-artifacts-nvr.json
```diff
@@ -0,0 +1,56 @@
...
+        "type": "array",
+
+        "items": {
+
+          "type": "object",
+
+          "description": "Matcher for archives in Koji build.",
+
+          "properties": {
+
```
twaugh: Why so many blank lines?

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
```diff
@@ -0,0 +1,216 @@
...
+
+        nvr_requests = self.read_nvr_requests()
+        url_requests = self.read_url_requests()
+
+        download_queue = (self.process_by_nvr(nvr_requests) +
+                          self.process_by_url(url_requests))
+
+        self.download_files(download_queue)
+
+        # TODO: Return a list of files for koji metadata
```
twaugh: Yes, until the format is decided I think this is fine as-is for this PR.

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
```diff
@@ -0,0 +1,216 @@
...
+    def process_by_url(self, url_requests):
+        download_queue = []
+
+        for url_request in url_requests:
+            url = url_request['url']
+
+            if self.allowed_domains:
+                netloc = urlparse(url).netloc.lower()
+                if netloc not in self.allowed_domains:
+                    raise ValueError('Domain {} is not in list of allowed domains: {}'
```
rcerven: maybe we can as well first collect all errors and raise after

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
```diff
@@ -0,0 +1,216 @@
...
+        return util.read_yaml(file_path, default=[],
+                              schema='schemas/fetch-artifacts-url.json')
+
+    def process_by_nvr(self, nvr_requests):
+        download_queue = []
+
+        for nvr_request in nvr_requests:
+            build_info = self.session.getBuild(nvr_request.nvr)
+            if not build_info:
+                raise ValueError('Build {} not found.'.format(nvr_request.nvr))
```
vrutkovs: Perhaps some errors should be the showstopper - i.e its dangerous to download jars from untrusted domain

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
rcerven: well it depends from where it will be downloading, but unless each source has upload speed limit, we are limited just by our download speed, which isn't as well limited per download, so one-by one is fine without any limitations

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
lcarva: Yes, we could do that. I didn't think it was worth it. String concatenation is what koji does btw: https://pagure.io/koji/blob/master/f/cli/koji#_6721

Should I change it?

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
lcarva: Usually, a generator is used to allow processing items while the collection itself is still being created. I think in this case, we don't want that. I'd rather make sure that all the DownloadRequest instances have been created *before* actually start any downloads.

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
lcarva: It's calculating the checksum (CPU) as it's downloading (Network). A thread pool wouldn't be beneficial for splitting up checksum calculations. If going down this approach, a process pool is better suited. I'm just not sure if this is overkill at this point.

atomic_reactor/plugins/pre_reactor_config.py
lcarva: I'll split it up.

tests/plugins/test_fetch_maven_artifacts.py
lcarva: Leftover from debugging. Uncommented out.

tests/plugins/test_fetch_maven_artifacts.py
lcarva: Fixed it.

atomic_reactor/schemas/fetch-artifacts-nvr.json
lcarva: Agree. Removed punctuation.

atomic_reactor/schemas/fetch-artifacts-nvr.json
lcarva: Tabs! Also fixed typo.

atomic_reactor/schemas/fetch-artifacts-nvr.json
lcarva: My futile approach into making this file more readable. Changed it so it's more consistent with config.json.

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
twaugh: It's probably fine.

atomic_reactor/plugins/pre_reactor_config.py
lcarva: Done.

atomic_reactor/schemas/fetch-artifacts-nvr.json
lcarva: Done.

atomic_reactor/schemas/fetch-artifacts-nvr.json
lcarva: Done.

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
lcarva: I'll change it so it collects all errors for better user experience.

@vrutkovs I still think we should fail if the user specified a build that doesn't exist.

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
```diff
@@ -0,0 +1,216 @@
...
+                    f.write(chunk)
+                    checksum.update(chunk)
+
+            if checksum.hexdigest() != download.md5:
+                raise ValueError('Computed md5 checksum, {}, does not match '
+                                 'expected checksum, {}'
+                                 .format(checksum.hexdigest(), download.md5))
+
+    def run(self):
+        self.session = create_koji_session(self.koji_info['hub'], self.koji_info.get('auth'))
```
twaugh: pylint suggests defining this in the constructor (i.e. self.session = None).

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
twaugh: We definitely should fail on error as currently. Just might be nice to report all the errors, then fail on one of them (the first?).

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
twaugh: Agreed.

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
lcarva: Done.

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
lcarva: Agreed, modified to collect all errors before raising exception.

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
lcarva: Agreed, modified to collect all errors prior to raising exception.

atomic_reactor/util.py
```diff
@@ -862,3 +866,49 @@ def are_plugins_in_order(plugins_conf, *plugins_names):
...
             return False
     return True
+
+
+def read_yaml(yaml_file_path, default=None, schema=None):
+    with open(yaml_file_path) as f:
+        data = yaml.safe_load(f)
+
+    if not schema:
+        return data
```
twaugh: Is this line ever hit?

lcarva: Hmm I guess not. I'll add a unit test for it.
I still think there's value in keep the functionality though.

lcarva: Actually, we don't have a use case for that. I'm going to make schema required to encourage always using a schema :)

Also noticed that default parameter is not used. This is leftover from a previous version. Removing it as well.

lcarva: Done!

twaugh:
Not sure whether we want to merge this into master yet.

lcarva:
Going over the design doc I realized the `allowed_domains` implementation was a little off. We want to be able to check the full URL prefix, not just the `netloc`. Updated accordingly, PTAL.

lcarva:
@twaugh docs updated. Let me know if that's sufficient.

PR #688, opened 2017-04-26
Remove code paths for py 2.6

vrutkovs:
I've also checked that no tested code paths are deprecated using 
PYTHONWARNINGS=default env var

### DIFF
--- a/atomic_reactor/plugins/pre_inject_yum_repo.py
+++ b/atomic_reactor/plugins/pre_inject_yum_repo.py
@@ -129,9 +129,7 @@ class InjectYumRepoPlugin(PreBuildPlugin):
         """
         run the plugin
         """
-        # dict comprehension is syntax error on 2.6
-        yum_repos = dict((k, v) for (k, v) in self.workflow.files.items()
-                                           if k.startswith(YUM_REPOS_DIR))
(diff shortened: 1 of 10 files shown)


twaugh:
Looks great! Other possibilities:

- urllib.parse
- backports.lzma
- StringIO
- BytesIO
- configparser
- six.integer_types?

vrutkovs:
All of those are for py2 and py3 compatibility.

Other possible ways to modernize the code (from [changelog](https://docs.python.org/3/whatsnew/2.7.html)):
* Dict and set comprehensions (`dict()` vs `{k: v for k, v in ..}`)
  Its debatable, imho `set(1, 2, 3)` would be converted to `{1, 2, 3}` - which makes the code less readable
* Nested context managers
  Found a couple of instances of that pattern, going to update the PR

rcerven:
so far looks good

atomic_reactor/plugins/pre_inject_yum_repo.py
```diff
@@ -130,8 +130,7 @@ def run(self):
         run the plugin
         """
         # dict comprehension is syntax error on 2.6
```
lcarva: Let's remove the comment.

vrutkovs: Fixed

PR #690, opened 2017-04-28
add equal label aliases

rcerven:

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -54,6 +54,7 @@ from atomic_reactor import start_time as atomic_reactor_start_time
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.constants import INSPECT_CONFIG
 from atomic_reactor.util import get_docker_architecture, df_parser
+from osbs.utils import Labels
 import json
 import datetime
 import string
@@ -62,20 +63,6 @@ import string
 class AddLabelsPlugin(PreBuildPlugin):
     key = "add_labels_in_dockerfile"
     is_allowed_to_fail = False
-
-    DEFAULT_ALIASES = {
-        'Name': 'name',
-        'Version': 'version',
-        'Release': 'release',
-        'Architecture': 'architecture',
-        'Vendor': 'vendor',
-        'RUN': 'run',
-        'INSTALL': 'install',
-        'UNINSTALL': 'uninstall',
-        'Authoritative_Registry': 'authoritative-source-url',
-        'BZComponent': 'com.redhat.component',
-        'Build_Host': 'com.redhat.build-host',
-    }
 
     def __init__(self, tasker, workflow, labels,
                  dont_overwrite=("Architecture", "architecture"),
@@ -86,7 +73,8 @@ class AddLabelsPlugin(PreBuildPlugin):
                               "com.redhat.build-host"),
                  aliases=None,
                  dont_overwrite_if_in_dockerfile=("distribution-scope",),
-                 info_url_format=None):
+                 info_url_format=None,
+                 equal_labels=None):
         """
         constructor
 
@@ -102,6 +90,7 @@ class AddLabelsPlugin(PreBuildPlugin):
         :param dont_overwrite_if_in_dockerfile : iterable, list of label keys which should not be
                                                  overwritten if they are present in dockerfile
         :param info_url_format : string, format for url dockerfile label
+        :param equal_labels: list, with equal labels groups as lists
         """
         # call parent constructor
         super(AddLabelsPlugin, self).__init__(tasker, workflow)
@@ -112,8 +101,11 @@ class AddLabelsPlugin(PreBuildPlugin):
         self.labels = labels
         self.dont_overwrite = dont_overwrite
         self.dont_overwrite_if_in_dockerfile = dont_overwrite_if_in_dockerfile
-        self.aliases = aliases or self.DEFAULT_ALIASES
+        self.aliases = aliases or Labels.get_new_names_by_old()
         self.info_url_format = info_url_format
+        self.equal_labels = equal_labels or []
+        if not isinstance(self.equal_labels, list):
+            raise RuntimeError("equal_labels have to be list")
 
         self.generate_auto_labels(auto_labels)
 
@@ -184,6 +176,7 @@ class AddLabelsPlugin(PreBuildPlugin):
             if new in new_labels:
                 if all_labels[old] != all_labels[new]:
                     if is_old_inherited:
+                        # set old label with value from new label if old was in base
                         add_as_an_alias(new, old, is_old_inherited)
                         continue
                     self.log.warning("labels %r=%r and %r=%r should probably have same value",
@@ -192,6 +185,7 @@ class AddLabelsPlugin(PreBuildPlugin):
                 self.log.debug("alias label %r for %r already exists, skipping", new, old)
                 continue
 
+            # new label is in base or doesn't exists and we have somewhere old label
             add_as_an_alias(new, old, is_old_inherited)
             applied_alias = True
             self.log.info(self.labels)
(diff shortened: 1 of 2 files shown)


rcerven:
for tests to pass, osbs-client PR has to be merged 1st https://github.com/projectatomic/osbs-client/pull/558

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -201,6 +190,56 @@ def add_as_an_alias(new, old, is_old_inherited):
...
+                    fail_build = True
+                    continue
+
+                # not all set
+                if found_labels_new != all_equal and not fail_build:
+                    set_missing_labels(found_labels_new, all_equal, new_labels)
+
+            elif found_labels_base:
+                if not check_if_all_same(found_labels_base, base_labels):
+                    self.log.error("labels in parent doesn't have same values %s", equal_list)
```
vrutkovs: typo, should be "don't"

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -201,6 +190,56 @@ def add_as_an_alias(new, old, is_old_inherited):
...
+            found_labels_new = set()
+            for equal_label in equal_list:
+                if equal_label in new_labels:
+                    found_labels_new.add(equal_label)
+                elif equal_label in base_labels:
+                    found_labels_base.add(equal_label)
+
+            if found_labels_new:
+                if not check_if_all_same(found_labels_new, new_labels):
+                    self.log.error("labels in dockerfile doesn't have same values %s", equal_list)
```
vrutkovs: typo, should be "don't"

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -201,6 +190,56 @@ def add_as_an_alias(new, old, is_old_inherited):
...
+                    set_missing_labels(found_labels_new, all_equal, new_labels)
+
+            elif found_labels_base:
+                if not check_if_all_same(found_labels_base, base_labels):
+                    self.log.error("labels in parent doesn't have same values %s", equal_list)
+                    fail_build = True
+                    continue
+
+                # not all set
+                if found_labels_base != all_equal and not fail_build:
```
vrutkovs: Perhaps its safe to leave this as `if not fail_build`: the `.difference(` call in `set_missing_labels` would return `[]` and return

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -201,6 +190,59 @@ def add_as_an_alias(new, old, is_old_inherited):
...
+                elif equal_label in base_labels:
+                    found_labels_base.add(equal_label)
+
+            if found_labels_new:
+                if not check_if_all_same(found_labels_new, new_labels):
+                    self.log.error("labels in dockerfile don't have same values %s", equal_list)
+                    fail_build = True
+                    continue
+
+                if not fail_build:
```
lcarva: Unnecessary `if-statement`. If it got here, `fail_build` will always be `False`

lcarva: Hmm I guess that's not true, because it may be the next iteration. Disregard previous comment.

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -201,6 +190,59 @@ def add_as_an_alias(new, old, is_old_inherited):
             self.log.debug("applied only some aliases, following old labels were not found: %s",
                            ", ".join(not_applied))
 
+        def check_if_all_same(labels_to_check, source):
+            if labels_to_check:
+                label_value = source[next(iter(labels_to_check))]
```
lcarva: Why use this instead of `label_value = labels_to_check[0]` ? Both seem to work the same way for empty iterablels:

```python
>>> l = []
>>> l[0]
IndexError: list index out of range
list index out of range
>>> next(iter(l))
StopIteration
```

rcerven: I am using sets not list

rcerven: which isn't iterable

atomic_reactor/plugins/pre_add_labels_in_df.py
rcerven: yes, and I don't want to set anymore anything once I've find out that something is wrong

atomic_reactor/plugins/pre_add_labels_in_df.py
vrutkovs: I think `check_if_all_same` should convert the argument to a list so that it would be easier to undestand that only the first value is used for comparison

vrutkovs: Another option is to check how many different values are stored in `source`:
```
def check_if_all_same(labels_to_check, source):
    source_values = set([source[x] for x in iter(labels_to_check)])
    return len(source_values) == 1
```

rcerven: yeah both are viable options, I just wanted to avoid to create additional list or set, and fail as soon I've find any non equal

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -86,7 +73,8 @@ def __init__(self, tasker, workflow, labels,
                               "com.redhat.build-host"),
                  aliases=None,
                  dont_overwrite_if_in_dockerfile=("distribution-scope",),
-                 info_url_format=None):
+                 info_url_format=None,
+                 equal_labels=[]):
```
twaugh: Shouldn't really declare default values as mutable objects such as lists. Instead, better to use `None` as the default value here and `or []` when setting the instance variable.

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -201,6 +195,61 @@ def add_as_an_alias(new, old, is_old_inherited):
             self.log.debug("applied only some aliases, following old labels were not found: %s",
                            ", ".join(not_applied))
 
+        def check_if_all_same(labels_to_check, source):
```
twaugh: I can't really follow what this method is trying to do. Could we have a docstring for it?

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -201,6 +195,61 @@ def add_as_an_alias(new, old, is_old_inherited):
...
+            if labels_to_check:
```
twaugh: Why do we need to explicitly test this? Won't it work correctly regardless?

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -201,6 +195,61 @@ def add_as_an_alias(new, old, is_old_inherited):
...
+                list_to_check = list(labels_to_check)
+
+                for equal_name in list_to_check[1:]:
+                    if source[list_to_check[0]] != source[equal_name]:
+                        return False
+            return True
+
+        def set_missing_labels(labels_set, all_labels, value_from, not_in=[], not_value=None):
```
twaugh: Default values should not be mutable.

atomic_reactor/plugins/pre_add_labels_in_df.py
rcerven: ok

atomic_reactor/plugins/pre_add_labels_in_df.py
rcerven: ok

atomic_reactor/plugins/pre_add_labels_in_df.py
rcerven: it won't because in set_missing_labels we are using value of 1st found equal label, so we have to do pre-check that all specified have same values, if not we raise

atomic_reactor/plugins/pre_add_labels_in_df.py
twaugh: Thanks, that's better.

rcerven:
added also test with 3 equal labels

tests/plugins/test_add_labels.py
```diff
@@ -302,6 +302,157 @@ def test_add_labels_aliases(tmpdir, docker_tasker, caplog,
     if expected_log:
         assert expected_log in caplog.text()
 
+@pytest.mark.parametrize('base_fst, base_snd, df_fst, df_snd, expected, expected_log', [
```
vrutkovs: How to fix `F811 redefinition of unused 'docker_tasker' from line 23`:
* Put `  # noqa` on that line
* Add a comment 'docker_tasker is a fixture, so its not a redefinition' above

PR #691, opened 2017-04-28
pre_add_labels_in_df: remove extra copy()

vrutkovs:
No need to copy label dicts when passing them to `add_aliases` or `add_info_url` - they
are being copied in these functions


vrutkovs:
This was moved to #692

PR #692, opened 2017-04-28
add_labels_in_df: several fixes

vrutkovs:
* remove extra `copy()`
* don't pass plugin_labels -  these are available in `self.labels`
* fix `add_as_an_alias` to show which label was added as an alias (previously it was showing that new was added as an alias always, although sometimes it was in reverse)
* make sure `applied_alias` is set only when `add_as_an_alias` is called

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -236,9 +236,9 @@ class AddLabelsPlugin(PreBuildPlugin):
 
         # changing dockerfile.labels writes out modified Dockerfile - err on
         # the safe side and make a copy
-        self.add_aliases(base_image_labels.copy(), dockerfile.labels.copy())
+        self.add_aliases(base_image_labels, dockerfile.labels)
         if self.info_url_format:
-            self.add_info_url(base_image_labels.copy(), dockerfile.labels.copy())
+            self.add_info_url(base_image_labels, dockerfile.labels)
 
         # correct syntax is:
         #   LABEL "key"="value" "key2"="value2"
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -154,13 +154,13 @@ class AddLabelsPlugin(PreBuildPlugin):
             else:
                 self.log.warning("requested automatic label %r is not available", lbl)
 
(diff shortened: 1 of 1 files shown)


atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -154,57 +154,54 @@ def generate_auto_labels(self, auto_labels):
...
                 not_applied.append(old)
                 continue
 
             is_old_inherited = old in base_labels and \
-                old not in df_labels and \
-                old not in plugin_labels
+                old not in new_labels
+
+            if is_old_inherited and new in all_labels:
+                add_as_an_alias(new, old)
```
rcerven: if all_labels[old] != all_labels[new]:
    add_as_an_alias(new, old)
    continue

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -154,57 +154,54 @@ def generate_auto_labels(self, auto_labels):
...
 
             if new in new_labels:
                 if all_labels[old] != all_labels[new]:
-                    if is_old_inherited:
-                        add_as_an_alias(new, old, is_old_inherited)
-                        continue
                     self.log.warning("labels %r=%r and %r=%r should probably have same value",
```
rcerven: log this only when

if not is_old_inherited:

we want to notify only in case that old & new in new_labels have different values

vrutkovs: Fixed

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -154,57 +154,54 @@ def generate_auto_labels(self, auto_labels):
...
+                if not is_old_inherited and all_labels[old] != all_labels[new]:
                     self.log.warning("labels %r=%r and %r=%r should probably have same value",
                                      old, all_labels[old], new, all_labels[new])
 
                 self.log.debug("alias label %r for %r already exists, skipping", new, old)
                 continue
 
-            add_as_an_alias(new, old, is_old_inherited)
-            applied_alias = True
+            if not is_old_inherited or new not in base_labels:
```
rcerven: change to the 

           else:
                if new not in all_labels:
                    add_as_an_alias(old, new)
                elif all_labels[old] != all_labels[new]:
                    add_as_an_alias(old, new)

vrutkovs:
This is not needed anymore

PR #693, opened 2017-05-03
tito: store version in version.py and generate it on a new release

vrutkovs:

### DIFF
--- a/atomic_reactor/__init__.py
+++ b/atomic_reactor/__init__.py
@@ -12,8 +12,8 @@ constants
 import logging
 import time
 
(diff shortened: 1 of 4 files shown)


atomic_reactor/__init__.py
```diff
@@ -12,8 +12,8 @@
 import logging
 import time
 
+from osbs.version import __version__
```
rcerven: why from osbs ???

vrutkovs: Oops, copypaste failed me again.

Fixed

PR #694, opened 2017-05-04
orchestration: don't set release for base image builds

vrutkovs:
Inspecting a dockerfile for base image before add_filesystem task has ran would 
crash with 'No such image: koji/image-build:latest'. This PR makes sure this 
method is not called and a bogus release is returned.

This is a temporary solution until `df_parser` is not fixed

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -13,6 +13,7 @@ import json
 import logging
 import tempfile
 import signal
+import docker
 
 from atomic_reactor.build import InsideBuilder
 from atomic_reactor.plugin import (
@@ -327,7 +328,13 @@ class DockerBuildWorkflow(object):
     @property
     def base_image_inspect(self):
         if self._base_image_inspect is None:
-            self._base_image_inspect = self.builder.tasker.inspect_image(self.builder.base_image)
+            try:
+                self._base_image_inspect = self.builder.tasker.inspect_image(
+                    self.builder.base_image)
+            except docker.errors.NotFound:
+                # If the base image cannot be found throw KeyError - as this property should behave
+                # like a dict
+                raise KeyError("Unprocessed base image Dockerfile cannot be inspected")
         return self._base_image_inspect
 
     def throw_canceled_build_exception(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


lcarva:
Update based on offline communication: The more desirable approach is to change `util.df_parser` to handle the case where the parent image has not been pulled previously. 

Looking at `df_parser` it appears that it was intended to always be "safe" - swallowing various Exceptions. The problem is that it's assuming `workflow.base_image_inspect[INSPECT_CONFIG]` will throw a `KeyError` if image is not found. Instead, this is a property that uses `docker-py` to fetch image info. In the case that it's not found, `docker.error.NotFound` is raised.

We should instead either 1) catch `docker.error.NotFound` in `df_parser`; or, 2) catch it in `base_image_inspect` property and wrap it in a `KeyError` exception. The second approach seems more reasonable as the property is meant to behave as a dict.

atomic_reactor/inner.py
```diff
@@ -327,7 +328,13 @@ def build_process_failed(self):
     @property
     def base_image_inspect(self):
         if self._base_image_inspect is None:
-            self._base_image_inspect = self.builder.tasker.inspect_image(self.builder.base_image)
+            try:
+                self._base_image_inspect = self.builder.tasker.inspect_image(
+                    self.builder.base_image)
+            except docker.errors.NotFound:
+                # If the base image cannot be found throw KeyError - as this property should behave
```
lcarva: Hey guy! Max is 99 here. :)

vrutkovs: I think adding `.flake8` file would fix it

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -163,7 +163,7 @@ def run(self):
 
         try:
             base_image_id = self.workflow.base_image_inspect['Id']
-        except docker.errors.NotFound:
+        except (docker.errors.NotFound, KeyError):
```
rcerven: only KeyError should be here, as base_image_inspect is reraising it

vrutkovs: Right, fixed

PR #696, opened 2017-05-11
don't log error when build image is already deleted

rcerven:


vrutkovs:
I don't really see a reason for that. This message is helpful, although it should probably be a `self.log.debug` or `info`

twaugh:
@vrutkovs, this gets logged at the end of every orchestrator build. It's a stale message from before this plugin was also able to remove other images than the built image, but just never got removed.

Note that each image it removes causes an INFO and a DEBUG log anyway, so it's quite clear which things are and are not being removed.

vrutkovs:
Oh, I see. That makes sense

Issue #698, opened 2017-05-24
Cannot build from scratch images with Koji/OSBS/atomic-reactor


nipakoo:
Unable to build with base image set to 'scratch' utilizing Koji, OSBS and atomic-reactor tool set at least without OSBS configuration adaptations. Would like to have atomic-reactor libary to recognize the special properties of a scratch image and adapt to the limitations that this sets i.e. we cannot perform all the checks and fetch same metadata as with other base images.

twaugh:
This should be fixed on the master branches (and in the latest releases) now.

PR #699, opened 2017-05-24
Recognize the special status of scratch image. Adjust plugins to chec…

nipakoo:
Initial support for allowing build with scratch image base.
https://github.com/projectatomic/atomic-reactor/issues/698

### DIFF
--- a/atomic_reactor/plugins/prepub_squash.py
+++ b/atomic_reactor/plugins/prepub_squash.py
@@ -14,7 +14,8 @@ import os
 from atomic_reactor.constants import EXPORTED_SQUASHED_IMAGE_NAME
 from atomic_reactor.plugin import PrePublishPlugin
 from atomic_reactor.plugins.exit_remove_built_image import defer_removal
-from atomic_reactor.util import get_exported_image_metadata
+from atomic_reactor.util import get_exported_image_metadata, \
+        base_image_is_scratch
 from docker_squash.squash import Squash
 
 __all__ = ('PrePublishSquashPlugin', )
@@ -70,6 +71,10 @@ class PrePublishSquashPlugin(PrePublishPlugin):
         self.image = self.workflow.builder.image_id
         self.tag = tag or str(self.workflow.builder.image)
         self.from_layer = from_layer
+
+        if base_image_is_scratch(self.workflow.builder.base_image.to_str()):
+            from_base = False
+
         if from_base and from_layer is None:
             try:
                 base_image_id = self.workflow.base_image_inspect['Id']
--- a/atomic_reactor/build.py
+++ b/atomic_reactor/build.py
@@ -13,7 +13,7 @@ import json
 
 import logging
 from atomic_reactor.core import DockerTasker, LastLogger
-from atomic_reactor.util import ImageName, print_version_of_tools, df_parser
+from atomic_reactor.util import ImageName, print_version_of_tools, df_parser, base_image_is_scratch
 
 logger = logging.getLogger(__name__)
 
@@ -154,7 +154,11 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
         :return: dict
         """
         logger.info("inspecting base image '%s'", self.base_image)
-        inspect_data = self.tasker.inspect_image(self.base_image)
+        if base_image_is_scratch(self.base_image.to_str()):
+            logger.debug("base image is scratch, setting empty inspect data")
+            inspect_data = {}
+        else:
+            inspect_data = self.tasker.inspect_image(self.base_image)
         return inspect_data
 
     def inspect_built_image(self):
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -28,7 +28,7 @@ from atomic_reactor.plugin import (
     PrePublishPluginsRunner,
 )
 from atomic_reactor.source import get_source_instance_for
-from atomic_reactor.util import ImageName
+from atomic_reactor.util import ImageName, base_image_is_scratch
 from atomic_reactor.build import BuildResult
 from atomic_reactor import get_logging_encoding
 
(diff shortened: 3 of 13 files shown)


lcarva:
Thanks for this!

There's currently an existing way of creating base images with `atomic-reactor`. 
Unfortunately, it's not really documented.... 
But the gist of it is that it uses `koji` via the `add_filesystem_plugin` to generate the initial filesystem tarball, which is then used as the starting point for building the image.
Let me know if this is of interest, I can provide more info.
Would this solve your use-case?

In most cases, the plugins are not aware of whether or not they should run. 
This is usually handled by manually configuring which plugins to run - simply omit the ones that are irrelevant.
Another way is to use [osbs-client](https://github.com/projectatomic/osbs-client) to configure the plugins.

nipakoo:
Thanks for the tip! Will definitely look into this plugin more closely in near future.

atomic_reactor/plugins/post_compress.py
```diff
@@ -47,7 +47,11 @@ def __init__(self, tasker, workflow, load_exported_image=False, method='gzip'):
                                     you may load the exported tar with this switch
         """
         super(CompressPlugin, self).__init__(tasker, workflow)
-        self.load_exported_image = load_exported_image
+        if base_image_is_scratch(self.workflow.builder.base_image.to_str()):
+            self.log.debug("build based on scratch image, do not load exported image")
+            self.load_exported_image = False
+        else:
+            self.load_exported_image = load_exported_image
```
nipakoo: Removed
- Do not squash images that have scratch base
- Do not load exported image when basing on scratch
from the first commit and cleaned history.

Now we utilize git history to find the bottom layer of built image and squash from that. Is this viable?

atomic_reactor/plugins/prepub_squash.py
```diff
@@ -70,13 +72,23 @@ def __init__(self, tasker, workflow, tag=None, from_base=True, from_layer=None,
...
-            except KeyError:
-                self.log.error("Missing Id in inspection: '%s'", self.workflow.base_image_inspect)
-                raise
-            self.log.info("will squash from base-image: '%s'", base_image_id)
-            self.from_layer = base_image_id
+            if base_image_is_scratch(self.workflow.builder.base_image.to_str()):
+                self.log.debug("build based on scratch, will not include base"
+                               "image in squash")
+                try:
+                    self.from_layer = find_bottom_layer_id(tasker, self.image)
```
twaugh: I think you can just specify `from_layer=None` to squash all the layers: https://github.com/goldmann/docker-squash/blob/05cd17fe639183e2986fa89b3ddb62bcf903cf2b/README.rst

nipakoo: Ah seems to do the job :) how about this?

lcarva:
Can you provide an example Dockerfile to clarify your use case?
We're trying to understand why some of the plugins are not needed (rpm_qa, inject_yum_repo, for example).

nipakoo:
1.
Unfortunately so far I have only tested this with very minimal Dockerfile in similar fashion with [fedora-cloud](https://github.com/fedora-cloud/docker-brew-fedora/blob/a7ae28fbed5ebf675d6a352f74d86149228e47f1/Dockerfile)

Basically:
FROM scratch
MAINTAINER [Niko Kortström <niko.kortstrom@nokia.com>]
LABEL Name="foo" \
       BZComponent="foo" \
       Version="0.2.0" \
       Release="1"
ADD foo-docker-base-0.2.0-2.x86_64.tar.xz /

2.
I believe the a limited amount of plugins were used in the atomic-reactor setup that I used to test the functionality so pulled off the minimal build with not all plugins checked if they would need adjustments to handle the notice the special case.

atomic_reactor/plugins/post_rpmqa.py
```diff
@@ -43,6 +44,9 @@ def __init__(self, tasker, workflow, image_id, ignore_autogenerated_gpg_keys=Tru
         self.ignore_autogenerated_gpg_keys = ignore_autogenerated_gpg_keys
 
     def run(self):
+        if base_image_is_scratch(self.workflow.builder.base_image.to_str()):
+            self.log.debug("image has scratch base, do not perform RPM check")
+            return
```
twaugh: We'd still want to run this though wouldn't we? I'm not sure why we would skip it.

atomic_reactor/plugins/pre_inject_yum_repo.py
```diff
@@ -129,6 +129,9 @@ def run(self):
         """
         run the plugin
         """
+        if base_image_is_scratch(self.workflow.builder.base_image.to_str()):
+            self.log.debug("base image is scratch, do not inject yum repo")
+            return
```
twaugh: Not sure why we would skip this.

atomic_reactor/plugins/post_rpmqa.py
nipakoo: Yea you're right. Cannot check this atm but I guess we would want to implement a non-fatal exit and warning log when rpm itself is not installed.

atomic_reactor/plugins/pre_inject_yum_repo.py
nipakoo: Isn't there yum commands run in this plugin? Which is not possible on a blank base image.

nipakoo:
Sure I will take a look at the tests also when I find a bit time :)

atomic_reactor/plugins/post_rpmqa.py
twaugh: Right now atomic-reactor assumes all images will have /bin/rpm installed. Unless/until that's changed, I'd rather the rpmqa plugin continues to fail when that's not the case.

atomic_reactor/plugins/pre_inject_yum_repo.py
twaugh: At the very end it runs /bin/rm -- but since the purpose is to build a usable base image, it seems likely this would be installed by the time the Dockerfile gets to the end.

bfontecc007:
Thank You for the contribution, @nipakoo

lcarva:
Closing this in favor of https://github.com/projectatomic/atomic-reactor/pull/924

PR #700, opened 2017-05-25
log load info, on worker clusters

rcerven:

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -172,6 +172,8 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
         osbs = OSBS(conf, conf)
         current_builds = self.get_current_builds(osbs)
         load = current_builds / cluster.max_concurrent_builds
+        self.log.debug('enabled cluster %s for platform %s has load %s and active builds %s/%s',
+                       cluster.name, platform, load, current_builds, cluster.max_concurrent_builds)
         return ClusterInfo(cluster, platform, osbs, load)
 
     def choose_cluster(self, platform):


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -186,6 +189,8 @@ def choose_cluster(self, platform):
         selected = min(clusters, key=lambda c: c.load)
         self.log.info('platform %s will use cluster %s',
```
vrutkovs: Remove this line?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -186,6 +189,8 @@ def choose_cluster(self, platform):
...
                       platform, selected.cluster.name)
+        self.log.debug('platform %s will use cluster %s with the lowest load %s',
+                       platform, selected.cluster.name, selected)
```
vrutkovs: `selected` here is a cluster name, it should be `selected.load`

atomic_reactor/plugins/build_orchestrate_build.py
vrutkovs: Lets use one message for picked cluster and its load

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -172,6 +172,9 @@ def get_cluster_info(self, cluster, platform):
         osbs = OSBS(conf, conf)
         current_builds = self.get_current_builds(osbs)
         load = current_builds / cluster.max_concurrent_builds
+        self.log.debug('enabled cluster %s for platform %s has load %s and active builds %s and '
+                       'max concurrent builds %s', cluster.name, platform, load, current_builds,
```
twaugh: How about `active builds %s/%s`?

atomic_reactor/plugins/build_orchestrate_build.py
twaugh: I don't really think we need this debug line if we have the get_cluster_info() one.

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: ok

PR #701, opened 2017-05-26
provide hash algo for fetch maven artifacts by URL

lcarva:
Plugin fetch-maven-artifacts has been enhanced to allow
user to specify the hashing algorithm to be used when
validating an artifact downloaded via URL.
Possible options are md5, sha1 and sha256.

In addition, when fetching archives from Koji, plugin
no longer assumes the hashing algorithm is md5. Instead,
it leverages the checksum_type attribute from the artifact
information from Koji.

### DIFF
--- a/atomic_reactor/plugins/pre_fetch_maven_artifacts.py
+++ b/atomic_reactor/plugins/pre_fetch_maven_artifacts.py
@@ -9,6 +9,7 @@ from __future__ import unicode_literals
 
 import fnmatch
 import hashlib
+import koji
 import os
 import requests
 
@@ -16,8 +17,6 @@ from atomic_reactor import util
 from atomic_reactor.koji_util import create_koji_session
 from atomic_reactor.plugin import PreBuildPlugin
 from collections import namedtuple
-from koji import PathInfo
-
 
 try:
     from urlparse import urlparse
@@ -61,7 +60,7 @@ class NvrRequest(object):
         return [archive for archive in self.archives if not archive['matched']]
 
 
-DownloadRequest = namedtuple('DownloadRequest', 'url dest md5')
+DownloadRequest = namedtuple('DownloadRequest', 'url dest checksums')
 
 
 class FetchMavenArtifactsPlugin(PreBuildPlugin):
@@ -104,7 +103,7 @@ class FetchMavenArtifactsPlugin(PreBuildPlugin):
             'root': koji_root,
             'auth': koji_auth or None
         }
-        self.path_info = PathInfo(topdir=self.koji_info['root'])
+        self.path_info = koji.PathInfo(topdir=self.koji_info['root'])
         self.allowed_domains = set(domain.lower() for domain in allowed_domains or [])
         self.workdir = self.workflow.source.get_dockerfile_path()[1]
         self.session = None
@@ -147,8 +146,9 @@ class FetchMavenArtifactsPlugin(PreBuildPlugin):
                 # not contain a trailing slash, which causes the last dir to
                 # be dropped.
                 url = maven_build_path + '/' + maven_file_path
-                download_queue.append(DownloadRequest(url, maven_file_path,
-                                                      build_archive['checksum']))
+                checksum_type = koji.CHECKSUM_TYPES[build_archive['checksum_type']]
+                checksums = {checksum_type: build_archive['checksum']}
+                download_queue.append(DownloadRequest(url, maven_file_path, checksums))
 
             unmatched_archive_requests = nvr_request.unmatched()
             if unmatched_archive_requests:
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/pre_fetch_maven_artifacts.py
```diff
@@ -147,8 +153,10 @@ def process_by_nvr(self, nvr_requests):
                 # not contain a trailing slash, which causes the last dir to
                 # be dropped.
                 url = maven_build_path + '/' + maven_file_path
+                checksum_type = koji.CHECKSUM_TYPES[build_archive['checksum_type']]
```
vrutkovs: Why are we using `koji.CHECKSUM_TYPES` here - and then just `CHECKSUM_TYPES` below?

lcarva: `koji.CHECKSUM_TYPES` is used to convert the value we get from koji into something meaningful. Koji reports the checksum_type as 0, we use `koji.CHECKSUM_TYPES` to get the string `'md5'` from it.

The `CHECKSUM_TYPES` dict defined in this plugin has a different purposed. It maps names of hashing algorithms  to the actual hashing function: `'md5' => haslib.md5`

vrutkovs: Oh, now I get it. Maybe a different name should be used for clarity - `CHECKSUM_DICT` for instance?

tests/plugins/test_fetch_maven_artifacts.py
```diff
@@ -590,6 +642,32 @@ def test_fetch_maven_artifacts_url_bad_checksum(tmpdir, docker_tasker):
...
+        }]
+    )
+
+    with pytest.raises(PluginFailedException) as e:
+        runner.run()
+
+    assert 'Missing checksum value attribute' in str(e)
+
+
+@responses.activate
```
vrutkovs: Stickler doesn't know how py.test fixtures work :(

This line should have `  # noqa` and a comment above that `docker_tasker` is a fixture

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
twaugh: Could `hashlib.algorithms` help?

lcarva: Yeah good point. Changed it to `CHECKSUM_ALGORITHMS`.

tests/plugins/test_fetch_maven_artifacts.py
```diff
@@ -590,7 +642,33 @@ def test_fetch_maven_artifacts_url_bad_checksum(tmpdir, docker_tasker):
...
```
lcarva: @vrutkovs how do I get rid of this?

vrutkovs: Weird, it should have gone away with `# noqa`. I think its safe to ignore it for now

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
lcarva: > Could hashlib.algorithms help?

Wasn't aware of that! Not sure if it would help though. I'm trying to avoid using `getattr(hashlib, checksum_type)`. Having an explicit map of which algorithms are supported seems safer to me.

twaugh: Instead of `getattr(hashlib, checksum_type)`, `hashlib.new(checksum_type)`? But otherwise I'm fine with this as it is if you are.

lcarva: That I like :)
Updated the code accordingly.

lcarva:
Marked as work in progress as we iron out the yaml schema.

lcarva:
@vrutkovs @twaugh PTAL. 
Removed the need to explicitly set hashing algorithm via `checksum_type`.
Instead, user simply needs to define one or more of the following attributes: md5, sha1, sha256.

If the user provides more than one hashing algorithm, then multiple hashes will be computed - one per algorithm. All hashes will be validated.

In most cases, users will only provide a single hash to be validated. 
However, when the output of this plugin is properly integrated with Koji's content generator metadata,
we'll always need to compute md5 hash for Koji data consistency.
This change provides the flexibility to users to choose desired hashing algorithm to be
used by OSBS, while also paving the way in ensuring data consistency when importing data into Koji.

Before merging I'd like to squash commits and reword commit message. 
Leaving it as is during review for easier rollback if needed.

tests/plugins/test_fetch_maven_artifacts.py
```diff
@@ -65,6 +63,7 @@
     'artifact_id': 'jaxb-core',
     'build_id': 472397,
     'checksum': '697317209103338c7c841e327bb6e7b0',
+    'checksum_type': 0,
```
twaugh: Should we use `koji.CHECKSUM_TYPES['md5']` here? (and add `CHECKSUM_TYPES` to our `koji/__init__.py` stub)

lcarva: Yes. Enhanced the stub to handle this.

lcarva:
Squashed commits, ready for merging.

PR #702, opened 2017-06-13
orchestrate_build: enhance tests

twaugh:
Make sure that Unicode logs will be handled correctly.

### DIFF
--- a/tests/plugins/test_orchestrate_build.py
+++ b/tests/plugins/test_orchestrate_build.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 """
 Copyright (c) 2017 Red Hat, Inc
 All rights reserved.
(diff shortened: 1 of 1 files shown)


tests/plugins/test_orchestrate_build.py
```diff
@@ -136,7 +137,7 @@ def mock_create_worker_build(**kwargs):
 
     (flexmock(OSBS)
         .should_receive('get_build_logs')
-        .and_yield('line-%d' % line for line in range(10)))
+        .and_yield('Lińe-%d' % line for line in range(10)))
```
lcarva: I don't think this would trigger the scenario we want.
Maybe we need to mock at the API level? But that seems something this plugin shouldn't be concerned with.

lcarva:
@twaugh, is this still relevant?

twaugh:
No.

PR #703, opened 2017-06-14
koji_promote: don't encode docker logs and final openshift logs if logs are bytestream

vrutkovs:

### DIFF
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -345,7 +345,11 @@ class KojiPromotePlugin(ExitPlugin):
             logfile = NamedTemporaryFile(prefix=self.build_id,
                                          suffix=".log",
                                          mode='wb')
-            logfile.write(logs.encode('UTF-8'))
+            try:
+                logfile.write(logs)
+            except (TypeError, UnicodeEncodeError):
+                # Older osbs-client versions returned Unicode objects
+                logfile.write(logs.encode('utf-8'))
             logfile.flush()
             metadata = self.get_output_metadata(logfile.name,
                                                 "openshift-final.log")
@@ -354,7 +358,7 @@ class KojiPromotePlugin(ExitPlugin):
         docker_logs = NamedTemporaryFile(prefix="docker-%s" % self.build_id,
                                          suffix=".log",
                                          mode='wb')
-        docker_logs.write("\n".join(self.workflow.build_result.logs).encode('UTF-8'))
+        docker_logs.write("\n".join(self.workflow.build_result.logs).encode('utf-8'))
         docker_logs.flush()
         output.append(Output(file=docker_logs,
                              metadata=self.get_output_metadata(docker_logs.name,
--- a/tests/plugins/test_koji_promote.py
+++ b/tests/plugins/test_koji_promote.py
@@ -179,7 +179,8 @@ def mock_environment(tmpdir, session=None, name=None,
                      is_rebuild=True, docker_registry=True,
                      pulp_registries=0, blocksize=None,
                      task_states=None, additional_tags=None,
-                     has_config=None):
+                     has_config=None,
+                     logs_return_bytes=True):
     if session is None:
         session = MockedClientSession('', task_states=None)
     if source is None:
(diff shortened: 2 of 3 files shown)


rel-eng/packages/atomic-reactor
```diff
@@ -1 +1 @@
-1.6.23.1-1 ./
+1.6.23-1 ./
```
rcerven: latest release version is 1.6.23.2

so not sure if there should be future version or current version

vrutkovs: Right, that doesn't belong here. Removed

twaugh:
Don't forget https://github.com/projectatomic/atomic-reactor/blob/master/tests/plugins/test_orchestrate_build.py#L139.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -354,7 +354,7 @@ def get_logs(self):
         docker_logs = NamedTemporaryFile(prefix="docker-%s" % self.build_id,
                                          suffix=".log",
                                          mode='wb')
-        docker_logs.write("\n".join(self.workflow.build_result.logs).encode('UTF-8'))
+        docker_logs.write(b"\n".join(self.workflow.build_result.logs))
```
twaugh: This value comes from atomic-reactor itself and wouldn't have changed type I think.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -345,7 +345,7 @@ def get_logs(self):
             logfile = NamedTemporaryFile(prefix=self.build_id,
                                          suffix=".log",
                                          mode='wb')
-            logfile.write(logs.encode('UTF-8'))
+            logfile.write(logs)
```
twaugh: I believe this should be compatible with old and new versions of osbs-client.

It's just that we need this new version of atomic-reactor when using the new osbs-client in the builder image they are both in.

vrutkovs: Hmm, right. Not sure how to do that correctly - use `six` find out the string type?

tests/plugins/test_koji_promote.py
```diff
@@ -261,10 +261,10 @@ def mock_environment(tmpdir, session=None, name=None,
         setattr(workflow, 'exported_image_sequence', [{'path': fp.name}])
 
     if build_process_failed:
-        workflow.build_result = BuildResult(logs=["docker build log - \u2018 \u2017 \u2019 \n'"],
+        workflow.build_result = BuildResult(logs=[b"docker build log - \u2018 \u2017 \u2019 \n'"],
```
twaugh: I think this should not change.

tests/plugins/test_koji_promote.py
```diff
@@ -261,10 +261,10 @@ def mock_environment(tmpdir, session=None, name=None,
...
                                             fail_reason="not built")
     else:
-        workflow.build_result = BuildResult(logs=["docker build log - \u2018 \u2017 \u2019 \n'"],
+        workflow.build_result = BuildResult(logs=[b"docker build log - \u2018 \u2017 \u2019 \n'"],
```
twaugh: I think this should not change.

atomic_reactor/plugins/exit_koji_promote.py
twaugh: I *think* it might be OK as it is.

tests/plugins/test_koji_promote.py
vrutkovs: Ah, correct, these are logs straight from container

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -345,7 +345,11 @@ def get_logs(self):
             logfile = NamedTemporaryFile(prefix=self.build_id,
                                          suffix=".log",
                                          mode='wb')
-            logfile.write(logs.encode('UTF-8'))
+            try:
+                logfile.write(logs)
+            except UnicodeEncodeError:
```
twaugh: No, `TypeError`.

twaugh: Or do different versions of Python behave differently with this? I hope not.

vrutkovs: It doesn't throw `TypeError` :( Not sure why though, unit test looks correct to me

twaugh: Python 3 throws TypeError:
```
>>> open('/tmp/foo', 'wb').write('ä')
TypeError: a bytes-like object is required, not 'str'
```

So maybe catch both.

vrutkovs: Oh, I see. So it throws `TypeError` in py3, as expected.

The problem that it currently throws `UnicodeEncodeError` in py2 in this case

twaugh: We can just `except (TypeError, UnicodeEncodeError):` our way around it I expect.

vrutkovs: Right, that worked

vrutkovs:
Integration tests seem to pass with osbs-clients PR and this PR.

Not sure how to verify py3 case though - I'll try to revive fedora-based osbs-box

lcarva:
> Not sure how to verify py3 case though

This plugin requires koji library which is only python 2, safe to ignore it?

twaugh:
I'd say so, yes.

vrutkovs:
>This plugin requires koji library which is only python 2, safe to ignore it?

Ah, yes. 

Koji module would be py3-compatible shortly though, so more fixes might be required

PR #705, opened 2017-06-19
[WIP] plugins/add_filesystem: split worker and orchestrator

jarodwilson:
In a case where we're building images for multiple architectures, we need
to be able to pass in a list of the relevant architectures we want to
build for. In order to properly create these multi-arch builds, we need to
split the add_filesystem plugin's code up a bit, into portions that run on
the orchestrator, which organizes which workers should be utilized for
different architectures specified to build arch-specific images, and the
portion that runs on those workers.

This adds new parameters to add_filesystem, arch and architectures, which
are consumed by the worker and orchestrator builds respectively, and
overwrite image-build.conf's arches var (which we do log).

Outside of that, the majority of the work is merely reorganizing code and
conditionalizing what gets run depending on being invoked as an
orchestrator or as a worker, and then tweaking self-tests to compensate
for certain paths requiring more information provided to complete
successfully.

Theoretically, this should also be backwards-compatible with an older
osbs-client that doesn't know about the orchestrator/worker split.

While we're at it, fix the link for koji's image_build docs.

- Adjust default values of arches we pass through for worker builds
- Support legacy osbs-client one-shot building
- Further refactor code to eliminate duplicate get_filesystem_regex
  and parse_image_build_config calls

### DIFF
--- a/tests/plugins/test_add_filesystem.py
+++ b/tests/plugins/test_add_filesystem.py
@@ -101,7 +101,7 @@ def mock_koji_session(koji_proxyuser=None, koji_ssl_certs_dir=None,
             .replace_with(get_task_result_mock).once())
 
     session.should_receive('listTaskOutput').and_return([
-        'fedora-23-1.0.tar.gz',
+        'fedora-23-1.0.x86_64.tar.gz',
     ])
     session.should_receive('getTaskChildren').and_return([
         {'id': 1234568},
@@ -143,7 +143,6 @@ def mock_image_build_file(tmpdir, contents=None):
             version = 1.0
             target = guest-fedora-23-docker
             install_tree = http://install-tree.com/$arch/fedora23/
-            arches = x86_64
 
             format = docker
             distro = Fedora-23
@@ -205,6 +204,7 @@ def test_add_filesystem_plugin_generated(tmpdir, docker_tasker, scratch):
         RUN dnf install -y python-django
         """)
     workflow = mock_workflow(tmpdir, dockerfile)
+    task_id = FILESYSTEM_TASK_ID
     mock_koji_session(scratch=scratch)
     mock_image_build_file(str(tmpdir))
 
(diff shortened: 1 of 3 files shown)


tests/plugins/test_add_filesystem.py
```diff
@@ -213,7 +213,42 @@ def test_add_filesystem_plugin_generated(tmpdir, docker_tasker, scratch):
...
+                'koji_hub': KOJI_HUB,
+                'from_task_id': task_id,
+                'architecture': 'x86_64'
+            }
+        }]
+    )
+
+    results = runner.run()
+    plugin_result = results[AddFilesystemPlugin.key]
+    assert 'base-image-id' in plugin_result
```
lcarva: Good idea asserting the plugin result!
Think we're missing checking the value for `'filesystem-koji-task-id'`.
You can simplify these assertion statements to something like this:
```python
expected_plugin_result = {
    'base-image-id': IMPORTED_IMAGE_ID,
    'filesystem-koji-task-id': FILESYSTEM_TASK_ID,
}
assert plugin_result == expected_plugin_result
```

tests/plugins/test_add_filesystem.py
```diff
@@ -407,12 +445,11 @@ def test_image_build_overwrites(tmpdir):
                     [fedora-os]
                     baseurl = http://default-repo.com/fedora/$basearch/os.repo
                     """))
-    plugin = create_plugin_instance(tmpdir, {'repos': repos})
+    plugin = create_plugin_instance(tmpdir, {'repos': repos, 'architectures': architectures})
     image_build_conf = dedent("""\
         [image-build]
         name = my-name
         version = 1.0
-        arches = i386,i486
```
lcarva: You could enhance this test to:

- Ensure arches when defined here are overwritten by `architectures` plugin parameter
- Ensure arches when defined here are overwritten by `architecture` plugin parameter
- Ensure that if neither `architecture` nor `architectures` is provided, this arches value is used

tests/plugins/test_add_filesystem.py
```diff
@@ -407,12 +445,11 @@ def test_image_build_overwrites(tmpdir):
...
         target = guest-fedora-23-docker
         install_tree = http://install-tree.com/$arch/fedora23/
         format = locker,mocker
```
lcarva: I can't comment further down than this, but `test_build_filesystem_from_task_id` can be enhanced by adding additional cases to ensure `self.architecture` is used in regex when set.

tests/plugins/test_add_filesystem.py
jarodwilson: Done.

tests/plugins/test_add_filesystem.py
jarodwilson: I'll see what I can do to enhance this self-test as well.

tests/plugins/test_add_filesystem.py
```diff
@@ -101,7 +101,7 @@ def _mockBuildImageOz(*args, **kwargs):
             .replace_with(get_task_result_mock).once())
 
     session.should_receive('listTaskOutput').and_return([
-        'fedora-23-1.0.tar.gz',
+        'fedora-23-1.0.x86_64.tar.gz',
```
twaugh: Could add some other architectures in here and check it picks the right one.

tests/plugins/test_add_filesystem.py
jarodwilson: Just pushed a commit that should implement this.

tests/plugins/test_add_filesystem.py
jarodwilson: So I didn't add additional cases, per se, but rather, took the existing ones, and inserted architecture into some of them. Seems to work!

tests/plugins/test_add_filesystem.py
jarodwilson: This particular bit of code seems rather tied up with several of the self-tests, but the additions pushed for test_build_filesystem_from_task_id go a ways to make sure we match arch properly too, when it's provided. I think we're good to go here, and will start working on cleaning things up for a formal pull request -- right after some functional testing with @lcarva.

lcarva:
@jarodwilson, one of the things we found out during integration tests is that `add_filesystem` plugin in worker build never receives the `from_task_id` parameter. I think the best way to address this is to modify the `orchestrate_build` plugin to read the plugin result of `add_filesystem` plugin, so it can pass it along when creating the worker build.

The sequence of events becomes this:

1. Orchestrator build is created
2. Orchestrator build runs `add_filesystem` plugin to execute the image task in koji. When done this plugin completes, it returns a dict with the key `filesystem-koji-task-id`.
3. Orchestrator build runs `orchestrate_build` plugin which reads the result from `add_filesystem` plugin and retrieves the `filesystem-koji-task-id` value. Then, the worker builds are [created](https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/build_orchestrate_build.py#L197) with this value for the `filesystem_koji_task_id` build param.
4. In worker build the `add_filesystem` plugin runs again, this time it streams the tarball file system from the given koji task ID.

Example of reading result from another plugin can be seen [here](https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/exit_store_metadata_in_osv3.py#L47).

PR #706, opened 2017-06-20
store_metadata_in_osv3: Convert KojiPromotePlugin.key to constant

dzickusrh:
The plugin exit_store_metadata_in_osv3 was directly importing the
KojiPromotePlugin plugin even though it could be packaged separately.

Now that atomic_reactor/constants.py has been created to store the
key values, it is easier to use that file instead of a plugin that may
not be around.

Convert 'KojiPromotePlugin.key' -> PLUGIN_KOJI_PROMOTE_PLUGIN_KEY

and convert the imports:

'from atomic_reactor.plugins.exit_koji_promote import KojiPromotePlugin'
to
'from atomic_reactor.constants import PLUGIN_KOJI_PROMOTE_PLUGIN_KEY'

This is done for the store_metadata_in_osv3 plugin and its unit
test case, test_store_metadata.

For the test_store_metadata, I ripped out the tricky 'faulting import
and append a test path to plugin' logic.

### DIFF
--- a/tests/plugins/test_store_metadata.py
+++ b/tests/plugins/test_store_metadata.py
@@ -15,26 +15,11 @@ from datetime import datetime, timedelta
 from copy import deepcopy
 from textwrap import dedent
 
-try:
-    import koji as koji
-except ImportError:
-    import inspect
-    import os
-    import sys
-
-    # Find our mocked koji module
-    import tests.koji as koji
-    mock_koji_path = os.path.dirname(inspect.getfile(koji.ClientSession))
-    if mock_koji_path not in sys.path:
-        sys.path.append(os.path.dirname(mock_koji_path))
-finally:
-    del koji
-    from atomic_reactor.plugins.exit_koji_promote import KojiPromotePlugin
-
 from flexmock import flexmock
 from osbs.api import OSBS
 import osbs.conf
 from osbs.exceptions import OsbsResponseException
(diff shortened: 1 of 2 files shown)


tests/plugins/test_store_metadata.py
```diff
@@ -15,26 +15,11 @@
...
-    import sys
-
-    # Find our mocked koji module
-    import tests.koji as koji
-    mock_koji_path = os.path.dirname(inspect.getfile(koji.ClientSession))
-    if mock_koji_path not in sys.path:
-        sys.path.append(os.path.dirname(mock_koji_path))
-finally:
-    del koji
-    from atomic_reactor.plugins.exit_koji_promote import KojiPromotePlugin
```
twaugh: Actually I think we need this part. The reason it's there is that the `koji` module is not yet available for Python 3, but we still want to test our own koji-using code under Python 3.

To this end, we have a pretend koji module: https://github.com/projectatomic/atomic-reactor/blob/master/tests/koji/__init__.py

The `ImportError` stuff above is to handle this case, and allow our pretend koji module to be used when testing a module that uses `import koji`.

Yes, this should probably have had a big comment explaining that.

twaugh: Aha... of course, we no longer need it in *this* module, because it doesn't need koji itself nor does the module under test.

PR #707, opened 2017-06-20
Koji upload

dzickusrh:
Depends on PR#706.  3 commits to implement koji_upload plugin

### DIFF
--- a/atomic_reactor/plugins/post_koji_upload.py
+++ b/atomic_reactor/plugins/post_koji_upload.py
@@ -1,0 +1,731 @@
+"""
+Copyright (c) 2015 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import unicode_literals
+
+from collections import namedtuple
+import json
+import os
+import random
+from string import ascii_letters
+import subprocess
+from tempfile import NamedTemporaryFile
+import time
+import copy
+
+from atomic_reactor import __version__ as atomic_reactor_version
+from atomic_reactor import start_time as atomic_reactor_start_time
+from atomic_reactor.plugin import ExitPlugin
+from atomic_reactor.source import GitSource
+from atomic_reactor.plugins.post_rpmqa import PostBuildRPMqaPlugin
+from atomic_reactor.plugins.pre_add_filesystem import AddFilesystemPlugin
+from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
+from atomic_reactor.plugins.pre_add_help import AddHelpPlugin
+from atomic_reactor.constants import (PROG, PLUGIN_KOJI_PROMOTE_PLUGIN_KEY,
+                                      PLUGIN_KOJI_TAG_BUILD_KEY)
+from atomic_reactor.util import (get_version_of_tools, get_checksums,
+                                 get_build_json, get_preferred_label,
+                                 get_docker_architecture, df_parser,
+                                 are_plugins_in_order)
+from atomic_reactor.koji_util import create_koji_session, tag_koji_build
+from osbs.conf import Configuration
+from osbs.api import OSBS
+from osbs.exceptions import OsbsException
+
+# An output file and its metadata
+Output = namedtuple('Output', ['file', 'metadata'])
+
+
+class KojiUploadLogger(object):
+    def __init__(self, logger, notable_percent=10):
+        self.logger = logger
+        self.notable_percent = notable_percent
+        self.last_percent_done = 0
+
+    def callback(self, offset, totalsize, size, t1, t2): # pylint: disable=W0613
+        if offset == 0:
+            self.logger.debug("upload size: %.1fMiB", totalsize / 1024.0 / 1024)
+
+        if not totalsize or not t1:
+            return
+
+        percent_done = 100 * offset / totalsize
+        if (percent_done >= 99 or
+                percent_done - self.last_percent_done >= self.notable_percent):
+            self.last_percent_done = percent_done
+            self.logger.debug("upload: %d%% done (%.1f MiB/sec)",
+                              percent_done, size / t1 / 1024 / 1024)
+
+
+class KojiPromotePlugin(ExitPlugin):
+    """
+    Promote this build to Koji
+
+    Submits a successful build to Koji using the Content Generator API,
+    https://fedoraproject.org/wiki/Koji/ContentGenerators
+
+    Authentication is with Kerberos unless the koji_ssl_certs
+    configuration parameter is given, in which case it should be a
+    path at which 'cert', 'ca', and 'serverca' are the certificates
+    for SSL authentication.
+
+    If Kerberos is used for authentication, the default principal will
+    be used (from the kernel keyring) unless both koji_keytab and
+    koji_principal are specified. The koji_keytab parameter is a
+    keytab name like 'type:name', and so can be used to specify a key
+    in a Kubernetes secret by specifying 'FILE:/path/to/key'.
+
+    If metadata_only is set, the 'docker save' image will not be
+    uploaded, only the logs. The import will be marked as
+    metadata-only.
+
+    Runs as an exit plugin in order to capture logs from all other
+    plugins.
+    """
+
+    key = PLUGIN_KOJI_PROMOTE_PLUGIN_KEY
+    is_allowed_to_fail = False
+
+    def __init__(self, tasker, workflow, kojihub, url,
+                 verify_ssl=True, use_auth=True,
+                 koji_ssl_certs=None, koji_proxy_user=None,
+                 koji_principal=None, koji_keytab=None,
+                 metadata_only=False, blocksize=None,
+                 target=None, poll_interval=5):
+        """
+        constructor
+
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param kojihub: string, koji hub (xmlrpc)
+        :param url: string, URL for OSv3 instance
+        :param verify_ssl: bool, verify OSv3 SSL certificate?
+        :param use_auth: bool, initiate authentication with OSv3?
+        :param koji_ssl_certs: str, path to 'cert', 'ca', 'serverca'
+        :param koji_proxy_user: str, user to log in as (requires hub config)
+        :param koji_principal: str, Kerberos principal (must specify keytab)
+        :param koji_keytab: str, keytab name (must specify principal)
+        :param metadata_only: bool, whether to omit the 'docker save' image
+        :param blocksize: int, blocksize to use for uploading files
+        :param target: str, koji target
+        :param poll_interval: int, seconds between Koji task status requests
+        """
+        super(KojiPromotePlugin, self).__init__(tasker, workflow)
+
+        self.kojihub = kojihub
+        self.koji_ssl_certs = koji_ssl_certs
+        self.koji_proxy_user = koji_proxy_user
+
+        self.koji_principal = koji_principal
+        self.koji_keytab = koji_keytab
+
+        self.metadata_only = metadata_only
+        self.blocksize = blocksize
+        self.target = target
+        self.poll_interval = poll_interval
+
+        self.namespace = get_build_json().get('metadata', {}).get('namespace', None)
+        osbs_conf = Configuration(conf_file=None, openshift_uri=url,
+                                  use_auth=use_auth, verify_ssl=verify_ssl,
+                                  namespace=self.namespace)
+        self.osbs = OSBS(osbs_conf, osbs_conf)
+        self.build_id = None
+        self.pullspec_image = None
+
+    @staticmethod
+    def parse_rpm_output(output, tags, separator=';'):
+        """
+        Parse output of the rpm query.
+
+        :param output: list, decoded output (str) from the rpm subprocess
+        :param tags: list, str fields used for query output
+        :return: list, dicts describing each rpm package
+        """
+
+        def field(tag):
+            """
+            Get a field value by name
+            """
+            try:
+                value = fields[tags.index(tag)]
+            except ValueError:
+                return None
+
+            if value == '(none)':
+                return None
+
+            return value
+
+        components = []
+        sigmarker = 'Key ID '
+        for rpm in output:
+            fields = rpm.rstrip('\n').split(separator)
+            if len(fields) < len(tags):
+                continue
+
+            signature = field('SIGPGP:pgpsig') or field('SIGGPG:pgpsig')
+            if signature:
+                parts = signature.split(sigmarker, 1)
+                if len(parts) > 1:
+                    signature = parts[1]
+
+            component_rpm = {
+                'type': 'rpm',
+                'name': field('NAME'),
+                'version': field('VERSION'),
+                'release': field('RELEASE'),
+                'arch': field('ARCH'),
+                'sigmd5': field('SIGMD5'),
+                'signature': signature,
+            }
+
+            # Special handling for epoch as it must be an integer or None
+            epoch = field('EPOCH')
+            if epoch is not None:
+                epoch = int(epoch)
+
+            component_rpm['epoch'] = epoch
+
+            if component_rpm['name'] != 'gpg-pubkey':
+                components.append(component_rpm)
+
+        return components
+
+    def get_rpms(self):
+        """
+        Build a list of installed RPMs in the format required for the
+        metadata.
+        """
+
+        tags = [
+            'NAME',
+            'VERSION',
+            'RELEASE',
+            'ARCH',
+            'EPOCH',
+            'SIGMD5',
+            'SIGPGP:pgpsig',
+            'SIGGPG:pgpsig',
+        ]
+
+        sep = ';'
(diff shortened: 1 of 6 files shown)


tests/plugins/test_store_metadata.py
```diff
@@ -233,15 +252,20 @@ def test_metadata_plugin_rpmqa_failure(tmpdir):
     workflow.prebuild_results = {}
     workflow.postbuild_results = {
         PostBuildRPMqaPlugin.key: RuntimeError(),
+        PLUGIN_KOJI_UPLOAD_PLUGIN_KEY: {'metadata_fragment_key': 'metadata.json',
+                                        'metadata_fragment': 'configmap/build-1-md'}
```
twaugh: Just an observation: both of these are post-build plugins and the rpm-qa one runs before koji-upload. If it fails, koji-upload will not be run at all.

That means this is testing a hypothetical situation in which they are deliberately run the other way around, to test whether the store-metadata plugin can deal with it.

Seems valid.

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -62,9 +59,9 @@ def callback(self, offset, totalsize, size, t1, t2): # pylint: disable=W0613
...
 
 
-class KojiPromotePlugin(ExitPlugin):
+class KojiUploadPlugin(PostBuildPlugin):
     """
-    Promote this build to Koji
+    Upload this build to Koji
 
     Submits a successful build to Koji using the Content Generator API,
     https://fedoraproject.org/wiki/Koji/ContentGenerators
```
twaugh: We should remove 'Submits a successful build to Koji using the Content Generator API', because that's what koji-import will do. How about this?:
```
Note: only the image archive is uploaded to Koji 
at this stage. Metadata about this image
is created and stored in a ConfigMap in 
OpenShift, ready for the orchestrator build to
collect and use to actually create the Koji Build 
together with the uploaded image archive(s).
```

We also no longer need the `metadata_only` parameter. It can be removed entirely from koji_upload.

tests/plugins/test_koji_upload.py
```diff
@@ -938,200 +967,14 @@ def test_koji_promote_image_tags(self, tmpdir, os_env, additional_tags):
...
-        (True,
-         True),
-        (False,
-         False),
-    ])
-    @pytest.mark.parametrize('tag_later', (True, False))
-    def test_koji_promote_success(self, tmpdir, apis, docker_registry,
-                                  pulp_registries, metadata_only, blocksize,
-                                  target, os_env, has_config, is_autorebuild,
-                                  tag_later):
```
twaugh: We still want this test. It should just do a bit less. It should:

- set up the mock environment and run the plugin of course
- verify the 'metadata_version', 'buildroots', and 'output' keys as before
- not verify 'build', as this is something that koji-import will create

tests/plugins/test_koji_upload.py
```diff
@@ -1179,36 +1023,15 @@ def test_koji_promote_pullspec(self, tmpdir, os_env, primary, unique, invalid):
...
-
     @pytest.mark.parametrize('expect_result', [
         'empty_config',
         'no_help_file',
         'skip',
         'pass',
         'unknown_status'
     ])
-    def test_koji_promote_add_help(self, tmpdir, os_env, expect_result):
+    def test_koji_upload_add_help(self, tmpdir, os_env, expect_result):
```
twaugh: This test is concerned with the 'build' key which koji_upload will not create, so it can be removed.

twaugh:
Now that https://github.com/projectatomic/atomic-reactor/pull/706 is merged you could rebase onto master to avoid conflicts.

tests/plugins/test_store_metadata.py
dzickusrh: Purely accidental.  I can modify to make it more realistic.

tests/plugins/test_koji_upload.py
dzickusrh: It seems like without 'build' in the metadata the following tests no longer make sense:
test_koji_upload_filesystem_koji_task_id
test_koji_upload_filesystem_koji_task_id_missing
test_koji_upload_log_task_id
test_koji_upload_wrong_source_type

tests/plugins/test_koji_upload.py
dzickusrh: Ok, I think I got most of what you wanted.  The 'build_id' piece relied on the promote plugin so I removed it.  Something to double check.

tests/plugins/test_koji_upload.py
```diff
@@ -0,0 +1,993 @@
...
+            .with_args(BUILD_ID+'-md', dict)
+            .replace_with(self.create_config_map))
+        (flexmock(OSBS)
+            .should_receive('get_config_map')
+            .with_args(BUILD_ID+'-md')
+            .replace_with(self.get_config_map))
+
+    def create_config_map(self, name, data):
+        assert isinstance(data, dict)
+        assert isinstance(name, basestring)
```
twaugh: This is causing Python3 tests to fail. Instead of using `isinstance(name, basestring)`, use `is_string_type(name)`.

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -0,0 +1,642 @@
...
+        try:
+            logs = self.osbs.get_build_logs(self.build_id)
+        except OsbsException as ex:
+            self.log.error("unable to get build logs: %r", ex)
+        else:
+            # Deleted once closed
+            logfile = NamedTemporaryFile(prefix=self.build_id,
+                                         suffix=".log",
+                                         mode='wb')
+            logfile.write(logs.encode('UTF-8'))
```
twaugh: This needs updating to match the recent change to koji_promote: https://github.com/projectatomic/atomic-reactor/commit/9e09d607d1c5b648fbbc92224ec300b930d0945e

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -0,0 +1,642 @@
...
+            logfile.flush()
+            metadata = self.get_output_metadata(logfile.name,
+                                                "openshift-final.log")
+            output.append(Output(file=logfile, metadata=metadata))
+
+        docker_logs = NamedTemporaryFile(prefix="docker-%s" % self.build_id,
+                                         suffix=".log",
+                                         mode='wb')
+        docker_logs.write("\n".join(self.workflow.build_result.logs).encode('UTF-8'))
```
twaugh: Change also needed here to match koji_promote. Probably in tests too.

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -0,0 +1,642 @@
...
+
+    def get_image_output(self, arch):
+        """
+        Create the output for the image
+
+        This is the Koji Content Generator metadata, along with the
+        'docker save' output to upload.
+
+        For metadata-only builds, an empty file is used instead of the
+        output of 'docker save'.
```
twaugh: This paragraph can be removed.

tests/plugins/test_koji_upload.py
```diff
@@ -0,0 +1,993 @@
...
+
+            # References one of the buildroots
+            assert len([buildroot for buildroot in buildroots
+                        if buildroot['id'] == buildroot_id]) == 1
+
+        files = session.uploaded_files
+
+        # There should be a file in the list for each output
+        # except for metadata-only imports, in which case there
+        # will be no upload for the image itself
```
twaugh: These two lines (about metadata-only) can be removed.

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -0,0 +1,642 @@
...
+        """
+
+        def add_buildroot_id(output):
+            logfile, metadata = output
+            metadata.update({'buildroot_id': buildroot_id})
+            return Output(file=logfile, metadata=metadata)
+
+        def add_log_type(output):
+            logfile, metadata = output
+            metadata.update({'type': 'log', 'arch': 'noarch'})
```
twaugh: We should move the `arch = os.uname()[4]` before this line and use it to set `'arch': arch` instead of 'noarch'.
http://osbs.readthedocs.io/en/latest/multiarch.html#koji-build

atomic_reactor/plugins/post_koji_upload.py
dzickusrh: I hand merged commit 9e09d607d1c5b648fbbc92224ec300b930d0945e, though I am not sure what the added test does, as it doesn't verify anything.

tests/plugins/test_store_metadata.py
twaugh: I think it's fine.

dzickusrh:
Alright, I can update those docs.

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -0,0 +1,643 @@
...
+        """
+        constructor
+
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param kojihub: string, koji hub (xmlrpc)
+        :param url: string, URL for OSv3 instance
+        :param verify_ssl: bool, verify OSv3 SSL certificate?
+        :param use_auth: bool, initiate authentication with OSv3?
+        :param koji_ssl_certs: str, path to 'cert', 'ca', 'serverca'
```
vrutkovs: Other plugin have `koji_ssl_certs_dir` paramer, only `koji_tag_build` and `koji_promote` have `koji_ssl_cert`.

Lets rename this parameter here, I'll make a PR to update the rest

tests/plugins/test_koji_upload.py
```diff
@@ -0,0 +1,1092 @@
...
+        logger.should_receive('debug').times(expected_times)
+        upload_logger = KojiUploadLogger(logger, notable_percent=notable)
+        for offset in range(0, totalsize + step, step):
+            upload_logger.callback(offset, totalsize, step, 1.0, 1.0)
+
+
+class TestKojiUpload(object):
+    def test_koji_upload_failed_build(self, tmpdir, os_env):
+        session = MockedClientSession('')
+        osbs = MockedOSBS()
```
vrutkovs: Stickler is right - `osbs` is not needed here

dzickusrh: Hi Vadim,

That should have been resolved in the latest version late last week.

dzickusrh: Sorry, resolved by actually using the variable 'osbs'.

atomic_reactor/plugins/post_koji_upload.py
dzickusrh: This is easy to do for koji_upload.  I still struggle where this input comes from.  Is it osbs-client's json file?  Do we need to make changes on the osbs-client too?

vrutkovs: Yeah, it comes from osbs-client. I'll prepare a PR for osbs-client after this one gets merged

twaugh: @vrutkovs: note that the osbs-client work is already tracked and currently assigned to Mark.

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -0,0 +1,643 @@
...
+        :param kojihub: string, koji hub (xmlrpc)
+        :param url: string, URL for OSv3 instance
+        :param verify_ssl: bool, verify OSv3 SSL certificate?
+        :param use_auth: bool, initiate authentication with OSv3?
+        :param koji_ssl_certs_dir: str, path to 'cert', 'ca', 'serverca'
+        :param koji_proxy_user: str, user to log in as (requires hub config)
+        :param koji_principal: str, Kerberos principal (must specify keytab)
+        :param koji_keytab: str, keytab name (must specify principal)
+        :param blocksize: int, blocksize to use for uploading files
+        :param target: str, koji target
```
lcarva: This parameter is not used by the plugin.

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -0,0 +1,643 @@
...
+        }
+
+        try:
+            self.osbs.create_config_map(md_fragment, cm_data)
+        except OsbsException:
+            self.log.debug("metadata: %r", koji_metadata)
+            self.log.debug("annotations: %r", annotations)
+            raise
+        except AttributeError as e:
+            self.log.debug("missing attribute: %s", e.strerror)
```
lcarva: `AttributeError` exception has no `strerror` attribute.

```python
>>> o = object()
>>> try:
...     o.yolo
... except AttributeError as e:
...     pass
>>> e.strerror
AttributeError: 'exceptions.AttributeError' object has no attribute 'strerror'
'exceptions.AttributeError' object has no attribute 'strerror'
```

lcarva: It's probably better to not catch this exception and just let it propagate normally.

atomic_reactor/plugins/post_koji_upload.py
dzickusrh: I removed target and poll_interval.

atomic_reactor/plugins/post_koji_upload.py
dzickusrh: I bet this was caught on python3? :-)  Because I never ran into that issue. :-/  I think I put strerror to make flake8 happy.
I will remove that Exception case.

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -0,0 +1,636 @@
...
+        self.koji_ssl_certs_dir = koji_ssl_certs_dir
+        self.koji_proxy_user = koji_proxy_user
+
+        self.koji_principal = koji_principal
+        self.koji_keytab = koji_keytab
+
+        self.blocksize = blocksize
+
+        self.namespace = get_build_json().get('metadata', {}).get('namespace', None)
+        osbs_conf = Configuration(conf_file=None, openshift_uri=url,
```
lcarva: We also need to set the `build_json_dir` parameter here.
This is because osbs library will load the `config_map.json` as part of creating the ConfigMap object in OpenShift.

See how `import_image` plugin does this: https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/post_import_image.py#L77

In osbs-client you also want to make sure this is set correctly. 
See how `import_image` plugin is rendered in `render_import_image` method of `build_request.py`.
(Not sure if you or @mlangsdorf is doing the osbs-client portion)

dzickusrh: Yeah @mlangsdorf is doing the osbs-client piece. 
Not sure how it all works, but I will blindly add the build_json_dir in there.  I guess it also needs to be added to osbs-client too.

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -0,0 +1,639 @@
...
+        md_fragment = "{}-md".format(get_build_json()['metadata']['name'])
+        md_fragment_key = 'metadata.json'
+        cm_data = {md_fragment_key: koji_metadata}
+        annotations = {
+            "metadata_fragment": "configmap/" + md_fragment,
+            "metadata_fragment_key": md_fragment_key
+        }
+
+        try:
+            self.osbs.create_config_map(md_fragment, cm_data)
```
lcarva: It appears that `ConfigMaps` do not allow nested objects, which is a problem here because md_fragment has multiple nested objects, including lists.

How about we base64 encode the data?
1. `create_config_map` encodes data as base64 string, and creates config map.
2. `ConfigMapResponse.get_data()` automatically base64 decodes data.
This way, the base64 encoding "trick" is hidden from the user.

@twaugh, @vrutkovs, @dzickusrh, thoughts?

twaugh: We can't use a JSON string?

lcarva: I guess we could. Not sure if that will cause any encoding issues though.
base64 would be more inline with the approach used by `Secret` Objects.

dzickusrh: So a trivial conversion like:
cm_data = {md_fragment_key: koji_metadata} to
cm_data = {md_fragment_key: json.dumps(koji_metadata)}

would do the trick right?

dzickusrh: Perhaps osbs-client can assert the data is a string, though it would have to read the dict.

lcarva: My concern with serializing the data as a json string from this plugin is that consumers of this value, must also know that the data must be deserialized. This might be ok, just not a very clean API IMO.

twaugh: How about this?:

- `create_config_map()` always JSON-serializes values that are instances of `collections.Mapping`
- `ConfigMapResponse` provides a `get_json()` method (cf requests.Response.json) to re-create the dict from the JSON string

Then koji_upload can continue to use `create_config_map()` the same way it is doing now, and fetch_worker_metadata can use the new ConfigMapResponse method instead of using get_data().

lcarva:
Sanity test in an integrated environment passed.

lcarva:
Updated release notes.

PR #708, opened 2017-06-20
add_labels_in_df: base image inspect not required

lcarva:
With the split of `add_filesystem` plugin to run on both orchestrator and worker builds, the tarball filesystem is not actually streamed into the docker engine for orchestrator build.
This causes the parent image to not be inspected when we're building base images. (The parent image in this case is just the loaded tarball filesystem).

This change is to allow the plugin to continue execution if the parent image has not been inspected.

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -271,18 +271,21 @@ class AddLabelsPlugin(PreBuildPlugin):
         info_url = MyFormatter().vformat(self.info_url_format, [], all_labels)
         self.labels['url'] = info_url
 
+    def _get_base_image_labels(self):
+        labels = {}
+
+        try:
+            labels = self.workflow.base_image_inspect[INSPECT_CONFIG]['Labels'] or {}
+        except (KeyError):
+            self.log.debug("base image was not inspected")
+
+        return labels
+
     def run(self):
         """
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -271,18 +271,21 @@ def get_field(self, field_name, args, kwargs):
         info_url = MyFormatter().vformat(self.info_url_format, [], all_labels)
         self.labels['url'] = info_url
 
+    def _get_base_image_labels(self):
+        labels = {}
+
+        try:
+            labels = self.workflow.base_image_inspect[INSPECT_CONFIG]['Labels'] or {}
+        except (KeyError):
```
twaugh: Any reason for the parentheses?

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -271,18 +271,21 @@ def get_field(self, field_name, args, kwargs):
         info_url = MyFormatter().vformat(self.info_url_format, [], all_labels)
         self.labels['url'] = info_url
 
+    def _get_base_image_labels(self):
+        labels = {}
```
twaugh: Setting here instead of in the except block?

lcarva:
Dropping this as we don't necessarily need it anymore.

PR #709, opened 2017-06-21
plugins/add_filesystem: split worker and orchestrator

jarodwilson:
In a case where we're building images for multiple architectures, we need
to be able to pass in a list of the relevant architectures we want to
build for. In order to properly create these multi-arch builds, we need to
split the add_filesystem plugin's code up a bit, into portions that run on
the orchestrator, which organizes which workers should be utilized for
different architectures specified to build arch-specific images, and the
portion that runs on those workers.

This adds new parameters to add_filesystem, arch and architectures, which
are consumed by the worker and orchestrator builds respectively, and
overwrite image-build.conf's arches var (which we do log).

Outside of that, the majority of the work is merely reorganizing code and
conditionalizing what gets run depending on being invoked as an
orchestrator or as a worker, and then tweaking self-tests to compensate
for certain paths requiring more information provided to complete
successfully. We also now match downloaded filesystems with architecture
added to the regex, so we don't try to do something like use a PowerPC
base image on x86_64, etc.

Theoretically, this should also be backwards-compatible with an older
osbs-client that doesn't know about the orchestrator/worker split, the
run_image_task and stream_filesystem functions should simply both get
called in a single pass.

While we're at it, fix the link for koji's image_build docs.

Also make sure the task ID from add_filesystem gets passed into
orchestrate_build, so that worker builds can properly download base
images from said task.

Successfully tested on staging server, both with the new arrangement
version 2 and the prior arrangement version 1 in the matching osbs-client
changes.


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -238,8 +239,25 @@ def _make_labels(self):
...
+                try:
+                    task_id = int(fs_task_id)
+                except ValueError:
+                    self.log.error("invalid taskID %r", fs_task_id, exc_info=1)
+
         try:
             kwargs = self.get_worker_build_kwargs(release, cluster_info.platform)
+            self.log.debug("%s: setting filesystem_koji_task_id to %d",
+                           AddFilesystemPlugin.key, task_id)
+            kwargs['filesystem_koji_task_id'] = task_id
```
lcarva: Should the value be passed to `get_worker_build_kwargs` and set there?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -238,8 +239,25 @@ def _make_labels(self):
 
     def do_worker_build(self, release, cluster_info):
         build = None
+        task_id = None
```
lcarva: Maybe move this logic to a helper method?

atomic_reactor/plugins/build_orchestrate_build.py
jarodwilson: Yeah, passing it and setting it in there would be more consistent, didn't look closely enough at that. Change pushed that does that. I'm not sure what you mean by "if it's truthy" though. Looks like we should throw an exception if we get a non-numeric task ID, not sure what other sort of verification we can do here on the fly?

atomic_reactor/plugins/build_orchestrate_build.py
jarodwilson: Done. Pushed a change with a broken-out get_fs_task_id() created.

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: By truthy I mean a value that evaluates to true in an expression, which should handle not passing the parameter if the value is None for example. 
A simple `if task_id` would do exactly that.

jarodwilson: Ah. So get_worker_build_kwargs() doesn't do that for release or platform, so I'd just went ahead with doing the same for them. If I were to add an 'if task_id' there, should the same be done for release and platform?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -236,10 +238,32 @@ def _make_labels(self):
...
+            try:
+                fs_task_id = fs_result['filesystem-koji-task-id']
+            except KeyError:
+                self.log.error("%s: expected filesystem-koji-task-id in result",
+                               AddFilesystemPlugin.key)
+            else:
+                try:
+                    task_id = int(fs_task_id)
+                except ValueError:
+                    self.log.error("invalid taskID %r", fs_task_id, exc_info=1)
```
twaugh: This block (from line 245) doesn't get execution coverage during unit testing. I think we need to have additional tests in tests/test_orchestrate_build.py for that.

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -298,12 +312,35 @@ def run(self):
...
+        if not image_build_conf or image_build_conf == 'latest':
+            image_build_conf = 'image-build.conf'
+
+        self.session = create_koji_session(self.koji_hub, self.koji_auth_info)
+
+        task_id, filesystem_regex = self.run_image_task(image_build_conf)
+
+        new_base_image = None
+        if not self.is_orchestrator:
+            new_base_image = self.stream_filesystem(task_id, filesystem_regex)
```
twaugh: Could we have a unit test to verify that it does not try to stream the filesystem (i.e. that `downloadTaskOutput` is never called) when `architectures` is set?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -194,13 +195,14 @@ def get_release(self):
...
 
-    def get_worker_build_kwargs(self, release, platform):
+    def get_worker_build_kwargs(self, release, platform, task_id):
         build_kwargs = deepcopy(self.build_kwargs)
 
         build_kwargs.pop('architecture', None)
 
         build_kwargs['release'] = release
         build_kwargs['platform'] = platform
+        build_kwargs['filesystem_koji_task_id'] = task_id
```
lcarva: Can we do this only if `task_id` is truthy?
```python
if task_id:
    build_kwargs['filesystem_koji_task_id'] = task_id
```

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: BTW, if you change this to exception level, the stack trace is automatically logged:
```python
self.log.exception("invalid taskID %r", fs_task_id)
```

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -236,10 +238,32 @@ def _make_labels(self):
 
         return labels
 
+    def get_fs_task_id(self):
+        task_id = None
+        fs_result = self.workflow.prebuild_results.get(AddFilesystemPlugin.key)
+        if fs_result is not None:
+            try:
```
lcarva: I wonder if this is easier to read:
```python
try:
    task_id = int(fs_result['filesystem-koji-task-id'])
except KeyError:
    self.log.error("expected filesystem-koji-task-id in result of %s plugin",
                   AddFilesystemPlugin.key)
except ValueError:
    self.log.exception("invalid taskID")
```

Also, there's no need to log the actual invalid task_id value. The `ValueError` exception will provide it:
```python
>>> int('asfdasfd')
ValueError: invalid literal for int() with base 10: 'asfdasfd'
invalid literal for int() with base 10: 'asfdasfd'
```

lcarva: If the add_filesystem plugin ran, do we want to fail if the task_id cannot be determined?
Meaning, should we silence the `KeyError` and `ValueError` exceptions here?

atomic_reactor/plugins/build_orchestrate_build.py
jarodwilson: Done.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -236,10 +239,29 @@ def _make_labels(self):
...
+        task_id = None
+
+        fs_result = self.workflow.prebuild_results.get(AddFilesystemPlugin.key)
+        try:
+            task_id = int(fs_result['filesystem-koji-task-id'])
+        except KeyError:
+            self.log.error("%s: expected filesystem-koji-task-id in result",
+                           AddFilesystemPlugin.key)
+        except ValueError:
+            self.log.exception("%s: returned an invalid taskID")
```
lcarva: Missing string format paramater here.

atomic_reactor/plugins/build_orchestrate_build.py
jarodwilson: So I'd initially dropped out the 'if fs_result is None' bit, and things went bad with self-tests. If there's no task_id, we should already return None before we go through the whole exception dance. If something is returned, and it's not a valid task ID, I think we do want to explode, because that shouldn't happen, right?

atomic_reactor/plugins/build_orchestrate_build.py
jarodwilson: D'oh.

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: Right, if there's a `fs_result`, meaning the `add_filesystem` plugin actually ran, and we can't grab the filesystem task ID something is wrong.

atomic_reactor/plugins/build_orchestrate_build.py
jarodwilson: Fumbling around working on a test for get_fs_task_id() right now...

atomic_reactor/plugins/pre_add_filesystem.py
jarodwilson: This would go into test_add_filesystem.py, I'm thinking, where there are some other checks to see that downloadTaskOutput *is* called, but I don't see offhand how to check that it is NOT called. Will have to read more, but pointers appreciated. :)

lcarva: flexmock can help you here. 
You can mock the function and ensure that it's never called:
```python
session.should_receive('downloadTaskOutput').never()
```

Looks like `mock_koji_session` can be enhanced to take a param to do this for you conditionally.

jarodwilson: Okay, I think I have something for this now.

jarodwilson: Note: I baked it into test_image_task_failure() though. Would it be preferred to just create a new test explicitly for this check?

atomic_reactor/plugins/build_orchestrate_build.py
jarodwilson: Okay, what did you have in mind here? I've been toying with this, and it seems I'm doing something wrong, since the self-tests I'm working on seem to never be able to catch an expected exception here.

atomic_reactor/plugins/pre_add_filesystem.py
twaugh: Yes please. If I deliberately vandalise https://github.com/projectatomic/atomic-reactor/pull/709/files#diff-3e989a171d89d87cc2197c178b8d0516R341 to make it always stream the tarball, the test should fail. Currently it still passes.

jarodwilson: Okay, I think the latest bits I just pushed actually finally implement this, and vandalism results in tests failing when they should.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -17,6 +17,7 @@
 from atomic_reactor.build import BuildResult
 from atomic_reactor.plugin import BuildStepPlugin
 from atomic_reactor.plugins.pre_reactor_config import get_config
+from atomic_reactor.plugins.pre_add_filesystem import AddFilesystemPlugin
```
lcarva: Let's avoid this import by moving the value of `AddFilesystemPlugin.key` to constants.py.
Both this plugin and `add_filesystem` should read from the constants file. 
See how `PLUGIN_PULP_SYNC_KEY` is done.

The `add_filesystem` plugin requires koji library. 
We try to isolate such plugin in its own subpackage.
Not all installation of atomic-reactor will include `add_filesystem` plugin, for example.

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -112,6 +115,13 @@ def __init__(self, tasker, workflow, koji_hub,
         self.poll_interval = poll_interval
         self.blocksize = blocksize
         self.repos = repos or []
+        if architectures:
```
lcarva: This could be simplified:
```python
self.architectures = architectures
self.is_orchestrator = True if self.architectures else False
```

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -98,6 +99,8 @@ def __init__(self, tasker, workflow, koji_hub,
                       base filesystem creation. First value will also
                       be used as install_tree. Only baseurl value is used
                       from each repo file.
+        :param architectures: str, list of arches to build on (orchestrator)
```
lcarva: Wrong type, should be `list<str>,`

jarodwilson: Done

atomic_reactor/plugins/pre_add_filesystem.py
jarodwilson: Done

atomic_reactor/plugins/build_orchestrate_build.py
jarodwilson: Okay, I think I have this mess sorted out now too.

twaugh:
Draft release notes for 1.6.25 updated.

PR #710, opened 2017-06-22
pre_cancel_builds_on_autorebuild plugin

maxamillion:


atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
```diff
@@ -0,0 +1,90 @@
...
+
+            osbs = OSBS(osbs_conf, osbs_conf)
+
+            try:
+                builds = osbs.list_builds(
+                    field_separator={"buildconfig": buildconfig}
+                )
+
+                for build in builds:
+                    if build.is_running():
```
lcarva: Don't we need to make sure build is not this current build? Otherwise, I think both builds will be cancelled.

PR #711, opened 2017-06-22
pre_cancel_builds_on_autorebuild plugin

maxamillion:

### DIFF
--- a/atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
+++ b/atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
@@ -47,7 +47,11 @@ class CancelBuildsOnAutoRebuild(PreBuildPlugin):
         buildconfig = labels["buildconfig"]
 
         self.log.info("This is a rebuild? %s", buildconfig)
-        if is_rebuild(self.workflow):
+        if not is_rebuild(self.workflow):
+            self.log.info(
+                'this is not an autorebuild, %s is doing nothing' % self.key
+            )
+        else:
             self.log.info(
                 'this is an autorebuild, determining if any previous builds need to be cancelled'
             )
@@ -77,10 +81,5 @@ class CancelBuildsOnAutoRebuild(PreBuildPlugin):
                         osbs.cancel_build(build.get_build_name())
 
             except OsbsResponseException as ex:
-                # FIXME - I'm not sure what a good 'graceful' handling of this
-                # would be.
-                raise
-        else:
-            self.log.info(
-                'this is not an autorebuild, %s is doing nothing' % self.key
-            )
+                log.exception("failed to cancel build %s", build.get_build_name())
+
--- a/atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
+++ b/atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
@@ -65,7 +65,7 @@ class CancelBuildsOnAutoRebuild(PreBuildPlugin):
 
             try:
                 builds = osbs.list_builds(
-                    field_separator={"buildconfig": buildconfig}
+                    field_selector={"buildconfig": buildconfig}
                 )
 
                 for build in builds:
@@ -77,13 +77,9 @@ class CancelBuildsOnAutoRebuild(PreBuildPlugin):
                         osbs.cancel_build(build.get_build_name())
 
             except OsbsResponseException as ex:
-                if ex.status_code == 409:
-                    # Someone else was modifying the build
-                    # configuration at the same time. Try again.
-                    self.log.debug("got status %d, retrying", ex.status_code)
-                    osbs.set_labels_on_build_config(buildconfig, labels)
-                else:
-                    raise
+                # FIXME - I'm not sure what a good 'graceful' handling of this
+                # would be.
+                raise
         else:
             self.log.info(
                 'this is not an autorebuild, %s is doing nothing' % self.key
--- a/atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
+++ b/atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
@@ -66,18 +66,35 @@ class CancelBuildsOnAutoRebuild(PreBuildPlugin):
             osbs = OSBS(osbs_conf, osbs_conf)
 
             try:
-                builds = osbs.list_builds(
-                    field_selector="buildconfig={}".format(buildconfig)
+                # Filter out builds that are not currently running
+                field_selector = ",".join(
+                    [
+                        "status!={status}".format(
+                            status=status.capitalize()
+                        )
+                     for status in BUILD_FINISHED_STATES
+                    ]
                 )
+                # Filter out builds that are not using this buildconfig
+                field_selector += ",buildconfig={}".format(buildconfig)
 
-                for build in builds:
+                builds = osbs.list_builds(field_selector=field_selector)
+            except OsbsResponseException as ex:
+                self.log.exception("failed to cancel build, unable to obtain list of builds from OSBS")
+                raise ex
+
+            last_ex = None
+            for build in builds:
+                try:
                     if build.is_running():
                         self.log.info(
                             "cancelling build %s in favor of autorebuild",
                             build.build_id
                         )
                         osbs.cancel_build(build.get_build_name())
+                except OsbsResponseException as ex:
+                    self.log.exception("failed to cancel build %s", build.get_build_name())
+                    last_ex = ex
 
-            except OsbsResponseException as ex:
-                self.log.exception("failed to cancel build %s", build.get_build_name())
-                raise ex
+            if last_ex is not None:
+                raise last_ex
--- a/atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
+++ b/atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
@@ -21,7 +21,6 @@ class CancelBuildsOnAutoRebuild(PreBuildPlugin):
     """
     key = 'cancel_builds_on_autorebuild'
 
-    # FIXME - Unsure if this should be allowed to fail
     is_allowed_to_fail = False
 
     def __init__(self, tasker, workflow, url, verify_ssl=True, use_auth=True):
@@ -58,7 +57,6 @@ class CancelBuildsOnAutoRebuild(PreBuildPlugin):
 
             osbs_conf = Configuration(
                 conf_file=None,
-                openshift_uri=self.url,
                 openshift_url=self.url,
                 use_auth=self.use_auth,
                 verify_ssl=self.verify_ssl,
@@ -69,7 +67,7 @@ class CancelBuildsOnAutoRebuild(PreBuildPlugin):
 
             try:
                 builds = osbs.list_builds(
-                    field_selector={"buildconfig": buildconfig}
+                    field_selector="buildconfig={}".format(buildconfig)
                 )
 
                 for build in builds:
@@ -82,3 +80,4 @@ class CancelBuildsOnAutoRebuild(PreBuildPlugin):
 
             except OsbsResponseException as ex:
                 self.log.exception("failed to cancel build %s", build.get_build_name())
+                raise ex
--- a/atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
+++ b/atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
@@ -81,5 +81,4 @@ class CancelBuildsOnAutoRebuild(PreBuildPlugin):
                         osbs.cancel_build(build.get_build_name())
 
             except OsbsResponseException as ex:
-                log.exception("failed to cancel build %s", build.get_build_name())
-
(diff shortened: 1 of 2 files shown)


tests/plugins/test_cancel_builds_on_autorebuild.py
```diff
@@ -0,0 +1,130 @@
...
+        build_json = {
+            "metadata": {
+                "labels": {
+                    "buildconfig": buildconfig,
+                    key: value,
+                }
+            }
+        }
+
+        build = flexmock(
```
maxamillion: This is flexmock'd but not actually used locally.

lcarva: Maybe call flexmock without storing return to a variable?

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
```diff
@@ -0,0 +1,90 @@
...
+                            build.build_id
+                        )
+                        osbs.cancel_build(build.get_build_name())
+
+            except OsbsResponseException as ex:
+                if ex.status_code == 409:
+                    # Someone else was modifying the build
+                    # configuration at the same time. Try again.
+                    self.log.debug("got status %d, retrying", ex.status_code)
+                    osbs.set_labels_on_build_config(buildconfig, labels)
```
twaugh: Copy and pasted? We want to retry the cancel_build(). Also this exception should be caught inside the loop, so we also continue to cancel any other running builds.

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
```diff
@@ -0,0 +1,90 @@
...
+                use_auth=self.use_auth,
+                verify_ssl=self.verify_ssl,
+                namespace=metadata.get('namespace', None)
+            )
+
+            osbs = OSBS(osbs_conf, osbs_conf)
+
+            try:
+                builds = osbs.list_builds(
+                    field_separator={"buildconfig": buildconfig}
```
twaugh: `field_selector`

maxamillion: Ugh ... typo. Thanks

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
maxamillion: Yeah, I pulled over the code to setup the OSBS session from another plugin. I very clearly didn't clean it up properly. Will fix.

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
```diff
@@ -0,0 +1,86 @@
...
+                        self.log.info(
+                            "cancelling build %s in favor of autorebuild",
+                            build.build_id
+                        )
+                        osbs.cancel_build(build.get_build_name())
+
+            except OsbsResponseException as ex:
+                # FIXME - I'm not sure what a good 'graceful' handling of this
+                # would be.
+                raise
```
twaugh: For the time being, let's just attempt each cancel_build call once, i.e.
```
try:
    osbs.cancel_build(build.get_build_name())
except OsbsResponseException:
    log.exception("failed to cancel build %s", build.get_build_name())
```

I'm going to write down a policy for retries etc.

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
```diff
@@ -0,0 +1,86 @@
...
+        else:
+            self.log.info(
+                'this is not an autorebuild, %s is doing nothing' % self.key
+            )
```
twaugh: Maybe the `if` condition should be reversed so the 'do nothing' block can exit early?
https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
maxamillion: +1

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
maxamillion: Makes sense. +1

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class CancelBuildsOnAutoRebuild(PreBuildPlugin):
+    """
+    If the current build is rebuild, we need to query OSBS to make sure there
+    are no other builds currently running for this container and build target
+    """
+    key = 'cancel_builds_on_autorebuild'
+
+    # FIXME - Unsure if this should be allowed to fail
```
lcarva: I think it should not be allowed to fail. If this plugins fails to cancel an already running build, it might as well cancel itself. Having two builds pushing content to the same repo simultaneously can cause further problems.

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
```diff
@@ -0,0 +1,84 @@
...
+                'this is not an autorebuild, %s is doing nothing' % self.key
+            )
+        else:
+            self.log.info(
+                'this is an autorebuild, determining if any previous builds need to be cancelled'
+            )
+
+            osbs_conf = Configuration(
+                conf_file=None,
+                openshift_uri=self.url,
```
lcarva: Just `openshift_url` should be sufficient.

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
```diff
@@ -0,0 +1,84 @@
...
+                openshift_url=self.url,
+                use_auth=self.use_auth,
+                verify_ssl=self.verify_ssl,
+                namespace=metadata.get('namespace', None)
+            )
+
+            osbs = OSBS(osbs_conf, osbs_conf)
+
+            try:
+                builds = osbs.list_builds(
```
lcarva: At this point, the list of builds includes the build running this plugin!
We definitely want to filter out this current build from the list of builds being processed.

Also, let's narrow down the query to only non-finalized builds.
See how osbs-client does: https://github.com/projectatomic/osbs-client/blob/master/osbs/cli/main.py#L134-L136

@twaugh do we need to move the "running" query logic to within `list_builds` for better reuse?

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
```diff
@@ -0,0 +1,84 @@
...
+                    field_selector={"buildconfig": buildconfig}
```
lcarva: The value is actually a string in the format: `'buildconfig={}'.format(buildconfig)`.

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
```diff
@@ -0,0 +1,84 @@
...
+                for build in builds:
+                    if build.is_running():
+                        self.log.info(
+                            "cancelling build %s in favor of autorebuild",
+                            build.build_id
+                        )
+                        osbs.cancel_build(build.get_build_name())
+
+            except OsbsResponseException as ex:
+                self.log.exception("failed to cancel build %s", build.get_build_name())
```
lcarva: Re-raise exception here to avoid two builds running at the same time?

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
maxamillion: +1

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
maxamillion: Will remove `openshift_uri`, I saw it in some other plugins and wasn't sure if it'd been fully deprecated.

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
maxamillion: +1 on filtering the query down as well as filtering out "this" build.

@twaugh thoughts on @lcarva's question about the `list_builds` logic? If it should be, I can include that in this patch set (if it makes sense to do so).

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
maxamillion: Ah, +1. My mistake. Thank you.

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
twaugh: Yes, we probably should. Then the CLI tool can also reuse it.

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
twaugh: We also still need to move this exception handling inside the loop so every build cancellation is attempted.

maxamillion: I'm sorry I don't follow, what loop?

twaugh: Right now it looks like:
```
try:
    for build in builds:
        osbs.cancel_build(build.get_build_name())
except:
    ...
```

but I think it should look more like this:

```
last_ex = None
for build in builds:
    try:
        osbs.cancel_build(build.get_build_name())
    except OsbsResponseException as ex:
        last_ex = ex
        ...

if last_ex is not None:
    raise last_ex
```

In other words, we should attempt to cancel every build in that list.

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
```diff
@@ -0,0 +1,83 @@
...
+    def run(self):
+
+        metadata = get_build_json().get("metadata", {})
+        labels = metadata.get("labels", {})
+        buildconfig = labels["buildconfig"]
+
+        self.log.info("This is a rebuild? %s", buildconfig)
+        if not is_rebuild(self.workflow):
+            self.log.info(
+                'this is not an autorebuild, %s is doing nothing' % self.key
```
twaugh: Because this is a logging method, the preferred way is `self.log.info('... %s ...', self.key)`.

tests/plugins/test_cancel_builds_on_autorebuild.py
```diff
@@ -0,0 +1,130 @@
...
+            build_id=self.test_build_name
+        )
+
+        atomic_reactor.plugins.pre_check_and_set_rebuild = flexmock(
+            is_rebuild=lambda x: True
+        )
+
+        flexmock(OSBS)
+        OSBS.should_receive("list_builds").and_return([build])
+        OSBS.should_receive("cancel_build").with_args(self.test_build_name)
```
twaugh: Also want to verify the situation where cancel_build raises an exception. Perhaps also use a list of builds?

tests/plugins/test_cancel_builds_on_autorebuild.py
maxamillion: @lcarva +1 makes sense

atomic_reactor/plugins/pre_cancel_builds_on_autorebuild.py
```diff
@@ -0,0 +1,104 @@
...
+                )
+                # Filter out builds that are not using this buildconfig
+                field_selector += ",buildconfig={}".format(buildconfig)
+
+                builds = osbs.list_builds(field_selector=field_selector)
+            except OsbsResponseException as ex:
+                self.log.exception(
+                    "failed to cancel build, unable to obtain list of builds from OSBS"
+                )
+                raise ex
```
twaugh: Can just `except OsbsResponseException: ... raise` when not using `ex` explicitly.

vrutkovs:
>do we need to set the builds policy in such a way that multiple builds are allowed in parallel for this plugin to have anything to do?

I think so, yes. Autorebuild builds should have `strategy: Parallel` - and this plugin would decide which build should be canceled.

Other builds should have `Serial` strategy

twaugh:
Does that mean we also have to update the BuildConfig to set that after explicitly instantiating a Build, in the same way the node selector for automated triggered rebuilds is set?

vrutkovs:
Yes, we'd have to update BuildConfig - that is done anyway when low priority nodelector is applied, right?

I don't think it would be necessary to update worker buildconfig - this plugin would ensure only build would start there, right?

maxamillion:
@twaugh @vrutkovs is this waiting on something from me?

twaugh:
No, waiting on a final decision on runPolicy (from me).

lcarva:
I'm going to close this as runPolicy has been decided to use "Serial". The use case handled by this plugin goes away since there will always ever be a single build at a time running for any given BuildConfig.

maxamillion:
:+1:

PR #712, opened 2017-06-23
Orchestration: exclude a cluster if builds cannot be listed

### LINKED ISSUE #664: build_orchestrate_build: get_current_builds should catch OsbsResponseException
If one of the workers cannot be contacted the build stops:
```
2017-03-08 08:43:33,546 - atomic_reactor.plugin - DEBUG - running plugin 'orchestrate_build'
2017-03-08 08:43:33,547 - atomic_reactor.plugin - INFO - running plugin instance with args: '{[...]}'
2017-03-08 08:43:33,548 - atomic_reactor.util - DEBUG - Parent Config ENV: [u'container=docker', u'PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin']
2017-03-08 08:43:33,550 - atomic_reactor.util - INFO - searching for dockerfile in '/tmp/tmpfst87u/rsyslog-docker' (local path None)
2017-03-08 08:43:33,550 - atomic_reactor.util - DEBUG - abs path = '/tmp/tmpfst87u/rsyslog-docker', local path = 'None'
2017-03-08 08:43:33,551 - atomic_reactor.util - DEBUG - Dockerfile found: '/tmp/tmpfst87u/rsyslog-docker/Dockerfile'
2017-03-08 08:43:34,605 - atomic_reactor.plugin - DEBUG - Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/atomic_reactor-1.6.22-py2.7.egg/atomic_reactor/plugin.py", line 239, in run
  File "/usr/lib/python2.7/site-packages/atomic_reactor-1.6.22-py2.7.egg/atomic_reactor/plugins/build_orchestrate_build.py", line 219, in run
    [self.choose_cluster(platform) for platform in platforms]
  File "/usr/lib/python2.7/site-packages/atomic_reactor-1.6.22-py2.7.egg/atomic_reactor/plugins/build_orchestrate_build.py", line 164, in choose_cluster
    config.get_enabled_clusters_for_platform(platform)]
  File "/usr/lib/python2.7/site-packages/atomic_reactor-1.6.22-py2.7.egg/atomic_reactor/plugins/build_orchestrate_build.py", line 157, in get_cluster_info
  File "/usr/lib/python2.7/site-packages/atomic_reactor-1.6.22-py2.7.egg/atomic_reactor/plugins/build_orchestrate_build.py", line 149, in get_current_builds
    return len(osbs.list_builds(field_selector=field_selector))
  File "build/bdist.linux-x86_64/egg/osbs/api.py", line 41, in catch_exceptions
    return func(*args, **kwargs)
  File "build/bdist.linux-x86_64/egg/osbs/api.py", line 109, in list_builds
    serialized_response = response.json()
  File "build/bdist.linux-x86_64/egg/osbs/http.py", line 380, in json
    raise OsbsResponseException(self.content, self.status_code)
OsbsResponseException: Unauthorized
```

The plugin should return exclude this worker and move on

vrutkovs:
Fixes #664 

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -25,6 +25,7 @@ from atomic_reactor.constants import PLUGIN_ADD_FILESYSTEM_KEY
 from osbs.api import OSBS
 from osbs.conf import Configuration
 from osbs.constants import BUILD_FINISHED_STATES
+from osbs.exceptions import OsbsException
 
 
 ClusterInfo = namedtuple('ClusterInfo', ('cluster', 'platform', 'osbs', 'load'))
@@ -141,6 +142,7 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
     """
 
     EXCLUDE_FILENAME = 'exclude-platform'
+    UNREACHABLE_CLUSTER_LOAD = object()
 
     key = 'orchestrate_build'
 
@@ -201,7 +203,13 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
 
         conf = Configuration(**kwargs)
         osbs = OSBS(conf, conf)
-        current_builds = self.get_current_builds(osbs)
+        try:
+            current_builds = self.get_current_builds(osbs)
+        except OsbsException:
+            self.log.exception("Error occurred while listing builds on %s",
+                               cluster.name)
+            return ClusterInfo(cluster, platform, osbs, self.UNREACHABLE_CLUSTER_LOAD)
+
         load = current_builds / cluster.max_concurrent_builds
         self.log.debug('enabled cluster %s for platform %s has load %s and active builds %s/%s',
                        cluster.name, platform, load, current_builds, cluster.max_concurrent_builds)
@@ -216,7 +224,14 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
             raise RuntimeError('No clusters found for platform {}!'
                                .format(platform))
 
-        selected = min(clusters, key=lambda c: c.load)
+        reachable_clusters = [cluster for cluster in clusters
+                              if cluster.load != self.UNREACHABLE_CLUSTER_LOAD]
+
+        if not reachable_clusters:
+            raise RuntimeError('All clusters for platform {} are unreachable!'
+                               .format(platform))
+
+        selected = min(reachable_clusters, key=lambda c: c.load)
         self.log.info('platform %s will use cluster %s',
                       platform, selected.cluster.name)
         return selected
--- a/tests/plugins/test_orchestrate_build.py
+++ b/tests/plugins/test_orchestrate_build.py
@@ -646,3 +646,47 @@ def test_orchestrate_build_get_fs_task_id(tmpdir, task_id, error):
     else:
         build_result = runner.run()
         assert not build_result.is_failed()
+
+
+@pytest.mark.parametrize('fail_at', ('all', 'first'))
+def test_orchestrate_build_failed_to_list_builds(tmpdir, fail_at):
+    workflow = mock_workflow(tmpdir)
+    mock_osbs()  # Current builds is a constant 2
+
+    mock_reactor_config(tmpdir, {
+        'x86_64': [
+            {'name': 'spam', 'max_concurrent_builds': 5},
+            {'name': 'eggs', 'max_concurrent_builds': 5}
+        ],
+    })
+
+    flexmock_chain = flexmock(OSBS).should_receive('list_builds').and_raise(OsbsException("foo"))
+
+    if fail_at == 'all':
+        flexmock_chain.and_raise(OsbsException("foo"))
+
+    if fail_at == 'first':
+        flexmock_chain.and_return(['a', 'b'])
+
+    runner = BuildStepPluginsRunner(
+        workflow.builder.tasker,
+        workflow,
+        [{
+            'name': OrchestrateBuildPlugin.key,
+            'args': {
+                'platforms': ['x86_64'],
+                'build_kwargs': make_worker_build_kwargs(),
+                'osbs_client_config': str(tmpdir),
+            }
+        }]
+    )
+    if fail_at != 'all':
+        build_result = runner.run()
+        assert not build_result.is_failed()
+
+        annotations = build_result.annotations
+        assert annotations['worker-builds']['x86_64']['build']['cluster-url'] == 'https://eggs.com/'
+    else:
+        with pytest.raises(PluginFailedException) as exc:
+            build_result = runner.run()
+        assert 'All clusters for platform x86_64 are unreachable' in str(exc)


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -160,7 +160,10 @@ def get_platforms(self):
     def get_current_builds(self, osbs):
         field_selector = ','.join(['status!={status}'.format(status=status.capitalize())
                                    for status in BUILD_FINISHED_STATES])
-        return len(osbs.list_builds(field_selector=field_selector))
+        try:
+            return len(osbs.list_builds(field_selector=field_selector))
+        except Exception:
+            return -1
```
twaugh: We must at least log the exception.

This is using an API method, so it's guaranteed to raise an OsbsException of some sort. Can we catch OsbsException instead of Exception?

tests/plugins/test_orchestrate_build.py
```diff
@@ -574,3 +574,41 @@ def mock_wait_for_build_to_finish(build_name):
...
+
+    mock_reactor_config(tmpdir, {
+        'x86_64': [
+            {'name': 'spam', 'max_concurrent_builds': 5},
+            {'name': 'eggs', 'max_concurrent_builds': 5}
+        ],
+    })
+
+    def mock_list_builds(**kwargs):
+        raise RuntimeError("foo")
```
twaugh: This should be an OsbsException of some sort for a real-world test case.

tests/plugins/test_orchestrate_build.py
```diff
@@ -574,3 +574,41 @@ def mock_wait_for_build_to_finish(build_name):
...
+
+    (flexmock(OSBS)
+        .should_receive('list_builds')
+        .replace_with(mock_list_builds)
+        .once())
```
twaugh: Why not `.and_raise(...)`?

atomic_reactor/plugins/build_orchestrate_build.py
vrutkovs: Good idea, implemented

tests/plugins/test_orchestrate_build.py
vrutkovs: Reworked this to raise first and return correct result the next time

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -110,6 +111,7 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
     """
 
     EXCLUDE_FILENAME = 'exclude-platform'
+    UNREACHABLE_CLUSTER_LOAD = -1
```
lcarva: Nitpick: if this is not meant to be used as a number, consider using a symbol-like value:
```python
UNREACHABLE_CLUSTER_LOAD = object()
```

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -160,7 +162,12 @@ def get_platforms(self):
     def get_current_builds(self, osbs):
         field_selector = ','.join(['status!={status}'.format(status=status.capitalize())
                                    for status in BUILD_FINISHED_STATES])
-        return len(osbs.list_builds(field_selector=field_selector))
+        try:
+            return len(osbs.list_builds(field_selector=field_selector))
+        except OsbsException:
+            self.log.error("Error occurred while listing builds on %s" % osbs.os_conf.conf_section,
```
lcarva: Use `self.log.exception` to print the stacktrace instead of `exc_info=1` param?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -185,7 +195,8 @@ def choose_cluster(self, platform):
             raise RuntimeError('No clusters found for platform {}!'
                                .format(platform))
 
-        selected = min(clusters, key=lambda c: c.load)
+        reachable_clusters = filter(lambda x: x.load != self.UNREACHABLE_CLUSTER_LOAD, clusters)
```
lcarva: What if all clusters are unreachable? This should be moved up so the `if not clusters` check is performed.

lcarva: Let's add a test case for this scenario as well.

atomic_reactor/plugins/build_orchestrate_build.py
vrutkovs: Fixed

atomic_reactor/plugins/build_orchestrate_build.py
vrutkovs: Good idea, added an additional test case and exception when all clusters are unreachable

atomic_reactor/plugins/build_orchestrate_build.py
vrutkovs: I'm not sure about that. Using a non-numeric value internally sounds like a good idea - but what if any other plugin would want to reuse that function?

I think its worth keeping it like this - using negative values as error codes

lcarva: Sure, but in either case it creates a special case that must be known to the caller. Having `-1` is no better than `object()` in that sense.

vrutkovs: Right, but at least it would be numerical. I'm thinking I should rewrite this and change `get_enabled_clusters_for_platform` instead

twaugh: It definitely doesn't belong in `get_enabled_clusters_for_platform`, which should only be concerned with the content of the configuration file.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -186,7 +196,13 @@ def choose_cluster(self, platform):
             raise RuntimeError('No clusters found for platform {}!'
                                .format(platform))
 
-        selected = min(clusters, key=lambda c: c.load)
+        reachable_clusters = list(filter(
+            lambda x: x.load != self.UNREACHABLE_CLUSTER_LOAD, clusters))
```
vrutkovs: Agree, there is no need to produce a generator here

atomic_reactor/plugins/build_orchestrate_build.py
vrutkovs: Updated

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -191,7 +193,12 @@ def get_platforms(self):
     def get_current_builds(self, osbs):
         field_selector = ','.join(['status!={status}'.format(status=status.capitalize())
                                    for status in BUILD_FINISHED_STATES])
-        return len(osbs.list_builds(field_selector=field_selector))
+        try:
+            return len(osbs.list_builds(field_selector=field_selector))
+        except OsbsException:
+            self.log.exception("Error occurred while listing builds on %s" %
+                               osbs.os_conf.conf_section)
+            return self.UNREACHABLE_CLUSTER_LOAD
```
twaugh: Maybe just `raise` here?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -202,6 +209,9 @@ def get_cluster_info(self, cluster, platform):
         conf = Configuration(**kwargs)
         osbs = OSBS(conf, conf)
         current_builds = self.get_current_builds(osbs)
+        if current_builds == self.UNREACHABLE_CLUSTER_LOAD:
```
twaugh: ...and maybe catch `OsbsException` here? i.e. wrap `self.get_current_builds(osbs)` in a `try:` block.

atomic_reactor/plugins/build_orchestrate_build.py
vrutkovs: I don't think `get_current_builds` should throw exceptions as `get_cluster_info` uses it - other functions might want some config params and ignore the fact its actually unreachable at the moment

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -201,7 +203,13 @@ def get_cluster_info(self, cluster, platform):
 
         conf = Configuration(**kwargs)
         osbs = OSBS(conf, conf)
-        current_builds = self.get_current_builds(osbs)
+        try:
+            current_builds = self.get_current_builds(osbs)
+        except OsbsException:
+            self.log.exception("Error occurred while listing builds on %s" %
```
lcarva: Preferred way is to use: `self.log.exception("...%s...", cluster.name)`
But it's really minor.

vrutkovs:
Right, fixed

lcarva:
Release notes updated.

PR #713, opened 2017-06-27
build_orchestrator: Save build_info in plugin's workspace

dzickusrh:
The build orchestrator plugin goes through a lengthy setup to establish
cluster and osbs connection info.  Instead of replicating the setup in
later plugins, just save the build_info and pass it along to other
plugins that need it.

Currently this is intended to be used by the new post_fetch_metadata
plugin to connect to the established osbs setup to retrieve configmap
data.

The data is saved as a dict with platform as the key in
workflow.plugin_workspace and provides a wrapper call to retrieve it,
given platform and workflow parameters.

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -26,6 +26,14 @@ from osbs.constants import BUILD_FINISHED_STATES
 ClusterInfo = namedtuple('ClusterInfo', ('cluster', 'platform', 'osbs', 'load'))
 
 
+def get_worker_build_info(workflow, platform):
+    """
+    Obtain worker build information for a given platform
+    """
+    workspace = workflow.plugin_workspace[OrchestrateBuildPlugin.key]
+    return workspace[platform]
+
+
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -26,6 +26,14 @@
 ClusterInfo = namedtuple('ClusterInfo', ('cluster', 'platform', 'osbs', 'load'))
 
 
+def get_worker_build_info(workflow, platform):
+    """
+    Obtain orchestrator workspace objects
```
twaugh: How about "Obtain worker build information for a given platform"?

lcarva:
travis is faling with:

> Could not find any downloads that satisfy the requirement docker-py (from -r requirements.txt (line 1))
> No distributions at all found for docker-py (from -r requirements.txt (line 1))

Not sure what that's all about.

vrutkovs:
Looks like a pypi outage - https://status.python.org/

lcarva:
Updated release notes.

PR #714, opened 2017-06-27
build_orchestrator: Add metadata fragments from koji_upload

dzickusrh:
When the new plugin, koji_upload, is enabled, it will pass metadata
from the worker to the orchestrator through an osbs configmap object.

The pointer to this configmap object will be passed through an annotations
object.  Currently, orchestrator plugins can not access this metadata
info because it is not saved.

Save the info inside get_annotations() if available.  This will allow
future plugins like post_fetch_worker_metadata to grab and utilize
this data.


lcarva:
Verified this is already in the release notes.

PR #715, opened 2017-06-27
gitignore: Add .cache and lib directories

dzickusrh:
These directories are transient from either running a local build or test
coverage.  Clean up git status by hiding them.

This mimics what osbs-client does.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,7 @@
 *.swp
 dist/
 *.egg-info/
+build/
+
+# Unit test / coverage reports
+.cache


dzickusrh:
Actually, I wanted build/ ignored but noticed osbs-client achieved this by ignoring lib and lib64.  I am open to whatever, just looking to hide build/. :-)

twaugh:
We could just ignore `build/`because we don't have any other directories with that name. In osbs-client it has an `osbs/build/` directory containing source modules, which is why it jumps through that particular hoop.

lcarva:
You can use a leading forward slash to specify an exact path:
```
/build
```
Will only ignore `build` dir at root of git repo.

> A leading slash matches the beginning of the pathname. For example, "/*.c" matches "cat-file.c" but not "mozilla-sha1/sha1.c".

https://git-scm.com/docs/gitignore

PR #716, opened 2017-06-28
add_labels_in_df: handle base images too

twaugh:
This is required due to:
https://github.com/projectatomic/osbs-client/pull/581

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -57,6 +57,7 @@ from atomic_reactor.util import get_docker_architecture, df_parser
 from osbs.utils import Labels
 import json
 import datetime
+import re
 import string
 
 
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -275,19 +276,22 @@ def run(self):
...
-            self.log.error(message)
-            raise RuntimeError(message)
-        else:
-            base_image_labels = config["Labels"] or {}
-
         dockerfile = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
 
         lines = dockerfile.lines
 
+        if re.match('^koji/image-build(:.*)?$', dockerfile.baseimage):
```
lcarva: I'd rather this case wasn't called out like this...
Why can't we make inspection of base image optional?
Falling back to an empty dict on `KeyError` seems like a much better solution.

twaugh: But what if the image inspection fails for some reason? e.g. devicemapper issue or whatever.

lcarva: Looking at the code, `KeyError` is only raised if `docker.errors.NotFound` is raised.
Any other errors should propagate, causing a build failure.

twaugh: OK, I misunderstood how it behaved. Feel free to change this...

lcarva:
Updated release notes.

PR #717, opened 2017-06-29
Make sure some labels - namely release and version - are always set

vrutkovs:
Release and version labels would be written explicitly even if any of
them has the same value as in base image.

This fixes an issue I've sometimes seen in integration test runs - base image has release = 1, so if we build a new image on top of it atomic-reactor won't write this label in the file:
```
label u'release' is already set to u'1'
```

In the end the image would have a correct label set, but `koji_promote` would fail, as it fetches labels from Dockerfile, not knowing the label is set in the base image.

That also means release and version won't be written in Dockerfile, stored in the image - this is why this PR sets the labels explicitely, instead of updating `koji_promote` code

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -64,6 +64,7 @@ import string
 class AddLabelsPlugin(PreBuildPlugin):
     key = "add_labels_in_dockerfile"
     is_allowed_to_fail = False
+    rewrite_whitelist = ['release', 'version']
 
     def __init__(self, tasker, workflow, labels,
                  dont_overwrite=("Architecture", "architecture"),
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -64,6 +64,7 @@
 class AddLabelsPlugin(PreBuildPlugin):
     key = "add_labels_in_dockerfile"
     is_allowed_to_fail = False
+    explicit_labels = ['release', 'version']
```
twaugh: Trying to think of the best name for this variable. What about `overwrite_whitelist`?

vrutkovs: Its still a bit different (the value of the label is not being overwritten, its just being set again for visibility), but its better than `explicit`. Updated

twaugh: `rewrite_whitelist` then? With a comment explaining?

vrutkovs: Yeah, that sounds about right

lcarva:
Release notes updated.

PR #719, opened 2017-07-03
sendmail: remove double slash from workdir and make sure it ends with a slash

vrutkovs:
This PR makes sure email notification contains a correct link

### DIFF
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -102,7 +102,8 @@ class SendMailPlugin(ExitPlugin):
         self.error_addresses = list(error_addresses)
         self.email_domain = email_domain
         self.koji_hub = koji_hub
-        self.koji_root = koji_root
+        # Make sure koji_root doesn't end with a slash for a prettier link
+        self.koji_root = koji_root[:-1] if koji_root and koji_root[-1] == '/' else koji_root
         self.koji_auth_info = {
             'proxyuser': koji_proxyuser,
             'ssl_certs_dir': koji_ssl_certs_dir,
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -243,7 +243,14 @@ def _get_logs_url(self):
             # We're importing this here in order to trap ImportError
             from koji import PathInfo
             pathinfo = PathInfo(topdir=self.koji_root)
-            url = urljoin(pathinfo.work(), pathinfo.taskrelpath(self.koji_task_id))
+            # Koji's work() outputs an invalid workdir, so we should fix it manually
+            # Replace first occurrence of double slash from the right
+            work_url = '/'.join(pathinfo.work().rsplit('//', 1))
```
lcarva: The double `//` is not what causes the broken link issue. But if you want to clean it, then in the constructor of this plugin just **remove** trailing slash. This would cause `PathInfo` to return the url without double forward slashes.
For instance:
```python
>>> koji.PathInfo(topdir='http://koji.example.com/kojiroot').work()
'http://koji.example.com/kojiroot/work'
>>> koji.PathInfo(topdir='http://koji.example.com/kojiroot/').work()
'http://koji.example.com/kojiroot//work'
```

The real problem is that when `urljoin` is used, it drops `'/work'` from the URL path. Regardless of whether or not the string returned by `PathInfo.work()` contains double forward slashes, `//`. 
Any variation of the following should fix the problem:
```python
url = '/'.join([pathinfo.work(), pathinfo.taskrelpath(self.koji_task_id)])
```

vrutkovs: Oh, that's a good idea, implemented that (including a prettier link)

tests/plugins/test_sendmail.py
```diff
@@ -205,7 +205,7 @@ class WF(object):
         }))
 
         session = MockedClientSession('', has_kerberos=True)
-        pathinfo = MockedPathInfo()
+        pathinfo = MockedPathInfo('https://koji')
```
lcarva: Can we add coverage for stripping trailing slash?

vrutkovs: Sure, implemented

lcarva:
Release notes updated.

PR #720, opened 2017-07-03
generate autolabels for aliases as well overwritting

rcerven:

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -64,10 +64,8 @@ import string
 class AddLabelsPlugin(PreBuildPlugin):
     key = "add_labels_in_dockerfile"
     is_allowed_to_fail = False
-    rewrite_whitelist = ['release', 'version']
 
-    def __init__(self, tasker, workflow, labels,
-                 dont_overwrite=("Architecture", "architecture"),
+    def __init__(self, tasker, workflow, labels, dont_overwrite=None,
                  auto_labels=("build-date",
                               "architecture",
                               "vcs-type",
@@ -86,6 +84,8 @@ class AddLabelsPlugin(PreBuildPlugin):
         :param dont_overwrite: iterable, list of label keys which should not be overwritten
                                if they are present in parent image
         :param auto_labels: iterable, list of labels to be determined automatically, if supported
+                            it should contain only new label names and not old label names,
+                            as they will be managed automatically
         :param aliases: dict, maps old label names to new label names - for each old name found in
                         base image, dockerfile, or labels argument, a label with the new name is
                         added (with the same value)
@@ -101,18 +101,20 @@ class AddLabelsPlugin(PreBuildPlugin):
         if not isinstance(labels, dict):
             raise RuntimeError("labels have to be dict")
         self.labels = labels
-        self.dont_overwrite = dont_overwrite
+        self.dont_overwrite = dont_overwrite or ()
         self.dont_overwrite_if_in_dockerfile = dont_overwrite_if_in_dockerfile
         self.aliases = aliases or Labels.get_new_names_by_old()
+        self.auto_labels = auto_labels or ()
         self.info_url_format = info_url_format
         self.equal_labels = equal_labels or []
         if not isinstance(self.equal_labels, list):
             raise RuntimeError("equal_labels have to be list")
 
-        self.generate_auto_labels(auto_labels)
-
-    def generate_auto_labels(self, auto_labels):
+    def generate_auto_labels(self, base_labels, df_labels, plugin_labels):
         generated = {}
+        all_labels = base_labels.copy()
+        all_labels.update(df_labels)
+        all_labels.update(plugin_labels)
 
         # build date
         dt = datetime.datetime.fromtimestamp(atomic_reactor_start_time)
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -138,14 +138,19 @@ def generate_auto_labels(self, auto_labels):
...
-            if lbl in self.labels:
-                self.log.info("label %r is set explicitly, not using generated value", lbl)
-                continue
-
-            if lbl in generated:
-                self.labels[lbl] = generated[lbl]
-            else:
-                self.log.warning("requested automatic label %r is not available", lbl)
+            for _, llist in Labels.LABEL_NAMES.items():
+                if lbl in llist:
```
vrutkovs: This is probably unnecessary - `auto_labels` is surely in `LABEL_NAMES`

rcerven: mmmm? well I am trying to find label and its full list of equivalent labels

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -138,14 +138,19 @@ def generate_auto_labels(self, auto_labels):
...
         for lbl in auto_labels:
-            if lbl in self.labels:
-                self.log.info("label %r is set explicitly, not using generated value", lbl)
-                continue
-
-            if lbl in generated:
-                self.labels[lbl] = generated[lbl]
-            else:
-                self.log.warning("requested automatic label %r is not available", lbl)
+            for _, llist in Labels.LABEL_NAMES.items():
```
vrutkovs: `Labels.LABEL_NAMES.values()`?

atomic_reactor/plugins/pre_add_labels_in_df.py
vrutkovs: `llist` here is a tuple of old-style and new-style labels from `LABEL_NAMES`. An autolabel would surely be in the tuple

rcerven: yes it will be but I need the 'correct' tuple so I can add all aliases

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -132,63 +134,57 @@ def generate_auto_labels(self, auto_labels):
...
             generated['vcs-ref'] = vcs.vcs_ref
 
-        for old, new in self.aliases.items():
-            self.log.info("old=%r new=%r", old, new)
-            if new in generated and old not in generated:
-                self.log.info("adding %r for compatibility", old)
-                generated[old] = generated[new]
+        for lbl in self.auto_labels:
+            if lbl not in generated:
+                self.log.warning("requested automatic label %r is not available", lbl)
```
vrutkovs: Not sure what this error is about

tests/plugins/test_add_labels.py
```diff
@@ -202,7 +202,6 @@ def test_add_labels_plugin(tmpdir, docker_tasker,
     ('vcs-url', DOCKERFILE_GIT),
     ('vcs-ref', DOCKERFILE_SHA1),
     ('com.redhat.build-host', 'the-build-host'),
-    ('Build_Host', 'the-build-host'),
```
vrutkovs: Why are we removing that?

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -319,29 +317,21 @@ def escape(s):
...
 
-            if (key in self.dont_overwrite_if_in_dockerfile) and (key in dockerfile.labels):
-                self.log.info("denying overwrite of label %r, using from Dockerfile", key)
-                continue
+            if key not in dockerfile.labels or dockerfile.labels[key] != value:
 
-            label = '"%s"="%s"' % (escape(key), escape(value))
-            self.log.info("setting label %r", label)
-            labels.append(label)
+                if key in self.dont_overwrite_if_in_dockerfile and key in dockerfile.labels:
```
vrutkovs: It doesn't make any sense to me:
```
if key not in dockerfile.labels or dockerfile.labels[key] != value:
    if key in self.dont_overwrite_if_in_dockerfile and key in dockerfile.labels:
```

Why are we first checking that key is not in `dont_overwrite` or value is different - and then checking if its there and value is the same?

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -64,10 +64,8 @@
 class AddLabelsPlugin(PreBuildPlugin):
     key = "add_labels_in_dockerfile"
     is_allowed_to_fail = False
-    rewrite_whitelist = ['release', 'version']
 
-    def __init__(self, tasker, workflow, labels,
-                 dont_overwrite=("Architecture", "architecture"),
+    def __init__(self, tasker, workflow, labels, dont_overwrite=None,
```
vrutkovs: Default `dont_overwrite` is not iterable as docstring claims to be

atomic_reactor/plugins/pre_add_labels_in_df.py
rcerven: well this was there before, and it might happen that you specify in auto_labels some labels, but you don't generate it, so it won't be found in generated

tests/plugins/test_add_labels.py
rcerven: auto_labels should use only the new Labels, and old alias names are sorted in add_aliases

atomic_reactor/plugins/pre_add_labels_in_df.py
rcerven: first we are checking IF we DONT have label in dockerfile OR value is different
in both cases we will be adding / overwriting 

2nd check is IF label is in dont_overwrite_if_in_dockerfile AND the label actually exists in dockerfile

atomic_reactor/plugins/pre_add_labels_in_df.py
rcerven: line 104 is making it iterable

lcarva: Nitpick, but if you want to simplify this, you can use `dont_overwrite=tuple()` to default to an empty tuple in method signature instead of in the method body.

rcerven: well I wanted to make it consistent with others, because we do it like this always

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -132,63 +134,57 @@ def generate_auto_labels(self, auto_labels):
...
-                        # set old label with value from new label if old was in base
-                        add_as_an_alias(new, old, is_old_inherited)
-                        continue
-                    self.log.warning("labels %r=%r and %r=%r should probably have same value",
-                                     old, all_labels[old], new, all_labels[new])
-
+            # new label doesn't exists but old label does
+            # add new label with value from old label
+            if new not in all_labels:
+                add_as_an_alias(new, old)
```
vrutkovs: +1

tests/plugins/test_add_labels.py
```diff
@@ -704,3 +706,125 @@ def test_add_labels_base_image(tmpdir, docker_tasker, parent, should_fail,
...
+    ('A',   'B',   'A',   'C',   'C',   'C',   'as an alias for label'),
+    ('A',   'B',   'B',   'A',   'A',   'A',   'as an alias for label'),
+    ('A',   'B',   'B',   'B',   'B',   'B',   'already exists'),
+    ('A',   'B',   'B',   'C',   'C',   'C',   'as an alias for label'),
+    ('A',   'B',   'C',   'A',   'A',   'A',   'as an alias for label'),
+    ('A',   'B',   'C',   'B',   'B',   'B',   'as an alias for label'),
+    ('A',   'B',   'C',   'C',   'C',   'C',   'already exists'),
+    ('A',   'B',   'C',   'D',   'D',   'D',   'as an alias for label'),
+])
+def test_release_label(tmpdir, docker_tasker, caplog,
```
lcarva: I don't think this test should have old vs new style labels. It should only be concerned with `'release'` value from base image and `'release'` value from the Dockerfile of the layered image.

rcerven: well that test is checking what is in final Dockerfile, so it covers both

lcarva: I see that, but it doesn't need to. This test is difficult to read, let's make it simplify it.
There's already a test for aliases.

tests/plugins/test_add_labels.py
```diff
@@ -233,64 +233,63 @@ def test_add_labels_plugin_generated(tmpdir, docker_tasker, auto_label, value_re
...
-    ('A',   'B',   'A',   'C',   'A',   'C',  'should probably have same value'),  # noqa
-    ('A',   'B',   'B',   'A',   'B',   'A',  'should probably have same value'),  # noqa
-    ('A',   'B',   'B',   'B',   'B',   'B',  'already exists'                 ),  # noqa
-    ('A',   'B',   'B',   'C',   'B',   'C',  'should probably have same value'),  # noqa
-    ('A',   'B',   'C',   'A',   'C',   'A',  'should probably have same value'),  # noqa
-    ('A',   'B',   'C',   'B',   'C',   'B',  'should probably have same value'),  # noqa
-    ('A',   'B',   'C',   'C',   'C',   'C',  'already exists'                 ),  # noqa
-    ('A',   'B',   'C',   'D',   'C',   'D',  'should probably have same value'),  # noqa
+@pytest.mark.parametrize('base_old, base_new, df_old, df_new, exp_old, exp_new, exp_log', [
+    (None,  None,  None,  None,  None,  None,  None),
```
vrutkovs: Extra space here?

tests/plugins/test_add_labels.py
```diff
@@ -704,3 +704,75 @@ def test_add_labels_base_image(tmpdir, docker_tasker, parent, should_fail,
...
+                       base_new, df_new, plugin_new,
+                       expected_in_df, expected_log):
+    if MOCK:
+        mock_docker()
+
+    df_content = "FROM fedora\n"
+    plugin_labels = {}
+
+    if df_new:
+        df_content += 'LABEL release="{0}"\n'.format(df_new)
```
vrutkovs: I'm not sure why is `release` being tested there. Having `rewrite_whitelist` removed will this behaviour work on any label? I don't think that's what we want

lcarva: I specifically asked for this. I'd rather have duplicated coverage.

PR #724, opened 2017-07-07
Try harder to report reasons for errors

twaugh:

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -23,6 +23,7 @@ from atomic_reactor.plugins.pre_reactor_config import get_config
 from atomic_reactor.util import get_preferred_label, df_parser
 from atomic_reactor.constants import PLUGIN_ADD_FILESYSTEM_KEY
 from osbs.api import OSBS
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -100,11 +101,25 @@ def get_fail_reason(self):
...
+        except KeyError:
+            try:
+                build_name = self.build.get_build_name()
+                pod = self.osbs.get_pod_for_build(build_name)
+            except OsbsException:
+                pass
+            else:
+                try:
+                    fail_reason['pod'] = pod.get_failure_reason()
+                except AttributeError:
```
lcarva: Can we combine these try/except blocks somehow. Something about having 3 nested of these seems odd. Maybe:
```python
try:
  build_name = self.build.get_build_name()
  pod = self.osbs.get_pod_for_build(build_name)
  fail_reason['pod'] = pod.get_failure_reason()
except (OsbsException, AttributeError):
  pass
```

twaugh:
Updated.

lcarva:
Updated release notes.

PR #726, opened 2017-07-07
Koji import

mlangsdorf:
Still in progress and needs non-unit testing.

Almost all the tests from koji_promote are still enabled. I did remove the tests that dealt with tagging for the most part.


atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,280 @@
...
+            raise
+
+        # Older versions of CGImport do not return a value.
+        build_id = build_info.get("id") if build_info else None
+
+        self.log.debug("Build information: %s",
+                       json.dumps(build_info, sort_keys=True, indent=4))
+
+        # If configured, koji_tag_build plugin will perform build tagging
+        tag_later = are_plugins_in_order(self.workflow.exit_plugins_conf,
```
lcarva: I think this code exists in `koji_import` for backwards compatibility with plugin configurations that do not yet use `koji_tag_build` plugin. Can we assume that if someone is using this new plugin, they have already switched over to using `koji_tag_build`? I'm suggesting we start this new plugin without tagging support.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,280 @@
...
+            self.build_id = metadata["name"]
+        except KeyError:
+            self.log.error("No build metadata")
+            raise
+
+        metadata_version = 0
+
+        build = self.get_build(metadata)
+        """
+        # get all the platform names
```
lcarva: Let's remove this commented out code.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,280 @@
...
+
+        metadatas = self.workflow.postbuild_results.get(FetchWorkerMetadataPlugin.key)
+        buildroot = self.get_buildroot(metadatas)
+        output = self.get_output(metadatas)
+
+        koji_metadata = {
+            'metadata_version': metadata_version,
+            'build': build,
+            'buildroots': [buildroot],
+            'output': [output],
```
lcarva: In `koji_promote` this is a list of multiple items, corresponding to each `output_file`. Here it appears to be constructing a list of a single item. Should this be `'output': output` instead?

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,280 @@
+"""
+Copyright (c) 2015 Red Hat, Inc
```

atomic_reactor/plugins/exit_koji_import.py

atomic_reactor/plugins/exit_koji_import.py

atomic_reactor/plugins/exit_koji_import.py
mlangsdorf: get_output() assumes multiple entries in metadatas['output'] and returns metadatas['output'], so I thought that [metadatas['output']] would be a list of multiple items.

as far as I can tell, it's just a pass-through of the metadata input with the modification of the buildroot_id.

lcarva: If `metadatas['output'] is a list: ['one', 'two', 'three'], and you wrap that in brackets: [['one', 'two', 'three']], you now have a list of one item. That item in itself is another list of multiple items. Don't think you want the surrounding brackets.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,262 @@
...
+
+    If Kerberos is used for authentication, the default principal will
+    be used (from the kernel keyring) unless both koji_keytab and
+    koji_principal are specified. The koji_keytab parameter is a
+    keytab name like 'type:name', and so can be used to specify a key
+    in a Kubernetes secret by specifying 'FILE:/path/to/key'.
+
+    If metadata_only is set, the 'docker save' image will not be
+    uploaded, only the logs. The import will be marked as
+    metadata-only.
```
twaugh: This paragraph, and all the parts relating to `metadata_only`, can be removed.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,262 @@
...
+    """
+
+    key = PLUGIN_KOJI_IMPORT_PLUGIN_KEY
+    is_allowed_to_fail = False
+
+    def __init__(self, tasker, workflow, kojihub, url,
+                 verify_ssl=True, use_auth=True,
+                 koji_ssl_certs=None, koji_proxy_user=None,
+                 koji_principal=None, koji_keytab=None,
+                 metadata_only=False, target=None,
```
twaugh: Remove metadata_only

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,262 @@
...
+        :param workflow: DockerBuildWorkflow instance
+        :param kojihub: string, koji hub (xmlrpc)
+        :param url: string, URL for OSv3 instance
+        :param verify_ssl: bool, verify OSv3 SSL certificate?
+        :param use_auth: bool, initiate authentication with OSv3?
+        :param koji_ssl_certs: str, path to 'cert', 'ca', 'serverca'
+        :param koji_proxy_user: str, user to log in as (requires hub config)
+        :param koji_principal: str, Kerberos principal (must specify keytab)
+        :param koji_keytab: str, keytab name (must specify principal)
+        :param metadata_only: bool, whether to omit the 'docker save' image
```
twaugh: Remove metadata_only

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,262 @@
...
+        super(KojiImportPlugin, self).__init__(tasker, workflow)
+
+        self.kojihub = kojihub
+        self.koji_ssl_certs = koji_ssl_certs
+        self.koji_proxy_user = koji_proxy_user
+
+        self.koji_principal = koji_principal
+        self.koji_keytab = koji_keytab
+
+        self.metadata_only = metadata_only
```
twaugh: Remove metadata_only

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,262 @@
...
+from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
+from atomic_reactor.plugins.pre_add_help import AddHelpPlugin
+from atomic_reactor.constants import PLUGIN_KOJI_IMPORT_PLUGIN_KEY
+from atomic_reactor.util import (get_build_json, get_preferred_label, df_parser)
+from atomic_reactor.koji_util import create_koji_session
+from osbs.conf import Configuration
+from osbs.api import OSBS
+
+# An output file and its metadata
+Output = namedtuple('Output', ['file', 'metadata'])
```
twaugh: This is no longer used.

atomic_reactor/constants.py
```diff
@@ -62,6 +62,7 @@
 TAG_NAME_REGEX = r'^[\w][\w.-]{0,127}$'
 
 PLUGIN_KOJI_PROMOTE_PLUGIN_KEY = 'koji_promote'
+PLUGIN_KOJI_IMPORT_PLUGIN_KEY = 'koji_import'
```
twaugh: The reason for doing this is so the exit_store_metadata_in_osv3.py plugin can collect the Koji build ID and store it in the OpenShift Build annotation.

Please alter that plugin so it sets koji-build-id from whichever plugin ran, either koji_import or (failing that) koji_promote.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,262 @@
...
+        except KeyError:
+            self.log.error("No build metadata")
+            raise
+
+        metadata_version = 0
+        build = self.get_build(metadata)
+
+        metadatas = self.workflow.postbuild_results.get(FetchWorkerMetadataPlugin.key)
+        buildroot = self.get_buildroot(metadatas)
+        output = self.get_output(metadatas)
```
twaugh: Very minor suggestion: 'buildroot' and 'output' could both be named as plurals, like 'buildroots = self.get_buildroots(metadatas)'.

It's really confusing and unfortunate that the CG metadata format requires 'buildroots' (correct, makes sense) but 'output' (which ought to be called 'outputs' as the value is a list).

Anyway, it's fine as-is, just might be easier to read if the only 'huh?' thing is `'output': outputs,`, which is the fault of the CG metadata format spec.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,262 @@
...
+                    extra['filesystem_koji_task_id'] = task_id
+
+        help_result = self.workflow.prebuild_results.get(AddHelpPlugin.key)
+        if isinstance(help_result, dict) and 'help_file' in help_result and 'status' in help_result:
+            if help_result['status'] == AddHelpPlugin.NO_HELP_FILE_FOUND:
+                extra['image']['help'] = None
+            elif help_result['status'] == AddHelpPlugin.HELP_GENERATED:
+                extra['image']['help'] = help_result['help_file']
+            else:
+                self.log.error("Unknown result from add_help plugin: %s", help_result)
```
twaugh: Hmm, the add_help plugin won't run in the orchestrator build. I think we'll need to marshal this information across from the worker builds via OpenShift Build annotations.

One possible way to do this:

- move this logic to exit_store_metadata_in_osv3.py, i.e. have it find the result from the add_help plugin and store the 'help_file' value in a new annotation, perhaps "help" (so we'd have e.g. `"help": None` if no help file found, `"help": pathname` if there was a help file, and no "help" annotation at all if the plugin didn't run).

- here in koji_import, examine the "help" annotations from the worker builds. If none are present, don't add an image.help key. If any are present, use any value that is not `None`, or else use `None`.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,262 @@
...
+            'version': version,
+            'release': release,
+            'source': "{0}#{1}".format(source.uri, source.commit_id),
+            'start_time': start_time,
+            'end_time': int(time.time()),
+            'extra': extra,
+        }
+
+        if self.metadata_only:
+            build['metadata_only'] = True
```
twaugh: Remove metadata_only

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def get_output(self, metadatas):
+        """
+        Build the output entry of the metadata.
+
+        :return: list, containing dicts of partial metadata
+        """
+
+        for instance in metadatas['output']:
+            instance['buildroot_id'] = instance['arch'] + '-' + repr(instance['buildroot_id'])
```
twaugh: Maybe use `"{arch}-{buildroot_id}".format(**instance)`?

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,262 @@
...
+        return metadatas['output']
+
+    def get_buildroot(self, metadatas):
+        """
+        Build the buildroot entry of the metadata.
+
+        :return: list, containing dicts of partial metadata
+        """
+        for instance in metadatas['buildroots']:
+            instance['id'] = instance['container']['arch'] + '-' + repr(instance['id'])
```
twaugh: Maybe use `"{arch}-{id}".format(arch=instance['container']['arch'], id=instance['id'])`?

Might avoid ugliness with quotes if instance['id'] turned out to be a string of some sort, and issues with concatenation if one of these things ended up being/not being a unicode string.

atomic_reactor/plugins/exit_koji_import.py
twaugh: Hmm, had a thought about this: I don't think we can store `None` in build annotations. So actually let's JSON-encode that value, e.g.

- "\"/path/to/help.1\"" for the case where there is a help.1 file
- "nil" for the case where there is not a help.1 file
- annotation not present for the case where the add_help plugin didn't run

twaugh: Described here now: https://github.com/projectatomic/osbs-docs/pull/25

mlangsdorf:
tested using
https://osbs.qa.engineering.redhat.com:8443/console/project/osbs-stage/browse/builds/osbs-buildroot-docker-private-mlangsdorf-increm-b2a1b/osbs-buildroot-docker-private-mlangsdorf-increm-b2a1b-32?tab=details

to produce
https://osbs.qa.engineering.redhat.com:8443/console/project/osbs-stage/browse/builds/helloworld-container-master-9bc2a/helloworld-container-master-9bc2a-18

with logs
2017-07-13 12:09:20,523 - atomic_reactor.plugin - DEBUG - running plugin 'koji_import'
2017-07-13 12:09:20,524 - atomic_reactor.plugin - INFO - running plugin instance with args: '{u'kojihub': u'https://brewhub.stage.engineering.redhat.com/brewhub', u'url': u'https://osbs.qa.engineering.redhat.com:8443/', u'koji_keytab': False, u'koji_ssl_certs': u'/var/run/secrets/atomic-reactor/kojisecret', u'verify_ssl': False, u'use_auth': True}'
2017-07-13 12:09:20,525 - atomic_reactor.koji_util - INFO - Using SSL certificates for Koji authentication
2017-07-13 12:09:21,861 - atomic_reactor.util - DEBUG - Parent Config ENV: [u'container=docker', u'PKGM=yum', u'PATH=/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin']
2017-07-13 12:09:25,893 - atomic_reactor.plugins.koji_import - DEBUG - Build information: {
    "build_id": 1015736, 
    "completion_time": "2017-07-13 12:09:21", 
    "completion_ts": 1499947761.0, 
    "creation_event_id": 13351746, 
    "creation_time": "2017-07-13 12:09:22.115564", 
    "creation_ts": 1499947762.11556, 
    "epoch": null, 
    "extra": {
        "image": {
            "autorebuild": false
        }
    }, 
    "id": 1015736, 
    "name": "chromium", 
    "nvr": "chromium-0.0.1-10", 
    "owner_id": 3436, 
    "owner_name": "osbs", 
    "package_id": 253, 
    "package_name": "chromium", 
    "release": "10", 
    "source": "git://git.engineering.redhat.com/users/mlangsdo/helloworld-container.git#67f8da9b0391499a0cd5de1e6f8a231bd0bef93a", 
    "start_time": "2017-07-13 12:05:16", 
    "start_ts": 1499947516.0, 
    "state": 1, 
    "task_id": null, 
    "version": "0.0.1", 
    "volume_id": 0, 
    "volume_name": "DEFAULT"
}
2017-07-13 12:09:26,159 - atomic_reactor.plugin - DEBUG - plugin 'koji_import' finished in 5s

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -191,6 +195,15 @@ def run(self):
             "plugins-metadata": json.dumps(self.get_plugin_metadata())
         }
 
+        help_result = self.workflow.prebuild_results.get(AddHelpPlugin.key)
+        if isinstance(help_result, dict) and 'help_file' in help_result and 'status' in help_result:
+            if help_result['status'] == AddHelpPlugin.NO_HELP_FILE_FOUND:
+               annotations['help_file'] = json.dump(None)
```
twaugh: Think you want `json.dumps(None)` here.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -191,6 +195,15 @@ def run(self):
...
+            elif help_result['status'] == AddHelpPlugin.HELP_GENERATED:
+                annotations['help_file'] = json.dump(help_result['help_file'])
```
twaugh: Think you want `json.dumps()` here.

mlangsdorf:
`        results = [
            {
                'id': 'x86_64-1',
                'container': {
                    'arch': 'x86_64',
                    'type': 'docker',
                }
             },
            {'id': 'ppc64le-1',
                'container': {
                    'arch': 'ppc64le',
                    'type': 'docker',
                }
             },
        ]
`

gets reordered between python2 and python3. is there any way to prevent this?

atomic_reactor/constants.py
mlangsdorf: fixed.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,264 @@
...
+
+        # get the annotations from the worker BuildResponses
+        valid_help_file = False
+        help_file = None
+        for platform in worker_metadatas:
+            worker_annotations = get_worker_build_info(self.workflow, platform)
+            if worker_annotations and 'help_file' in worker_annotations:
+                valid_help_file = True
+                tmp_help_file = json.loads(worker_annotations['help_file'])
+                if tmp_help_file is not None:
```
twaugh: If the annotation has `None`, we want to store that in the CG metadata.

Rationale: it should be clear for Koji clients from the Koji Build metadata whether there is a help file in the git repository. If 'help' says None then they know there isn't and won't have to go and look for it themselves.

They should only have to look for it themselves if the annotation is not present at all, meaning 'I don't know'.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -191,6 +195,15 @@ def run(self):
...
         }
 
+        help_result = self.workflow.prebuild_results.get(AddHelpPlugin.key)
+        if isinstance(help_result, dict) and 'help_file' in help_result and 'status' in help_result:
+            if help_result['status'] == AddHelpPlugin.NO_HELP_FILE_FOUND:
+                annotations['help_file'] = json.dumps(None)
+            elif help_result['status'] == AddHelpPlugin.HELP_GENERATED:
+                annotations['help_file'] = json.dumps(help_result['help_file'])
+            else:
+                self.log.error("Unknown result from add_help plugin: %s", help_result)
```
twaugh: Does this get test coverage? coveralls thinks it doesn't:
https://coveralls.io/builds/12383762/source?filename=atomic_reactor%2Fplugins%2Fexit_store_metadata_in_osv3.py#L200

atomic_reactor/plugins/exit_koji_import.py
mlangsdorf: I thought I got the logic right here.
help_file is None unless tmp_help_file is not None. But valid_help_file is set to True is there's any help_file in annotations, even if that help_file is None.

extra['image']['help'] is set if valid_help_file is True. If all instances of helpfile were None, valid_help_file will still be True and extra['image']['help'] will be set to the default value of help_file which is None.

Did I get the logic wrong?

twaugh: Oh, I see. Do we need the `if tmp_help_file is not None:` then? We could just always assign that variable to whatever `json.loads()` returns.

I wonder if it would be more readable like this:

```
help_metadata = {}
for ... in ...:
    if worker_annotations and 'help_file' in worker_annotations:
        help_metadata['help'] = json.loads(....)

extra['image'].update(help_metadata)
```

twaugh:
I think this all looks great, and the only reason I haven't approved it is the missing coverage. I think the readability could be improved for 'help' but it looks correct as-is.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
mlangsdorf: Fixed I think.

atomic_reactor/plugins/exit_koji_import.py
mlangsdorf: I knew there was a better Python way to do it but didn't know what it was. Thanks.

Fixed.

tests/plugins/test_koji_import.py
```diff
@@ -0,0 +1,1280 @@
...
+                    'filesize': 71268781
+                }
+            ]
+        }
+    }
+    workflow.plugin_workspace = {
+        OrchestrateBuildPlugin.key: {
+            WORKSPACE_KEY_UPLOAD_DIR: 'test-dir',
+            WORKSPACE_KEY_BUILD_INFO: {
+               'x86_64': {'help_file': json.dumps('help.md')}
```
twaugh: You actually want an object, like:
```
class BuildInfo(object):
    build = BuildResponse({'metadata': {'annotations': {'help_file': ...}}})

workflow.plugin_workspace = {
    OrchestrateBuildPlugin.key: {
        ...
        WORKSPACE_KEY_BUILD_INFO: {
            'x86_64': BuildInfo(),
        }
    }
}
```

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,257 @@
...
+                    self.log.error("invalid task ID %r", fs_task_id, exc_info=1)
+                else:
+                    extra['filesystem_koji_task_id'] = task_id
+
+        help_metadata = {}
+        for platform in worker_metadatas:
+            worker_annotations = get_worker_build_info(self.workflow, platform)
+            if worker_annotations and 'help_file' in worker_annotations:
+                help_metadata['help'] = json.loads(worker_annotations['help_file'])
+        extra['image'].update(help_metadata)
```
twaugh: I now realize why you wrote it the original way you did. But how about this?:
```
    all_annotations = [worker['metadata']['annotations'] for worker in get_worker_build_info(...).values()]
    help_known = ['help' in annotations for annotations in all_annotations]
    # Only set the 'help' key when any 'help_file' annotation is set
    if any(help_known):
        # See if any are not None
        for known, annotations in zip(help_known, all_annotations):
            if known:
                if annotations['help'] is not None:
                    metadata['help'] = annotatations['help']
                    break
        else:
            # They are all None
            metadata['help'] = None
```

twaugh:
Let's make sure to update the docs/plugins.md documentation and also the spec file (to package the plugin in the koji integration subpackage).

atomic_reactor/plugins/exit_koji_import.py
mlangsdorf: This clearly doesn't work because get_worker_build_info(...) requires a platform.

```
all_annotations = []
            all_annotations.append([worker['metadata']['annotations'] for worker in 
                           get_worker_build_info(self.workflow, platform)])
```

is a syntax error "TypeError: string indices must be integers" in get_worker_build_info()

since the old version was
```
            worker_annotations = get_worker_build_info(self.workflow, platform)
```

I'm confused about what the problem is.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def get_buildroot(self, worker_metadatas):
+        """
+        Build the buildroot entry of the metadata.
+
+        :return: list, containing dicts of partial metadata
+        """
+        buildroots = []
+
+        # force ordering to work around python2 to python3 bug
```
lcarva: Not really a bug. A `dict` is not guaranteed to be ordered. The order will be different when the hashing algorithm changes. For retaining order, use `ordereddict`.

In this case, if all you want is to use a sorted list of platforms:
```python
for platform in sorted(worker_metadatas.keys()):
   for instance in worker_metadatas[platform]['buildroots']:
      ...
```

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -0,0 +1,257 @@
...
+        platforms = []
+        for platform in worker_metadatas:
+            platforms.append(platform)
+        platforms.sort()
+        for platform in platforms:
+            for instance in worker_metadatas[platform]['buildroots']:
+                instance['id'] = platform + '-' + repr(instance['id'])
```
lcarva: Why use `repr` here?

atomic_reactor/plugins/exit_koji_import.py
lcarva: I think the problem that Tim is pointing out is that because the koji build is now being created from the orchestrator build,  the `'help'` value needs to be pulled from each worker build, and we're trying to combine those values into one.

In normal cases, the `'help'` value should be the same across all worker builds. But since each worker build will have its own binary buildroot image, there's the case in which for one platform the libraries needed for generating the help file are not installed. Tim's algorithm attempts to address that case.

Here's my understanding of what this block of code should do:

1. Only set the `'help'` key if any of the worker builds have set it
2. If multiple worker builds have set it, use any of the values; but the value of `None` should only be used if all worker builds have set it to `None`. In other words, a `str` value takes precedence over `None`.
3. The `'help'` key in orchestrator build will have a single value for help file (`str`, not a `list`)

I think the only thing needing change from Tim's suggestion is how `all_annotations` is computed. This might do it:
```python
    all_annotations = [get_worker_buildinfo(self.workflow, platform) for platform in worker_metadatas]
    help_known = ['help' in annotations for annotations in all_annotations]
    # Only set the 'help' key when any 'help_file' annotation is set
    if any(help_known):
        # See if any are not None
        for known, annotations in zip(help_known, all_annotations):
            if known:
                if annotations['help'] is not None:
                    metadata['help'] = annotatations['help']
                    break
        else:
            # They are all None
            metadata['help'] = None
```

Note: I think this whole block should be in its own method.

atomic_reactor/plugins/exit_koji_import.py
mlangsdorf: E       instance['id'] = platform + '-' instance['id']
E                                              ^
E   SyntaxError: invalid syntax

lcarva: Ah ok
btw, you can use string formatting which is preferred over string concatenation:
```python
instance['id'] = '{}-{}'.format(platform, instance['id'])
```

mlangsdorf:
```
2017-07-17 20:11:47,430 - atomic_reactor.plugin - DEBUG - running plugin 'koji_import'
2017-07-17 20:11:47,431 - atomic_reactor.plugin - INFO - running plugin instance with args: '{u'kojihub': u'https://brewhub.stage.engineering.redhat.com/brewhub', u'url': u'https://osbs.qa.engineering.redhat.com:8443/', u'koji_keytab': False, u'koji_ssl_certs': u'/var/run/secrets/atomic-reactor/kojisecret', u'verify_ssl': False, u'use_auth': True}'
2017-07-17 20:11:47,433 - atomic_reactor.koji_util - INFO - Using SSL certificates for Koji authentication
2017-07-17 20:11:48,757 - atomic_reactor.util - DEBUG - Parent Config ENV: [u'container=docker', u'PKGM=yum', u'PATH=/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin']
2017-07-17 20:11:52,035 - atomic_reactor.plugins.koji_import - DEBUG - Build information: {
    "build_id": 1015875,
    "completion_time": "2017-07-17 20:11:48",
    "completion_ts": 1500322308.0,
    "creation_event_id": 13352421,
    "creation_time": "2017-07-17 20:11:49.010259",
    "creation_ts": 1500322309.01026,
    "epoch": null,
    "extra": {
        "image": {
            "autorebuild": false,
            "help": "null"
        }
    },
    "id": 1015875,
    "name": "chromium",
    "nvr": "chromium-0.0.1-16",
    "owner_id": 3436,
    "owner_name": "osbs",
    "package_id": 253,
    "package_name": "chromium",
    "release": "16",
    "source": "git://git.engineering.redhat.com/users/mlangsdo/helloworld-container.git#67f8da9b0391499a0cd5de1e6f8a231bd0bef93a",
    "start_time": "2017-07-17 20:06:55",
    "start_ts": 1500322015.0,
    "state": 1,
    "task_id": null,
    "version": "0.0.1",
    "volume_id": 0,
    "volume_name": "DEFAULT"
}
2017-07-17 20:11:52,196 - atomic_reactor.plugin - DEBUG - plugin 'koji_import' finished in 4s
```

lcarva:
> Let's make sure to update the docs/plugins.md documentation and also the spec file (to package the plugin in the koji integration subpackage).

I'll take care of this as it also needs to be done for koji_parent plugin.

lcarva:
Release notes updated. 

Created https://github.com/projectatomic/atomic-reactor/pull/738 to address packaging and documentation.

PR #727, opened 2017-07-07
build_orchestrator_build: parse container.yaml for excluded and extra…

mlangsdorf:
… platforms

use PyYaml to get the platform.not and platform.only values. remove the
platform.not platforms from the set of available platforms and add the
platform.only platforms.

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -11,6 +11,7 @@ from collections import namedtuple
 from copy import deepcopy
 from multiprocessing.pool import ThreadPool
 
+import yaml
 import json
 import os
 import random
@@ -26,7 +27,6 @@ from osbs.api import OSBS
 from osbs.exceptions import OsbsException
 from osbs.conf import Configuration
 from osbs.constants import BUILD_FINISHED_STATES
-from osbs.exceptions import OsbsException
 
 
 ClusterInfo = namedtuple('ClusterInfo', ('cluster', 'platform', 'osbs', 'load'))
@@ -153,7 +153,7 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
     build_image from kwargs
     """
 
-    EXCLUDE_FILENAME = 'exclude-platform'
+    CONTAINER_FILENAME = 'container.yaml'
     UNREACHABLE_CLUSTER_LOAD = object()
 
     key = 'orchestrate_build'
@@ -185,22 +185,25 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
 
         self.worker_builds = []
 
-    def get_excluded_platforms(self):
-        df_dir = self.workflow.source.get_dockerfile_path()[1]
-        exclude_platforms = set()
-        exclude_path = os.path.join(df_dir, self.EXCLUDE_FILENAME)
-        if os.path.exists(exclude_path):
-            with open(exclude_path) as f:
-                for platform in f:
-                    platform = platform.strip()
-                    if not platform:
-                        continue
-                    exclude_platforms.add(platform)
-
-        return exclude_platforms
+    def make_list(self, value):
+        if not isinstance(value, list):
+            value = [value]
+        return value
 
     def get_platforms(self):
-        return self.platforms - self.get_excluded_platforms()
+        df_dir = self.workflow.source.get_dockerfile_path()[1]
+        excluded_platforms = set()
+        container_path = os.path.join(df_dir, self.CONTAINER_FILENAME)
+        if os.path.exists(container_path):
+            with open(container_path) as f:
+                data = yaml.load(f)
+                if data is None or 'platforms' not in data or data['platforms'] is None:
+                    return self.platforms
+                excluded_platforms = set(self.make_list(data['platforms'].get('not', [])))
+                only_platforms = set(self.make_list(data['platforms'].get('only', [])))
+                if only_platforms:
+                    self.platforms = self.platforms & only_platforms
+        return self.platforms - excluded_platforms
 
     def get_current_builds(self, osbs):
         field_selector = ','.join(['status!={status}'.format(status=status.capitalize())
--- a/tests/plugins/test_orchestrate_build.py
+++ b/tests/plugins/test_orchestrate_build.py
@@ -33,6 +33,7 @@ import json
 import os
 import pytest
 import time
+import yaml
 
 
 class MockSource(object):
@@ -480,7 +481,21 @@ def test_orchestrate_build_choose_clusters(tmpdir, clusters_x86_64,
         assert plat_annotations['build']['cluster-url'] == 'https://chosen_{}.com/'.format(platform)
 
 
-def test_orchestrate_build_exclude_platforms(tmpdir):
+@pytest.mark.parametrize(('platforms', 'platform_exclude', 'platform_only', 'result'), [
+    (['x86_64', 'powerpc64le'], '', 'powerpc64le', ['powerpc64le']),
+    (['x86_64', 'spam', 'bacon', 'toast', 'powerpc64le'], ['spam', 'bacon', 'eggs', 'toast'], '',
+     ['x86_64', 'powerpc64le']),
+    (['powerpc64le', 'spam', 'bacon', 'toast'], ['spam', 'bacon', 'eggs', 'toast'], 'powerpc64le',
+     ['powerpc64le']),
+    (['x86_64', 'bacon', 'toast'], 'toast', ['x86_64', 'powerpc64le'], ['x86_64']),
+    (['x86_64', 'toast'], 'toast', 'x86_64', ['x86_64']),
+    (['x86_64', 'spam', 'bacon', 'toast'], ['spam', 'bacon', 'eggs', 'toast'], ['x86_64',
+                                                                                'powerpc64le'],
+     ['x86_64']),
+    (['x86_64', 'powerpc64le'], '', '', ['x86_64', 'powerpc64le'])
+])
+def test_orchestrate_build_exclude_platforms(tmpdir, platforms, platform_exclude, platform_only,
+                                             result):
     workflow = mock_workflow(tmpdir)
     mock_osbs()
 
@@ -488,6 +503,12 @@ def test_orchestrate_build_exclude_platforms(tmpdir):
         'x86_64': [
             {
                 'name': 'worker01',
+                'max_concurrent_builds': 3
+            }
+        ],
+        'powerpc64le': [
+            {
+                'name': 'worker02',
                 'max_concurrent_builds': 3
             }
         ]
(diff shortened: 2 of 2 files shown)


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -171,22 +171,30 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
-        if os.path.exists(exclude_path):
-            with open(exclude_path) as f:
-                for platform in f:
-                    platform = platform.strip()
-                    if not platform:
+        excluded_platforms = set()
+        container_path = os.path.join(df_dir, self.CONTAINER_FILENAME)
+        add_or_exclude = 0
+        if os.path.exists(container_path):
+            with open(container_path) as f:
```
lcarva: Let's use PyYaml here for parsing.

mlangsdorf: revised with PyYaml

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -171,22 +172,29 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+        excluded_platforms = set()
+        container_path = os.path.join(df_dir, self.CONTAINER_FILENAME)
+        if os.path.exists(container_path):
+            with open(container_path) as f:
+                data = yaml.load(f)
+                if 'platforms' not in data:
+                    return self.platforms
+                for action, platforms in data['platforms'].items():
+                    if 'not' in action:
+                        if type(platforms) == str:
```
twaugh: I'd prefer a test like `isinstance(platforms, list)`.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -171,22 +172,29 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+                            excluded_platforms.add(platforms)
+                        else:
+                            for x in platforms:
+                                excluded_platforms.add(x)
+                    elif 'only' in action:
```
twaugh: I can see a use-case for having both `not` and `only`: one of them is temporary.

Should this be `if` instead of `elif`?

mlangsdorf: any given instance of action is one of 'not', 'only', or something else that can be ignored. I think elif is appropriate here.

if you check the test case, config.yaml file contains both not: and only: platforms.

atomic_reactor/plugins/build_orchestrate_build.py
mlangsdorf: fixed.

atomic_reactor/plugins/build_orchestrate_build.py
twaugh: Oh, right, of course. :-)

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -171,22 +172,29 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+        df_dir = self.workflow.source.get_dockerfile_path()[1]
+        excluded_platforms = set()
+        container_path = os.path.join(df_dir, self.CONTAINER_FILENAME)
+        if os.path.exists(container_path):
+            with open(container_path) as f:
+                data = yaml.load(f)
+                if 'platforms' not in data:
+                    return self.platforms
+                for action, platforms in data['platforms'].items():
+                    if 'not' in action:
```
twaugh: Should this be `if action == 'not':`?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -171,22 +172,29 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+                        if isinstance(platforms, str):
+                            excluded_platforms.add(platforms)
+                        else:
+                            for x in platforms:
+                                excluded_platforms.add(x)
+                    elif 'only' in action:
```
twaugh: `elif action == 'only':`?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -185,22 +186,29 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+        excluded_platforms = set()
+        container_path = os.path.join(df_dir, self.CONTAINER_FILENAME)
+        if os.path.exists(container_path):
+            with open(container_path) as f:
+                data = yaml.load(f)
+                if 'platforms' not in data:
+                    return self.platforms
+                for action, platforms in data['platforms'].items():
+                    if action == 'not':
+                        if isinstance(platforms, str):
```
twaugh: I think it would be safer to test `isinstance(platforms, list)` because in Python 2 `unicode()` is not an instance of `str` and `yaml.load()` will create unicode strings if regular strings won't hold the characters. For example with this input:
```
platforms:
  not: x86⚬64
```

...which `yaml.load()` on Python 2 will return as:
```
{'platforms': {'not': u'x86\u26ac64'}}
```

So I think it would be better to use:
```
if isinstance(platforms, list):
    for platform in platforms:
        excluded_platforms.add(platform)
else:
    ...
```

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -185,22 +186,29 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
     def get_platforms(self):
-        return self.platforms - self.get_excluded_platforms()
+        df_dir = self.workflow.source.get_dockerfile_path()[1]
+        excluded_platforms = set()
+        container_path = os.path.join(df_dir, self.CONTAINER_FILENAME)
+        if os.path.exists(container_path):
+            with open(container_path) as f:
+                data = yaml.load(f)
+                if 'platforms' not in data:
+                    return self.platforms
```
twaugh: This case doesn't get covered by unit testing:
https://coveralls.io/builds/12327768/source?filename=atomic_reactor%2Fplugins%2Fbuild_orchestrate_build.py#L197

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -185,22 +186,29 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+                for action, platforms in data['platforms'].items():
+                    if action == 'not':
+                        if isinstance(platforms, str):
+                            excluded_platforms.add(platforms)
```
twaugh: This case doesn't get covered by unit testing:
https://coveralls.io/builds/12327768/source?filename=atomic_reactor%2Fplugins%2Fbuild_orchestrate_build.py#L201

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -185,22 +186,29 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+                        else:
+                            for x in platforms:
+                                excluded_platforms.add(x)
+                    elif action == 'only':
+                        if isinstance(platforms, str):
+                            self.platforms.add(platforms)
+                        else:
+                            for x in platforms:
+                                self.platforms.add(x)
```
twaugh: This case doesn't get covered by unit testing:
https://coveralls.io/builds/12327768/source?filename=atomic_reactor%2Fplugins%2Fbuild_orchestrate_build.py#L209

atomic_reactor/plugins/build_orchestrate_build.py
mlangsdorf: All the unit tests are hopefully fixed.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -185,22 +185,29 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+        df_dir = self.workflow.source.get_dockerfile_path()[1]
+        excluded_platforms = set()
+        container_path = os.path.join(df_dir, self.CONTAINER_FILENAME)
+        if os.path.exists(container_path):
+            with open(container_path) as f:
+                data = yaml.load(f)
+                if data is None or 'platforms' not in data or data['platforms'] is None:
+                    return self.platforms
+                for action, platforms in data['platforms'].items():
+                    if action == 'not':
```
lcarva: There's some duplication here when processing the platforms value. Consider adding a helper method to ensure that value is a list:
```python
def make_list(self, value):
    if not isinstance(value, list):
        value = [value]
    return value
```

I also think that looping over the actions is not really necessary since we know which keys we're looking for:

```python
excluded_platforms = set(make_list(data['platforms'].get('not', [])))
self.platforms += set(make_list(data['platforms'].get('only', [])))
```

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -185,22 +185,29 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+                        if isinstance(platforms, list):
+                            for x in platforms:
+                                excluded_platforms.add(x)
+                        else:
+                            excluded_platforms.add(platforms)
+                    elif action == 'only':
```
lcarva: Shouldn't this cause `self.platforms` to be cleared, and only add the items specified here?

tests/plugins/test_orchestrate_build.py
```diff
@@ -480,7 +481,16 @@ def test_orchestrate_build_choose_clusters(tmpdir, clusters_x86_64,
         assert plat_annotations['build']['cluster-url'] == 'https://chosen_{}.com/'.format(platform)
 
 
-def test_orchestrate_build_exclude_platforms(tmpdir):
+@pytest.mark.parametrize(('platforms', 'platform_exclude', 'platform_only'), [
```
lcarva: This test should probably specify a list of platforms that are expected to be set, instead of repeating logic inside the unit test.

atomic_reactor/plugins/build_orchestrate_build.py
twaugh: My fault, the docs say to do it this way. Docs fix: https://github.com/projectatomic/osbs-docs/pull/26

What we want is `self.platforms &= platforms` I think.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -185,22 +185,25 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+        container_path = os.path.join(df_dir, self.CONTAINER_FILENAME)
+        if os.path.exists(container_path):
+            with open(container_path) as f:
+                data = yaml.load(f)
+                if data is None or 'platforms' not in data or data['platforms'] is None:
+                    return self.platforms
+                excluded_platforms = set(self.make_list(data['platforms'].get('not', [])))
+                only_platforms = set(self.make_list(data['platforms'].get('only', [])))
+                if only_platforms:
+                    self.platforms = self.platforms & only_platforms
```
twaugh: Could use `self.platforms &= only_platforms` to do it in-place, but this is also fine.

mlangsdorf: I don't think you can. If only_platforms is () because no one put an only: field in the yaml, you want to use the default platforms, not no platforms. I thought not: and only: were optional.

mlangsdorf: I got a syntax error the first time I tried that and shied away in favor of something I knew would work.

twaugh:
Draft release notes updated.

PR #728, opened 2017-07-07
logging: Add 'arch' to all log output

dzickusrh:
To help with multiplexing orchestrator logs and worker build logs,
add an 'arch' component to the log output.

Orchestrator has no 'arch' and outputs a '-'.

This is done by using the LoggerAdapter for the worker builds.  Instead
of littering LoggerAdapter all of the git tree, just use a custom LogFormatter
and default arch to '-' for the orchestrator logs.

Output goes from:

2017-07-07 16:10:52,214 - atomic_reactor.plugins.orchestrate_build - INFO - platform x86_64 will use cluster worker_x86_64
2017-07-07 16:10:52,214 - atomic_reactor.plugins.orchestrate_build - INFO - x86_64 - created build worker-build-x86_64
2017-07-07 16:10:52,215 - atomic_reactor.plugins.orchestrate_build - INFO - <generator object <genexpr> at 0x7f8fbe0a6460>

To:

2017-07-07 16:10:52,214 platform:- - atomic_reactor.plugins.orchestrate_build - INFO - platform x86_64 will use cluster worker_x86_64
2017-07-07 16:10:52,214 platform:- - atomic_reactor.plugins.orchestrate_build - INFO - x86_64 - created build worker-build-x86_64
2017-07-07 16:10:52,215 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO - <generator object <genexpr> at 0x7f8fbe0a6460>

Unit test updated to parse the capture log and verify record.arch is set correctly.

This change only affects atomic_reactor project logging.  Other modules are not
impacted and will not have the 'arch' set (ie Docker output).

### DIFF
--- a/atomic_reactor/__init__.py
+++ b/atomic_reactor/__init__.py
@@ -14,7 +14,22 @@ import time
 
 from atomic_reactor.version import __version__
 
+try:
+    from osbs.constants import ATOMIC_REACTOR_LOGGING_FMT
+except ImportError:
+    # not available in osbs < 0.41
+    fmt = '%(asctime)s platform:%(arch)s - %(name)s - %(levelname)s - %(message)s'
+    ATOMIC_REACTOR_LOGGING_FMT = fmt
+
 start_time = time.time()
+
+
+class ArchFormatter(logging.Formatter):
+    def format(self, record):
+        if not hasattr(record, 'arch'):
+            record.arch = '-'
+
+        return super(ArchFormatter, self).format(record)
 
 
 def set_logging(name="atomic_reactor", level=logging.DEBUG, handler=None):
@@ -29,7 +44,7 @@ def set_logging(name="atomic_reactor", level=logging.DEBUG, handler=None):
         handler.setLevel(logging.DEBUG)
 
         # create formatter
-        formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+        formatter = ArchFormatter(ATOMIC_REACTOR_LOGGING_FMT)
 
         # add formatter to ch
         handler.setFormatter(formatter)
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -16,6 +16,7 @@ import os
 import random
 from string import ascii_letters
 import time
+import logging
 
 from atomic_reactor.build import BuildResult
 from atomic_reactor.plugin import BuildStepPlugin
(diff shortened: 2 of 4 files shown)


lcarva:
> I think we want self.log.info("%s", line) here, to avoid any issues with percent characters in line.

That's not needed:
```python
>>> import logging, sys
>>> logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
>>> logging.info('yolo')
INFO:root:yolo
>>> logging.info('yolo %s')
INFO:root:yolo %s
>>> logging.info('yolo %s', '!!')
INFO:root:yolo !!

```

twaugh:
I think I remember you showing me that before. I'll pay attention this time.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -66,13 +68,9 @@ def wait_to_finish(self):
...
 
-    def watch_logs(self, logger):
+    def watch_logs(self):
         for line in self.osbs.get_build_logs(self.name, follow=True):
-            # TODO: This is a little clunky:
-            # 2017-02-24 14:22:51,472 - atomic_reactor.plugins.orchestrate_build - INFO -
-            # x86_64 - 2017-02-24 14:22:51,314 - atomic_reactor.plugin - INFO -
-            # <worker build log message>
-            logger.info('%s - %s', self.platform, line)
+            self.log.info("%s", line)
```
twaugh: So actually this was fine the way you had it, sorry I asked you to change it. :-/

dzickusrh: Hmm, I jumped to quickly.  I made the changed and pushed it.  I can revert the change and re-push if you want.

twaugh: Yes, let's make it `self.log.info(line)` again.

atomic_reactor/__init__.py
```diff
@@ -29,7 +37,10 @@ def set_logging(name="atomic_reactor", level=logging.DEBUG, handler=None):
         handler.setLevel(logging.DEBUG)
 
         # create formatter
-        formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+        # NOTE: "platform:%(arch)s" is expected to be parsed, update parser in
+        # NOTE: osbs-client if this needs updating
+        fmt = '%(asctime)s platform:%(arch)s - %(name)s - %(levelname)s - %(message)s'
```
vrutkovs: Since it depends on osbs-client implementation it might be a good idea to move it to `atomic_reactor/constants.py`?

twaugh: Yes, I think that makes sense.

Could define
```
ATOMIC_REACTOR_LOGGING_FMT = '%(asctime)s platform:%(arch)s ...'
```
in osbs/constants.py and have atomic-reactor fetch it from there carefully:

```
try:
    from osbs.constants import ATOMIC_REACTOR_LOGGING_FMT
except ImportError:
    # not available in osbs < 0.41
    ATOMIC_REACTOR_LOGGING_FMT = '%(asctime)s platform:%(arch)s ...'
```

PR #729, opened 2017-07-12
orchestrate: reraise BuildCanceledException if its raised when fetching builds from worker cluster

vrutkovs:
This resolves the following bug:
* an orchestration build is started with one of the faulty worker clusters is started
* during 'get_current_builds' function the build gets cancelled
* http.py wraps atomic-reactor's BuildCanceledException in OsbsException
* osbs-client's 'get_current_builds' skips OsbsException and moves on to next cluster

The fix is to check the cause of OsbsException and reraise the exception if its 
BuildCanceledException

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -19,7 +19,7 @@ import time
 import logging
 
 from atomic_reactor.build import BuildResult
-from atomic_reactor.plugin import BuildStepPlugin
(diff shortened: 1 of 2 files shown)


vrutkovs:
Not sure if an additional test for this is needed

twaugh:
Would be nice to, so we have coverage over that line.

vrutkovs:
Test added, PTAL

twaugh:
Doesn't need updates to the draft release notes.

PR #730, opened 2017-07-13
Import primary image tags

bfontecc007:
*Added unit test for tag ensure
*Added code for ensuring primary image tags on import

### DIFF
--- a/atomic_reactor/plugins/post_import_image.py
+++ b/atomic_reactor/plugins/post_import_image.py
@@ -15,7 +15,7 @@ from osbs.conf import Configuration
 from osbs.exceptions import OsbsResponseException
 
 from atomic_reactor.plugin import PostBuildPlugin
-from atomic_reactor.util import get_build_json
+from atomic_reactor.util import get_build_json, ImageName
 
 
 class ImportImagePlugin(PostBuildPlugin):
@@ -77,9 +77,9 @@ class ImportImagePlugin(PostBuildPlugin):
                                   build_json_dir=self.build_json_dir,
                                   namespace=metadata.get('namespace', None))
         osbs = OSBS(osbs_conf, osbs_conf)
-
+        imagestream = None
         try:
-            osbs.get_image_stream(self.imagestream)
+            imagestream = osbs.get_image_stream(self.imagestream)
         except OsbsResponseException:
             if self.insecure_registry is not None:
                 kwargs['insecure_registry'] = self.insecure_registry
@@ -87,20 +87,41 @@ class ImportImagePlugin(PostBuildPlugin):
             self.log.info("Creating ImageStream %s for %s", self.imagestream,
                           self.docker_image_repo)
 
-            # Tags are imported automatically on creation
-            osbs.create_image_stream(self.imagestream, self.docker_image_repo,
-                                     **kwargs)
-        else:
-            self.log.info("Importing new tags for %s", self.imagestream)
+            imagestream = osbs.create_image_stream(self.imagestream,
+                                                   self.docker_image_repo,
+                                                   **kwargs)
+        self.log.info("Importing new tags for %s", self.imagestream)
 
-            attempts = 0
-            while not osbs.import_image(self.imagestream):
-                attempts += 1
+        primaries = None
+        try:
+            primaries = self.workflow.build_result.annotations['repositories']['primary']
+        except (TypeError, KeyError):
+            self.log.exception('Unable to read primary repositories annotations')
 
-                if attempts >= self.import_attempts:
-                    msg = "Failed to import new tags for %s"
-                    raise RuntimeError(msg % self.imagestream)
+        if not primaries:
+            raise RuntimeError('Could not find primary images in workflow')
 
-                self.log.info("no new tags, will retry after %d seconds (%d/%d)",
-                              self.retry_delay, attempts, self.import_attempts)
-                sleep(self.retry_delay)
+        failures = False
+        for s in primaries:
+            tag_image_name = ImageName.parse(s)
+            tag = tag_image_name.tag
+            try:
+                osbs.ensure_image_stream_tag(imagestream.json(), tag)
+                self.log.info("Imported ImageStreamTag: (%s)", tag)
+            except OsbsResponseException:
+                failures = True
+                self.log.info("Could not import ImageStreamTag: (%s)", tag)
+        if failures:
+            raise RuntimeError("Failed to import ImageStreamTag(s). Check logs")
+
+        attempts = 0
+        while not osbs.import_image(self.imagestream):
+            attempts += 1
+
+            if attempts >= self.import_attempts:
+                msg = "Failed to import new tags for %s"
+                raise RuntimeError(msg % self.imagestream)
+
+            self.log.info("no new tags, will retry after %d seconds (%d/%d)",
+                          self.retry_delay, attempts, self.import_attempts)
+            sleep(self.retry_delay)
--- a/tests/plugins/test_import_image.py
+++ b/tests/plugins/test_import_image.py
@@ -15,6 +15,7 @@ from atomic_reactor.inner import DockerBuildWorkflow
 from atomic_reactor.plugin import PostBuildPluginsRunner, PluginFailedException
 from atomic_reactor.util import ImageName
 from atomic_reactor.plugins.post_import_image import ImportImagePlugin
+from atomic_reactor.build import BuildResult
 
 import osbs.conf
 from osbs.api import OSBS
@@ -37,7 +38,16 @@ class X(object):
     base_image = ImageName(repo="qwe", tag="asd")
 
 
-def prepare(insecure_registry=None, retry_delay=0, import_attempts=3, namespace=None):
+class ImageStreamResponse:
+    '''
+    Mocks a get_image_stream response
+    '''
+    def __init__(self):
+        self.json = lambda: {'hello': 'howdy'}
+
+
+def prepare(insecure_registry=None, retry_delay=0, import_attempts=3,
+            namespace=None):
     """
     Boiler-plate test set-up
     """
@@ -50,6 +60,13 @@ def prepare(insecure_registry=None, retry_delay=0, import_attempts=3, namespace=
     setattr(workflow.builder, 'source', X())
     setattr(workflow.builder.source, 'dockerfile_path', None)
     setattr(workflow.builder.source, 'path', None)
+
+    a = {
+        'repositories': {
+            'primary': ['one', 'two', 'three', 'four', 'five', 'six']}}
+    build_result = BuildResult(annotations=a, image_id='foo')
+    setattr(workflow, 'build_result', build_result)
+
     fake_conf = osbs.conf.Configuration(conf_file=None, openshift_uri='/')
 
     expectation = flexmock(osbs.conf).should_receive('Configuration').and_return(fake_conf)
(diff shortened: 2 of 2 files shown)


atomic_reactor/plugins/post_import_image.py
```diff
@@ -77,9 +77,9 @@ def run(self):
                                   build_json_dir=self.build_json_dir,
                                   namespace=metadata.get('namespace', None))
         osbs = OSBS(osbs_conf, osbs_conf)
-
+        i = None
```
vrutkovs: It would be better to rename this variable to smth like 'imagestream', as its complicated to find all instances of 'i' in the code

atomic_reactor/plugins/post_import_image.py
```diff
@@ -88,19 +88,28 @@ def run(self):
...
-                              self.retry_delay, attempts, self.import_attempts)
-                sleep(self.retry_delay)
+            i = osbs.create_image_stream(self.imagestream, self.docker_image_repo,
+                                         **kwargs)
+        self.log.info("Importing new tags for %s", self.imagestream)
+
+        primaries = None
+        try:
+            primaries = self.workflow.build_result.annotations['repositories']['primary']
+        except:
```
vrutkovs: Catching all exceptions is dangerous (e.g. if the build is canceled, the exception would be thrown - and it would be ignored).

Should be `except KeyError` most likely

atomic_reactor/plugins/post_import_image.py
```diff
@@ -88,19 +88,28 @@ def run(self):
...
+            pass
+        if primaries:
+            for s in primaries:
+                tag_image_name = ImageName.parse(s)
+                osbs.ensure_image_stream_tag(i.json(), tag_image_name.tag)
```
vrutkovs: This throws at least `OsbsResponseException` - it probably should be trapped. @twaugh @lcarva what do you think?

vrutkovs: In order to get a list of all missing imagestream tags we should be catching `OsbsResponseException` and logging it. If some exceptions were thrown in the loop a `RuntimeError` should be thrown with a message `some imagestreams were not created, check the logs`. This would allow us to get a full list of missing imagestreams, not just the first one.

atomic_reactor/plugins/post_import_image.py
```diff
@@ -88,19 +88,40 @@ def run(self):
                           self.docker_image_repo)
 
             # Tags are imported automatically on creation
```
twaugh: Is this comment still correct?

atomic_reactor/plugins/post_import_image.py
```diff
@@ -88,19 +88,40 @@ def run(self):
...
+                                                   **kwargs)
+        self.log.info("Importing new tags for %s", self.imagestream)
+
+        primaries = None
+        try:
+            primaries = self.workflow.build_result.annotations['repositories']['primary']
+        except TypeError:
+            pass
+        except KeyError:
+            pass
```
twaugh: Worth collapsing this?:
```
except (TypeError, KeyError):
    pass
```

atomic_reactor/plugins/post_import_image.py
```diff
@@ -88,19 +88,40 @@ def run(self):
...
+        if primaries:
```
twaugh: If you initialise it to `[]` instead of `None`, can we remove this test?

atomic_reactor/plugins/post_import_image.py
```diff
@@ -88,19 +88,40 @@ def run(self):
...
+                tag_image_name = ImageName.parse(s)
+                tag = tag_image_name.tag
+                failures = False
+                try:
+                    osbs.ensure_image_stream_tag(imagestream.json(), tag)
+                    self.log.info("Imported ImageStreamTag: " + str(tag))
+                except OsbsResponseException:
+                    failures = True
+                    self.log.info("Could not import ImageStreamTag: " + str(tag))
+            if failures:
```
twaugh: If `primaries` is an empty list, this will not have been set.

atomic_reactor/plugins/post_import_image.py
```diff
@@ -88,19 +88,40 @@ def run(self):
...
+            primaries = self.workflow.build_result.annotations['repositories']['primary']
+        except TypeError:
+            pass
+        except KeyError:
+            pass
+        if primaries:
+            for s in primaries:
+                tag_image_name = ImageName.parse(s)
+                tag = tag_image_name.tag
+                failures = False
```
twaugh: Is this being initialised in the right place? Shouldn't it be outside the loop?

atomic_reactor/plugins/post_import_image.py
```diff
@@ -88,19 +88,40 @@ def run(self):
...
+                try:
+                    osbs.ensure_image_stream_tag(imagestream.json(), tag)
+                    self.log.info("Imported ImageStreamTag: " + str(tag))
```
twaugh: Tiny nitpick: might be nicer to use `self.log.info("Imported ImageStreamTag: %s", tag)`, just because then the log record has an `args`.

atomic_reactor/plugins/post_import_image.py
```diff
@@ -88,19 +88,40 @@ def run(self):
...
+        while not osbs.import_image(self.imagestream):
+            attempts += 1
+
+            if attempts >= self.import_attempts:
+                msg = "Failed to import new tags for %s"
+                raise RuntimeError(msg % self.imagestream)
+
+            self.log.info("no new tags, will retry after %d seconds (%d/%d)",
+                          self.retry_delay, attempts, self.import_attempts)
+            sleep(self.retry_delay)
```
twaugh: I sort of feel like the retry handling will be better performed by osbs.import_image(), and we have work tracked to add that there.

Is this getting hit a lot? Will it still work if the retry handling is removed?

atomic_reactor/plugins/post_import_image.py
```diff
@@ -88,19 +88,40 @@ def run(self):
...
-                sleep(self.retry_delay)
+            imagestream = osbs.create_image_stream(self.imagestream,
+                                                   self.docker_image_repo,
+                                                   **kwargs)
+        self.log.info("Importing new tags for %s", self.imagestream)
+
+        primaries = None
+        try:
+            primaries = self.workflow.build_result.annotations['repositories']['primary']
+        except TypeError:
```
vrutkovs: `except (TypeError, KeyError)` would do the same, but shorter

atomic_reactor/plugins/post_import_image.py
bfontecc007: This is probably a question for Luiz

atomic_reactor/plugins/post_import_image.py
bfontecc007: It might be misleading. I can take it out.

atomic_reactor/plugins/post_import_image.py
vrutkovs: This can easily happen if Openshift's load-balancer is unavailable at some point.

I think its worth keeping, at least until a proper retry strategy is implemented - currently the code is using `sleep` or `attempt` everywhere, so it would be easy to replace these parts later on

lcarva: Yes, I'd rather the retry logic was addressed separately from what these changes are trying to achieve.

atomic_reactor/plugins/post_import_image.py
```diff
@@ -77,30 +77,48 @@ def run(self):
...
+        except (TypeError, KeyError):
+            pass
+
+        failures = False
+        for s in primaries:
+            tag_image_name = ImageName.parse(s)
+            tag = tag_image_name.tag
+            try:
+                osbs.ensure_image_stream_tag(imagestream.json(), tag)
+                self.log.info("Imported ImageStreamTag: (%s)", str(tag))
```
lcarva: Why `str(tag)` ? `tag` should already be `str`. Even if it isn't, the logging module should perform the needed conversion.

atomic_reactor/plugins/post_import_image.py
```diff
@@ -77,30 +77,48 @@ def run(self):
...
+            except OsbsResponseException:
+                failures = True
+                self.log.info("Could not import ImageStreamTag: (%s)", str(tag))
```
lcarva: Should the be `error` level instead of `info`?

*Same comment regarding unnecessary usage of `str(tag)`

atomic_reactor/plugins/post_import_image.py
```diff
@@ -77,30 +77,48 @@ def run(self):
...
-                if attempts >= self.import_attempts:
-                    msg = "Failed to import new tags for %s"
-                    raise RuntimeError(msg % self.imagestream)
-
-                self.log.info("no new tags, will retry after %d seconds (%d/%d)",
-                              self.retry_delay, attempts, self.import_attempts)
-                sleep(self.retry_delay)
+        primaries = []
+        try:
+            primaries = self.workflow.build_result.annotations['repositories']['primary']
```
lcarva: In the case where there are no `primaries`, do we want to bail out immediately? The `for-loop` below will be skipped since `primaries` is an empty list, but `osbs.import_image` will still be called.

If the use case for no `primaries` is only for scratch builds, then this is not a valid use case since the plugin is disabled for scratch builds.
I'm wondering if we should fail the plugin if there are no `primaries`, as that's not expected.
If not, then we should probably make sure that `osbs.import_image` is not called if there are not primaries.

atomic_reactor/plugins/post_import_image.py
bfontecc007: I got an error when I tried that.

lcarva: Which error?

bfontecc007: Ok, the error was actually when I still had a '+' instead of passing it as another arg. I'll take out str now.

atomic_reactor/plugins/post_import_image.py
```diff
@@ -77,30 +77,52 @@ def run(self):
...
-                    raise RuntimeError(msg % self.imagestream)
-
-                self.log.info("no new tags, will retry after %d seconds (%d/%d)",
-                              self.retry_delay, attempts, self.import_attempts)
-                sleep(self.retry_delay)
+        primaries = None
+        try:
+            primaries = self.workflow.build_result.annotations['repositories']['primary']
+        except (TypeError, KeyError):
+            pass
```
lcarva: Probably don't need to initialize `primaries` to None in this case. And do something like this in the except block:
```python
except (TypeError, KeyError):
   msg = 'Could not find primary images in workflow'
   self.log.exception(msg)
   raise RuntimeError(msg)
```

This has the benefit of logging the originating stack trace for easier debugging, while providing a meaningful error message for better usability.

bfontecc007: I thought about that but I was thinking it could be there but just be an empty list or be None.

lcarva: aha! yes, you're right.
Still would be beneficial to log the exception though.

atomic_reactor/plugins/post_import_image.py
```diff
@@ -77,30 +77,51 @@ def run(self):
...
-                    raise RuntimeError(msg % self.imagestream)
-
-                self.log.info("no new tags, will retry after %d seconds (%d/%d)",
-                              self.retry_delay, attempts, self.import_attempts)
-                sleep(self.retry_delay)
+        primaries = None
+        try:
+            primaries = self.workflow.build_result.annotations['repositories']['primary']
+        except (TypeError, KeyError):
+            raise RuntimeError('Could not find primary images in workflow (1)')
```
lcarva: What about logging the original stack trace and not re-raising any exceptions here? The if-block should take care of checking if there are any primaries. Something like:

```python
...
except (TypeError, KeyError):
  # This actually logs the full stack trace, helpful in debugging.
   self.log.exception('Unable to read primary repositories annotations')

if not primaries:
   ...

bfontecc007: Sure, we can do that

lcarva:
Release notes updated.

PR #731, opened 2017-07-13
Sort out log encoding

dzickusrh:
Keep the log output in a consistent output format by forcing logging to
use UTF-8.  Forcing stderr to accomplish this wasn't straightforward, so
a new streamHandler was created to create the effect we want.  And an
appropriate test case to verify.

Based on Tim Waugh's original work.

I tweaked the test case to handle the different names for 'UTF8' and
added a check to not encode byte data for writer().

### DIFF
--- a/atomic_reactor/__init__.py
+++ b/atomic_reactor/__init__.py
@@ -9,7 +9,10 @@ of the BSD license. See the LICENSE file for details.
 constants
 """
 
+from locale import nl_langinfo, CODESET
 import logging
+from os import fdopen, dup
+import sys
 import time
 
 from atomic_reactor.version import __version__
@@ -32,6 +35,33 @@ class ArchFormatter(logging.Formatter):
         return super(ArchFormatter, self).format(record)
 
 
+class EncodedStream(object):
+    # The point of this class is to force python to enocde UTF-8
+    # over stderr.  Normal techniques were not working, so we dup
+    # the file handler and force it UTF-8.  :-(
+    # Any attempts to fix should run test_cli.py::TestCLISuite::test_log_encoding
+    # to verify.
+    def __init__(self, fileno, encoding):
+        self.binarystream = fdopen(dup(fileno), 'wb')
+        self.encoding = encoding
+
+    def write(self, text):
+        if not isinstance(text, bytes):
+            self.binarystream.write(text.encode(self.encoding))
+        else:
+            self.binarystream.write(text)
+
+    def __del__(self):
+        try:
+            self.binarystream.close()
+        except AttributeError:
+            pass
+
+
+def get_logging_encoding(name="atomic_reactor"):
+    return logging.getLogger(name).handlers[0].stream.encoding
+
+
 def set_logging(name="atomic_reactor", level=logging.DEBUG, handler=None):
     # create logger
     logger = logging.getLogger(name)
@@ -42,7 +72,9 @@ def set_logging(name="atomic_reactor", level=logging.DEBUG, handler=None):
 
     if not handler:
         # create console handler and set level to debug
-        handler = logging.StreamHandler()
+        log_encoding = nl_langinfo(CODESET)
+        encoded_stream = EncodedStream(sys.stderr.fileno(), log_encoding)
+        handler = logging.StreamHandler(encoded_stream)
         handler.setLevel(logging.DEBUG)
 
         # create formatter
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -11,11 +11,15 @@ from __future__ import print_function, unicode_literals
 import logging
 import os
 import sys
+import codecs
+import encodings
 
 import pytest
+import flexmock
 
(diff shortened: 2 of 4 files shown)


twaugh:
"53" is coming from performing that split on this line:
https://travis-ci.org/projectatomic/atomic-reactor/jobs/253288475#L2021

It's a warning, emitted via logging, that we didn't close a file.

I think https://github.com/projectatomic/atomic-reactor/pull/732 might fix it.

twaugh:
Please re-base against master and let's see what Travis CI thinks of it now.

twaugh:
Hmm, there's still another of these warnings. Investigating.

twaugh:
See utf8-logs branch.

atomic_reactor/__init__.py
```diff
@@ -32,6 +35,28 @@ def format(self, record):
...
+            self.binarystream.write(text)
+
+    def __del__(self):
+        try:
+            self.binarystream.close()
+        except AttributeError:
+            pass
+
+
+def get_encoding(name=None):
```
twaugh: Wonder if it might be nice to use the same default value set_logging() uses here.

lcarva: Should this be `get_logging_encoding` instead?

atomic_reactor/__init__.py
```diff
@@ -32,6 +35,28 @@ def format(self, record):
         return super(ArchFormatter, self).format(record)
 
 
+class EncodedStream(object):
+    def __init__(self, fileno, encoding):
+        self.binarystream = fdopen(dup(fileno), 'wb')
```
lcarva: There's probably a very good reason why the file descriptor is being duplicated here, but won't this cause overlapping writing issues?

atomic_reactor/__init__.py
dzickusrh: I can use that default.  This was left over from my playing with root logging.

atomic_reactor/__init__.py
dzickusrh: Tim and I couldn't figure out how to force stderr to encode UTF-8 for output.  Tim stumbled upon this trick.  Suggestions welcome? :-)
Not sure what would be the symptom of an overlapping write issues.  My log output seems fine.

lcarva: Unfortunately I have no suggestions, only questions :(

I ask because the encoding is used when writing to the file, not when opening the file.
So in theory, the same file descriptor should work.

In any case, it would be nice to have a warning comment in there explaining the intent.
This will, at least attempt to, prevent someone, like me, breaking things in the future.

atomic_reactor/__init__.py
```diff
@@ -32,6 +35,33 @@ def format(self, record):
         return super(ArchFormatter, self).format(record)
 
 
+class EncodedStream(object):
+    # The point of this class is to force pyton to enocde UTF-8
+    # over stderr.  Normal techniques were not working, so we dup
+    # the file handler and force it UTF-8.  :-(
+    # Any attempts to fix should run test_cli.py::test_log_encoding to
```
lcarva: `test_cli.py::TestCLISuite::test_log_encoding`

atomic_reactor/__init__.py
```diff
@@ -32,6 +35,33 @@ def format(self, record):
         return super(ArchFormatter, self).format(record)
 
 
+class EncodedStream(object):
+    # The point of this class is to force pyton to enocde UTF-8
```
lcarva: Typo: 'pyton'

tests/test_cli.py
```diff
@@ -173,3 +177,28 @@ def test_create_build_image(self, temp_image_name):  # noqa
...
+            .and_raise(RuntimeError))
+
+        command = [
+            "main.py",
+            "--verbose",
+            "inside-build",
+        ]
+        with caplog.atLevel(logging.INFO):
+            with pytest.raises(RuntimeError):
+                self.exec_cli(command)
```
lcarva: @dzickusrh has this been addressed?

atomic_reactor/__init__.py
lcarva: Removing the file descriptor duplication bit causes this error: `ValueError: I/O operation on closed file`. That looks like a good reason to keep this. Thank you for adding the big fat warning comment.

tests/test_cli.py
dzickusrh: Yeah, Tim added that last week and I pulled and pushed it out on Friday.  Those lines are easily seen in test_cli.py.  Did I miss something?

lcarva: Nope. Just wanted to confirm.

lcarva:
Updated release notes.

atomic_reactor/__init__.py
twaugh: We dup() the file descriptor here because the `file` returned by `os.fdopen()` will call `close()` on its file descriptor when it is deleted -- and so will `sys.stderr`.

PR #735, opened 2017-07-14
different equal labels value, wont fail the build

rcerven:


atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -194,64 +194,35 @@ def add_as_an_alias(set_to, set_from):
...
-                    self.labels[set_label] = value_from[list_labels_set[0]]
+                    self.labels[set_label] = value_from[labels_found[0]]
                     self.log.warning("adding equal label %r with value %r",
-                                     set_label, value_from[list_labels_set[0]])
+                                     set_label, value_from[labels_found[0]])
 
-        fail_build = False
         for equal_list in self.equal_labels:
             all_equal = set(equal_list)
-            found_labels_base = set()
```
vrutkovs: Why would we not want `set` here?

rcerven: because in set_missing_labels it is converted to set for comparison, but as list still can be used for indexing

tests/plugins/test_add_labels.py
```diff
@@ -417,32 +417,31 @@ def test_add_labels_equal_aliases(tmpdir, docker_tasker, caplog,
...
-    (None,  None,  None,  None,  None,  None,  None,  None),
-    (None,  None,  None,  None,  None,  'A',   'A',   'adding equal label'),
-    ('A',   'B',   'B',   None,  None,  None,  None,  RuntimeError()),
-    (None,  None,  None,  'A',   'B',   'B',   None,  RuntimeError()),
-    ('A',   'A',   'A',   None,  None,  None,  'A',   None),
-    ('A',   None,  'A',   'A',   None,  'A',   'A',   'adding equal label'),
-    ('A',   None,  None,  None,  'A',   'A',   'A',   'skipping label'),
-    (None,  'A',   'A',   'A',   'A',   None,  'A',   'skipping label'),
+    (None,  None,  None,  None,  None,  None,  (None, None, None),  None),
+    (None,  None,  None,  None,  None,  'A',   ('A', 'A', 'A'),     'adding equal label'),
```
twaugh: Putting the base and df parts into tuples as well would help readability a lot here.

tests/plugins/test_add_labels.py
```diff
@@ -417,32 +417,31 @@ def test_add_labels_equal_aliases(tmpdir, docker_tasker, caplog,
...
+    ('A',   'B',   'B',   None,  None,  None,  ('A', 'B', 'B'),     None),
+    (None,  None,  None,  'A',   'B',   'B',   ('A', 'B', 'B'),     None),
+    ('A',   'A',   'A',   None,  None,  None,  ('A', 'A', 'A'),     None),
+    ('A',   None,  'A',   'A',   None,  'A',   ('A', 'A', 'A'),     'adding equal label'),
+    ('A',   None,  None,  None,  'A',   'A',   ('A', 'A', 'A'),     'skipping label'),
+    (None,  'A',   'A',   'A',   'A',   None,  ('A', 'A', 'A'),     'skipping label'),
+    ('A',   'A',   'A',   'B',   'C',   None,  ('B', 'C', 'B'),     'adding equal label'),
```
twaugh: Is the rule for deciding which value to use here described in words anywhere?

rcerven: no it isn't, not sure where to put it

atomic_reactor/plugins/pre_add_labels_in_df.py
rcerven: and the main point was to keep ordered list, so we can always use the 1st equal label to copy from to others

tests/plugins/test_add_labels.py
twaugh: The module docstring at the top of the plugin module would do.

rcerven:
base and df are groups as well now

tests/plugins/test_add_labels.py
rcerven: ok got it

twaugh:
Draft release notes updated.

PR #736, opened 2017-07-17
Fix default poll timeout for koji_parent

lcarva:


lcarva:
Not updating release notes because this fix affects a plugin that has not yet been released.

PR #739, opened 2017-07-18
koji_import: adjust exit_sendmail and exit_koji_tag_build

mlangsdorf:
make them aware of koji_import and prefer to use it over
koji_promote.

### DIFF
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -16,6 +16,7 @@ except ImportError:
 
 from atomic_reactor.plugin import ExitPlugin, PluginFailedException
 from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
+from atomic_reactor.plugins.exit_koji_import import KojiImportPlugin
 from atomic_reactor.plugins.exit_koji_promote import KojiPromotePlugin
 from atomic_reactor.koji_util import create_koji_session
 from atomic_reactor.util import get_build_json
@@ -123,11 +124,13 @@ class SendMailPlugin(ExitPlugin):
         else:
             self.log.info("Koji task ID: %s", self.koji_task_id)
 
-        try:
+        self.koji_build_id = self.workflow.exit_results.get(KojiImportPlugin.key)
+        if not self.koji_build_id:
             self.koji_build_id = self.workflow.exit_results.get(KojiPromotePlugin.key)
-        except Exception:
-            self.log.exception("Failed to fetch koji build ID")
-            self.koji_build_id = None
+            if not self.koji_build_id:
+                self.log.info("Failed to fetch koji build ID")
+            else:
+                self.log.info("Koji build ID: %s", self.koji_build_id)
         else:
             self.log.info("Koji build ID: %s", self.koji_build_id)
 
--- a/tests/plugins/test_sendmail.py
+++ b/tests/plugins/test_sendmail.py
@@ -26,6 +26,7 @@ except ImportError:
 from atomic_reactor.plugin import PluginFailedException
 from atomic_reactor.plugins.pre_check_and_set_rebuild import CheckAndSetRebuildPlugin
 from atomic_reactor.plugins.exit_sendmail import SendMailPlugin
+from atomic_reactor.plugins.exit_koji_import import KojiImportPlugin
 from atomic_reactor.plugins.exit_koji_promote import KojiPromotePlugin
 from atomic_reactor import util
 from smtplib import SMTPException
@@ -114,7 +115,10 @@ class MockedPathInfo(object):
 
 class TestSendMailPlugin(object):
     def test_fails_with_unknown_states(self):
-        p = SendMailPlugin(None, None,
+        class WF(object):
+            exit_results = {}
+
+        p = SendMailPlugin(None, WF(),
                            smtp_host='smtp.bar.com', from_address='foo@bar.com',
                            send_on=['unknown_state', MS])
         with pytest.raises(PluginFailedException) as e:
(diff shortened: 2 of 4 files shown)


atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -123,13 +124,19 @@ def __init__(self, tasker, workflow,
...
-        try:
-            self.koji_build_id = self.workflow.exit_results.get(KojiPromotePlugin.key)
-        except Exception:
+        valid_id = False
+        for key in [KojiImportPlugin.key, KojiPromotePlugin.key]:
+            try:
+                self.koji_build_id = self.workflow.exit_results.get(key)
+            except Exception:
+                continue
+            else:
```
lcarva: Unnecessary `else`. If there's an exception it's going to skip over next item.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -123,13 +124,19 @@ def __init__(self, tasker, workflow,
         else:
             self.log.info("Koji task ID: %s", self.koji_task_id)
 
-        try:
-            self.koji_build_id = self.workflow.exit_results.get(KojiPromotePlugin.key)
-        except Exception:
+        valid_id = False
+        for key in [KojiImportPlugin.key, KojiPromotePlugin.key]:
```
lcarva: I'm curious as to why we're using two different approaches for performing the same logical operation. Why catch exceptions here, and not in `koji_tag` ?

tests/plugins/test_sendmail.py
```diff
@@ -193,6 +194,7 @@ class WF(object):
             build_canceled = manual_cancel
             tag_conf = TagConf()
             exit_results = {
+                KojiImportPlugin.key: MOCK_KOJI_BUILD_ID,
```
lcarva: Why do we need both here?

atomic_reactor/plugins/exit_sendmail.py
mlangsdorf: original code was written that way. sendmail used a try/catch pair while koji_tag just did an assignment.

tests/plugins/test_sendmail.py
mlangsdorf: bad logic. fixed.

atomic_reactor/plugins/exit_sendmail.py
mlangsdorf: fixed.

atomic_reactor/plugins/exit_sendmail.py
lcarva: I'm trying to understand why this line was even wrapped in a try-except block. I think it's because some of the unit tests for this plugin are using a mocked version of `DockerBuildWorkflow` which does not always have an `exit_results` attribute. This, of course, is not a valid scenario because `DockerBuildWorkflow` always has this attribute (https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/inner.py#L273).

I propose removing this try-except block. Some of the tests for this plugin may fail. To fix, simply set `exist_results = {}` in the `WF` class for the corresponding tests. Thoughts?

vrutkovs: +1 to removing `try..catch`

mlangsdorf: fixed. Had to add WF() to all the tests that didn't have it.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -123,11 +124,13 @@ def __init__(self, tasker, workflow,
...
 
-        try:
+        self.koji_build_id = self.workflow.exit_results.get(KojiImportPlugin.key)
+        if self.koji_build_id is None:
             self.koji_build_id = self.workflow.exit_results.get(KojiPromotePlugin.key)
-        except Exception:
-            self.log.exception("Failed to fetch koji build ID")
-            self.koji_build_id = None
+            if self.koji_build_id is None:
+                self.log.exception("Failed to fetch koji build ID")
```
lcarva: I think you want `self.log.error` here. `exception` is used to automatically log the latest stack trace. Since there are not exceptions that triggered this code path, it may actually log an exception that is completed unrelated.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -123,11 +124,13 @@ def __init__(self, tasker, workflow,
         else:
             self.log.info("Koji task ID: %s", self.koji_task_id)
 
-        try:
+        self.koji_build_id = self.workflow.exit_results.get(KojiImportPlugin.key)
+        if self.koji_build_id is None:
```
lcarva: Let's keep these checks similar to avoid unexpected behavior. Either `koji_tag` performs `if build_id is None` check, or this plugin performs `if not self.koji_build_id`.

Checking for falsy value (`if not foo`) actually checks for things like empty string and None. If, for some reason, `koji_import` returned an empty string, `koji_tag` would attempt to get value from `koji_promote`. But `exit_sendmail` would be perfectly happy with an empty string value.

mlangsdorf: I think i want if not, because there's a possibility that self.workflow.exit_results.get(KojiImportPlugin.key) might not return anything because arrangement{1,2} calls don't even have a koji_import plugin.

The logic should be get a result from koji_import which might not be valid. if it isn't valid, check koji_promote. Only if both give an invalid result should we believe that there's no koji_build_id.

atomic_reactor/plugins/exit_sendmail.py
mlangsdorf: is that really an error at that point? You said that a null return is legitimate. I'm going with info.

lcarva: Info sounds good.

atomic_reactor/plugins/exit_sendmail.py
lcarva: Agree. `if not` seems to be the right approach for this.

lcarva:
Updated release notes.

PR #740, opened 2017-07-19
Registry fixes

owtaylor:
This PR has fixes for two small corner cases when communicating with the registry:
 * Handling falling back to http for insecure registries
 * What happens if a v2 manifest can't be converted into a v1 manifest

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -15,6 +15,7 @@ import os
 import re
 from pipes import quote
 import requests
+from requests.exceptions import ConnectionError, SSLError
 import shutil
 import subprocess
 import tempfile
@@ -667,22 +668,39 @@ def query_registry(image, registry, digest=None, insecure=False, dockercfg_path=
         if username and password:
             auth = requests.auth.HTTPBasicAuth(username, password)
 
+    # In the insecure case, if the registry is just a hostname:port, we don't
+    # know whether to talk HTTPS or HTTP to it, so try both ways
     if not re.match('http(s)?://', registry):
-        registry = 'https://{}'.format(registry)
+        if insecure:
+            registries = ('https://{}'.format(registry), 'http://{}'.format(registry))
+        else:
+            registries = ('https://{}'.format(registry),)
+    else:
+        registries = (registry,)
 
     context = '/'.join([x for x in [image.namespace, image.repo] if x])
     reference = digest or image.tag or 'latest'
     object_type = 'manifests'
     if is_blob:
         object_type = 'blobs'
-    url = '{}/v2/{}/{}/{}'.format(registry, context, object_type, reference)
-    logger.debug("url: {}".format(url))
 
     headers = {'Accept': (get_manifest_media_type(version))}
     kwargs = {'verify': not insecure, 'headers': headers, 'auth': auth}
 
-    response = requests.get(url, **kwargs)
-    response.raise_for_status()
+    for idx, r in enumerate(registries):
+        url = '{}/v2/{}/{}/{}'.format(r, context, object_type, reference)
+        logger.debug("url: {}".format(url))
+
+        try:
+            response = requests.get(url, **kwargs)
+            response.raise_for_status()
+            break
+        except (ConnectionError, SSLError):
+            # If there are no more registry URLs to try, let the exception
+            # propagate, otherwise we'll try again
+            if idx == len(registries) - 1:
+                raise
+
     return response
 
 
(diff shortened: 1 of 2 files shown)


owtaylor:
Let me know if you'd like this split into two PR's.

tests/test_util.py
```diff
@@ -365,21 +377,29 @@ def test_get_manifest_digests_missing(tmpdir, v1_digest, v2_digest):
...
-            raise ValueError('Unexpected media type {}'.format(media_type))
+            if media_type.endswith('v2+json'):
+                digest = 'v2-digest'
+            else:
+                if not can_convert_v2_v1:
+                    # In some cases it's not possible to do the V2 => V1 conversion. For example,
+                    # if the V2 image doesn't have the legacy history information. This is the
+                    # response that the docker registry gives in that case
+                    headers = {}
+                    return (400, headers, '{"errors":[{"code":"MANIFEST_INVALID","message":"manifest invalid","detail":{}}]}')
```
owtaylor: Hmm https://github.com/projectatomic/atomic-reactor/blob/master/CONTRIBUTING.md shouldn't recommend running pyflakes and pep8 only on the atomic_reactor subdirectory ;-) Will fix these and update the PR

atomic_reactor/util.py
```diff
@@ -667,22 +668,38 @@ def query_registry(image, registry, digest=None, insecure=False, dockercfg_path=
...
     kwargs = {'verify': not insecure, 'headers': headers, 'auth': auth}
 
-    response = requests.get(url, **kwargs)
-    response.raise_for_status()
+    for r in registries:
+        url = '{}/v2/{}/{}/{}'.format(r, context, object_type, reference)
+        logger.debug("url: {}".format(url))
+
+        try_again = False
+        try:
```
lcarva: An error should probably be raised if all the requests raise `ConnectionError`.

atomic_reactor/util.py
```diff
@@ -708,6 +725,16 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
             insecure=insecure, dockercfg_path=dockercfg_path,
             version=version)
 
+        # If the registry has a v2 manifest that can't be converted into a v1
+        # manifest, the registry fails with status=400, and a error code of
+        # MANIFEST_INVALID.
+        if response.status_code == 400:
```
lcarva: Does this happen only when requesting version v1? If so, maybe also check if `version == 'v1'` ?

atomic_reactor/util.py
owtaylor: Yeah. If I do a fixup patch for that, is it preferred style to provide it as separate commit, or to push a new version of the patches with the fixes squashed?

atomic_reactor/util.py
owtaylor: The http standard gives two options when the Accept header cannot be honored: return '406 - Not Acceptable' or ignore the Accept header. So considering that, checking for v1 isn't unreasonable, and might avoid hiding actual client errors (400 is supposed to be when the request is in error.)

The question of what status codes can be returned in similar cases is complicated:
 * If conversion fails because data on the server has invalid references (that should have been caught at upload), then 400 is also returned. 500 can be returned in other cases of conversion failure.
 * If there is a manifest list on the server, but that doesn't contain any manifest for the default arch+os, then 404 is returned.
 * In the draft patch for OCI support, then requesting a docker manifest when only an OCI manifest is available or vice-versa also returns a 404.

The most conservative thing in some ways would to be to treat a 404 or a 406 as a not-log-and-continue, and to treat any other error status as a log-and-continue.

atomic_reactor/util.py
vrutkovs: I think its okay to push a new version, as we prefer to squash PRs into one commit

atomic_reactor/util.py
lcarva: Right, so in this case, we want to handle `400` differently because we're assuming that the v2 content cannot be converted to v1. To minimize the potential of hiding some other issue, it seems to me we should isolate this check to only apply when v1 content was requested.

owtaylor: I added the simple check for 'v1' now to decrease the chance of false positives. If there upstream situation with respect to Accept handling is clarified (https://github.com/docker/distribution/pull/2076#discussion_r128424684) , I'll revisit this - I'll have to revisit it anyways for OCI support.

vrutkovs:
Release notes updated

PR #741, opened 2017-07-20
plugins: Support orchestrator node in delete_from_registry plugin

dzickusrh:
The orchestrator node needs to process some of the digests before they
are deleted.  As a result, the delete_from_registry plugin needs to
move from the worker node to the orchestrator node.

This requires detecting which node the plugin is running from and using
the correct data lookup to find the registry information.

This plugin should still function correctly for older arrangements.

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -217,6 +217,9 @@ class PushConf(object):
         for registry_uri in registry_uris:
             self.add_docker_registry(registry_uri, insecure=insecure)
 
+    def remove_docker_registry(self, docker_registry):
+        self._registries["docker"].remove(docker_registry)
+
     def add_pulp_registry(self, name, crane_uri, server_side_sync=True):
         if crane_uri is None:
             raise RuntimeError("registry URI cannot be None")
--- a/atomic_reactor/plugins/exit_delete_from_registry.py
+++ b/atomic_reactor/plugins/exit_delete_from_registry.py
@@ -42,8 +42,136 @@ class DeleteFromRegistryPlugin(ExitPlugin):
 
         self.registries = deepcopy(registries)
 
+    def setup_secret(self, registry, secret_path):
+        auth = None
+
+        if secret_path:
+            self.log.debug("registry %s secret %s", registry, secret_path)
+            dockercfg = Dockercfg(secret_path).get_credentials(registry)
+            try:
+                username = dockercfg['username']
+                password = dockercfg['password']
+            except KeyError:
+                self.log.error("credentials for registry %s not found in %s",
+                               registry, secret_path)
+            else:
+                self.log.debug("found user %s for registry %s", username, registry)
+                auth = requests.auth.HTTPBasicAuth(username, password)
+
+        return auth
+
+    def request_delete(self, url, manifest, insecure, auth):
+        response = requests.delete(url, verify=not insecure, auth=auth)
+
+        if response.ok:
+            self.log.info("deleted manifest %s", manifest)
+            return True
+        elif response.status_code == requests.codes.NOT_FOUND:
+            self.log.warning("cannot delete %s: not found", manifest)
+        elif response.status_code == requests.codes.METHOD_NOT_ALLOWED:
+            self.log.warning("cannot delete %s: image deletion disabled on registry",
+                             manifest)
+        else:
+            msg = "failed to delete %s: %s" % (manifest, response.reason)
+            self.log.error("%s\n%s", msg, response.text)
+            raise PluginFailedException(msg)
+
+        return False
+
+    def make_manifest(self, registry, repo, digest):
+        return "{registry}/{repo}@{digest}".format(**vars())
+
+    def make_url(self, registry, repo, digest):
+        return "{registry}/v2/{repo}/manifests/{digest}".format(**vars())
+
+    def make_registry_noschema(self, registry):
+        return urlparse(registry).netloc
+
+    def find_registry(self, registry_noschema, workflow):
+        for push_conf_registry in workflow.push_conf.docker_registries:
+            if push_conf_registry.uri == registry_noschema:
+                return push_conf_registry
+
+        return None
+
+    def handle_registry(self, registry, push_conf_registry, auth, deleted_digests):
+        registry_noschema = self.make_registry_noschema(registry)
+        deleted = False
+
+        for tag, digests in push_conf_registry.digests.items():
+            digest = digests.default
+            if digest in deleted_digests:
+                # Manifest schema version 2 uses the same digest
+                # for all tags
+                self.log.info('digest already deleted %s', digest)
+                deleted = True
+                continue
+
+            repo = tag.split(':')[0]
+            url = self.make_url(registry, repo, digest)
+            manifest = self.make_manifest(registry_noschema, repo, digest)
+
+            # override insecure if passed
+            insecure = push_conf_registry.insecure
+
+            if self.request_delete(url, manifest, insecure, auth):
+                deleted_digests.add(digest)
+                deleted = True
+
+        return deleted
+
+    def get_worker_digests(self):
+        """
+         If we are being called from an orchestrator build, collect the worker
+         node data and recreate the data locally.
+        """
+        try:
+            builds = self.workflow.build_result.annotations['worker-builds']
+        except(TypeError, KeyError):
+            # This annotation is only set for the orchestrator build.
+            # It's not present, so this is a worker build.
+            return {}
+
+        worker_digests = {}
+
+        for plat, annotation in builds.items():
+            digests = annotation['digests']
+            self.log.debug("build %s has digests: %s", plat, digests)
+
+            for digest in digests:
+                reg = digest['registry']
+                worker_digests.setdefault(reg, [])
+                worker_digests[reg].append(digest)
+
+        return worker_digests
+
+    def handle_worker_digests(self, worker_digests, registry, insecure, auth, deleted_digests):
+        registry_noschema = self.make_registry_noschema(registry)
+
+        if registry_noschema not in worker_digests:
+            return False
+
+        digests = worker_digests[registry_noschema]
+        for digest in digests:
+            if digest['digest'] in deleted_digests:
+                # Manifest schema version 2 uses the same digest
+                # for all tags
+                self.log.info('digest already deleted %s', digest['digest'])
+                return True
+
+            url = self.make_url(registry, digest['repository'], digest['digest'])
+            manifest = self.make_manifest(registry_noschema, digest['repository'],
+                                          digest['digest'])
+
+            if self.request_delete(url, manifest, insecure, auth):
+                deleted_digests.add(digest['digest'])
+
+        return True
+
     def run(self):
         deleted_digests = set()
+
+        worker_digests = self.get_worker_digests()
 
         for registry, registry_conf in self.registries.items():
             if not registry.startswith('http://') and not registry.startswith('https://'):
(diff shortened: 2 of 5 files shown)


atomic_reactor/plugins/exit_delete_from_registry.py
```diff
@@ -42,6 +44,39 @@ def __init__(self, tasker, workflow, registries):
...
+
+            conf = {}
+            builds = self.workflow.build_result.annotations['worker-builds']
+
+            for plat, annotation in builds.items():
+                digests = json.loads(annotation['digests'])
+
+                for digest in digests:
+                    if digest['registry'] == registry:
+                        conf[digest['tag']] = ManifestDigest(v2=digest['digest'])
```
lcarva: What if we build the list of docker_registries once?
```python
if are_plugins_in_order(self.workflow.buildstep_plugins_conf, 'orchestrate_build'):
   ...
   for plat, annotation in builds.items():
      for digest in digests:
         ...
         self.docker_registries.append(DockerRegistry(registry))
else:
   ...
   self.docker_registries = self.workflow.push_conf.docker_registries
```

Then the find docker registry procedure is more straightforward as you can simply look in self.docker_registries?

dzickusrh: I started thinking down that road and got distracted.  But yeah I agree with your thought process there. :-)  Let me poke at it.

dzickusrh:
I haven't tested in an integrated environment yet.  I will have to hack up the osbs-client (which needs to get done anyway).

dzickusrh:
This commit has debug log stuff in it for now.  I have submitted it for an initial review as I can't imagine an ssl error changing the overall solution that much.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -84,6 +84,8 @@ def get_annotations(self):
             },
             'digests': json.loads(
                 build_annotations.get('digests', '[]')),
+            'all-digests': json.loads(
```
lcarva: What's the purpose of this param?

atomic_reactor/plugins/exit_delete_from_registry.py
```diff
@@ -42,9 +42,38 @@ def __init__(self, tasker, workflow, registries):
...
+                digests = annotation['all-digests']
+
+                for digest in digests:
+                    registry = None
+                    for reg in self.workflow.push_conf.docker_registries:
+                        if reg.uri == digest['registry']:
+                            registry = reg
+
+                    if not registry:
+                        registry = self.workflow.push_conf.add_docker_registry(digest['registry'])
```
lcarva: I don't think we want to alter the `push_conf` in this plugin.

atomic_reactor/plugins/exit_delete_from_registry.py
```diff
@@ -42,9 +42,38 @@ def __init__(self, tasker, workflow, registries):
...
+
+                    registry.digests[digest['tag']] = ManifestDigest(v2=digest['digest'])
+
     def run(self):
         deleted_digests = set()
 
+        self.check_for_orchestrator_build()
```
lcarva: How about having an attribute called: `self.docker_registries`.
If it's in orchestrator mode, it must be populated from workers metadata, otherwise it's just a reference to `self.workflow.push_conf.docker_registries` ?

The the main code in this plugin can iterate over `self.docker_registries`.

atomic_reactor/plugins/exit_delete_from_registry.py
twaugh: I thought that too at first, but actually it makes sense. In the orchestrator build, push_conf will only have the spec for the manifest list (or image manifest) it creates in the registry. By changing this here, we'll add to that the specs for the image manifests the worker builds created.

Then they are all deleted from the registry together.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -82,8 +82,7 @@ def _get_registries(self):
         Return a list of registries that are exposed to user
         If pulp registries are used, only those will be returned
         """
-        return (self.workflow.push_conf.pulp_registries or
-                self.workflow.push_conf.all_registries)
+        return (self.workflow.push_conf.all_registries)
```
twaugh: Method's docstring is wrong. "Return a list of all the registries this build updated"?
Unnecessary parentheses.

atomic_reactor/plugins/exit_delete_from_registry.py
```diff
@@ -42,16 +42,72 @@ def __init__(self, tasker, workflow, registries):
...
+                             manifest)
+        else:
+            msg = "failed to delete %s: %s" % (manifest, response.reason)
+            self.log.error("%s\n%s", msg, response.text)
+            raise PluginFailedException(msg)
+
+        return False
+
+    def make_manifest(self, registry, repo, digest):
+        manifest = registry + "/" + repo + "@" + digest
```
twaugh: Nitpick: to avoid unicode/str issues it's often wise to use string formatting for concatenation, e.g.
`manifest = "{registry}/{repo}@{digest}".format(**vars())`

atomic_reactor/inner.py
```diff
@@ -212,6 +212,12 @@ def add_docker_registries(self, registry_uris, insecure=False):
         for registry_uri in registry_uris:
             self.add_docker_registry(registry_uri, insecure=insecure)
 
+    def remove_docker_registry(self, docker_registry):
+        if docker_registry in self._registries["docker"]:
+            self._registries["docker"].remove(docker_registry)
+        else:
+            raise RuntimeError("removing bad docker_registry")
```
vrutkovs: Maybe `"Tried to remove docker registry outside of the list: %s" % self._registries["docker"].keys()`?

atomic_reactor/plugins/exit_delete_from_registry.py
```diff
@@ -42,16 +42,70 @@ def __init__(self, tasker, workflow, registries):
...
+        except(TypeError, KeyError):
+            # This annotation is only set for the orchestrator build.
+            # It's not present, so this is a worker build.
+            return {}
+
+        worker_digests = {}
+
+        for plat, annotation in builds.items():
+            digests = annotation['digests']
+            self.log.debug("build %s has digests: %s" % (plat, digests))
```
twaugh: nit: `self.log.debug("...", plat, digests)`

atomic_reactor/plugins/exit_delete_from_registry.py
```diff
@@ -74,6 +148,11 @@ def run(self):
                                  registry_noschema)
                 continue
 
+            if push_conf_registry.insecure:
+                # override insecure if passed
+                insecure = push_conf_registry.insecure
```
twaugh: Should `push_conf_registry` should be initialised to `None` before the loop? Currently static analysis tools complain that this might not be defined.

atomic_reactor/plugins/exit_delete_from_registry.py
```diff
@@ -42,16 +42,70 @@ def __init__(self, tasker, workflow, registries):
...
+        for plat, annotation in builds.items():
+            digests = annotation['digests']
+            self.log.debug("build %s has digests: %s" % (plat, digests))
+
+            for digest in digests:
+                reg = digest['registry']
+                if reg in worker_digests:
+                    worker_digests[reg].append(digest)
+                else:
+                    worker_digests[reg] = [digest]
```
dzickusrh: Hmm, you skipped the part where 'reg = digest['registry']', so the above won't quite work that way. :-)

I think you need something like
worker_digests = {}
worker_digests = [{ digest['registry'] : [digest] } for digest in digests]
or something like that.  But you need to handle duplicate digest['registry'] but different digest.

I am sure there is a pythonic way to do that, I just don't know how. :-(

atomic_reactor/plugins/exit_delete_from_registry.py
dzickusrh: i forgot how to run the static analysis tools locally.  Pointers?

atomic_reactor/plugins/exit_delete_from_registry.py
twaugh: I see. I think you do need a `for` loop then, but how about this:

```
for digest in digests:
    reg = digest['registry']
    worker_digests.setdefault(reg, [])
    worker_digests[reg].append(digest)
```

atomic_reactor/plugins/exit_delete_from_registry.py
twaugh: I didn't run it locally myself, I just read what landscape.io reported (see the 'Show all checks' link below).

twaugh: FWIW pylint catches this particular issue.

atomic_reactor/inner.py
```diff
@@ -212,6 +212,13 @@ def add_docker_registries(self, registry_uris, insecure=False):
         for registry_uri in registry_uris:
             self.add_docker_registry(registry_uri, insecure=insecure)
 
+    def remove_docker_registry(self, docker_registry):
+        if docker_registry in self._registries["docker"]:
+            self._registries["docker"].remove(docker_registry)
+        else:
+            raise RuntimeError("Tried to remove docker registry outside of the list: %s"
+                               % self._registries["docker"].keys())
```
twaugh: The `else` block doesn't get covered in unit testing.

Not sure if we need to do this checking anyway. Won't the `KeyError` that would get raised be sufficient?
```
def remove_docker_registry(self, docker_registry):
    self._registries['docker'].remove(docker_registry)
```

atomic_reactor/plugins/exit_delete_from_registry.py
```diff
@@ -42,64 +42,163 @@ def __init__(self, tasker, workflow, registries):
...
+            else:
+                self.log.debug("found user %s for registry %s", username, registry)
+                auth = requests.auth.HTTPBasicAuth(username, password)
+
+        return auth
+
+    def request_delete(self, url, manifest, insecure, auth):
+        response = requests.delete(url, verify=not insecure, auth=auth)
+
+        if response.status_code == requests.codes.ACCEPTED:
```
twaugh: I wonder if we need to be so strict. Would `if response.ok:` be alright here?

atomic_reactor/plugins/exit_delete_from_registry.py
```diff
@@ -42,64 +42,163 @@ def __init__(self, tasker, workflow, registries):
...
+            return True
+        elif response.status_code == requests.codes.NOT_FOUND:
+            self.log.warning("cannot delete %s: not found", manifest)
+        elif response.status_code == requests.codes.METHOD_NOT_ALLOWED:
+            self.log.warning("cannot delete %s: image deletion disabled on registry",
+                             manifest)
+        else:
+            msg = "failed to delete %s: %s" % (manifest, response.reason)
+            self.log.error("%s\n%s", msg, response.text)
+            raise PluginFailedException(msg)
```
twaugh: The `elif`/`else` blocks don't get covered during unit testing.

twaugh:
@mlangsdorf FYI

atomic_reactor/inner.py
dzickusrh: ok.

atomic_reactor/plugins/exit_delete_from_registry.py
dzickusrh: this is copied-n-pasted from the original code, but we can change that if you want.

atomic_reactor/plugins/exit_delete_from_registry.py
dzickusrh: this is also copied-n-pasted from the original code (as part of the code movement).  I can look to add coverage here.

atomic_reactor/plugins/exit_delete_from_registry.py
dzickusrh: actually ok==200 and ACCEPTED==202.  test_delete_from_registry.py uses 202 which fails with 'ok'.  What do we expect from trying to delete from the pulp?? http address, 200 or 202?

twaugh: I was talking about `response.ok`, not `response.status_code == requests.codes.ok`. Despite the naming, these are different: `response.ok` is only false when `raise_for_status()` would raise an exception.

That's what I meant about being strict. Currently we are checking for a particular response code, but I wonder if we should just accept any success code. We could do that with `if response.ok:`.

twaugh:
Please re-base against master.

twaugh:
Draft release notes updated.

PR #742, opened 2017-07-20
Increase chunk for fetch_maven_artifacts downloads

lcarva:
By default requests module uses 1 byte chunks when iterating
over streamed content. This is optimal for responsive use
cases, not so much for downloading large files. The chunk size
is now set to 10Mb to improve throughput.

### DIFF
--- a/atomic_reactor/plugins/pre_fetch_maven_artifacts.py
+++ b/atomic_reactor/plugins/pre_fetch_maven_artifacts.py
@@ -14,6 +14,7 @@ import os
 import requests
 
 from atomic_reactor import util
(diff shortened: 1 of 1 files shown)


lcarva:
Download speeds went from over 10 minutes to 2 seconds.

vrutkovs:
>The chuck size is not set to 10Mb to improve throughput.

*now* set to 10Mb, right?

lcarva:
> now set to 10Mb, right?

doh! of course, fixed commit message

lcarva:
Updated release notes.

PR #744, opened 2017-07-21
pass is_auto param to worker builds

rcerven:

### DIFF
--- a/tests/plugins/test_orchestrate_build.py
+++ b/tests/plugins/test_orchestrate_build.py
@@ -17,6 +17,7 @@ from atomic_reactor.plugins.build_orchestrate_build import (OrchestrateBuildPlug
                                                             get_worker_build_info,
                                                             get_koji_upload_dir)
 from atomic_reactor.plugins.pre_reactor_config import ReactorConfig
+from atomic_reactor.plugins.pre_check_and_set_rebuild import CheckAndSetRebuildPlugin
 from atomic_reactor.util import ImageName, df_parser
 from atomic_reactor.constants import PLUGIN_ADD_FILESYSTEM_KEY
 from flexmock import flexmock
@@ -28,6 +29,7 @@ from osbs.exceptions import OsbsException
 from tests.constants import MOCK_SOURCE, TEST_IMAGE, INPUT_IMAGE, SOURCE
 from tests.docker_mock import mock_docker
 from textwrap import dedent
+from copy import deepcopy
 
 import json
 import os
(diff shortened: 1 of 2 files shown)


lcarva:
It would be really great to verify that the return value if `is_rebuild` is the value actually used.

rcerven:
unit test added

tests/plugins/test_orchestrate_build.py
```diff
@@ -823,3 +831,72 @@ def test_orchestrate_build_failed_to_list_builds(tmpdir, fail_at):
...
+
+    runner = BuildStepPluginsRunner(
+        workflow.builder.tasker,
+        workflow,
+        [{
+            'name': OrchestrateBuildPlugin.key,
+            'args': plugin_args,
+        }]
+    )
+    workflow.prebuild_results[CheckAndSetRebuildPlugin.key] = is_auto
```
twaugh: This is the only time `is_auto` is used in this function. Is that intentional?

tests/plugins/test_orchestrate_build.py
```diff
@@ -823,3 +831,72 @@ def test_orchestrate_build_failed_to_list_builds(tmpdir, fail_at):
...
+
+    for record in caplog.records():
+        if not record.name.startswith("atomic_reactor"):
+            continue
+
+        assert hasattr(record, 'arch')
+        if record.funcName == 'watch_logs':
+            assert record.arch == 'x86_64'
+        else:
+            assert record.arch == '-'
```
twaugh: Most of this seems to be a duplicate of tests elsewhere.

tests/plugins/test_orchestrate_build.py
rcerven: it is used in expected_kwargs as well

twaugh: Ok.

tests/plugins/test_orchestrate_build.py
twaugh: Can we remove the duplicate testing here? Looks like everything from line 873.

rcerven: ok removed the same things, which were in different test and kept just the kwargs check

tests/plugins/test_orchestrate_build.py
rcerven: and the test is testing if create_worker receives expected arguments

lcarva:
Updated release notes for 1.6.26.

PR #745, opened 2017-07-21
pulp_pull, koji_import: pass pulp build_id schema information

mlangsdorf:
pulp_pull, koji_import: pass pulp build_id schema information

Change pulp_pull to return build_id and a string list that describes
the type of image schema. store_metadata_in_osv3 consumes the list
and stores the version in the annotations. koji_import reads the
annotations and stores the version in build.extra.image.media_types.

Return the v2schemav2 information if there is a v2 digest, v2schema1
if pulp_sync ran, and v1 information if pulp_pull ran. Concat and 
sort the response so it is possible to return v2schema2, v2schema1,
and v1 information.

### DIFF
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -135,6 +135,13 @@ class KojiImportPlugin(ExitPlugin):
                 # They are all None
                 extra['image']['help'] = None
 
+    def set_media_types(self, extra, worker_metadatas):
+        for platform in worker_metadatas:
+            annotations = get_worker_build_info(self.workflow, platform).build.get_annotations()
+            if annotations.get('media-types'):
+                extra['image']['media_types'] = annotations['media-types']
+                return
+
     def get_build(self, metadata, worker_metadatas):
         start_time = int(atomic_reactor_start_time)
 
@@ -174,6 +181,8 @@ class KojiImportPlugin(ExitPlugin):
                     extra['filesystem_koji_task_id'] = task_id
 
         self.set_help(extra, worker_metadatas)
+
+        self.set_media_types(extra, worker_metadatas)
 
         build = {
             'name': component,
--- a/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
+++ b/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
@@ -15,6 +15,7 @@ from osbs.conf import Configuration
 from osbs.exceptions import OsbsResponseException
 
 from atomic_reactor.plugins.pre_add_help import AddHelpPlugin
+from atomic_reactor.plugins.post_pulp_pull import PulpPullPlugin
 from atomic_reactor.constants import (PLUGIN_KOJI_IMPORT_PLUGIN_KEY,
                                       PLUGIN_KOJI_PROMOTE_PLUGIN_KEY,
                                       PLUGIN_KOJI_UPLOAD_PLUGIN_KEY)
@@ -203,6 +204,11 @@ class StoreMetadataInOSv3Plugin(ExitPlugin):
                 annotations['help_file'] = json.dumps(help_result['help_file'])
             else:
                 self.log.error("Unknown result from add_help plugin: %s", help_result)
+
+        pulp_pull_results = self.workflow.postbuild_results.get(PulpPullPlugin.key)
+        if pulp_pull_results:
+            _, media_types = pulp_pull_results
+            annotations['media-types'] = media_types
 
         tar_path = tar_size = tar_md5sum = tar_sha256sum = None
         if len(self.workflow.exported_image_sequence) > 0:
--- a/atomic_reactor/plugins/post_pulp_pull.py
+++ b/atomic_reactor/plugins/post_pulp_pull.py
@@ -15,6 +15,7 @@ discover the image ID Docker will give it.
 
 from __future__ import unicode_literals
 
+from atomic_reactor.constants import PLUGIN_PULP_PUSH_KEY, PLUGIN_PULP_SYNC_KEY
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.plugins.exit_remove_built_image import defer_removal
 from atomic_reactor.util import get_manifest_digests
@@ -64,10 +65,18 @@ class PulpPullPlugin(PostBuildPlugin):
         pullspec = image.copy()
         pullspec.registry = registry.uri  # the image on Crane
 
+        media_types = []
+        for plugin in self.workflow.postbuild_plugins_conf:
+            if plugin['name'] is PLUGIN_PULP_SYNC_KEY:
+                media_types.append('application/vnd.docker.distribution.manifest.v1+json')
+            if plugin['name'] is PLUGIN_PULP_PUSH_KEY:
+                media_types.append('application/json')
+
         digests = get_manifest_digests(pullspec, registry.uri, self.insecure, self.secret)
         if digests.v2:
             self.log.info("V2 schema 2 digest found, returning %s", self.workflow.builder.image_id)
-            return self.workflow.builder.image_id
+            media_types.append('application/vnd.docker.distribution.manifest.v2+json')
+            return self.workflow.builder.image_id, sorted(media_types)
         else:
             self.log.info("V2 schema 2 digest is not available")
 
@@ -97,4 +106,4 @@ class PulpPullPlugin(PostBuildPlugin):
                        image_id)
         self.workflow.builder.image_id = image_id
 
-        return image_id
+        return image_id, sorted(media_types)
--- a/tests/plugins/test_koji_import.py
+++ b/tests/plugins/test_koji_import.py
@@ -176,11 +176,14 @@ def is_string_type(obj):
 
 
 class BuildInfo(object):
-    def __init__(self, value, valid=True):
-        if valid:
-            self.build = BuildResponse({'metadata': {'annotations': {'help_file': value}}})
-        else:
-            self.build = BuildResponse({'metadata': {'annotations': {}}})
+    def __init__(self, help_file=None, help_valid=True, media_types=None):
+        annotations = {}
+        if media_types:
+            annotations['media-types'] = media_types
+        if help_valid:
+            annotations['help_file'] = help_file
+
+        self.build = BuildResponse({'metadata': {'annotations': annotations}})
 
 
 def mock_environment(tmpdir, session=None, name=None,
@@ -1272,3 +1275,50 @@ class TestKojiImport(object):
             assert image['help'] is None
         elif expect_result in ['skip', 'unknown_status']:
             assert 'help' not in image.keys()
+
+    @pytest.mark.parametrize('version', [
+        # no pulp plugin used
+        False,
+        # V1-only image
+        ["application/json"],
+        # V1+V2schema1 image, i.e. what we are building today
+        ["application/json", "application/vnd.docker.distribution.manifest.v1+json"],
+        # V1+V2schema2 image
+        ["application/json", "application/vnd.docker.distribution.manifest.v1+json",
+         "application/vnd.docker.distribution.manifest.v2+json"],
+        # manifest lists, as well as compatibility formats back to V1
+        ["application/json", "application/vnd.docker.distribution.manifest.v1+json",
+         "application/vnd.docker.distribution.manifest.v2+json",
+         "application/vnd.docker.distribution.manifest.list.v2+json"],
+        # manifest lists, with compatibility formats back to V2schema1, without V1
+        ["application/vnd.docker.distribution.manifest.v1+json",
+         "application/vnd.docker.distribution.manifest.v2+json",
+         "application/vnd.docker.distribution.manifest.list.v2+json"],
+    ])
+    def test_koji_import_set_media_types(self, tmpdir, os_env, version):
+        session = MockedClientSession('')
+        tasker, workflow = mock_environment(tmpdir,
+                                            name='ns/name',
+                                            version='1.0',
+                                            release='1',
+                                            session=session)
+        workflow.plugin_workspace[OrchestrateBuildPlugin.key][WORKSPACE_KEY_BUILD_INFO]['x86_64'] = BuildInfo(media_types=version)  # noqa
+
+        runner = create_runner(tasker, workflow)
+        runner.run()
+
+        data = session.metadata
+        assert 'build' in data
+        build = data['build']
+        assert isinstance(build, dict)
+        assert 'extra' in build
+        extra = build['extra']
+        assert isinstance(extra, dict)
+        assert 'image' in extra
+        image = extra['image']
+        assert isinstance(image, dict)
+        if version:
+            assert 'media_types' in image
+            assert image['media_types'] == version
+        else:
+            assert 'media_types' not in image
--- a/tests/plugins/test_pulp_pull.py
+++ b/tests/plugins/test_pulp_pull.py
@@ -19,6 +19,7 @@ import requests
 DIGEST_V1 = 'sha256:7de72140ec27a911d3f88d60335f08d6530a4af136f7beab47797a196e840afd'
 DIGEST_V2 = 'sha256:85a7e3fb684787b86e64808c5b91d926afda9d6b35a0642a72d7a746452e71c1'
 
+
 class MockerTasker(object):
     def __init__(self):
         self.pulled_images = []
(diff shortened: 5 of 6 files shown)


atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -135,6 +136,10 @@ def set_help(self, extra, worker_metadatas):
                         # They are all None
                         extra['image']['help'] = None
 
+    def set_brew_versions(self, extra, worker_metadatas):
+        build_id, brew_version = self.workflow.postbuild_results.get(PulpPullPlugin.key)
```
lcarva: You wanna read from `worker_metadatas` right?

`pulp_pull` runs in worker build, and `koji_import` in orchestrator build.
`self.workflow.postbuild_results.get(PulpPullPlugin.key)` will always return None, I think.

atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -84,4 +84,6 @@ def run(self):
                        image_id)
         self.workflow.builder.image_id = image_id
 
-        return image_id
+        # currently we only support v1+v2 schema1
```
lcarva: Sure, but this is configuration driven. This should be dynamically computed.

mlangsdorf: Isn't that what osbs-3983 is for? Right now, it's a static value.

atomic_reactor/plugins/exit_koji_import.py
mlangsdorf: How does it get added to the worker_metadatas?
Does it happen automatically, or do I have to add code to store_metadata_in_osv3.py similar to the code for the help file?

lcarva: To get the value from worker to orchestrator build, you'd need something like this:

In worker build:
1. `pulp_pull` plugin returns the information needed
2. `store_metadata` reads value from `pulp_pull` and stores it as an annotation, say `registry_content_type`

In orchestrator build:
1. `orchestrate_build` plugin additionally reads `registry_content_type` annotation for each worker build and makes it available in `self.workflow.build_result.annotations`
2. `koji_import` reads from `self.workflow.build_result.annotations['registry_content_type']` to get the information needed.

atomic_reactor/plugins/post_pulp_pull.py
lcarva: Consider the `adjust_for_registry_api_versions` method in `BuildRequest`: https://github.com/projectatomic/osbs-client/blob/master/osbs/build/build_request.py#L441
It embeds the following heuristic:

- If `pulp_push` ran, we expect `v1` content.
- If `pulp_sync` ran, we expect `v2` content (currently only schema 1).

Obviously, that is not sufficient since it cannot know whether schema 1 or schema 2 is being used for v2 content. IMO, this plugin should query the registry to figure out what exactly is there. The changes to `koji_import` should happen after that's done.

atomic_reactor/plugins/exit_koji_import.py
mlangsdorf: is the orchestrate_build step necessary? the help_file is just stored as annotations['help_file'] and gets passed on to koji_import without any specific handling by orchestrate_build.

atomic_reactor/plugins/post_pulp_pull.py
twaugh: I'd previously written this: "To handle these cases, pulp_pull should directly request the v2 schema 2 image manifest from Crane. If the v2 schema 2 image manifest is available, the plugin should take no further action."

By "directly request" I meant that it should contact the registry directly using the Docker Registry HTTP API V2, accepting v2 schema 2 image manifests, yes.

mlangsdorf: the question of the various v2 schemas is addressed in osbs-3983. Right now, I'm focused on what Brew Pulp currently supports, which is v2 schema 1.

pulp_pull always runs, as far as I know, and it doesn't handle v1 vs v2 differently that I can see.

How do I query the registry to determine v1 vs v2?

atomic_reactor/plugins/exit_koji_import.py
lcarva: I think the confusion is that `koji_import` needs to know the list of all content_types available. And `pulp_pull`, at least for now, only needs to know whether or not v2 schema 2 is available.

Making `pulp_pull` perform the checks seems reasonable to me. We should that first, if that's the desired approach. 

To query the registry, there's the helper method: `util.get_manifest_digests`
But that only queries for v2 (both schema 1 and schema 2). 
Something similar must be added for v1 content, as I'm not aware of existing code that does this.

atomic_reactor/plugins/post_pulp_pull.py
lcarva: I think the confusion is that koji_import needs to know the list of all content_types available. And pulp_pull, at least for now, only needs to know whether or not v2 schema 2 is available.

Making pulp_pull perform the checks seems reasonable to me. We should that first, if that's the desired approach.

To query the registry, there's the helper method: util.get_manifest_digests
But that only queries for v2 (both schema 1 and schema 2).
Something similar must be added for v1 content, as I'm not aware of existing code that does this.

atomic_reactor/plugins/exit_koji_import.py
lcarva: For help_file, it appears to follow the same path, but it uses the `metadata_fragment` data structure which is already handled by `orchestrate_build`.

It could be a bad case of Mondays, but I'm not really seeing how the help_file key is actually added to `metadata_fragment` :(

lcarva: Update: the worker build sets the `help_file` in its annotations. Then in `koji_import`, the annotations are fetched again from the worker build (build.get_annotations() vs .get_annotations()). So yes, for help_file, it bypasses the need to make `orchestrate_build` aware of it. I'm not sure if this is actually better though. Seems to me, the initial idea was to have annotations being fetched from one place, `orchestrate_build`.

atomic_reactor/plugins/post_pulp_pull.py
mlangsdorf: I don't want the manifest_digests, and rewriting that code to determine whether we have v2schema1 manifest_digests or v2schema2 manifest_digests should be outside the scope of this task. I just want to know if we've run pulp_push and pulp_pull should return ['application/json'], or if we've run pulp_sync and pulp_pull should return ['application/json', 'application/vnd.docker.distribution.manifest.v1+json']

is there a way for the orchestrator build to know what plugins the worker builds ran?

mlangsdorf: fixed

atomic_reactor/plugins/exit_koji_import.py
mlangsdorf: Do you want me to change the current code or is the worker->store_metadata_in_osv3->orchestrator->koji_import work flow sufficient for now?

I think if we want to insert orchestrate_build into the process, it should be done the same way for both media_version and help_file and there should be a single small story to unify that handling.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -135,6 +135,13 @@ def set_help(self, extra, worker_metadatas):
                         # They are all None
                         extra['image']['help'] = None
 
+    def set_brew_versions(self, extra, worker_metadatas):
```
twaugh: Let's call it `set_media_types`.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -135,6 +135,13 @@ def set_help(self, extra, worker_metadatas):
...
+        for platform in worker_metadatas:
+            annotations = get_worker_build_info(self.workflow, platform).build.get_annotations()
+            if 'media_version' in annotations and annotations['media_version']:
+                extra['image']['media_types'] = annotations['media_version']
```
twaugh: Any reason not to call the annotation "media-types"? ('types' for consistency with the CG metadata name, and '-' for consistency with other annotations)

atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -84,4 +84,9 @@ def run(self):
                        image_id)
         self.workflow.builder.image_id = image_id
 
-        return image_id
+        media_version = ['application/json']
```
twaugh: `media_types`?

tests/plugins/test_koji_import.py
```diff
@@ -176,11 +176,16 @@ def is_string_type(obj):
 
 
 class BuildInfo(object):
-    def __init__(self, value, valid=True):
-        if valid:
-            self.build = BuildResponse({'metadata': {'annotations': {'help_file': value}}})
-        else:
-            self.build = BuildResponse({'metadata': {'annotations': {}}})
+    def __init__(self, help_file=None, help_valid=True, brew_version=None):
```
twaugh: `media_types`?

atomic_reactor/plugins/exit_koji_import.py

atomic_reactor/plugins/exit_koji_import.py

atomic_reactor/plugins/post_pulp_pull.py
mlangsdorf: fixed.

tests/plugins/test_koji_import.py
mlangsdorf: fixed.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -135,6 +135,13 @@ def set_help(self, extra, worker_metadatas):
                         # They are all None
                         extra['image']['help'] = None
 
+    def set_media_types(self, extra, worker_metadatas):
+        for platform in worker_metadatas:
+            annotations = get_worker_build_info(self.workflow, platform).build.get_annotations()
+            if 'media-types' in annotations and annotations['media-types']:
```
lcarva: This can be simplified as: `if annotations.get('media-types'):`
The `.get` dict method will return a default value, `None` in this case, if key is not set.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -204,6 +205,11 @@ def run(self):
             else:
                 self.log.error("Unknown result from add_help plugin: %s", help_result)
 
+        pulp_pull_results = self.workflow.postbuild_results.get(PulpPullPlugin.key)
+        if pulp_pull_results:
+            build_id, media_types = pulp_pull_results
```
lcarva: This will change the value of `build_id` that was set at the beginning of this method? If this is not intentional, please use something like this: `_, media_types = pulp_pull_results`.
Otherwise, I think a comment is needed explaining why we want to reset the `build_id` in this case.

Reading pulp_pull, this appears to actually be the `image_id`.

atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -84,4 +84,9 @@ def run(self):
                        image_id)
         self.workflow.builder.image_id = image_id
 
-        return image_id
+        media_types = ['application/json']
```
lcarva: Should this be conditionally based on whether or not `pulp_push` ran?

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -204,6 +205,11 @@ def run(self):
             else:
                 self.log.error("Unknown result from add_help plugin: %s", help_result)
 
+        pulp_pull_results = self.workflow.postbuild_results.get(PulpPullPlugin.key)
```
lcarva: Let's use the key as provided by constants: `PLUGIN_PULP_PUSH_KEY`

mlangsdorf: PulpPullPlugin.key != PLUGIN_PULP_PUSH_KEY

atomic_reactor/plugins/post_pulp_pull.py
mlangsdorf: if neither pulp_push nor pulp_sync ran, then pulp_pull won't run either.
if pulp_push or pulp_sync ran, the first element in media_types will be ['application/json'] - at least until we handle "manifest lists, with compatibility formats back to V2schema1, without V1" which is more forward-proofing than I think we need to be concerned with at this time.

In short, returning ['application/json'] always appears to be correct unless I'm misreading https://projects.engineering.redhat.com/browse/OSBS-3874

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
mlangsdorf: Fixed.

atomic_reactor/plugins/exit_koji_import.py
mlangsdorf: Fixed

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
lcarva: doh! of course.

atomic_reactor/plugins/post_pulp_pull.py
lcarva: We currently don't run a configuration in which we do not produce v1 content only (pulp_push always runs). But that won't always be the case. Hard coding it like this adds an incorrect dependency IMO.

twaugh: In the near future, pulp_pull will need to move into the orchestrator build, and there it won't know whether pulp_pull ran. Maybe it should be told via parameter, based on what api versions osbs.conf says to support?

twaugh: Thinking more about this, I think the answer right now is to make "application/json" conditional on whether pulp_push ran (in the same build as it is running).

Later, to support pulp_publish in the orchestrator build, we'll need pulp_push to store annotations in the worker build that say what v1 content was uploaded, and when we get to that point we can use those annotations to decide whether to include "application/json" in the list.

tests/plugins/test_pulp_pull.py
```diff
@@ -100,13 +101,25 @@ def custom_get_v1(self, url, headers, **kwargs):
     def custom_get_v2(self, url, headers, **kwargs):
         return self.config_response_config_v2
 
-    @pytest.mark.parametrize(('insecure', 'config_version'), [
-        (True, 'v1'),
-        (True, 'v2'),
-        (False, 'v1'),
-        (False, 'v2'),
+    @pytest.mark.parametrize('insecure', [(True, False),])
```
mlangsdorf: Fixed

atomic_reactor/plugins/post_pulp_pull.py
mlangsdorf: Fixed, I think. Commit message updated.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
mlangsdorf: Fixed

tests/plugins/test_pulp_pull.py
```diff
@@ -100,13 +101,29 @@ def custom_get_v1(self, url, headers, **kwargs):
...
+        ('v2', [{'name': 'pulp_sync'}],
+         ['application/vnd.docker.distribution.manifest.v1+json',
+          'application/vnd.docker.distribution.manifest.v2+json']),
+        ('v2', [{'name': 'pulp_sync'}, {'name': 'pulp_push'}],
+         ['application/json',
+          'application/vnd.docker.distribution.manifest.v1+json',
+          'application/vnd.docker.distribution.manifest.v2+json']),
     ])
-    def test_pull_first_time(self, insecure, config_version):
+    def test_pull_first_time(self, insecure, config_version, pulp_plugin, expected_version):
```
twaugh: Can we rename `config_version` to `schema_version`? Sorry, I know you didn't add it.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -204,6 +205,11 @@ def run(self):
...
+        if pulp_pull_results:
+            image_id, media_types = pulp_pull_results
```
twaugh: Since we're not using `image_id`, we can write this as `_, media_types = pulp_pull_results`.

lcarva: I'm actually surprised stickler didn't complain about this.

atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -64,10 +64,18 @@ def run(self):
         pullspec = image.copy()
         pullspec.registry = registry.uri  # the image on Crane
 
+        media_types = []
+        for plugin in self.workflow.postbuild_plugins_conf:
+            if 'pulp_sync' in plugin['name']:
```
lcarva: Nitpick, but shouldn't this be an equality check? It can also leverage the constants: `PLUGIN_PULP_PUSH_KEY, PLUGIN_PULP_SYNC_KEY`

tests/plugins/test_koji_import.py
```diff
@@ -176,11 +176,16 @@ def is_string_type(obj):
...
 
 class BuildInfo(object):
-    def __init__(self, value, valid=True):
-        if valid:
-            self.build = BuildResponse({'metadata': {'annotations': {'help_file': value}}})
-        else:
-            self.build = BuildResponse({'metadata': {'annotations': {}}})
+    def __init__(self, help_file=None, help_valid=True, media_types=None):
+        annotations = {}
+        if media_types is None:
```
lcarva: Just curious, but why not `if not media_types`? Are we intentionally handling an empty list differently?

atomic_reactor/plugins/post_pulp_pull.py
mlangsdorf: Fixed.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
mlangsdorf: Fixed.

tests/plugins/test_pulp_pull.py
mlangsdorf: Fixed.

tests/plugins/test_koji_import.py
mlangsdorf: Yes. The pulp_plugin does not need to be run, so there needs to be a distinction between media_types is False (and no annotation is created) and media_types is [] (create a V1 image).

But looking over the code, that's an overly complicated way of expressing it and it's better to just drop both lines.

lcarva:
Release notes updated.

PR #746, opened 2017-07-21
Add unit tests for squash plugin

lcarva:

### DIFF
--- a/tests/plugins/test_squash.py
+++ b/tests/plugins/test_squash.py
@@ -1,0 +1,149 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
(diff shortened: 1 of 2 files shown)


tests/plugins/test_squash.py
```diff
@@ -0,0 +1,139 @@
...
+import logging
+import os
+import pytest
+
+from flexmock import flexmock
+
+from atomic_reactor.constants import EXPORTED_SQUASHED_IMAGE_NAME
+from atomic_reactor.core import DockerTasker
+from atomic_reactor.inner import DockerBuildWorkflow
+from atomic_reactor.plugin import PrePublishPluginsRunner, PluginFailedException
```
lcarva: It is now.

lcarva:
One line is not covered, but I believe it's just an invalid scenario in the plugin.

PR #747, opened 2017-07-24
Flatpak support

owtaylor:
This is an initial version of changes for flatpak support. It builds workable Flatpak applications and runtimes, and has pretty complete test cases. Things I know that need improvement:

 * It only has been tested with a hand-crafted json file and 'atomic-reactor build json --mode=here'. I think it should work fine in a containerized build, but haven't tested it. orchestrated builds will need more work, something that I'm planning to look at next.
 * It depends on a manually-created yum repository of the module that was built through the Module Build Service and koji, because no repository is built automatically. The code will need to be modified to ask the "On Demand Compose Service" for a compose once that lands within the Fedora repository.

Please let me know about all the things that I don't know about yet that need improvement :-).


owtaylor:
Note that travis-ci failures are expected because 'flatpak' and 'desktop-file-utils' need to be installed within the test environment.

lcarva:
> Note that travis-ci failures are expected because 'flatpak' and 'desktop-file-utils' need to be installed within the test environment.

You can control that by specifying it in requirements.txt or test/requirements.txt

If that's not possible, then those calls need to be mocked.

tests/plugins/test_flatpak_create_dockerfile.py
```diff
@@ -0,0 +1,124 @@
...
+from urlparse import urlparse, parse_qs
+
+from atomic_reactor.inner import DockerBuildWorkflow
+from atomic_reactor.plugins.pre_flatpak_create_dockerfile import FlatpakCreateDockerfilePlugin
+from atomic_reactor.plugin import PreBuildPluginsRunner
+from atomic_reactor.source import VcsInfo
+from atomic_reactor.util import ImageName
+
+from tests.constants import (MOCK_SOURCE, FLATPAK_GIT, FLATPAK_SHA1)
+from tests.fixtures import docker_tasker
```
vrutkovs: Stickler CI doesn't know about py.test fixtures. Lets add `# noqa` to every line it complains about unused fixture (or redefinition of unused fixture)

vrutkovs:
Does ubuntu has flatpak ppa? It should be installed in `.travis.yml` to have flatpak binary.

Another way to resolve test failures is to create dummy flatpak script

atomic_reactor/plugins/post_flatpak_create_oci.py
```diff
@@ -0,0 +1,308 @@
...
+            shutil.move(full_path,
+                        os.path.join(os.path.dirname(full_path),
+                                     add_app_prefix(app_id,
+                                                    basename[:-len('.desktop')],
+                                                    basename)))
+
+
+class FlatpakCreateOciPlugin(PostBuildPlugin):
+    key = 'flatpak_create_oci'
+    is_allowed_to_fail = False
```
vrutkovs: This should be set to `True` so that the exception in the plugin would fail the build

atomic_reactor/plugins/post_flatpak_create_oci.py
```diff
@@ -0,0 +1,308 @@
+"""
+Copyright (c) 2015 Red Hat, Inc
```
vrutkovs: Nitpick: 2017

atomic_reactor/plugins/post_flatpak_create_oci.py
```diff
@@ -0,0 +1,308 @@
...
+    def __init__(self, tasker, workflow):
+        """
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        """
+        super(FlatpakCreateOciPlugin, self).__init__(tasker, workflow)
+
+    def _get_target_path(self, export_path):
+        if self.source.runtime:
+            if export_path == "var/tmp/flatpak-build":
```
vrutkovs:
```
return {
 'var/tmp/flatpak-build': 'files',
 ...
}[export_path]
```
would look simpler. The dict should probably moved outside of the function

atomic_reactor/plugins/post_flatpak_create_oci.py
```diff
@@ -0,0 +1,308 @@
...
+        basename = os.path.basename(full_path)
+        if not basename.startswith(app_id):
+            shutil.move(full_path,
+                        os.path.join(os.path.dirname(full_path),
+                                     add_app_prefix(app_id,
+                                                    basename[:-len('.desktop')],
+                                                    basename)))
+
+
+class FlatpakCreateOciPlugin(PostBuildPlugin):
```
vrutkovs: This probably should be a `PrePublishPlugin`, similar to `prepub_squash`

atomic_reactor/plugins/post_flatpak_create_oci.py
```diff
@@ -0,0 +1,308 @@
...
+        image = self.workflow.image
+        self.log.info("Creating temporary docker container")
+        container_dict = self.tasker.d.create_container(image)
+        container_id = container_dict['Id']
+
+        try:
+            return self._export_container(container_id)
+        finally:
+            self.log.info("Cleaning up docker container")
+#            self.tasker.d.remove_container(container_id)
```
vrutkovs: IIUC this was commented for testing, right?

atomic_reactor/plugins/post_flatpak_create_oci.py
```diff
@@ -0,0 +1,308 @@
...
+        os.mkdir(builddir)
+
+        repo = os.path.join(self.workflow.source.workdir, "repo")
+        subprocess.check_call(['ostree', 'init', '--mode=archive-z2', '--repo', repo])
+
+        runtime_id = info['runtime']
+        runtime_version = info['runtime-version']
+
+        with open(os.path.join(builddir, 'metadata'), 'w') as f:
+            f.write("""[Runtime]
```
vrutkovs: `from textwrap import dedent` and `dedent("""long text""")` should be used here to preserve the identation

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -43,14 +45,62 @@ def __init__(self, tasker, workflow, registries):
...
+            if last_image['type'] == IMAGE_TYPE_OCI or last_image['type'] == IMAGE_TYPE_OCI_TAR:
+                return True
+
+        return False
+
+    def push_with_skopeo(self, registry_image, insecure, docker_push_secret):
+        # If the last image has type OCI_TAR, then hunt back and find the
+        # the untarred version, since skopeo only supports OCI's as an
+        # untarred directory
+        index = 1
```
vrutkovs: Maybe
```
image = [x for x in self.workflow.exported_image_sequence if x['type'] != IMAGE_TYPE_OCI_TAR][-1]
```
would be prettier?

atomic_reactor/plugins/pre_flatpak_create_dockerfile.py
```diff
@@ -0,0 +1,208 @@
...
+import json
+import os
+from pdc_client import PDCClient
+from modulemd import ModuleMetadata
+
+from atomic_reactor.constants import FLATPAK_FILENAME, DOCKERFILE_FILENAME, YUM_REPOS_DIR
+from atomic_reactor.plugin import PreBuildPlugin
+from atomic_reactor.util import render_yum_repo
+
+BASE_IMAGE = 'registry.fedoraproject.org/fedora:26'
```
vrutkovs: This probably should be a plugin parameter

owtaylor:
> Does ubuntu has flatpak ppa? It should be installed in .travis.yml to have flatpak binary.

There is a flatpak ppa, but looking at the [Travis docs](https://docs.travis-ci.com/user/reference/precise/), it seems like that would require switching the Travis run from a container (sudo: false) to a VM (sudo: required) - possibly slowing down the Travis runs significantly. Which I feel would be presumptuous for just my own thing. :-)

> Another way to resolve test failures is to create dummy flatpak script

I'll keep that in mind, but mocking ostree/flatpak via subprocess seems more straightforward - keeps everything in one place.

vrutkovs:
There is another PR - https://github.com/projectatomic/atomic-reactor/pull/751 - created to resolve version problem in Travis. When it gets merged the script might include flatpak to make tests pass.

For now it would be best to disable tests if flatpak is not available, see `tests/test_http.py`:
```
def has_connection():
    # In case we run tests in an environment without internet connection.

    if sys.version_info < (2, 7):
        # py 2.6 doesn't have SNI support, required for httpbin, as it has SSLv3 certificates
        return False

    try:
        HttpStream("https://httpbin.org/get", "get")
        return True
    except OsbsNetworkException:
        return False

@pytest.mark.skipif(not has_connection(),
                    reason="requires internet connection")
class TestHttpSession(object):
...
```

atomic_reactor/plugins/post_tag_and_push.py
owtaylor: My version certainly is a bit ugly. It's also wrong - the assertion doesn't work right. Your version looks good. The O(1) version would be something like:

        image = next((x in reversed(self.workflow.exported_image_sequence) if
                      x['type'] != IMAGE_TYPE_OCI_TAR), None)
        assert image is not None

But the use of the next() built-in is a bit cryptic (I had to look it up), and exported_image_sequence is going to be super short.

atomic_reactor/plugins/post_flatpak_create_oci.py
owtaylor: Isn't that backwards? It's set to False now, so it's *not* allowed to fail.

vrutkovs: Oh, correct, my fault

atomic_reactor/plugins/post_flatpak_create_oci.py
owtaylor: This was a pretty confusing function.... I rewrote it for clarity, then it bothered me that it was doing a whole bunch of string concatenation and slicing for tens of thousands of files, so I rewrote it again. Took half an afternoon to get back to something that does the same thing. :-/

atomic_reactor/plugins/post_flatpak_create_oci.py
owtaylor: I'm not sure I entirely understand the distinction - the pre-publish plugins are run first, but the post-build plugins need a particular ordering from the build.json anyways. But the thought did occur to me "since it modifies the image, it's like squash"

The reason that I didn't do it that way is the post_rpmqa plugin. The "modified image" created by flatpak_create_oci no longer has rpm in it, so you can't run 'rpm -qa' in it - so from that perspective, post_rpmqa should conceptually run *before* flatpak_create_oci. In practical terms, it doesn't matter, since the original image is still there, and post_rpmqa is just going to run against that.

I'll go ahead and change it.

atomic_reactor/plugins/post_tag_and_push.py
vrutkovs: I think `next(` version is fine as long as there is comment explaining what's happening there

owtaylor:
This update should fix all the comments from reviews so far.

I went ahead and mocked the ostree and flatpak usage. I tried changing the .travis.yml to pull in the (available everywhere, whitelisted in Travis) desktop-file-utils - I'll have to see if that works. I left the mocking of flatpak/ostree as a separate commit so it can be dropped if atomic-reactor is switched to testing in VM's before this lands and can pull in the necessary PPA.

This push also contains commits from PR #764 and PR #767 that it depends on. I'll rebase once those land (or things are resolved a different way.)

owtaylor:
The latest push push adds support for looking up a yum repository for a module build via the "On Demand Compose Server". I left this as separate commits for clarity, though it involves some amount of restructuring of the earlier commits.

atomic_reactor/util.py
```diff
@@ -719,24 +734,27 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
...
+        # v2 manifest into a v1 manifest as the default type, so the same
+        # thing occurs.
+        if version != 'v2' and response.status_code == 400:
             logger.warning('Unable to fetch digest for %s, got error %s',
                            media_type, response.status_code)
             continue
 
+        # Returned if the manifest could not be retrieved for the given
+        # media type
+        if response.status_code == 404 or response.status_code == 406:
```
twaugh: Could we use symbolic names like `requests.codes.not_found` and `requests.codes.not_acceptable` here?

atomic_reactor/plugins/prepub_flatpak_create_oci.py
```diff
@@ -0,0 +1,334 @@
...
+        runtime_version = info['runtime-version']
+
+        builddir = os.path.join(self.workflow.source.workdir, "build")
+        os.mkdir(builddir)
+
+        repo = os.path.join(self.workflow.source.workdir, "repo")
+
+        subprocess.check_call(['flatpak', 'build-init',
+                               builddir, app_id, runtime_id, runtime_id, runtime_version])
+        subprocess.check_call(['tar', 'xCf', builddir, tarfile])
```
owtaylor: The question is performance - from experience writing this plugin _compression_ with tarfile is very slow - the reason for the dance with streaming to gzip when extracting and munging the filesystem. But what about decompression? Testing decompression with an actual 91-megabyte filesystem tarball:

```
$ time tar xfz filesystem.tar.gz
real	0m2.369s
user	0m1.988s
sys	0m1.038s

python:
    import sys, tarfile;
    tf = tarfile.open(sys.argv[1]);
    tf.extractall();
    tf.close();
real	0m4.803s
user	0m3.577s
sys	0m1.105s
```

Which is sort of indeterminate - I don't like slowing down a build of a large module by 3 seconds, but it's not that bad. For comparison, compression is a much bigger deal:

```
$ time tar cfz foo.tar.gz file
real	0m12.276s
user	0m12.137s
sys	0m0.498s

python:
     import tarfile, sys;
     tf = tarfile.open(sys.argv[1], "w:gz");
     tf.add(sys.argv[2]);
     tf.close();
real	0m44.857s
user	0m44.317s
sys	0m0.406s
```

On the other hand, when tarring up the resulting OCI image - a few large files with no compression - the Python vs. tar difference is irrelevant:

```
$ time tar cf foo.tar org.fedoraproject.Platform
real	0m0.122s
user	0m0.005s
sys	0m0.116s


Python:
real	0m0.157s
user	0m0.038s
sys	0m0.115s
```

(The conclusion from this might be to drop the compression for the intermediate filesystem tarball, but laptop SSD IO is not the same as IO on the build system, so hard to know...)

Long story short, my plan is:
 * Drop all other usages of subprocess.check_call(['tar', ...])
 * But keep the one you commented in with the comment 'With gzip'ed tarball, tar is several seconds faster than tarfile.extractall'

atomic_reactor/util.py
owtaylor: Sure - your call. Once you get beyond 200/404 the numeric status codes are arguably less self-documenting ;-)

atomic_reactor/plugins/pre_resolve_module_compose.py
```diff
@@ -0,0 +1,266 @@
...
+        :param pdc_insecure: If True, don't check SSL certificates for `pdc_url`
+        :
+        """
+        # call parent constructor
+        super(ResolveModuleComposePlugin, self).__init__(tasker, workflow)
+
+        if not pdc_url:
+            raise RuntimeError("pdc_url is required")
+        if not odcs_url:
+            raise RuntimeError("odcs_url is required")
```
owtaylor: Grouping is certainly one consideration. Another consideration that may have been in my head is that  python2, you get very unclear messages when trying to pass non-kwargs as kwargs goes wrong (foo() takes at least 4 arguments (6 given) - python3 is better. I think I'll keep it this way since I do like having the parameters grouped.

atomic_reactor/plugins/pre_resolve_module_compose.py
```diff
@@ -0,0 +1,266 @@
...
+        :param module_name: Module name to look up in PDC
+        :param module_stream: Module stream to look up in PDC
+        :param module_version: Module version to look up in PDC (optional)
+        :param compose_id: ID of compose in ODCS (optional - will only be set for workers)
+        :param odcs_url: URL of ODCS (On Demand Compose Service)
+        :param odcs_insecure: If True, don't check SSL certificates for `odcs_url`
+        :param odcs_openidc_secret_path: directory to look in for a `token` file (optional)
+        :param pdc_url: URL of PDC (Product Definition center))
+        :param pdc_insecure: If True, don't check SSL certificates for `pdc_url`
+        :
```
owtaylor: Rather, a stray leftover in the parameters list.

atomic_reactor/plugins/pre_flatpak_create_dockerfile.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        # call parent constructor
+        super(FlatpakCreateDockerfilePlugin, self).__init__(tasker, workflow)
+
+        self.base_image = base_image
+
+    def _load_source(self):
+        flatpak_path = os.path.join(self.workflow.builder.df_dir, FLATPAK_FILENAME)
+        if not os.path.isfile(flatpak_path):
+            raise RuntimeError("{} not found".format(flatpak_path))
```
owtaylor: the thought process was that this is the place where you'll get a failure if you try to build a non-flatpak repository with a Dockerfile as a flatpak, so didn't want to let a raw error happen, but agreed that the RuntimeError that I added here is no friendlier than the default IOError - I'll just remove it.

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -43,14 +45,62 @@ def __init__(self, tasker, workflow, registries):
...
+            source_img = 'docker-archive://' + image['path']
+        else:
+            raise RuntimeError("Attempt to push unsupported image type %s with skopeo",
+                               image['type'])
+
+        dest_img = 'docker://' + registry_image.to_str()
+
+        cmd += [source_img, dest_img]
+
+        self.log.info("Calling: %s", ' '.join(cmd))
```
owtaylor: Thanks for catching this - had noticed that, but didn't do anything hoping to find some better way to send auth to skopeo. If I don't log, and make sure that failure of skopeo doesn't log, do you think this is good enough (probably don't need to worry about ps on the build system)? The only alternative I can find is to create a FAKE_HOMEDIR/.docker/config.json and point to that with $HOME when calling skopeo, which is a lot of work, and somewhat fragile w.r.t. to the exact handling of config.json.

owtaylor:
The latest push should address all review comments from @twaugh - I squashed most of the fixes with the original, but there are three loose commits at the end, one unrelated to the review, but that I found I needed as I kept testing this:

 * A commit to add an append parameter to the bump_release plugin so that we can have versions like 201407040100203.2 in Koji and avoid conflicts. *needs review*

And two fixes from the review that are larger and could use a look-over:

 * Changes to use 'tarfile' instead calling tar via subprocess
 * Changes to avoid logging the password when we need to pass it on the command line to skopeo.

owtaylor:
Latest changes:
 * A requirements.txt addition to workaround a missing requirement in modulemd (was causing Travis CI to fail ... thought I pushed this earlier, but forgot)
 * A fix for 'flatpak build-init' requiring the runtime to be installed to install the app
 * A fix for a problem with an extra level of directory in the tarball

atomic_reactor/plugins/pre_resolve_module_compose.py
```diff
@@ -0,0 +1,272 @@
...
+                         burst_length=30,
+                         slow_retry=10,
+                         timeout=300):
+        self.log.debug("Getting compose information for information for compose_id={}"
+                       .format(compose_id))
+        url = self.url + 'composes/' + str(compose_id)
+        headers = self._auth_headers()
+        start_time = time.time()
+        while True:
+            response = requests.get(url, headers=headers)
```
twaugh: Instead of using `requests` directly, `from atomic_reactor.util import get_retrying_requests_session` and use e.g.
```
session = get_retrying_requests_session()
response = session.get(...)
```

That way, retries will be handled transparently (for retryable requests).

atomic_reactor/plugins/pre_resolve_module_compose.py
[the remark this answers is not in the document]
twaugh: OK.

atomic_reactor/plugins/pre_resolve_module_compose.py
```diff
@@ -0,0 +1,272 @@
...
+        if self.odcs_openidc_secret_path:
+            token_path = os.path.join(self.odcs_openidc_secret_path, 'token')
+            with open(token_path, "r") as f:
+                odcs_token = f.read().strip()
+        else:
+            odcs_token = None
+
+        odcs_client = ODCSClient(self.odcs_url, insecure=self.odcs_insecure, token=odcs_token,
+                                 log=self.log)
+        pdc_client = PDCClient(server=self.pdc_url, ssl_verify=not self.pdc_insecure, develop=True)
```
twaugh: `develop` is for using in a development environment? Should this be a module parameter?

atomic_reactor/plugins/pre_resolve_module_compose.py
```diff
@@ -0,0 +1,272 @@
...
+                'source': source
+            }
+        }
+
+        self.log.info("Starting compose for source_type={source_type}, source={source}"
+                      .format(source_type=source_type, source=source))
+        response = requests.post(self.url + 'composes/',
+                                 json=body,
+                                 headers=self._auth_headers())
+        response.raise_for_status()
```
twaugh: Are there any circumstances in which this POST request can be safely retried?

atomic_reactor/plugins/prepub_flatpak_create_oci.py
```diff
@@ -0,0 +1,361 @@
...
+def get_arch():
+    return subprocess.check_output(['flatpak', '--default-arch'],
+                                   universal_newlines=True).strip()
+
+
+# flatpak build-init requires the sdk and runtime to be installed on the
+# build system (so that subsequent build steps can execute things with
+# the SDK). While it isn't impossible to download the runtime image and
+# install the flatpak, that would be a lot of unnecessary complexity
+# since we our build step is just unpacking the filesystem we've already
```
twaugh: Extra word (we?)

atomic_reactor/plugins/prepub_flatpak_create_oci.py
```diff
@@ -0,0 +1,361 @@
...
+            source = re.sub("^ROOT", ROOT, source)
+            if source.endswith("/"):
+                patterns.append((re.compile(source + "(.*)"), target))
+                patterns.append((source[:-1], target))
+            else:
+                patterns.append((source, target))
+
+        def get_target_func(self, path):
+            for source, target in patterns:
+                if isinstance(source, str):
```
twaugh: This is slightly fragile because of the differences between Python 2 (`str` and `unicode`) and Python 3. Possibly more robust approach:
```
try:
    m = source.match(path)
    if m:
        return os.path.join(target, m.group(1))
except AttributeError:
    # Not a regular expression, just a filename
    if source == path:
        return target
```

tests/plugins/test_bump_release.py
```diff
@@ -187,3 +190,34 @@ def ssl_login(self, cert, ca, serverca, proxyuser=None):
             assert 'Release' not in parser.labels
         else:
             assert parser.labels['Release'] == next_release['expected']
+
+    @pytest.mark.parametrize('base_release,builds,expected', [
+        ('42', [], '42.1'),
+        ('42', ['42.1', '42.2'], '42.3'),
```
twaugh: Could `base_release` already be in the form `'42.1'`?

atomic_reactor/plugins/pre_resolve_module_compose.py
owtaylor: 'develop' actually means "try API requests without authentication" - since all the PDC access we need is public and possible without authentication, this makes things a lot simpler and avoids us having to set up OSBS <=> PDC authentication. Clearly needs a comment.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -342,11 +354,18 @@ def get_fs_task_id(self):
         return task_id
 
     def do_worker_build(self, release, cluster_info, koji_upload_dir, task_id):
+        if self.key in self.workflow.plugin_workspace:
+            workspace = self.workflow.plugin_workspace[self.key]
+            override_kwargs = workspace.get(WORKSPACE_KEY_OVERRIDE_KWARGS, {})
+        else:
+            override_kwargs = {}
```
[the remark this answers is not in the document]
owtaylor: Sure, fixed

atomic_reactor/util.py
```diff
@@ -461,16 +468,17 @@ def get_docker_architecture(tasker):
...
-    if not path or not os.path.isfile(path):
-        logger.error('%s is not a file', path)
-        return
+def get_exported_image_metadata(path, image_type):
+    logger.info('getting metadata for exported image %s (%s)', path, image_type)
+    metadata = {'path': path, 'type': image_type}
+    if image_type != IMAGE_TYPE_OCI:
+        if not path or not os.path.isfile(path):
+            logger.error('%s is not a file', path)
+            return
```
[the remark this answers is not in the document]
owtaylor: Not something that I added, but sure, I agree. I'll stick a small fixup patch for this into the next push.

atomic_reactor/plugins/pre_flatpak_create_dockerfile.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def __init__(self, tasker, workflow,
+                 base_image=None):
+        """
+        constructor
+
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param base_image: host image used to install packages when creating the Flatpak
+        :
```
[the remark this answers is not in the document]
owtaylor: Nope

atomic_reactor/plugins/pre_resolve_module_compose.py
owtaylor: Hmm, this is not retrying on failure, it is polling against the json content of successful requests. I don't think there's any way obvious way to use the urllib3 Retry functionality for this. [I also would be reluctant to have exponential back-off here... I think it would cause unnecessary slow waits.] The individual requests could be retried, of course.

atomic_reactor/plugins/pre_resolve_module_compose.py
owtaylor: Define "safe" :-) It's clearly safe to retry on a connection error. On other errors, it's hard to say - nothing horrible will go wrong if multiple composes are queued, but extra work will be done. 

(To be precise, only composes in the *done* state are reused, so additional copies of the compose are started until the first copy completes - but there's a configurable limit to how many composes are run in parallel - by default 2. If we retry at 5 seconds, 10 seconds, and 20 seconds, as I believe the retry default is, probably at most one extra compose will be done.)

atomic_reactor/plugins/prepub_flatpak_create_oci.py
owtaylor: Fixed

atomic_reactor/plugins/prepub_flatpak_create_oci.py
owtaylor: Good point to be cautious about `isinstance(x, str)`, but since the rules are hardcoded in the source code, we're just counting that `isinstance("somestring", str)`, which should be safe both python2 and python3. I'd be inclined to leave it as is, but if you want, I'll switch to compiling patterns for everything and skip the literal-matches-as-strings "optimization". Rough timings

  1.5us - isinstance(str) and string compare
  4.7us - try match, catch AttributeError, string compare
  3.5us - compile a ^foo$ regex and just use that

 (I think we'd do max. 100,000 rule / filename matches in any case, so the worst/best difference is < .3s, and not a big deal.)

twaugh: Maybe using a pattern for everything would be better then. As it stands, even with hardcoded strings, a `from __future__ import unicode_literals` statement (used in lots of other modules) would break it.

Using patterns even for fixed strings might make the code easier to follow anyway.

atomic_reactor/plugins/pre_resolve_module_compose.py
twaugh: We do also need to retry on failure though, to allow for ODCS to be offline for short periods.
http://osbs.readthedocs.io/en/latest/contributors.html#retry-on-error

atomic_reactor/util.py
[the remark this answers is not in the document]
twaugh: Oh, sorry, I didn't realised that's just pre-existing code that moved.

owtaylor:
Just pushed an updated version with fixes for your comments and rebased to the latest master. I flattened most of the trivial stuff into existing commits. New commits:

 util.py: Fix swapped parameters in an error message - small fix found when debugging rebase

 util.get_exported_image_metadata: handle bad paths better - removes the pointless return of None

 flatpak_create_oci: Avoid isinstance(x, str) - ended up just adding an extra boolean

 resolve_module_compose: retry HTTP requests - retrying POST and GET. Couldn't figure out how to tests the retries in the unit tests because the retry logic is so deep into the structure of urllib3 - you'd pretty much have to spin up a HTTP server. 

Beyond that, the main thing of note is that get_manifest_digests() got uglier with this rebase because post_pulp_pull and its tests want there to be a difference between A) a registry returning an empty ManifestDigests object because the registry had the manifest but it couldn't be processed [fail immediately] B) The registry not having the manifest at all. [retry]. Saving a 404 error to reraise in the second case causes the ugliness.

atomic_reactor/plugins/pre_resolve_module_compose.py
twaugh: Looks like this still needs a comment.

atomic_reactor/plugins/prepub_flatpak_create_oci.py
```diff
@@ -0,0 +1,361 @@
...
+
+        # with gzip'ed tarball, tar is several seconds faster than tarfile.extractall
+        subprocess.check_call(['tar', 'xCfz', builddir, tarred_filesystem])
+
+        update_desktop_files(app_id, builddir)
+
+        subprocess.check_call(['flatpak', 'build-finish'] + info['finish-args'] + [builddir])
+        subprocess.check_call(['flatpak', 'build-export', repo, builddir])
+
+        # FIXME version?
```
twaugh: Is this about checking the version of the flatpak binary?

tests/plugins/test_bump_release.py
twaugh: Would like to see a test case where `base_release` is e.g. `'42.1'`, or a comment explaining why it's not valid to test it.

twaugh:
(Also needs to be re-based against master.)

atomic_reactor/plugins/pre_resolve_module_compose.py
owtaylor: Sorry, missed that - added, will be in my next push.

tests/plugins/test_bump_release.py
owtaylor: OK, will add:

```
        # No interpretation of the base release when appending - just treated as string                                                                    
        ('42.1', ['42.2'], '42.1.1'),
        ('42.1', ['42.1.1'], '42.1.2'),
```

atomic_reactor/plugins/prepub_flatpak_create_oci.py
owtaylor: I have no idea at this point. I might have meant that I thought that the bundle filename (outfile) should contain some version information, but it doesn't actually matter since the tarball uploaded to Koji ends up as docker-image-{id}... anyways. [Something that is on my list of things to fix after this PR lands.] Will just remove.

lcarva:
Updated release notes draft:
> - Support for building Flatpaks has been added

PR #749, opened 2017-07-25
post_tag_from_config: take tag_suffixes from osbs-client

mlangsdorf:
add an optional tag_suffixes plugin parameter to post_tag_from_config. If
it is passed, do not get tag_suffixes from the git repository. Instead,
parse the tags dictionary for 'unique' and 'primary', which are
lists of strings of tag_suffixes. Append the unique tag_suffixes verbatim
to the component name and apply the resulting tags; expand the primary 
tag_suffixes, append them to the component name, and apply them. Return the 
list of tags applied.

### DIFF
--- a/atomic_reactor/plugins/post_tag_from_config.py
+++ b/atomic_reactor/plugins/post_tag_from_config.py
@@ -8,11 +8,10 @@ of the BSD license. See the LICENSE file for details.
 
 import os
 import re
-import string
 
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.constants import INSPECT_CONFIG, TAG_NAME_REGEX
-from atomic_reactor.util import get_preferred_label_key, df_parser
+from atomic_reactor.util import get_preferred_label_key, df_parser, LabelFormatter
 
 
 class TagFromConfigPlugin(PostBuildPlugin):
@@ -49,15 +48,6 @@ class TagFromConfigPlugin(PostBuildPlugin):
         self.tag_suffixes = tag_suffixes
 
     def parse_and_add_tags(self):
-        class MyFormatter(string.Formatter):
-            """
-            using this because str.format can't handle keys with dots and dashes
-            which are included in some of the labels, such as
-            'authoritative-source-url', 'com.redhat.component', etc
-            """
-            def get_field(self, field_name, args, kwargs):
-                return (self.get_value(field_name, args, kwargs), field_name)
-
         tags = []
         name = self.get_component_name()
         labels = df_parser(self.workflow.builder.df_path, workflow=self.workflow,
@@ -69,7 +59,7 @@ class TagFromConfigPlugin(PostBuildPlugin):
             tags.append(tag)
 
         for tag_suffix in self.tag_suffixes.get('primary', []):
-            p_suffix = MyFormatter().vformat(tag_suffix, [], labels)
+            p_suffix = LabelFormatter().vformat(tag_suffix, [], labels)
             p_tag = '{}:{}'.format(name, p_suffix)
             self.log.debug('Using additional primary tag %s', p_tag)
             self.workflow.tag_conf.add_primary_image(p_tag)
--- a/atomic_reactor/plugins/post_tag_from_config.py
+++ b/atomic_reactor/plugins/post_tag_from_config.py
@@ -8,10 +8,11 @@ of the BSD license. See the LICENSE file for details.
 
 import os
 import re
+import string
 
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.constants import INSPECT_CONFIG, TAG_NAME_REGEX
-from atomic_reactor.util import get_preferred_label_key
+from atomic_reactor.util import get_preferred_label_key, df_parser
 
 
 class TagFromConfigPlugin(PostBuildPlugin):
@@ -43,7 +44,40 @@ class TagFromConfigPlugin(PostBuildPlugin):
 
     TAGS_FILENAME = 'additional-tags'
 
-    def get_tags(self):
+    def __init__(self, tasker, workflow, tag_suffixes=None):
+        super(TagFromConfigPlugin, self).__init__(tasker, workflow)
+        self.tag_suffixes = tag_suffixes
+
+    def parse_and_add_tags(self):
+        class MyFormatter(string.Formatter):
+            """
+            using this because str.format can't handle keys with dots and dashes
+            which are included in some of the labels, such as
+            'authoritative-source-url', 'com.redhat.component', etc
+            """
+            def get_field(self, field_name, args, kwargs):
+                return (self.get_value(field_name, args, kwargs), field_name)
+
+        tags = []
+        name = self.get_component_name()
+        labels = df_parser(self.workflow.builder.df_path, workflow=self.workflow,
+                           env_replace=True).labels
+        for tag_suffix in self.tag_suffixes.get('unique', []):
+            tag = '{}:{}'.format(name, tag_suffix)
+            self.log.debug('Using additional unique tag %s', tag)
+            self.workflow.tag_conf.add_unique_image(tag)
+            tags.append(tag)
+
+        for tag_suffix in self.tag_suffixes.get('primary', []):
+            p_suffix = MyFormatter().vformat(tag_suffix, [], labels)
+            p_tag = '{}:{}'.format(name, p_suffix)
+            self.log.debug('Using additional primary tag %s', p_tag)
+            self.workflow.tag_conf.add_primary_image(p_tag)
+            tags.append(p_tag)
+
+        return tags
+
+    def get_and_add_tags(self):
         tags = []
 
         df_dir = self.workflow.source.get_dockerfile_path()[1]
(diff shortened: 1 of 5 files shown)


atomic_reactor/plugins/post_tag_from_config.py
```diff
@@ -43,7 +44,51 @@ class TagFromConfigPlugin(PostBuildPlugin):
...
+            def get_field(self, field_name, args, kwargs):
+                return (self.get_value(field_name, args, kwargs), field_name)
+
+        tags = []
+        labels = df_parser(self.workflow.builder.df_path, workflow=self.workflow,
+                           env_replace=True).labels
+
+        for tag in self.tags.get('unique', []):
+            try:
+                u_tag = MyFormatter().vformat(tag, [], labels)
```
twaugh: We don't want to perform substitution on the unique tags. I think lines 66-74 should be removed.

atomic_reactor/plugins/post_tag_from_config.py
```diff
@@ -43,7 +44,51 @@ class TagFromConfigPlugin(PostBuildPlugin):
...
+                continue
+            if u_tag != tag:
+                self.log.error('%s submitted as unique tag', tag)
+                continue
+            self.log.debug('Using additional unique tag %s', u_tag)
+            self.workflow.tag_conf.add_unique_image(tag)
+            tags.append(tag)
+
+        if self.tags.get('primary', []):
+            for tag in self.tags.get('primary', []):
```
twaugh: Don't need the `if` on line 78. If there is no 'primary' key, or it has an empty list as its value, the `for` loop will have nothing to do anyway.

atomic_reactor/plugins/post_tag_from_config.py
```diff
@@ -43,7 +44,51 @@ class TagFromConfigPlugin(PostBuildPlugin):
...
+                try:
+                    p_tag = MyFormatter().vformat(tag, [], labels)
+                except KeyError:
+                    self.log.error('%s submitted as primary tag with no matching label', tag)
+                    continue
```
twaugh: I wonder if it's better to just let this exception propagate rather than discarding the tag. If we have a bad tag here it means something is badly wrong with either the ATOMIC_REACTOR_PLUGINS config (bad tag template) or the Dockerfile (missing essential label).

atomic_reactor/plugins/post_tag_from_config.py
```diff
@@ -83,15 +137,9 @@ def get_component_name(self):
...
-
-        if tags:
-            name = self.get_component_name()
-            for i, tag_suffix in enumerate(tags):
-                tag = '{}:{}'.format(name, tag_suffix)
-                self.log.debug('Using additional tag: %s', tag)
-                self.workflow.tag_conf.add_primary_image(tag)
-                # Store modified name.
-                tags[i] = tag
+        if self.tags:
```
twaugh: Nit-pick, but I think this ought to be `if self.tags is not None`. If `tags=[]` is supplied, we should add no tags.

atomic_reactor/plugins/post_tag_from_config.py
mlangsdorf: I'm testing whether the unique tab can be expanded via substitution (because if it can be, it needs to be discarded). Is there a better way to detect improper unique tags?

twaugh: Just letting tag_and_push try to use those tags exactly as they are will eventually cause a failure:
```
$ docker tag docker.io/fedora:latest 'anything:{braced}'
Error parsing reference: "anything:{braced}" is not a valid repository/tag
```

mlangsdorf: self.workflow.tag_conf.add_unique_image('{version}') is a valid operation that does not raise any exceptions, as far as I can tell.

twaugh: That's correct. It just notes which tags should be applied, later, when the tag_and_push plugin runs. It is during the execution of that tag_and_push plugin that I expect '{version}' to cause a failure.

mlangsdorf:
I've got most of the requested fixes, but I'm having problems with the from expansion in my tests.
```
DF_CONTENT_LABELS = '''\
FROM fedora
LABEL "version"="version_value"
LABEL "release"="$parentenv"'''

@pytest.mark.parametrize(('unique_tags', 'primary_tags', 'expected'), [  # noqa
    (['foo', 'bar'], ['{version}-{release}'], ['foo', 'bar', 'version_value-7.4.1']),
])
def test_tag_parse(tmpdir, docker_tasker, unique_tags, primary_tags, expected):
    df = df_parser(str(tmpdir))
    df.content = DF_CONTENT_LABELS

    workflow = mock_workflow(tmpdir)
    setattr(workflow.builder, 'df_path', df.dockerfile_path)
    workflow.built_image_inspect = {
        INSPECT_CONFIG: {
            'Labels': {'Name': 'get_tags'},
            'Env': {'parentenv', '7.4.1'}
        }
    }
```

I think that's the correct format for DF_CONTENT_LABELS but python disagrees with me. But I can't figure out what else it should be:
LABEL label="release $parentenv"
isn't right either.

atomic_reactor/plugins/post_tag_from_config.py
mlangsdorf: fixed

atomic_reactor/plugins/post_tag_from_config.py
mlangsdorf: fixed

atomic_reactor/plugins/post_tag_from_config.py
mlangsdorf: fixed

atomic_reactor/plugins/post_tag_from_config.py
mlangsdorf: Okay.
Fixed.

atomic_reactor/plugins/post_tag_from_config.py
```diff
@@ -43,7 +44,38 @@ class TagFromConfigPlugin(PostBuildPlugin):
 
     TAGS_FILENAME = 'additional-tags'
 
-    def get_tags(self):
+    def __init__(self, tasker, workflow, tags=None):
+        super(TagFromConfigPlugin, self).__init__(tasker, workflow)
+        self.tags = tags
+
+    def parse_and_add_tags(self):
+        class MyFormatter(string.Formatter):
```
lcarva: Can we do a bit of refactoring, and instead of copying this class, move it to a common place? Maybe utils.py, or a new formatters.py. Let's also call it something useful, like `LabelFormater`.

atomic_reactor/plugins/post_tag_from_config.py
```diff
@@ -43,7 +44,38 @@ class TagFromConfigPlugin(PostBuildPlugin):
...
+            def get_field(self, field_name, args, kwargs):
+                return (self.get_value(field_name, args, kwargs), field_name)
+
+        tags = []
+        labels = df_parser(self.workflow.builder.df_path, workflow=self.workflow,
+                           env_replace=True).labels
+
+        for tag in self.tags.get('unique', []):
+            self.log.debug('Using additional unique tag %s', tag)
+            self.workflow.tag_conf.add_unique_image(tag)
```
lcarva: The term `tag` sometimes is a little ambiguous. In some cases, it refers to just the thing after `':'` (for instance, `'latest'`). In other cases, it includes the repo, `'spam/bacon:latest'`.
Previously, this plugin made the distinction by using `'tag_suffix'` terminology.

So... let's define this first, do we want the `tags` plugin param to contain tags, or tag-suffixes? I'm thinking tag-suffixes is safer. (this would require pre-pending the component to each tag as it previously did).

mlangsdorf: from osbs-4135 in the acceptance criteria: "Verify unique tags are applied verbatim without substitution"
If you want to go back to pre-pending component tags, then you need to argue the design there.

lcarva: My comment is not to substitute values in the tag provided, but to actually add a prefix to it.
This also applies to primary tags.

twaugh: @lcarva: "let's define  this first" -- yes, we're talking about tag suffixes i.e. the part after ":" in the image name.
Sorry, I didn't remember that tag_conf.add_*_image() needs *image* names, not just the part after ":". You're quite right, we need to add those in too.

twaugh: Updated the requirements for this story.

atomic_reactor/plugins/post_tag_from_config.py
twaugh: +1

twaugh:
Note: we want to define 'Env' for the *base* (parent) image, not for the *built* image.

atomic_reactor/plugins/post_tag_from_config.py
mlangsdorf: I'll do that as part of this story, but first let me finish the test logic.

atomic_reactor/plugins/post_tag_from_config.py
mlangsdorf: Okay, fixed to match the new requirements.

atomic_reactor/plugins/post_tag_from_config.py
mlangsdorf: Fixed.

mlangsdorf:
After some hammering, the pass through from Env test case has been added successfully.

twaugh:
Now that we've changed the parameter name to tag_suffixes, let's make sure the commit message agrees.

mlangsdorf:
I updated the commit message to the new text.

PR #751, opened 2017-07-26
Rework travis to make it run in different Fedora/CentOS containers

vrutkovs:
* Travis failures can be reproduced locally: 
`sudo env OS=centos OS_VERSION=7 ./test.sh` or 
`sudo env OS=fedora OS_VERSION=26 PYTHON_VERSION=3 ./test.sh`. 
By default it would run tests on CentOS 7
* Tests would run against actual packages in distributions, not latest from pypi (with exception of osbs-client and dockerfile-parse)

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+set -eux
+
+# Prepare env vars
+OS=${OS:="centos"}
+OS_VERSION=${OS_VERSION:="7"}
+PYTHON_VERSION=${PYTHON_VERSION:="2"}
+IMAGE="$OS:$OS_VERSION"
+# Pull fedora images from registry.fedoraproject.org
+if [[ $OS == "fedora" ]]; then
+  IMAGE="registry.fedoraproject.org/$IMAGE"
+fi
+
+CONTAINER_NAME="atomic-reactor-$OS-$OS_VERSION-py$PYTHON_VERSION"
+RUN="docker exec -ti $CONTAINER_NAME"
+if [[ $OS == "fedora" ]]; then
+  if [[ $OS_VERSION == 25 && $PYTHON_VERSION == 2 ]]; then PIP_PKG="python-pip"; else PIP_PKG="python$PYTHON_VERSION-pip"; fi
+  PIP="pip$PYTHON_VERSION"
+  PKG="dnf"
+  # F25 has outdated flatpak, don't run flatpak tests on it
+  if [[ $OS_VERSION == 25 ]]; then PKG_EXTRA="dnf-plugins-core desktop-file-utils"; else PKG_EXTRA="dnf-plugins-core desktop-file-utils flatpak ostree"; fi
+  BUILDDEP="dnf builddep"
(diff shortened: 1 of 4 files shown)


test.sh
```diff
@@ -0,0 +1,73 @@
...
+# Install package
+$RUN $PKG install -y $PIP_PKG
+if [[ $PYTHON_VERSION == 3 && $OS_VERSION == rawhide ]]; then
+  # https://fedoraproject.org/wiki/Changes/Making_sudo_pip_safe
+  $RUN mkdir -p /usr/local/lib/python3.6/site-packages/
+fi
+
+# Install osbs-client and docker_squash
+$RUN $PKG install -y osbs-client
+$RUN $PIP install --upgrade --no-deps --force-reinstall git+https://github.com/projectatomic/osbs-client
```
lcarva: Why do we need to install osbs-client twice?
Seems to me that the `requirements.txt` file is not being processed at all anymore.

vrutkovs: This is required to a) setup all deps for osbs-client from packages and b) update osbs-client from latest master without attempting to install deps from PIP.

This is done mostly for docker-squash, which tries to pull latest docker-py and fails as there is already one installed from RPM. Other deps might want gcc etc.

lcarva: Right, but it means that if we add a new dependency to osbs-client, it won't be installed until it's actually released for the underlying OS.

vrutkovs: Yes, the `test.sh` packages in `PKG_EXTRA` with specific packages names for CentOS/Fedora version, as installing deps from pip would take too much time (most of these require `gcc` etc.)

lcarva: @twaugh, thoughts?

vrutkovs: The way to resolve this:
* pull latest code from osbs-client master branch
* install the deps from the spec
* install fresh osbs-client via `pip --upgrade --no-deps --force-reinstall`

This would ensure the latest deps are recorded in the spec of osbs-client

twaugh: @vrutkovs, that looks like a good way to do it IMHO.

vrutkovs: Implemented that, PTAL

tests/plugins/test_resolve_module_compose.py
```diff
@@ -108,7 +108,11 @@ def test_resolve_module_compose(tmpdir, docker_tasker, specify_version):
     def handle_composes_post(request):
         assert request.headers['OIDC_access_token'] == 'green_eggs_and_ham'
 
-        body_json = json.loads(request.body.decode())
+        if isinstance(request.body, str):
```
twaugh: `isinstance(_, str)` is usually unsafe because of the differences between Python versions. Is there any other way of phrasing this? The example usage says it should be fine to give directly to `json.loads()`: https://github.com/getsentry/responses/blob/master/README.rst

twaugh: Maybe use `six.text_type`?

vrutkovs: Good idea, `six.text_type` worked

PR #754, opened 2017-07-27
get_manifest_digests: skip version if registry/pulp didn't return required headers

vrutkovs:
Pulp won't return headers for schema 1 images, so get_manifest_digests should 
check those and not throw exception if throw_exception param is False

### DIFF
--- a/atomic_reactor/plugins/post_pulp_pull.py
+++ b/atomic_reactor/plugins/post_pulp_pull.py
@@ -72,7 +72,8 @@ class PulpPullPlugin(PostBuildPlugin):
             if plugin['name'] is PLUGIN_PULP_PUSH_KEY:
                 media_types.append('application/json')
 
-        digests = get_manifest_digests(pullspec, registry.uri, self.insecure, self.secret)
+        digests = get_manifest_digests(pullspec, registry.uri, self.insecure, self.secret,
+                                       require_digest=False)
         if digests.v2:
             self.log.info("V2 schema 2 digest found, returning %s", self.workflow.builder.image_id)
             media_types.append('application/vnd.docker.distribution.manifest.v2+json')
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -31,7 +31,8 @@ from dockerfile_parse import DockerfileParser
 from pkg_resources import resource_stream
 
 from importlib import import_module
-
+from requests.utils import guess_json_utf
+from time import sleep
 
 logger = logging.getLogger(__name__)
 
@@ -705,7 +706,7 @@ def query_registry(image, registry, digest=None, insecure=False, dockercfg_path=
 
 
 def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
-                         versions=('v1', 'v2')):
+                         versions=('v1', 'v2'), require_digest=True):
     """Return manifest digest for image.
 
     :param image: ImageName, the remote image to inspect
@@ -714,6 +715,8 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
     :param insecure: bool, when True registry's cert is not verified
     :param dockercfg_path: str, dirname of .dockercfg location
     :param versions: tuple, which manifest schema versions to fetch digest
+    :param require_digest: bool, when True exception is thrown if no digest is
+                                 set in the headers.
 
     :return: dict, versions mapped to their digest
     """
@@ -721,10 +724,16 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
     for version in versions:
         media_type = get_manifest_media_type(version)
         headers = {'Accept': media_type}
-        response = query_registry(
-            image, registry, digest=None,
-            insecure=insecure, dockercfg_path=dockercfg_path,
-            version=version)
+
+        for attempts in range(10):
+            response = query_registry(
+                image, registry, digest=None,
+                insecure=insecure, dockercfg_path=dockercfg_path,
+                version=version)
+            if response.status_code == 404:
+                sleep(1)
+            else:
+                break
 
         # If the registry has a v2 manifest that can't be converted into a v1
         # manifest, the registry fails with status=400, and a error code of
@@ -736,14 +745,41 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
 
         response.raise_for_status()
 
+        received_media_type = None
+        try:
+            received_media_type = response.headers['Content-Type']
+        except KeyError:
+            # Guess content_type from contents
+            try:
+                encoding = guess_json_utf(response.content)
+                manifest = json.loads(response.content.decode(encoding))
+                received_media_type = manifest['mediaType']
+            except (ValueError,  # not valid JSON
+                    KeyError):   # no mediaType key
+                logger.warning("Unable to fetch media type: neither Content-Type header "
+                               "nor mediaType in output was found")
+
+        if not received_media_type:
+            continue
+
         # Only compare prefix as response may use +prettyjws suffix
         # which is the case for signed manifest
-        response_h_prefix = response.headers['Content-Type'].rsplit('+', 1)[0]
+        response_h_prefix = received_media_type.rsplit('+', 1)[0]
         request_h_prefix = media_type.rsplit('+', 1)[0]
         if response_h_prefix != request_h_prefix:
             logger.debug('request headers: %s', headers)
             logger.debug('response headers: %s', response.headers)
-            logger.warning('Unable to fetch digest for %s', media_type)
+            logger.warning('Received media type %s mismatches the expected %s',
+                           media_type, received_media_type)
+            continue
+
+        # set it to truthy value so that koji_import would know pulp supports these digests
+        digests[version] = True
+        logger.warning('received_media_type=%s', received_media_type)
+
+        if not response.headers.get('Docker-Content-Digest'):
+            logger.warning('Unable to fetch digest for %s, no Docker-Content-Digest header',
+                           media_type)
             continue
 
         digests[version] = response.headers['Docker-Content-Digest']
@@ -752,7 +788,7 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
         logger.debug('Image %s:%s has %s manifest digest: %s',
                      context, tag, version, digests[version])
 
-    if not digests:
+    if not digests and require_digest:
         raise RuntimeError('No digests found for {}'.format(image))
 
     return ManifestDigest(**digests)
--- a/tests/plugins/test_pulp_pull.py
+++ b/tests/plugins/test_pulp_pull.py
@@ -15,6 +15,7 @@ from docker.errors import NotFound
 from flexmock import flexmock
 import pytest
 import requests
+import json
 
 DIGEST_V1 = 'sha256:7de72140ec27a911d3f88d60335f08d6530a4af136f7beab47797a196e840afd'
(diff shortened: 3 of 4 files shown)


twaugh:
If it tried to parse the response content as UTF-x-encoded JSON, we'd be able to read the mediaType key to get the media type. That's all pulp_pull wants anyway, not the digest.

```
try:
    encoding = guess_json_utf(response.content)
    manifest = json.loads(response.content.decode(encoding))
    media_type = manifest['mediaType']
except (ValueError,  # not valid JSON
        KeyError):   # no mediaType key
    ...
```

atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -64,7 +64,8 @@ def run(self):
         pullspec = image.copy()
         pullspec.registry = registry.uri  # the image on Crane
 
-        digests = get_manifest_digests(pullspec, registry.uri, self.insecure, self.secret)
+        digests = get_manifest_digests(pullspec, registry.uri, self.insecure, self.secret,
+                                       throw_error=False)
```
twaugh: What about just catching `RuntimeError` (or even `Exception`) here?

vrutkovs: In this case we won't return which digests are supported by registry/pulp, as it sets `digests[version]=True`

atomic_reactor/util.py
```diff
@@ -738,12 +739,33 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
...
+            try:
+                encoding = guess_json_utf(response.content)
+                manifest = json.loads(response.content.decode(encoding))
+                received_media_type = manifest['mediaType']
+            except (ValueError,  # not valid JSON
+                    KeyError):   # no mediaType key
+                logger.warning("Unable to fetch media type: neither Content-Type header "
+                               "nor mediaType in output was found")
+
+        if received_media_type:
```
lcarva: Don't we want: `if not received_media_type: continue` ?

atomic_reactor/util.py
```diff
@@ -738,12 +739,33 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
...
+            # set it to truthy value so that koji_import would know pulp supports these digests
+            digests[version] = True
+            response_h_prefix = received_media_type.rsplit('+', 1)[0]
+            request_h_prefix = media_type.rsplit('+', 1)[0]
+            if response_h_prefix != request_h_prefix:
```
lcarva: Are they ever going to be the same in the case media type was guessed?

atomic_reactor/plugins/post_pulp_pull.py
twaugh: Not sure what you mean. Couldn't you just `pass` if you catch `RuntimeError`, and have the same effect as passing through a `throw_error` parameter?

atomic_reactor/util.py
```diff
@@ -714,17 +715,24 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
...
+
+        for attempts in range(10):
+            response = query_registry(
+                image, registry, digest=None,
+                insecure=insecure, dockercfg_path=dockercfg_path,
+                version=version)
+            if response.status_code == 404:
+                sleep(1)
+            else:
+                break
```
twaugh: Oh, right, retries.

This looks a little wonky though. Shouldn't query_registry() do this using a requests.Session? (I'd be happy for this to be filed as something to get to later.)

atomic_reactor/util.py
twaugh: Why wouldn't they be?

atomic_reactor/util.py
```diff
@@ -738,21 +746,51 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
...
+                continue
+
+        try
+            content_digest = response.headers['Docker-Content-Digest']:
+        except KeyError:
+            # Guess content_type from contents
+            try:
+                encoding = guess_json_utf(response.content)
+                manifest = json.loads(response.content.decode(encoding))
+                content_digest = manifest['config']['digest']
```
twaugh: No, that's something different. If we aren't told the Docker-Content-Digest, we'd have to calculate it ourselves to know it.

Luckily, we don't need to know if in that case. (Real registries always return it.)

atomic_reactor/util.py
vrutkovs: I'm thinking it should be updated for all plugins automatically

atomic_reactor/util.py
vrutkovs: Oh, okay. In that case its useless to use `get_manifest_digests` in pulp_pull - Pulp doesn't set this header

atomic_reactor/util.py
vrutkovs: Probably yes. I can't think of a situation where digest is in headers and media type isn't. Updated

atomic_reactor/util.py
lcarva: Because the media type in the request is something like: `'application/vnd.docker.distribution.manifest.v1+json'`
and we're comparing the segment before `'+'`. 
Will `requests` guess the media_type in such way that it is prefixed with `'application/vnd.docker.distribution.manifest.v1'` ?

atomic_reactor/util.py
```diff
@@ -738,21 +746,44 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
...
+            except (ValueError,  # not valid JSON
+                    KeyError):   # no mediaType key
+                logger.warning("Unable to fetch media type: neither Content-Type header "
+                               "nor mediaType in output was found")
+
+        if not received_media_type:
+            continue
+
+        # set it to truthy value so that koji_import would know pulp supports these digests
+        digests[version] = True
```
lcarva: Shouldn't this only be set if the media type was guessed?

vrutkovs: It is being set only after `received_media_type` is True

lcarva: Right, but this line exists only to handle the case where the digest was guessed, right?

vrutkovs: It doesn't matter - if the digest would be fetched later on from header it would be overwritten

atomic_reactor/util.py
twaugh: I've made a note in JIRA to revisit it later.

atomic_reactor/util.py
```diff
@@ -738,21 +747,45 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
...
+        except KeyError:
+            # Guess content_type from contents
+            try:
+                encoding = guess_json_utf(response.content)
+                manifest = json.loads(response.content.decode(encoding))
+                received_media_type = manifest['mediaType']
+            except (ValueError,  # not valid JSON
+                    KeyError):   # no mediaType key
+                logger.warning("Unable to fetch media type: neither Content-Type header "
+                               "nor mediaType in output was found")
```
twaugh: We need unit test coverage on this `except KeyError` block.

atomic_reactor/util.py
```diff
@@ -738,21 +747,45 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
...
+
+        if not received_media_type:
+            continue
```
twaugh: This `continue` doesn't get executed during unit testing.

atomic_reactor/util.py
twaugh: That's odd, this still doesn't seem to get coverage.

tests/plugins/test_pulp_pull.py
```diff
@@ -89,19 +91,30 @@ def workflow(self):
...
               headers={
                 'Content-Type': 'application/vnd.docker.distribution.manifest.v2+json',
                 'Docker-Content-Digest': DIGEST_V2
               }))
 
+    config_response_config_v2_no_headers = requests.Response()
+    (flexmock(config_response_config_v2_no_headers,
+              raise_for_status=lambda: None,
+              status_code=requests.codes.ok,
+              _content=bytes(json.dumps(get_response_config_json(media_type_v2))),
```
twaugh: Maybe `_content=json.dumps(...).encode('utf-8')`?

vrutkovs: Right, that worked, thanks!

atomic_reactor/util.py
vrutkovs: Its covered now, added a special case to `test_pulp_pull` to send a broken response

vrutkovs:
Test cases were updated, PTAL

atomic_reactor/util.py
owtaylor: In some cases the docker/distribution code already returns 404 if it can't satisfy the Accept header (manifest list where an appropriate image can't be extracted, for example), and the current draft of OCI support extends that behavior. This looks like a good recipe for slowing builds down by 10 or 20 seconds.

Is this some sort of race condition workaround? A workaround for a bug in some component? In general, there should be never any reason to retry a 404 - it's not a '503 Service Unavailable' or something.

tests/test_util.py
```diff
@@ -402,25 +421,38 @@ def request_callback(request):
 
                 digest = 'v1-converted-digest'
 
-        headers = {
-            'Content-Type': '{}+jsonish'.format(media_type_prefix),
-            'Docker-Content-Digest': digest
-        }
+        headers = {}
+        if digest_is_v1 or has_content_type_header:
```
owtaylor: The digest_is_v1 check here is extremely hard to understand without a comment - my only guess is that some external component (pulp?) has the missing header behavior for v2 digests not v1 digests? If there's some parameterized combinations you want to skip, wouldn't it be better to bypass the whole test at the top?

tests/test_util.py
```diff
@@ -402,25 +421,38 @@ def request_callback(request):
...
+        with pytest.raises(RuntimeError):
+            get_manifest_digests(**kwargs)
+        return
+    else:
+        actual_digests = get_manifest_digests(**kwargs)
 
     if digest_is_v1:
         assert actual_digests.v1 == 'v1-digest'
-        assert actual_digests.v2 is None
+        assert actual_digests.v2 in [True, None]
```
owtaylor: Why would v2 by True here? (tests pass fine if the check is just against None)

tests/test_util.py
vrutkovs: Yes, Pulp is not sending schema v1 digest in the output as this digest is cannot be used for pulling later on - https://github.com/pulp/pulp_docker/blob/master/plugins/pulp_docker/plugins/registry.py#L384. A correct manifest would be returned however, so its possible to fetch mediaType from the manifest even if the header is missing.

tests/test_util.py
vrutkovs: `digests.v2` would return `None` if registry/pulp don't support v2 schema 2 images at all. It would return `True` if it does support, but the image could not be converted, and finally it would return the digest string if the digest is in the headers/manifest.

The tests should be rewritten to have this tested for each specific case though

owtaylor: > It would return True if it does support, but the image could not be converted

As far as I understand what you implemented in this PR, True means that the registry replied with something but didn't provide a digest. If the registry has a 'v1' manifest, it doesn't make any sense that it would serve up a v2 manifest without a digest. Let's take this to PR #764 in any case.

tests/test_util.py
owtaylor: The code you are pointing to seems to be the code to import docker images into pulp. Wouldn't the relevant code be what pulp-crane does? (And pulp-crane *could* return the digest since the v2 protocol specifies that the digest is returned with the redirect and not the final content.)

PR #756, opened 2017-07-27
group_manifests: create minimal plugin for group manifests

mlangsdorf:
group_manifests: create plugin for group manifests

Add a group_manifests plugin. For now, it fails if asked to create
a group of manifests.

If asked to create a single manifest, it pulls the x86_64 image manifest
from a repository passed as an argument. It then re-uploads the image
manifest to a new location for each image in the tag_conf. It returns the
image manifest.

### DIFF
--- a/atomic_reactor/plugins/post_group_manifests.py
+++ b/atomic_reactor/plugins/post_group_manifests.py
@@ -1,0 +1,148 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+
+get the image manifest lists from the worker builders. If possible, group them together
+and return them. if not, return the x86_64/amd64 image manifest instead after re-uploading
+it for all existing image tags.
+"""
+
+
+from __future__ import unicode_literals
+import requests
+import requests.auth
+
+from six.moves.urllib.parse import urlparse
+
+from atomic_reactor.plugin import PostBuildPlugin, PluginFailedException
+from atomic_reactor.util import Dockercfg
+
+
+class GroupManifestsPlugin(PostBuildPlugin):
+    key = 'group_manifests'
+    is_allowed_to_fail = False
+
+    def __init__(self, tasker, workflow, registries, group=True, goarch=None):
+        """
+        constructor
+
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param registries: dict, keys are docker registries, values are dicts containing
+                           per-registry parameters.
+                           Params:
+                            * "secret" optional string - path to the secret, which stores
+                              login and password for remote registry
+        :param group: bool, if true, create a manifest list; otherwise only add tags to
+                      amd64 image manifest
+        :param goarch: dict, keys are platform, values are go language platform names
+        """
+        # call parent constructor
+        super(GroupManifestsPlugin, self).__init__(tasker, workflow)
+        self.group = group
+        self.goarch = goarch or {}
+        self.registries = registries
+        self.worker_registries = {}
+
+    def get_worker_manifest(self, worker_data):
+        worker_digests = worker_data['digests']
+        worker_manifest = []
+
+        msg = "worker_registries {0}".format(self.worker_registries)
+        self.log.debug(msg)
+
+        for registry, registry_conf in self.registries.items():
+            if registry_conf.get('version') == 'v1':
+                continue
+
+            if not registry.startswith('http://') and not registry.startswith('https://'):
+                registry = 'https://' + registry
+
+            registry_noschema = urlparse(registry).netloc
+            self.log.debug("evaluating registry %s", registry_noschema)
+
+            insecure = registry_conf.get('insecure', False)
+            auth = None
+            secret_path = registry_conf.get('secret')
+            if secret_path:
+                self.log.debug("registry %s secret %s", registry_noschema, secret_path)
+                dockercfg = Dockercfg(secret_path).get_credentials(registry_noschema)
+                try:
+                    username = dockercfg['username']
+                    password = dockercfg['password']
+                except KeyError:
+                    self.log.error("credentials for registry %s not found in %s",
+                                   registry_noschema, secret_path)
+                else:
+                    self.log.debug("found user %s for registry %s", username, registry_noschema)
+                    auth = requests.auth.HTTPBasicAuth(username, password)
+
+            if registry_noschema in self.worker_registries:
+                self.log.debug("getting manifests from %s", registry_noschema)
+                digest = worker_digests[0]['digest']
+                repo = worker_digests[0]['repository']
+
+                # get a v2 schemav2 response for now
+                v2schema2 = 'application/vnd.docker.distribution.manifest.v2+json'
+                headers = {'accept': v2schema2}
+                kwargs = {'verify': not insecure, 'headers': headers, 'auth': auth}
+
+                url = '{0}/v2/{1}/manifests/{2}'.format(registry, repo, digest)
+                self.log.debug("attempting get from %s", url)
+                response = requests.get(url, **kwargs)
+
+                image_manifest = response.json()
+
+                if image_manifest['schemaVersion'] == '1':
+                    msg = 'invalid schema from {0}'.format(url)
+                    raise PluginFailedException(msg)
+
+                headers = {'Content-Type': v2schema2}
+                kwargs = {'verify': not insecure, 'headers': headers, 'auth': auth}
+
+                for image in self.workflow.tag_conf.images:
+                    image_tag = image.to_str(registry=False).split(':')[1]
+                    url = '{0}/v2/{1}/manifests/{2}'.format(registry, repo, image_tag)
+                    self.log.debug("for image_tag %s, putting at %s", image_tag, url)
+                    response = requests.put(url, json=image_manifest, **kwargs)
+
+                    if not response.ok:
+                        msg = "PUT failed: {0},\n manifest was: {1}".format(response.json(),
+                                                                            image_manifest)
+                        self.log.error(msg)
+                    response.raise_for_status()
+
+                worker_manifest.append(image_manifest)
+                self.log.debug("appending an image_manifest")
+                break
+
+        return worker_manifest
+
+    def run(self):
+        if self.group:
+            raise NotImplementedError('group=True is not supported in group_manifests')
+        grouped_manifests = []
+
+        valid = False
+        all_annotations = self.workflow.build_result.annotations['worker-builds']
+        for plat, annotation in all_annotations.items():
+            digests = annotation['digests']
+            for digest in digests:
+                registry = digest['registry']
+                self.worker_registries.setdefault(registry, [])
+                self.worker_registries[registry].append(registry)
+
+        for platform in all_annotations:
+            if self.goarch.get(platform, platform) == 'amd64':
+                valid = True
+                grouped_manifests = self.get_worker_manifest(all_annotations[platform])
+                break
+
+        if valid:
+            self.log.debug("found an x86_64 platform and grouped its manifest")
+            return grouped_manifests
+        else:
+            raise ValueError('failed to find an x86_64 platform')
--- a/tests/plugins/test_group_manifests.py
+++ b/tests/plugins/test_group_manifests.py
@@ -1,0 +1,222 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import print_function, unicode_literals
+import pytest
+import json
+import responses
+from copy import deepcopy
+from tempfile import mkdtemp
+import os
+
+from tests.constants import SOURCE, INPUT_IMAGE, MOCK, DOCKER0_REGISTRY
+
+from atomic_reactor.core import DockerTasker
+from atomic_reactor.build import BuildResult
+from atomic_reactor.plugin import PostBuildPluginsRunner, PluginFailedException
+from atomic_reactor.inner import DockerBuildWorkflow, TagConf
+from atomic_reactor.util import ImageName
+from atomic_reactor.plugins.post_group_manifests import GroupManifestsPlugin
+
+if MOCK:
+    from tests.docker_mock import mock_docker
+
+
+DIGEST1 = 'sha256:28b64a8b29fd2723703bb17acf907cd66898440270e536992b937899a4647414'
+DIGEST2 = 'sha256:0000000000000000000000000000000000000000000000000000000000000000'
+
+
+class Y(object):
+    pass
+
+
+class X(object):
+    image_id = INPUT_IMAGE
+    source = Y()
+    source.dockerfile_path = None
+    source.path = None
+    base_image = ImageName(repo="qwe", tag="asd")
+
+
+X86_DIGESTS = [
+    {
+        'digest': 'sha256:worker-build-x86_64-digest',
+        'tag': 'worker-build-x86_64-latest',
+        'registry': DOCKER0_REGISTRY,
+        'repository': 'worker-build-x86_64-repository',
+    },
+]
+X86_ANNOTATIONS = {
+    'build': {
+        'build-name': 'worker-build-x86_64',
+        'cluster-url': 'https://worker_x86_64.com/',
+        'namespace': 'worker_x86_64_namespace'
+    },
+    'digests': X86_DIGESTS,
+    'plugins-metadata': {},
+}
+PPC_DIGESTS = [
+    {
+        'digest': 'sha256:worker-build-ppc64le-digest',
+        'tag': 'worker-build-ppc64le-latest',
+        'registry': 'worker-build-ppc64le-registry',
+        'repository': 'worker-build-ppc64le-repository',
+    },
+]
+PPC_ANNOTATIONS = {
+    'build': {
+        'build-name': 'worker-build-ppc64le',
+        'cluster-url': 'https://worker_ppc64le.com/',
+        'namespace': 'worker_ppc64le_namespace'
+    },
+    'digests': PPC_DIGESTS,
+    'plugins-metadata': {}
+}
+
+BUILD_ANNOTATIONS = {
+        'worker-builds': {
+        },
+        'repositories': {
+            'unique': [
+                'worker-build-ppc64le-unique',
+                'worker-build-x86_64-unique',
+            ],
+            'primary': [
+                'worker-build-ppc64le-primary',
+                'worker-build-x86_64-primary',
+            ],
+        },
+    }
+V1_REGISTRY = "172.17.42.2:5000"
+
+
+def mock_environment(tmpdir, docker_registry=None, primary_images=None,
+                     worker_annotations={}):
+    if MOCK:
+        mock_docker()
+    tasker = DockerTasker()
+    workflow = DockerBuildWorkflow(SOURCE, "test-image")
+    base_image_id = '123456parent-id'
+    setattr(workflow, '_base_image_inspect', {'Id': base_image_id})
+    setattr(workflow, 'builder', X())
+    setattr(workflow.builder, 'image_id', '123456imageid')
+    setattr(workflow.builder, 'base_image', ImageName(repo='Fedora', tag='22'))
+    setattr(workflow.builder, 'source', X())
+    setattr(workflow.builder, 'built_image_info', {'ParentId': base_image_id})
+    setattr(workflow.builder.source, 'dockerfile_path', None)
+    setattr(workflow.builder.source, 'path', None)
+    setattr(workflow, 'tag_conf', TagConf())
+    if primary_images:
+        workflow.tag_conf.add_primary_images(primary_images)
+
+    annotations = deepcopy(BUILD_ANNOTATIONS)
+    if not worker_annotations:
+        worker_annotations = {'ppc64le': PPC_ANNOTATIONS}
+    for worker in worker_annotations:
+        annotations['worker-builds'][worker] = deepcopy(worker_annotations[worker])
+
+    workflow.build_result = BuildResult(image_id='123456', annotations=annotations)
+
+    return tasker, workflow
+
+
+def mock_url_responses(docker_registry, test_images, worker_digests, version='2', respond=True):
+    responses.reset()
+    for worker_digest in worker_digests:
+        digest = worker_digest[0]['digest']
+        repo = worker_digest[0]['repository']
+        for registry in docker_registry:
+            if not registry.startswith('http://') and not registry.startswith('https://'):
+                registry = 'https://' + registry
+            url = '{0}/v2/{1}/manifests/{2}'.format(registry, repo, digest)
+            body = json.dumps({'tag': 'testtag', 'schemaVersion': version})
+            responses.add(responses.GET, url, body=body)
+            if respond:
+                status = 200
+            else:
+                status = 400
+                body = json.dumps({'error': 'INVALID MANIFEST'})
+            for image_tag in test_images:
+                url = '{0}/v2/{1}/manifests/{2}'.format(registry, repo, image_tag.split(':')[1])
+                responses.add(responses.PUT, url, status=status, json=body)
+
+
+class TestGroupManifests(object):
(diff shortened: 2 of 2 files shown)


mlangsdorf:
This code is most likely wrong, but it's close enough for review.

I suspect that this code:
```
            digest = worker_digests[0]['digest']
            tag = worker_digests[0]['tag']
            repo = tag.split(':')[0]
                image_tag = image.to_str(registry=False)

                url = '{}/v2/{}/manifests/{}'.format(registry, repo, digest)

                response = requests.head(url, verify=not insecure, auth=auth)
                response = requests.get(url, verify=not insecure, auth=auth)
```
should use the image name in place of the digest, but I'm not 100% sure.

I can't get this code to fail if there's an amd64 platform in the annotations
```
        for platform in all_annotations['worker-builds']:
            if self.goarch.get(platform, platform) == 'amd64':
```
I suspect that is not the desired behavior, but I don't understand what goarch is for.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,112 @@
...
+from atomic_reactor.plugin import PostBuildPlugin
+from atomic_reactor.plugins.build_orchestrate_build import OrchestrateBuildPlugin
+from atomic_reactor.util import Dockercfg
+
+
+class GroupManifestsPlugin(PostBuildPlugin):
+    key = 'group_manifests'
+    is_allowed_to_fail = False
+
+    def __init__(self, tasker, workflow, registries, group=True, goarch={}):
```
lcarva: Avoid using mutable objects as default values. Instead, set it to `None` and create a new instance in method body.

```python
..., goarch=None):
  goarch = goarch or {}
```

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,112 @@
...
+
+            digest = worker_digests[0]['digest']
+            tag = worker_digests[0]['tag']
+            repo = tag.split(':')[0]
+            for image in self.workflow.tag_conf.images:
+                image_tag = image.to_str(registry=False)
+
+                url = '{}/v2/{}/manifests/{}'.format(registry, repo, digest)
+
+                response = requests.head(url, verify=not insecure, auth=auth)
```
lcarva: Why `HEAD`?

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,112 @@
...
+                response = requests.get(url, verify=not insecure, auth=auth)
```
lcarva: Do we need to do this for each iteration in this for-loop? Parameters don't seem to change.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,112 @@
...
+    key = 'group_manifests'
+    is_allowed_to_fail = False
+
+    def __init__(self, tasker, workflow, registries, group=True, goarch={}):
+        """
+        constructor
+
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param timeout: int, maximum number of seconds to wait
```
lcarva: outdated params doc

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def run(self):
+        if self.group:
+            raise NotImplementedError('group=True is not supported in group_manifest')
+        grouped_manifests = []
+
+        build_result = self.workflow.buildstep_result.get(OrchestrateBuildPlugin.key)
+        all_annotations = build_result.annotations
+        for platform in all_annotations['worker-builds']:
+            if self.goarch.get(platform, platform) == 'amd64':
```
lcarva: What's the purpose of `goarch`? A way to control arch to platform mapping?

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,112 @@
...
+                grouped_manifests = self.get_worker_manifest(all_annotations['worker-builds'][platform])  # noqa
+                break
+
```
lcarva: We should probably handle the case where the platform mapping to 'amd64' was not found.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,112 @@
...
+            for image in self.workflow.tag_conf.images:
+                image_tag = image.to_str(registry=False)
+
+                url = '{}/v2/{}/manifests/{}'.format(registry, repo, digest)
+
+                response = requests.head(url, verify=not insecure, auth=auth)
+                response = requests.get(url, verify=not insecure, auth=auth)
+
+                image_manifest = json.loads(response.text)
+                image_manifest['tag'] = image_tag
```
owtaylor: This is only part of the v1 schema, not part of the v2 schema (or manifest list schema), and in any case, the tag that is applied in the registry is the reference from the URL. If all tags can share the exact same manifest content/digest, that seems preferable.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,112 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+
+create a group of the image manifests from the worker builds
```
owtaylor: This isn't very clear to me at all. "a group of the image manifest" or "group manifest" aren't - as far as I know - standard things. Is the final intent, something like:

 Creates a manifest list that includes the manifests from the worker builds and uploads that to the configured registries

? (I think the group_manifests name for the plugin is also perhaps not the best one!) If the final intent is that if there is only one architecture, it creates a single manifest (essentially retagging the worker build), that should also be indicated here.

owtaylor:
> I suspect that this code:
> 
>             digest = worker_digests[0]['digest']
>             tag = worker_digests[0]['tag']
>             repo = tag.split(':')[0]
>             for image in self.workflow.tag_conf.images:
>                 image_tag = image.to_str(registry=False)
> 
>                 url = '{}/v2/{}/manifests/{}'.format(registry, repo, digest)
> 
>                 response = requests.head(url, verify=not insecure, auth=auth)
>                 response = requests.get(url, verify=not insecure, auth=auth)
> 
> should use the image name in place of the digest, but I'm not 100% sure.

Why? I think the assumption can be made that the worker pushed the *same* image for all the tags, and the digest that the worker returned is the reliable way to get exactly that image for that architecture. As @lcarva said, I think the head can be removed and this can be pulled out of the loop.

atomic_reactor/plugins/post_group_manifests.py
twaugh: That's right. http://osbs.readthedocs.io/en/latest/multiarch.html#group-manifests

atomic_reactor/plugins/post_group_manifests.py
vrutkovs: `HEAD` should be sufficient for checking that the manifest exists, but it also should have `response.raise_for_status()` to throw exception on HTTP 404.

Anyway, since `get` is performed later on, its not needed

twaugh:
What `goarch` is for: manifest lists describe platforms in terms of the strings used by the Go language to describe operating systems (GOOS) and architectures (GOARCH). We're using platform names such as 'x86_64' to say GOOS=linux, GOARCH=amd64. So far, the GOOS is assumed to be 'linux', but to work out the GOARCH we need a translation dictionary to change 'x86_64' into 'amd64'. Hopefully other platform names will be the same as the GOARCH names. FWIW I just submitted a docs update to describe how this translation dictionary will be configured: https://github.com/projectatomic/osbs-docs/pull/31

atomic_reactor/plugins/post_group_manifests.py
twaugh: The idea behind naming it "group_manifests" is that in the future it may need to group them together in an OCI Image Index, as well as (or instead of) a Docker Manifest List.

It's unfortunate that "group" is a noun as well as a verb. It's intended as a verb in this plugin name: bring together the image manifests in some way.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: It currently returns an empty list in that case. What else should it do?

lcarva: If this plugin ran with `group=False`, and the `amd64` goarch was not found, it indicates an error IMO. @twaugh, thoughts?

twaugh: Yes, agree. Raise an exception.

atomic_reactor/plugins/post_group_manifests.py
owtaylor: If I think of it as a verb, the plugin name seems fine. But this comment needs to be improved to be in line with that and to reference back to concepts (a manifest list) that people reading the code can be expected to know or able to look up.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,112 @@
...
+
+        return worker_manifest
+
+    def run(self):
+        if self.group:
+            raise NotImplementedError('group=True is not supported in group_manifest')
+        grouped_manifests = []
+
+        build_result = self.workflow.buildstep_result.get(OrchestrateBuildPlugin.key)
+        all_annotations = build_result.annotations
```
mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,112 @@
...
+"""
+
+
+from __future__ import unicode_literals
+from copy import deepcopy
+import requests
+import requests.auth
+import json
+
+try:
```
mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: fixed. possibly.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: Better examples of goarch added to the tests.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,121 @@
...
+        for platform in all_annotations['worker-builds']:
+            if self.goarch.get(platform, platform) == 'amd64':
+                valid = True
+                grouped_manifests = self.get_worker_manifest(all_annotations['worker-builds'][platform])  # noqa
+                break
+
+        if valid:
+            return grouped_manifests
+        else:
+            raise PluginFailedException('failed to find an x86_64 platform')
```
lcarva: The exception raised here will be wrapped in a `PluginFailedException`. Let's use something else. Seems like a good case for `RuntimeError` or `ValueError`.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,121 @@
...
+        if self.group:
+            raise NotImplementedError('group=True is not supported in group_manifest')
+        grouped_manifests = []
+
+        valid = False
+        all_annotations = self.workflow.build_result.annotations
+        for platform in all_annotations['worker-builds']:
+            if self.goarch.get(platform, platform) == 'amd64':
+                valid = True
+                grouped_manifests = self.get_worker_manifest(all_annotations['worker-builds'][platform])  # noqa
```
lcarva: Why `# noqa` ?

mlangsdorf: line is too long and there's no good place to split it.

vrutkovs:
```
                worker_data = all_annotations['worker-builds'][platform]
                grouped_manifests = self.get_worker_manifest(worker_data)
```

mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,122 @@
...
+
+        for registry, registry_conf in self.registries.items():
+            if registry_conf.get('version') == 'v1':
+                continue
+
+            if not registry.startswith('http://') and not registry.startswith('https://'):
+                registry = 'https://' + registry
+
+            registry_noschema = urlparse(registry).netloc
+            for push_conf_registry in self.workflow.push_conf.docker_registries:
```
vrutkovs: `self.workflow.push_conf` would always return an empty list on orchestrator, as it is being filled by `tag_and_push` plugin.

Probably workers should also store the registry they've pushed the image to during the sync

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,122 @@
...
+            url = '{}/v2/{}/manifests/{}'.format(registry, repo, digest)
+            response = requests.get(url, verify=not insecure, auth=auth)
+
+            image_manifest = json.loads(response.text)
+
+            if image_manifest['schemaVersion'] == '1':
+                msg = 'invalid schema from {}'.format(url)
+                raise PluginFailedException(msg)
+
+            for image in self.workflow.tag_conf.images:
```
vrutkovs: Same here - this property is being set by `tag_and_push`. I think the plugin should iterate over `worker_digest` instead

twaugh: I think it's right as it is -- `tag_from_config` will set these tags.

vrutkovs: Right, I've assumed this function will be called on orchestrator - it probably won't, as for manifest list we need just a digest and goarch

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: I know delete_from_registry runs on the orchestrator and worker in arrangement 3 and later, and I took the registry handling code from delete_from_registry. But I don't know that the push_conf is the same between worker and orchestrator. I'm going to leave this alone for now.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: tag_from_config only runs on worker builds, so will the workflow.tag_conf on the orchestrator build inherit the tags?

twaugh: In the next arrangement version we'll be running `tag_from_config` in both the worker build (for the unique tags) and the orchestrator build (for all the other tags).

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,122 @@
...
+                    auth = requests.auth.HTTPBasicAuth(username, password)
+
+            digest = worker_digests[0]['digest']
+            tag = worker_digests[0]['tag']
+            repo = tag.split(':')[0]
+
+            url = '{}/v2/{}/manifests/{}'.format(registry, repo, digest)
+            response = requests.get(url, verify=not insecure, auth=auth)
+
+            image_manifest = json.loads(response.text)
```
twaugh: `response.json()`

atomic_reactor/plugins/post_group_manifests.py
lcarva: `delete_from_registry` is actually a noop in orchestrator build because as @vrutkovs pointed out, push_conf is empty in this case. This is being address by https://github.com/projectatomic/atomic-reactor/pull/741.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,138 @@
...
+
+                if image_manifest['schemaVersion'] == '1':
+                    msg = 'invalid schema from {0}'.format(url)
+                    raise PluginFailedException(msg)
+
+                for image in self.workflow.tag_conf.images:
+                    image_tag = image.to_str(registry=False)
+                    url = '{0}/v2/{1}/manifests/{2}'.format(registry, repo, image_tag)
+                    self.log.debug("for image_tag %s, putting at %s", image_tag, url)
+                    response = requests.put(url, image_manifest, **kwargs)
```
twaugh: We should call `response.raise_for_status()` here I think.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,138 @@
...
+
+                image_manifest = response.json()
+
+                if image_manifest['schemaVersion'] == '1':
+                    msg = 'invalid schema from {0}'.format(url)
+                    raise PluginFailedException(msg)
+
+                for image in self.workflow.tag_conf.images:
+                    image_tag = image.to_str(registry=False)
+                    url = '{0}/v2/{1}/manifests/{2}'.format(registry, repo, image_tag)
```
twaugh: `{2}` must be only the suffix part, so `image.tag` I think.

mlangsdorf:
tested with brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888/mlangsdo/buildroot:eng-rhel-7-docker-candidate-94693-20170816174925, which is a build of pkgs.devel.redhat.com/rpms/osbs-buildroot-docker@private-mlangsdorf-incremental

that uses:
https://github.com/mlangsdorf/atomic-reactor.git@arrangement4
https://github.com/mlangsdorf/osbs-client.git@arrangement4

build logs are here:
https://osbs.qa.engineering.redhat.com:8443/console/project/osbs-stage/browse/builds/helloworld-container-master-9bc2a/helloworld-container-master-9bc2a-77?tab=logs

please merge thanks.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+from __future__ import unicode_literals
+from copy import deepcopy
+import requests
+import requests.auth
+
+from six.moves.urllib.parse import urlparse
+
+from atomic_reactor.plugin import PostBuildPlugin, PluginFailedException
```
twaugh: It is now.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,149 @@
...
+        constructor
+
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param registries: dict, keys are docker registries, values are dicts containing
+                           per-registry parameters.
+                           Params:
+                            * "secret" optional string - path to the secret, which stores
+                              login and password for remote registry
+        :param group: bool, if true, return the grouped image manifests from the workers
```
twaugh: Nit: better description might be "if true, create a manifest list; otherwise only add tags to amd64 image manifest"

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,149 @@
...
+        :param goarch: dict, keys are platform, values are go language platform names
+        """
+        # call parent constructor
+        super(GroupManifestsPlugin, self).__init__(tasker, workflow)
+        self.group = group
+        self.goarch = goarch or {}
+        self.registries = deepcopy(registries)
```
twaugh: Not sure what the copy is for. We don't modify it.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,149 @@
...
+
+                worker_manifest.append(image_manifest)
+                self.log.debug("appending an image_manifest")
+                break
+
+        return worker_manifest
+
+    def run(self):
+        if self.group:
+            raise NotImplementedError('group=True is not supported in group_manifest')
```
twaugh: Typo: should be "group_manifests"

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,149 @@
...
+        valid = False
+        all_annotations = self.workflow.build_result.annotations['worker-builds']
+        for plat, annotation in all_annotations.items():
+            digests = annotation['digests']
+            for digest in digests:
+                registry = digest['registry']
+                if registry in self.worker_registries:
+                    self.worker_registries[registry].append(registry)
+                else:
+                    self.worker_registries[registry] = [registry]
```
twaugh: Unnecessary `if`.

```
registry = digest['registry']
self.worker_registries.setdefault(registry, [])
self.worker_registries[registry].append(registry)
```

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,149 @@
...
+                for image in self.workflow.tag_conf.images:
+                    image_tag = image.to_str(registry=False).split(':')[1]
+                    url = '{0}/v2/{1}/manifests/{2}'.format(registry, repo, image_tag)
+                    self.log.debug("for image_tag %s, putting at %s", image_tag, url)
+                    response = requests.put(url, json=image_manifest, **kwargs)
+
+                    if not response.ok:
+                        msg = "PUT failed: {0},\n manifest was: {1}".format(response.json(),
+                                                                            image_manifest)
+                        self.log.error(msg)
```
twaugh: These lines don't get covered in unit testing.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: fixed.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: fixed.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: fixed.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: fixed.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: fixed.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,147 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+
+get the image manifest lists from the worker builders. If possible, group them together
+and return them. if not, return the x86_64/amd64 one after tagging it.
```
twaugh: Could use clarification to avoid putting 'manifest' and 'lists' together when not talking about 'manifest list' objects.
```
Get the image manifest digests from the worker builds and group them together into a manifest list.
When told not to group them (group=False), just tag the amd64 image manifest instead.
```

mlangsdorf: Fixed. Also slightly clarified the commit message.

twaugh: It still says "image manifest lists" which could be ambiguous.

twaugh:
@mlangsdorf, is this OK to merge? I think it is, just checking.

vrutkovs:
IIUC https://github.com/projectatomic/atomic-reactor/pull/778 supersedes this PR

twaugh:
I'd like to be able to perform testing for arrangement version 4 which requires group_manifests, but does not require group=True to work.

twaugh:
Draft release notes updated.

PR #759, opened 2017-07-28
Fix pulp_pull media type checking

twaugh:

Not sure why the tests didn't spot this.

### DIFF
--- a/atomic_reactor/plugins/post_pulp_pull.py
+++ b/atomic_reactor/plugins/post_pulp_pull.py
@@ -67,9 +67,9 @@ class PulpPullPlugin(PostBuildPlugin):
 
         media_types = []
         for plugin in self.workflow.postbuild_plugins_conf:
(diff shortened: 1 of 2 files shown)


twaugh:
The tests are fine, there doesn't seem to be any way to test for this failure. With recent Python versions these strings just get put into the same objects, even with explicit copy.

twaugh:
@mlangsdorf, the reason this fix is required is that `is` is a test to see if the two things are the same object (i.e. `a is b` means `id(a) == id(b)`), whereas `==` is the less strict test to see if the two things compare as equal.

In day-to-day code, `is None` is the main case for using `is`.

lcarva:
I saw this and thought it should work because the plugin uses the same constant. I realize now that it's not the same object because the value is being passed across systems.

twaugh:
Ah, figured out how to make the test catch this: using `json.loads()`.

PR #764, opened 2017-07-31
test_util: Fix logic in get_manifest_digests_missing

owtaylor:
[ This fixes up some issues introduced in PR #754 ]

If there's no 'Content-Type' header, the responses module adds
Content-Type: text-plain. This was confusing the test code, and the checks
were made to pass in a way that didn't fully make sense.

Add a hook into the get_manifest_digests() code to treat
X-Ignore-Content-Type as a missing content type, then fix up the logic
in the test.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -12,6 +12,7 @@ import json
 import os
 import tempfile
 import pytest
+import requests
 import responses
 from requests.exceptions import ConnectionError
 import six
@@ -381,7 +382,6 @@ def test_get_manifest_digests(tmpdir, image, registry, insecure, creds,
     (False, True),
     (False, False),
 ])
-@responses.activate
 def test_get_manifest_digests_missing(tmpdir, has_content_type_header, has_content_digest,
                                       digest_is_v1, can_convert_v2_v1):
     kwargs = {}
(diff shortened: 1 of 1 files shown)


vrutkovs:
The test should be rewritten actually, a better way to mock `requests.Response` is in https://github.com/projectatomic/atomic-reactor/blob/master/tests/plugins/test_pulp_pull.py#L85:
```
    config_response_config_v2_no_headers = requests.Response()
    (flexmock(config_response_config_v2_no_headers,
              raise_for_status=lambda: None,
              status_code=requests.codes.ok,
              _content=json.dumps({'foo': 'bar'}).encode('utf-8'),
              headers={}))
     ...
    def custom_get_v2_no_headers(self, url, headers, **kwargs):
        return self.config_response_config_v2_no_headers
    ...
    (flexmock(requests)
            .should_receive('get')
            .replace_with(custom_get_v2_no_headers))
```

owtaylor:
OK, redid it directly mocking requests.

lcarva:
Unit tests are failing.

Btw, the easiest way to mock `requests` module is to use, well... `responses` module!
See example here: https://github.com/projectatomic/atomic-reactor/blob/master/tests/plugins/test_add_filesystem.py#L383

owtaylor:
> Unit tests are failing.

Only tested with python2 locally. :-( - need to set up a local python3 test environment, but for now, pushed a "blind" fix that hopefully does the trick.

> Btw, the easiest way to mock requests module is to use, well... responses module!
> See example here: https://github.com/projectatomic/atomic-reactor/blob/master/tests/plugins/test_add_filesystem.py#L383

The problem is that responses is unable to simulate returning a response with no Content-Type header - text/plain is added by default, so the test wasn't working as expected. My first attempt hacked around this with responses, but @vrutkovs suggested directly mocking on responses, and in this case, it ended up equally simple and cleaner.

twaugh:
No release notes changes needed.

PR #765, opened 2017-08-01
Pulp tag

mlangsdorf:
pulp_tag: add post_pulp_tag plugin
    
the pulp_tag plugin runs in the orchestrator build. When push_to_pulp runs,
it now also returns the top_layer of the pushed image.
store_metadata_in_osv3 stores that value as v1-image-id. pulp_tag takes
that value from the worker metadata and tags the image in pulp with it.
It returns a dict of the repos and tags.

if there are multiple instances of v1-image-id, pulp_tag raises an
exception.
    

atomic_reactor/util_pulp.py
```diff
@@ -0,0 +1,232 @@
+"""
```
twaugh: For the Koji equivalent of this we named it `koji_util.py`. We should be consistent, one way or the other.

mlangsdorf: Fixed.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -205,6 +206,11 @@ def run(self):
             else:
                 self.log.error("Unknown result from add_help plugin: %s", help_result)
 
+        pulp_push_results = self.workflow.postbuild_results.get(PLUGIN_PULP_PUSH_KEY)
+        if pulp_push_results:
+            top_layer, _ = pulp_push_results
+            annotations['v1-image-id'] = json.dumps(top_layer)
```
twaugh: Since top_layer will always be a string, there is no need to wrap it in JSON.

atomic_reactor/plugins/post_pulp_tag.py
```diff
@@ -0,0 +1,90 @@
...
+
+from atomic_reactor.plugin import PostBuildPlugin
+from atomic_reactor.plugins.build_orchestrate_build import get_worker_build_info
+from atomic_reactor.constants import PLUGIN_PULP_TAG_KEY
+from atomic_reactor.pulp_util import PulpHandler
+
+
+class PulpTagPlugin(PostBuildPlugin):
+    """
+    Find a platform with a v1-image-id annotation and tag it in pulp with the value of that
```
twaugh: "... and tag that docker_image in pulp as directed by tag_conf"?

tests/plugins/test_pulp_tag.py
```diff
@@ -0,0 +1,254 @@
...
+class BuildInfo(object):
+    def __init__(self, v1_image_id=None):
+        annotations = {'meta': 'test'}
+        if v1_image_id:
+            annotations['v1-image-id'] = json.dumps(v1_image_id)
+
+        self.build = BuildResponse({'metadata': {'annotations': annotations}})
+
+
+def prepare(check_repo_retval=0, v1_tags={}):
```
twaugh: There's naming confusing between tags and image IDs here. These `v1_tags` get set as the value of the "v1-image-id" annotation.

An image ID looks like: `"002c95e6c2b4a56113d0d5cdeab7d203646125287b5f753345b5e6a4491c2fec"`.

The tags applied to that image ID might look like `["7.1-29", "7.1", "latest"]`.

This would result in an updateRepo call with a parameter like: `{"tags": "7.1-29,7.1,latest:002c95e6c2b4a56113d0d5cdeab7d203646125287b5f753345b5e6a4491c2fec"}`

(Yes, that way round... no, I don't know why)

So this parameter should be named `v1_image_ids`.

atomic_reactor/plugins/post_pulp_tag.py
```diff
@@ -0,0 +1,90 @@
...
+        self.username = username
+        self.password = password
+
+        self.dockpulp_loglevel = dockpulp_loglevel
+
+    def set_v1_tag(self, v1_image_id):
+        image_names = self.workflow.tag_conf.images[:]
+        # Work out image ID
+        image = self.workflow.image
+        self.log.info("fetching image %s from docker", image)
```
twaugh: No, we don't need to do that. :-)

Just call updateRepo.

atomic_reactor/plugins/post_pulp_tag.py
```diff
@@ -0,0 +1,90 @@
...
+            annotations = build_info.build.get_annotations()
+            v1_image_id = annotations.get('v1-image-id')
+            if v1_image_id:
+                v1_image_id = json.loads(v1_image_id)
+                if has_v1_image_id:
+                    msg = "two platforms with v1-image-ids: {0} and {0}".format(platform,
+                                                                                has_v1_image_id)
+                    raise RuntimeError(msg)
+                has_v1_image_id = platform
+                self.log.info("tagging v1-image-id %s to platform %s", v1_image_id, platform)
```
twaugh: "... for platform %s"?

tests/plugins/test_pulp_tag.py
```diff
@@ -0,0 +1,254 @@
...
+        return
+
+    msg = None
+    for platform, v1_tag in v1_tags.items():
+        if v1_tag:
+            msg = "tagging v1-image-id {0} to platform {1}".format(v1_tag, platform)
+            break
+
+    runner.run()
+    assert PLUGIN_PULP_TAG_KEY is not None
```
twaugh: This line is a little odd.

tests/plugins/test_pulp_tag.py
```diff
@@ -0,0 +1,254 @@
...
+    if msg:
+        assert msg in caplog.text()
+    else:
+        assert "tagging v1-image-id" not in caplog.text()
```
twaugh: Don't we also want to make sure updateRepo() was called with the arguments we expect?

tests/plugins/test_pulp_tag.py
mlangsdorf: it's a straight translation of 
assert PulpPushPlugin.key is not None
from test_pulp.py

I confess I don't see the logic and will experiment with removing it.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
mlangsdorf: Fixed.

atomic_reactor/plugins/post_pulp_tag.py
mlangsdorf: Fixed.

atomic_reactor/plugins/post_pulp_tag.py
mlangsdorf: Fixed.

_check_file got moved to an externally callable function and got a test, which simplified the test code for test_create_dockpulp_and_repos

atomic_reactor/plugins/post_pulp_tag.py
mlangsdorf: Fixed.

tests/plugins/test_pulp_tag.py
mlangsdorf: Fixed.

tests/plugins/test_pulp_tag.py
mlangsdorf: Fixed in test_pulp_tag ; test_pulp is still weird.

tests/plugins/test_pulp_tag.py
mlangsdorf: Fixed.

atomic_reactor/plugins/post_pulp_tag.py
```diff
@@ -0,0 +1,85 @@
...
+        # call parent constructor
+        super(PulpTagPlugin, self).__init__(tasker, workflow)
+        self.pulp_registry_name = pulp_registry_name
+        self.pulp_secret_path = pulp_secret_path
+        self.username = username
+        self.password = password
+
+        self.dockpulp_loglevel = dockpulp_loglevel
+
+    def set_v1_tag(self, v1_image_id):
```
twaugh: Nitpick: this could be named `set_v1_tags`, because it applies the tags from `tag_conf` to the image ID.

atomic_reactor/plugins/post_pulp_tag.py
```diff
@@ -0,0 +1,85 @@
...
+        image_names = self.workflow.tag_conf.images[:]
+        handler = PulpHandler(self.workflow, self.pulp_registry_name, self.log,
+                              pulp_secret_path=self.pulp_secret_path, username=self.username,
+                              password=self.password, dockpulp_loglevel=self.dockpulp_loglevel)
+
+        pulp_repos = handler.create_dockpulp_and_repos(image_names)
+        repo_tags = {}
+        for repo_id, pulp_repo in pulp_repos.items():
+            handler.update_repo(repo_id, {"tag": "%s:%s" % (",".join(pulp_repo.tags), v1_image_id)})
+            repo_tags[repo_id] = {"tag": "%s:%s" % (",".join(pulp_repo.tags), v1_image_id)}
```
twaugh: Any reason to rebuild that dict again?

atomic_reactor/plugins/post_pulp_tag.py
```diff
@@ -0,0 +1,85 @@
...
+                if has_v1_image_id:
+                    msg = "two platforms with v1-image-ids: {0} and {0}".format(platform,
+                                                                                has_v1_image_id)
+                    raise RuntimeError(msg)
+                has_v1_image_id = platform
+                self.log.info("tagging v1-image-id %s for platform %s", v1_image_id, platform)
+                ret_val = self.set_v1_tag(v1_image_id)
+                if ret_val:
+                    repo_tags = ret_val
+        return repo_tags
```
twaugh: Could lines 82-85 be replaced by `return self.set_v1_tag(v1_image_id)`?

tests/plugins/test_pulp_tag.py
```diff
@@ -0,0 +1,262 @@
...
+                'redhat-prefix-image-name2': {'tag': latest_tag}
+            }
+            break
+
+    results = runner.run()
+    if msg:
+        assert msg in caplog.text()
+    else:
+        assert "tagging v1-image-id" not in caplog.text()
+    assert results['pulp_tag'] == expected_results
```
twaugh: This is good, but we should also be using `(flexmock(dockpulp.Pulp).should_receive('updateRepo').with_args(...)` with the arguments we expect, I think.

atomic_reactor/plugins/post_pulp_tag.py
mlangsdorf: No; can't find the double v1-image-id annotation that way.

tests/plugins/test_pulp_tag.py
mlangsdorf:
```
    (flexmock(dockpulp.Pulp)
     .should_receive('updateRepo')
     .with_args(unicode, dict))
```
that's already in the code. updateRepo is called multiple times, so I don't think I can specify the arguments more than unicode, dict.

atomic_reactor/plugins/post_pulp_tag.py
mlangsdorf: Fixed

atomic_reactor/plugins/post_pulp_tag.py
mlangsdorf: Fixed

atomic_reactor/plugins/post_pulp_tag.py
twaugh: In the unit tests, you mean? You can tell flexmock to expect updateRepo to be called exactly once.

tests/plugins/test_pulp_tag.py
twaugh: So, at the moment, it is just checking that the *types* of the arguments are correct. It isn't checking that values, and that what I'm suggesting we want to change.

Why would updateRepo be called multiple times for a given test?

twaugh: Oh, I see, because there may be multiple repos.

But in fact, in the test data, we can make there only be one repo, right?

tests/plugins/test_pulp_tag.py
```diff
@@ -0,0 +1,262 @@
...
+    mock_docker()
+    return tasker, workflow
+
+
+@pytest.mark.skipif(dockpulp is None,
+                    reason='dockpulp module not available')
+@pytest.mark.parametrize(("v1_image_ids", "should_raise"), [
+    ({'x86_64': None, 'ppc64le': None}, False),
+    ({'x86_64': None, 'ppc64le': 'ppc64le_v1_tag'}, False),
+    ({'x86_64': 'ppc64le_v1_tag', 'ppc64le': 'ppc64le_v1_tag'}, True),
```
twaugh: These are image IDs, not tags, so perhaps:
```
({'x86_64': 'ppc64le_v1_image_id', 'ppc64le': 'ppc64le_v1_image_id'}, True),
```
etc

tests/plugins/test_pulp_tag.py
```diff
@@ -0,0 +1,262 @@
...
+    workflow = DockerBuildWorkflow(SOURCE, "test-image")
+    setattr(workflow, 'builder', X())
+    setattr(workflow.builder, 'source', X())
+    setattr(workflow.builder.source, 'dockerfile_path', None)
+    setattr(workflow.builder.source, 'path', None)
+    setattr(workflow, 'tag_conf', X())
+    setattr(workflow.tag_conf, 'images', [ImageName(repo="image-name1"),
+                                          ImageName(namespace="prefix",
+                                                    repo="image-name2"),
+                                          ImageName(repo="image-name3", tag="asd")])
```
twaugh: We can set this to e.g.
```
setattr(workflow.tag_conf, 'images', [ImageName(namespace="prefix", repo="repository", tag="1.0"),
                                      ImageName(namespace="prefix", repo="repository", tag="1.0-1"),
                                      ImageName(namespace="prefix", repo="repository", tag="latest")])
```

to control what we expect updateRepo to be called with.

mlangsdorf: Right. Fixed.

tests/plugins/test_pulp_tag.py
mlangsdorf: Fixed.

tests/plugins/test_pulp_tag.py
mlangsdorf: I think test_pulp_util and test_pulp are adequately testing the multiple repo scenario. Additional repos deleted from the tests and with_args specified exactly.

atomic_reactor/plugins/post_push_to_pulp.py
```diff
@@ -39,156 +39,78 @@
...
+        :param password: pulp password, used in preference to certificate and key
+        """
+        # call parent constructor
+        super(PulpPushPlugin, self).__init__(tasker, workflow)
+        self.pulp_registry_name = pulp_registry_name
+        self.image_names = image_names
+        if load_squashed_image is not None and load_exported_image is not None and \
+                (load_squashed_image != load_exported_image):
+            raise RuntimeError(
+                'Can\'t use load_squashed_image and load_exported_image with different values')
```
vrutkovs: Nitpick: `RuntimeError("Can't use..`

lcarva:
Updated release notes.

PR #767, opened 2017-08-03
Pulp pull retries

owtaylor:
The main point of this PR is to move the retrying when pulp_pull calls get_manifest_digests() back into the plugin, since it's not actually generic. The patch uses the same retry logic as was already in the plugin for the pull.

However, since the default parameters for the plugin wait for 10 minutes rather than the hardcoded 10 second retry that was added to get_manifest_digests(), this makes a problem much worse - if no manifests ever show up, then we'll retry until we time out. In my understanding, this will be the case when only pulp_push is configured (osbs-client is configured with registry_api_versions=v1).

The second patch in this PR deals with this by recording whether a pulp registry was used for syncing or just for pushing. If this problem is considered unimportant, it could be dropped for simplicity.

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -176,6 +176,11 @@ class PulpRegistry(Registry):
         super(PulpRegistry, self).__init__(crane_uri, insecure=insecure)
         self.name = name
 
+        # True if we've synced an image to the pulp repository using the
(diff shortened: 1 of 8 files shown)


owtaylor:
(rebased to resolve merge conflict)

twaugh:
No user-visible changes so I haven't updated the draft release notes.

PR #768, opened 2017-08-03
Pulp publish

mlangsdorf:
pulp_publish: new plugin to handle delayed pulp publish
    
Add a parameter to post_push_to_pulp, publish, which is AND'd with
the existing publish paramater. In future arrangements, this parameter
will be false and repos will not be pushed to crane.
    
Create a new plugin, exit_pulp_publish, which runs in the orchestrator.
If the build was successful, it completes the push to crane that was
deferred by post_push_to_pulp. Otherwise, it searches through the
annotations for a v1-image-ids key and calls dockpulp.remove for
all repos for that image.
    
pulp_utils gets a new method to encapsulate dockpulp.remove.
    
(this looks like a huge commit set, but the first three commits are in pulp_tag
and should go away when pulp_tag https://github.com/projectatomic/atomic-reactor/pull/765
is committed. The meaningful changes are in exit_pulp_publish and 
test_pulp_publish)


atomic_reactor/pulp_util.py
```diff
@@ -0,0 +1,179 @@
...
+        self.p.copy(repo_id, layer)
+
+    def update_repo(self, repo_id, tag):
+        self.p.updateRepo(repo_id, tag)
+
+    def remove_image(self, repo_id, image):
+        self.p.remove(repo_id, image)
+
+    def publish(self, keys):
+        task_ids = self.p.crane(keys, wait=True)
```
twaugh: Because of dockpulp's unexpected behaviour when `len(keys) == 0` (it calls publish for every single repository), we should probably `assert keys` before calling `crane()`.

tests/plugins/test_pulp_publish.py
```diff
@@ -0,0 +1,186 @@
+"""
+Copyright (c) 2015 Red Hat, Inc
```
twaugh: 2017

atomic_reactor/plugins/exit_pulp_publish.py
```diff
@@ -0,0 +1,122 @@
...
+                                        dockpulp_loglevel=self.dockpulp_loglevel)
+
+    def publish_to_crane(self, repo_prefix="redhat-"):
+        image_names = self.workflow.tag_conf.images[:]
+        # Find out how to publish this image.
+        self.log.info("image names: %s", [str(image_name) for image_name in image_names])
+
+        pulp_repos = self.pulp_handler.create_dockpulp_and_repos(image_names, repo_prefix)
+
+        self.pulp_handler.publish(pulp_repos.keys())
```
twaugh: I think this beginning section should read:

```
pulp_repos = set([image.pulp_repo for image in self.workflow.tag_conf.images])
self.pulp_handler.publish(pulp_repos)
```

`.pulp_repo` is an attribute of a tag_conf 'image', i.e. if the tag is named foo/bar:baz then the "pulp_repo" for it is named redhat-foo-bar. We're just trying to find out which repositories need to get published, given the tag_conf.

Note: tag_conf will be set by the group_manifests plugin.

tests/plugins/test_pulp_publish.py
mlangsdorf: fixed.

atomic_reactor/pulp_util.py
mlangsdorf: fixed.

atomic_reactor/plugins/exit_pulp_publish.py
```diff
@@ -0,0 +1,122 @@
...
+                    self.pulp_handler.remove_image(repo_id, v1_image_id)
+                pulp_registry = self.pulp_handler.get_registry_hostname()
+                return [ImageName(registry=pulp_registry, repo=repodata.registry_id, tag=tag)
+                        for dummy_repo, repodata in pulp_repos.items()
+                        for tag in repodata.tags]  # noqa
+        return []
+
+    def run(self):
+        if self.workflow.build_process_failed:
+            return self.delete_v1_layers()
```
mlangsdorf: fixed.

atomic_reactor/plugins/exit_pulp_publish.py
mlangsdorf: fixed.

atomic_reactor/plugins/exit_pulp_publish.py
```diff
@@ -0,0 +1,118 @@
...
+                                        username=self.username, password=self.password,
+                                        dockpulp_loglevel=self.dockpulp_loglevel)
+
+    def publish_to_crane(self, repo_prefix="redhat-"):
+        image_names = self.workflow.tag_conf.images[:]
+        # Find out how to publish this image.
+        self.log.info("image names: %s", [str(image_name) for image_name in image_names])
+
+        self.pulp_handler.create_dockpulp()
+        pulp_repos = [image.pulp_repo for image in image_names]
```
twaugh: I see from log failures that you'll need to add the repo_prefix here:
```
if not repo_prefix:
    repo_prefix = ''

pulp_repos = ['%s%s' % (repo_prefix, image.pulp_repo) for image in image_names]
```

mlangsdorf: Fixed.

mlangsdorf:
should pulp_publish only publish v1 images, the same way it only deletes v1 images?

twaugh:
No, both v1 and v2 should be published. In fact we don't have a choice about this in any case because 'publish' is a repository-level operation and applies to all content within the repository, v1 and/or v2.

The v1 content takes a different route into Pulp than the v2 content.

### API v1 images

* Worker build uploads directly to Pulp with a platform-specific tag (`pulp_push`)
* Orchestrator build updates tags on it (there is only one) so they are platform-independent (`pulp_tag`)
* Orchestrator build calls publish (`pulp_publish`)

### API v2 images

* Worker build pushes to a Docker registry using a platform-specific tag (`tag_and_push`)
* Orchestrator build creates Manifest List in Docker registry, with platform-independent tags (`group_manifests`)
* Orchestrator build syncs into Pulp from Docker Registry (`pulp_sync`)
* Orchestrator build calls publish (`pulp_publish`)

Because `pulp_publish` is the comment part used for both routes, that's where the error handling for v1 goes. For v2, if any worker build failed then `pulp_sync` won't get called, so v2 content will only remain in the Pulp repository if something goes wrong after syncing. (Which we ought to handle at some point, but it's a much less likely scenario.)

mlangsdorf:
tested with brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888/mlangsdo/buildroot:eng-rhel-7-docker-candidate-94693-20170816174925, which is a build of pkgs.devel.redhat.com/rpms/osbs-buildroot-docker@private-mlangsdorf-incremental

that uses:
https://github.com/mlangsdorf/atomic-reactor.git@arrangement4
https://github.com/mlangsdorf/osbs-client.git@arrangement4

build logs are here:
https://osbs.qa.engineering.redhat.com:8443/console/project/osbs-stage/browse/builds/helloworld-container-master-9bc2a/helloworld-container-master-9bc2a-77?tab=logs

please merge thanks.

atomic_reactor/plugins/exit_pulp_publish.py
```diff
@@ -0,0 +1,123 @@
...
+        :param username: pulp username, used in preference to certificate and key
+        :param password: pulp password, used in preference to certificate and key
+        """
+        # call parent constructor
+        super(PulpPublishPlugin, self).__init__(tasker, workflow)
+        self.pulp_registry_name = pulp_registry_name
+        self.pulp_secret_path = pulp_secret_path
+        self.username = username
+        self.password = password
+        self.dockpulp_loglevel = dockpulp_loglevel
```
twaugh: Nit: don't need to set these 5 instance variables as they are not used after this method.

atomic_reactor/plugins/exit_pulp_publish.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def delete_v1_layers(self, repo_prefix="redhat-"):
+        worker_builds = self.workflow.build_result.annotations['worker-builds']
+
+        for platform in worker_builds:
+            build_info = get_worker_build_info(self.workflow, platform)
+            annotations = build_info.build.get_annotations()
+            v1_image_id = annotations.get('v1-image-id')
+            if v1_image_id:
+                image_names = self.workflow.tag_conf.images[:]
```
twaugh: Nit: no need for this explicit copy (`[:]`), since we don't modify it.

mlangsdorf: fixed.

atomic_reactor/plugins/exit_pulp_publish.py
mlangsdorf: fixed.

twaugh:
Draft release notes updated.

PR #769, opened 2017-08-04
bump_release: increment release correctly with a patched value

vrutkovs:
`bump_release` would now adjust the value returned by `getNextRelease`:

20 -> 20 (if build 20 doesn't exist)
20 -> 21 (if build 20 exists)
20.1 -> 21
20.fc25 -> 20.fc25 (if build 20.fc25 doesn't exist)
20.fc25 -> 21.fc25 (if build 20.fc25 exists)
20.1.fc25 -> 21.fc25

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -46,6 +46,26 @@ class BumpReleasePlugin(PreBuildPlugin):
             }
         self.xmlrpc = create_koji_session(hub, koji_auth_info)
 
+    def get_patched_release(self, original_release, force_increment=False):
+        # Split the original release by dots, make sure there at least 3 items in parts list
+        parts = original_release.split('.', 2) + [None, None]
+        release, suffix, rest = parts[:3]
+
+        # Is this a "dot" release, i.e. second part is a number?
+        is_dotrel = suffix is not None and suffix.isdigit()
+
+        if is_dotrel or force_increment:
+            # Increment first part as a number
+            release = str(int(release) + 1)
+
+        # Remove second part if it's a number
+        if is_dotrel:
+            suffix = None
+
+        # Recombine the parts
+        return '.'.join([part for part in [release, suffix, rest]
+                         if part is not None])
+
     def run(self):
         """
         run the plugin
@@ -74,7 +94,7 @@ class BumpReleasePlugin(PreBuildPlugin):
 
         build_info = {'name': component, 'version': version}
         self.log.debug('getting next release from build info: %s', build_info)
-        next_release = self.xmlrpc.getNextRelease(build_info)
+        next_release = self.get_patched_release(self.xmlrpc.getNextRelease(build_info))
 
         # getNextRelease will return the release of the last successful build
         # but next_release might be a failed build. Koji's CGImport doesn't
@@ -87,7 +107,7 @@ class BumpReleasePlugin(PreBuildPlugin):
             if not build:
                 break
 
-            next_release = str(int(next_release) + 1)
+            next_release = self.get_patched_release(next_release, force_increment=True)
 
         # Always set preferred release label - other will be set if old-style
         # label is present
--- a/tests/plugins/test_bump_release.py
+++ b/tests/plugins/test_bump_release.py
@@ -111,8 +111,29 @@ class TestBumpRelease(object):
         False
     ])
     @pytest.mark.parametrize('next_release', [
-        {'actual': '1', 'expected': '1'},
-        {'actual': '1', 'expected': '2'},
+        {'actual': '1', 'builds': [], 'expected': '1'},
+        {'actual': '1', 'builds': ['1'], 'expected': '2'},
+        {'actual': '1', 'builds': ['1', '2'], 'expected': '3'},
+        {'actual': '20', 'builds': ['19.1'], 'expected': '20'},
+        {'actual': '20', 'builds': ['20', '20.1'], 'expected': '21'},
+        {'actual': '20.1', 'builds': ['20'], 'expected': '21'},
+        {'actual': '20.1', 'builds': ['20', '20.1'], 'expected': '21'},
+        {'actual': '20.2', 'builds': ['20', '20.1'], 'expected': '21'},
+        {'actual': '20.2', 'builds': ['20', '20.1', '20.2'], 'expected': '21'},
+        {'actual': '20.fc25', 'builds': ['20.fc24'], 'expected': '20.fc25'},
+        {'actual': '20.fc25', 'builds': ['20.fc25'], 'expected': '21.fc25'},
+        {'actual': '20.foo.fc25',
+         'builds': ['20.foo.fc25'],
+         'expected': '21.foo.fc25'},
+        {'actual': '20.1.fc25',
+         'builds': ['20.fc25', '20.1.fc25'],
+         'expected': '21.fc25'},
+        {'actual': '20.1.fc25',
+         'builds': ['20.fc25', '20.1.fc25', '21.fc25'],
+         'expected': '22.fc25'},
+        {'actual': '20.1.fc25',
+         'builds': ['20.1.fc25', '21.fc25'],
+         'expected': '22.fc25'},
     ])
     def test_increment(self, tmpdir, component, version, next_release,
                        include_target):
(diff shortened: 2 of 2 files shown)


atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -46,6 +46,32 @@ def __init__(self, tasker, workflow, hub, target=None, koji_ssl_certs_dir=None):
...
+
+        def _is_int(part):
+            try:
+                int(part)
+                return True
+            except ValueError:
+                return False
+
+        if original_release.count('.') > 1:
+            release, suffix, postfix = original_release.split('.', 2)
```
vrutkovs: Changed `postfix` to `remainder`

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -46,6 +46,32 @@ def __init__(self, tasker, workflow, hub, target=None, koji_ssl_certs_dir=None):
...
+        else:
+            postfix = ''
+            if original_release.count('.') == 1:
+                release, suffix = original_release.split('.', 1)
+            else:
+                release = original_release
+                suffix = ''
+
+        result = [str(int(release) + 1)]
+        if not _is_int(suffix):
```
vrutkovs: Oh, good idea

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -87,7 +113,7 @@ def run(self):
             if not build:
                 break
```
vrutkovs: `getNextRelease` won't increment the second part - https://pagure.io/koji/blob/master/f/hub/kojihub.py#_3590, for `20.2` it would return `21.2`, which would be adjusted to `21`

lcarva: > which would be adjusted to 21

Not sure if this would actually happen. If `getNextRelease` returns `21.2` and that build doesn't exist, this `break` statement will execute - meaning that `self.increment_release` is not invoked, thus release value won't be adjusted.

vrutkovs: Oh, correct, will rework this then

vrutkovs: Right, reworked this, PTAL

lcarva: That seems correct. Can we make sure the unit tests are reflecting this properly now? IIRC, they were passing with the previous behavior as well.

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -46,6 +46,25 @@ def __init__(self, tasker, workflow, hub, target=None, koji_ssl_certs_dir=None):
...
+            release, suffix, remainder = original_release.split('.', 2)
+        else:
+            remainder = ''
+            if original_release.count('.') == 1:
+                release, suffix = original_release.split('.', 1)
+            else:
+                release = original_release
+                suffix = ''
+
+        result = [str(int(release) + 1)] if increment else [release]
```
rcerven: this will increment even that 20.f25 to 21.f25, is this what you want? if yeah maybe acceptance criteria should be updated?

lcarva: If `increment` is `True` then it should update as such.
Note that the first time this method is called, `increment` is `False.

It will only increment `20.f25` to `21.f25`, if `getNextRelease` has returned `20.f25` and for some reason that build already exists in Koji. In which case, this is an enhancement of a supposedly already handled scenario.

atomic_reactor/plugins/pre_bump_release.py
vrutkovs: Right, added new tests parameter - `exists`. Now `getBuild` would return more realistic results - `True` if 'major' part of the releases matches (and if `exists` is set a full release string is matched)

tests/plugins/test_bump_release.py
```diff
@@ -129,9 +139,19 @@ def getNextRelease(self, build_info):
             def getBuild(self, build_info):
                 assert build_info['name'] == list(component.values())[0]
                 assert build_info['version'] == list(version.values())[0]
-                if build_info['release'] >= next_release['expected']:
-                    return None
-                return True
+
+                # Check 'major' part of the releases
+                if '.' in next_release['actual'] and \
```
lcarva: I don't think we want to perform splitting in the mock `getBuild`.
It should return build info solely based on `next_release['exists']`

I've altered the plugin code, as such:
```python
# next_release = self.get_patched_release(self.xmlrpc.getNextRelease(build_info))
```
and these tests still pass. They should fail.

I think the plugin code is right. It's just the testing that needs to be a little more robust.

vrutkovs: Yes, the tests were needlessly complex - now instead of `exists` flag the tests would receive the list of build releases which exist. Changing back to `next_release = self.xmlrpc.getNextRelease(build_info)` fails the tests:
```
next_release = {'actual': '20.1', 'builds': ['20'], 'expected': '21'}
....
assert '20.1' == '21'
```

Tests now also verify that the release is being incremented more than once

atomic_reactor/plugins/pre_bump_release.py
twaugh: I find this quite hard to follow because of all the cases for splitting. How about this?
```
parts = original_release.split('.', 2) + [None, None]
release, suffix, rest = parts[:3]
if increment:
    # Increment first part as a number
    release = str(int(release) + 1)

# Remove second part if it's a number
if suffix is not None and suffix.isdigit():
    suffix = None

# Recombine the parts
return '.'.join([part for part in [release, suffix, rest]
                 if part is not None])
```

tests/plugins/test_bump_release.py
```diff
@@ -111,8 +111,26 @@ def test_missing_labels(self, tmpdir, caplog, labels):
...
+        {'actual': '20.fc25', 'builds': ['20.fc25'], 'expected': '21.fc25'},
+        {'actual': '20.foo.fc25',
+         'builds': ['20.foo.fc25'],
+         'expected': '21.foo.fc25'},
+        {'actual': '20.1.fc25',
+         'builds': ['20.fc25', '20.1.fc25'],
+         'expected': '21.fc25'},
+        {'actual': '20.1.fc25',
+         'builds': ['20.fc25', '20.1.fc25', '21.fc25'],
+         'expected': '22.fc25'},
```
twaugh: Can we add a test like this?:
```
{'actual': '20.1.fc25',
 'builds': ['20.1.fc25', '21.fc25'],
 'expected': '22.fc25'},
```

At the moment that test fails.

vrutkovs: That's an incorrect testcase - build `20.1.fc25` cannot be created without `20.fc25`.

With `'builds': ['20.fc25', '20.1.fc25', '21.fc25'],` test passes - added this testcase

atomic_reactor/plugins/pre_bump_release.py
vrutkovs: It looks better, but I'm not sure if:
```if increment or is_dotrel:``` is a good idea - this would cause increment in first `self.get_patched_release(self.xmlrpc.getNextRelease(build_info))` call. That would fail the case from https://github.com/projectatomic/atomic-reactor/pull/769#discussion_r132246260, but I think its safe to assume that `20.fc25` exists if `20.1.fc25` is returned by `getNextRelease`, isn't it?

tests/plugins/test_bump_release.py
twaugh: Why wouldn't 20.1.fc25 be possible without 20.fc25? In any case, my point is that it shouldn't transform 20.1.fc25 into 20.fc25.

The test case you added is identical to the previous test case.

atomic_reactor/plugins/pre_bump_release.py
twaugh: No, it's not safe to assume that. As a trivial case, what if the release label had been explicitly set to 20.1.fc25 in the most recent build, and then the release label had been removed?

Without this change, we will actually *decrement* the release value, and we never want to do that.

tests/plugins/test_bump_release.py
vrutkovs: I think this assumption is indeed wrong - some packages indeed use custom format for release and it can accidentally match the format for isolated builds. I'll revert this part then

atomic_reactor/plugins/pre_bump_release.py
vrutkovs: Reworked code from https://github.com/projectatomic/atomic-reactor/pull/769#discussion_r132243638 a bit - renamed `increment` to `force_increment` and added the testcase

lcarva:
Release notes updated.

PR #770, opened 2017-08-07
New inject_parent_image plugin

lcarva:


atomic_reactor/util.py
```diff
@@ -75,7 +77,9 @@ def to_str(self, registry=True, tag=True, explicit_tag=False,
 
         result = self.repo
 
-        if tag and self.tag:
+        if tag and self.tag and ':' in self.tag:
```
vrutkovs: `tag.startswith('sha256:')`? That might limit us to one hashing method, but would be a bit more explicit

vrutkovs: Another option is to add a new `self.sep` and fill it in `def parse`

lcarva: Oooh I like the idea of self.sep. Will do that in the morning.

lcarva: Turns out that using `self.sep` is a little awkward because it must be computed in `__init__` as well as `parse` methods. Switched to using your initial suggestion.

atomic_reactor/plugins/pre_inject_parent_image.py
```diff
@@ -0,0 +1,110 @@
...
+                self._archive_repositories = repositories
+                break
+
+        if not self._archive_repositories:
+            raise RuntimeError('A suitable archive for Koji build {[nvr]} was not found'
+                               .format(self._koji_build_info))
+
+    def select_new_parent_image(self):
+        for repository in self._archive_repositories:
+            if '@sha256' in repository:
```
twaugh: I'm not sure we want to hard-code the checksum type do we?

twaugh: The way I've always done this is to test `'@' in ...` to know whether it's a by-digest pullspec.

vrutkovs: This was implemented after I suggested in https://github.com/projectatomic/atomic-reactor/pull/770#discussion_r131769180, it probably was a bad idea. Since we can't store `self.sep` lets just revert back to original code

lcarva: I'll change it to check `if '@' in repository`

lcarva: Ah! I can't do that, because the '@' has been stripped off.
Ok, reverting back to original?

atomic_reactor/plugins/pre_inject_parent_image.py
```diff
@@ -0,0 +1,110 @@
...
+from atomic_reactor.plugins.exit_remove_built_image import defer_removal
+from osbs.utils import graceful_chain_get
+
+
+class InjectParentImage(PreBuildPlugin):
+    """
+    Modifies parent image to be used based on given Koji build.
+
+    For each listed archive in Koji build, chose the archive that
+    defines a non-empty '.extra.docker.repositories' list.
```
twaugh: Maybe check build.extra.image.index.pull first? If that doesn't exist, the current approach is fine.

atomic_reactor/plugins/pre_inject_parent_image.py
```diff
@@ -0,0 +1,110 @@
...
+    This list provides the pull reference for the container image
+    associated with Koji build. If it contains multiple item, the
+    manifest digest, @sha256, is preferred. Otherwise, the first
+    repository in list is used.
+
+    The namespace and repository for the new parent image must match
+    the namespace and repostirory for the parent image defined in
```
twaugh: Typo

atomic_reactor/plugins/pre_inject_parent_image.py
lcarva: Done.

atomic_reactor/plugins/pre_inject_parent_image.py
lcarva: Fixed.

lcarva:
Updated release notes.

PR #771, opened 2017-08-08
Add filesystem koji task id to labels

bfontecc007:
This is related to story 3955. The idea here is to add the koji filestystem task id to the labels (in the orchestrator as well as worker).

### DIFF
--- a/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
+++ b/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
@@ -13,13 +13,15 @@ import os
 from osbs.api import OSBS
 from osbs.conf import Configuration
 from osbs.exceptions import OsbsResponseException
+from osbs.utils import graceful_chain_get
 
 from atomic_reactor.plugins.pre_add_help import AddHelpPlugin
 from atomic_reactor.plugins.post_pulp_pull import PulpPullPlugin
 from atomic_reactor.constants import (PLUGIN_KOJI_IMPORT_PLUGIN_KEY,
                                       PLUGIN_KOJI_PROMOTE_PLUGIN_KEY,
                                       PLUGIN_KOJI_UPLOAD_PLUGIN_KEY,
-                                      PLUGIN_PULP_PUSH_KEY)
+                                      PLUGIN_PULP_PUSH_KEY,
+                                      PLUGIN_ADD_FILESYSTEM_KEY)
 from atomic_reactor.plugin import ExitPlugin
 from atomic_reactor.util import get_build_json
 
@@ -61,6 +63,10 @@ class StoreMetadataInOSv3Plugin(ExitPlugin):
             return {}
 
         return annotations
+
+    def get_filesystem_koji_task_id(self):
+        res = self.get_result(self.workflow.prebuild_results.get(PLUGIN_ADD_FILESYSTEM_KEY))
+        return graceful_chain_get(res, 'filesystem-koji-task-id')
 
     def get_digests(self):
         """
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -62,6 +64,10 @@ def get_config_map(self):
 
         return annotations
 
+    def get_filesystem_koji_task_id(self):
+        res = self.get_result(self.workflow.prebuild_results.get(PLUGIN_ADD_FILESYSTEM_KEY, ''))
```
lcarva: Don't think you need the default empty string here. Omit second parameter to `get` letting it default to `None`.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -62,6 +64,10 @@ def get_config_map(self):
...
+        return graceful_chain_get(res, 'filesystem-koji-task-id') or ''
```
lcarva: Unnecessary `or ''` here.

tests/plugins/test_store_metadata.py
```diff
@@ -288,6 +289,40 @@ def test_metadata_plugin(tmpdir, br_annotations, expected_br_annotations,
...
+    )
+    output = runner.run()
+    labels = output[StoreMetadataInOSv3Plugin.key]["labels"]
+
+    if 'filesystem-koji-task-id' in res:
+        assert 'filesystem-koji-task-id' in labels
+        assert labels['filesystem-koji-task-id'] == 'example-fs-taskid'
+
+
+
```
twaugh: Just need 2 blank lines here, not 3.

tests/plugins/test_store_metadata.py
```diff
@@ -288,6 +289,40 @@ def test_metadata_plugin(tmpdir, br_annotations, expected_br_annotations,
...
+            }
+        }]
+    )
+    output = runner.run()
+    labels = output[StoreMetadataInOSv3Plugin.key]["labels"]
+
+    if 'filesystem-koji-task-id' in res:
+        assert 'filesystem-koji-task-id' in labels
+        assert labels['filesystem-koji-task-id'] == 'example-fs-taskid'
+
```
twaugh: We don't assert anything if `'filesystem-koji-task-id' not in res`. Are we just making sure the plugin copes with it OK? Maybe we should assert that no 'filesystem-koji-task-id' label is added.

tests/plugins/test_store_metadata.py
```diff
@@ -288,6 +289,40 @@ def test_metadata_plugin(tmpdir, br_annotations, expected_br_annotations,
         assert json.loads(annotations['media-types']) == expected_pulp_pull_results
 
 
+@pytest.mark.parametrize(('res'), (
+    {
+        'filesystem-koji-task-id': 'example-fs-taskid',
+        'base-image-id': 'foobar'
```
twaugh: Nitpick: I usually put a comma after each entry in a sequence or dict when they're on separate lines like this, even the last, to make diffs easier to read if we add/remove/swap lines afterwards.

tests/plugins/test_store_metadata.py
```diff
@@ -288,6 +289,41 @@ def test_metadata_plugin(tmpdir, br_annotations, expected_br_annotations,
...
+            }
+        }]
+    )
+    output = runner.run()
+    labels = output[StoreMetadataInOSv3Plugin.key]["labels"]
+
+    if 'filesystem-koji-task-id' in res:
+        assert 'filesystem-koji-task-id' in labels
+        assert labels['filesystem-koji-task-id'] == 'example-fs-taskid'
+    if 'filesystem-koji-task-id' not in res:
```
twaugh: nit: could just use `else:` here

vrutkovs: +1

twaugh:
Draft release notes updated.

PR #774, opened 2017-08-11
Tag base images with the OpenShift build name

twaugh:
This avoids a race condition in which two builds are contending over the as-pulled image tag:

1. build A starts
2. build B starts
3. build A is nearly finished and is about to remove the parent image
4. build B tags the parent image
5. build A removes the parent image
6. build B tries to e.g. inspect the parent image, or build the Dockerfile

To avoid this race, each build adds an additional tag to the parent image it pulls, using the value of the OpenShift build name. This keeps the image around until the build is done with it at the end.

The race condition is then between pulling the image and tagging it with this unique ID. This is addressed by retrying the pull if tagging fails.

Also included in this pull request is a plugin for helping to test this.

### DIFF
--- a/atomic_reactor/plugins/pre_pull_base_image.py
+++ b/atomic_reactor/plugins/pre_pull_base_image.py
@@ -11,7 +11,10 @@ Pull base image to build our layer on.
 
 from __future__ import unicode_literals
 
+import docker
+
 from atomic_reactor.plugin import PreBuildPlugin
+from atomic_reactor.util import get_build_json, ImageName
 
 
 class PullBaseImagePlugin(PreBuildPlugin):
@@ -71,9 +74,32 @@ class PullBaseImagePlugin(PreBuildPlugin):
 
         self.workflow.pulled_base_images.add(pulled_base)
 
-        if not base_image.registry:
-            response = self.tasker.tag_image(base_image_with_registry, base_image, force=True)
-            self.workflow.pulled_base_images.add(response)
-            pulled_base = response
+        # Attempt to tag it using a unique ID. We might have to retry
+        # if another build with the same parent image is finishing up
+        # and removing images it pulled.
 
+        # Use the OpenShift build name as the unique ID
+        unique_id = get_build_json()['metadata']['name']
+        base_image = ImageName(repo=unique_id)
+
+        for _ in range(20):
+            try:
+                self.log.info("tagging pulled image")
+                response = self.tasker.tag_image(base_image_with_registry,
+                                                 base_image)
+                self.workflow.pulled_base_images.add(response)
+                break
+            except docker.errors.NotFound:
+                # If we get here, some other build raced us to remove
+                # the parent image, and that build won.
+                # Retry the pull immediately.
+                self.log.info("re-pulling removed image")
+                self.tasker.pull_image(base_image_with_registry,
+                                       insecure=self.parent_registry_insecure)
+        else:
+            # Failed to tag it
+            self.log.error("giving up trying to pull image")
+            raise RuntimeError("too many attempts to pull and tag image")
+
+        self.workflow.builder.set_base_image(base_image.to_str())
         self.log.debug("image '%s' is available", pulled_base)
--- a/tests/docker_mock.py
+++ b/tests/docker_mock.py
@@ -193,6 +193,7 @@ mock_inspect_container = {
 def _find_image(img, ignore_registry=False):
     global mock_images
 
+    img = ImageName.parse(img).to_str(explicit_tag=True)
     for im in mock_images:
         im_name = im['RepoTags'][0]
         if im_name == img:
--- a/tests/docker_mock.py
+++ b/tests/docker_mock.py
@@ -193,14 +193,14 @@ mock_inspect_container = {
 def _find_image(img, ignore_registry=False):
     global mock_images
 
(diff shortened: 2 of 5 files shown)


atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -71,9 +74,29 @@ def run(self):
...
+        self.workflow.builder.set_base_image(base_image)
+
+        for _ in range(20):
+            try:
+                self.log.info("tagging pulled image")
+                response = self.tasker.tag_image(base_image_with_registry,
+                                                 base_image)
+                self.workflow.pulled_base_images.add(response)
+                break
+            except docker.errors.NotFound:
```
vrutkovs: Maybe a `sleep(1)` here?

tests/docker_mock.py
```diff
@@ -193,6 +193,7 @@
 def _find_image(img, ignore_registry=False):
     global mock_images
 
+    img = ImageName.parse(img).to_str(explicit_tag=True)
```
vrutkovs: This is masking the original `img` parameter - its probably a bad idea

atomic_reactor/plugins/pre_pull_base_image.py
twaugh: Would why we need to delay? The reason for the error is surely that we raced some other build and they already won.

vrutkovs: Hmm, agreed

tests/docker_mock.py
twaugh: Updated.

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -71,9 +74,32 @@ def run(self):
...
-            self.workflow.pulled_base_images.add(response)
-            pulled_base = response
+        # Attempt to tag it using a unique ID. We might have to retry
+        # if another build with the same parent image is finishing up
+        # and removing images it pulled.
+
+        # Use the OpenShift build name as the unique ID
+        unique_id = get_build_json()['metadata']['name']
+        base_image = ImageName(repo=unique_id)
+        self.workflow.builder.set_base_image(base_image)
```
rcerven: maybe add this after the loop? as at this point we still don't know if tagging will be successful

tests/plugins/test_pull_base_image.py
```diff
@@ -101,7 +115,7 @@ def test_pull_base_image_plugin(parent_registry, df_base, expected, not_expected
     workflow.builder.base_image = ImageName.parse(df_base)
 
     expected = set(expected)
-    expected.add(df_base)
```
rcerven: wouldn't it exist as well, we added tag, so now is image accessible via both tags

atomic_reactor/plugins/pre_pull_base_image.py
vrutkovs: Unsuccessfull tagging would cause exception and fail the build, but it does make sense to move it below the `for` loop to separate it from actual tagging part of the code

twaugh: Sure, I can do that.

tests/plugins/test_pull_base_image.py
twaugh: That's a good point. For some reason when I add that line back in I get test failures... will investigate.

twaugh:
Updated.

tests/plugins/test_pull_base_image.py
twaugh: `df_base` doesn't have the registry name included, which is why it was missing. Pulling e.g. library/foo results in example.registry.com/library/foo being present.

twaugh:
Re-based against master.

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -71,9 +74,32 @@ def run(self):
...
+                # Retry the pull immediately.
+                self.log.info("re-pulling removed image")
+                self.tasker.pull_image(base_image_with_registry,
+                                       insecure=self.parent_registry_insecure)
+        else:
+            # Failed to tag it
+            self.log.error("giving up trying to pull image")
+            raise RuntimeError("too many attempts to pull and tag image")
 
+        self.workflow.builder.set_base_image(base_image)
```
vrutkovs: This throws exception:
```
2017-08-15 11:41:25,162 platform:- - atomic_reactor.plugin - DEBUG - Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/atomic_reactor/plugin.py", line 242, in run
  File "/usr/lib/python2.7/site-packages/atomic_reactor/plugins/pre_pull_base_image.py", line 104, in run
    self.workflow.builder.set_base_image(base_image)
  File "/usr/lib/python2.7/site-packages/atomic_reactor/build.py", line 148, in set_base_image
    self.base_image = ImageName.parse(base_image)
  File "/usr/lib/python2.7/site-packages/atomic_reactor/util.py", line 52, in parse
    s = image_name.split('/', 2)
AttributeError: 'ImageName' object has no attribute 'split'
```

twaugh:
Updated again to fix set_base_image() traceback.

vrutkovs:
Tested on osbs-box, the image is being tagged and untagged correctly

vrutkovs:
Release notes updated

PR #776, opened 2017-08-16
pre_koji_parent: don't check Koji if the base image has no NVR labels

owtaylor:
If the base image doesn't have the com.redhat.component, version, or
release labels, then it wasn't built via OSBS and there's no point
in checking that it was imported into Koji or waiting for that to
happen. (This fixes a KeyError when using a Fedora base image as
the base image.)

### DIFF
--- a/atomic_reactor/plugins/pre_koji_parent.py
+++ b/atomic_reactor/plugins/pre_koji_parent.py
@@ -65,7 +65,8 @@ class KojiParentPlugin(PreBuildPlugin):
         self._poll_start = None
 
     def run(self):
-        self.detect_parent_image_nvr()
+        if not self.detect_parent_image_nvr():
+            return
         self.wait_for_parent_image_build()
         self.verify_parent_image_build()
         return self.make_result()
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -74,8 +75,17 @@ def detect_parent_image_nvr(self):
         config = self.workflow.base_image_inspect[INSPECT_CONFIG]
         labels = config['Labels'] or {}
 
+        label_names = 'com.redhat.component', 'version', 'release'
+        for label_name in label_names:
+            if label_name not in labels:
+                self._parent_image_nvr = None
+                self.log.info("Failed to find label '%s' in parent image. "
+                              "Not waiting for Koji build." % label_name)
```
twaugh: landscape.io suggests passing `label_name` as a separate parameter to the format string, rather than performing the formatting before the call, e.g. `self.log.info("Failed to ... %s ...", label_name)`.

owtaylor: Fixed.

twaugh:
"wasn't built via OSBS" -- it may have been built by an older version of OSBS which used a different set of label names.

Thinking about it, which labels is Fedora using for this? Not sure if a better approach is to check new/old aliases.

cc @maxamillion

maxamillion:
@twaugh Fedora isn't building base images via OSBS because of multi-arch. Our current labels are set [here](https://pagure.io/pungi-fedora/blob/master/f/fedora.conf#_296) but we can add whatever we need to there. Is there a set of labels that OSBS now expects to be in place for base images?

twaugh:
@maxamillion, got it. So we definitely need to handle images that were not built in OSBS. Thanks.

maxamillion:
thank you :+1:

twaugh:
Draft release notes updated.

PR #777, opened 2017-08-17
retry policy in atomic reactor

rcerven:


atomic_reactor/core.py
```diff
@@ -236,7 +240,36 @@ def __init__(self, base_url=None, timeout=120, **kwargs):
         except AttributeError:
             # docker-py 1.x
             self.d = docker.Client(**client_kwargs)
-
+        self.retry_times = retry_times
```
twaugh: Do we want to enforce a minimum of 1 here?

atomic_reactor/core.py
```diff
@@ -236,7 +240,36 @@ def __init__(self, base_url=None, timeout=120, **kwargs):
...
+        self.retry_delay = DOCKER_BACKOFF_FACTOR
+        self.retry_client_statuses = DOCKER_CLIENT_STATUS_RETRY
+
+
+    def retry(function):
```
twaugh: Is this a static method? Confused by the signature not including 'self', but the function body using it.

atomic_reactor/core.py
```diff
@@ -301,6 +334,7 @@ def build_image_from_git(self, url, image, git_path=None, git_commit=None,
         logger.info("build finished")
         return response
 
+    @retry
```
twaugh: This method uses two Docker API calls, so if the second one (start the container) fails the first one (create the container) will be executed again. I think we probably want to retry at the API call level, i.e. a retry class wrapping docker.APIClient, rather than at the level of methods of DockerTasker.

atomic_reactor/core.py
```diff
@@ -579,6 +623,7 @@ def remove_container(self, container_id, force=False):
         logger.debug("container_id = '%s'", container_id)
         self.d.remove_container(container_id, force=force)  # returns None
 
+    @retry
     def logs(self, container_id, stderr=True, stream=True):
```
twaugh: This one's interesting, because it's a generator function. Calling the function will always succeed and return a generator; it's when iterating through it that API calls may fail. I don't think wrapping this generator function by `retry` will have any effect.

atomic_reactor/util.py
```diff
@@ -953,3 +953,25 @@ class LabelFormatter(string.Formatter):
     """
     def get_field(self, field_name, args, kwargs):
         return (self.get_value(field_name, args, kwargs), field_name)
+
+
+def do_retries(method, url, client_statuses=HTTP_CLIENT_STATUS_RETRY,
```
twaugh: Why not use the built-in support for retries provided by `requests`?
https://github.com/projectatomic/osbs-client/blob/master/osbs/http.py#L99-L107

atomic_reactor/plugins/pre_check_and_set_rebuild.py
```diff
@@ -94,15 +98,17 @@ def run(self):
...
-                if ex.status_code == 409:
-                    # Someone else was modifying the build
-                    # configuration at the same time. Try again.
-                    self.log.debug("got status %d, retrying", ex.status_code)
+
+            for counter in xrange(times):
+                if counter > 0:
+                    sleep(delay)
+                try:
                     osbs.set_labels_on_build_config(buildconfig, labels)
```
twaugh: This already performs retries for 409 conflict: https://github.com/projectatomic/osbs-client/blob/master/osbs/core.py#L688

rcerven: 409 isn't included in osbs-client retries
https://github.com/projectatomic/osbs-client/blob/master/osbs/constants.py#L84

tests/plugins/test_check_and_set_rebuild.py
```diff
@@ -144,7 +145,6 @@ def test_409_response(self, monkeypatch):
 
         (flexmock(OSBS)
             .should_receive('set_labels_on_build_config')
-            .twice()
```
vrutkovs: This should be replaced with according `.times(<n>)` instruction

tests/plugins/test_add_yum_repo_by_url.py
```diff
@@ -34,7 +34,7 @@ class X(object):
 def prepare():
     if MOCK:
         mock_docker()
-    tasker = DockerTasker()
+    tasker = DockerTasker(retry_times=1)
```
vrutkovs: A cleaner way of doing this is putting:
```
fake_docker_tasker = DockerTasker(retry_times=1)
flexmock(DockerTasker).new_instances = fake_docker_tasker
```
in `mock_docker`, as we might want to test the actual docker here with long retries

tests/plugins/test_delete_from_registry.py
```diff
@@ -109,11 +121,11 @@ def test_delete_from_registry_plugin(saved_digests, req_registries, tmpdir):
...
             auth_type = requests.auth.HTTPBasicAuth if req_registries[reg] else None
-            (flexmock(requests)
-                .should_receive('delete')
-                .with_args(url, verify=bool, auth=auth_type)
+            (flexmock(requests.Session)
+                .should_receive('request')
+                .with_args('delete', url, verify=bool, auth=auth_type)
                 .once()
-                .and_return(flexmock(status_code=202)))
+                .and_return(flexmock(status_code=202, raise_for_status=lambda: None)))
```
vrutkovs: Why is `raise_for_status` being replaced?

tests/test_util.py
```diff
@@ -408,6 +408,16 @@ def test_get_manifest_digests_missing(tmpdir, has_content_type_header, has_conte
 
     url = 'https://example.com/v2/spam/manifests/latest'
 
+    def custom_retries(*args, **kwargs):
```
vrutkovs: Does this functions makes retries skip the function at all? Why is it needed here?

atomic_reactor/core.py
```diff
@@ -211,8 +213,54 @@ def build_image_privileged_container(self, build_image, json_args_path):
...
 
+def retry(function, *args, **kwargs):
+    retry_times = kwargs.pop('retry', None)
+    retry_delay = DOCKER_BACKOFF_FACTOR
+    retry_client_statuses = DOCKER_CLIENT_STATUS_RETRY
+
+    final_excep = None
+
+    for counter in range(retry_times):
+        if counter > 0:
```
vrutkovs: >it makes more sense to do sleep here Before executing function if I'd place it to 'except' that would mean sleep after last try as well

sounds good to me

atomic_reactor/plugins/pre_check_and_set_rebuild.py
```diff
@@ -94,15 +93,6 @@ def run(self):
                                       namespace=metadata.get('namespace', None))
             osbs = OSBS(osbs_conf, osbs_conf)
             labels = {self.label_key: self.label_value}
-            try:
```
rcerven: I was thinking the same, but Tim pointed out, that it actually does, search for 'retry_on_conflict'

twaugh: It does. See `retry_on_conflict`.

vrutkovs: Ah, fair enough

tests/plugins/test_check_and_set_rebuild.py
rcerven: well it should be now once()

vrutkovs: Its not being implied by default, it has to be stated explicitely

tests/plugins/test_delete_from_registry.py
rcerven: not needed when using Session

tests/test_util.py
rcerven: this is to set retries to 0, because if I don't do this, default retry values from constants will be used

tests/plugins/test_check_and_set_rebuild.py
rcerven: yes I've already added once :)

atomic_reactor/core.py
```diff
@@ -211,8 +213,54 @@ def build_image_privileged_container(self, build_image, json_args_path):
...
+            sleep(retry_delay)
```
rcerven: changed

atomic_reactor/util.py
```diff
@@ -719,25 +723,27 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
...
                 image, registry, digest=None,
                 insecure=insecure, dockercfg_path=dockercfg_path,
                 version=version)
-            if response.status_code == 404:
-                sleep(1)
+        except (ConnectionError, HTTPError, RetryError) as ex:
+            # If the registry has a v2 manifest that can't be converted into a v1
+            # manifest, the registry fails with status=400, and a error code of
+            # MANIFEST_INVALID.
+            print("====================================================ex")
```
rcerven: done

atomic_reactor/util.py
rcerven: done

tests/plugins/test_add_yum_repo_by_url.py
twaugh: I don't mind.

tests/test_util.py
vrutkovs: This resets the remaining amount of retries to 0, so with a proper backoff formula it should wait the longest amount.

Perhaps replacing `retries` to `1` would be faster

atomic_reactor/constants.py
```diff
@@ -70,3 +70,16 @@
 PLUGIN_PULP_TAG_KEY = 'pulp_tag'
 PLUGIN_ADD_FILESYSTEM_KEY = 'add_filesystem'
 PLUGIN_FETCH_WORKER_METADATA_KEY = 'fetch_worker_metadata'
+
+# max retries for docker requests
+DOCKER_MAX_RETRIES = 3
+# how many seconds should wait before another try of docker request
+DOCKER_BACKOFF_FACTOR = 5
+# docker client retries statuses
+DOCKER_CLIENT_STATUS_RETRY = [404, 408, 500, 502, 503, 504]
```
vrutkovs: 404 should be retried only in some cases, e.g. pulling from registry, which is not related to docker

atomic_reactor/core.py
```diff
@@ -211,8 +213,54 @@ def build_image_privileged_container(self, build_image, json_args_path):
...
+def retry(function, *args, **kwargs):
+    retry_times = kwargs.pop('retry', None)
+    retry_delay = DOCKER_BACKOFF_FACTOR
+    retry_client_statuses = DOCKER_CLIENT_STATUS_RETRY
+
+    final_excep = None
+
+    for counter in range(retry_times):
+        if counter > 0:
+            sleep(retry_delay * counter)
```
vrutkovs: Wrong formula, should be:
`retry_delay * (2 ** counter)`

tests/test_util.py
rcerven: remaining amount???
do_retry is called once and Session with Retry set to 'total=0' will be used, which mean no retry at all

vrutkovs: Ah, I see, each function has a different naming for params, so its hard to follow. Resolved

tests/plugins/test_add_yum_repo_by_url.py
rcerven: changed

atomic_reactor/core.py
rcerven: done

atomic_reactor/constants.py
twaugh: Agreed, we shouldn't be retrying 404 for the Docker Engine API.

atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -81,24 +81,12 @@ def run(self):
...
-                                            .format(self.timeout))
-
-                self.log.info("will try again in %ss", self.retry_delay)
-                sleep(self.retry_delay)
-                continue
-
-            defer_removal(self.workflow, name)
-            break
+        # Pull the image from Crane
+        name = self.tasker.pull_image(pullspec, insecure=self.insecure)
```
twaugh: I think this ought to perform special handling for not_found, since it "knows" that it expects the content to appear at a later time.

atomic_reactor/core.py
```diff
@@ -211,8 +213,57 @@ def build_image_privileged_container(self, build_image, json_args_path):
...
+            self.retry_times = 1
+        try:
+            # docker-py 2.x
+            self.wrapped = docker.APIClient(**kwargs)
+        except AttributeError:
+            # docker-py 1.x
+            self.wrapped = docker.Client(**kwargs)
+
+    def __getattr__(self, attr):
+
```
twaugh: Unnecessary blank line.

atomic_reactor/core.py
```diff
@@ -230,12 +278,14 @@ def __init__(self, base_url=None, timeout=120, **kwargs):
...
             client_kwargs['version'] = 'auto'
 
-        try:
-            # docker-py 2.x
-            self.d = docker.APIClient(**client_kwargs)
-        except AttributeError:
-            # docker-py 1.x
-            self.d = docker.Client(**client_kwargs)
+        self.retry_times = retry_times
+        if not str(retry_times).isdigit() or retry_times == 0:
```
rcerven: well it can be passed as anything, so this just makes sure that wouldn't break retries, if for some reason somebody set that constant incorrectly

vrutkovs: I believe its a simple input sanitation - I'd rather have it during `self.retry_times` setup (as this value is currently not being sanitized), but this way is fine too

twaugh: But if it's a string, it will break, right?

tests/plugins/test_check_and_set_rebuild.py
rcerven: ok

atomic_reactor/constants.py
rcerven: ok

atomic_reactor/constants.py
```diff
@@ -70,3 +70,16 @@
...
+# max retries for docker requests
+HTTP_MAX_RETRIES = 3
+# how many seconds should wait before another try of docker request
+HTTP_BACKOFF_FACTOR = 5
+# docker client retries statuses
```
rcerven: yeah will change

atomic_reactor/constants.py
```diff
@@ -70,3 +70,16 @@
...
+HTTP_CLIENT_STATUS_RETRY = [404, 408, 500, 502, 503, 504]
```
rcerven: ok

atomic_reactor/core.py
```diff
@@ -211,8 +213,54 @@ def build_image_privileged_container(self, build_image, json_args_path):
...
+
+    for counter in range(retry_times):
+        if counter > 0:
+            sleep(retry_delay * (2 ** counter))
+            final_excep = None
+        try:
+            return function(*args, **kwargs)
+        except APIError as e:
+            if e.response.status_code in retry_client_statuses:
+                logger.info("retrying '%s' on '%s'", function, e.response.status_code)
```
rcerven: sure

atomic_reactor/core.py
```diff
@@ -211,8 +213,54 @@ def build_image_privileged_container(self, build_image, json_args_path):
...
+        try:
+            # docker-py 2.x
+            self.wrapped = docker.APIClient(**kwargs)
+        except AttributeError:
+            # docker-py 1.x
+            self.wrapped = docker.Client(**kwargs)
+
+    def __getattr__(self, attr):
+
+        orig_attr = self.wrapped.__getattribute__(attr)
```
rcerven: can do

atomic_reactor/core.py
rcerven: no it won't, this is why I have it there :)

atomic_reactor/util.py
```diff
@@ -953,3 +953,24 @@ class LabelFormatter(string.Formatter):
...
         return (self.get_value(field_name, args, kwargs), field_name)
+
+
+def do_retries(method, url, client_statuses=HTTP_CLIENT_STATUS_RETRY,
+               times=HTTP_MAX_RETRIES, delay=HTTP_BACKOFF_FACTOR, **kwargs):
+
+    if method not in ['post', 'get', 'put', 'delete']:
+        raise RuntimeError("Unsupported method '%s' for requests call!" % method)
+
+    if not str(times).isdigit():
```
rcerven: again to sanitize input, in case string was passed

atomic_reactor/util.py
```diff
@@ -953,3 +953,24 @@ class LabelFormatter(string.Formatter):
     """
     def get_field(self, field_name, args, kwargs):
         return (self.get_value(field_name, args, kwargs), field_name)
+
+
+def do_retries(method, url, client_statuses=HTTP_CLIENT_STATUS_RETRY,
```
rcerven: it can return session

I don't think you will be able to use 'raise_for_status' as you don't have request anymore but session, which raises by itself already

atomic_reactor/core.py
twaugh: But let's say it's `"4"`. The isdigit() test will succeed so we won't change it. Then we'll pass a string straight though... won't that break?

atomic_reactor/util.py
twaugh: `raise_for_status` is provided as a method on the Response class, so you can use it as above even when using a session.

atomic_reactor/core.py
rcerven: ahhh ok I see now what you mean, I'll change the last value with int()

atomic_reactor/util.py
```diff
@@ -947,3 +955,21 @@ class LabelFormatter(string.Formatter):
     """
     def get_field(self, field_name, args, kwargs):
         return (self.get_value(field_name, args, kwargs), field_name)
+
+
+def do_retries(client_statuses=HTTP_CLIENT_STATUS_RETRY,
+               times=HTTP_MAX_RETRIES, delay=HTTP_BACKOFF_FACTOR):
+
+    if not str(times).isdigit():
+        times = 0
```
vrutkovs: This line needs coverage

atomic_reactor/core.py
```diff
@@ -214,8 +216,56 @@ def build_image_privileged_container(self, build_image, json_args_path):
         return container_id
 
 
+def retry(function, *args, **kwargs):
```
vrutkovs: Actual retry functionality is not being tested - coveralls doesn't show `final_excep` line being triggered at all

atomic_reactor/util.py
rcerven: added

atomic_reactor/core.py
rcerven: added

tests/test_tasker.py
```diff
@@ -369,3 +374,33 @@ def version(self):
...
+    my_kwargs = {'one': 'first', 'two': 'second'}
+    assert args == my_args
+    assert kwargs == my_kwargs
+    response = requests.Response()
+    response.status_code = 408
+    raise APIError("test fail", response)
+
+
+@pytest.mark.parametrize('retry_times', [0, 1, 2, 3])
+def test_retry_method(retry_times):
```
vrutkovs: This test seems to verify that function is not being called when `retry_times == 0` and being called at least once in other cases. 

It should also verify that the wrapped function is being called several times, should check that timeout is calculated correctly, should check that the exception is propagated after a number of attempts etc.

rcerven: it is checking how many times wrapped function, so you mean it should check also the wrapper function? well if wrapped function is called 3 times, wrapper has to be called as well

vrutkovs:
Other plugins which need retry functions:
 * pre_add_filesystem
 * pre_add_yum_repo_by_url

tests/test_tasker.py
vrutkovs: It should at least check that the exception is thrown after a number of retries

rcerven: mmm? there is pytest.raises which is expecting APIError if retry_times is >0

rcerven:
added retry for : 

    pre_add_filesystem
    pre_add_yum_repo_by_url

atomic_reactor/util.py
```diff
@@ -947,3 +955,21 @@ class LabelFormatter(string.Formatter):
     """
     def get_field(self, field_name, args, kwargs):
         return (self.get_value(field_name, args, kwargs), field_name)
+
+
+def do_retries(client_statuses=HTTP_CLIENT_STATUS_RETRY,
```
twaugh: I'd prefer if this was called something more descriptive or specific. It sounds more generic than it is. How about `get_retrying_requests_session`?

tests/plugins/test_add_filesystem.py
```diff
@@ -36,6 +36,7 @@
     PreBuildPluginsRunner, PluginFailedException, BuildCanceledException)
 from atomic_reactor.plugins.pre_add_filesystem import AddFilesystemPlugin
 from atomic_reactor.util import ImageName, df_parser
+import atomic_reactor.util  # noqa
```
twaugh: `noqa` needed? If there's a particular warning to quell we ought to name it (e.g. "# noqa:E501")

tests/plugins/test_pulp_pull.py
```diff
@@ -49,6 +49,15 @@ def workflow(self, push=True, sync=True, build_process_failed=False):
...
             push_conf.add_pulp_registry('pulp', crane_uri=self.CRANE_URI, server_side_sync=True)
 
+        def custom_retries(*args, **kwargs):
+            kwargs['times'] = 0
+            return retry_fnc(*args, **kwargs)
+
+        retry_fnc = atomic_reactor.util.do_retries
+        (flexmock(atomic_reactor.util)
+            .should_receive('do_retries')
+            .replace_with(custom_retries))
```
twaugh: This is repeated a lot. I wonder if it should be somewhere common.

tests/plugins/test_pulp_pull.py
```diff
@@ -57,6 +66,8 @@ def workflow(self, push=True, sync=True, build_process_failed=False):
                         build_process_failed=build_process_failed,
                         plugin_workspace={})
 
+
+
```
twaugh: Too many blank lines?

tests/plugins/test_add_filesystem.py
rcerven: mmm maybe it isn't needed I've copy pasted it from somewhere

tests/plugins/test_pulp_pull.py
rcerven: yop

atomic_reactor/util.py
rcerven: yeah now it isn't as generic as it was before :)

tests/test_tasker.py
vrutkovs: Right, but does it check that exception is thrown after a specific number of attempts? It doesn't, but it should be, in order to verify that the request is being retried more than once.

atomic_reactor/util.py
vrutkovs: Since it returns `requests.Session` lets call it `get_requests_session` or similar, as later on we might add some more requests features here

tests/plugins/test_rpmqa.py
```diff
@@ -53,14 +52,14 @@ def mock_logs(cid, **kwargs):
...
-def test_rpmqa_plugin(remove_container_error, ignore_autogenerated):
+def test_rpmqa_plugin(docker_tasker, remove_container_error, ignore_autogenerated):
     if MOCK:
         should_raise_error = {}
         if remove_container_error:
             should_raise_error['remove_container'] = None
         mock_docker(should_raise_error=should_raise_error)
 
-    tasker = DockerTasker()
+    tasker = docker_tasker
```
vrutkovs: There is no point in additional variable here

atomic_reactor/core.py
```diff
@@ -214,8 +215,56 @@ def build_image_privileged_container(self, build_image, json_args_path):
...
+        except AttributeError:
+            # docker-py 1.x
+            self.wrapped = docker.Client(**kwargs)
+
+    def __getattr__(self, attr):
+        orig_attr = getattr(self.wrapped, attr)
+
+        if callable(orig_attr):
+            def hooked(*args, **kwargs):
+                return retry(orig_attr, *args, retry=self.retry_times, **kwargs)
```
lcarva: What if `retry` is given in the `kwargs` of the original function?
I believe python will complain about two values being given to the `retry` parameter.
Maybe this?
```python
kwargs.setdefault('retry', self.retry_times)
```
That way it gives the flexibility to overwrite the retry_times value for a specific call.

atomic_reactor/core.py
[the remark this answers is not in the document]
lcarva: Honestly, I think type checking shouldn't be done here at all unless it'll silently cause an issue later on which I don't think it's the case (`range('2')` causes `TypeError`).

We also don't do that for `timeout`, or any other param. IMO it just adds confusion to what the code is intended to do.

Additionally, `self.retry_times` should be allowed to be None, or 0 in this class. Adjustment of that value is something that should be handled by `WrappedDocker` as it's the one that actually uses it in a meaningful way.

tests/plugins/test_pulp_pull.py
rcerven: added to retry_mock

tests/test_tasker.py
rcerven: this is checking that function is called X times, and after that whole retry will raise

 (flexmock(sys.modules[__name__])
        .should_call('my_func')
        .with_args(*my_args, **my_kwargs)
        .times(retry_times)

atomic_reactor/core.py
```diff
@@ -214,8 +215,56 @@ def build_image_privileged_container(self, build_image, json_args_path):
...
+            else:
+                raise
+
+    if final_excep is not None:
+        raise final_excep
+
+
+class WrappedDocker(object):
+    def __init__(self, **kwargs):
+        self.retry_times = kwargs.pop('retry')
```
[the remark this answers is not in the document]
rcerven: when the retry is passed and is set to None, IT won't set it to 1 !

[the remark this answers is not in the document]
lcarva: Is `None` a valid value?

tests/plugins/test_rpmqa.py
rcerven: ok

atomic_reactor/core.py
```diff
@@ -214,8 +215,56 @@ def build_image_privileged_container(self, build_image, json_args_path):
...
+def retry(function, *args, **kwargs):
+    retry_times = kwargs.pop('retry', None)
+    retry_delay = DOCKER_BACKOFF_FACTOR
+    retry_client_statuses = DOCKER_CLIENT_STATUS_RETRY
+
+    final_excep = None
+
+    for counter in range(int(retry_times)):
+        if counter > 0:
+            time.sleep(retry_delay * (2 ** counter))
```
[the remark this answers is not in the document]
rcerven: ok

atomic_reactor/core.py
```diff
@@ -214,8 +215,56 @@ def build_image_privileged_container(self, build_image, json_args_path):
...
+        try:
+            return function(*args, **kwargs)
+        except APIError as e:
+            if e.response.status_code in retry_client_statuses:
+                logger.info("retrying %s on %s", function, e.response.status_code)
+                final_excep = e
+            else:
+                raise
+
+    if final_excep is not None:
```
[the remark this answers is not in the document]
rcerven: removed

atomic_reactor/core.py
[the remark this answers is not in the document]
rcerven: changed

atomic_reactor/core.py
rcerven: no need anymore

rcerven: if any argument which isn't accepted with docker.APIClient / Client will be given in kwargs
init will complain:
TypeError: __init__() got an unexpected keyword argument 'retry'

tests/test_util.py
```diff
@@ -654,3 +658,15 @@ def test_label_formatter(labels, test_string, expected):
...
             LabelFormatter().vformat(test_string, [], labels)
+
+
+@pytest.mark.parametrize(('retry', 'expected'), (
+    (5, 5),
+    (0, 0),
+    (1, 1),
+    (12, 12),
+))
+def test_retry_value(retry, expected):
```
vrutkovs: What does this test do? `expected` is not used anywhere

tests/test_tasker.py
```diff
@@ -369,3 +374,33 @@ def version(self):
...
+    my_args = ('some', 'new')
+    my_kwargs = {'one': 'first', 'two': 'second'}
+    assert args == my_args
+    assert kwargs == my_kwargs
+    response = requests.Response()
+    response.status_code = 408
+    raise APIError("test fail", response)
+
+
+@pytest.mark.parametrize('retry_times', [-1, 0, 1, 2, 3])
```
vrutkovs: Why is `retry_times==-1` being tested here? Retries are not part of public API, it cannot be negative (and a string btw) outside of atomic-reactor

rcerven: that is testing that function isn't run at all, when retry_times = 0, func will run 1 time, same logic as with session

tests/test_util.py
rcerven: I guess it is useless now, when we don't have any sanity check and not allowing strings, will delete that

atomic_reactor/core.py
```diff
@@ -214,8 +215,47 @@ def build_image_privileged_container(self, build_image, json_args_path):
         return container_id
 
 
+def retry(function, *args, **kwargs):
+    retry_times = kwargs.pop('retry', None)
+    retry_delay = DOCKER_BACKOFF_FACTOR
+    retry_client_statuses = DOCKER_CLIENT_STATUS_RETRY
+
+    for counter in range(int(retry_times)+1):
```
lcarva: If `retry_times` ends up being None due to `pop` call earlier, this will throw an exception.

It's a little messy to convert this parameter to `int` multiple times in this function.
Would it make more sense to default to `0` instead? This for-loop would still execute once as expected.

atomic_reactor/core.py
lcarva: That's not what I mean. If `orig_attr` is, let's say, `build` function of docker-py's `Client` class and it takes a `retry` parameter, we won't be able to use it.

AFAIK, there are no methods there that take a `retry` parameter so this is probably fine. But it's the kind of thing that increases difficulty in debugging.

atomic_reactor/plugins/post_import_image.py
```diff
@@ -29,7 +29,7 @@ class ImportImagePlugin(PostBuildPlugin):
 
     def __init__(self, tasker, workflow, imagestream, docker_image_repo,
                  url, build_json_dir, verify_ssl=True, use_auth=True,
-                 insecure_registry=None, retry_delay=30, import_attempts=3):
+                 insecure_registry=None):
```
lcarva: Let's remove reference to these params in docstring as well.

atomic_reactor/core.py
rcerven: ok will make it 0 by default and use 'int' just once

atomic_reactor/plugins/post_import_image.py
rcerven: done

atomic_reactor/core.py
rcerven: yeah docker method can't have 'retry' parameter, as we have no way passing 2 same named parameters to that new class, unless I'd make extra method just to set retry

tests/plugins/test_check_and_set_rebuild.py
```diff
@@ -156,4 +157,5 @@ def test_409_response(self, monkeypatch):
                 }
             }
         }))
-        runner.run()
+        with pytest.raises(PluginFailedException):
```
lcarva: Why is it raising an exception now? Sounds like we've changed behavior.

twaugh: The test scenario changed and now no longer correctly tests 409 retries. @rcerven, I think we just need to undo these changes. This retry-on-conflict testing really belongs in osbs-client in any case.

tests/plugins/test_delete_from_registry.py
```diff
@@ -209,10 +213,28 @@ def test_delete_from_registry_failures(tmpdir, status_codes):
             ok = True if status_codes == requests.codes.ACCEPTED else False
             url = "https://" + reg + "/v2/" + tag.split(":")[0] + "/manifests/" + dig
             auth_type = requests.auth.HTTPBasicAuth if req_registries[reg] else None
-            (flexmock(requests)
+
+            class myHTTPError(HTTPError):
```
twaugh: I don't understand why this needs a class. Isn't this sufficient?:
```
response = requests.Response()
response.status_code = status_codes  # also maybe this param should be named 'status_code'
(flexmock(requests.Session)
 ...
 .and_return(response))
```

tests/plugins/test_rpmqa.py
```diff
@@ -104,11 +102,9 @@ def test_rpmqa_plugin_exception(docker_tasker):  # noqa
         runner.run()
 
 
-def test_dangling_volumes_removed(request):
+def test_dangling_volumes_removed(docker_tasker, request):  # noqa
```
twaugh: Nit: ought to name the diagnostic code

tests/plugins/test_check_and_set_rebuild.py
lcarva: Ok I get why it's raising the exception now. Previously this test was testing the retry logic built into the plugin. Since that's gone and handled by `osbs-client`, it's no longer needed to test retry logic here.

The reason this test exists is because the plugin used to care about the 409 status code. It no longer does. I think it provides no value as is. Probably better to remove it, but I'm fine either way.

lcarva:
merge conflicts.

tests/plugins/test_delete_from_registry.py
rcerven: raise_for_status in that session response has to raise

rcerven: ok removed test for 999, and I can use response() now

lcarva:
Release notes updated.

PR #778, opened 2017-08-17
Implement group=True for group_manifests plugin

mlangsdorf:
Add support for creating manifest lists. Includes four commits:

    utils: allow get_manifest_digests() to return a digest from a manifest list
    
    get_manifest_digests now supports v1, v2, and list.v2.

and

    post_group_manifests: implement creation of the manifest list
    
    when group=True, group_manifest is meant to create a manifest list.
    It loops through all the platforms in the worker annotations and
    creates a manifest_list_spec, using that platform's worker image
    to create the image entry and the goarch information to create
    the platform information.
    
    The manifest_list_spec is then repeatedly retagged with the tags
    from tag_conf.images, written to a yaml file, and pushed via
    manifest-list. After all the images are tagged, the last image
    is used to create a manifest digest of the manifest list. The manifest
    digest is added to push_conf under the registry's name and the
    image's tag, so that delete_from_registry can delete the
    manifest list later.
    
    A list of the manifest digest from each registry is returned.

and

    group_manifests: add koji_import support for manifest lists
    
    If group_manifests created a manifest list, koji_import should create a dict
    in 'extras': 'index' that contains the pullspec and tags for the manifest
    list. Have koji_import check the return of group_manifests, and if it
    got a list of manifest digests, have it set up the 'extras': 'index'
    dictionary.

and

    delete_from_registry: delete manifest lists if they exist
    
    If group_manifests creates a manifest list, it stores the registry and tag
    in push_conf. If handle_worker_digests works correctly, we need to check to
    see if there is a push_conf registry. If there isn't, there's no need to
    post a warning.


twaugh:
I just spotted this: https://github.com/estesp/manifest-tool/pull/32

Since manifest-tool-0.6.0 you can add a list of additional tags! So we'd just need a single call to manifest-tool.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -46,6 +49,66 @@ def __init__(self, tasker, workflow, registries, group=True, goarch=None):
...
+                                                        insecure=insecure,
+                                                        dockercfg_path=secret_path,
+                                                        versions=('list',))
+            if not manifest_list_digest.list:
+                raise PluginFailedException('no manifest list digest for %s', registry)
+            self.log.debug('Digest for registry %s is %s', registry, manifest_list_digest.list)
+            push_conf_registry = self.workflow.push_conf.add_docker_registry(registry,
+                                                                             insecure=insecure)
+            push_conf_registry.config = get_config_from_registry(registry_image, registry, None,
+                                                                 insecure, secret_path, 'v2')
```
twaugh: This will fetch the 'config' blob for the architecture=amd64 image. Are we only fetching 'config' because of the way push_conf works?

The only thing atomic-reactor does with 'config' is store an edited version of it in the Koji build metadata, and the worker builds will already have taken care of that.

mlangsdorf: If we don't need the config, I am happy to not get it. I was just copying the code from post_tag_and_push but looking at the workflow, it doesn't look like anyone requires a config for the docker registries.

mlangsdorf: Fixed.

atomic_reactor/util.py
```diff
@@ -620,9 +620,10 @@ def get_credentials(self, docker_registry):
 class ManifestDigest(object):
     """Wrapper for digests for a docker manifest."""
 
-    def __init__(self, v1=None, v2=None):
+    def __init__(self, v1=None, v2=None, list=None):
```
mlangsdorf: Fixed.

tests/plugins/test_group_manifests.py
```diff
@@ -0,0 +1,286 @@
...
+
+        (flexmock(subprocess)
+         .should_receive("check_call"))
+        (flexmock(utils)
+         .should_receive("get_config_from_registry"))
+
+        runner = PostBuildPluginsRunner(tasker, workflow, plugins_conf)
+        if valid:
+            result = runner.run()
+            test_results = []
```
mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,211 @@
...
+                    'platform': {
+                        'os': 'linux',
+                        'architecture': self.goarch.get(platform, platform)
+                    }
+                }
+                manifest_list_spec['manifests'].append(arch_entry)
+
+            image_tags = []
+            for image in self.workflow.tag_conf.images:
+                image_tags.append(image.to_str(registry=False))
```
twaugh: We want just the suffix part here I think, not the repository name. Also, we only need to set additional tags other than the one used in 'image'.
```
registry_image = self.workflow.tag_conf.images[0].copy()
image_tags = [image.tag for image in self.workflow.tag_conf.images[1:]]
```

atomic_reactor/util.py
```diff
@@ -633,10 +634,12 @@ def default(self):
         In such case, the v1 schema should be used when interacting
         with the registry.
         """
-        return self.v2 or self.v1
+        return self.v2_list or self.v2 or self.v1
 
 
 def get_manifest_media_type(version):
+    if version == 'list':
```
twaugh: Maybe use 'v2_list' throughout, instead of 'list'? There's another type in the works as well, 'oci', so at some point we might need to define some constants to hold these names.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,212 @@
...
+#        dockercfg = "{0}/.dockercfg".format(registry_conf.get('secret', None))
+        
+        with NamedTemporaryFile(prefix='manifest-list', suffix=".yml", mode='w') as fp:
+            yaml.dump(manifest_list_spec, stream=fp)
+            fp.flush()
+            self.log.debug("Wrote to file %s with config %s", fp.name, docker_secret_path)
+
+            cmd = ['manifest-tool', '--docker-cfg=%s' % docker_secret_path,
+                   'push', 'from-spec', fp.name]
+            subprocess.check_call(cmd, env={'HOME': docker_secret_path})
```
twaugh: Could add a comment explaining that docker/cli (used by manifest-tool) only reads .dockercfg from $HOME.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,212 @@
...
+    def submit_manifest_list(self, registry, registry_conf, manifest_list_spec):
+        docker_secret_path = registry_conf.get('secret', None)
+#        dockercfg = "{0}/.dockercfg".format(registry_conf.get('secret', None))
+        
+        with NamedTemporaryFile(prefix='manifest-list', suffix=".yml", mode='w') as fp:
+            yaml.dump(manifest_list_spec, stream=fp)
+            fp.flush()
+            self.log.debug("Wrote to file %s with config %s", fp.name, docker_secret_path)
+
+            cmd = ['manifest-tool', '--docker-cfg=%s' % docker_secret_path,
```
twaugh: Do we think `--docker-cfg` is useful here at all, given how docker/cli behaves with .dockercfg?

It seems to me that only the newer-style 'config.json' file is read from the --docker-cfg location.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,212 @@
...
+                           Params:
+                            * "secret" optional string - path to the secret, which stores
+                              login and password for remote registry
+        :param group: bool, if true, create a manifest list; otherwise only add tags to
+                      amd64 image manifest
+        :param goarch: dict, keys are platform, values are go language platform names
+        """
+        # call parent constructor
+        super(GroupManifestsPlugin, self).__init__(tasker, workflow)
+        self.group = True # group
```
twaugh: Let's not hard-code it to True. :-)

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,212 @@
...
+                    'image': '{0}/{1}@{2}'.format(registry, repository, digest),
+                    'platform': {
+                        'os': 'linux',
+                        'architecture': self.goarch.get(platform, platform)
+                    }
+                }
+                manifest_list_spec['manifests'].append(arch_entry)
+
+            manifest_list_spec['tags'] = [image.tag for image in self.workflow.tag_conf.images]
+            # use a unique image tag because manifest_list doesn't like digests
```
twaugh: It's manifest-tool that doesn't like digests. I still don't understand why.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,212 @@
...
+            registry_image = self.workflow.tag_conf.unique_images[0]
+            registry_image.registry = registry
+            manifest_list_spec['image'] = registry_image.to_str()
+            self.log.info("Submitting manifest-list %s", manifest_list_spec)
```
twaugh: The manifest-tool documentation calls this input file a 'spec'. Let's call it that in the log message too, to avoid confusion with the actual 'manifest list' that manifest-tool creates as its output.

atomic_reactor/plugins/post_group_manifests.py
twaugh: Ah... we need to use a tag for the output we want manifest-tool to create because the output doesn't exist yet, so we can't very well know its digest.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -0,0 +1,212 @@
...
+        for registry, registry_conf in self.registries.items():
+            manifest_list_spec = {}
+            manifest_list_spec['manifests'] = []
+            all_annotations = self.workflow.build_result.annotations['worker-builds']
+            for platform in all_annotations:
+                worker_image = all_annotations[platform]['digests'][0]
+                digest = worker_image['digest']
+                repository = worker_image['repository']
+                arch_entry = {
+                    'image': '{0}/{1}@{2}'.format(registry, repository, digest),
```
twaugh: I think we need to use the unique tag instead of the digest here, and move the comment about manifest-tool not liking digests to here.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -149,6 +151,25 @@ def set_media_types(self, extra, worker_metadatas):
                 extra['image']['media_types'] = json.loads(annotations['media-types'])
                 return
 
+    def set_manifest_list_info(self, extra, worker_metadatas):
+        manifest_list_digests = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY)
+        # group_manifests returns either a list of manifest digests of the manifest lists,
+        # or a list of the image_manifests
```
twaugh: By the way, can we fix this? I think it should always return a list of manifest list digests, and that this list should be empty it didn't create any manifest list (i.e. when group=False). That way the test on the following line can be much simpler: `if manifest_list_digests:`.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -149,6 +151,25 @@ def set_media_types(self, extra, worker_metadatas):
...
+            repositories = self.workflow.build_result.annotations['repositories']['unique']
+            repo = ImageName.parse(repositories[0]).repo
+            # group_manifests added the registry, so this should be valid
+            for registry in self.workflow.push_conf.docker_registries:
+                pullspec = "{0}/{1}@{2}".format(registry.uri, repo,
+                                                manifest_list_digests[0].v2_list)
+                index['pull'] = [pullspec]
+                for tag in index['tags']:
+                    pullspec = "{0}/{1}:{2}".format(registry.uri, repo, tag)
+                    index['pull'].append(pullspec)
```
twaugh: We must not put "floating" tags here.  Ideally we should put the {version}-{release} tag here but no other tags, so it looks like:
pull: [ "registry.example.com/foo@sha256:...", "registry.example.com/foo:1.0-1" ]
tags: [ "1.0-1", "latest", ... ]

One way to arrange this is to use the knowledge that only the {version}-{release} tag may contain a dash character:
```
...
index['pull'] = [pullspec]
for tag in index['tags']:
    if '-' in tag:  # {version}-{release}
        index['pull'].append("{0}/{1}:{2}".format(registry.uri, repo, tag))
        break
```

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -47,6 +51,68 @@ def __init__(self, tasker, workflow, registries, group=True, goarch=None):
...
+        docker_secret_path = registry_conf.get('secret', None)
+        with NamedTemporaryFile(prefix='manifest-list', suffix=".yml", mode='w') as fp:
+            yaml.dump(manifest_list_spec, stream=fp)
+            fp.flush()
+            self.log.debug("Wrote to file %s with config %s", fp.name, docker_secret_path)
+
+            # --docker-cfg may be rendundant here, but it's how the tool should work
+            cmd = ['manifest-tool', '--docker-cfg=%s' % docker_secret_path,
+                   'push', 'from-spec', fp.name]
+            # docker always looks in $HOME for the .docker_cfg, so set $HOME to the path
```
twaugh: nit: typo: ".dockercfg"

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -47,6 +51,68 @@ def __init__(self, tasker, workflow, registries, group=True, goarch=None):
...
+                    'image': '{0}/{1}:{2}'.format(registry, repository, tag),
+                    'platform': {
+                        'os': 'linux',
+                        'architecture': self.goarch.get(platform, platform)
+                    }
+                }
+                manifest_list_spec['manifests'].append(arch_entry)
+
+            manifest_list_spec['tags'] = [image.tag for image in self.workflow.tag_conf.images]
+            # use a unique image tag because manifest_list can't accept a digest that
```
twaugh: Nit: typo: "... because manifest-tool can't... "

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -47,6 +51,68 @@ def __init__(self, tasker, workflow, registries, group=True, goarch=None):
...
+            # isn't in the respoistory yet
+            registry_image = self.workflow.tag_conf.unique_images[0]
+            registry_image.registry = registry
+            manifest_list_spec['image'] = registry_image.to_str()
+            self.log.info("Submitting manifest-list %s", manifest_list_spec)
```
twaugh: Can we say "spec" in this log message? Submitting manifest-list spec %s

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -149,6 +151,25 @@ def set_media_types(self, extra, worker_metadatas):
...
+        manifest_list_digests = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY)
+        # group_manifests returns either a list of manifest digests of the manifest lists,
+        # or a list of the image_manifests
+        if manifest_list_digests and not isinstance(manifest_list_digests[0], dict):
+            index = {}
+            index['tags'] = [image.tag for image in self.workflow.tag_conf.images]
+            repositories = self.workflow.build_result.annotations['repositories']['unique']
+            repo = ImageName.parse(repositories[0]).repo
+            # group_manifests added the registry, so this should be valid
+            for registry in self.workflow.push_conf.docker_registries:
```
twaugh: If Pulp is used, we only want to include the Pulp registry. Otherwise, we want to include all registries we pushed to. So it should be:

```
registries = self.workflow.push_conf.pulp_registries
if not registries:
for registry in registry:
```

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -47,6 +51,68 @@ def __init__(self, tasker, workflow, registries, group=True, goarch=None):
...
+            # --docker-cfg may be rendundant here, but it's how the tool should work
+            cmd = ['manifest-tool', '--docker-cfg=%s' % docker_secret_path,
+                   'push', 'from-spec', fp.name]
+            # docker always looks in $HOME for the .docker_cfg, so set $HOME to the path
+            subprocess.check_call(cmd, env={'HOME': docker_secret_path})
+            self.log.info("Manifest list submitted for %s", registry)
+
+    def get_grouped_manifests(self):
+        grouped_manifests = []
+        for registry, registry_conf in self.registries.items():
```
twaugh: Skip registries where `registry.get('version') == 'v1'`

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -47,6 +51,68 @@ def __init__(self, tasker, workflow, registries, group=True, goarch=None):
...
+                    'platform': {
+                        'os': 'linux',
+                        'architecture': self.goarch.get(platform, platform)
+                    }
+                }
+                manifest_list_spec['manifests'].append(arch_entry)
+
+            manifest_list_spec['tags'] = [image.tag for image in self.workflow.tag_conf.images]
+            # use a unique image tag because manifest_list can't accept a digest that
+            # isn't in the respoistory yet
```
twaugh: Nit: typo: repository

atomic_reactor/plugins/exit_koji_import.py

atomic_reactor/plugins/exit_koji_import.py

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: fixed.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: fixed.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: fixed.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: fixed.

atomic_reactor/plugins/exit_koji_import.py

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: fixed.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -47,6 +51,71 @@ def __init__(self, tasker, workflow, registries, group=True, goarch=None):
...
+        with NamedTemporaryFile(prefix='manifest-list', suffix=".yml", mode='w') as fp:
+            yaml.dump(manifest_list_spec, stream=fp)
+            fp.flush()
+            self.log.debug("Wrote to file %s with config %s", fp.name, docker_secret_path)
+
+            # --docker-cfg may be rendundant here, but it's how the tool should work
+            cmd = ['manifest-tool', '--docker-cfg=%s' % docker_secret_path,
+                   'push', 'from-spec', fp.name]
+            # docker always looks in $HOME for the .dockercfg, so set $HOME to the path
+            subprocess.check_call(cmd, env={'HOME': docker_secret_path})
```
lcarva: In the case this command fails, will we see any output from the tool?
This would be ideal for debugging purposes.

mlangsdorf: check_call raises a CalledProcessError on failure which gets caught normally and any error message displayed. we don't have to do anything special.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -149,6 +151,30 @@ def set_media_types(self, extra, worker_metadatas):
...
+            repositories = self.workflow.build_result.annotations['repositories']['unique']
+            repo = ImageName.parse(repositories[0]).repo
+            # group_manifests added the registry, so this should be valid
+            registries = self.workflow.push_conf.pulp_registries
+            if not registries:
+                registries = self.workflow.push_conf.all_registries
+            for registry in registries:
+                pullspec = "{0}/{1}@{2}".format(registry.uri, repo,
+                                                manifest_list_digests[0].v2_list)
+                index['pull'] = [pullspec]
```
mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
vrutkovs: `CalledProcessError` would not show process stdout. A better idea would be:
* use `check_output`
* catch `CalledProcessError as exc`
* output `exc.output`

mlangsdorf: fixed.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -47,6 +52,76 @@ def __init__(self, tasker, workflow, registries, group=True, goarch=None):
...
+
+            # --docker-cfg may be rendundant here, but it's how the tool should work
+            cmd = ['manifest-tool', '--docker-cfg=%s' % docker_secret_path,
+                   'push', 'from-spec', fp.name]
+            # docker always looks in $HOME for the .dockercfg, so set $HOME to the path
+            # subprocess.check_call(cmd, env={'HOME': docker_secret_path})
+            try:
+                subprocess.check_call(cmd)
+            except CalledProcessError as exc:
+                self.log.error("manifest-tool failed with %s", exc.output)
```
vrutkovs: `check_output(cmd)` must be used to have `output` in exception, `check_call` won't cut it

mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -47,6 +51,75 @@ def __init__(self, tasker, workflow, registries, group=True, goarch=None):
...
+
+            # --docker-cfg may be rendundant here, but it's how the tool should work
+            cmd = ['manifest-tool', '--docker-cfg=%s' % docker_secret_path,
+                   'push', 'from-spec', fp.name]
+            # docker always looks in $HOME for the .dockercfg, so set $HOME to the path
+            try:
+                check_output(cmd, stderr=STDOUT, env={'HOME': docker_secret_path})
+            except CalledProcessError as exc:
+                self.log.error("manifest-tool failed with %s", exc.output)
+                raise RuntimeError
```
lcarva: How about just re-raising the previous exception?
As it is, the exception that gets propagated contains no useful information.

mlangsdorf: Fixed.

tests/plugins/test_koji_import.py
```diff
@@ -1345,3 +1350,46 @@ def test_koji_import_set_media_types(self, tmpdir, os_env, version):
...
+        extra = build['extra']
+        assert isinstance(extra, dict)
+        assert 'image' in extra
+        image = extra['image']
+        assert isinstance(image, dict)
+        if digests:
+            assert 'index' in image.keys()
+            expected_results = {}
+            expected_results['tags'] = [tag.tag for tag in workflow.tag_conf.images]
+            pullspec = "docker.example.com/hello-world@{0}".format(digests[0].v2_list)
```
twaugh: This is incorrect! It should be "docker.example.com/myproject/hello-world@{0}". Without the namespace part included, the image is not pullable.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -173,6 +173,29 @@ def set_media_types(self, extra, worker_metadatas):
                 extra['image']['media_types'] = json.loads(annotations['media-types'])
                 return
 
+    def set_manifest_list_info(self, extra, worker_metadatas):
+        manifest_list_digests = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY)
+        if manifest_list_digests:
+            index = {}
+            index['tags'] = [image.tag for image in self.workflow.tag_conf.images]
+            repositories = self.workflow.build_result.annotations['repositories']['unique']
+            repo = ImageName.parse(repositories[0]).repo
```
twaugh: This should be: `repo = ImageName.parse(repositories[0]).to_str(registry=False, tag=False)`.

mlangsdorf: Fixed.

tests/plugins/test_koji_import.py
mlangsdorf: Fixed.

tests/plugins/test_group_manifests.py
```diff
@@ -173,16 +275,17 @@ def test_group_manifests_unimplemented(self, tmpdir):
...
             mock_docker()
 
         if version == '1':
             valid = False
 
-        test_images = ['namespace/httpd:2.4', 'namespace/sshd:2.4']
-        test_results = [{'tag': 'testtag', 'schemaVersion': '2'}]
+        test_images = ['registry.example.com/namespace/httpd:2.4',
+                       'registry.example.com/namespace/httpd:latest']
+#        test_results = [{'tag': 'testtag', 'schemaVersion': '2'}]
```
twaugh: Not needed?

mlangsdorf: Fixed.

lcarva:
Updated release notes.

PR #779, opened 2017-08-18
Logging: try to disentangle overlapping log entries

jarodwilson:
We have a few issues with the way we do logging right now. For the most part, the crux of the problem seems to be that python only serializes log entries from a single process when logging to a file or file-like handler, such as the StreamHandler we currently log to. Multiple threads are fine, but when you fire off a subprocess call, all bets are off, and we do actually have a fair number of those throughout the code, which appear to be the primary offender for getting themselves logged right in the middle of other log entries.

### DIFF
--- a/atomic_reactor/__init__.py
+++ b/atomic_reactor/__init__.py
@@ -50,6 +50,9 @@ class EncodedStream(object):
             self.binarystream.write(text.encode(self.encoding))
         else:
             self.binarystream.write(text)
+        # We need to flush regularly, because launching plugins or running
+        # subprocess calls breaks serialization of logging output otherwise
+        self.binarystream.flush()
 
     def __del__(self):
         try:
--- a/atomic_reactor/__init__.py
+++ b/atomic_reactor/__init__.py
@@ -86,5 +86,8 @@ def set_logging(name="atomic_reactor", level=logging.DEBUG, handler=None):
     # add ch to logger
     logger.addHandler(handler)
 
+    logger = logging.getLogger('osbs')
+    for hdlr in list(logger.handlers):  # make a copy so it doesn't change
+        hdlr.setFormatter(formatter)
 
 set_logging(level=logging.WARNING)  # override this however you want


jarodwilson:
Note: if the .flush() route doesn't work, we're probably looking at having to implement a SocketHandler logger, some of which is detailed at https://docs.python.org/2/howto/logging-cookbook.html#network-logging

atomic_reactor/buildimage.py
```diff
@@ -112,6 +113,7 @@ def get_reactor_tarball_path(self, tmpdir):
         os.chdir(local_reactor_git_path)
         try:
             logger.debug("executing sdist command in directory '%s'", os.getcwd())
+            sys.stderr.flush()
```
twaugh: Rather than flushing sys.stderr after every logging call, wouldn't it be more robust to modify EncodedStream (from __init__.py) so it calls `self.binarystream.flush()` after writing?

atomic_reactor/plugins/post_push_to_pulp.py
```diff
@@ -127,7 +128,9 @@ def push_tar(self, filename, image_names=None, repo_prefix="redhat-"):
                 cmd = "set -o pipefail; {0} {1} | tar --delete {2} | gzip - > {3}".format(
                     unpacker, filename, ' '.join(remove_layers), outfile.name)
                 self.log.debug("running %s", cmd)
+                sys.stderr.flush()
                 subprocess.check_call(cmd, shell=True)
+                sys.stderr.flush()
```
twaugh: At this point, the subprocess has exited and so anything it wrote to stderr has definitely been flushed.

atomic_reactor/buildimage.py
jarodwilson: My initial thought was "only flush when we have to", rather than always flushing, thinking there could be a performance impact, but that does seem a lot more fragile with how many seems like they might be necessary, so centralizing it is probably the way to go. I'll adjust accordingly.

tests/test_cli.py
```diff
@@ -197,9 +197,14 @@ def test_log_encoding(self, caplog, monkeypatch):
                 self.exec_cli(command)
 
         # first message should be 'log encoding: <encoding>'
-        match = caplog.records()[0].message.split(':')
+        match = None
+        for record in caplog.records():
```
vrutkovs: Why not check first message received here? Log encoding should be the first record, this test was designed to ensure that

twaugh: Agreed. I think we should drop the 'flexibility when testing for log encoding' commit.

jarodwilson: Wasn't sure about that one myself, haven't hit a problem with it personally. Okay, have dropped that commit from the branch, leaving just the bit to force log format (I've seen things from osbs logged w/o the proper format) and the flush.

twaugh:
No longer [WIP], right?

tests/test_cli.py
dzickusrh: I hit this problem when playing with the root logger.  Basically I enabled DEBUG on root logger, which enabled DEBUG logging for every python library.  In this case, Docker debug messages popped up before the 'log encoding' and the test failed.  I agree this is overkill, however I didn't think we could guarantee that no other 'info' message from say docker would ever pop up first.

Feel free to nak it, I don't mind.  Just wanted to state my original reason.

twaugh: Alright. If it turns out to be a problem we can address that separately.

twaugh:
Draft release notes updated.

PR #780, opened 2017-08-22
Update docker image ID for x86_64 if pulp_pull has ran before

vrutkovs:


atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -98,9 +98,15 @@ def get_output(self, worker_metadatas):
         :return: list, containing dicts of partial metadata
         """
         outputs = []
+        has_pulp_pull = PLUGIN_PULP_PULL_KEY in self.workflow.postbuild_results
         for platform in worker_metadatas:
             for instance in worker_metadatas[platform]['output']:
                 instance['buildroot_id'] = '{}-{}'.format(platform, instance['buildroot_id'])
+
+                # update image ID with pulp_pull results
+                if 'extra' in instance and platform == "x86_64" and has_pulp_pull:
```
twaugh: I'd have preferred to pick the worker build whose goarch is "amd64", but since it's awkward to get at it that way I'm fine with doing it via platform name.

vrutkovs: IIRC `koji_upload` expects platform name to be the architecture of the build, so its safe to rely on platform (although indeed feels bad). Using goarch for that would bump the required osbs-client version

twaugh:
Draft release notes updated.

PR #781, opened 2017-08-22
Only publish each Pulp repository once

twaugh:

### DIFF
--- a/atomic_reactor/plugins/exit_pulp_publish.py
+++ b/atomic_reactor/plugins/exit_pulp_publish.py
@@ -81,7 +81,8 @@ class PulpPublishPlugin(ExitPlugin):
         self.pulp_handler.create_dockpulp()
         if not repo_prefix:
             repo_prefix = ''
-        pulp_repos = ['%s%s' % (repo_prefix, image.pulp_repo) for image in image_names]
+        pulp_repos = set(['%s%s' % (repo_prefix, image.pulp_repo)
+                          for image in image_names])
         self.pulp_handler.publish(pulp_repos)
 
         pulp_registry = self.pulp_handler.get_registry_hostname()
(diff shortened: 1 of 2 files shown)


twaugh:
@mlangsdorf

atomic_reactor/plugins/exit_pulp_publish.py
```diff
@@ -81,8 +81,9 @@ def publish_to_crane(self, repo_prefix="redhat-"):
         self.pulp_handler.create_dockpulp()
         if not repo_prefix:
             repo_prefix = ''
-        pulp_repos = ['%s%s' % (repo_prefix, image.pulp_repo) for image in image_names]
-        self.pulp_handler.publish(pulp_repos)
+        pulp_repos = set(['%s%s' % (repo_prefix, image.pulp_repo)
```
lcarva: Out of curiosity, when will this list contain more than one item?

atomic_reactor/plugins/exit_pulp_publish.py
```diff
@@ -81,8 +81,9 @@ def publish_to_crane(self, repo_prefix="redhat-"):
...
+                          for image in image_names])
+        self.pulp_handler.publish(sorted(list(pulp_repos)))
```
vrutkovs: Not sure if `sorted` and `list` are required - dockpulp's `crane` seems to accept any iterable

twaugh: Right you are. Updated.

atomic_reactor/plugins/exit_pulp_publish.py
twaugh: It iterates through the image names, so there is an `image.pulp_repo` for each tag we made.

lcarva: Sure, but isn't the pulp_repo name derived from the `name` label in Dockerfile?
`pulp_repo` does not include the registry, nor the tag.

twaugh:
No release note.

atomic_reactor/plugins/exit_pulp_publish.py
twaugh: It is. So yes, we expect the resulting set to have a single element.

PR #782, opened 2017-08-22
koji_import: use get_orchestrator_build_logs in plugin

jarodwilson:
Make use of osbs-client's new get_orchestrator_build_logs api to acquire
orchestrator and per-arch log files when importing a build.

Signed-off-by: Jarod Wilson <jarod@redhat.com>

### DIFF
--- a/atomic_reactor/koji_util.py
+++ b/atomic_reactor/koji_util.py
@@ -9,6 +9,7 @@ of the BSD license. See the LICENSE file for details.
 from __future__ import print_function
 
 
+from collections import namedtuple
 import koji
 import logging
 import os
@@ -18,6 +19,28 @@ from atomic_reactor.constants import DEFAULT_DOWNLOAD_BLOCK_SIZE
 
 
 logger = logging.getLogger(__name__)
+Output = namedtuple('Output', ['file', 'metadata'])
+
+
+class KojiUploadLogger(object):
+    def __init__(self, logger, notable_percent=10):
+        self.logger = logger
+        self.notable_percent = notable_percent
+        self.last_percent_done = 0
+
+    def callback(self, offset, totalsize, size, t1, t2):  # pylint: disable=W0613
+        if offset == 0:
+            self.logger.debug("upload size: %.1fMiB", totalsize / 1024.0 / 1024)
+
+        if not totalsize or not t1:
+            return
+
+        percent_done = 100 * offset / totalsize
+        if (percent_done >= 99 or
+                percent_done - self.last_percent_done >= self.notable_percent):
+            self.last_percent_done = percent_done
+            self.logger.debug("upload: %d%% done (%.1f MiB/sec)",
+                              percent_done, size / t1 / 1024 / 1024)
 
 
 def koji_login(session,
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -9,7 +9,9 @@ of the BSD license. See the LICENSE file for details.
 from __future__ import unicode_literals
 
 import json
+import os
 import time
+from tempfile import NamedTemporaryFile
 
 from atomic_reactor import start_time as atomic_reactor_start_time
 from atomic_reactor.plugin import ExitPlugin
@@ -23,10 +25,12 @@ from atomic_reactor.constants import (PLUGIN_KOJI_IMPORT_PLUGIN_KEY,
                                       PLUGIN_PULP_SYNC_KEY,
                                       PLUGIN_FETCH_WORKER_METADATA_KEY,
                                       PLUGIN_GROUP_MANIFESTS_KEY)
-from atomic_reactor.util import (get_build_json, get_preferred_label, df_parser, ImageName)
-from atomic_reactor.koji_util import create_koji_session
+from atomic_reactor.util import (get_build_json, get_preferred_label,
+                                 df_parser, ImageName, get_checksums)
+from atomic_reactor.koji_util import (create_koji_session, Output, KojiUploadLogger)
 from osbs.conf import Configuration
 from osbs.api import OSBS
+from osbs.exceptions import OsbsException
 
 
 class KojiImportPlugin(ExitPlugin):
@@ -58,6 +62,7 @@ class KojiImportPlugin(ExitPlugin):
                  verify_ssl=True, use_auth=True,
                  koji_ssl_certs=None, koji_proxy_user=None,
                  koji_principal=None, koji_keytab=None,
+                 blocksize=None,
                  target=None, poll_interval=5):
         """
         constructor
@@ -72,6 +77,7 @@ class KojiImportPlugin(ExitPlugin):
         :param koji_proxy_user: str, user to log in as (requires hub config)
         :param koji_principal: str, Kerberos principal (must specify keytab)
         :param koji_keytab: str, keytab name (must specify principal)
+        :param blocksize: int, blocksize to use for uploading files
         :param target: str, koji target
         :param poll_interval: int, seconds between Koji task status requests
         """
@@ -84,6 +90,7 @@ class KojiImportPlugin(ExitPlugin):
         self.koji_principal = koji_principal
         self.koji_keytab = koji_keytab
 
+        self.blocksize = blocksize
         self.target = target
         self.poll_interval = poll_interval
 
@@ -148,6 +155,46 @@ class KojiImportPlugin(ExitPlugin):
                 buildroots.append(instance)
 
         return buildroots
+
+    def get_logs(self):
+        """
+        Build list of log files
+
+        :return: list, of log files
+        """
+
+        logs = None
+        output = []
+
+        # Collect logs from server
+        try:
+            logs = self.osbs.get_orchestrator_build_logs(self.build_id)
+        except OsbsException as ex:
+            self.log.error("unable to get build logs: %r", ex)
+            return output
+        except TypeError:
+            # Older osbs-client has no get_orchestrator_build_logs
+            self.log.error("OSBS client does not support get_orchestrator_build_logs")
+            return output
+
+        platform_logs = {}
+        for entry in logs:
+            platform = entry.platform
+            if platform not in platform_logs:
+                filename = 'orchestrator' if platform is None else platform
+                platform_logs[platform] = NamedTemporaryFile(prefix="%s-%s" %
+                                                             (self.build_id, filename),
+                                                             suffix=".log", mode='wb')
+            platform_logs[platform].write(entry.line.encode('utf-8'))
+
+        for platform, logfile in platform_logs.items():
+            logfile.flush()
+            filename = 'orchestrator' if platform is None else platform
+            metadata = self.get_output_metadata(logfile.name, "%s.log" % filename)
+            output.append(Output(file=logfile, metadata=metadata))
+
+        return output
+
 
     def set_help(self, extra, worker_metadatas):
         all_annotations = [get_worker_build_info(self.workflow, platform).build.get_annotations()
@@ -217,6 +264,21 @@ class KojiImportPlugin(ExitPlugin):
                             self.log.debug("reset tags to so that docker is %s",
                                            instance['extra']['docker'])
 
+    def get_output_metadata(self, path, filename):
+        """
+        Describe a file by its metadata.
+
+        :return: dict
+        """
+
+        checksums = get_checksums(path, ['md5'])
+        metadata = {'filename': filename,
+                    'filesize': os.path.getsize(path),
+                    'checksum': checksums['md5sum'],
+                    'checksum_type': 'md5'}
+
+        return metadata
+
     def get_build(self, metadata, worker_metadatas):
         start_time = int(atomic_reactor_start_time)
 
@@ -272,6 +334,16 @@ class KojiImportPlugin(ExitPlugin):
         return build
 
     def combine_metadata_fragments(self):
+        def add_buildroot_id(output, buildroot_id):
+            logfile, metadata = output
+            metadata.update({'buildroot_id': buildroot_id})
+            return Output(file=logfile, metadata=metadata)
+
+        def add_log_type(output):
+            logfile, metadata = output
+            metadata.update({'type': 'log', 'arch': 'noarch'})
+            return Output(file=logfile, metadata=metadata)
+
         try:
             metadata = get_build_json()["metadata"]
             self.build_id = metadata["name"]
@@ -284,7 +356,11 @@ class KojiImportPlugin(ExitPlugin):
         worker_metadatas = self.workflow.postbuild_results.get(PLUGIN_FETCH_WORKER_METADATA_KEY)
         build = self.get_build(metadata, worker_metadatas)
         buildroot = self.get_buildroot(worker_metadatas)
+        buildroot_id = buildroot[0]['id']
         output = self.get_output(worker_metadatas)
+        output_files = [add_log_type(add_buildroot_id(md, buildroot_id))
+                        for md in self.get_logs()]
+        output.extend([of.metadata for of in output_files])
 
         koji_metadata = {
             'metadata_version': metadata_version,
@@ -292,7 +368,7 @@ class KojiImportPlugin(ExitPlugin):
             'buildroots': buildroot,
             'output': output,
         }
-        return koji_metadata
+        return koji_metadata, output_files
 
     def login(self):
         """
@@ -309,6 +385,28 @@ class KojiImportPlugin(ExitPlugin):
             "krb_keytab": str(self.koji_keytab)
         }
         return create_koji_session(str(self.kojihub), auth_info)
+
+    def upload_file(self, session, output, serverdir):
+        """
+        Upload a file to koji
+
+        :return: str, pathname on server
+        """
+        name = output.metadata['filename']
+        self.log.debug("uploading %r to %r as %r",
+                       output.file.name, serverdir, name)
+
+        kwargs = {}
+        if self.blocksize is not None:
+            kwargs['blocksize'] = self.blocksize
+            self.log.debug("using blocksize %d", self.blocksize)
+
+        upload_logger = KojiUploadLogger(self.log)
+        session.uploadWrapper(output.file.name, serverdir, name=name,
+                              callback=upload_logger.callback, **kwargs)
+        path = os.path.join(serverdir, name)
+        self.log.debug("uploaded %r", path)
+        return path
 
     def run(self):
         """
@@ -329,7 +427,16 @@ class KojiImportPlugin(ExitPlugin):
 
         server_dir = get_koji_upload_dir(self.workflow)
 
-        koji_metadata = self.combine_metadata_fragments()
+        koji_metadata, output_files = self.combine_metadata_fragments()
+
+        try:
+            for output in output_files:
+                if output.file:
+                    self.upload_file(session, output, server_dir)
+        finally:
+            for output in output_files:
+                if output.file:
+                    output.file.close()
 
         try:
             build_info = session.CGImport(koji_metadata, server_dir)
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -8,7 +8,6 @@ of the BSD license. See the LICENSE file for details.
 
 from __future__ import unicode_literals
 
-from collections import namedtuple
 import json
 import os
 import random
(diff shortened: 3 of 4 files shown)


atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -37,6 +37,12 @@
 from osbs.api import OSBS
 from osbs.exceptions import OsbsException
 
+have_get_orchestrator_build_logs = True
+try:
+    from osbs import get_orchestrator_build_logs
+except (ImportError, SyntaxError):
```
twaugh: Don't think we need to trap SyntaxError here. For dockpulp we did that just because dockpulp isn't valid Python 3.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -326,43 +332,93 @@ def get_buildroot(self, build_id):
...
+
+        :return: list, Output instances
+        """
+        output = []
+
+        # Build list of arches
+        arches = []
+        for log in logs:
+            if log.platform is not None and log.platform not in arches:
+                arches.append(log.platform)
```
twaugh: Could also write this as `arches = set([log.platform for log in logs if log.platform is not None])`

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -37,6 +37,12 @@
 from osbs.api import OSBS
 from osbs.exceptions import OsbsException
 
+have_get_orchestrator_build_logs = True
+try:
+    from osbs import get_orchestrator_build_logs
```
twaugh: This isn't a module-level function though, it's a method on a class instance (instances of osbs.api.OSBS).

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -326,43 +332,93 @@ def get_buildroot(self, build_id):
...
         """
 
+        logs = None
         output = []
 
         # Collect logs from server
         try:
-            logs = self.osbs.get_build_logs(self.build_id)
+            if have_get_orchestrator_build_logs:
+                logs = get_orchestrator_build_logs(self.build_id)
```
twaugh: Maybe this should be a `try:`/`except AttributeError:` instead?
```
try:
    logs = self.osbs.get_orchestrator_build_logs(self.build_id)
except AttributeError:
    # get_orchestrator_build_logs was introduced in ...
    logs = self.osbs.get_build_logs(...)
```

twaugh: Maybe do the parsing inside each block too.

atomic_reactor/plugins/exit_koji_promote.py
twaugh: In fact, we should avoid iterating more than once. It's a generator so further iterations will produce no items.

How about
```
platform_logs = {}
for log in logs:
    platform_logs.setdefault(log.platform, [])
    platform_logs[log.platform].append(log.line)
```
etc

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -326,43 +332,93 @@ def get_buildroot(self, build_id):
...
+        """
+        output = []
+
+        # Deleted once closed
+        logfile = NamedTemporaryFile(prefix=self.build_id, suffix=".log", mode='wb')
+        try:
+            logfile.write(logs)
+        except (TypeError, UnicodeEncodeError):
+            # Older osbs-client versions returned Unicode objects
+            logfile.write(logs.encode('utf-8'))
```
vrutkovs: Maybe it should first try to write UTF-8 encoded line and then fall back to `write(logs)`?

twaugh: I think this is pre-existing code which was just moved around.

jarodwilson: Correct, this was just some pre-existing code that got moved around a little bit.

jarodwilson:
So I had a suspicion I might not be putting this stuff in the right place. Not quite sure yet where to shoehorn this into koji_import at though -- possibly into get_output()?

twaugh:
Yes, `get_output()` is probably as good a place as any.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -20,10 +23,35 @@
...
 from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
 from atomic_reactor.constants import PLUGIN_KOJI_IMPORT_PLUGIN_KEY
-from atomic_reactor.util import (get_build_json, get_preferred_label, df_parser)
+from atomic_reactor.util import (get_build_json, get_preferred_label, df_parser, get_checksums)
 from atomic_reactor.koji_util import create_koji_session
 from osbs.conf import Configuration
 from osbs.api import OSBS
+from osbs.exceptions import OsbsException
+
+LogEntry = namedtuple('LogEntry', ['platform', 'line'])
```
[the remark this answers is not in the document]
jarodwilson: Oops, stray relic. Removing.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -20,10 +23,35 @@
...
+Output = namedtuple('Output', ['file', 'metadata'])
+
+
+class KojiUploadLogger(object):
```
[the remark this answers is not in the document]
jarodwilson: Yeah, was thinking that was the way to go as well, since it's 100% copy-n-paste, just wasn't sure if it was the right thing to copy-n-paste yet. I'll go ahead and make that change.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -55,6 +83,7 @@ def __init__(self, tasker, workflow, kojihub, url,
                  verify_ssl=True, use_auth=True,
                  koji_ssl_certs=None, koji_proxy_user=None,
                  koji_principal=None, koji_keytab=None,
+                 metadata_only=False, blocksize=None,
```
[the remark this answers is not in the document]
jarodwilson: Done.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -69,6 +98,8 @@ def __init__(self, tasker, workflow, kojihub, url,
         :param koji_proxy_user: str, user to log in as (requires hub config)
         :param koji_principal: str, Kerberos principal (must specify keytab)
         :param koji_keytab: str, keytab name (must specify principal)
+        :param metadata_only: bool, whether to omit the 'docker save' image
```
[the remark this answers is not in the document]
jarodwilson: Done.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -81,6 +112,8 @@ def __init__(self, tasker, workflow, kojihub, url,
         self.koji_principal = koji_principal
         self.koji_keytab = koji_keytab
 
+        self.metadata_only = metadata_only
```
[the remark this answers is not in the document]
jarodwilson: Done.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -119,6 +152,65 @@ def get_buildroot(self, worker_metadatas):
...
+            logs = self.osbs.get_orchestrator_build_logs(self.build_id)
+        except OsbsException as ex:
+            self.log.error("unable to get build logs: %r", ex)
+            return output
+        except TypeError:
+            # Older osbs-client has no get_orchestrator_build_logs
+            self.log.error("Your osbs client doesn't support this functionality")
+            return output
+
+        self.log.info("logs=%s", logs)
```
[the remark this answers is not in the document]
jarodwilson: Stray debug logging, oops.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -119,6 +152,65 @@ def get_buildroot(self, worker_metadatas):
...
+        platform_logs = {}
+        for entry in logs:
+            self.log.info("platform: %s, line: %s", entry.platform, entry.line)
```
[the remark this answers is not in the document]
jarodwilson: More stray logging added while debugging things getting it to a working state, will remove.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -119,6 +152,65 @@ def get_buildroot(self, worker_metadatas):
...
+            for entry in platform_logs[arch]:
+                self.log.info("platform_logs[%s]: %s", arch, entry)
+                logfile.write(entry)
+            logfile.flush()
+            metadata = self.get_output_metadata(logfile.name, "%s.log" % arch)
+            output.append(Output(file=logfile, metadata=metadata))
+
+        logfile = NamedTemporaryFile(prefix="%s-orchestrator" % self.build_id,
+                                     suffix=".log", mode='wb')
+        orchestrator_logs = (log.line for log in logs if log.platform is None)
```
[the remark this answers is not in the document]
jarodwilson: Ah. Right. An earlier version of the code built an arches list that excluded None, so that was added for the None logs, but now we're building a list that includes None... Okay, I'll figure out something for that shortly here.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -143,6 +235,24 @@ def set_media_types(self, extra, worker_metadatas):
...
+        """
+
+        checksums = get_checksums(path, ['md5'])
+        metadata = {'filename': filename,
+                    'filesize': os.path.getsize(path),
+                    'checksum': checksums['md5sum'],
+                    'checksum_type': 'md5'}
+
+        if self.metadata_only:
+            metadata['metadata_only'] = True
```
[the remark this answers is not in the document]
jarodwilson: Removed.

lcarva:
merge conflicts

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -149,6 +156,55 @@ def get_buildroot(self, worker_metadatas):
...
+
+        # Collect logs from server
+        try:
+            logs = self.osbs.get_orchestrator_build_logs(self.build_id)
+        except OsbsException as ex:
+            self.log.error("unable to get build logs: %r", ex)
+            return output
+        except TypeError:
+            # Older osbs-client has no get_orchestrator_build_logs
+            self.log.error("Your osbs client doesn't support this functionality")
```
lcarva: Maybe it's worth describing what the functionality is?

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -149,6 +156,55 @@ def get_buildroot(self, worker_metadatas):
...
+            return output
+
+        platform_logs = {}
+        for entry in logs:
+            platform_logs.setdefault(entry.platform, [])
+            platform_logs[entry.platform].append(entry.line)
```
lcarva: Should we be concerned with loading the whole log into memory?
Not sure if writing each line per iteration is needed.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -149,6 +156,55 @@ def get_buildroot(self, worker_metadatas):
...
+                                         suffix=".log", mode='wb')
+            for entry in platform_logs[arch]:
+                logfile.write(entry)
+            logfile.flush()
+            metadata = self.get_output_metadata(logfile.name, "%s.log" % filename)
+            output.append(Output(file=logfile, metadata=metadata))
+
+        docker_logs = NamedTemporaryFile(prefix="docker-%s" % self.build_id,
+                                         suffix=".log", mode='wb')
+        docker_logs.write("\n".join(self.workflow.build_result.logs).encode('utf-8'))
```
lcarva: Isn't this empty on orchestrator builds?

`koji_upload` currently handles this file. We should probably change it there so it reports a build.log for each platform.

atomic_reactor/util.py
```diff
@@ -34,6 +35,28 @@
 from requests.utils import guess_json_utf
 
 logger = logging.getLogger(__name__)
+Output = namedtuple('Output', ['file', 'metadata'])
+
+
+class KojiUploadLogger(object):
```
lcarva: Maybe `koji_util.py` is a better home?

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -211,14 +331,16 @@ def combine_metadata_fragments(self):
...
         buildroot = self.get_buildroot(worker_metadatas)
         output = self.get_output(worker_metadatas)
+        output_files = [add_log_type(add_buildroot_id(md))
+                        for md in self.get_logs()]
 
         koji_metadata = {
             'metadata_version': metadata_version,
             'build': build,
             'buildroots': buildroot,
             'output': output,
```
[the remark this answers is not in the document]
jarodwilson: This should be implemented (and working!) in the latest push.

atomic_reactor/plugins/exit_koji_import.py
jarodwilson: Will make that message less ambiguous and more useful.

atomic_reactor/plugins/exit_koji_import.py
jarodwilson: A prior iteration of the code did loop through the log just trying to build a list of arches, and then wrote out the log lines to the files, but it ended up looping through the logs once to build the list of arches, then one time for each of the arches as well, and @twaugh actually suggested the above. I suppose one question that should be asked here is if we should be trying to dynamically build the list of arches, just assume a static list of possible arches, and not send any empty logs, so we only need to loop over the combined log once to split out each line and write them out to individual files. Might be possible to do all in a single loop, including creating a dynamic arch list, by someone a bit more adept at python than I am... :)

atomic_reactor/plugins/exit_koji_import.py
jarodwilson: Ah. It could well be empty, you'd know better than I would. Some of this is just copy-n-paste as a base from other logging code, it's entirely possible I've brought over bits that aren't needed. I *think* the $arch.log files are supposed to be the per-arch build.log files, aren't they? I could be woefully mistaken though.

atomic_reactor/util.py
jarodwilson: Whoops, that's actually what @twaugh had suggested as well, I'd moved it from being in two plugins to util instead of koji_util, will relocate that now.

atomic_reactor/plugins/exit_koji_import.py
twaugh: I think it can be done in a single loop. How about `.setdefault(entry.platform, NamedTemporaryFile(...))`...?

atomic_reactor/koji_util.py
```diff
@@ -18,6 +18,28 @@
 
 
 logger = logging.getLogger(__name__)
+Output = namedtuple('Output', ['file', 'metadata'])
```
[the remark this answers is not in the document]
jarodwilson: This was fixed, I dunno why stickler hasn't hidden this, other than probably because the fix was a few lines north, adding the import.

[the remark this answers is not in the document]
lcarva: Yeah, stickler doesn't remove its previous comments. At the bottom of this PR it shows that stickler is happy now.

atomic_reactor/plugins/exit_koji_import.py
jarodwilson: Okay, per discussion on IRC, I think I've whipped up something that seems to work with a single pass through the logs, and without loading the whole log into memory before writing it out to temp files.

atomic_reactor/koji_util.py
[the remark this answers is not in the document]
jarodwilson: No, I know it doesn't remove them, but it usually hides them under a "Show outdated" collapsible thingamajig if there are code changes in the area it was complaining about. Whatever the case, yeah, this is fixed, ignoring it now. :)

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -149,6 +156,52 @@ def get_buildroot(self, worker_metadatas):
...
+        for entry in logs:
+            platform = entry.platform
+            platform_logs.setdefault(platform, [])
+            if platform not in platforms:
+                platforms.append(platform)
+                platform_name[platform] = platform
+                if platform is None:
+                    platform_name[platform] = "orchestrator"
+                platform_logs[platform] = NamedTemporaryFile(prefix="%s-%s" %
+                                                             (self.build_id, platform_name[platform]),
```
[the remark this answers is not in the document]
jarodwilson: Look here, stickler. You can't have it both ways. You tell me my line is under-indented, so I indent it further, and now you complain that the line is too long. I'm kind of running out of options here, and anything further I do to try to silence you will do nothing but make the code harder to read.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -149,6 +156,53 @@ def get_buildroot(self, worker_metadatas):
...
+            return output
+
+        platforms = []
+        platform_name = {}
+        platform_logs = {}
+        for entry in logs:
+            platform = entry.platform
+            platform_logs.setdefault(platform, [])
+            if platform not in platforms:
+                platforms.append(platform)
```
twaugh: So `platform_logs` is a dict of "platform (str)" -> "file-like obj". Do you also need to keep a list of platforms? I'd expect to see something like:
```
for entry in logs:
    platform = entry.platform
    if platform not in platform_logs:
        filename = 'orchestrator' if platform is None else platform
        prefix = "%s-%s" % (self.build_id, filename)
        platform_logs[platform] = NamedTemporaryFile(prefix=prefix, ...)
    platform_logs[platform].write(entry.line)

for platform, logfile in platforms.items():
    logfile.flush()
    ...
```

jarodwilson: Seems to be working now, with 'if platform not in platform_logs.items():'.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -149,6 +156,47 @@ def get_buildroot(self, worker_metadatas):
...
+
+        platform_logs = {}
+        for entry in logs:
+            platform = entry.platform
+            platform_logs.setdefault(platform, [])
+            if platform not in platform_logs.items():
+                filename = 'orchestrator' if platform is None else platform
+                platform_logs[platform] = NamedTemporaryFile(prefix="%s-%s" %
+                                                             (self.build_id, filename),
+                                                             suffix=".log", mode='wb')
```
twaugh: I think this wants to be `mode='wt'`. The interface for `get_orchestrator_build_logs()` tells us that `.line` will be a 'str' ('unicode' in Python 2), not a 'bytes', so in order to write it to a file in binary mode we'd need to encode it (to e.g. UTF-8). Using `mode='wt'` means this happens automatically for us.

tests/plugins/test_koji_import.py
```diff
@@ -239,12 +241,9 @@ def mock_environment(tmpdir, session=None, name=None,
     flexmock(subprocess, Popen=fake_Popen)
     flexmock(koji, ClientSession=lambda hub, opts: session)
     flexmock(GitSource)
-    if logs_return_bytes:
-        logs = b'build logs - \xe2\x80\x98 \xe2\x80\x97 \xe2\x80\x99'
-    else:
-        logs = 'build logs - \u2018 \u2017 \u2019'
+    logs = [LogEntry('x86_64', 'Hurray for bacon')]
```
twaugh: Maybe include a Unicode non-ASCII  code point like \u2017 to catch encoding issues.

twaugh: I'm starting to think that automatic UTF-8 encoding for files only works properly in Python 3. I can't seem to get it to work right in Python2.

So I think we should use an explicitly Unicode log message like `u'Hurray for bacon \u2017'` and fix up the `.write()` call so that it encodes the log line to 'utf-8' explicitly.

twaugh:
Looks great! Needs some squashing, and still has '[wip]' in the PR title.

jarodwilson:
All squashed and fixed up to remove the [wip] prefixes.

atomic_reactor/plugins/exit_koji_import.py
lcarva: What's the resolution here?

twaugh: It had to be 'wb', with explicit `.encode()` calls before writing, to make Python 2 happy -- do you know another way?

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -149,6 +156,47 @@ def get_buildroot(self, worker_metadatas):
...
+        except TypeError:
+            # Older osbs-client has no get_orchestrator_build_logs
+            self.log.error("OSBS client does not support get_orchestrator_build_logs")
+            return output
+
+        platform_logs = {}
+        for entry in logs:
+            platform = entry.platform
+            platform_logs.setdefault(platform, [])
+            if platform not in platform_logs.items():
```
lcarva: When will `platform` ever be in `platform_logs.items()`?

Should this be something like this:

```python
platform = entry.platform
# platform_logs.setdefault(platform, [])  # Don't use setdefault if we need to check if key exists
if platform not in platform_logs:  # This actually checks if platform is in keys of platform_logs, not items
  platorm_logs[plaform] = NamedTemporaryFile(...)
  
```

atomic_reactor/plugins/exit_koji_import.py
lcarva: No, I do not. Currently resolution seems good to me.

atomic_reactor/plugins/exit_koji_import.py
twaugh: Good catch.

jarodwilson: Ah, I forgot about something here when I ran into it last night. With just 'if platform not in platform_logs:', the log file was never getting created, for whatever reason -- the self-tests would bomb out on the following .write() call, saying that you can't use .write on a dict, iirc, so something wasn't quite working, but it seems .items() wasn't the right fix either. Prior code using a separate list of platforms did seem to work correctly, though I guess better understanding why the code as suggested by both @twaugh and @lcarva isn't behaving as expected.

twaugh: Original discussion: https://github.com/projectatomic/atomic-reactor/pull/782#discussion_r135921167

jarodwilson: It would help if I'd read @lcarva 's comment closer, it shows that the check would never be true, since the setdefault call right before it made sure that was never true.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -149,6 +156,48 @@ def get_buildroot(self, worker_metadatas):
...
+        platform_logs = {}
+        for entry in logs:
+            platform = entry.platform
+            if platform not in platform_logs:
+                filename = 'orchestrator' if platform is None else platform
+                platform_logs[platform] = NamedTemporaryFile(prefix="%s-%s" %
+                                                             (self.build_id, filename),
+                                                             suffix=".log", mode='wb')
+            else:
+                platform_logs.setdefault(platform, [])
```
twaugh: This looks wrong. This is saying that if `platform` is not a key in the `platform_logs` dict, it should be added with the value being an empty list.

But we already know it is a key in the dict, or we wouldn't get here.

Also why `[]`? The values are meant to be file-like objects, not lists.

jarodwilson: Bah. I'm not sure why I thought that snippet needed to stay in some form. Will just remove it.

lcarva:
I'm tempted to think this doesn't need release notes update. Let me know otherwise.

PR #783, opened 2017-08-23
Allow pulp_pull to act as an exit plugin

twaugh:
This is required for arrangement version 4.

This is required for arrangement version 4, in which pulp_pull runs as an exit plugin. It must run in that phase because it has to run after pulp_publish, also an exit plugin -- pulp_publish has work to do when a build is failing, i.e. remove v1 content.
    
For arrangement versions earlier than 4, pulp_pull must still be able to run as a post-build plugin.

### DIFF
--- a/atomic_reactor/plugins/post_pulp_pull.py
+++ b/atomic_reactor/plugins/post_pulp_pull.py
@@ -16,7 +16,7 @@ discover the image ID Docker will give it.
 from __future__ import unicode_literals
 
 from atomic_reactor.constants import PLUGIN_PULP_PUSH_KEY, PLUGIN_PULP_SYNC_KEY
-from atomic_reactor.plugin import PostBuildPlugin
+from atomic_reactor.plugin import PostBuildPlugin, ExitPlugin
 from atomic_reactor.plugins.exit_remove_built_image import defer_removal
 from atomic_reactor.util import get_manifest_digests
 from docker.errors import NotFound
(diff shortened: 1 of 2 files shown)


vrutkovs:
The reason to make it an exit plugin is that it has to run after `pulp_publish`, which is exit plugin

twaugh:
Updated commit message.

atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -28,7 +28,11 @@ class CraneTimeoutError(Exception):
     pass
 
 
-class PulpPullPlugin(PostBuildPlugin):
+# Note: We use multiple inheritance here only to make it explicit that
```
lcarva: This comment contradicts the commit message in the sense that it states it's only done for clarity. 
The commit message makes it seem that this is actually required for functionality.

twaugh:
Inheriting from `ExitPlugin` allows it to run as an exit plugin. Additionally inheriting from `PostBuildPlugin` is only done for clarity.

atomic_reactor/plugins/post_pulp_pull.py
twaugh: Inheriting from `ExitPlugin` allows it to run as an exit plugin. Additionally inheriting from `PostBuildPlugin` is only done for clarity.

lcarva: Ah right, misread the diff. Thought it was already inheriting from `ExitPlugin`.

twaugh:
Anyone have objections to merging?

lcarva:
Release notes updated.

PR #784, opened 2017-08-23
Fix pulp pull

twaugh:
It was needlessly retrying its requests for media types. We only need to retry if we see a 404 response.

### DIFF
--- a/atomic_reactor/plugins/post_pulp_pull.py
+++ b/atomic_reactor/plugins/post_pulp_pull.py
@@ -77,6 +77,11 @@ class PulpPullPlugin(ExitPlugin, PostBuildPlugin):
             sleep(self.retry_delay)
 
     def run(self):
+        # Only run if the build was successful
+        if self.workflow.build_process_failed:
+            self.log.info("Not running for failed build")
+            return None, []
+
         # Work out the name of the image to pull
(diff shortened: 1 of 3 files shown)


twaugh:
Hmm, needs work to fix coverage.

rcerven:
why do we again wanna put retry logic into plugin? and only for 404?

twaugh:
404 is not something that should be automatically retried in general. However, pulp_pull knows that there may be a good reason to retry when it sees 404 (see comment).

So the expectation here is that pulp_pull will handle retries for 404, only for the specific case of Crane not having spotted the new Pulp content yet.

For everything else: the GET requests used by get_manifest_digests(), the docker API calls, retries for retryable errors should be handled automatically "behind the scenes".

Does that explain it adequately?

rcerven:
ok in that case at least move the retry function into utils so it can be used for anything else if needed

twaugh:
I could move it there, but what else would use it?

rcerven:
don't know, but seems weird put put again retry logic in plugin, when mine story is trying to get rid of it

twaugh:
How about we leave this retry logic as-is for this PR, but if your retry-logic PR allows re-use of this type of code, let's try to merge them there?

PR #785, opened 2017-08-24
Add 'publish' parameter to pulp_sync plugin

twaugh:
When set to False this parameter prevents the plugin from publishing the repository.


twaugh:
@mlangsdorf

PR #786, opened 2017-08-24
koji_import tries to read media-types from pulp_pull in exit plugins first, then worker annotations

vrutkovs:


atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -143,11 +143,9 @@ def set_help(self, extra, worker_metadatas):
                 extra['image']['help'] = None
 
     def set_media_types(self, extra, worker_metadatas):
```
lcarva: `worker_metadata` parameter is no loner used.

mlangsdorf: shouldn't the logic be 
`if pulp_pull_results and pulp_pull_results[1]:
      extra['image']['media_types'] = json.loads(pulp_pull_results[1])
      return`

and then the original logic otherwise? osbs-4285 says "otherwise, look for a media-types annotation on the worker build"

lcarva: that's a good point, not falling back to worker_metadata would break arrangement version 3, I think

vrutkovs: Right, reworked this - first `pulp_sync` is checked and then `worker_annotations`

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -143,11 +143,15 @@ def set_help(self, extra, worker_metadatas):
...
-        for platform in worker_metadatas:
-            annotations = get_worker_build_info(self.workflow, platform).build.get_annotations()
-            if annotations.get('media-types'):
-                extra['image']['media_types'] = json.loads(annotations['media-types'])
-                return
+        pulp_pull_results = self.workflow.exit_results.get(PLUGIN_PULP_PULL_KEY)
+        if pulp_pull_results and pulp_pull_results[1]:
+            extra['image']['media_types'] = json.loads(pulp_pull_results[1])
```
twaugh: Just `extra['image']['media_types'] = pulp_pull_results[1]`. It's already a list.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -143,11 +143,15 @@ def set_help(self, extra, worker_metadatas):
                 extra['image']['help'] = None
 
     def set_media_types(self, extra, worker_metadatas):
-        for platform in worker_metadatas:
-            annotations = get_worker_build_info(self.workflow, platform).build.get_annotations()
-            if annotations.get('media-types'):
-                extra['image']['media_types'] = json.loads(annotations['media-types'])
-                return
+        pulp_pull_results = self.workflow.exit_results.get(PLUGIN_PULP_PULL_KEY)
+        if pulp_pull_results and pulp_pull_results[1]:
```
twaugh: If `pulp_pull` ran *at all* it means we're running arrangement version >= 4, where we cannot be certain there is only a single worker build. So I think this should be `if pulp_pull_results:`, and if no media types were reported then we set an empty list.

tests/plugins/test_koji_import.py
```diff
@@ -1303,16 +1304,21 @@ def test_koji_import_add_help(self, tmpdir, os_env, expect_result):
...
-        build_info = BuildInfo(media_types=version)
-        orchestrate_plugin = workflow.plugin_workspace[OrchestrateBuildPlugin.key]
-        orchestrate_plugin[WORKSPACE_KEY_BUILD_INFO]['x86_64'] = build_info
+        if has_pulp_sync:
+            if version:
+                workflow.exit_results[PLUGIN_PULP_PULL_KEY] = ("acbdef1234", json.dumps(version))
+        else:
+            build_info = BuildInfo(media_types=version)
+            orchestrate_plugin = workflow.plugin_workspace[OrchestrateBuildPlugin.key]
+            orchestrate_plugin[WORKSPACE_KEY_BUILD_INFO]['x86_64'] = build_info
```
twaugh: No need for this entire block to be an `else`, but the media_types could be set to some canary value like `['look', 'elsewhere']` if `has_pulp_sync`.

atomic_reactor/plugins/exit_koji_import.py
vrutkovs: `pulp_pull` would return a tuple of `build_id, media_types`, this is why `[1]` is used here

atomic_reactor/plugins/exit_koji_import.py
vrutkovs: Fixed

atomic_reactor/plugins/exit_koji_import.py
twaugh: But you're testing whether `pulp_pull_results[1]` is truthy, and if not then you're checking worker builds. We shouldn't check worker builds in that case.

vrutkovs: Oh, agree, its safe to assume that if pulp_pull has returned a result, then checking worker builds is pointless. Updated

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -166,11 +166,15 @@ def set_help(self, extra, worker_metadatas):
...
-                return
+        pulp_pull_results = self.workflow.exit_results.get(PLUGIN_PULP_PULL_KEY)
+        if pulp_pull_results:
+            extra['image']['media_types'] = pulp_pull_results[1]
+        else:
+            for platform in worker_metadatas:
+                annotations = get_worker_build_info(self.workflow, platform).build.get_annotations()
+                if annotations.get('media-types'):
+                    extra['image']['media_types'] = json.loads(annotations['media-types'])
+                    return
```
lcarva: Nitpick: a `break` statement here would look more natural.

twaugh:
Please re-base against master to fix stickler-ci.

twaugh:
Fix for unreleased bug, so no release notes update needed.

PR #787, opened 2017-08-24
koji_import: re-write pull specifications for Crane as needed

twaugh:
When Pulp and Koji integration are both in use, Koji metadata created by worker builds has incorrect pullspecs in the extra.docker.repositories field for the docker-image output. This is because the worker build only knows about the 'sync' Docker registry (used for syncing images into Pulp).

In this situation, koji_import needs to re-write these pull specifications so they use Pulp's Crane instead of the Docker registry.

### DIFF
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -19,8 +19,11 @@ from atomic_reactor.plugins.build_orchestrate_build import (get_worker_build_inf
 from atomic_reactor.plugins.post_fetch_worker_metadata import FetchWorkerMetadataPlugin
 from atomic_reactor.plugins.pre_add_filesystem import AddFilesystemPlugin
 from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
-from atomic_reactor.constants import PLUGIN_KOJI_IMPORT_PLUGIN_KEY, PLUGIN_PULP_PULL_KEY
-from atomic_reactor.util import (get_build_json, get_preferred_label, df_parser)
+from atomic_reactor.constants import (PLUGIN_KOJI_IMPORT_PLUGIN_KEY,
+                                      PLUGIN_PULP_PULL_KEY,
+                                      PLUGIN_PULP_SYNC_KEY)
+from atomic_reactor.util import (get_build_json, get_preferred_label,
+                                 df_parser, ImageName)
 from atomic_reactor.koji_util import create_koji_session
 from osbs.conf import Configuration
 from osbs.api import OSBS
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -99,14 +102,34 @@ def get_output(self, worker_metadatas):
...
+
         for platform in worker_metadatas:
             for instance in worker_metadatas[platform]['output']:
                 instance['buildroot_id'] = '{}-{}'.format(platform, instance['buildroot_id'])
 
+                if instance['type'] != 'docker-image':
+                    continue
+
                 # update image ID with pulp_pull results
-                if 'extra' in instance and platform == "x86_64" and has_pulp_pull:
```
vrutkovs: `'extra' in instance` check is required - as logs are being listed here

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -99,14 +102,34 @@ def get_output(self, worker_metadatas):
...
+                if platform == "x86_64" and has_pulp_pull:
                     image_id, _ = self.workflow.postbuild_results[PLUGIN_PULP_PULL_KEY]
                     instance['extra']['docker']['id'] = image_id
+
+                # update repositories to point to Crane
+                if crane_registry:
+                    pulp_pullspecs = []
+                    for pullspec in instance['extra']['docker']['repositories']:
```
vrutkovs: Same here: `instance` without `'extra'` key should be filtered out

atomic_reactor/plugins/exit_koji_import.py
twaugh: But see a few lines above where I filter on 'type'.

atomic_reactor/plugins/exit_koji_import.py
twaugh: See a few lines above where I filter on 'type'.

atomic_reactor/plugins/exit_koji_import.py
vrutkovs: Oh, its a bit subtle, but yes, would work

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -473,7 +473,7 @@ def add_log_type(output, arch):
 
         digests = self.get_digests()
         repositories = self.get_repositories(digests)
-        tags = set(image.tag for image in self.workflow.tag_conf.primary_images)
+        tags = set(image.tag for image in self.workflow.tag_conf.images)
```
lcarva: Why is this needed now?

twaugh: There are literally no primary tags for worker builds any more, so the metadata included "tags: []". Hmm, this should go in a separate commit I guess.

twaugh: Separated into its own commit.

lcarva:
Bugfix to unreleased code - not updating release notes.

PR #788, opened 2017-08-24
Add 'sha256:' prefix to image ID if needed

twaugh:
Older versions of Docker need this.


twaugh:
No release note needed.

PR #791, opened 2017-08-25
koji_import: don't overwrite 'id' with None

twaugh:
When pulp_pull sees a v2 digest from Crane it returns None. In this case, don't overwrite the reported image ID.


twaugh:
Fix for unreleased bug so no release note needed.

PR #792, opened 2017-08-25
squash: fix up image ID to include sha256: prefix as needed

twaugh:


lcarva:
Release note not needed.

PR #793, opened 2017-08-25
exit_store_metadata: use repos from pulp_pull in exit plugins if available

vrutkovs:
Not sure if tests were updated correctly, but actual code was tested with `next` branch and works

### DIFF
--- a/atomic_reactor/constants.py
+++ b/atomic_reactor/constants.py
@@ -73,6 +73,7 @@ PLUGIN_PULP_TAG_KEY = 'pulp_tag'
 PLUGIN_ADD_FILESYSTEM_KEY = 'add_filesystem'
 PLUGIN_FETCH_WORKER_METADATA_KEY = 'fetch_worker_metadata'
 PLUGIN_GROUP_MANIFESTS_KEY = 'group_manifests'
+PLUGIN_BUILD_ORCHESTRATE_KEY = 'orchestrate_build'
 
 # max retries for docker requests
 DOCKER_MAX_RETRIES = 3
--- a/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
+++ b/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
@@ -21,7 +21,8 @@ from atomic_reactor.constants import (PLUGIN_KOJI_IMPORT_PLUGIN_KEY,
                                       PLUGIN_KOJI_PROMOTE_PLUGIN_KEY,
                                       PLUGIN_KOJI_UPLOAD_PLUGIN_KEY,
                                       PLUGIN_PULP_PUSH_KEY,
-                                      PLUGIN_ADD_FILESYSTEM_KEY)
+                                      PLUGIN_ADD_FILESYSTEM_KEY,
+                                      PLUGIN_BUILD_ORCHESTRATE_KEY)
 from atomic_reactor.plugin import ExitPlugin
 from atomic_reactor.util import get_build_json
 
@@ -86,8 +87,17 @@ class StoreMetadataInOSv3Plugin(ExitPlugin):
     def _get_registries(self):
         """
         Return a list of registries that this build updated
+
+        For orchestrator it should attempt to filter out non-pulp registries, on worker - return
+        all registries
         """
-        return self.workflow.push_conf.all_registries
+        if self.workflow.buildstep_result.get(PLUGIN_BUILD_ORCHESTRATE_KEY):
+            registries = self.workflow.push_conf.pulp_registries
+            if not registries:
+                registries = self.workflow.push_conf.all_registries
+            return registries
+        else:
+            return self.workflow.push_conf.all_registries
 
     def get_repositories(self):
         # usually repositories formed from NVR labels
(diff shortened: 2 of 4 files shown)


atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -87,7 +87,10 @@ def _get_registries(self):
         """
         Return a list of registries that this build updated
         """
-        return self.workflow.push_conf.all_registries
+        registries = self.workflow.push_conf.pulp_registries
+        if not registries:
+            registries = self.workflow.push_conf.all_registries
+        return registries
```
twaugh: This is wrong in general. We needed to change this to return all registries for the benefit of delete_from_registry, see https://github.com/projectatomic/atomic-reactor/commit/e73808a5.

twaugh:
Let me look at the docs for how this was supposed to work.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -219,7 +222,7 @@ def run(self):
             top_layer, _ = pulp_push_results
             annotations['v1-image-id'] = top_layer
 
-        pulp_pull_results = self.workflow.postbuild_results.get(PulpPullPlugin.key)
+        pulp_pull_results = self.workflow.exit_results.get(PulpPullPlugin.key)
```
twaugh: No, this is specifically for arrangement versions < 4 which ran pulp_pull as a post-build plugin in the worker build. It could definitely use a comment though.

twaugh:
The way this is meant to work is that tag_from_config updates the tag_conf and group_manifests updates the push_conf. Then, pulp_sync should update push_conf with information about Pulp. When merged together we should get pull specifications for each tag the manifest list is accessible through, although it will also include the v2 docker registry.

Because delete_from_registry needs to know details of the image manifests in the docker registry, we'll need to have exit_store_metadata_in_osv3 act differently depending on whether it is running in the orchestrator or the worker build.

Orchestrator build: use pulp registries if there are any, otherwise use all registries
Worker build: use all registries

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
vrutkovs: Oh, that makes sense. In this case a PR to delete manifest lists from registry would resolve that.

Renaming this to `[WIP]` until this is implemented

vrutkovs:
>we'll need to have exit_store_metadata_in_osv3 act differently depending on whether it is running in the orchestrator or the worker build

Should it be an additional param to `exit_store_metadata_in_osv3` plugin, or just checking for `pulp_pull` results in exit plugins is sufficient? In the latter case it should be enough to add a comment

twaugh:
I don't think checking for pulp_pull is the best way. I'd suggest either checking for orchestrate_build (make it's plugin key a constant), or being told by parameter. @lcarva?

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
lcarva: +1 for adding a comment

lcarva:
> I don't think checking for pulp_pull is the best way. I'd suggest either checking for orchestrate_build (make it's plugin key a constant), or being told by parameter.

I think checking for `orchestrate_build` seems reasonable. Maybe a helper utility function in utils.py is worth adding.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
vrutkovs: `pulp_pull_results` would now check both postbuild and exit plugins lists. Added a comment

vrutkovs:
Updated `_get_repositories()` to check if its in orchestrator build flow or not and added a test for that

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -223,7 +233,9 @@ def run(self):
         if pulp_push_results:
             media_types += ['application/json']
 
-        pulp_pull_results = self.workflow.postbuild_results.get(PulpPullPlugin.key)
+        # pulp_pull may run on worker as a postbuild plugin or on orchestrator as an exit plugin
+        pulp_pull_results = self.workflow.postbuild_results.get(PulpPullPlugin.key) or \
+            self.workflow.exit_results.get(PulpPullPlugin.key)
```
twaugh: Nit: nicer to use parentheses and line these up, instead of backslash

vrutkovs: Good idea, implemented

vrutkovs:
Now Travis fails on `test_orchestrate_build`, but I can't reproduce it locally :/

lcarva:
I have absolutely no idea why tests are failing in travis...

vrutkovs:
Tests fixed, don't import `OrchestrateBuildPlugin` - this seems to break the tests :(

atomic_reactor/constants.py
```diff
@@ -73,3 +73,4 @@
 PLUGIN_ADD_FILESYSTEM_KEY = 'add_filesystem'
 PLUGIN_FETCH_WORKER_METADATA_KEY = 'fetch_worker_metadata'
 PLUGIN_GROUP_MANIFESTS_KEY = 'group_manifests'
+PLUGIN_BUILD_ORCHESTRATE_KEY = 'orchestrate_build'
```
lcarva: orchestrate build plugin should also use this.

vrutkovs: Fixed

twaugh:
Needs re-basing.

vrutkovs:
Rebased on master

twaugh:
Fix for unreleased bug, no need for release notes update.

PR #795, opened 2017-08-28
Store media-type from worker's  pulp_push in annotations

vrutkovs:


atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -222,7 +222,9 @@ def run(self):
         pulp_pull_results = self.workflow.postbuild_results.get(PulpPullPlugin.key)
         if pulp_pull_results:
             _, media_types = pulp_pull_results
-            annotations['media-types'] = json.dumps(media_types)
+            if pulp_push_results:
+                media_types = list(set(media_types + ['application/json']))
```
lcarva: I don't really like setting the v1 content media type here, but I can't really think of an alternative :(

lcarva: Wait, is this code path only for arrangement 3? It requires both `pulp_pull` and `pulp_push` to run on the same build. In arrangement 4, this is no longer true. If so, shouldn't pulp_pull already include `'application/json'` media type?

vrutkovs: This one is for both:
* in arrangement 3 the build annotations would be read by `koji_upload` from this build
* in arrangement 4 worker build's `pulp_push` would write the annotation to the worker build. 
In `koji_import` this annotation from x86-64 would be combined with `pulp_pull` result

twaugh: Yes, this looks like a fine approach.

lcarva: > in arrangement 4 worker build's pulp_push would write the annotation to the worker build.

I must be missing something, but the conditional for pulp_push_results is nested under pulp_pull_results. This is a no-op in arrangement 4 since pulp_pull and pulp_push do not run on the same build.

tests/plugins/test_koji_import.py
```diff
@@ -1295,43 +1295,39 @@ def test_koji_import_add_help(self, tmpdir, os_env, expect_result):
...
+        (["application/json",
+          "application/vnd.docker.distribution.manifest.v1+json",
+          "application/vnd.docker.distribution.manifest.v2+json"],
+         ["application/vnd.docker.distribution.manifest.v1+json",
+          "application/vnd.docker.distribution.manifest.v2+json",
+          "application/vnd.docker.distribution.manifest.list.v2+json"],
+         ["application/json",
+          "application/vnd.docker.distribution.manifest.list.v2+json",
+          "application/vnd.docker.distribution.manifest.v1+json",
+          "application/vnd.docker.distribution.manifest.v2+json"]),
```
twaugh: This needs to be easier to read so we can clearly see what we're testing/expecting.

What about this sort of layout?:
```
@pytest.mark.parametrize(('config', 'expected'), [
    # API V1-only build, arrangement version < 4
    ({'pulp_push': True,
      'pulp_pull_in_worker': True,
      'pulp_pull_in_orchestrator': False,
      'pulp_pull_media_types': []},
      ["application/json"]),

    ...
    # API V1+V2 build, arrangement version >= 4, schema2-capable Pulp
    ({'pulp_push': True,
       'pulp_pull_in_worker': False,
       'pulp_pull_in_orchestrator': True,
       'pulp_pull_media_types': [
           "application/vnd.docker.distribution.manifest.v1+json",
           "application/vnd.docker.distribution.manifest.v2+json",
           "application/vnd.docker.distribution.manifest.v2.list+json",
       ]},
      ["application/json",
       "application/vnd.docker.distribution.manifest.v1+json",
       "application/vnd.docker.distribution.manifest.v2+json",
       "application/vnd.docker.distribution.manifest.v2.list+json"]),
...
```

vrutkovs: Agree, its a bit excessive. Is it worth setting `pulp_pull: {'schema1': True, 'schema2': True, 'list': False}`?

twaugh: Yes, that sounds even better.

vrutkovs: Implemented

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
vrutkovs: Agree, implemented independent setting of media types. The tests cannot properly verify that (as those have a single instance of plugins list) though

tests/plugins/test_koji_import.py
```diff
@@ -1300,43 +1300,105 @@ def test_koji_import_add_help(self, tmpdir, os_env, expect_result):
...
+        ({'pulp_push': True,
+          'pulp_pull_in_worker': False,
+          'pulp_pull_in_orchestrator': True,
+          'schema1': True,
+          'schema2': True,
+          'list': True},
+         ["application/json",
+          "application/vnd.docker.distribution.manifest.v1+json",
+          "application/vnd.docker.distribution.manifest.v2+json",
+          "application/vnd.docker.distribution.manifest.list.v2+json"]),
```
twaugh: Duplicated test case?

tests/plugins/test_koji_import.py
```diff
@@ -1300,43 +1300,105 @@ def test_koji_import_add_help(self, tmpdir, os_env, expect_result):
...
+        ({'pulp_push': False,
+          'pulp_pull_in_worker': False,
+          'pulp_pull_in_orchestrator': True,
+          'schema1': True,
+          'schema2': True,
+          'list': True},
+         ["application/vnd.docker.distribution.manifest.v1+json",
+          "application/vnd.docker.distribution.manifest.v2+json",
+          "application/vnd.docker.distribution.manifest.list.v2+json"]),
```
twaugh: Can we also test pulp_push=False, pulp_pull_in_worker=True ?

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -219,10 +219,17 @@ def run(self):
             top_layer, _ = pulp_push_results
             annotations['v1-image-id'] = top_layer
 
+        media_types = []
```
twaugh: If you make it a set to start with here, you don't need to create a set from it later.

tests/plugins/test_koji_import.py
vrutkovs: Reworked to test "schema 2 + manifest list" case

tests/plugins/test_koji_import.py
vrutkovs: Done

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
vrutkovs: `set` cannot concatenate lists (coming from `pulp_pull_results`). I could turn them all into sets, but it feels a bit like cheating

twaugh: Fair enough.

twaugh:
Fix for unreleased bug so no release notes update required.

PR #797, opened 2017-08-28
Add new plugin compare_components

dzickusrh:
This plugin takes the metadata from the worker and compares the components
version with that of other arches metadata.  The version info should be
identical otherwise fail the plugin.

The only component type supported initially is 'rpm'.

Test case attached using a copy of live koji data.

### DIFF
--- a/atomic_reactor/plugins/post_compare_components.py
+++ b/atomic_reactor/plugins/post_compare_components.py
@@ -1,0 +1,112 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+from atomic_reactor.plugin import PostBuildPlugin
+from atomic_reactor.constants import (PLUGIN_COMPARE_COMPONENTS_KEY,
+                                      PLUGIN_FETCH_WORKER_METADATA_KEY)
+
+
+SUPPORTED_TYPES = ("rpm",)
+
+
+class CompareComponentsPlugin(PostBuildPlugin):
+    """
+    Compare components from each worker build and verify the same version was
+    on each worker.
+    """
+
+    key = PLUGIN_COMPARE_COMPONENTS_KEY
+    is_allowed_to_fail = False
+
+    def rpm_compare(self, a, b):
+        """
+        Compare rpm component version 'a' with 'b'.
+        'name' is implied equal as it is the key() lookup
+        """
+        if a['version'] == b['version'] and \
+           a['release'] == b['release'] and \
+           a['signature'] == b['signature']:
+            return
+
+        raise ValueError("%s != %s" % (a, b))
+
+    def get_component_list_from_workers(self, worker_metadatas):
+        """
+        Find the component lists from each worker build.
+
+        The components that are interesting are under the 'output' key.  The
+        buildhost's components are ignored.
+
+        Inside the 'output' key are various 'instances'.  The only 'instance'
+        with a 'component list' is the 'docker-image' instance.  The 'log'
+        instances are ignored for now.
+
+        Reference plugin post_koji_upload for details on how this is created.
+
+        :return: list of component lists
+        """
+        comp_list = []
+        for platform in sorted(worker_metadatas.keys()):
+            for instance in worker_metadatas[platform]['output']:
+                if instance['type'] == 'docker-image':
+                    if 'components' not in instance or not instance['components']:
+                        self.log.warn("Missing 'components' key in 'output' metadata instance: %s",
+                                      instance)
+                        continue
+
+                    comp_list.append(instance['components'])
+
+        return comp_list
+
+    def run(self):
+        """
+        Run the plugin.
+        """
+
+        worker_metadatas = self.workflow.postbuild_results.get(PLUGIN_FETCH_WORKER_METADATA_KEY)
+        comp_list = self.get_component_list_from_workers(worker_metadatas)
+
+        if not comp_list:
+            raise ValueError("No components to compare")
+
+        # master compare list
+        master_comp = {}
+
+        # The basic strategy is to start with empty lists and add new component
+        # versions as we find them.  Upon next iteration, we should notice
+        # duplicates and be able to compare them.  If the match fails, we raise
+        # an exception.  If the component name does not exist, assume it was an
+        # arch dependency, add it to list and continue.  By the time we get to
+        # the last arch, we should have every possible component in the master
+        # list to compare with.
+
+        # Keep everything separated by component type
+        failed = False
+        for components in comp_list:
+            for component in components:
+                t = component['type']
+                name = component['name']
+
+                if t not in SUPPORTED_TYPES:
+                    raise ValueError("Type %s not supported")
+
+                identifier = (t, name)
+                if identifier not in master_comp:
+                    master_comp[identifier] = component
+                    continue
+
+                try:
+                    mc = master_comp[identifier]
+
+                    if t == 'rpm':
+                        self.rpm_compare(mc, component)
+                except ValueError as ex:
+                    self.log.warn("Comparison mismatch for component %s: %s", name, ex)
+                    failed = True
+
+        if failed:
+            raise ValueError("Failed component comparison")
--- a/tests/plugins/test_compare_components.py
+++ b/tests/plugins/test_compare_components.py
@@ -1,0 +1,162 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import print_function, unicode_literals
+
+import os
+import json
+
+from flexmock import flexmock
+
+from atomic_reactor.constants import (PLUGIN_FETCH_WORKER_METADATA_KEY,
+                                      PLUGIN_COMPARE_COMPONENTS_KEY)
+from atomic_reactor.core import DockerTasker
+from atomic_reactor.inner import DockerBuildWorkflow
+from atomic_reactor.plugin import PostBuildPluginsRunner, PluginFailedException
+from atomic_reactor.util import ImageName
+
+from tests.constants import MOCK_SOURCE, TEST_IMAGE, INPUT_IMAGE, FILES
+from tests.docker_mock import mock_docker
+
+import pytest
+
+
+class MockSource(object):
+
+    def __init__(self, tmpdir):
+        tmpdir = str(tmpdir)
+        self.dockerfile_path = os.path.join(tmpdir, 'Dockerfile')
+        self.path = tmpdir
+
+    def get_dockerfile_path(self):
+        return self.dockerfile_path, self.path
+
+
+class MockInsideBuilder(object):
+
+    def __init__(self):
+        mock_docker()
+        self.tasker = DockerTasker()
+        self.base_image = ImageName(repo='fedora', tag='25')
+        self.image_id = 'image_id'
+        self.image = INPUT_IMAGE
+        self.df_path = 'df_path'
+        self.df_dir = 'df_dir'
+
+        def simplegen(x, y):
+            yield "some\u2018".encode('utf-8')
+        flexmock(self.tasker, build_image_from_path=simplegen)
+
+    def get_built_image_info(self):
+        return {'Id': 'some'}
+
+    def inspect_built_image(self):
+        return None
+
+    def ensure_not_built(self):
+        pass
+
+
+def mock_workflow(tmpdir):
+    workflow = DockerBuildWorkflow(MOCK_SOURCE, TEST_IMAGE)
+    setattr(workflow, 'builder', MockInsideBuilder())
+    setattr(workflow, 'source', MockSource(tmpdir))
+    setattr(workflow.builder, 'source', MockSource(tmpdir))
+    setattr(workflow, 'postbuild_result', {})
+    return workflow
+
+
+def mock_metadatas():
+    json_x_path = os.path.join(FILES, "example-koji-metadata-x86_64.json")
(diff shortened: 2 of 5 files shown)


tests/plugins/test_compare_components.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+@pytest.mark.parametrize('fail', [True, False])
+def test_fetch_worker_plugin(tmpdir, fail):
+    workflow = mock_workflow(tmpdir)
+
+    # data copied from a live brew build
+    worker_metadatas = {
+        'x86_64': {
+            "buildroots": [
```
lcarva: Can we store this data in its own file(s) somewhere under `tests/files` ?

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -0,0 +1,81 @@
...
+        return buildroots
+
+    def run(self):
+        """
+        Run the plugin.
+        """
+
+        worker_metadatas = self.workflow.postbuild_results.get(PLUGIN_FETCH_WORKER_METADATA_KEY)
+        buildroots = self.get_buildroot(worker_metadatas)
+        self.log.debug("DON: md is: %s and br are %s" % (worker_metadatas, buildroots))
```
lcarva: :)

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -0,0 +1,81 @@
...
+
+        # master compare list
+        master_comp = {}
+
+        # The basic strategy is to start with empty lists and add new component
```
lcarva: +1 for this algorithm

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def get_buildroot(self, worker_metadatas):
+        """
+        Find the buildroot entry of the metadata.
+
+        :return: list, containing dicts of partial metadata
+        """
+        buildroots = []
+        for platform in sorted(worker_metadatas.keys()):
+            for instance in worker_metadatas[platform]['buildroots']:
```
lcarva: We want to compare the output files, not buildroots.

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -0,0 +1,81 @@
...
+    is_allowed_to_fail = False
+
+    def rpm_compare(self, a, b):
+        """
+        Compare rpm component version 'a' with 'b'.
+        'name' is implied equal as it is the key() lookup
+        """
+        if a['version'] == b['version'] and \
+           a['release'] == b['release'] and \
+           a['signature'] == b['signature']:
```
lcarva: For completeness, let's compare the `'type'` key as well.

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -0,0 +1,90 @@
...
+        'name' is implied equal as it is the key() lookup
+        """
+        self.log.debug("%s: version %s == %s" % (a['name'], a['version'], b['version']))
+        if a['version'] == b['version'] and \
+           a['release'] == b['release'] and \
+           a['signature'] == b['signature']:
+            return
+
+        self.log.debug("Comparing %s to %s" % (a, b))
+        raise PluginFailedException("Version mismatch for package %s" % a['name'])
```
lcarva: Let's not raise `PluginFailedException`. That is used for wrapping the exception raised from plugins. How about `ValueError` ?

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -0,0 +1,90 @@
...
+            for component in components:
+                t = component['type']
+                name = component['name']
+
+                if name not in master_comp:
+                    master_comp[name] = component
+                    continue
+
+                if t == 'rpm':
+                    self.rpm_compare(master_comp[name], component)
```
lcarva: Do we want to collect all errors then raise exception? Might be a better user experience.

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -0,0 +1,105 @@
...
+    is_allowed_to_fail = False
+
+    def rpm_compare(self, a, b):
+        """
+        Compare rpm component version 'a' with 'b'.
+        'name' is implied equal as it is the key() lookup
+        """
+        if a['version'] == b['version'] and \
+           a['release'] == b['release'] and \
+           a['type'] == b['type'] and \
```
twaugh: We're already comparing 'type' in the main loop.

tests/plugins/test_compare_components.py
```diff
@@ -0,0 +1,783 @@
...
+      "buildroot_id": 1
+    },
+    {
+      "extra": {
+        "image": {
+          "arch": "x86_64"
+        },
+        "docker": {
+          "repositories": [
+            "brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888/rcm/buildroot:1.0-71"
```
twaugh: Can we use a *.example.com hostname here?

tests/plugins/test_compare_components.py
```diff
@@ -0,0 +1,783 @@
...
+      "buildroot_id": 1
+    },
+    {
+      "extra": {
+        "image": {
+          "arch": "ppc64le"
+        },
+        "docker": {
+          "repositories": [
+            "brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888/rcm/buildroot:1.0-71"
```
twaugh: Can we use a *.example.com hostname here?

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -0,0 +1,105 @@
...
+                try:
+                    mc = master_comp[name]
+
+                    if t != mc['type']:
+                        raise ValueError("Bad type %s != %s" % (t, mc['type']))
+
+                    if t == 'rpm':
+                        self.rpm_compare(mc, component)
+                    else:
+                        self.log.warn("Unknown component type %s, skipping comparison", t)
```
twaugh: Would be nice to get test coverage for this line.

atomic_reactor/plugins/post_compare_components.py
lcarva: It seems to belong here, and not in the main loop.

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -0,0 +1,105 @@
...
+                except ValueError as ex:
+                    self.log.warn("Version mismatch for package %s: %s", name, ex)
```
lcarva: Nitpick: but the version is not necessarily the cause of the mismatch.

tests/plugins/test_compare_components.py
```diff
@@ -0,0 +1,783 @@
...
+from atomic_reactor.util import ImageName
+
+from tests.constants import MOCK_SOURCE, TEST_IMAGE, INPUT_IMAGE
+from tests.docker_mock import mock_docker
+
+import pytest
+
+
+# data copied from a live brew build
+metadata_x86_64_json = """\
```
lcarva: Can we store this data in its own file(s) somewhere under `tests/files` ?

atomic_reactor/plugins/post_compare_components.py
dzickusrh: My counter argument is 
pkgA['type'] = 'rpm' and pkgB['type'] = 'foo'.  pkgA makes it to master_comp list (new pkg), but pkgB triggers 'unknown' type because it fails to match type == 'rpm'.

So the warning is misleading one to think pkgB is new and incorrectly labeled when one can conclude it is a 'version' mismatch. 

Now swap the types pkgA = foo and pkgB = rpm, you get the version mismatch error as expected because pkgA is blindly added to master_comp list.

This is why I put the 'type' check in the 'main' function.  I thought it would explicitly catch the bad scenarios as opposed to relying on a random order to determine what warning is generated.

Feel free to disagree and I can make the changes.

twaugh: I'm not so sure we should even be comparing components of different types, just like we don't compare components with different names.

I tihnk 'type' and 'name' should be used together as the key for comparison.

atomic_reactor/plugins/post_compare_components.py
twaugh: I think `master_comp` keys ought to be tuples of (type,  name).

lcarva: Ah! So it can be used like this?
```python
identifier = (t, name)
if identifier not in master_comp:
  master_comp[identifier] = component
  continue
```

Type wasn't compared before, but since we're comparing it now, I think tuples as keys make a lot of sense.

twaugh: Yes, exactly like that.

dzickusrh: Hmm, bad 'types' don't fail like they are supposed to.  And maybe we don't care in some scenarios.

idA = (foo, perl)
idB = (rpm, perl)

should this fail?  We can pre-filter certain 'type' before saving, ie

if t not in accepted_types:
  raise(Bad type)

But
idA = (log, perl)
idB = (rpm, perl)

would pass because 'log' would be acceptable type or replace 'log' with 'jar'??  Are we ok with that limitation?

Cheers,
Don

lcarva: The idea is that the unique identifier for a component is *both* the type and the name. In your examples above, `idA` and `idB` would never be compared to each other. However, both can still be stored in `master_comp` for later usage when another component with same type and name is found.

dzickusrh: Sure, I get that.  My point was (and I am probably going overboard with this scenario), if/could a component come in with a bad component type?  If so, that gets missed because it would be assumed to be a different component.  Now if we accept that atomic_reactor provides the component type correctly on the worker (and a corrupt json file doesn't happen in between), then we have nothing to worry about.

I just wanted to throw that out there.  I believe my latest code implements the scheme you are requesting with the addition of a sanity 'type' check right above it.

Cheers,
Don

twaugh: We need to trust that the data is accurate. What we're trying to look for is an RPM component that has unexpectedly been installed with a different n-v-r.

dzickusrh: I guess bad experience has taught to me careful with pulling data over the network.  In retrospect, I wish we implemented a simple checksum scheme for the metadata to verify it with fetch_worker_metadata.  That would make my argument moot.

Regardless of my opinion, I believe current code implements what you requested?  Or is there something else I need to change?

Cheers,
Don

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -0,0 +1,106 @@
...
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+from atomic_reactor.plugin import PostBuildPlugin
+from atomic_reactor.constants import (PLUGIN_COMPARE_COMPONENTS_KEY,
+                                      PLUGIN_FETCH_WORKER_METADATA_KEY)
+
+
+supported_types = ["rpm"]
```
twaugh: Usually a module-level constant variable would be named in UPPER_CASE_WITH_UNDERSCORES.
https://www.python.org/dev/peps/pep-0008/#constants

Also, here `["rpm"]` is mutable, whereas `("rpm",)` is immutable. Using the immutable style (i.e. a 1-tuple) might make it clearer that this is never updated at runtime.

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # Keep everything separated by component type
+        failed = False
+        for components in comp_list:
+            for component in components:
+                t = component['type']
+                name = component['name']
+
+                if t not in supported_types:
+                    raise ValueError("Type %s not supported")
```
lcarva: Do we want an exception here? Seems unnecessarily making the code brittle.

twaugh: I guess this is fine. It means we'll need to add support for non-RPM content to this plugin at the same time as adding metadata tracking to fetch_maven_artifacts. @lcarva, OK with that?

atomic_reactor/plugins/post_compare_components.py
twaugh: (This looks fine now we're using (type,name) as the identifier.)

atomic_reactor/plugins/post_compare_components.py
lcarva: I guess that 's fine.

dzickusrh: I am fine with either an exception or warning.  I have to go into the code to change it to immutable anyway..

twaugh: Let's make it an exception then. Adding non-RPM content will be a code change in any case, so if we decide otherwise when we do that we could always change this to a warning at that time.

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -0,0 +1,106 @@
...
+        raise ValueError("%s != %s" % (a, b))
+
+    def get_component_list(self, worker_metadatas):
+        """
+        Find the worker build's component list.
+        The strategy is to find the 'output' key in the metadata and then
+        search the array for a 'docker-image' type.  The 'log' types are not
+        interesting.
+
+        :return: array of components
```
twaugh: I think this docstring could do with tightening up.

This is returning a list of lists of "component" dicts. The method name "get_component_list" suggests returning a list of "components". Should it be "get_components_from_workers" instead?

The doc string says "Find the worker build's component list", making it seem like there is a single worker build. Should it say "Find the components from each worker build" instead?

The ":return:" is inaccurate. It's a list of component lists.

lcarva:
Updated release notes draft:
> New `compare_components` plugin is used to ensure different platforms have a common set of components installed

PR #798, opened 2017-08-29
pulp_pull: fetch media type for manifest list too

vrutkovs:
Not sure if the tests are correct, though, as currently its impossible to verify "list and v2 schema" case now

### DIFF
--- a/atomic_reactor/plugins/post_pulp_pull.py
+++ b/atomic_reactor/plugins/post_pulp_pull.py
@@ -107,15 +107,19 @@ class PulpPullPlugin(ExitPlugin, PostBuildPlugin):
                                               pullspec, registry.uri,
                                               self.insecure, self.secret,
                                               require_digest=False)
-            if digests and digests.v2:
-                self.log.info("V2 schema 2 digest found, returning %s",
-                              self.workflow.builder.image_id)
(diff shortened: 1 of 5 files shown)


atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -107,13 +107,17 @@ def run(self):
...
+                    media_types.append('application/vnd.docker.distribution.manifest.list.v2+json')
+                if digests.v2:
+                    self.log.info("V2 schema 2 digest found, returning %s",
+                                  self.workflow.builder.image_id)
+                    media_types.append('application/vnd.docker.distribution.manifest.v2+json')
+                    # No need to pull the image to work out the image ID as
+                    # we already know it.
+                    return self.workflow.builder.image_id, sorted(media_types)
             else:
                 self.log.info("V2 schema 2 digest is not available")
```
lcarva: This log line seems to no longer be valid.  At this point it'll only be longed if no digests at all were found, not just v2 schema 2.

It's probably worth just dropping it altogether?

vrutkovs: Updated

twaugh:
Fix for unreleased bug so not updating release notes.

PR #799, opened 2017-08-29
pulp_publish: handle there being no worker builds

twaugh:
When a build has failed it may be that no worker builds were created. In that case, attempting to find their annotations would result in a traceback. Check there are annotations first.


lcarva:
Not updating release notes. Bugfix of unreleased changes

PR #800, opened 2017-08-29
pulp_publish: only remove each docker_image unit once

twaugh:


lcarva:
Not updating release notes. Bugfix of unreleased feature.

PR #804, opened 2017-08-30
Remove error log for exit plugin in post-build

twaugh:
The pulp_pull plugin is an ExitPlugin, but is allowed to run in the post-build phase for arrangement versions < 4.


twaugh:
Updated.

lcarva:
Not updating release notes. Bugfix to unreleased content.

PR #805, opened 2017-08-30
group_manifests: tag existing image manifest using identical content

twaugh:
Parsing to JSON and writing back to a byte stream causes the digest to change.


lcarva:
Not updating release notes. Bugfix to unreleased content.

PR #806, opened 2017-08-30
pulp_pull: add a new param expect_v2schema2 to wait until schema 2 digest is found

vrutkovs:
Pulp seems to rush redirecting to schema 1 images before manifest list is fully published. `pulp_pull` should optionally wait until schema 2 image is available

### DIFF
--- a/atomic_reactor/plugins/post_pulp_pull.py
+++ b/atomic_reactor/plugins/post_pulp_pull.py
@@ -38,7 +38,8 @@ class PulpPullPlugin(ExitPlugin, PostBuildPlugin):
 
     def __init__(self, tasker, workflow,
                  timeout=600, retry_delay=30,
-                 insecure=False, secret=None):
+                 insecure=False, secret=None,
+                 expect_v2schema2=False):
         """
         constructor
 
@@ -48,6 +49,8 @@ class PulpPullPlugin(ExitPlugin, PostBuildPlugin):
         :param retry_delay: int, seconds between pull attempts
         :param insecure: bool, allow non-https pull if true
         :param secret: str, path to secret
+        :param expect_v2schema2: bool, require Pulp to return a schema 2 digest and
+                                       retry until it does
         """
         # call parent constructor
         super(PulpPullPlugin, self).__init__(tasker, workflow)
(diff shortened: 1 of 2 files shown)


tests/plugins/test_pulp_pull.py
```diff
@@ -299,13 +299,17 @@ def test_pull_push_vs_sync(self, push, sync):
...
+        (True, False),
+        (True, True),
+    ])
     @pytest.mark.parametrize('timeout,retry_delay,failures,expect_success', [
         (0.1, 0.06, 1, True),
         (0.1, 0.06, 1, True),
         (0.1, 0.06, 3, False),
     ])
-    def test_pull_retry(self, v2, timeout, retry_delay, failures,
+    def test_pull_retry(self, expect_v2schema2, v2, timeout, retry_delay, failures,
```
twaugh: This method tests how the plugin reacts to 404 responses. Would be great to have another one that tests how it reacts to v2schema1 responses when expect_v2schema2 is True.

vrutkovs: It does happen for any `v2 == True` - first a number of 404's returned, then a schema 1manifest in returned (`expectation.and_return(self.config_response_config_v1)`) and after that a schema 2 manifest is returned

twaugh: Not sufficient. If I undo the relevant post_pulp_pull.py changes, the tests still pass.

atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -48,26 +49,34 @@ def __init__(self, tasker, workflow,
         :param retry_delay: int, seconds between pull attempts
         :param insecure: bool, allow non-https pull if true
         :param secret: str, path to secret
+        :param expect_v2schema2: bool, retry until Pulp retries schema 2 digest
```
lcarva: I don't understand this description

atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -48,26 +49,34 @@ def __init__(self, tasker, workflow,
...
-                return func(*args, **kwargs)
+                digests = func(*args, **kwargs)
             except requests.exceptions.HTTPError as ex:
                 # Retry for 404 not-found because we assume Crane has
                 # not spotted the new Pulp content yet. For all other
                 # errors, give up.
                 if ex.response.status_code != requests.codes.not_found:
                     raise
+            else:
+                if not self.expect_v2schema2:
```
lcarva: Would it be easier to read as: `if not self.expect_v2schema2 or digests.v2:` ?

tests/plugins/test_pulp_pull.py
vrutkovs: Oh, indeed. Added a warning log message and a test which is expected to fail due to the timeout, as pulp never returns schema 2, but its expected

atomic_reactor/plugins/post_pulp_pull.py
vrutkovs: It would, but for debugging purposes I added a warning message to identify we've encountered this bug - and now a full `if` is required

atomic_reactor/plugins/post_pulp_pull.py
vrutkovs: How about `require Pulp to return a schema 2 digest and retry until it does`?

atomic_reactor/plugins/post_pulp_pull.py
lcarva: Maybe an `elif` then.

atomic_reactor/plugins/post_pulp_pull.py
vrutkovs: Fixed

atomic_reactor/plugins/post_pulp_pull.py
vrutkovs: Right, looks better, fixed

twaugh:
Draft release notes updated.

PR #808, opened 2017-08-31
Build logs fixes

twaugh:
When koji_import writes logs, make sure to include newlines.

When koji_upload uploads logs, it now optionally only uploads buildstep logs using a provided name. This allows multiple workers to upload buildstep logs.


twaugh:
Draft release notes updated.

PR #809, opened 2017-08-31
koji_upload takes platform parameter instead of buildstep_logs

twaugh:
This way it can also use the platform name in the image archive filename.

### DIFF
--- a/atomic_reactor/plugins/post_koji_upload.py
+++ b/atomic_reactor/plugins/post_koji_upload.py
@@ -80,7 +80,7 @@ class KojiUploadPlugin(PostBuildPlugin):
                  koji_ssl_certs_dir=None, koji_proxy_user=None,
                  koji_principal=None, koji_keytab=None,
                  blocksize=None, prefer_schema1_digest=True,
-                 buildstep_logs=None):
+                 platform='x86_64'):
         """
         constructor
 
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -466,7 +467,7 @@ def add_log_type(output, arch):
         digests = self.get_digests()
         repositories = self.get_repositories(digests)
         tags = set(image.tag for image in self.workflow.tag_conf.images)
-        metadata, output = self.get_image_output(arch)
```
lcarva: Should `arch` not be used up above either? (around line 448: `add_log_type(add_buildroot_id(metadata), arch)`

lcarva: Otherwise, there can be a disparity with the `extra.image.arch` value.

twaugh: `arch` values in Koji are the same as those reported by uname. Any `arch` in the metadata should be from a uname value.

I don't think there's any disparity is there? It's only the filenames that use the platform instead of the uname.

lcarva: In the build you created with the fake `amd64` platform, the archive is named with amd64, but its metadata has `extra.image.arch`. I'm not sure if that's ok.

twaugh: It's what I intended.

At the moment we're free to choose platform names, aside from a few references to "x86_64" for legacy things. But we could Have a new platform name x86_64_v2 that differs only in enable_v1, or decide to switch to GOARCH names.

However, "arch" in Koji means CPU architecture, ie what uname reports.

twaugh:
Draft release notes updated.

PR #810, opened 2017-09-01
Make HTTP status code lists immutable

twaugh:


lcarva:

PR #811, opened 2017-09-01
retries for group_manifests plugin

rcerven:

### DIFF
--- a/atomic_reactor/plugins/post_group_manifests.py
+++ b/atomic_reactor/plugins/post_group_manifests.py
@@ -21,7 +21,7 @@ from subprocess import check_output, CalledProcessError, STDOUT
 from six.moves.urllib.parse import urlparse
 
 from atomic_reactor.plugin import PostBuildPlugin, PluginFailedException
-from atomic_reactor.util import Dockercfg, get_manifest_digests
+from atomic_reactor.util import Dockercfg, get_manifest_digests, get_retrying_requests_session
 from atomic_reactor.constants import PLUGIN_GROUP_MANIFESTS_KEY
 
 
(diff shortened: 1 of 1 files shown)


vrutkovs:
Need retries for:
* `check_output` of manifest-tool runs
* replace retries code in `get_manifest_digests`
* no need to `response.raise_for_status()` in `get_worker_manifest` if it 404 - otherwise it would take 15 mins to realize that Pulp doesn't support schema 2

twaugh:
It might be easier to add `check_output` retries once we have a solution for `git clone` as it is a similar problem.

Not sure whta you mean by replacing retries code in `get_manifest_digests`.

Why isn't `response.raise_for_status()` needed for 404? This plugin does not interact with Pulp, only with the Docker distribution registry.

vrutkovs:
>Not sure whta you mean by replacing retries code in get_manifest_digests.

Right, this code was already updated

>Why isn't response.raise_for_status() needed for 404? This plugin does not interact with Pulp, only with the Docker distribution registry.

I was thinking registry (and pulp) would return 404 when a different schema is tried, but its actually 400, so nevermind

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -164,7 +165,8 @@ def get_worker_manifest(self, worker_data):
 
                 url = '{0}/v2/{1}/manifests/{2}'.format(registry, repo, digest)
                 self.log.debug("attempting get from %s", url)
-                response = requests.get(url, **kwargs)
+                response = session.get(url, **kwargs)
+                response.raise_for_status()
```
lcarva: @mlangsdorf , we noticed that for this particular `get` API call the status code is not checked.
We're adding it now. Can you confirm this was not a deliberate decision?

lcarva:

PR #812, opened 2017-09-01
Koji Parent Content Generator

bfontecc007:
Store parent Koji build ID in Koji Content Generator metadata for build

### DIFF
--- a/atomic_reactor/constants.py
+++ b/atomic_reactor/constants.py
@@ -74,6 +74,7 @@ PLUGIN_ADD_FILESYSTEM_KEY = 'add_filesystem'
 PLUGIN_FETCH_WORKER_METADATA_KEY = 'fetch_worker_metadata'
 PLUGIN_GROUP_MANIFESTS_KEY = 'group_manifests'
 PLUGIN_BUILD_ORCHESTRATE_KEY = 'orchestrate_build'
+PLUGIN_KOJI_PARENT_KEY = 'koji_parent'
 
 # max retries for docker requests
 DOCKER_MAX_RETRIES = 3
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -24,7 +24,8 @@ from atomic_reactor.constants import (PLUGIN_KOJI_IMPORT_PLUGIN_KEY,
                                       PLUGIN_PULP_PULL_KEY,
                                       PLUGIN_PULP_SYNC_KEY,
                                       PLUGIN_FETCH_WORKER_METADATA_KEY,
-                                      PLUGIN_GROUP_MANIFESTS_KEY)
+                                      PLUGIN_GROUP_MANIFESTS_KEY,
+                                      PLUGIN_KOJI_PARENT_KEY)
 from atomic_reactor.util import (get_build_json, get_preferred_label,
                                  df_parser, ImageName, get_checksums)
 from atomic_reactor.koji_util import (create_koji_session, Output, KojiUploadLogger)
@@ -141,6 +142,10 @@ class KojiImportPlugin(ExitPlugin):
                 outputs.append(instance)
 
         return outputs
+
+    def get_parent_image_koji_build_id(self):
+        res = self.workflow.prebuild_results.get(PLUGIN_KOJI_PARENT_KEY, {})
+        return res.get('parent-image-koji-build-id')
 
     def get_buildroot(self, worker_metadatas):
         """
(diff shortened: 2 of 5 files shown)


atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -142,6 +143,14 @@ def get_output(self, worker_metadatas):
 
         return outputs
 
+    def get_parent_image_koji_build_id(self, workflow):
```
twaugh: Unused argument 'workflow' (from landscape.io)

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -142,6 +143,14 @@ def get_output(self, worker_metadatas):
...
+        res = self.workflow.prebuild_results.get(PLUGIN_KOJI_PARENT_KEY)
+        if res is None:
+            return
+        if 'parent-image-koji-build-id' in res:
+            return res['parent-image-koji-build-id']
+        return
```
twaugh: Could reduce method returns like this:
```
res = self.workflow.prebuild_results.get(PLUGIN_KOJI_PARENT_KEY, {})
return res.get('parent-image-koji-build-id')
```

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -317,6 +326,16 @@ def get_build(self, metadata, worker_metadatas):
                 else:
                     extra['filesystem_koji_task_id'] = task_id
 
+        parent_id = self.get_parent_image_koji_build_id(self.workflow)
+        if parent_id is not None:
+            try:
+                parent_id = int(parent_id)
+            except ValueError:
+                self.log.exception("invalid koji parent id %r", parent_id)
```
twaugh: This block isn't covered by unit tests.

bfontecc007: added.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -142,6 +143,13 @@ def get_output(self, worker_metadatas):
 
         return outputs
 
+    def get_parent_image_koji_build_id(self):
+        res = self.workflow.prebuild_results.get(PLUGIN_KOJI_PARENT_KEY)
+        if res is None:
+            return
```
twaugh: Are the above three lines needed?

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -142,6 +143,13 @@ def get_output(self, worker_metadatas):
...
+        res = self.workflow.prebuild_results.get(PLUGIN_KOJI_PARENT_KEY, {})
```
lcarva: Why compute `res` again?

If you're passing a default param to `get`, `{}`, might as well just change this whole method to:
```python
def get_parent_image_koji_build_id(self):
  return self.workflow.prebuild_results.get(PLUGIN_KOJI_PARENT_KEY, {}).get('parent-image-koji-build-id')
```

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -317,6 +322,16 @@ def get_build(self, metadata, worker_metadatas):
...
 
+        parent_id = self.get_parent_image_koji_build_id()
+        if parent_id is not None:
+            try:
+                parent_id = int(parent_id)
+            except ValueError:
+                self.log.exception("invalid koji parent id %r", parent_id)
+            else:
+                extra.setdefault('image', {})
+                extra['image']['parent-image-koji-build-id'] = parent_id
```
twaugh: Other metadata keys are in "snake_case" not "kebab-case". I think we should use underscores in the name, not dashes.

atomic_reactor/constants.py
```diff
@@ -74,6 +74,7 @@
 PLUGIN_FETCH_WORKER_METADATA_KEY = 'fetch_worker_metadata'
 PLUGIN_GROUP_MANIFESTS_KEY = 'group_manifests'
 PLUGIN_BUILD_ORCHESTRATE_KEY = 'orchestrate_build'
+PLUGIN_KOJI_PARENT_KEY = 'koji-parent'
```
twaugh: Should be an underscore, not a dash. Also, if adding this please also change the koji_parent plugin so it uses it, otherwise it gets defined in two places.

atomic_reactor/plugins/exit_koji_import.py
twaugh: Still needs underscores, not dashes, here.

twaugh:
Still needs work for https://github.com/projectatomic/atomic-reactor/pull/812/files#r137030657.

lcarva:
Updated release notes draft:
> `koji_import` will add the ID of parent image's koji build to metadata.

PR #813, opened 2017-09-02
fix tests so they don't fail on non-x86_64 systems

maxamillion:


lcarva:
Unit test fixes, not updating release notes.

maxamillion:
Thanks! :)

PR #814, opened 2017-09-04
Omit unique tag from Koji metadata

twaugh:
The unique tag has never been included in the 'tags' list in the Koji metadata. In order to preserve compatibility and reduce surprises, omit the unique tag from the 'tags' field in both the output (for compatibility) and the image.index map (for consistency).


twaugh:
Needs release note:
* Koji integration: with group_manifests=false in osbs.conf, the Koji output metadata "extra.docker.tags" list included the unique tag in 1.6.26. This was an unintended change from previous releases and has now been reverted.

PR #816, opened 2017-09-06
Fix build annotations when Pulp integration disabled

twaugh:


twaugh:
Needs release note:
* In 1.6.26 orchestrator build annotations were only accurate when Pulp integration was enabled. They are now correct regardless of whether Pulp integration is enabled

PR #817, opened 2017-09-06
Avoid traceback when pulp_pull fails

twaugh:


twaugh:
Release note:
* Bugfix:
** failure of the pulp_pull plugin no longer causes the store_metadata_in_osv3 plugin to fail

PR #818, opened 2017-09-06
koji_upload: do not include unique tag in metadata

twaugh:


twaugh:
Oh, hold on. We were putting that in deliberately for arrangement version 4...

PR #819, opened 2017-09-07
Remove debugging now it is not needed

twaugh:
This extra logging was added to aid debugging an issue which is now understood.


twaugh:
No release note needed.

PR #820, opened 2017-09-07
docker_api: report syntax errors

twaugh:
When the Dockerfile is incorrectly formed, use the exception's "explanation" (the syntax error message) instead of the HTTP error (500 status code).

This includes an expected-fail commit adding the test-case, followed by the actual fix, showing that tests now pass.


lcarva:
Updated release notes draft:
> - `docker_api` build plugin will provide explanation when build api call fails.

PR #821, opened 2017-09-07
Fix Koji metadata when group_manifests=false is set

twaugh:
The by-tag pull specification in output[].extra.docker.repositories should use one of the tags from extra.docker.tags in that output.


atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -230,11 +230,20 @@ def set_media_types(self, extra, worker_metadatas):
             extra['image']['media_types'] = sorted(list(set(media_types)))
 
     def set_group_manifest_info(self, extra, worker_metadatas):
+        version_release = None
+        for image in self.workflow.tag_conf.primary_images:
+            if '-' in image.tag:  # {version}-{release} only, and only one instance
+                version_release = image.tag
+                break
+
+        assert version_release is not None
```
rcerven: wouldn't be cleaner some pluginexception raise?

PR #822, opened 2017-09-08
Set Koji pull-specification metadata according to availability in the registry

twaugh:
In order to include the correct image manifest digests in the Koji metadata (`output[].extra.docker.repositories`), we need to collect both the v2 schema 1 and v2 schema 2 image digests in the worker build, then after syncing to Pulp in the orchestrator build, remove any that do not appear in the Pulp repository.

To achieve this:
* koji_upload should be told to report multiple digests based on whether Pulp is in use
* pulp_sync should collect the content of the repository after sync
* koji_import should remove any digests not present after sync

Additionally, update koji_promote to mirror this behaviour.

### DIFF
--- a/atomic_reactor/plugins/post_koji_upload.py
+++ b/atomic_reactor/plugins/post_koji_upload.py
@@ -80,7 +80,7 @@ class KojiUploadPlugin(PostBuildPlugin):
                  koji_ssl_certs_dir=None, koji_proxy_user=None,
                  koji_principal=None, koji_keytab=None,
                  blocksize=None, prefer_schema1_digest=True,
-                 platform='x86_64'):
+                 platform='x86_64', report_multiple_digests=False):
         """
         constructor
 
@@ -100,6 +100,8 @@ class KojiUploadPlugin(PostBuildPlugin):
         :param prefer_schema1_digest: bool, when True, v2 schema 1 digest will
             be preferred as the built image digest
         :param platform: str, platform name for this build
+        :param report_multiple_digests: bool, whether to report both schema 1
+            and schema 2 digests; if truthy, prefer_schema1_digest is ignored
         """
         super(KojiUploadPlugin, self).__init__(tasker, workflow)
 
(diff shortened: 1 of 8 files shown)


vrutkovs:
It would be nice to merge two parameters into one - e.g. `report_digests=['v1', 'v2']` and deprecate `prefer_schema1_digest`, not sure if that can easily be done though

twaugh:
Hard to do while still maintaining compatibility. Once we drop arrangement version 3 we can remove a lot of code.

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -374,16 +377,20 @@ def get_image_output(self):
...
+        digests = {}  # repository -> digests
         for registry in self.workflow.push_conf.docker_registries:
             for image in self.workflow.tag_conf.images:
                 image_str = image.to_str()
                 if image_str in registry.digests:
-                    digest = self.select_digest(registry.digests[image_str])
-                    digests[image.to_str(registry=False)] = digest
+                    image_digests = registry.digests[image_str]
+                    if self.report_multiple_digests:
+                        digest_list = [image_digests.v1, image_digests.v2]
```
lcarva: Any chance either `.v1` or `.v2` is `None`?

atomic_reactor/plugins/post_pulp_sync.py
```diff
@@ -236,5 +245,13 @@ def run(self):
             for image_name in images:
                 self.log.info("image available at %s", image_name.to_str())
 
+        # Fetch the repository content so we can remove v2 schema 2
+        # manifests from Koji metadata if they are not present
+        # (i.e. if Pulp does not have v2 schema 2 support).
```
lcarva: After a sync, can we reliably use the presence of v2 schema 2 content to determine pulp support?
If so, may we can expose this information to `koji_upload` instead of using the plugin arg `report_multiple_digests` ?

twaugh: The sync doesn't happen until after `koji_upload` runs.

atomic_reactor/plugins/post_koji_upload.py
twaugh: Only if the registry doesn't support v2, which I guess is a possibility we need to deal with.

atomic_reactor/plugins/post_pulp_sync.py
lcarva: bummer..

twaugh:
Updated the description to make it clearer what this is for.

atomic_reactor/plugins/post_pulp_sync.py
twaugh: That's a fine idea for koji_promote though. Implemented.

twaugh:
Draft release notes updated.

PR #824, opened 2017-09-11
Fix by-tag pullspec metadata

twaugh:
Signed-off-by: Tim Waugh <twaugh@redhat.com>

@csomh

PR #827, opened 2017-09-12
Fix repo content processing in pulp_sync

lcarva:


twaugh:
Fix was for issue that had not yet made it into a release, so not updating release notes.

twaugh:
This is a fix to https://github.com/projectatomic/atomic-reactor/pull/824.

PR #828, opened 2017-09-12
WIP - Add retry logic to cluster selection

bfontecc007:
Try a (configurable) number of times to reach a cluster that can handle additional load.

### DIFF
--- a/tests/plugins/test_orchestrate_build.py
+++ b/tests/plugins/test_orchestrate_build.py
@@ -893,6 +893,8 @@ def test_orchestrate_build_failed_to_list_builds(tmpdir, fail_at):
                 'platforms': ['x86_64'],
                 'build_kwargs': make_worker_build_kwargs(),
                 'osbs_client_config': str(tmpdir),
+                'unreachable_cluster_retry_delay': .1,
+                'unreachable_cluster_retry_count': 2
             }
         }]
     )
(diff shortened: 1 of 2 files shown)


twaugh:
Firstly, we need to disable the regular retries (using `with cluster_info.osbs.retries_disabled():`, https://github.com/projectatomic/osbs-client/blob/master/osbs/api.py#L1143) for the list-builds and create-build operations. If we don't do that, it will take a long time to move onto the next cluster in the list. We'll have two levels of retrying for the same thing.

Also, we need to apply retry logic to more than just cluster selection. What if creating the build on the selected cluster fails? In that situation we need to try another cluster.

Imagine this situation: clusters A, B, and C are for the same platform, but cluster A is unreachable. What we want to happen then is: try listing builds for cluster A once only -- i.e. discount the cluster for this selection attempt when it fails, then try listing builds for cluster B, then cluster C. After this first pass of attempting to list builds, pick a cluster based on load (e.g. B), then attempt to create the build on it, once only. If creating the build fails, try creating a build on cluster C, once only.

If all those things failed, start again, trying to list builds for cluster A as well as clusters B and C again.

bfontecc007:
Luiz and Vadim, I hadn't seen Tim's comment when I requested your reviews (had to refresh). Looks like I'll probably close this PR and make Tim's changes.

tests/plugins/test_orchestrate_build.py
```diff
@@ -500,6 +500,83 @@ def test_orchestrate_build_choose_clusters(tmpdir, clusters_x86_64,
...
+            }
+        }]
+    )
+
+    runner.run()
+
+
+def test_orchestrate_choose_cluster_retry_timeout(tmpdir):
+
+    (flexmock(OSBS).should_receive('list_builds')
```
vrutkovs: It would be nice to test a case where the results are eventually returned, but after the deadline

bfontecc007:
I have a new branch for this now.

PR #829, opened 2017-09-13
add-help plugin automatic metadata generation

ficap:


atomic_reactor/plugins/pre_add_help.py
```diff
@@ -66,6 +68,29 @@ def run(self):
             result['status'] = self.NO_HELP_FILE_FOUND
             return result
 
+        dockerfile = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        labels = dockerfile.labels
+
+        try:
+            help_file = open(help_path, 'r+')
```
twaugh: Maybe use `with open(help_path, 'r+') as help_file:` here, and remove the explicit `.close()` call?

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -66,6 +68,29 @@ def run(self):
...
+            lines = help_file.readlines()
+
+            if not lines[0].startswith("% "):
+                lines.insert(0, "%% %s (1) Container Image Pages\n" % labels.get("name") or "")
```
twaugh: Rather than `labels.get("name")` this should use `get_preferred_label(labels, 'name')`.

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -66,6 +68,29 @@ def run(self):
...
+                lines.insert(1, "%% %s\n" % labels.get("maintainer") or "")
+                lines.insert(2, "%% %s\n" % dt.fromtimestamp(atomic_reactor_start_time)
+                             .strftime(format="%B %-d, %Y"))
```
twaugh: Would be great to have a unit test to verify the content gets written correctly.

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -66,6 +69,33 @@ def run(self):
...
+                    lines.insert(1, "%% %s\n" %
+                                 (get_preferred_label(labels, "maintainer") or ""))
+                    lines.insert(2, "%% %s\n" % dt.fromtimestamp(atomic_reactor_start_time)
+                                 .strftime(format="%B %-d, %Y"))
+
+                help_file.seek(0)
+                help_file.truncate()
+                help_file.writelines(lines)
+
+                self.log.info("Added metadata to %s for generating nicer manpages." % help_path)
```
twaugh: We only need to write the file if we changed `lines`, so these four lines could be moved into the `if not ...:` block.

ficap: you're right my fault i'll fix that

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -66,6 +69,33 @@ def run(self):
             result['status'] = self.NO_HELP_FILE_FOUND
             return result
 
+        dockerfile = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        labels = dockerfile.labels
+
+        with open(help_path, 'r+') as help_file:
+            try:
```
twaugh: Thought I added a comment about this but I can't find it.

IMHO we shouldn't even use `try:` here. If opening and/or rw-ing the file doesn't work for some reason, I think the build should just fail.

ficap: Yes you did but with statement handles only file closing but what in case that theres IOError which can occur we should catch it so the whole plugin doesn't fail.

ficap: I tried it of course and it works maybe you're using python < 2.6 which doesn't support this feature. But as I looked into travis tasks it tests only python 2.6 and greater so olders shouldn't be supported, should they?

twaugh: What I'm suggesting is that we take the `try:` out, as well as the entire `except ...:` block, and just have:
```
with open(...) as ...:
    # read it
    if changes_needed:
        # adjust it
        ...
```

If adjusting that file fails, I think the plugin should fail. Maybe others disagree (@vrutkovs, @lcarva ?)

lcarva: Agree that it should fail the plugin.

vrutkovs: It seems if the resulting help file cannot be successfully written during build it's probably broken - the build should fail then IMHO

This metadata doesn't look critical for man pages, it just makes the display nicer, right?

tests/plugins/test_add_help.py
```diff
@@ -240,3 +243,55 @@ def check_popen_fail(*args, **kwargs):
...
+            'name': AddHelpPlugin.key,
+            'args': {'help_file': filename}
+        }]
+    )
+    runner.run()
+    lines = ""
+    with open(help_markdown_path) as f:
+        lines = "".join(f.readlines())
+
+    example = """%% test (1) Container Image Pages
```
vrutkovs: nitpick: wrap the text in `textwrap.dedent` to preserve identation (example: https://github.com/projectatomic/atomic-reactor/blob/master/tests/plugins/test_add_filesystem.py#L144)

ficap: I did this intentionally because I wanted to workarout this situation
```
dedent(
        """        %% test (1) Container Image Pages       <--  this line as it should contain exact amount of spaces
        %% me
        %% %s
        foo""")
```
but if you want this solution there is no problem to fix that :)

atomic_reactor/plugins/pre_add_help.py
ficap: ok, this is more logical way. it seems I misunderstanded some concepts of plugins. I'll fix that.

tests/plugins/test_add_help.py
vrutkovs: >this line as it should contain the exact amount of spaces prefixing the real content

Not sure if I understood this correctly - the spaces must be there? If yes the text is not explicit about it.

>but I don't want empty line at start of the string
Use `"""\` to avoid adding a new line

Its a minor thing anyway, I'm fine with current implementation, unless the first line has to contain spaces - this might be important

atomic_reactor/plugins/pre_add_help.py
ficap: Already done but GH doesn't want to mark it as outdated.

ficap:
@twaugh Should I squash these commits into one and force push?

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -66,6 +69,26 @@ def run(self):
...
+                lines.insert(1, "%% %s\n" %
+                             (get_preferred_label(labels, "maintainer") or ""))
+                lines.insert(2, "%% %s\n" % dt.fromtimestamp(atomic_reactor_start_time)
+                             .strftime(format="%B %-d, %Y"))
+
+                help_file.seek(0)
+                help_file.truncate()
+                help_file.writelines(lines)
+
+                self.log.info("Added metadata to %s for generating nicer manpages." % help_path)
```
twaugh: For logging methods, it's better to pass formatting arguments separately. Also, the style for logging messages is not to use title-case for the first word, and not to end with a full-stop:
```
self.log.info("added metadata to %s for generating nicer manpages", help_path)
```

lcarva: @ficap, will you have time to address this minor change?

twaugh:
Draft release notes updated.

PR #830, opened 2017-09-13
Ensure koji-import can resolve primary images 

lcarva:
In arrangement 3, tag_conf is empty in orchestrator build.
This caused this plugin to always fail in this case as the
list of primary images is always empty. This changes addresses
the problem by falling back to the primary images as reported
by worker builds.


twaugh:
Looks good.

lcarva:

PR #831, opened 2017-09-14
new remove worker metadata plugin

rcerven:

### DIFF
--- a/atomic_reactor/plugins/exit_remove_worker_metadata.py
+++ b/atomic_reactor/plugins/exit_remove_worker_metadata.py
@@ -1,0 +1,39 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+from atomic_reactor.plugin import ExitPlugin
+from atomic_reactor.constants import PLUGIN_REMOVE_WORKER_METADATA_KEY
+from osbs.exceptions import OsbsResponseException
+
+
+def defer_removal(workflow, cf_map, osbs):
+    key = RemoveWorkerMetadataPlugin.key
+    workspace = workflow.plugin_workspace.setdefault(key, {})
+    workspace.setdefault('cf_maps_to_remove', set())
+    workspace['cf_maps_to_remove'].add((cf_map, osbs))
+
+
+class RemoveWorkerMetadataPlugin(ExitPlugin):
+    """
+    Remove worker metadata for each platform.
+    """
+
+    key = PLUGIN_REMOVE_WORKER_METADATA_KEY
+
+    def run(self):
+        """
+        Run the plugin.
+        """
+
+        workspace = self.workflow.plugin_workspace.get(self.key, {})
+        cf_maps_to_remove = workspace.get('cf_maps_to_remove', [])
+        for cm_key, osbs in cf_maps_to_remove:
+            try:
+                osbs.delete_config_map(cm_key)
+                self.log.debug("ConfigMap %s deleted", cm_key)
+            except OsbsResponseException as ex:
+                self.log.warning("Failed to delete ConfigMap %s: %s", cm_key, ex.message)
--- a/atomic_reactor/plugins/post_fetch_worker_metadata.py
+++ b/atomic_reactor/plugins/post_fetch_worker_metadata.py
@@ -8,6 +8,7 @@ of the BSD license. See the LICENSE file for details.
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.plugins.build_orchestrate_build import get_worker_build_info
 from atomic_reactor.constants import PLUGIN_FETCH_WORKER_METADATA_KEY
+from atomic_reactor.plugins.exit_remove_worker_metadata import defer_removal
 
 
 class FetchWorkerMetadataPlugin(PostBuildPlugin):
@@ -55,5 +56,7 @@ class FetchWorkerMetadataPlugin(PostBuildPlugin):
             metadata = cm_data.get_data_by_key(cm_frag_key)
 
             metadatas[platform] = metadata
+
+            defer_removal(self.workflow, cm_key, osbs)
 
         return metadatas
--- a/tests/plugins/test_remove_worker_metadata.py
+++ b/tests/plugins/test_remove_worker_metadata.py
@@ -1,0 +1,131 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import print_function, unicode_literals
+
+import os
+
+from flexmock import flexmock
+
+from atomic_reactor.core import DockerTasker
+from atomic_reactor.constants import PLUGIN_REMOVE_WORKER_METADATA_KEY
+from atomic_reactor.inner import DockerBuildWorkflow
+from atomic_reactor.plugin import ExitPluginsRunner
+from atomic_reactor.util import ImageName
+from atomic_reactor.plugins.exit_remove_worker_metadata import (defer_removal)
+from osbs.exceptions import OsbsResponseException
+
+from tests.constants import MOCK_SOURCE, TEST_IMAGE, INPUT_IMAGE
+from tests.docker_mock import mock_docker
+import pytest
+
+
+class MockConfigMapResponse(object):
+    def __init__(self, data):
+        self.data = data
+
+    def get_data_by_key(self, key):
+        return self.data[key]
+
+
+class MockOSBS(object):
+    def __init__(self, config_map):
+        self.config_map = config_map
+
+    def delete_config_map(self, name):
+        if name:
+            return name in self.config_map
+        else:
+            raise OsbsResponseException("Failed to delete config map", 404)
+
+
+class MockSource(object):
+
+    def __init__(self, tmpdir):
+        tmpdir = str(tmpdir)
+        self.dockerfile_path = os.path.join(tmpdir, 'Dockerfile')
+        self.path = tmpdir
+
+    def get_dockerfile_path(self):
+        return self.dockerfile_path, self.path
+
+
+class MockInsideBuilder(object):
+
+    def __init__(self):
+        mock_docker()
+        self.tasker = DockerTasker()
+        self.base_image = ImageName(repo='fedora', tag='25')
+        self.image_id = 'image_id'
+        self.image = INPUT_IMAGE
+        self.df_path = 'df_path'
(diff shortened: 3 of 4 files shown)


rcerven:
unit tests not done yet

atomic_reactor/plugins/exit_remove_worker_metadata.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class RemoveWorkerMetadataPlugin(ExitPlugin):
+    """
+    Remove worker metadata for each platform.
+    """
+
+    key = PLUGIN_REMOVE_WORKER_METADATA_KEY
+
+    def run(self):
```
lcarva: Most of the code in this plugin is a direct duplication of `fetch_worker_metadata`.
Let's abstract it.

atomic_reactor/plugins/exit_remove_worker_metadata.py
```diff
@@ -0,0 +1,46 @@
...
+            cmlen = len(kind)
+            cm_key_tmp = build_annotations['metadata_fragment']
+
+            if not cm_key_tmp or cm_key_tmp[:cmlen] != kind:
+                self.log.warning("Bad ConfigMap %s for platform %s", cm_key_tmp, platform)
+                continue
+
+            cm_key = cm_key_tmp[cmlen:]
+            osbs.delete_config_map(cm_key)
+            self.log.debug("Deleting ConfigMap %s", cm_key)
```
lcarva: At this point you have already deleted, so "Deleting" is inaccurate.

rcerven: ok

atomic_reactor/plugins/exit_remove_worker_metadata.py
rcerven: done

atomic_reactor/plugins/exit_remove_worker_metadata.py
```diff
@@ -0,0 +1,38 @@
...
+
+__all__ = ('RemoveWorkerMetadataPlugin', )
+
+
+def defer_removal(workflow, cf_map, osbs):
+    key = RemoveWorkerMetadataPlugin.key
+    workflow.plugin_workspace.setdefault(key, {})
+    workspace = workflow.plugin_workspace[key]
+    workspace.setdefault('cf_maps_to_remove', set())
+    workspace['cf_maps_to_remove'].add((cf_map, osbs))
```
vrutkovs: There is no need to store `osbs` here - it would be the same for all configmaps. Its easier to fetch `osbs` in `def run` once

atomic_reactor/plugins/exit_remove_worker_metadata.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def run(self):
+        """
+        Run the plugin.
+        """
+
+        workspace = self.workflow.plugin_workspace.get(self.key, {})
+        cf_maps_to_remove = workspace.get('cf_maps_to_remove', [])
+        for cm_key, osbs in cf_maps_to_remove:
+            osbs.delete_config_map(cm_key)
```
vrutkovs: Needs exception handling, so that one wrong ConfigMap wouldn't stop removing others

atomic_reactor/plugins/exit_remove_worker_metadata.py
```diff
@@ -0,0 +1,38 @@
...
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+from atomic_reactor.plugin import ExitPlugin
+from atomic_reactor.constants import PLUGIN_REMOVE_WORKER_METADATA_KEY
+
+__all__ = ('RemoveWorkerMetadataPlugin', )
+
+
+def defer_removal(workflow, cf_map, osbs):
```
vrutkovs: This function should be moved to `utils/` as its being used in a different plugin - or just to `atomic_reactor/plugins/post_fetch_worker_metadata.py`

tests/plugins/test_remove_worker_metadata.py
```diff
@@ -0,0 +1,127 @@
...
+    runner = ExitPluginsRunner(
+        None,
+        workflow,
+        [{
+            'name': PLUGIN_REMOVE_WORKER_METADATA_KEY,
+            "args": {}
+        }]
+    )
+
+    (flexmock(osbs)
```
vrutkovs: No need to mock `delete_config_map` twice:
```
.should_call("delete_config_map")
.with_args('build-1-ppc64le-md')
.with_args('build-1-x86_64-md')
```

This would assume `ordered()` too

atomic_reactor/plugins/post_fetch_worker_metadata.py
```diff
@@ -56,4 +57,6 @@ def run(self):
 
             metadatas[platform] = metadata
 
+            defer_removal(self.workflow, cm_key, osbs)
```
lcarva: +1 for this approach

atomic_reactor/plugins/exit_remove_worker_metadata.py
```diff
@@ -0,0 +1,38 @@
...
+    key = RemoveWorkerMetadataPlugin.key
+    workflow.plugin_workspace.setdefault(key, {})
```
lcarva: fyi, `setdefault` returns the value. This can be simplified to: 
```python
workspace = workflow.plugin_workspace.setdefault(key, {})
```

atomic_reactor/plugins/exit_remove_worker_metadata.py
lcarva: That's not true. Each configmap comes from a different cluster, requiring a different osbs api object.

tests/plugins/test_remove_worker_metadata.py
vrutkovs: The tests should also make sure a correct - worker-specific - `osbs` instance is used here

atomic_reactor/plugins/exit_remove_worker_metadata.py
lcarva: Nack on that. We store the osbs object from `orchestrate_build` exactly for this reason. Any plugin that needs to reach out to worker clusters will use it, just like `fetch_metadata_worker` does.
Spawning a new osbs instance is not as straight forward as you may think, as it requires reading the reactor config and client config again.

vrutkovs: Hmm, okay, lets leave it for further improvements then - I assumes its nearly effortless

atomic_reactor/plugins/exit_remove_worker_metadata.py
rcerven: ok

atomic_reactor/plugins/exit_remove_worker_metadata.py
rcerven: ok

tests/plugins/test_remove_worker_metadata.py
rcerven: ok

atomic_reactor/plugins/exit_remove_worker_metadata.py
twaugh: You think? It works like this for remove_built_image.

lcarva: Keeping this function here seems reasonable to me.

atomic_reactor/plugins/exit_remove_worker_metadata.py
```diff
@@ -0,0 +1,41 @@
...
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+from atomic_reactor.plugin import ExitPlugin
+from atomic_reactor.constants import PLUGIN_REMOVE_WORKER_METADATA_KEY
+from osbs.exceptions import OsbsResponseException
+
+__all__ = ('RemoveWorkerMetadataPlugin', )
```
lcarva: I'd just remove this as it's not really needed since we don't do `import *` anywhere. In which case, we'd probably want to include `defer_removal` in it.

atomic_reactor/plugins/exit_remove_worker_metadata.py
```diff
@@ -0,0 +1,41 @@
...
+        """
+
+        workspace = self.workflow.plugin_workspace.get(self.key, {})
+        cf_maps_to_remove = workspace.get('cf_maps_to_remove', [])
+        for cm_key, osbs in cf_maps_to_remove:
+            try:
+                osbs.delete_config_map(cm_key)
+                self.log.debug("ConfigMap %s deleted", cm_key)
+            except OsbsResponseException as ex:
+                self.warning("Failed to delete ConfigMap %s: %s", cm_key, ex.message)
```
lcarva: Don't think this is covered by unit tests.

tests/plugins/test_remove_worker_metadata.py
twaugh: We want each flexmock...should_call to be applied to the correct osbs instance. So something like:
```
for name in ('build-1-x86_64-md', 'build-1-ppc64le-md'):
    osbs = MockOSBS({name: metadata})
    (flexmock(osbs)
     .should_call("delete_config_map")
     .with_args(name)
     .once()
     .and_return(True))
```

The `.once()` is important to make sure it's really called.

atomic_reactor/plugins/exit_remove_worker_metadata.py
rcerven: ok

tests/plugins/test_remove_worker_metadata.py
rcerven: ok

atomic_reactor/plugins/exit_remove_worker_metadata.py
rcerven: ok

tests/plugins/test_remove_worker_metadata.py
```diff
@@ -0,0 +1,125 @@
...
+def mock_workflow(tmpdir):
+    workflow = DockerBuildWorkflow(MOCK_SOURCE, TEST_IMAGE)
+    setattr(workflow, 'builder', MockInsideBuilder())
+    setattr(workflow, 'source', MockSource(tmpdir))
+    setattr(workflow.builder, 'source', MockSource(tmpdir))
+
+    return workflow
+
+
+@pytest.mark.parametrize('name', ['build-1-x86_64-md', 'build-1-ppc64le-md', None])
```
vrutkovs: This would call the test with one name only, not a combination of both. As a result we don't verify if correct osbs was used

vrutkovs:
Release notes updated

PR #832, opened 2017-09-14
pre_bump_release: make sure base_release is always a string

vrutkovs:
/cc @owtaylor

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -90,9 +90,10 @@ class BumpReleasePlugin(PreBuildPlugin):
         # This is brute force, but trying to use getNextRelease() would be fragile
         # magic depending on the exact details of how koji increments the release,
         # and we expect that the number of builds for any one base_release will be small.
+        release = base_release or '1'
         suffix = 1
         while True:
-            next_release = base_release + '.' + str(suffix)
+            next_release = '%s.%s' % (release, suffix)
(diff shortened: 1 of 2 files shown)


tests/plugins/test_bump_release.py
```diff
@@ -215,8 +218,9 @@ def getBuild(self, build_info):
         labels = {
             'com.redhat.component': 'component1',
             'version': 'fc26',
-            'release': base_release
         }
+        if base_release:
```
rcerven: why this if? 

if base_release will be None, it will be set to '0'

vrutkovs: Without this block the tests would pass `base_release` as a string `'None'`, but not `None`

rcerven: ahh ok

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -90,9 +90,10 @@ def get_next_release_append(self, component, version, base_release):
         # This is brute force, but trying to use getNextRelease() would be fragile
         # magic depending on the exact details of how koji increments the release,
         # and we expect that the number of builds for any one base_release will be small.
+        release = base_release or '0'
```
lcarva: Should it be 1 instead? At least that's what Koji does.

vrutkovs: Oh, agreed. Fixed

lcarva:

PR #833, opened 2017-09-18
Handle koij-parent without result

lcarva:
The koji-parent plugin may return None, instead of a dict,
if it the base image does not have the required Labels.


lcarva:

PR #834, opened 2017-09-19
layer sizes in image archive metadata

rcerven:

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -28,6 +28,7 @@ from atomic_reactor.plugin import (
     PrePublishPluginsRunner,
 )
 from atomic_reactor.source import get_source_instance_for
+from atomic_reactor.constants import INSPECT_ROOTFS, INSPECT_ROOTFS_LAYERS
 from atomic_reactor.util import ImageName
 from atomic_reactor.build import BuildResult
 from atomic_reactor import get_logging_encoding
@@ -310,6 +311,7 @@ class DockerBuildWorkflow(object):
 
         self.builder = None
         self.built_image_inspect = None
+        self.layer_sizes = []
         self._base_image_inspect = None
 
         self.pulled_base_images = set()
@@ -413,6 +415,16 @@ class DockerBuildWorkflow(object):
             except PluginFailedException as ex:
                 logger.error("one or more prepublish plugins failed: %s", ex)
                 raise
+
+            if self.build_result.is_image_available():
+                history = self.builder.tasker.d.history(self.builder.image_id)
+                diff_ids = self.built_image_inspect[INSPECT_ROOTFS][INSPECT_ROOTFS_LAYERS]
+
+                # diff_ids is ordered oldest first
+                # history is ordered newest first
+                # We want layer_sizes to be ordered oldest first
+                self.layer_sizes = [{"diff_id": diff_id, "size": layer['Size']}
+                                    for (diff_id, layer) in zip(diff_ids, reversed(history))]
 
             postbuild_runner = PostBuildPluginsRunner(self.builder.tasker, self,
                                                       self.postbuild_plugins_conf,
--- a/docs/koji.md
+++ b/docs/koji.md
@@ -73,6 +73,7 @@ The docker map has these entries:
 - `repositories` (string list): docker pull specifications for the name:version-release image in the docker registry (or in Crane, if Pulp/Crane integration is used), by tag and by digests (there may be multiple digests, e.g. v2 schema 1 and v2 schema 2)
 - `config` (map): the [v2 schema 2 'config' object](https://docs.docker.com/registry/spec/manifest-v2-2/#image-manifest-field-descriptions) but with the 'container_config' entry removed
 - `tags` (string list): the image tags (i.e. the part after the ":") applied to this image when it was tagged and pushed
+- `layer_sizes` (map list): the image layer uncompressed sizes, the oldest layer first (the size information comes from docker history command)
 
 ## Example
 
@@ -131,6 +132,12 @@ Example type-specific content generator metadata in context:
       "docker": {
         "id": "sha256:abc123def...",
         "parent_id": "sha256:123def456...",
+        "layer_sizes": [
+            {"diff_id": "sha256:123def456...",
+             "size": 1234556},
+            {"diff_id": "sha256:456789013...",
+             "size": 9494949}
+        ],
         "repositories": [
           "registry.example.com/product/package-name:1.0.0-1",
           "registry.example.com/product/package-name@sha256:123def..."
--- a/tests/plugins/test_docker_api.py
+++ b/tests/plugins/test_docker_api.py
@@ -17,6 +17,7 @@ from atomic_reactor.plugin import PluginFailedException
 from atomic_reactor.build import InsideBuilder, BuildResult
 from atomic_reactor.util import ImageName, CommandResult
 from atomic_reactor.inner import DockerBuildWorkflow
+from atomic_reactor.constants import INSPECT_ROOTFS, INSPECT_ROOTFS_LAYERS
 
 from tests.docker_mock import mock_docker
 from flexmock import flexmock
@@ -24,7 +25,14 @@ import pytest
 from tests.constants import MOCK_SOURCE
 
 
+class MockDocker(object):
+    def history(self, name):
+        return []
+
 class MockDockerTasker(object):
+    def __init__(self):
+        self.d = MockDocker()
+
     def inspect_image(self, name):
         return {}
 
@@ -64,7 +72,7 @@ class MockInsideBuilder(object):
         return {'Id': 'some'}
 
     def inspect_built_image(self):
-        return None
+        return {INSPECT_ROOTFS: {INSPECT_ROOTFS_LAYERS: []}}
 
     def ensure_not_built(self):
         pass
--- a/tests/test_inner.py
+++ b/tests/test_inner.py
@@ -32,6 +32,7 @@ import signal
 
 from atomic_reactor.inner import BuildResults, BuildResultsEncoder, BuildResultsJSONDecoder
 from atomic_reactor.inner import DockerBuildWorkflow
+from atomic_reactor.constants import INSPECT_ROOTFS, INSPECT_ROOTFS_LAYERS
 
 
 BUILD_RESULTS_ATTRS = ['build_logs',
@@ -68,7 +69,18 @@ def test_build_results_decoder():
         assert getattr(results, attr) == getattr(expected_results, attr)
 
 
+class MockDocker(object):
+    def history(self, name):
+        return [{'Size': 1, 'Id': "sha256:layer1-newest"},
+                {'Size': 2, 'Id': "sha256:layer2"},
+                {'Size': 3, 'Id': "sha256:layer3"},
+                {'Size': 4, 'Id': "sha256:layer4-oldest"}]
+
+
 class MockDockerTasker(object):
+    def __init__(self):
+        self.d = MockDocker()
+
     def inspect_image(self, name):
         return {}
 
@@ -117,7 +129,10 @@ class MockInsideBuilder(object):
         return {'Id': 'some'}
 
     def inspect_built_image(self):
-        return None
+        return {INSPECT_ROOTFS: {INSPECT_ROOTFS_LAYERS: ['sha256:diff_id1-oldest',
+                                                         'sha256:diff_id2',
+                                                         'sha256:diff_id3',
+                                                         'sha256:diff_id4-newest']}}
 
     def ensure_not_built(self):
         pass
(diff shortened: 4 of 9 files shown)


twaugh:
Don't forget to update docs/koji.md.

rcerven:
ahhh ok will do that thanks for reminding

tests/plugins/test_docker_api.py
```diff
@@ -31,6 +31,9 @@ def inspect_image(self, name):
     def build_image_from_path(self):
         return True
 
+    def history(self, name):
```
vrutkovs: That fixes unittest, but doesn't fix smoketests:
```
2017-09-19 14:50:28,082 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO - Traceback (most recent call last):
2017-09-19 14:50:28,083 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO -   File "/usr/bin/atomic-reactor", line 9, in <module>
2017-09-19 14:50:28,084 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO -     load_entry_point('atomic-reactor==1.6.26.3', 'console_scripts', 'atomic-reactor')()
2017-09-19 14:50:28,085 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO -   File "/usr/lib/python2.7/site-packages/atomic_reactor/cli/main.py", line 311, in run
2017-09-19 14:50:28,085 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO -     cli.run()
2017-09-19 14:50:28,087 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO -   File "/usr/lib/python2.7/site-packages/atomic_reactor/cli/main.py", line 294, in run
2017-09-19 14:50:28,087 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO -     args.func(args)
2017-09-19 14:50:28,088 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO -   File "/usr/lib/python2.7/site-packages/atomic_reactor/cli/main.py", line 95, in cli_inside_build
2017-09-19 14:50:28,089 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO -     substitutions=args.substitute)
2017-09-19 14:50:28,094 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO -   File "/usr/lib/python2.7/site-packages/atomic_reactor/inner.py", line 487, in build_inside
2017-09-19 14:50:28,094 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO -     build_result = dbw.build_docker_image()
2017-09-19 14:50:28,096 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO -   File "/usr/lib/python2.7/site-packages/atomic_reactor/inner.py", line 419, in build_docker_image
2017-09-19 14:50:28,097 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO -     for layer_info in self.builder.tasker.history(self.builder.image_id):
2017-09-19 14:50:28,098 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO - AttributeError: 'DockerTasker' object has no attribute 'history'
```

docs/koji.md
```diff
@@ -131,6 +131,10 @@ Example type-specific content generator metadata in context:
       "docker": {
         "id": "sha256:abc123def...",
         "parent_id": "sha256:123def456...",
+        "layer_sizes": [
+            {"sha256:123def456...": "1234556"},
+            {"sha256:456789013...": "1234"}
```
twaugh: If we make this a list and not a map, we'd better be clear in the documentation about which order it is.

rcerven: ahh right missed the description section,
yeah I wanted to keep the order

rcerven: added

atomic_reactor/inner.py
```diff
@@ -414,6 +416,14 @@ def build_docker_image(self):
                 logger.error("one or more prepublish plugins failed: %s", ex)
                 raise
 
+            if self.build_result.is_image_available():
+                history = self.builder.tasker.d.history(self.builder.image_id)
+                diff_ids = self.built_image_inspect[INSPECT_ROOTFS]['Layers']
+                diff_ids.reverse()
+                for layer_idx in range(len(history)):
+                    self.built_image_layer_sizes.insert(0, {diff_ids[layer_idx]:
+                                                                history[layer_idx]['Size']})
```
twaugh: This might be easier to follow as a list comprehension, and definitely needs comments explaining which order things start off in.

e.g. (not sure if I have the order right)
```
# diff_ids is ordered oldest first
# history is ordered newest first
# We want layer_sizes to be ordered oldest first
layer_sizes = [{diff_id: layer['Size']}
               for (diff_id, layer) in zip(diff_ids, reversed(history))]
```

docs/koji.md
twaugh: This has strings for the values, but aren't they really integers?

tests/plugins/test_docker_api.py
```diff
@@ -17,14 +17,22 @@
...
 
 from tests.docker_mock import mock_docker
 from flexmock import flexmock
 import pytest
 from tests.constants import MOCK_SOURCE
 
 
+class MockDocker(object):
+    def history(self, name):
+        return []
```
twaugh: Don't we want to include some mocked data here, to test that we get the order right?

docs/koji.md
```diff
@@ -73,6 +73,8 @@ The docker map has these entries:
 - `repositories` (string list): docker pull specifications for the name:version-release image in the docker registry (or in Crane, if Pulp/Crane integration is used), by tag and by digests (there may be multiple digests, e.g. v2 schema 1 and v2 schema 2)
 - `config` (map): the [v2 schema 2 'config' object](https://docs.docker.com/registry/spec/manifest-v2-2/#image-manifest-field-descriptions) but with the 'container_config' entry removed
 - `tags` (string list): the image tags (i.e. the part after the ":") applied to this image when it was tagged and pushed
+- `layer_sizes` (map list): the image layer uncompressed sizes, the oldest layer first (the size information comes from docker history command),
```
vrutkovs: `layer_sizes` is a list

twaugh: A list of maps, i.e. a map list. We use that elsewhere e.g. "string list".

vrutkovs: This sounds a bit confusing to me, maybe we should expand this to 'a list of layers, each represented by a map of layer ID and its size`?

also `map` is not Pythonic, wouldn't it be better to use `dict` instead?

twaugh: Technically we're talking about JSON here, not Python. I used "map" originally following the content generator metadata format description here: https://docs.pagure.org/koji/content_generator_metadata/

How about we follow the existing style in this PR, but have a new PR to switch to something more clear? We can talk about ways of expressing it better in that new PR. One idea is to mix Python typing hints like "List[Mapping]" with JSON schema type names, giving something like "array[object]".

Thinking about this, though, makes me wonder if we have the format right to begin with. Perhaps we should name the fields like this?:
```
"layer_sizes": [
  {
    "diff_id": "sha256:012345...",
    "size": 12345
  },
  {
    "diff_id": "sha256:123456...",
    size: 23456
  }
]
```

vrutkovs: >Technically we're talking about JSON here, not Python

Oh, right. Agree

> have a new PR to switch to something more clear

Good idea, other fields might need some clarification too.

>Perhaps we should name the fields like this?

its good idea in case some additional fields would be added - but I can't think of any. Making `diff_id` the key in the dict sounds fine to me

docs/koji.md
rcerven: yes will fix

atomic_reactor/inner.py
```diff
@@ -414,6 +416,14 @@ def build_docker_image(self):
                 logger.error("one or more prepublish plugins failed: %s", ex)
                 raise
 
+            if self.build_result.is_image_available():
+                history = self.builder.tasker.d.history(self.builder.image_id)
+                diff_ids = self.built_image_inspect[INSPECT_ROOTFS]['Layers']
```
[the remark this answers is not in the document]
rcerven: yes will make constant

vrutkovs:
Tests should verify that the layers are returned in correct order and correct dict is set in metadata

rcerven:
added test to check layer_sizes metadata

atomic_reactor/inner.py
```diff
@@ -414,6 +416,16 @@ def build_docker_image(self):
...
 
+            if self.build_result.is_image_available():
+                history = self.builder.tasker.d.history(self.builder.image_id)
+                diff_ids = self.built_image_inspect[INSPECT_ROOTFS][INSPECT_ROOTFS_LAYERS]
+
+                # diff_ids is ordered oldest first
+                # history is ordered newest first
+                # We want layer_sizes to be ordered oldest first
+                self.layer_sizes = [{"diff_id": diff_id, "size": layer['Size']}
+                                    for (diff_id, layer) in zip(diff_ids, reversed(history))]
```
twaugh: By the way, is there a particular reason to do this in `inner.py` instead of `post_koji_upload.py`?

rcerven: apart from the information gathered after squash and be available to other plugins when needed, no

twaugh: Alright. I'd rather have this method (`build_docker_image`) only be concerned with dispatching the plugins for each phase in order, and move the code inspecting layer sizes into koji_upload.

twaugh: Now that I've remembered that exit_koji_promote.py will also need this, maybe it's fine to leave this where it is.

rcerven: koji_promote plugin isn't used since arrangement 3

rcerven:
exit_koji_promote.py updated

tests/test_inner.py
```diff
@@ -68,7 +69,18 @@ def test_build_results_decoder():
         assert getattr(results, attr) == getattr(expected_results, attr)
 
 
+class MockDocker(object):
```
vrutkovs: Why would we want to create a new copy of MockDocker and not reuse one in `test_docker_api`? `inner.py` might not always use docker btw

tests/test_inner.py
```diff
@@ -117,7 +129,10 @@ def get_built_image_info(self):
         return {'Id': 'some'}
 
     def inspect_built_image(self):
-        return None
+        return {INSPECT_ROOTFS: {INSPECT_ROOTFS_LAYERS: ['sha256:1234',
```
vrutkovs: `def history` returns size for `Id: "sha256:1"` and `ROOTFS_LAYERS` returns `'sha256:1234'` - and tests pass? This is weird, must be a bug somewhere

rcerven: what?? we need diff_ids from rootfs, and sizes from history, those are different IDs, one is diff_ids and another is regular id

twaugh: History describes the layers in terms of their image IDs; inspect describes them in terms of diff_ids. Both look like "sha256:...".

Perhaps the test would look less confusing if the "sha256:" strings described themselves better, e.g.

- sha256:diff_id1
- sha256:diff_id2
- sha256:diff_id3
- ...

and

- sha256:layer0-oldest
- sha256:layer1
- sha256:layer2
- sha256:layer3-newest

vrutkovs: >"sha256:" strings described themselves better

good idea

tests/test_inner.py
rcerven: well they both can share that class, but we don't need data in that docker_api, but we need it in inner

tests/plugins/test_docker_api.py
twaugh: Oh, this is in the generic test machinery, not in the testing for layer sizes, I get it now.

tests/test_inner.py
```diff
@@ -68,7 +69,18 @@ def test_build_results_decoder():
         assert getattr(results, attr) == getattr(expected_results, attr)
 
 
+class MockDocker(object):
+    def history(self, name):
+        return [{'Size': 1, 'Id': "sha256:layer1-oldest"},
+                {'Size': 2, 'Id': "sha256:layer2"},
+                {'Size': 3, 'Id': "sha256:layer3"},
+                {'Size': 4, 'Id': "sha256:layer4-newest"}]
```
twaugh: History returns layers newest first.

tests/plugins/test_docker_api.py
rcerven: yeah test for layer sizes is in test_inner

tests/test_inner.py
```diff
@@ -1304,3 +1319,34 @@ def test_add_pulp_registry():
...
+                                                       }}],
+                                   plugin_files=[this_file])
+
+    workflow.build_docker_image()
+
+    expected = [
+        {'diff_id': u'sha256:diff_id1', 'size': 4},
+        {'diff_id': u'sha256:diff_id2', 'size': 3},
+        {'diff_id': u'sha256:diff_id3', 'size': 2},
+        {'diff_id': u'sha256:diff_id4', 'size': 1}
```
twaugh: Wrong expectation. We expect sha256:diff_id1 with 1 (oldest layer), sha256:diff_id2 with 2, etc.

tests/test_inner.py
rcerven: yeah wrong order

tests/test_inner.py
rcerven: expectation is good, but because of that wrong history order it seemed wrong

tests/test_inner.py
vrutkovs: It would be easier to unify those, so that `MockDocker` class from `tests/plugins/test_docker_api.py` would not return empty history always

rcerven: it might as well break other test, and data are really only needed for inner

PR #836, opened 2017-09-20
Apply image ID from pulp_pull for all single-arch builds

twaugh:
If building for a single architecture, always look for pulp_pull's re-calculated image ID regardless of architecture.


twaugh:
Not sure what docs to update really, we don't even mention pulp_pull in docs/plugins.md. :-|

Release notes:
* `pulp_pull` plugin's image ID from pulling the image is used regardless of platform name; this only applies to installations using Pulp < 2.14

lcarva:
Updated release notes draft with suggested text.

PR #837, opened 2017-09-20
requests are using timeout

rcerven:

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -15,7 +15,7 @@ import os
 import re
 from pipes import quote
 import requests
-from requests.exceptions import ConnectionError, SSLError, HTTPError, RetryError
+from requests.exceptions import ConnectionError, SSLError, HTTPError, RetryError, Timeout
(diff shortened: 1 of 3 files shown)


atomic_reactor/util.py
```diff
@@ -994,6 +994,26 @@ def get_field(self, field_name, args, kwargs):
...
 
 
+class SessionWithTimeout(requests.Session):
+    """
+    requests Session with added timeout
+    """
+    def __init__(self, *args, **kwargs):
+        super(SessionWithTimeout, self).__init__(*args, **kwargs)
+
+    def get(self, *args, **kwargs):
```
vrutkovs: Wouldn't it be easier to override just `def request` there?

rcerven: not sure

lcarva: That's worth investigating.

rcerven: changed to override just request

vrutkovs:
Release notes updated

PR #838, opened 2017-09-20
Import image tag conf

lcarva:
Read primary images from either tag_conf, or worker builds annotations.


lcarva:
Updated release notes draft:
> - `import_image` reads primary tags from either tag_conf or worker builds annotations

PR #840, opened 2017-09-21
koji_promote adds media_type to build metadata

rcerven:

### DIFF
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -134,10 +134,8 @@ class KojiImportPlugin(ExitPlugin):
                     # building for many, we know Pulp has schema 2
                     # support.
                     if len(worker_metadatas) == 1 and has_pulp_pull:
-                        exit_results = self.workflow.exit_results
-                        image_id, _ = exit_results[PLUGIN_PULP_PULL_KEY]
-                        if image_id is not None:
-                            instance['extra']['docker']['id'] = image_id
+                        if self.workflow.builder.image_id is not None:
+                            instance['extra']['docker']['id'] = self.workflow.builder.image_id
 
                     # update repositories to point to Crane
                     if crane_registry:
@@ -240,7 +238,7 @@ class KojiImportPlugin(ExitPlugin):
         # Append media_types from pulp pull
         pulp_pull_results = self.workflow.exit_results.get(PLUGIN_PULP_PULL_KEY)
         if pulp_pull_results:
-            media_types += pulp_pull_results[1]
+            media_types += pulp_pull_results
 
         if media_types:
             extra['image']['media_types'] = sorted(list(set(media_types)))
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -33,7 +33,8 @@ except ImportError:
         raise KeyError
 
 from atomic_reactor.constants import (PROG, PLUGIN_KOJI_PROMOTE_PLUGIN_KEY,
-                                      PLUGIN_KOJI_TAG_BUILD_KEY)
+                                      PLUGIN_KOJI_TAG_BUILD_KEY,
+                                      PLUGIN_PULP_PULL_KEY)
 from atomic_reactor.util import (get_version_of_tools, get_checksums,
                                  get_build_json, get_preferred_label,
                                  get_docker_architecture, df_parser,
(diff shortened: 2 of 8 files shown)


atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -563,6 +564,13 @@ def get_build(self, metadata):
                 else:
                     extra['filesystem_koji_task_id'] = task_id
 
+        media_types = []
```
vrutkovs: This should also append `['application/json']` if pulp_push was running

twaugh: I think pulp_pull already does that when reporting media types. For prod builds, pulp_push runs in the same OpenShift Build as pulp_pull.
https://github.com/projectatomic/atomic-reactor/blob/1.6.26.3/atomic_reactor/plugins/post_pulp_pull.py#L105-L109

vrutkovs: Oh, I see, I assumed `koji_promote` code is more recent. Its not needed then, yes

rcerven:
test added

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -134,10 +134,8 @@ def get_output(self, worker_metadatas):
                     # building for many, we know Pulp has schema 2
                     # support.
                     if len(worker_metadatas) == 1 and has_pulp_pull:
-                        exit_results = self.workflow.exit_results
-                        image_id, _ = exit_results[PLUGIN_PULP_PULL_KEY]
-                        if image_id is not None:
-                            instance['extra']['docker']['id'] = image_id
+                        if self.workflow.builder.image_id is not None:
```
vrutkovs: This change should be moved to a new PR

rcerven: agreed with Tim & Lui that it might be in this PR as 2nd commit

atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -89,7 +89,7 @@ def run(self):
         # Only run if the build was successful
         if self.workflow.build_process_failed:
             self.log.info("Not running for failed build")
-            return None, []
+            return []
```
vrutkovs: Should we reset `self.workflow.builder.image_id` here just in case then?

tests/plugins/test_koji_import.py
```diff
@@ -1152,6 +1152,9 @@ def test_koji_import_success(self, tmpdir, apis, docker_registry,
         workflow.prebuild_results[CheckAndSetRebuildPlugin.key] = is_autorebuild
         if pulp_pull:
             workflow.exit_results[PLUGIN_PULP_PULL_KEY] = pulp_pull
+
+        workflow.builder.image_id = expect_id
```
vrutkovs: The test should verify that `expect_id` is the same as `workflow.builder.image_id` in the end, no?

atomic_reactor/plugins/post_pulp_pull.py
rcerven: I can do that, but self.workflow.builder.image_id is by default None

vrutkovs: It could have been changed in any plugin

tests/plugins/test_koji_import.py
rcerven: test is verifying that metadata was set, verifying functionality of pulp pull is up to pulp_pull test
https://github.com/projectatomic/atomic-reactor/pull/840/files#diff-7a4c2d84966bd20f8d8c269e05f6df49R1184

vrutkovs: Okay, got it

twaugh:
Typo in commit message "unsued" -> "unused"

tests/plugins/test_pulp_pull.py
```diff
@@ -247,10 +247,8 @@ def test_pull_first_time(self, no_headers, broken_response, insecure, schema_ver
         # fall through to pulp_pull.
         plugin = PulpPullPlugin(tasker, workflow, insecure=insecure,
                                 timeout=0.1, retry_delay=0.25)
-        results, version = plugin.run()
+        version = plugin.run()
 
-        # Plugin return value is the new ID and schema
-        assert results == test_id
```
twaugh: We should be verifying that the image_id is updated in the workflow, shouldn't we?

twaugh: nm, I see it now

atomic_reactor/plugins/post_pulp_pull.py
rcerven: ok will reset it

rcerven:
typo fixed, stickler silenced

rcerven:
test works, but seems to be problem with that travis configuration

lcarva:
> test works, but seems to be problem with that travis configuration

Looks like a timeout when pulling from fedora's container registry. Just push something to retrigger it.

vrutkovs:
Release notes updated

PR #841, opened 2017-09-22
Implement Retry Logic for Cluster Selection

bfontecc007:
This is the second PR opened for retries. Closed #828

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -14,13 +14,17 @@ from multiprocessing.pool import ThreadPool
 import yaml
 import json
 import os
+from operator import attrgetter
 import random
 from string import ascii_letters
 import time
 import logging
+from datetime import timedelta
+import datetime as dt
+import copy
 
 from atomic_reactor.build import BuildResult
-from atomic_reactor.plugin import BuildStepPlugin, BuildCanceledException
+from atomic_reactor.plugin import BuildStepPlugin
 from atomic_reactor.plugins.pre_reactor_config import get_config
 from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
 from atomic_reactor.util import get_preferred_label, df_parser
@@ -35,6 +39,9 @@ ClusterInfo = namedtuple('ClusterInfo', ('cluster', 'platform', 'osbs', 'load'))
 WORKSPACE_KEY_BUILD_INFO = 'build_info'
 WORKSPACE_KEY_UPLOAD_DIR = 'koji_upload_dir'
 WORKSPACE_KEY_OVERRIDE_KWARGS = 'override_kwargs'
+FIND_CLUSTER_RETRY_DELAY = 15.0
+FAILURE_RETRY_DELAY = 10.0
+MAX_CLUSTER_FAILS = 20
 
 
 def get_worker_build_info(workflow, platform):
@@ -62,6 +69,61 @@ def override_build_kwarg(workflow, k, v):
     workspace = workflow.plugin_workspace.setdefault(key, {})
     override_kwargs = workspace.setdefault(WORKSPACE_KEY_OVERRIDE_KWARGS, {})
     override_kwargs[k] = v
+
+
+class UnknownPlatformException(Exception):
+    """ No clusters could be found for a platform """
+
+
+class AllClustersFailedException(Exception):
+    """ Each cluster has reached max_cluster_fails """
+
+
+class ClusterRetryContext(object):
+    def __init__(self, max_cluster_fails):
+        # how many times this cluster has failed
+        self.fails = 0
+
+        # datetime at which attempts can resume
+        self.retry_at = dt.datetime.utcfromtimestamp(0)
+
+        # the number of fail counts before this cluster is considered dead
+        self.max_cluster_fails = max_cluster_fails
+
+    @property
+    def failed(self):
+        """Is this cluster considered dead?"""
+        return self.fails >= self.max_cluster_fails
+
+    @property
+    def in_retry_wait(self):
+        """Should we wait before trying this cluster again?"""
+        return dt.datetime.now() < self.retry_at
+
+    def try_again_later(self, seconds):
+        """Put this cluster in retry-wait (or consider it dead)"""
+        if not self.failed:
+            self.fails += 1
+            self.retry_at = (dt.datetime.now() + timedelta(seconds=seconds))
+
+
+def wait_for_any_cluster(contexts):
+    """
+    Wait until any of the clusters are out of retry-wait
+
+    :param contexts: List[ClusterRetryContext]
+    :raises: AllClustersFailedException if no more retry attempts allowed
+    """
+    try:
+        earliest_retry_at = min(ctx.retry_at for ctx in contexts.values()
+                                if not ctx.failed)
+    except ValueError:  # can't take min() of empty sequence
+        raise AllClustersFailedException(
+            "Could not find appropriate cluster for worker build."
+        )
+
+    time_until_next = earliest_retry_at - dt.datetime.now()
+    time.sleep(max(timedelta(seconds=0), time_until_next).seconds)
 
 
 class WorkerBuildInfo(object):
@@ -171,7 +233,10 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
 
     def __init__(self, tasker, workflow, platforms, build_kwargs,
                  osbs_client_config=None, worker_build_image=None,
-                 config_kwargs=None):
+                 config_kwargs=None,
+                 find_cluster_retry_delay=FIND_CLUSTER_RETRY_DELAY,
+                 failure_retry_delay=FAILURE_RETRY_DELAY,
+                 max_cluster_fails=MAX_CLUSTER_FAILS):
         """
         constructor
 
@@ -183,12 +248,22 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
         :param worker_build_image: str, the builder image to use for worker builds
                                   (deprecated, use config_kwargs instead)
         :param config_kwargs: dict, keyword arguments to override worker configuration
+        :param find_cluster_retry_delay: the delay in seconds to try again reaching a cluster
+        :param failure_retry_delay: the delay in seconds to try again starting a build
+        :param max_cluster_fails: the maximum number of times a cluster can fail before being
+                                  ignored
         """
         super(OrchestrateBuildPlugin, self).__init__(tasker, workflow)
         self.platforms = set(platforms)
         self.build_kwargs = build_kwargs
         self.osbs_client_config = osbs_client_config
         self.config_kwargs = config_kwargs or {}
+        self.find_cluster_retry_delay = find_cluster_retry_delay
+        self.failure_retry_delay = failure_retry_delay
+        self.max_cluster_fails = max_cluster_fails
+        self.koji_upload_dir = self.get_koji_upload_dir()
+        self.fs_task_id = self.get_fs_task_id()
+        self.release = self.get_release()
 
         if worker_build_image:
             self.log.warning('worker_build_image is deprecated, use config_kwargs instead')
@@ -219,7 +294,8 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
     def get_current_builds(self, osbs):
         field_selector = ','.join(['status!={status}'.format(status=status.capitalize())
                                    for status in BUILD_FINISHED_STATES])
-        return len(osbs.list_builds(field_selector=field_selector))
+        with osbs.retries_disabled():
+            return len(osbs.list_builds(field_selector=field_selector))
 
     def get_cluster_info(self, cluster, platform):
         kwargs = deepcopy(self.config_kwargs)
@@ -229,42 +305,38 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
 
         conf = Configuration(**kwargs)
         osbs = OSBS(conf, conf)
-        try:
-            current_builds = self.get_current_builds(osbs)
-        except OsbsException as e:
-            # If the build is canceled reraise the error
-            if isinstance(e.cause, BuildCanceledException):
-                raise e
 
-            self.log.exception("Error occurred while listing builds on %s",
-                               cluster.name)
-            return ClusterInfo(cluster, platform, osbs, self.UNREACHABLE_CLUSTER_LOAD)
+        current_builds = self.get_current_builds(osbs)
 
         load = current_builds / cluster.max_concurrent_builds
         self.log.debug('enabled cluster %s for platform %s has load %s and active builds %s/%s',
                        cluster.name, platform, load, current_builds, cluster.max_concurrent_builds)
         return ClusterInfo(cluster, platform, osbs, load)
 
-    def choose_cluster(self, platform):
-        config = get_config(self.workflow)
-        clusters = [self.get_cluster_info(cluster, platform) for cluster in
-                    config.get_enabled_clusters_for_platform(platform)]
+    def get_clusters(self, platform, retry_contexts, all_clusters):
+        ''' return clusters sorted by load. '''
 
-        if not clusters:
-            raise RuntimeError('No clusters found for platform {}!'
-                               .format(platform))
+        possible_cluster_info = {}
+        candidates = set(copy.copy(all_clusters))
+        while candidates and not possible_cluster_info:
+            wait_for_any_cluster(retry_contexts)
 
-        reachable_clusters = [cluster for cluster in clusters
-                              if cluster.load != self.UNREACHABLE_CLUSTER_LOAD]
+            for cluster in sorted(candidates, key=attrgetter('priority')):
+                ctx = retry_contexts[cluster.name]
+                if ctx.in_retry_wait:
+                    continue
+                if ctx.failed:
+                    continue
+                try:
+                    cluster_info = self.get_cluster_info(cluster, platform)
+                    possible_cluster_info[cluster] = cluster_info
+                except OsbsException:
+                    ctx.try_again_later(self.find_cluster_retry_delay)
+            candidates -= set([c for c in candidates if retry_contexts[c.name].failed])
 
-        if not reachable_clusters:
-            raise RuntimeError('All clusters for platform {} are unreachable!'
-                               .format(platform))
-
-        selected = min(reachable_clusters, key=lambda c: c.load)
-        self.log.info('platform %s will use cluster %s',
-                      platform, selected.cluster.name)
-        return selected
+        ret = sorted(possible_cluster_info.values(), key=lambda c: c.cluster.priority)
+        ret = sorted(ret, key=lambda c: c.load)
+        return ret
 
     def get_release(self):
         labels = df_parser(self.workflow.builder.df_path, workflow=self.workflow).labels
@@ -353,17 +425,22 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
 
         return task_id
 
-    def do_worker_build(self, release, cluster_info, koji_upload_dir, task_id):
+    def do_worker_build(self, cluster_info):
         workspace = self.workflow.plugin_workspace.get(self.key, {})
         override_kwargs = workspace.get(WORKSPACE_KEY_OVERRIDE_KWARGS, {})
 
         build = None
 
         try:
-            kwargs = self.get_worker_build_kwargs(release, cluster_info.platform,
-                                                  koji_upload_dir, task_id)
+            kwargs = self.get_worker_build_kwargs(self.release, cluster_info.platform,
+                                                  self.koji_upload_dir, self.fs_task_id)
             kwargs.update(override_kwargs)
-            build = cluster_info.osbs.create_worker_build(**kwargs)
+            with cluster_info.osbs.retries_disabled():
+                build = cluster_info.osbs.create_worker_build(**kwargs)
+        except OsbsException:
+            self.log.exception('%s - failed to create worker build.',
+                               cluster_info.platform)
+            raise
         except Exception:
             self.log.exception('%s - failed to create worker build',
                                cluster_info.platform)
@@ -389,29 +466,45 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
                 except OsbsException:
                     pass
 
+    def select_and_start_cluster(self, platform):
+        ''' Choose a cluster and start a build on it '''
+
+        config = get_config(self.workflow)
+        clusters = config.get_enabled_clusters_for_platform(platform)
+
+        if not clusters:
+            raise UnknownPlatformException('No clusters found for platform {}!'
+                                           .format(platform))
+
+        retry_contexts = {
+            cluster.name: ClusterRetryContext(self.max_cluster_fails)
+            for cluster in clusters
+        }
+
+        while True:
+            possible_cluster_info = self.get_clusters(platform, retry_contexts, clusters)
+
+            for cluster_info in possible_cluster_info:
+                ctx = retry_contexts[cluster_info.cluster.name]
+                try:
+                    self.do_worker_build(cluster_info)
+                    return
+                except OsbsException:
+                    ctx.try_again_later(self.failure_retry_delay)
+                    # this will put the cluster in retry-wait when get_clusters runs
+
     def run(self):
-        release = self.get_release()
         platforms = self.get_platforms()
-        koji_upload_dir = self.get_koji_upload_dir()
-        task_id = self.get_fs_task_id()
 
         thread_pool = ThreadPool(len(platforms))
-        result = thread_pool.map_async(
-            lambda cluster_info: self.do_worker_build(release, cluster_info,
-                                                      koji_upload_dir, task_id),
-            [self.choose_cluster(platform) for platform in platforms]
-        )
+        result = thread_pool.map_async(self.select_and_start_cluster, platforms)
 
         try:
-            while not result.ready():
-                # The wait call is a blocking call which prevents signals
-                # from being processed. Wait for short intervals instead
-                # of a single long interval, so build cancellation can
-                # be handled virtually immediately.
-                result.wait(1)
+            result.get()
         # Always clean up worker builds on any error to avoid
         # runaway worker builds (includes orchestrator build cancellation)
         except Exception:
+            thread_pool.terminate()
             self.log.info('build cancelled, cancelling worker builds')
             if self.worker_builds:
                 ThreadPool(len(self.worker_builds)).map(
@@ -419,6 +512,9 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
             while not result.ready():
                 result.wait(1)
             raise
+        else:
+            thread_pool.close()
+            thread_pool.join()
 
         annotations = {'worker-builds': {
             build_info.platform: build_info.get_annotations()
@@ -436,7 +532,7 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
         }
 
         workspace = self.workflow.plugin_workspace.setdefault(self.key, {})
-        workspace[WORKSPACE_KEY_UPLOAD_DIR] = koji_upload_dir
+        workspace[WORKSPACE_KEY_UPLOAD_DIR] = self.koji_upload_dir
         workspace[WORKSPACE_KEY_BUILD_INFO] = {build_info.platform: build_info
                                                for build_info in self.worker_builds}
 
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -171,11 +171,17 @@ class WorkerBuildInfo(object):
         return annotations
 
     def get_fail_reason(self):
+        fail_reason = {}
+        if self.monitor_exception:
+            fail_reason['general'] = repr(self.monitor_exception)
+        elif not self.build:
+            fail_reason['general'] = 'build not started'
+
         if not self.build:
-            return {'general': 'build not started'}
+            return fail_reason
+
         build_annotations = self.build.get_annotations() or {}
         metadata = json.loads(build_annotations.get('plugins-metadata', '{}'))
-        fail_reason = {}
         if self.monitor_exception:
             fail_reason['general'] = repr(self.monitor_exception)
 
@@ -482,7 +488,18 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
         }
 
         while True:
-            possible_cluster_info = self.get_clusters(platform, retry_contexts, clusters)
+            try:
+                possible_cluster_info = self.get_clusters(platform,
+                                                          retry_contexts,
+                                                          clusters)
+            except AllClustersFailedException as ex:
+                cluster = ClusterInfo(None, platform, None, None)
+                build_info = WorkerBuildInfo(build=None,
+                                             cluster_info=cluster,
+                                             logger=self.log)
+                build_info.monitor_exception = repr(ex)
+                self.worker_builds.append(build_info)
+                return
 
             for cluster_info in possible_cluster_info:
                 ctx = retry_contexts[cluster_info.cluster.name]
--- a/tests/plugins/test_orchestrate_build.py
+++ b/tests/plugins/test_orchestrate_build.py
@@ -324,6 +324,8 @@ def test_orchestrate_build_annotations_and_labels(tmpdir, metadata_fragment):
                 'platforms': ['x86_64', 'ppc64le'],
                 'build_kwargs': make_worker_build_kwargs(),
                 'osbs_client_config': str(tmpdir),
+                'max_cluster_fails': 2,
+                'unreachable_cluster_retry_delay': .1
             }
         }]
     )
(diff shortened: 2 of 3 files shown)


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -231,40 +243,44 @@ def get_cluster_info(self, cluster, platform):
...
-        if not reachable_clusters:
-            raise RuntimeError('All clusters for platform {} are unreachable!'
-                               .format(platform))
-
-        selected = min(reachable_clusters, key=lambda c: c.load)
-        self.log.info('platform %s will use cluster %s',
-                      platform, selected.cluster.name)
-        return selected
+        reachable_clusters = sorted(reachable_clusters, key=lambda c: c.load)
+        return reachable_clusters
```
vrutkovs: nitpick: no need for additional var here

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -389,26 +406,45 @@ def do_worker_build(self, release, cluster_info, koji_upload_dir, task_id):
...
+
+        cluster_fails = Counter()
+        retry_at = {}
+        possible_clusters = None
+
+        while True:
+            if not clusters:
+                raise RuntimeError('Could not find appropriate cluster for worker build.')
+            possible_clusters = self.get_clusters(clusters, cluster_fails, platform, retry_at)
+            for cluster in possible_clusters or []:
```
vrutkovs: Instead of `or []` get clusters should return `[]` if no clusters found in `get_clusters`

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -389,26 +406,45 @@ def do_worker_build(self, release, cluster_info, koji_upload_dir, task_id):
...
+
+        config = get_config(self.workflow)
+        clusters = config.get_enabled_clusters_for_platform(platform)
+
+        if not clusters:
+            raise UnknownPlatformException('No clusters found for platform {}!'
+                                           .format(platform))
+
+        cluster_fails = Counter()
+        retry_at = {}
```
twaugh: Would be good to have a comment explaining what the keys are for these two.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -389,26 +406,45 @@ def do_worker_build(self, release, cluster_info, koji_upload_dir, task_id):
...
+        while True:
+            if not clusters:
+                raise RuntimeError('Could not find appropriate cluster for worker build.')
+            possible_clusters = self.get_clusters(clusters, cluster_fails, platform, retry_at)
+            for cluster in possible_clusters or []:
+                try:
+                    self.do_worker_build(self.release, cluster, self.koji_upload_dir,
+                                         self.fs_task_id)
+                    return
+                except RuntimeError:
```
twaugh: Is this the right exception to catch? It looks to me like do_worker_build doesn't raise an exception if creating the build failed -- it probably should, so we get into this retry loop. (Once created, it shouldn't raise an exception if monitoring fails.)

But we don't necessarily know what exception it will raise so maybe we need to catch `Exception` here? But also have a clause for `BuildCanceledException` to make sure that gets propagated immediately.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -231,40 +243,44 @@ def get_cluster_info(self, cluster, platform):
...
-            raise RuntimeError('No clusters found for platform {}!'
-                               .format(platform))
+    def get_clusters(self, all_clusters, cluster_fails, platform, retry_at):
+        ''' return clusters sorted by load '''
+
+        clusters = []
+        while all_clusters and not clusters:
+            for cluster in all_clusters:
+                if cluster_fails[cluster.name] >= self.max_cluster_fails:
+                    all_clusters.remove(cluster)
```
twaugh: This logic, of removing clusters once they have failed the maximum number of times, seems to be both here and in select_and_start_cluster. Is it needed in both places?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -389,26 +406,45 @@ def do_worker_build(self, release, cluster_info, koji_upload_dir, task_id):
...
+                raise RuntimeError('Could not find appropriate cluster for worker build.')
+            possible_clusters = self.get_clusters(clusters, cluster_fails, platform, retry_at)
+            for cluster in possible_clusters or []:
+                try:
+                    self.do_worker_build(self.release, cluster, self.koji_upload_dir,
+                                         self.fs_task_id)
+                    return
+                except RuntimeError:
+                    cluster_fails[cluster.name] += 1
+                    retry_at[cluster.name] = datetime.datetime.now()\
```
twaugh: Nit: can backslash-line-continuation be avoided here? Might look nicer to enclose the expression in (), or rephrase it somehow.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -231,40 +243,44 @@ def get_cluster_info(self, cluster, platform):
...
+                    if datetime.datetime.now() < retry_at[cluster.name]:
+                        continue
+                try:
+                    clusters.append(self.get_cluster_info(cluster, platform))
+                except BuildCanceledException:
+                    raise
+                except OsbsException:
+                    cluster_fails[cluster.name] += 1
+                    retry_at[cluster.name] = datetime.datetime.now()\
+                        + timedelta(seconds=self.find_cluster_retry_delay)
```
twaugh: This calculation is repeated elsewhere. Is it possible to avoid duplication?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -231,40 +243,44 @@ def get_cluster_info(self, cluster, platform):
...
+            if not clusters:
+                time.sleep((max(timedelta(seconds=0),
+                           min(retry_at.values()) - datetime.datetime.now())).seconds)
```
twaugh: The indentation alignment seems wrong here.

bfontecc007: How would you do it? This passes flake8

atomic_reactor/plugins/build_orchestrate_build.py
bfontecc007: The only way I can think of is to pull the do_worker_build logic into select clusters or pull select clusters into the main loop. Both would make a rather large and hard to understand loop.

twaugh: What about making a class to handle the retry context? e.g.
```
class ClusterRetryContexts(object):
    def __init__(self, clusters, max_cluster_fails):
        self.clusters = list(clusters)
        self.retry_at = {}
        self.failed = {}
        self.fails = Counter()
        self.max_cluster_fails = max_cluster_fails

    def in_retry_wait(self, cluster):
        if cluster.name in self.retry_at:
            return datetime.datetime.now() < self.retry_at[cluster.name]

        return False

    def try_again_later(self, cluster, seconds):
        name = cluster.name
        self.fails[name] += 1
        if self.fails[name] >= self.max_cluster_fails:
            self.failed[name] = True
        else:
            self.retry_at[name] = (datetime.datetime.now() +
                                   timedelta(seconds=seconds))

        return not self.failed[name]

    def wait_for_any(self):
        time_until_next = (min(retry_at.values()) -
                           datetime.datetime.now().seconds)
        time.sleep(max(timedelta(seconds=0), time_until_next))
```

Then you'd be able to do this sort of thing:
```
clusters = []
while all_clusters and not clusters:
    if retry_info.in_retry_wait(cluster):
        continue
    try:
        ...
    except OsbsException:
        if not retry_info.try_again_later(cluster, self.find_cluster_retry_delay):
            all_clusters.remove(cluster)
    if not clusters:
        retry_info.wait_for_any()
```

and similarly in select_and_start_cluster().

atomic_reactor/plugins/build_orchestrate_build.py
twaugh: What makes it hard to read is that the indented line does not line up with the parenthesis we're still in.

I'd personally prefer to the min and max calls in different statements, e.g.
```
next_retry_in = min(retry_at.values()) - datetime.datetime.now()
time.sleep(max(0, next_retry_in.seconds)
```

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -189,13 +195,48 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+            self.max_cluster_fails = max_cluster_fails
+            self.delay = find_cluster_retry_delay
+
+        def in_retry_wait(self, cluster):
+            if cluster.name in self.retry_at:
+                return datetime.datetime.now() < self.retry_at[cluster.name]
+
+            return False
+
+        def try_again_later(self, cluster, seconds):
```
twaugh: `seconds` is unused

bfontecc007: changed it to use seconds.

atomic_reactor/plugin.py
```diff
@@ -45,6 +45,10 @@ class BuildCanceledException(Exception):
     """Build was canceled"""
 
 
+class UnknownPlatformException(Exception):
```
lcarva: This seems out of scope here. Should it be defined in `build_orchestrate_build.py`?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -189,13 +195,47 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+        self.fs_task_id = self.get_fs_task_id()
+        self.release = self.get_release()
 
         if worker_build_image:
             self.log.warning('worker_build_image is deprecated, use config_kwargs instead')
             self.config_kwargs.setdefault('build_image', worker_build_image)
 
         self.worker_builds = []
 
+    class ClusterRetryContexts(object):
```
lcarva: Why is this class defined inside the plugin class?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -189,13 +195,47 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+            if self.fails[name] >= self.max_cluster_fails:
+                self.failed.add(name)
+            else:
+                self.retry_at[name] = (datetime.datetime.now() +
+                                       timedelta(seconds=seconds))
+
+            return name not in self.failed
+
+        def wait_for_any(self):
+            time_until_next = (min(self.retry_at.values()) - datetime.datetime.now())
```
lcarva: Outer parenthesis not needed?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -231,40 +272,38 @@ def get_cluster_info(self, cluster, platform):
         osbs = OSBS(conf, conf)
         try:
             current_builds = self.get_current_builds(osbs)
-        except OsbsException as e:
+        except OsbsException:
```
lcarva: Unnecessary try-except block.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -231,40 +272,38 @@ def get_cluster_info(self, cluster, platform):
...
-        reachable_clusters = [cluster for cluster in clusters
-                              if cluster.load != self.UNREACHABLE_CLUSTER_LOAD]
+        clusters = []
+        while all_clusters and not clusters:
+            for cluster in all_clusters:
+                if retry_info.in_retry_wait(cluster):
+                    continue
+                try:
+                    clusters.append(self.get_cluster_info(cluster, platform))
+                except BuildCanceledException:
```
lcarva: Unnecessary except block.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -231,40 +272,38 @@ def get_cluster_info(self, cluster, platform):
...
+                    raise
+                except OsbsException:
+                    if not retry_info.try_again_later(cluster, self.find_cluster_retry_delay):
+                        all_clusters.remove(cluster)
```
lcarva: Isn't this going to cause issues because you're altering the list being iterated over?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -231,40 +272,38 @@ def get_cluster_info(self, cluster, platform):
...
+                if not clusters:
+                    retry_info.wait_for_any()
 
-        if not reachable_clusters:
-            raise RuntimeError('All clusters for platform {} are unreachable!'
-                               .format(platform))
+        clusters = [cluster for cluster in clusters
```
lcarva: What if it gets here and there are no `clusters`? For instance, all the clusters have been removed from `all_clusters`, but none have been added to `clusters`.

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: Why not just pass the information to this class's constructor?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -389,26 +429,42 @@ def do_worker_build(self, release, cluster_info, koji_upload_dir, task_id):
...
+                raise RuntimeError('Could not find appropriate cluster for worker build.')
+            possible_clusters = self.get_clusters(clusters, platform, retry_info)
+            for cluster in possible_clusters:
+                try:
+                    self.do_worker_build(self.release, cluster, self.koji_upload_dir,
+                                         self.fs_task_id)
+                    return
+                except BuildCanceledException:
+                    raise
+                except Exception:
```
lcarva: Would it be safer to only retry on `OsbsException`?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -389,26 +429,42 @@ def do_worker_build(self, release, cluster_info, koji_upload_dir, task_id):
...
+        result = thread_pool.map_async(self.select_and_start_cluster, platforms)
 
         try:
-            while not result.ready():
-                # The wait call is a blocking call which prevents signals
-                # from being processed. Wait for short intervals instead
-                # of a single long interval, so build cancellation can
-                # be handled virtually immediately.
-                result.wait(1)
+            result.get()
```
lcarva: Based on the removed comment, will `get` get interrupted if there's an exception in the main thread?
I know it will if there's an exception in the worker thread.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -389,26 +429,42 @@ def do_worker_build(self, release, cluster_info, koji_upload_dir, task_id):
...
         # Always clean up worker builds on any error to avoid
         # runaway worker builds (includes orchestrator build cancellation)
         except Exception:
```
lcarva: The handling in this block is no longer correct.
Previously, `Exception` really meant `BuildCanceledException`. 
This is no longer true with the usage of `result.get` since it'll re-raise any exception in the worker threads.

Also, I'm not sure if something happens in a worker thread, the remaining worker builds will carry on as it did previously.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -189,13 +195,47 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
 
         if worker_build_image:
             self.log.warning('worker_build_image is deprecated, use config_kwargs instead')
             self.config_kwargs.setdefault('build_image', worker_build_image)
 
         self.worker_builds = []
 
+    class ClusterRetryContexts(object):
+        def __init__(self, clusters, max_cluster_fails):
+            self.clusters = list(clusters)
```
lcarva: This is unused.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -189,13 +195,47 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+            self.retry_at = {}
```
lcarva: Can we document the intent of these 4 variables? (for instance, the names `failed` vs `fails` doesn't help understanding)

atomic_reactor/plugins/build_orchestrate_build.py
bfontecc007: That's how I had it but I just changed it back lol

atomic_reactor/plugins/build_orchestrate_build.py
bfontecc007: Tim had me change it to Exception

atomic_reactor/plugins/build_orchestrate_build.py
bfontecc007: wait_for_any now raises an exception if they all fail. It should never get to that line without clusters now.

atomic_reactor/plugins/build_orchestrate_build.py
bfontecc007: This is all the documentation says: "If the remote call raised an exception then that exception will be reraised by get()."

atomic_reactor/plugins/build_orchestrate_build.py
bfontecc007: I added a terminate() statement to address this

atomic_reactor/plugins/build_orchestrate_build.py
twaugh: here: https://github.com/projectatomic/atomic-reactor/pull/841#discussion_r140527626

I'd be fine with changing this to OsbsException instead (and removing the BuildCanceledException clause) if everyone's happy with that and agrees that do_worker_build() shouldn't ever raise anything else.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -64,6 +68,47 @@ def override_build_kwarg(workflow, k, v):
...
+    """ No clusters could be found for a platform """
+
+
+class AllClustersFailedException(Exception):
+    """ Each cluster has reached max_cluster_fails """
+
+
+class ClusterRetryContexts(object):
+    def __init__(self, clusters, max_cluster_fails, find_cluster_retry_delay):
+        self.clusters = set(clusters)
```
twaugh: I originally put a self.clusters here thinking it would be useful to store the cluster info inside this class. If we're not using it that way though, let's just remove that attribute.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -64,6 +68,47 @@ def override_build_kwarg(workflow, k, v):
...
+        self.retry_at = {}
+        self.failed = set()
+        self.fails = Counter()
+        self.max_cluster_fails = max_cluster_fails
+        self.delay = find_cluster_retry_delay
```
twaugh: We had comments about some of these attributes in a previous iteration. Can we bring those back please?
https://github.com/projectatomic/atomic-reactor/pull/841/commits/bfc011d0635d8a17b1d7ad0763d5f4924a20b8c4

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -229,42 +282,35 @@ def get_cluster_info(self, cluster, platform):
...
-            current_builds = self.get_current_builds(osbs)
-        except OsbsException as e:
-            # If the build is canceled reraise the error
-            if isinstance(e.cause, BuildCanceledException):
-                raise e
 
-            self.log.exception("Error occurred while listing builds on %s",
-                               cluster.name)
-            return ClusterInfo(cluster, platform, osbs, self.UNREACHABLE_CLUSTER_LOAD)
+        current_builds = self.get_current_builds(osbs)
```
twaugh: Does this want to be `with osbs.retries_disabled()`?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -229,42 +282,35 @@ def get_cluster_info(self, cluster, platform):
...
+            for cluster in all_clusters:
+                if retry_info.in_retry_wait(cluster):
+                    continue
+                try:
+                    clusters.append(self.get_cluster_info(cluster, platform))
+                except OsbsException:
+                    if not retry_info.try_again_later(cluster):
+                        all_clusters.remove(cluster)
+                if not clusters:
+                    retry_info.wait_for_any()
```
twaugh: `if not clusters` is inside the loop that builds the clusters list. That looks wrong. Shouldn't it be like this?:
```
while all_clusters and not clusters:
    for cluster in all_clusters:
        ...

    if not clusters:
        retry_info.wait_for_any()
```

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -389,36 +436,57 @@ def do_worker_build(self, release, cluster_info, koji_upload_dir, task_id):
...
+            raise UnknownPlatformException('No clusters found for platform {}!'
+                                           .format(platform))
+
+        possible_clusters = None
+        retry_info = ClusterRetryContexts(clusters, self.max_cluster_fails,
+                                          self.find_cluster_retry_delay)
+
+        while True:
+            if not clusters:
+                raise RuntimeError('Could not find appropriate cluster for worker build.')
```
twaugh: Should this be `AllClustersFailedException`?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -389,36 +436,57 @@ def do_worker_build(self, release, cluster_info, koji_upload_dir, task_id):
...
+            possible_clusters = self.get_clusters(clusters, platform, retry_info)
```
twaugh: Maybe a comment here to explain that self.get_clusters() may actually modify `clusters` here?

atomic_reactor/plugins/build_orchestrate_build.py
twaugh: Oh, I see, it's used in `wait_for_any()` just to see if we've run out of clusters to try.

I wonder if, instead, we should change the caller of this method (i.e. `get_clusters`) so it raises an exception if it knows there are no clusters left to try -- it already knows this, without having to calculate it.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -64,6 +68,47 @@ def override_build_kwarg(workflow, k, v):
...
+            self.failed.add(name)
+        else:
+            self.retry_at[name] = (datetime.datetime.now() +
+                                   timedelta(seconds=self.delay))
+
+        return name not in self.failed
+
+    def wait_for_any(self):
+        if not len(self.clusters - self.failed):
+            raise AllClustersFailedException
```
twaugh: I don't really like this. I'd prefer `get_clusters` to have something like:
```
while all_clusters and not clusters:
    for cluster in all_clusters:
        ...

    if all_clusters and not clusters:
        retry_info.wait_for_any()

clusters = [filter out unreachable-load ones]
if not clusters:
    raise AllClustersFailedException
...
```

atomic_reactor/plugins/build_orchestrate_build.py
bfontecc007: self.get_current_builds has retries disabled L273

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -229,42 +285,42 @@ def get_cluster_info(self, cluster, platform):
...
+        possible_clusters = set()
+        candidates = copy.copy(retry_info.clusters)
+        while candidates and not possible_clusters:
+            for cluster in candidates:
+                if retry_info.in_retry_wait(cluster):
+                    continue
+                try:
+                    cluster_info = self.get_cluster_info(cluster, platform)
+                    possible_cluster_info[cluster.name] = cluster_info
+                    if cluster_info.load >= self.UNREACHABLE_CLUSTER_LOAD:
```
twaugh: This is wrong, it's an object not an int.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -229,42 +301,44 @@ def get_cluster_info(self, cluster, platform):
...
+    def get_clusters(self, platform, retry_contexts, all_clusters):
+        ''' return clusters sorted by load. '''
+
+        possible_cluster_info = {}
+        possible_clusters = set()
+        candidates = set(copy.copy(all_clusters))
+        while candidates and not possible_clusters:
+            for cluster in sorted(candidates, key=attrgetter('priority')):
+                ctx = retry_contexts[cluster.name]
+                if ctx.in_retry_wait:
```
twaugh: Just spotted that we also want to skip those that are `ctx.failed`.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -229,42 +301,44 @@ def get_cluster_info(self, cluster, platform):
...
+                    #     ctx.try_again_later(self.find_cluster_retry_delay)
+                    else:
+                        possible_clusters.add(cluster)
+                except OsbsException:
+                    ctx.try_again_later(self.failure_retry_delay)
+            candidates -= set([c for c in candidates if retry_contexts[c.name].failed])
+            candidates -= possible_clusters
+            # if we have ready clusters, let's use them, if not check the retries
+            if candidates and not possible_clusters:
+                wait_for_any_cluster(retry_contexts) # only if candidate?
```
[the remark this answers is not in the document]
twaugh: This is fine as it is, but a suggestion for readability is to call this unconditionally at the start of the loop. If any cluster is not in retry-wait it will return immediately.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -229,42 +301,44 @@ def get_cluster_info(self, cluster, platform):
...
+                ctx = retry_contexts[cluster.name]
+                if ctx.in_retry_wait:
+                    continue
+                try:
+                    cluster_info = self.get_cluster_info(cluster, platform)
+                    possible_cluster_info[cluster.name] = cluster_info
+                    # TODO: this is a problem area
+                    # if cluster_info.load >= 1:
+                    #     ctx.try_again_later(self.find_cluster_retry_delay)
+                    else:
```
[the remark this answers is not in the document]
twaugh: I'm fine with whatever @lcarva suggests here. If just sorting by load, and not checking if the load is too high, is the better plan then let's do that.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -35,6 +39,9 @@
 WORKSPACE_KEY_BUILD_INFO = 'build_info'
 WORKSPACE_KEY_UPLOAD_DIR = 'koji_upload_dir'
 WORKSPACE_KEY_OVERRIDE_KWARGS = 'override_kwargs'
+FIND_CLUSTER_RETRY_DELAY = 1.0
+FAILURE_RETRY_DELAY = 1.0
+MAX_CLUSTER_FAILS = 2
```
twaugh: Is this value realistic? I'd suggest 15, so that `MAX_CLUSTER_FAILS * FAILURE_RETRY_DELAY` takes us to 15 minutes. @lcarva, agree?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -35,6 +39,9 @@
 WORKSPACE_KEY_BUILD_INFO = 'build_info'
 WORKSPACE_KEY_UPLOAD_DIR = 'koji_upload_dir'
 WORKSPACE_KEY_OVERRIDE_KWARGS = 'override_kwargs'
+FIND_CLUSTER_RETRY_DELAY = 1.0
```
twaugh: This is now redundant so let's remove it.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -171,7 +233,10 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
 
     def __init__(self, tasker, workflow, platforms, build_kwargs,
                  osbs_client_config=None, worker_build_image=None,
-                 config_kwargs=None):
+                 config_kwargs=None,
+                 find_cluster_retry_delay=FIND_CLUSTER_RETRY_DELAY,
```
twaugh: Let's remove this redundant parameter.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -171,7 +233,10 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
...
+                 failure_retry_delay=FAILURE_RETRY_DELAY,
+                 max_cluster_fails=MAX_CLUSTER_FAILS):
```
twaugh: Let's include descriptions of these two in the docstring below.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -189,6 +254,12 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
         self.build_kwargs = build_kwargs
         self.osbs_client_config = osbs_client_config
         self.config_kwargs = config_kwargs or {}
+        self.find_cluster_retry_delay = find_cluster_retry_delay
```
twaugh: This is redundant and can be removed.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -229,42 +301,43 @@ def get_cluster_info(self, cluster, platform):
...
+            for cluster in sorted(candidates, key=attrgetter('priority')):
+                ctx = retry_contexts[cluster.name]
+                if ctx.in_retry_wait:
+                    continue
+                try:
+                    cluster_info = self.get_cluster_info(cluster, platform)
+                    possible_cluster_info[cluster.name] = cluster_info
+                    # TODO: this is a problem area
+                    # if cluster_info.load >= 1:
+                    #     ctx.try_again_later(self.find_cluster_retry_delay)
```
twaugh: Let's remove this now we've decided not to go that way.

tests/plugins/test_orchestrate_build.py
```diff
@@ -391,6 +393,84 @@ def mock_wait_for_build_to_finish(build_name):
...
+    runner = BuildStepPluginsRunner(
+        workflow.builder.tasker,
+        workflow,
+        [{
+            'name': OrchestrateBuildPlugin.key,
+            'args': {
+                'platforms': ['x86_64', 'ppc64le'],
+                'build_kwargs': make_worker_build_kwargs(),
+                'osbs_client_config': str(tmpdir),
+                'find_cluster_retry_delay': .1,
```
twaugh: Let's remove this redundant parameter.

tests/plugins/test_orchestrate_build.py
```diff
@@ -324,6 +324,8 @@ def mock_wait_for_build_to_finish(build_name):
                 'platforms': ['x86_64', 'ppc64le'],
                 'build_kwargs': make_worker_build_kwargs(),
                 'osbs_client_config': str(tmpdir),
+                'max_cluster_fails': 2,
+                'unreachable_cluster_retry_delay': 1
```
twaugh: Do we want this to be 0.1 in the tests, to make them quicker?

tests/plugins/test_orchestrate_build.py
```diff
@@ -391,6 +393,84 @@ def mock_wait_for_build_to_finish(build_name):
...
+    runner = BuildStepPluginsRunner(
+        workflow.builder.tasker,
+        workflow,
+        [{
+            'name': OrchestrateBuildPlugin.key,
+            'args': {
+                'platforms': ['x86_64', 'ppc64le'],
+                'build_kwargs': make_worker_build_kwargs(),
+                'osbs_client_config': str(tmpdir),
+                'find_cluster_retry_delay': .1,
```
twaugh: Let's remove this redundant parameter.

tests/plugins/test_orchestrate_build.py
```diff
@@ -431,19 +511,19 @@ class RaiseOnce(object):
...
-    (flexmock(AsyncResult).should_receive('wait')
-        .replace_with(raise_once.wait))
+    (flexmock(AsyncResult).should_receive('get')
+        .replace_with(raise_once.get))
 
     # Required due to python3 implementation.
-    (flexmock(AsyncResult).should_receive('get')
-        .and_return(None))
+    # (flexmock(AsyncResult).should_receive('get')
+    #     .and_return(None))
```
twaugh: If this isn't needed let's remove it.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -35,6 +39,9 @@
 WORKSPACE_KEY_BUILD_INFO = 'build_info'
 WORKSPACE_KEY_UPLOAD_DIR = 'koji_upload_dir'
 WORKSPACE_KEY_OVERRIDE_KWARGS = 'override_kwargs'
+FIND_CLUSTER_RETRY_DELAY = 1.0
+FAILURE_RETRY_DELAY = 1.0
```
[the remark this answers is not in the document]
bfontecc007: This was just temporary for testing.

atomic_reactor/plugins/build_orchestrate_build.py
bfontecc007: This is different than FAILURE_RETRY_DELAY (or it will be). I just changed line 333 to use it instead of the failure delay.

twaugh: So the delay when attempting to list builds fails is different to the delay when attempting to create a build fails? Not sure I understand why.

atomic_reactor/plugins/build_orchestrate_build.py
bfontecc007: done

atomic_reactor/plugins/build_orchestrate_build.py
bfontecc007: Right now it's designed to be configurable because we haven't figured out exactly how we want it to work in production. We can always just set them to the same value if that's the desired behavior.

lcarva: I'm fine either way.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -389,36 +462,55 @@ def do_worker_build(self, release, cluster_info, koji_upload_dir, task_id):
...
+        }
+
+        while True:
+            possible_cluster_info = self.get_clusters(platform, retry_contexts, clusters)
+
+            for cluster_info in possible_cluster_info:
+                ctx = retry_contexts[cluster_info.cluster.name]
+                try:
+                    self.do_worker_build(cluster_info)
+                    return
```
twaugh: Looking at `do_worker_build()`, I don't think it will ever raise `OsbsException`. If creating the build fails due to `OsbsResponseException` it looks like it will just create a `WorkerBuildInfo` with build=None.

Doesn't the `return` here assume that `do_worker_build()` succeeded in creating a build?

Do we have a unit test to verify this situation is handled correctly?

lcarva: Ha! Good catch. Let's change `do_worker_build` to not catch exceptions when `cluster_info.osbs.create_worker_build(**kwargs)` is called. This should cause the exception to propagate as needed.
What do you think?

bfontecc007:
I’m just on my phone but I’m guessing that having do_worker_build raise
OsbsException would make the most sense.

atomic_reactor/plugins/build_orchestrate_build.py
bfontecc007: Should do_worker_build no longer try and catch? Or should it just reraise OsbsException and handle all others in its usual way?

twaugh: I think we want to have do_worker_build()
* always append to self.worker_builds, as it does now
* raise any exception that occurs when starting a build or monitoring it

So where it currently catches `Exception`, I think it should note what the exception is and arrange for `build_info.monitor_exception` to be set to it. This may mean having WorkerBuildInfo's constructor not take 'build' but instead just set it to None, and have do_worker_build() setting build_info.build when it's known.

Then, when monitoring the build, it should catch the exception as before and set build_info.monitor_exception to it. It should also try to cancel the build, as it currently does. But at the end of all that, it should raise that exception.

@lcarva, what do you think?

twaugh: We discussed this and came to the conclusion that catching and re-raising OsbsException is the better approach.

lcarva:
Updated release notes draft:
> - Retry approach has been applied to interaction with worker clusters

PR #842, opened 2017-09-22
Import image ignore version release

lcarva:
This PR has two commit. The first one has functional changes.
The second is a cleanup of the code with no functional changes.

I figured now it's a good time to do this code cleanup since this
plugin needs to be tested again due to recent changes (including
the 1st commit in this PR).


lcarva:
> s it verified by not having a flexmocked ensure_image_stream_tag with args? If yes, could we add a comment about it?

Precisely. Added comment explaining.

lcarva:
Updated release notes draft:
> - `import_image` does not explicitly track version-release tags

PR #843, opened 2017-09-25
Flatpak: export a correct list of rpms to Koji

owtaylor:
This PR makes the list of RPMs that make up the archive that is sent to Koji correct for Flatpaks - previously it was the list of RPMs for the "base image" that is used to install RPMs into `/var/tmp/flatpak-build`.

Flatpak-specific checks and post-processing on the `rpm -qa` output are needed, so rather than extending the rpmqa plugin to be able to run `rpm -qa` on `/var/lib/flatpak-build`, logic is added into the `flatpak_create_oci plugin` instead, and the rpm list is moved to `workflow.image_components`, so different plugins can fill it in.


atomic_reactor/plugins/prepub_flatpak_create_oci.py
```diff
@@ -247,6 +254,108 @@ def _export_filesystem(self):
...
+        modules = self.source.compose.modules
+        base_module = self.source.compose.base_module
+
+        # Identify the module for the Flatpak runtime that this app runs against
+        runtime_module = None
+        for key in base_module.mmd.buildrequires.keys():
+            try:
+                module = modules[key]
+                if 'runtime' in module.mmd.profiles:
+                    runtime_module = module
```
twaugh: Should we `break` here?

atomic_reactor/rpm_util.py
```diff
@@ -0,0 +1,89 @@
...
+]
+
+
+def rpm_qf_args(tags=image_component_rpm_tags, separator=';'):
+    """
+    Return the arguments to pass to rpm to list RPMs in the format expected
+    by parse_rpm_output.
+    """
+
+    fmt = separator.join(["%%{%s}" % tag for tag in image_component_rpm_tags])
```
twaugh: Should this use `tags`?

atomic_reactor/rpm_util.py
```diff
@@ -0,0 +1,89 @@
...
+    'EPOCH',
+    'SIZE',
+    'SIGMD5',
+    'BUILDTIME',
+    'SIGPGP:pgpsig',
+    'SIGGPG:pgpsig',
+]
+
+
+def rpm_qf_args(tags=image_component_rpm_tags, separator=';'):
```
twaugh: Landscape.io complains that tags takes a dangerous default value because it is mutable. We should either default to `None` and handle the real default in the body, or add a comment to stop that warning.

atomic_reactor/rpm_util.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    Return the arguments to pass to rpm to list RPMs in the format expected
+    by parse_rpm_output.
+    """
+
+    fmt = separator.join(["%%{%s}" % tag for tag in image_component_rpm_tags])
+    return r"-qa --qf '{0}\n'".format(fmt)
+
+
+def parse_rpm_output(output, tags=image_component_rpm_tags, separator=';'):
```
twaugh: Landscape.io complains that tags takes a dangerous default value because it is mutable.

atomic_reactor/plugins/post_rpmqa.py
```diff
@@ -43,10 +32,13 @@ def __init__(self, tasker, workflow, image_id, ignore_autogenerated_gpg_keys=Tru
         self.ignore_autogenerated_gpg_keys = ignore_autogenerated_gpg_keys
 
     def run(self):
-        fmt = self.sep.join(["%%{%s}" % tag for tag in self.rpm_tags])
+        # If another component has already filled in the image component list, skip
+        if self.workflow.image_components is not None:
+            return None
```
twaugh: This line doesn't seem to get test coverage.

atomic_reactor/plugins/prepub_flatpak_create_oci.py
```diff
@@ -247,6 +254,108 @@ def _export_filesystem(self):
...
+            except KeyError:
+                pass
+
+        if runtime_module is None:
+            raise RuntimeError("Failed to identify runtime module in the buildrequires for {}"
+                               .format(base_module.name))
```
twaugh: This outcome isn't covered in unit tests.

atomic_reactor/plugins/prepub_flatpak_create_oci.py
```diff
@@ -247,6 +254,108 @@ def _export_filesystem(self):
...
+
+            # If it's not from the runtime, check that the specific package
+            # version was built in one of the app modules
+            if component['epoch'] is not None:
+                component_filename = \
+                    "{name}-{epoch}:{version}-{release}.{arch}.rpm".format(**component)
+            else:
+                # The PDC data has the misapprehension that epoch 0 is the same as no epoch
+                component_filename = \
+                    "{name}-0:{version}-{release}.{arch}.rpm".format(**component)
```
twaugh: This line isn't covered in unit tests.

atomic_reactor/plugins/prepub_flatpak_create_oci.py
owtaylor: Really, really precise handling would be to check that only one of the build dependencies is a runtime module, but I'm not expecting people to make that mistake in practice, so adding another error path seems unnecessary. First is  more self-explanatory than last, so I'll add the break.

atomic_reactor/plugins/prepub_flatpak_create_oci.py
owtaylor: Fixed.

atomic_reactor/plugins/prepub_flatpak_create_oci.py
owtaylor: Fixed.

atomic_reactor/rpm_util.py
owtaylor: Went with the None default, since it will read transparently and a comment would require more thinking on the part of a reader.

atomic_reactor/rpm_util.py
owtaylor: Good catch! this happened not to break anything because the tags list that koji_upload/koji_promote use are *smaller* than the default. No function is too small to unit test, apparently.

atomic_reactor/plugins/post_rpmqa.py
owtaylor: Fixed.

owtaylor:
OK, after some adventures in git rebasing, I think this covers all the requested changes.

PR #844, opened 2017-09-25
Export appropriate metadata to Koji for flatpaks

owtaylor:
In the extra data we send to Koji for a Flatpak build, include:

```    
'flatpak': True,
'source_modules': [<specified module for the build>],
'modules': [<resolved list of module name-stream-version>]
```
    
This is sufficient information to know when a rebuild needs to be triggered and to start a rebuild.

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_module_compose.py
+++ b/atomic_reactor/plugins/pre_resolve_module_compose.py
@@ -43,11 +43,21 @@ class ModuleInfo(object):
 
 
 class ComposeInfo(object):
-    def __init__(self, compose_id, base_module, modules, repo_url):
+    def __init__(self, source_spec, compose_id, base_module, modules, repo_url):
+        self.source_spec = source_spec
         self.compose_id = compose_id
         self.base_module = base_module
         self.modules = modules
         self.repo_url = repo_url
+
+    def koji_metadata(self):
+        sorted_modules = [self.modules[k] for k in sorted(self.modules.keys())]
+
+        return {
+            'source_modules': [self.source_spec],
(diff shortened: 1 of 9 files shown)


atomic_reactor/plugins/pre_resolve_module_compose.py
```diff
@@ -206,13 +216,14 @@ def _resolve_compose(self):
         # makes things simpler.
         pdc_client = PDCClient(server=self.pdc_url, ssl_verify=not self.pdc_insecure, develop=True)
 
-        if self.compose_id is None:
-            if self.module_version is not None:
-                source = self.module_name + '-' + self.module_stream + '-' + self.module_version
-            else:
-                source = self.module_name + '-' + self.module_stream
+        if self.module_version is not None:
+            source_spec = self.module_name + '-' + self.module_stream + '-' + self.module_version
```
vrutkovs: What about:
```
source_spec = "{0}-{1}".format(self.module_name, self.module_stream)
if self.module_version:
    source_spec = "{0}-{1}".format(source_spec, self.module_version)
```

This would avoid concatenation and convert into strings properly

owtaylor:
I think string concatenation is pretty optimized, probably as fast as getting the format machinery involved for this, and the arguments are supposed to be strings, but sure, I've switched it to:

```
fmt = '{n}-{s}' if self.module_version is None else '{n}-{s}-{v}'
source_spec = fmt.format(n=self.module_name, s=self.module_stream, v=self.module_version)
```

(optimizing for readability.)

Also added information to koji.md.

vrutkovs:
Needs a rebase to resolve conflicts

owtaylor:
Rebased

PR #845, opened 2017-09-25
Give OCI images uploaded to Koji an appropriate filename

owtaylor:
Instead of docker-image-{id}.tar.gz, call uploaded OCI images oci-image-{id}.tar.gz.

This involves moving the logic of creating the filename into a new utility function util.get_image_upload_filename()


owtaylor:
Rebased to fix merge conflict.

PR #846, opened 2017-09-25
post_group_manifests: add support for OCI Manifests and Indices

owtaylor:
This PR adds support for OCI Manifests and Indices to post_group_manifests. The introductory commits fix some bugs in HTTP communication with the registry, and abstracts the registry specifics to a RegistrySession() wrapper around requests.Session(). Then the final patch restructures post_group_manifests to drop the use of manifest-tool and create and post manifests directly. The group=true and group=false cases are more unified and a single parameterized test tests them both.

A lot of the complexity (and the need to restructure) is from handling the case where the manifests that are being grouped and the resulting manifest lists aren't all in the same repository. This is possible in the local view of 'group manifests' - it has as input a list of repository:tag combinations from the workers, and a list of repository:tag combinations that it needs to apply to the resulting output, but after the fact I examined things thoroughly and I don't think it can actually happen with the current atomic-reactor code. (Previously, it superficially looked like the plugin handled multiple such multiple-repository situations, I think the code would have likely worked for group=true, assuming that manifest-tool did the right thing, but would not have worked for group=false.)

Things could be simplified by removing this code and simply asserting if the inputs and outputs aren't in the same repository. On the other, hand the unit tests cover this code pretty well, and it might be useful to have this logic in the future.

### DIFF
--- a/atomic_reactor/plugins/post_group_manifests.py
+++ b/atomic_reactor/plugins/post_group_manifests.py
@@ -12,13 +12,22 @@ it for all existing image tags.
 
 
 from __future__ import unicode_literals
-from tempfile import NamedTemporaryFile
-import yaml
-from subprocess import check_output, CalledProcessError, STDOUT
+import json
+import requests
 
 from atomic_reactor.plugin import PostBuildPlugin, PluginFailedException
-from atomic_reactor.util import RegistrySession, registry_hostname, get_manifest_digests
-from atomic_reactor.constants import PLUGIN_GROUP_MANIFESTS_KEY, MEDIA_TYPE_DOCKER_V2_SCHEMA2
+from atomic_reactor.util import RegistrySession, registry_hostname, ManifestDigest
+from atomic_reactor.constants import (PLUGIN_GROUP_MANIFESTS_KEY, MEDIA_TYPE_DOCKER_V2_SCHEMA2,
+                                      MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST, MEDIA_TYPE_OCI_V1,
+                                      MEDIA_TYPE_OCI_V1_INDEX)
+
+
+# The plugin requires that the worker builds have already pushed their images into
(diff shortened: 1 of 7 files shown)


vrutkovs:
LGTM, but one test for F25 with py3 fails due to encoding issues

owtaylor:
Added a fix for `json.loads(<bytes>)` not working with older Python 3

twaugh:
Tests still fail for Fedora 25 + Python 3.

owtaylor:
OK, seem to have gotten *all* the encoding errors now.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -47,144 +59,298 @@ def __init__(self, tasker, workflow, registries, group=True, goarch=None):
...
+            # A 202-Accepted would mean that the source blob didn't exist and
+            # we're starting an upload - but we've checked that above
+            raise RuntimeError("Blob mount had unexpected status {}".format(result.status_code))
+
+    def link_manifest_references_into_repository(self, session, manifest, media_type,
+                                                 source_repo, target_repo):
+        """
+        Links all the blobs referenced by the manifest from source_repo into target_repo. If
+        manifest_digest is passed, also links the manifest itself. (As a *blob* not as
+        a manifest... to use it directly as a manifest, it needs to be uploaded with a tag.)
```
twaugh: Not sure I understand the manifest_digest remark.

twaugh:
There are merge conflicts, likely due to merging https://github.com/projectatomic/atomic-reactor/pull/857.

atomic_reactor/plugins/post_group_manifests.py
owtaylor: Sorry, a left-over from older versions of the code. Removed in the latest push.

owtaylor:
Stray comment fixed and rebased, fixing conflicts.

PR #847, opened 2017-09-25
flatpak: fix errors if exit_store_metadata_in_osv3 runs on early failure

owtaylor:
If the build fails before the Dockerfile has been created, then `exit_store_metadata_in_osv3` was failing .

Fix problems if `build.base_image` and `build._df_path` are not set.

### DIFF
--- a/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
+++ b/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
@@ -193,13 +193,25 @@ class StoreMetadataInOSv3Plugin(ExitPlugin):
         except AttributeError:
             commit_id = ""
 
-        try:
-            base_image_id = self.workflow.base_image_inspect['Id']
-        except KeyError:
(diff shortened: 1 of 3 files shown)


atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -193,13 +193,24 @@ def run(self):
...
+            try:
+                base_image_id = self.workflow.base_image_inspect['Id']
+            except KeyError:
+                base_image_id = ""
+        else:
+            base_image_name = ""
             base_image_id = ""
 
+        try:
+            dockerfile_contents = open(self.workflow.builder.df_path).read()
```
twaugh: Should we use `with open(...) as ...` here?

owtaylor: Existing code, just moved around, but sure, I'll squeeze a fix for that in. (Will just include the fix in the patch, rather splitting it out, I think.)

PR #848, opened 2017-09-26
Move flatpak deps in a separate file

vrutkovs:
This would simplify CentOS/RHEL buildroot

### DIFF
--- a/requirements-flatpak.txt
+++ b/requirements-flatpak.txt
@@ -1,0 +1,4 @@
+modulemd
+pdc-client
+# https://pagure.io/modulemd/pull-request/46
+python-dateutil
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,8 +2,4 @@ docker-py
 docker-squash>=1.0.0rc3
 dockerfile-parse>=0.0.5
 jsonschema
-modulemd
-pdc-client
-# https://pagure.io/modulemd/pull-request/46
-python-dateutil
 PyYAML
--- a/test.sh
+++ b/test.sh
@@ -67,6 +67,7 @@ if [[ $PYTHON_VERSION == 2* ]]; then
   $RUN $PIP install git+https://github.com/release-engineering/dockpulp
   $RUN $PIP install -r requirements-py2.txt
 fi
+$RUN $PIP install -r requirements-flatpak.txt
 $RUN $PIP install docker-squash
 $RUN $PYTHON setup.py install
 


tests/constants.py
```diff
@@ -24,7 +24,7 @@
 DOCKERFILE_ERROR_BUILD_PATH = os.path.join(FILES, 'docker-hello-world-error-build')
 DOCKERFILE_SUBDIR_PATH = os.path.join(FILES, 'df-in-subdir')
 
-FLATPAK_GIT = "git://pkgs.fedoraproject.org/modules/eog.git"
+FLATPAK_GIT = "http://pkgs.fedoraproject.org/modules/eog.git"
```
owtaylor:
```
$ git clone "git://pkgs.fedoraproject.org/modules/eog.git"
Cloning into 'eog'...
remote: Counting objects: 25, done.
remote: Compressing objects: 100% (24/24), done.
remote: Total 25 (delta 8), reused 0 (delta 0)
Receiving objects: 100% (25/25), done.
Resolving deltas: 100% (8/8), done.
```

Maybe there were transient problems - the service wasn't running? If we're going to change this, probably better to use: `https://src.fedoraproject.org/modules/eog.git` to get HTTPS. (Compare what https://src.fedoraproject.org/modules/eog shows for GIT URLs.) In any case, the commit subject doesn't match the commit.

atomic_reactor/plugins/pre_resolve_module_compose.py
```diff
@@ -25,8 +25,11 @@
 
 import os
 import re
-from modulemd import ModuleMetadata
-from pdc_client import PDCClient
+try:
+    from modulemd import ModuleMetadata
+    from pdc_client import PDCClient
```
owtaylor: is this just to avoid ` logger.warning("can't load module '%s': %r", f, ex)` in plugin.py or does it cause a failure somewhere? I'm basically OK with this, though it's a bit ugly. This approach does seem to be inconsistent with the handling of dockpulp, where the warning is left.

tests/constants.py
vrutkovs: Correct, I had some problems with Fedora Infra yesterday - I heard `git://` was disabled for outside hosts. Now it works fine, reverted that

atomic_reactor/plugins/pre_resolve_module_compose.py
vrutkovs: Agree, this change is not needed, removed

lcarva:
@rcerven, can you cherry-pick this into latest release branch.

lcarva:
Updated release notes draft:
> - flatpak dependencies are now in a separate file

PR #849, opened 2017-09-26
koji_promote adds media_type to build metadata fix

rcerven:

in prod build, pulp_pull runs as postbuild plugin and not as exit plugin in orchestrator

### DIFF
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -566,7 +566,7 @@ class KojiPromotePlugin(ExitPlugin):
                     extra['filesystem_koji_task_id'] = task_id
 
         # Append media_types from pulp pull
-        pulp_pull_results = self.workflow.exit_results.get(PLUGIN_PULP_PULL_KEY)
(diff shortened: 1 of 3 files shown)


atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -566,7 +566,7 @@ def get_build(self, metadata):
                     extra['filesystem_koji_task_id'] = task_id
 
         # Append media_types from pulp pull
-        pulp_pull_results = self.workflow.exit_results.get(PLUGIN_PULP_PULL_KEY)
+        pulp_pull_results = self.workflow.postbuild_results.get(PLUGIN_PULP_PULL_KEY)
```
lcarva: We should attempt to grab either from postbuild_results or exist_results.

rcerven: koji_promote runs only in prod, and pulp_pull in prod runs only as postbuild

lcarva: ah right, ok

lcarva:
Test failures.

rcerven:
just fix for what is already in release note

PR #850, opened 2017-09-26
koji_promote adds koji_parent_id to build metadata additional metadata

rcerven:

### DIFF
--- a/tests/plugins/test_koji_promote.py
+++ b/tests/plugins/test_koji_promote.py
@@ -27,7 +27,8 @@ except ImportError:
     del koji
     import koji
 
-from atomic_reactor.constants import (PLUGIN_PULP_SYNC_KEY, PLUGIN_PULP_PULL_KEY)
+from atomic_reactor.constants import (PLUGIN_PULP_SYNC_KEY, PLUGIN_PULP_PULL_KEY,
+                                      PLUGIN_KOJI_PARENT_KEY)
 from atomic_reactor.core import DockerTasker
 from atomic_reactor.plugins.exit_koji_promote import (KojiUploadLogger,
                                                       KojiPromotePlugin)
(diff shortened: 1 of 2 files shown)


tests/plugins/test_koji_promote.py
```diff
@@ -1364,3 +1365,44 @@ def test_koji_promote_set_media_types(self, tmpdir, os_env, expect_result):
...
+    ])
+    def test_koji_promote_set_parent_id(self, tmpdir, os_env, parent_output, expect_result):
+        session = MockedClientSession('')
+        tasker, workflow = mock_environment(tmpdir,
+                                            name='ns/name',
+                                            version='1.0',
+                                            release='1',
+                                            session=session)
+
+        if parent_output:
```
vrutkovs: Needs a test when `koji_parent` didn't run at all

tests/plugins/test_koji_promote.py
```diff
@@ -1364,3 +1365,44 @@ def test_koji_promote_set_media_types(self, tmpdir, os_env, expect_result):
             assert sorted(image['media_types']) == sorted(expect_result)
         else:
             assert 'media_types' not in image.keys()
+
+    @pytest.mark.parametrize(('parent_output', 'expect_result'), [
```
vrutkovs: nitpick: `expected_result`?

rcerven: ok

tests/plugins/test_koji_promote.py
rcerven: I am mocking what is in workflow, and runner is running only koji_promote

so how such a test has value?

rcerven:
commit message changed

rcerven:
smoketests fail because of https://github.com/release-engineering/dockpulp/pull/147#issuecomment-332434655

tests/plugins/test_koji_promote.py
lcarva: Having a use case where `PLUGIN_KOJI_PARENT_KEY` key is not set at all in `prebuild_results` would cover that.

rcerven: added

vrutkovs:
Release notes updated

PR #851, opened 2017-09-27
Add isolated variable to metadata.json

bfontecc007:
This adds isolated: True|False to the build metadata.

### DIFF
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -361,6 +361,10 @@ class KojiImportPlugin(ExitPlugin):
             except ValueError:
                 self.log.error("invalid task ID %r", koji_task_id, exc_info=1)
 
+        isolated = metadata.get('labels', {}).get('isolated', False)
+        self.log.info("build is isolated: {}".format(isolated))
+        extra['image']['isolated'] = isolated
+
         fs_result = self.workflow.prebuild_results.get(AddFilesystemPlugin.key)
         if fs_result is not None:
             try:
--- a/tests/plugins/test_koji_import.py
+++ b/tests/plugins/test_koji_import.py
@@ -622,6 +622,43 @@ class TestKojiImport(object):
             runner.run()
         assert "plugin 'koji_import' raised an exception: RuntimeError" in str(exc)
 
+    @pytest.mark.parametrize(('isolated'), [
+        False,
+        True,
+        None
+    ])
+    def test_isolated_metadata_json(self, tmpdir, monkeypatch, os_env, isolated):
+        session = MockedClientSession('')
+        tasker, workflow = mock_environment(tmpdir,
+                                            session=session,
+                                            name='ns/name',
+                                            version='1.0',
+                                            release='1')
+        runner = create_runner(tasker, workflow)
+
+        patch = {
+            'metadata': {
+                'creationTimestamp': '2015-07-27T09:24:00Z',
+                'namespace': NAMESPACE,
+                'name': BUILD_ID,
+                'labels': {
+                },
+            }
+        }
+
+        if isolated is not None:
+            patch['metadata']['labels']['isolated'] = isolated
+
+        monkeypatch.setenv("BUILD", json.dumps(patch))
+
+        runner.run()
+
+        build_metadata = session.metadata['build']['extra']['image']['isolated']
+        if isolated:
+            assert build_metadata is True
+        else:
+            assert build_metadata is False
+
     @pytest.mark.parametrize(('koji_task_id', 'expect_success'), [
         (12345, True),
         ('x', False),


bfontecc007:
Added unit test

tests/plugins/test_koji_import.py
```diff
@@ -622,6 +622,46 @@ def test_koji_import_wrong_source_type(self, tmpdir, os_env):
             runner.run()
         assert "plugin 'koji_import' raised an exception: RuntimeError" in str(exc)
 
+    @pytest.mark.parametrize(('isolated'), [
```
lcarva: Maybe also handle `None`, in which case the `'isolated'` label is not set at all?

tests/plugins/test_koji_import.py
```diff
@@ -622,6 +622,46 @@ def test_koji_import_wrong_source_type(self, tmpdir, os_env):
...
+                'namespace': NAMESPACE,
+                'name': BUILD_ID,
+                'labels': {
+                    'isolated': isolated
+                },
+            }
+        }))
+
+        runner.run()
+        metadata = session.metadata
```
lcarva: Can't we just run:
```python
build_metadata = metadata['build']['extra']['image']['isolated']
assert build_metadata is isolated
```
Might be a little bit easier to read.

tests/plugins/test_koji_import.py
bfontecc007: If it's None, should we assume false?

lcarva: Yes. Similarly to `autorebuild`, it's `False` unless proven otherwise. But it's always there.

bfontecc007: added.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -361,6 +361,14 @@ def get_build(self, metadata, worker_metadatas):
             except ValueError:
                 self.log.error("invalid task ID %r", koji_task_id, exc_info=1)
 
+        isolated = metadata.get('labels', {}).get('isolated', False)
+        if isolated is not None:
```
vrutkovs: Not sure if `isolated` can ever be `None`, `get` would set it to `False`. Just `if isolated` should be sufficient. `try..catch` is not necessary as well

tests/plugins/test_koji_import.py
```diff
@@ -622,6 +622,39 @@ def test_koji_import_wrong_source_type(self, tmpdir, os_env):
...
+                                            release='1')
+        runner = create_runner(tasker, workflow)
+
+        monkeypatch.setenv("BUILD", json.dumps({
+            'metadata': {
+                'creationTimestamp': '2015-07-27T09:24:00Z',
+                'namespace': NAMESPACE,
+                'name': BUILD_ID,
+                'labels': {
+                    'isolated': isolated
```
vrutkovs: `isolated` should not be set if its `None`. Instead the test should set this value only if `isolated` is not `None`

tests/plugins/test_koji_import.py
```diff
@@ -622,6 +622,39 @@ def test_koji_import_wrong_source_type(self, tmpdir, os_env):
...
+                },
+            }
+        }))
+
+        runner.run()
+
+        build_metadata = session.metadata['build']['extra']['image']['isolated']
+        if isolated:
```
vrutkovs: `assert build_metadata is True if isolated else False` might also work

tests/plugins/test_koji_import.py
lcarva: Let's not set this key at all if `isolated` is `None`. That would test the case in which the label has not been set on the OpenShift build object.

atomic_reactor/plugins/exit_koji_import.py
bfontecc007: I'm not sure I agree on this. What if we get a bad value from OSBS?

vrutkovs: Good point. Maybe `.get()` is not necessary then:
```
try:
    isolated = str(metadata['labels']['isolated']).lower() == 'true'
except IndexError:
    isolated = False
```

`IndexError` would guard against missing keys and `AttributeError` would ensure the returned value (forcibly converted into string) is True. This way even a string would be converted to a boolean, although I'm not sure if its intended.

In any case `bool("False")` would not throw exception and evaluate into `True`.

bfontecc007: Does this cover cases the current code doesn't cover? If not, I'd advocate for leaving it as-is just because I feel that this is hard to read.

vrutkovs: It does - if `isolated` there is `foo` it would evaluate into `True` with current code.

Suggested code is a bit complex, yes, maybe @lcarva has ideas about improving it.

bfontecc007: Hmm, you're right about the 'foo' case. String 'False' would result in True too.

bfontecc007: A lot of unit tests don't set 'labels' so I added an exception for KeyError.

lcarva: We expect that value to be a boolean. If it's not, let's treat it as an error instead of trying to mangling it into a string...
IMO opinion the original version of this code was the easiest to follow.

vrutkovs: Okay, agree. `bool(isolated)` doesn't do anything useful though

bfontecc007:
OK I pushed something that is similar to the original, but in consideration of Vadim's feedback.

lcarva:
Updated release notes draft:
> - Metadata for Koji builds will indicate whether or not a build is isolated: `extra.image.isolated`

lcarva:
Aye just realized we didn't update `docs/koij.md`. @bfontecc007 could you take care of that?

twaugh:
What about koji_promote?

lcarva:
> What about koji_promote?

Hmm. It would always be `False` in our deployment. But yes, we should do it in there as well.

PR #853, opened 2017-09-28
flatpak_create_oci: Sanitize file permissions and ownership

owtaylor:
When we are exporting the filesystem, match the permission and
ownership changes done by 'flatpak build-export'. In addition to
replacing the existing fixup for read-only files, this handles other
cases that could cause problems, such as setuid binaries or
world-writable directories.

### DIFF
--- a/atomic_reactor/plugins/prepub_flatpak_create_oci.py
+++ b/atomic_reactor/plugins/prepub_flatpak_create_oci.py
@@ -210,10 +210,23 @@ class FlatpakCreateOciPlugin(PrePublishPlugin):
             if target_name is None:
                 continue
 
-            # flatpak tries to create files/.ref and behaves badly if the root directory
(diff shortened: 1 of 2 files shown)


vrutkovs:
Needs a rebase to resolve merge conflicts

vrutkovs:
That's weird - Travis build has started failing after the rebase

owtaylor:
Yeah, I made a mistake in the rebase, pushed a new version now. The problem was that the patch fixed a bug:

    - if config_name is 'app'
    + if config_name == 'app'

which *sometimes* failed to match, depending on the details of the python version, and I lost that fix with the rebase (it was in an area with conflicts.)

vrutkovs:
Ah, `is` is tricky indeed. I'll merge it after tests are complete

PR #854, opened 2017-09-29
Add isolated variable to metadata.json

bfontecc007:
This is a fix for the already merged PR #851 

Signed-off-by: Bret Fontecchio <bfontecc@redhat.com>

PR #855, opened 2017-10-02
Add isolated boolean to koji_promote metadata

bfontecc007:
This commit is similar to #854, but it adds the same logic to the koji_promote plugin.

### DIFF
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -582,6 +582,14 @@ class KojiPromotePlugin(ExitPlugin):
             else:
                 extra['image']['parent_build_id'] = parent_id
 
+        # Append isolated build flag
+        try:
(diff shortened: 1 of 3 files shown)


atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -582,6 +582,14 @@ def get_build(self, metadata):
             else:
                 extra['image']['parent_build_id'] = parent_id
 
+        # Append isolated build flag
+        try:
+            isolated = str(metadata['labels']['isolated']).lower() == 'true'
+        except (IndexError, AttributeError, KeyError):
+            isolated = False
+        self.log.info("build is isolated: {}".format(isolated))
```
twaugh: Would be nice to change both this and the koji_import instance of this line to avoid the landscape.io warning:
https://landscape.io/github/projectatomic/atomic-reactor/2109/modules/atomic_reactor/plugins/exit_koji_promote.py#L590

lcarva:

PR #857, opened 2017-10-04
Add digests-version map in koji image extras

vrutkovs:
TODO:
* [x] Update tests to cover all lines
* [x] koji-promote
* [x] update docs/koji.md

### DIFF
--- a/atomic_reactor/constants.py
+++ b/atomic_reactor/constants.py
@@ -101,3 +101,11 @@ HTTP_BACKOFF_FACTOR = 5
 HTTP_CLIENT_STATUS_RETRY = (408, 500, 502, 503, 504)
 # requests timeout in seconds
 HTTP_REQUEST_TIMEOUT = 600
+
+
+# Media types
+MEDIA_TYPE_DOCKER_V1 = "application/json"
+MEDIA_TYPE_DOCKER_V2_SCHEMA1 = "application/vnd.docker.distribution.manifest.v1+json"
+MEDIA_TYPE_DOCKER_V2_SCHEMA2 = "application/vnd.docker.distribution.manifest.v2+json"
+MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST = "application/vnd.docker.distribution.manifest.list.v2+json"
+MEDIA_TYPE_OCI_V1 = "application/vnd.oci.image.manifest.v1+json"
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -35,7 +35,9 @@ from atomic_reactor.constants import (PLUGIN_KOJI_IMPORT_PLUGIN_KEY,
                                       PLUGIN_GROUP_MANIFESTS_KEY,
                                       PLUGIN_KOJI_PARENT_KEY)
 from atomic_reactor.util import (get_build_json, get_preferred_label,
-                                 df_parser, ImageName, get_checksums, get_primary_images)
+                                 df_parser, ImageName, get_checksums, get_primary_images,
+                                 get_manifest_media_type,
+                                 get_digests_map_from_annotations)
 from atomic_reactor.koji_util import (create_koji_session, Output, KojiUploadLogger)
 from osbs.conf import Configuration
 from osbs.api import OSBS
@@ -210,7 +212,6 @@ class KojiImportPlugin(ExitPlugin):
 
         return output
 
-
     def set_help(self, extra, worker_metadatas):
         all_annotations = [get_worker_build_info(self.workflow, platform).build.get_annotations()
                            for platform in worker_metadatas]
@@ -300,6 +301,13 @@ class KojiImportPlugin(ExitPlugin):
                 pullspec = "{0}/{1}:{2}".format(registry.uri, repo,
                                                 version_release)
                 index['pull'].append(pullspec)
+
+                # Store each digest with according media type
+                index['digests'] = {}
+                for version, digest in manifest_list_digests[0].items():
+                    if digest:
+                        media_type = get_manifest_media_type(version)
+                        index['digests'][media_type] = digest
                 break
             extra['image']['index'] = index
         # group_manifests returns None if didn't run, [] if group=False
@@ -323,6 +331,13 @@ class KojiImportPlugin(ExitPlugin):
                             instance['extra']['docker']['repositories'] = repositories
                             self.log.debug("reset tags to so that docker is %s",
                                            instance['extra']['docker'])
+                            annotations = get_worker_build_info(self.workflow, platform).\
+                                build.get_annotations()
+                            digests = {}
+                            if 'digests' in annotations:
+                                digests = get_digests_map_from_annotations(annotations['digests'])
+                                instance['extra']['docker']['digests'] = digests
+
 
     def get_output_metadata(self, path, filename):
         """
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -41,7 +41,8 @@ from atomic_reactor.util import (get_version_of_tools, get_checksums,
                                  get_build_json, get_preferred_label,
                                  get_docker_architecture, df_parser,
                                  are_plugins_in_order,
-                                 get_image_upload_filename)
+                                 get_image_upload_filename,
+                                 get_digests_map_from_annotations)
 from atomic_reactor.koji_util import (create_koji_session, tag_koji_build,
                                       Output, KojiUploadLogger)
 from atomic_reactor.rpm_util import parse_rpm_output, rpm_qf_args
@@ -451,6 +452,10 @@ class KojiPromotePlugin(ExitPlugin):
                 },
             },
         })
+        annotations = self.workflow.build_result.annotations
+        if annotations and 'digests' in annotations:
+            digests = get_digests_map_from_annotations(annotations['digests'])
+            metadata['extra']['docker']['digests'] = digests
 
         if not config:
             del metadata['extra']['docker']['config']
--- a/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
+++ b/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
@@ -26,7 +26,7 @@ from atomic_reactor.constants import (PLUGIN_KOJI_IMPORT_PLUGIN_KEY,
                                       PLUGIN_GROUP_MANIFESTS_KEY,
                                       MEDIA_TYPE_DOCKER_V1)
 from atomic_reactor.plugin import ExitPlugin
-from atomic_reactor.util import get_build_json
+from atomic_reactor.util import get_build_json, get_manifest_media_version
 
 
 class StoreMetadataInOSv3Plugin(ExitPlugin):
@@ -81,7 +81,7 @@ class StoreMetadataInOSv3Plugin(ExitPlugin):
             for image in self.workflow.tag_conf.images:
                 image_str = image.to_str()
                 if image_str in registry.digests:
-                    digest = registry.digests[image_str].default
+                    digest = registry.digests[image_str]
                     digests[image.to_str(registry=False)] = digest
 
         return digests
@@ -129,13 +129,16 @@ class StoreMetadataInOSv3Plugin(ExitPlugin):
         pullspecs = []
         for registry in self._get_registries():
             for image in self.workflow.tag_conf.images:
-                if image.to_str() in digests:
+                image_str = image.to_str()
+                if image_str in digests:
                     pullspecs.append({
                         "registry": registry.uri,
                         "repository": image.to_str(registry=False, tag=False),
                         "tag": image.tag or 'latest',
-                        "digest": digests[image.to_str()]
+                        "digest": digests[image_str].default,
+                        "version": get_manifest_media_version(digests[image_str])
                     })
+
         return pullspecs
 
     def get_plugin_metadata(self):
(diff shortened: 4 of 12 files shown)


atomic_reactor/constants.py
```diff
@@ -101,3 +101,11 @@
 HTTP_CLIENT_STATUS_RETRY = (408, 500, 502, 503, 504)
 # requests timeout in seconds
 HTTP_REQUEST_TIMEOUT = 600
+
+
+# Media types
+MEDIA_TYPE_V1 = "application/json"
+MEDIA_TYPE_V2_SCHEMA1 = "application/vnd.docker.distribution.manifest.v1+json"
+MEDIA_TYPE_V2_SCHEMA2 = "application/vnd.docker.distribution.manifest.v2+json"
+MEDIA_TYPE_V2_MANIFEST_LIST = "application/vnd.docker.distribution.manifest.list.v2+json"
```
twaugh: Wonder if we should call these `MEDIA_TYPE_DOCKER_...`?

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -300,6 +300,13 @@ def set_group_manifest_info(self, extra, worker_metadatas):
                 pullspec = "{0}/{1}:{2}".format(registry.uri, repo,
                                                 version_release)
                 index['pull'].append(pullspec)
+
+                # Store each digest with according media type
+                index['digests'] = {}
+                for version, digest in manifest_list_digests[0].__dict__.items():
```
twaugh: Let's avoid this and make `ManifestDigest` a bit friendlier to work with, e.g. have it inherit from `UserDict`, have the v1, v2, v2_list etc attributes look up the value from the dict.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -320,9 +327,21 @@ def set_group_manifest_info(self, extra, worker_metadatas):
...
 
+                            annotations = get_worker_build_info(self.workflow, platform).\
+                                build.get_annotations()
+
                             instance['extra']['docker']['repositories'] = repositories
                             self.log.debug("reset tags to so that docker is %s",
                                            instance['extra']['docker'])
+                            instance['extra']['docker']['digests'] = {}
+
+                            if 'digests' in annotations.keys():
```
twaugh: Could be just `if 'digests' in annotations:`

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -320,9 +327,21 @@ def set_group_manifest_info(self, extra, worker_metadatas):
 
                                 repositories.append(pullspec)
 
+                            annotations = get_worker_build_info(self.workflow, platform).\
+                                build.get_annotations()
+
                             instance['extra']['docker']['repositories'] = repositories
                             self.log.debug("reset tags to so that docker is %s",
                                            instance['extra']['docker'])
+                            instance['extra']['docker']['digests'] = {}
```
twaugh: Could have a local variable for this dict until it's finished, then assign it to instance[...][...][...] to save repeated look-ups.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -133,8 +134,10 @@ def get_pullspecs(self, digests):
                         "registry": registry.uri,
                         "repository": image.to_str(registry=False, tag=False),
                         "tag": image.tag or 'latest',
-                        "digest": digests[image.to_str()]
+                        "digest": digests[image.to_str()].default,
+                        "version": get_manifest_media_version(digests[image.to_str()])
```
twaugh: Can we cache `digests[image.to_str()]` in a local variable?

atomic_reactor/constants.py
vrutkovs: Agree, fixed

atomic_reactor/plugins/exit_koji_import.py
vrutkovs: Fixed

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
vrutkovs: Fixed

atomic_reactor/constants.py
```diff
@@ -101,3 +101,11 @@
 HTTP_CLIENT_STATUS_RETRY = (408, 500, 502, 503, 504)
 # requests timeout in seconds
 HTTP_REQUEST_TIMEOUT = 600
+
+
+# Media types
+MEDIA_TYPE_DOCKER_V1 = "application/json"
```
lcarva: What about turning all this into a single data structure? It'll make usage a bit simpler (one thing to import)
```python
MEDIA_TYPES = {
   'v1': 'application/json',
   'v2schema2': 'application/vnd.docker.distribution.manifest.v1+json'
   ....
```

or if we're concerned with mutability:
```python
_MediaTypes = namedtuple('MediaTypes', 'v1,v2schema2,...')
MEDIA_TYPES = _MediaTypes(v1='application/json', ...)
```

twaugh:
Let's make sure to document this in docs/koji.md.

atomic_reactor/constants.py
vrutkovs: I like the idea, but in most places only one or two constants are used, so it would just make things look longer with no real improvement.

@twaugh @rcerven thoughts?

atomic_reactor/util.py
```diff
@@ -640,14 +642,18 @@ def get_credentials(self, docker_registry):
...
-        self.oci = oci
+    content_type = {
+        'v1': MEDIA_TYPE_DOCKER_V2_SCHEMA1,
+        'v2': MEDIA_TYPE_DOCKER_V2_SCHEMA2,
+        'v2_list': MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST,
+        'oci': MEDIA_TYPE_OCI_V1
+    }
+
+    def __init__(self, **kwargs):
+        self.update(**kwargs)
```
twaugh: Think this is how `dict` already behaves. Maybe we don't need `__init__` here?

atomic_reactor/util.py
```diff
@@ -661,18 +667,23 @@ def default(self):
         """
         return self.v2_list or self.oci or self.v2 or self.v1
 
+    def __getattr__(self, attr):
+        return self.get(attr, None)
```
twaugh: Should this raise `AttributeError` if `attr not in self`?

atomic_reactor/util.py
```diff
@@ -661,18 +667,23 @@ def default(self):
...
 
 def get_manifest_media_type(version):
-    if version in ('v1', 'v2'):
-        return 'application/vnd.docker.distribution.manifest.{}+json'.format(version)
-    elif version == 'v2_list':
-        return 'application/vnd.docker.distribution.manifest.list.v2+json'
-    elif version == 'oci':
-        return 'application/vnd.oci.image.manifest.v1+json'
+    if version in ManifestDigest.content_type.keys():
+        return ManifestDigest.content_type[version]
```
twaugh: Maybe `try:`/`except KeyError:`?

atomic_reactor/util.py
```diff
@@ -661,18 +667,23 @@ def default(self):
...
     else:
         raise RuntimeError("Unknown manifest schema type")
 
 
+def get_manifest_media_version(digest):
+    for version in ManifestDigest.content_type.keys():
+        if getattr(digest, version) == digest.default:
```
twaugh: or `version[digest] == digest.default`

atomic_reactor/util.py
```diff
@@ -661,18 +667,23 @@ def default(self):
...
-    elif version == 'oci':
-        return 'application/vnd.oci.image.manifest.v1+json'
+    if version in ManifestDigest.content_type.keys():
+        return ManifestDigest.content_type[version]
     else:
         raise RuntimeError("Unknown manifest schema type")
 
 
+def get_manifest_media_version(digest):
+    for version in ManifestDigest.content_type.keys():
```
twaugh: Don't need `.keys()` here.

atomic_reactor/util.py
```diff
@@ -661,18 +667,23 @@ def default(self):
...
+        if getattr(digest, version) == digest.default:
+            return version
```
twaugh: Should it raise an exception if it wasn't found?

docs/koji.md
```diff
@@ -55,6 +55,7 @@ The index map has these entries:
 
 - `pull` (str list): docker pull specifications for the manifest list, by tag and by digest
 - `tags` (str list): tags applied to the manifest list when it was created
+- `digests` (map): a dict of media types and digests of the image
```
twaugh: https://docs.pagure.org/koji/content_generator_metadata/ has this:

> digests: A map of media type (such as “application/vnd.docker.distribution.manifest.v2+json”) to manifest digest (a string usually starting “sha256:”), for each available media type.

I'd like to avoid "digests of the image" since it's the manifest in particular that it's a digest of.

docs/koji.md
```diff
@@ -157,7 +159,10 @@ Example type-specific content generator metadata in context:
           "latest",
           "1.0.0-1",
           "1.0.0"
-        ]
+        ],
+        "digests": {
+          "application/vnd.docker.distribution.manifest.v2+json": "sha256:123def"
```
twaugh: To make things really clear, we could also add an example entry for manifest.v1, both here and in 'repositories' 12 lines above.

atomic_reactor/constants.py
twaugh: I'm fine with either way.

atomic_reactor/util.py
vrutkovs: It shouldn't, the previous version have set other digest versions to `None` by default, so this is what it should return if digest was not set for this version

docs/koji.md
vrutkovs: Good idea, updated

atomic_reactor/util.py
vrutkovs: Agreed, fixed

atomic_reactor/util.py
vrutkovs: I'm not sure if this situation is realistic.  If no version has been set, `default` would return `None` - and it would match `v1` as its the first tag.

The only case where exception can be thrown is `def default` doesn't list all items from `content_type` - in this case tests would be broken.

docs/koji.md
vrutkovs: Added an example with `v1` manifest (although the code would always prefer either v1 or v2 IIRC)

atomic_reactor/util.py
vrutkovs: That didn't work, both tests and smoketests fail with `KeyError`, I don't think its worth adding an exception handler for this

atomic_reactor/util.py
vrutkovs: Implemented

atomic_reactor/util.py
vrutkovs: Done

atomic_reactor/constants.py
lcarva: Let's drop this suggestion then.

atomic_reactor/util.py
twaugh: Maybe `AttributeError` if `attr not in self.content_type` then? Otherwise `digest.defalut` (e.g. a typo) will evaluate to None instead of raising an exception.

docs/koji.md
```diff
@@ -145,6 +147,7 @@ Example type-specific content generator metadata in context:
         ],
         "repositories": [
           "registry.example.com/product/package-name:1.0.0-1",
+          "registry.example.com/product/package-name@sha256:123abc..."
```
twaugh: Needs a comma

atomic_reactor/util.py
vrutkovs: Good idea. Added a test for that

docs/koji.md
vrutkovs: Done

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -323,6 +330,17 @@ def set_group_manifest_info(self, extra, worker_metadatas):
                             instance['extra']['docker']['repositories'] = repositories
                             self.log.debug("reset tags to so that docker is %s",
                                            instance['extra']['docker'])
+                            annotations = get_worker_build_info(self.workflow, platform).\
```
lcarva: In the near future, we should consider breaking up `set_group_manifest_info` into smaller methods. As it stands it's quite difficult to follow along.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -451,6 +452,16 @@ def add_log_type(output):
                 },
             },
         })
+        annotations = self.workflow.build_result.annotations
+        if annotations and 'digests' in annotations:
+            digests = {}
+            digests_annotations = json.loads(annotations['digests'])
+            for digest_annotation in digests_annotations:
```
lcarva: Code duplication. Consider moving this for-loop to helper method.

vrutkovs: Moved to a separate method in `utils`

atomic_reactor/plugins/exit_koji_import.py
vrutkovs: Agree, we should run cyclomatic complexity checker (`mccabe` from PyPi) and split large functions into smaller ones. I don't think its worth doing in this PR though

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -300,6 +300,13 @@ def set_group_manifest_info(self, extra, worker_metadatas):
                 pullspec = "{0}/{1}:{2}".format(registry.uri, repo,
                                                 version_release)
                 index['pull'].append(pullspec)
+
+                # Store each digest with according media type
+                index['digests'] = {}
+                for version, digest in manifest_list_digests[0].items():
+                    if digest:
+                        media_type = get_manifest_media_type(version)
```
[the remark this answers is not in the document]
vrutkovs: Resolved

lcarva:
Updated release notes draft:
> - Koji build metadata now includes digests-version map

PR #860, opened 2017-10-06
Ensure krb principal/keytab are str type in koji_util

maxamillion:
Thanks to @puiterwijk for the patch

### DIFF
--- a/atomic_reactor/koji_util.py
+++ b/atomic_reactor/koji_util.py
@@ -86,8 +86,8 @@ def koji_login(session,
         # Use Kerberos
         logger.info("Using Kerberos for Koji authentication")
         if krb_principal and krb_keytab:
-            kwargs['principal'] = krb_principal
-            kwargs['keytab'] = krb_keytab
+            kwargs['principal'] = str(krb_principal)
+            kwargs['keytab'] = str(krb_keytab)
 
         result = session.krb_login(**kwargs)
 
--- a/atomic_reactor/plugins/exit_koji_tag_build.py
+++ b/atomic_reactor/plugins/exit_koji_tag_build.py
@@ -57,13 +57,13 @@ class KojiTagBuildPlugin(ExitPlugin):
             raise RuntimeError('specify both koji_principal and koji_keytab '
                                'or neither')
 
-        self.kojihub = kojihub
+        self.kojihub = str(kojihub)
         self.koji_auth = {
-            "proxyuser": koji_proxy_user,
-            "ssl_certs_dir": koji_ssl_certs,
+            "proxyuser": str(koji_proxy_user) if koji_proxy_user else None,
+            "ssl_certs_dir": str(koji_ssl_certs) if koji_ssl_certs else None,
             # krbV python library throws an error if these are unicode
-            "krb_principal": str(koji_principal),
-            "krb_keytab": str(koji_keytab)
+            "krb_principal": str(koji_principal) if koji_principal else None,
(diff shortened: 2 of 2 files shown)


atomic_reactor/koji_util.py
```diff
@@ -75,8 +75,8 @@ def koji_login(session,
         # Use Kerberos
         logger.info("Using Kerberos for Koji authentication")
         if krb_principal and krb_keytab:
-            kwargs['principal'] = krb_principal
-            kwargs['keytab'] = krb_keytab
+            kwargs['principal'] = str(krb_principal)
```
lcarva: I'm assuming this is because the underlying kerberos c api doesn't like unicode?
If so, can we use something like [six](https://pythonhosted.org/six/#binary-and-text-data) to handle the conversion in both python 2 and 3?

Also, we can do the same to `hub_url` parameter in `create_koji_session`, hiding this from the plugins.

maxamillion: Yes, it appears the krb c api doesn't like unicode. We could do something like six but what's the concern about python2 to python3 here? Isn't `str()` valid in both?

I don't have a real preference on how it's handled, if you'd prefer `hub_url` be a param in `create_koji_session` then :+1:

puiterwijk: @maxamillion The last point I had to fix was exactly the `hub_url` in `create_koji_session` :).

lcarva: If `str` works for both python versions, let's just go with that. But I don't think it will on python3. Since str there means unicode...

maxamillion: @lcarva nvm, I see what you mean. It's not that it would/wouldn't work syntactically for py2/py3 but the issue of the krb5 libs not liking unicode. I'll have to find a way to test this.

lcarva:
PR unit tests are failing. Once that's fixed, we can merge this PR as is. (Might need to rebase with master).

lcarva:
> PR unit tests are failing. Once that's fixed, we can merge this PR as is. (Might need to rebase with master).

@maxamillion, have a chance to look into that?

maxamillion:
@lcarva I'll get this fixed up in just a bit.

atomic_reactor/koji_util.py
puiterwijk: Sorry for the late reply, but there's two points of update on this now:
- The code that was broken was the krbV codepaths (which were required for keytab use), and that just does not exist under py3. The GSSAPI codepaths are totally fine with getting unicode.
- As of Koji 1.15.0, the GSSAPI codepaths are also used for keytab auth, which are happy with unicode, which makes this patch no longer needed for Fedora. The patch might still be useful for setups without gssapi (thus only krbV support), which only works with py2 anyway. And as said, GSSAPI is not picky whether it gets basestring or unicode.

puiterwijk:
@maxamillion and @lcarva, information regarding the py2/py3 discussions in my last comment. I'll still rebase and fix the tests in case you decide you still want it.

puiterwijk:
These test failures are Fedora rawhide DNF/RPM issues

atomic_reactor/koji_util.py
lcarva: Fedora was the only one affected by this. If that's no longer true, let's drop this PR. Agree?

puiterwijk: Agreed. I was thinking there might've been other people, but those are likely still on Py2.

lcarva:
After talking to @puiterwijk, we agreed to close this.

PR #863, opened 2017-10-13
Koji ssl auth cleanup

lcarva:


lcarva:
Updated release notes draft:
> - Koji ssl authentication no longer requires `ca` and `serverca` files.

PR #864, opened 2017-10-16
Store digests for all versions in koji metadata.

vrutkovs:
This PR changes exit_store_metadata, group_manifest and koji_import plugins 
to store all available image digests in worker annotations, group only v2 images 
(if needed) and save all image digests in koji image metadata

### DIFF
--- a/atomic_reactor/plugins/post_group_manifests.py
+++ b/atomic_reactor/plugins/post_group_manifests.py
@@ -16,7 +16,8 @@ import json
 import requests
 
 from atomic_reactor.plugin import PostBuildPlugin, PluginFailedException
-from atomic_reactor.util import RegistrySession, registry_hostname, ManifestDigest
+from atomic_reactor.util import (RegistrySession, registry_hostname, ManifestDigest,
+                                 get_manifest_media_type)
 from atomic_reactor.constants import (PLUGIN_GROUP_MANIFESTS_KEY, MEDIA_TYPE_DOCKER_V2_SCHEMA2,
                                       MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST, MEDIA_TYPE_OCI_V1,
                                       MEDIA_TYPE_OCI_V1_INDEX)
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -340,7 +348,7 @@ def run(self):
             session = self.get_registry_session(registry)
 
             if self.group:
-                digest = self.group_manifests_and_tag(session, source)
+                return self.group_manifests_and_tag(session, source)
```
twaugh: This is within a loop.

Don't we want to append to a list here, and return the list when the loop finishes? (This looks to be a bug that slipped through in https://github.com/projectatomic/atomic-reactor/pull/846)

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -33,6 +34,12 @@
 class GroupManifestsPlugin(PostBuildPlugin):
     is_allowed_to_fail = False
     key = PLUGIN_GROUP_MANIFESTS_KEY
+    manifest_media_types = [
+        MEDIA_TYPE_DOCKER_V2_SCHEMA2,
+        MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST,
+        MEDIA_TYPE_OCI_V1,
+        MEDIA_TYPE_OCI_V1_INDEX
```
twaugh: Probably want to leave MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST and MEDIA_TYPE_OCI_V1_INDEX out of this list.

atomic_reactor/plugins/post_group_manifests.py
vrutkovs: Ah, yes, gonna fix it.

Its a bit weird though - we always use just first result from it anyway

twaugh: I'd also be happy with a check that there is only one item to iterate over, and a NotImplemented exception raised if there are more. :)

atomic_reactor/plugins/post_group_manifests.py
vrutkovs: Agree, workers wouldn't produce manifest lists / OCI indexes anyway. Implemented

vrutkovs: Had to revert this - these media types are also used to fetch the digest of the created manifest list / OCI index, so it has to include media types for those - otherwise tests fail

atomic_reactor/plugins/post_group_manifests.py
vrutkovs: This seems tricky - and also arguable.

We're preparing a manifest list using same worker builds - we're just pushing those to different registries, so essentially there is no difference which manifest list should be used later on.

Raising NotImplementedError here causes a lot of test failures - lets implement this in a separate story.

lcarva:
Not updating release notes draft because this is bugfix to an unreleased feature: `Koji build metadata now includes digests-version map`

PR #865, opened 2017-10-17
pulp_pull: retry on 403 but log an error

twaugh:
Occasionally a 403 status code response has been seen from Crane. Whenever it has happened, retrying the same request manually at a later time shows no error.

On the theory that this only occurs during a small window just after publish, retry on 403 status code responses. However, when doing this, also log an error with information about the final request and the response, to help us understand more about it.


lcarva:
Updated release notes draft:
> - Workaround to retry on 403 for crane requests

PR #868, opened 2017-10-18
Add logging to report cluster used to start build in orchestrate build

bfontecc007:
This reports which cluster is being used as starting the build is attempted. It is possible that this will not be the cluster ultimately used, due to retry policy. Logging a cluster with certainty would involve logging it after the build is already done so this is a tradeoff.

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -456,8 +456,8 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
 
         if build_info.build:
             try:
-                self.log.info('%s - created build %s', cluster_info.platform,
-                              build_info.name)
+                self.log.info('%s - created build %s on cluster %s.', cluster_info.platform,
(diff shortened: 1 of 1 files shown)


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -504,6 +504,8 @@ def select_and_start_cluster(self, platform):
             for cluster_info in possible_cluster_info:
                 ctx = retry_contexts[cluster_info.cluster.name]
                 try:
+                    self.log.info('Attempting to start build for platform %s on cluster %s',
```
lcarva: Do we need a log message for when the build was actually started? Right before `build_info.watch_logs()` ?

bfontecc007:
Sure I can add more logging. Should we keep this one too?

bfontecc007:
End to End testing is fine on this

csomh:
like these logs +1

lcarva:
Not updating release notes draft because this is an enhancement to an unreleased feature.

PR #869, opened 2017-10-19
retry on docker pull and push

rcerven:

### DIFF
--- a/atomic_reactor/core.py
+++ b/atomic_reactor/core.py
@@ -30,16 +30,18 @@ import json
 import requests
 import time
 import docker
+import atomic_reactor.util
 from docker.errors import APIError
+from functools import wraps
 
 from atomic_reactor.constants import CONTAINER_SHARE_PATH, CONTAINER_SHARE_SOURCE_SUBDIR,\
         BUILD_JSON, DOCKER_SOCKET_PATH, DOCKER_MAX_RETRIES, DOCKER_BACKOFF_FACTOR,\
         DOCKER_CLIENT_STATUS_RETRY
 from atomic_reactor.source import get_source_instance_for
 from atomic_reactor.util import (
-    ImageName, wait_for_command, clone_git_repo, figure_out_build_file, Dockercfg)
+    ImageName, clone_git_repo, figure_out_build_file, Dockercfg)
 
-from requests.packages.urllib3.exceptions import InsecureRequestWarning
+from requests.packages.urllib3.exceptions import InsecureRequestWarning, ProtocolError
 
 requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
 
@@ -231,6 +233,13 @@ def retry(function, *args, **kwargs):
                 raise
 
 
+class RetryGeneratorException(Exception):
+    """Retry generator pull/push failed"""
+    def __init__(self, message, error, *args, **kwargs):
+        super(RetryGeneratorException, self).__init__(message, *args, **kwargs)
+        self.error = error
+
+
 class WrappedDocker(object):
     def __init__(self, **kwargs):
         self.retry_times = kwargs.pop('retry', None)
@@ -246,6 +255,7 @@ class WrappedDocker(object):
         orig_attr = getattr(self.wrapped, attr)
 
         if callable(orig_attr):
+            @wraps(orig_attr)
             def hooked(*args, **kwargs):
                 return retry(orig_attr, *args, retry=self.retry_times, **kwargs)
             return hooked
@@ -272,10 +282,46 @@ class DockerTasker(LastLogger):
 
         if hasattr(docker, 'AutoVersionClient'):
             client_kwargs['version'] = 'auto'
-        client_kwargs['retry'] = retry_times
+        self.retry_times = retry_times
+        client_kwargs['retry'] = self.retry_times
 
         self.d = WrappedDocker(**client_kwargs)
 
+    def retry_generator(self, function, *args, **kwargs):
+        retry_times = int(kwargs.pop('retry_times', self.retry_times))
+        retry_delay = DOCKER_BACKOFF_FACTOR
+        retry_client_statuses = DOCKER_CLIENT_STATUS_RETRY
+
+        for counter in range(retry_times + 1):
+            exc = None
+            context = None
+            try:
+                logs_gen = function(*args, **kwargs)
+                cmd_result = atomic_reactor.util.wait_for_command(logs_gen)
+            except ProtocolError as e:
+                exc = e
+                context = e.args
+            except APIError as e:
+                exc = e
+                context = e.response.content
+            else:
+                if cmd_result.is_failed():
+                    exc = cmd_result.error_detail
+                    context = cmd_result.error
+
+            if exc:
+                if counter == retry_times or \
+                    (isinstance(exc, APIError) and
+                     exc.response.status_code not in retry_client_statuses):
+                    raise RetryGeneratorException("Failed to %s image %s: %r" %
+                                                  (function.__name__, args, context),
+                                                  exc)
+
+                logger.info("retrying %s - %s on %r", function.__name__, args, context)
+                time.sleep(retry_delay * (2 ** counter))
+                continue
+
+            return cmd_result
 
     def build_image_from_path(self, path, image, stream=False, use_cache=False, remove_im=True):
         """
@@ -471,12 +517,16 @@ class DockerTasker(LastLogger):
         logger.info("pulling image '%s' from registry", image)
         logger.debug("image = '%s', insecure = '%s'", image, insecure)
         try:
-            logs_gen = self.d.pull(image.to_str(tag=False), tag=image.tag,
-                                   insecure_registry=insecure, decode=True, stream=True)
+            command_result = self.retry_generator(self.d.pull,
+                                                  image.to_str(tag=False),
+                                                  tag=image.tag, insecure_registry=insecure,
+                                                  decode=True, stream=True)
         except TypeError:
             # because changing api is fun
-            logs_gen = self.d.pull(image.to_str(tag=False), tag=image.tag, decode=True, stream=True)
-        command_result = wait_for_command(logs_gen)
+            command_result = self.retry_generator(self.d.pull,
+                                                  image.to_str(tag=False),
+                                                  tag=image.tag, decode=True, stream=True)
+
         self.last_logs = command_result.logs
         return image.to_str()
 
@@ -546,17 +596,17 @@ class DockerTasker(LastLogger):
         try:
             # push returns string composed of newline separated jsons; exactly what 'docker push'
             # outputs
-            logs = self.d.push(image.to_str(tag=False), tag=image.tag, insecure_registry=insecure,
-                               decode=True, stream=True)
+            command_result = self.retry_generator(self.d.push,
+                                                  image.to_str(tag=False),
+                                                  tag=image.tag, insecure_registry=insecure,
+                                                  decode=True, stream=True)
         except TypeError:
             # because changing api is fun
-            logs = self.d.push(image.to_str(tag=False), tag=image.tag, decode=True, stream=True)
+            command_result = self.retry_generator(self.d.push,
+                                                  image.to_str(tag=False),
+                                                  tag=image.tag, decode=True, stream=True)
 
-        command_result = wait_for_command(logs)
         self.last_logs = command_result.logs
-        if command_result.is_failed():
-            detail = command_result.error_detail
-            raise RuntimeError("Failed to push image %s: %s" % (image, detail))
         return command_result.parsed_logs
 
     def tag_and_push_image(self, image, target_image, insecure=False, force=False,
--- a/atomic_reactor/plugins/pre_pull_base_image.py
+++ b/atomic_reactor/plugins/pre_pull_base_image.py
@@ -15,6 +15,7 @@ import docker
 
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.util import get_build_json, ImageName
+from atomic_reactor.core import RetryGeneratorException
 
 
 class PullBaseImagePlugin(PreBuildPlugin):
@@ -62,16 +63,26 @@ class PullBaseImagePlugin(PreBuildPlugin):
 
             base_image_with_registry.registry = self.parent_registry
 
-        pulled_base = self.tasker.pull_image(base_image_with_registry,
-                                             insecure=self.parent_registry_insecure)
-        if (base_image_with_registry.namespace != 'library' and
-                not self.tasker.image_exists(base_image_with_registry.to_str())):
+        try:
+            self.tasker.pull_image(base_image_with_registry,
+                                   insecure=self.parent_registry_insecure)
+
+        except RetryGeneratorException as original_exc:
+            if base_image_with_registry.namespace == 'library':
+                raise
+
             self.log.info("'%s' not found", base_image_with_registry.to_str())
             base_image_with_registry.namespace = 'library'
             self.log.info("trying '%s'", base_image_with_registry.to_str())
-            pulled_base = self.tasker.pull_image(base_image_with_registry,
-                                                 insecure=self.parent_registry_insecure)
 
+            try:
+                self.tasker.pull_image(base_image_with_registry,
+                                       insecure=self.parent_registry_insecure)
+
+            except RetryGeneratorException:
+                raise original_exc
+
+        pulled_base = base_image_with_registry.to_str()
         self.workflow.pulled_base_images.add(pulled_base)
 
         # Attempt to tag it using a unique ID. We might have to retry
--- a/tests/test_tasker.py
+++ b/tests/test_tasker.py
@@ -10,9 +10,10 @@ from __future__ import print_function, unicode_literals
 
 from tests.fixtures import temp_image_name, docker_tasker  # noqa
 
-from atomic_reactor.core import DockerTasker, retry
-from atomic_reactor.util import ImageName, clone_git_repo
+from atomic_reactor.core import DockerTasker, retry, RetryGeneratorException
+from atomic_reactor.util import ImageName, clone_git_repo, CommandResult
 from tests.constants import LOCALHOST_REGISTRY, INPUT_IMAGE, DOCKERFILE_GIT, MOCK, COMMAND
+from requests.packages.urllib3.exceptions import ProtocolError
 from tests.util import requires_internet
 
 import docker
@@ -20,6 +21,7 @@ import docker.errors
 import requests
 import sys
 import time
+import atomic_reactor
 from docker.errors import APIError
 
 from flexmock import flexmock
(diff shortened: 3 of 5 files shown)


atomic_reactor/core.py
```diff
@@ -272,10 +277,60 @@ def __init__(self, base_url=None, retry_times=DOCKER_MAX_RETRIES,
...
+            exc = None
+            try:
+                logs_gen = function(*args, **kwargs)
+                cmd_result = atomic_reactor.util.wait_for_command(logs_gen)
+
+            except ProtocolError as e:
+                if counter != retry_times:
+                    exc = e
+                else:
+                    if hasattr(exc, 'message'):
```
vrutkovs: `Exception.message` is [deprecated](https://www.python.org/dev/peps/pep-0352/#transition-plan) and not available in py3, we shouldn't use it

atomic_reactor/core.py
```diff
@@ -272,10 +277,60 @@ def __init__(self, base_url=None, retry_times=DOCKER_MAX_RETRIES,
 
         if hasattr(docker, 'AutoVersionClient'):
             client_kwargs['version'] = 'auto'
-        client_kwargs['retry'] = retry_times
+        self.retry_times = retry_times
+        client_kwargs['retry'] = self.retry_times
 
         self.d = WrappedDocker(**client_kwargs)
 
+    def retry_generator(self, fnc_name, function, *args, **kwargs):
```
vrutkovs: Since `fnc_name` is not human-readable anyway, lets use `function.__name__`, since all the functions are from `self.d`

atomic_reactor/core.py
```diff
@@ -272,10 +277,60 @@ def __init__(self, base_url=None, retry_times=DOCKER_MAX_RETRIES,
...
+                if (e.response.status_code in retry_client_statuses and counter != retry_times):
+                    exc = e
+                else:
+                    raise RetryGeneratorException("Failed to %s image %s: %s" %
+                                                  (fnc_name, args, e.response.content))
+
+            if exc:
+                if (isinstance(exc, ProtocolError)):
+                    if hasattr(exc, 'message'):
+                        logger.info("retrying %s - %s on %s", fnc_name, args, exc.message)
```
vrutkovs: `Exception.message` is deprecated

atomic_reactor/core.py
```diff
@@ -471,14 +526,18 @@ def pull_image(self, image, insecure=False):
...
             # because changing api is fun
-            logs_gen = self.d.pull(image.to_str(tag=False), tag=image.tag, decode=True, stream=True)
-        command_result = wait_for_command(logs_gen)
+            command_result = self.retry_generator('pull_image', self.d.pull,
+                                                  image.to_str(tag=False),
+                                                  tag=image.tag, decode=True, stream=True)
+
         self.last_logs = command_result.logs
-        return image.to_str()
+        return command_result.parsed_logs
```
vrutkovs: Why is return value changed? Are we using pull logs somewhere?

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -62,16 +63,21 @@ def run(self):
...
-            base_image_with_registry.namespace = 'library'
-            self.log.info("trying '%s'", base_image_with_registry.to_str())
-            pulled_base = self.tasker.pull_image(base_image_with_registry,
-                                                 insecure=self.parent_registry_insecure)
+        try:
+            self.tasker.pull_image(base_image_with_registry,
+                                   insecure=self.parent_registry_insecure)
 
+        except RetryGeneratorException:
+            if base_image_with_registry.namespace != 'library':
```
vrutkovs: Why is `not self.tasker.image_exists(base_image_with_registry.to_str()))` omitted here?

atomic_reactor/core.py
rcerven: I can't use __name__ as that will just show 'hooked some reference'

atomic_reactor/core.py
rcerven: python2 has message, python3 has that traceback

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: pull will now raise exceptions

atomic_reactor/core.py
rcerven: just to make it same as for push_image, and return value which is known even before call isn't very useful at all

atomic_reactor/core.py
vrutkovs: I'm pretty sure it works:
```
>>> import docker
>>> client = docker.APIClient()
>>> client.pull.__name__
'pull'
```

atomic_reactor/core.py
vrutkovs: That's how deprecation works - its still there, but no point to be used.

If we really want a traceback (I doubt that it would be useful) `traceback` module should be used instead

atomic_reactor/core.py
vrutkovs: Well, we're not using `push_image` logs anywhere as well. Maybe it should return `None` for both then?

atomic_reactor/plugins/pre_pull_base_image.py
vrutkovs: Could you elaborate?

atomic_reactor/core.py
rcerven: ahh ok will use .args

atomic_reactor/core.py
rcerven: ahh ok will use .args

atomic_reactor/core.py
rcerven: yeah but self.d is already wrapped, I've tried using __ name __ and didn't work

atomic_reactor/core.py
rcerven: yeah output of either isn't used anywhere

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: well we don't have to check image existence anymore because if 'pull'  is successful it won't be raising any error, but if there is problem it will raise, unlike before when it was just logging errors but didn't stop

atomic_reactor/core.py
rcerven: what do you think @twaugh @lcarva  should we leave logs or None?

atomic_reactor/core.py
```diff
@@ -231,6 +232,10 @@ def retry(function, *args, **kwargs):
                 raise
 
 
+class RetryGeneratorException(Exception):
+    """Retry generator pull/push failed"""
```
twaugh: Don't you want a constructor that takes a `reason` or `cause` parameter or something, being the original exception? e.g. https://github.com/projectatomic/osbs-client/blob/0.43/osbs/exceptions.py#L20, or https://github.com/shazow/urllib3/blob/1.22/urllib3/exceptions.py#L76

atomic_reactor/core.py
twaugh: Is leaving the return value the same as it was an option?

atomic_reactor/core.py
```diff
@@ -30,16 +30,17 @@
 import requests
 import time
 import docker
+import atomic_reactor.util
```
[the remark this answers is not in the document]
rcerven: yes but if we import wait_for_command directly we can't flexmock it as it is in local namespace

atomic_reactor/core.py
rcerven: yeah right I can leave it as it was

atomic_reactor/plugins/pre_pull_base_image.py
vrutkovs: Ah, this part is removed as `image_exists` is assumed to be True if `pull_image` has worked

atomic_reactor/core.py
rcerven: ok will safe original error, not necessary only exception but in case cmd_result.is_failed() will safe cmd_result

atomic_reactor/core.py
vrutkovs: That's because `WrappedDocker.__getattr__`'s `hooked` doesn't preseve `orig_attr`'s `__name__` - it seems to be a one-line change to me

rcerven: ok will preserve it

atomic_reactor/core.py
```diff
@@ -248,6 +256,7 @@ def __getattr__(self, attr):
         if callable(orig_attr):
             def hooked(*args, **kwargs):
                 return retry(orig_attr, *args, retry=self.retry_times, **kwargs)
+            hooked.__name__ = orig_attr.__name__
```
twaugh: Just use `functools.wraps`?
```
@wraps(orig_attr)
def hooked(*args, **kwargs):
...
```

atomic_reactor/core.py
```diff
@@ -272,10 +281,57 @@ def __init__(self, base_url=None, retry_times=DOCKER_MAX_RETRIES,
...
+                else:
+                    raise RetryGeneratorException("Failed to %s image %s: %s" %
+                                                  (function.__name__, args, e.args),
+                                                  e)
+
+            except APIError as e:
+                if (e.response.status_code in retry_client_statuses and counter != retry_times):
+                    exc = e
+                else:
+                    raise RetryGeneratorException("Failed to %s image %s: %s" %
```
twaugh: Use %r for e.response.content, since it has not been decoded.

atomic_reactor/core.py
```diff
@@ -272,10 +281,57 @@ def __init__(self, base_url=None, retry_times=DOCKER_MAX_RETRIES,
...
+                    logger.info("retrying %s - %s on %s", function.__name__, args,
+                                exc.response.content)
+                time.sleep(retry_delay * (2 ** counter))
+
+            elif (cmd_result.is_failed() and counter != retry_times):
+                logger.info("retrying %s - %s on %s", function.__name__, args, cmd_result.error)
+                time.sleep(retry_delay * (2 ** counter))
+
+            elif cmd_result.is_failed():
+                raise RetryGeneratorException("Failed to %s image %s: %s" %
```
twaugh: Use %r for .content.

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -62,16 +63,21 @@ def run(self):
...
+                self.log.info("'%s' not found", base_image_with_registry.to_str())
+                base_image_with_registry.namespace = 'library'
+                self.log.info("trying '%s'", base_image_with_registry.to_str())
+                self.tasker.pull_image(base_image_with_registry,
+                                       insecure=self.parent_registry_insecure)
+            else:
+                raise
```
twaugh: Could put the `raise` block first to avoid needing to indent the rest, e.g.
```
if ... == 'library':
    # Nothing else to try
    raise

self.log.info(...)
...
```

Also, I think we want the error to look different for these two cases:
- retries failed due to some sort of network issue (this is a potentially serious system issue)
- the image just isn't there (this is just user error)

At the very least the exception's message should be different... even better would be a different exception. We seem to use RuntimeError for that sort of thing, but even that seems pretty generic.

atomic_reactor/core.py
rcerven: ok

atomic_reactor/core.py
rcerven: ok

atomic_reactor/core.py
rcerven: ok

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: ok will put raise block first

atomic_reactor/core.py
```diff
@@ -272,10 +282,57 @@ def __init__(self, base_url=None, retry_times=DOCKER_MAX_RETRIES,
...
+            except APIError as e:
+                if (e.response.status_code in retry_client_statuses and counter != retry_times):
+                    exc = e
+                else:
+                    raise RetryGeneratorException("Failed to %s image %s: %r" %
+                                                  (function.__name__, args, e.response.content),
+                                                  e)
+
+            if exc:
+                if (isinstance(exc, ProtocolError)):
```
lcarva: Does it make sense to move these log messages to their corresponding except-block?
This seems error prone (what if start catching another exception?)

atomic_reactor/core.py
```diff
@@ -272,10 +282,57 @@ def __init__(self, base_url=None, retry_times=DOCKER_MAX_RETRIES,
...
+        retry_client_statuses = DOCKER_CLIENT_STATUS_RETRY
+
+        for counter in range(retry_times + 1):
+            exc = None
+            try:
+                logs_gen = function(*args, **kwargs)
+                cmd_result = atomic_reactor.util.wait_for_command(logs_gen)
+
+            except ProtocolError as e:
+                if counter != retry_times:
```
lcarva: There's quite a bit of repetition in this method. Can we take the approach of only checking `counter != retry_times` only once. Something like this:

```python
exc = None
context = None

try:
  ...
except Protocol as e:
  exc = e
  context = e.args
exception APIError as e:
  exc = e
  context = e.reponse.content

if exc or cmd_result.is_failed():
  if counter == retry_times:
    raise RetryGeneratorException("Failed to %s image %s: %r" %" % 
                                  (function.__name__, args, context),
                                  exc)
  logger.info("retrying %s - %s on %s", function.__name__, args, context)
  time.sleep(retry_delay * (2 ** counter))

# Also handle return cases.
```

rcerven: ok reworked

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -62,16 +63,21 @@ def run(self):
...
-                                             insecure=self.parent_registry_insecure)
-        if (base_image_with_registry.namespace != 'library' and
-                not self.tasker.image_exists(base_image_with_registry.to_str())):
+        try:
+            self.tasker.pull_image(base_image_with_registry,
+                                   insecure=self.parent_registry_insecure)
+
+        except RetryGeneratorException:
+            if base_image_with_registry.namespace == 'library':
+                raise
```
twaugh: One thing I just thought of: this will report a failure to pull e.g. "library/mything" whereas the developer might have "FROM mything". Should we make any attempt to raise the *original* exception when both pulls fail?

rcerven: which original exception ? you mean protocol & api error? and in case it didn't raise but was is_failed()?

1st it tries to pull image regardless if there is 'library'

if that fails, that mean that image isn't accessible, so this tries another pull with 'library', and if that works we will continue to work with that

twaugh: I mean that when a developer tries to build from a Dockerfile that has "FROM mything:1.0" but mything:1.0 is not present in the registry, we try again by pulling library/mything:1.0 (this "library/" bit is a left-over from the API v1 days).

So the error the developer will see is something like:
```
failed to pull image ("library/mything:1.0",): {"message": "unknown: Not Found")}
```

but a more useful message might be:
```
failed to pull image ("mything:1.0",): {"message": "unknown: Not Found")}
```

i.e. try pulling mything:1.0, but if the CommandResult.is_failed() then try pulling library/mything:1.0 and if it works then great. If that also fails, though, report the original CommandResult failure from pulling mything:1.0, not from pulling library/mything:1.0.

Without doing this, the error message says that library/mything:1.0 is not found even though this is expected to be the case (since only mything:1.0 is present), and developers would not be clear on what the problem actually is.

Make sense?

tests/test_tasker.py
```diff
@@ -403,3 +404,56 @@ def test_retry_method(retry_times):
...
+            .times(retry_times + 1)
+            .and_raise(ProtocolError, "error"))
+    else:
+        (flexmock(atomic_reactor.util)
+            .should_receive('wait_for_command')
+            .times(retry_times + 1)
+            .and_return(cr))
+
+    if retry_times >= 0:
+        with pytest.raises(RetryGeneratorException):
```
twaugh: Maybe use `pytest.raises(..., message=...):` to make sure the exception description matches what we expect?

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: yeah it does now

tests/test_tasker.py
```diff
@@ -403,3 +404,61 @@ def test_retry_method(retry_times):
...
+        t = DockerTasker()
+
+    (flexmock(time)
+        .should_receive('sleep')
+        .and_return(None))
+
+    if not exc:
+        cr = CommandResult()
+        cr._error = "cmd_error"
+        cr._error_detail = "error_detail"
```
twaugh: Nit: in reality this isn't a string, it's a dict, e.g. `{"message": "error_detail"}`

tests/test_tasker.py
```diff
@@ -403,3 +404,61 @@ def test_retry_method(retry_times):
...
+            .times(retry_times + 1)
+            .and_return(cr))
+        error_message = 'cmd_error'
+
+    if retry_times >= 0:
+        with pytest.raises(RetryGeneratorException) as exc:
+            t.retry_generator(lambda *args, **kwargs: simplegen(),
+                              *my_args, **my_kwargs)
+
+        assert exc.match(repr(error_message))
```
twaugh: Doesn't `pytest.raises(RetryGeneratorException, match=error_message)` work?

twaugh: Minor: also, I think we should be able to expect the full exact message here.

rcerven: it didn't for me

tests/test_tasker.py
```diff
@@ -403,3 +404,61 @@ def test_retry_method(retry_times):
...
+            .times(retry_times + 1)
+            .and_return(cr))
+        error_message = 'cmd_error'
+
+    if retry_times >= 0:
+        with pytest.raises(RetryGeneratorException) as exc:
+            t.retry_generator(lambda *args, **kwargs: simplegen(),
+                              *my_args, **my_kwargs)
+
+        assert repr(error_message) in exc.value.args[0]
```
twaugh: I still think we ought to be making an assertion about what `repr(exc)` looks like, since that's what the user will see, rather than rummaging inside `exc`, but I'm happy enough with this.

rcerven: but repr(exc) is this : <ExceptionInfo RetryGeneratorException tblen=2>
that isn't what user will see

twaugh: Oh, it's an ExceptionInfo (I thought it was an Exception because it is named `exc` not `exc_info`). Then `repr(exc.value)` is what ought to be tested.

The existing test verifies that the exception message contains the actual error detail somewhere, but doesn't verify that the resulting exception shown in a Koji task result will actually be intelligible to users.

rcerven: ok will use repr(exc.value)

tests/test_tasker.py
```diff
@@ -236,7 +238,6 @@ def test_pull_image():
     remote_img.registry = LOCALHOST_REGISTRY
     t.tag_and_push_image(local_img, remote_img, insecure=True)
     got_image = t.pull_image(remote_img, insecure=True)
-    assert remote_img.to_str() == got_image
```
vrutkovs: Not sure why this check was removed - it seem valid to me

rcerven: it was removed before, when I changed return value of pull_image as it wasn't used anyway, I will return this assert

rcerven:
release note updated

PR #870, opened 2017-10-23
pull_base_image: make sure correct base image is stored in the annotations

vrutkovs:
Base image is now being tagged with buildid name to ensure base image
would not be removed in concurrent builds. This, however, makes
store_metadata plugin to save a wrong base-image-id annotation.

The solution to resolve this is to append a tagged base image to
workflow.pulled_base_images list and set_base_image to actual
base image. During exit_remove_built_image only images from
pulled_base_images would be attempted to be removed.

This fixes a regression introduced in https://github.com/projectatomic/atomic-reactor/pull/774


twaugh:
Draft release notes updated.

PR #871, opened 2017-10-24
pull_base_image: tag image with base_image name too

vrutkovs:
If a non-default docker registry is used, the base image
would not be inspectable using plain base image name.

The pulled image should be tagged with base name without
registry, as it would later on be used by other plugins.

### DIFF
--- a/atomic_reactor/plugins/pre_pull_base_image.py
+++ b/atomic_reactor/plugins/pre_pull_base_image.py
@@ -101,5 +101,11 @@ class PullBaseImagePlugin(PreBuildPlugin):
             self.log.error("giving up trying to pull image")
             raise RuntimeError("too many attempts to pull and tag image")
 
+        # Make sure registry image is tagged with plain base image name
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -101,5 +101,11 @@ def run(self):
             self.log.error("giving up trying to pull image")
             raise RuntimeError("too many attempts to pull and tag image")
 
+        # Make sure registry image is tagged with plain base image name
+        # so that it would be inspectable
```
twaugh: It's inspectable anyway, but is the issue that other plugins are looking for it using the plain image name without a registry? If so, let's make this comment say that.

vrutkovs: Changed to "other plugins would use base_image to inspect it"

lcarva:

PR #872, opened 2017-10-25
Fix test

twaugh:
Instead of directly comparing the lists of dicts, check every element of each is in the other.


lcarva:
Not updating release notes draft because this is a unit test only change.

PR #873, opened 2017-10-27
group_manifest: return ManifestDigest object when group=True

vrutkovs:
The plugin should return a list of ManifestDigest objects when
group=True, as its later on used by koji_import.

ManifestDigest is now a hashable type (derived from `HashableDict` class, which assigns a unique hash based on dict items).

### DIFF
--- a/atomic_reactor/plugins/post_group_manifests.py
+++ b/atomic_reactor/plugins/post_group_manifests.py
@@ -242,22 +242,22 @@ class GroupManifestsPlugin(PostBuildPlugin):
                                               target_repo, target_repo, tag=image.tag)
         # Get the digest of the manifest list using one of the tags
         registry_image = self.workflow.tag_conf.unique_images[0]
-        _, digest, _, _ = self.get_manifest(session,
-                                            registry_image.to_str(registry=False, tag=False),
-                                            registry_image.tag)
+        _, digest_str, _, _ = self.get_manifest(session,
+                                                registry_image.to_str(registry=False, tag=False),
+                                                registry_image.tag)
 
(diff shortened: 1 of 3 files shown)


atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -345,14 +345,14 @@ def get_registry_session(self, registry):
         return RegistrySession(registry, insecure=insecure, dockercfg_path=secret_path)
 
     def run(self):
-        digests = set()
+        digests = list()
```
lcarva: Why change it to a `list`? I believe `set` was used to ensure no duplicates are added.

vrutkovs: `dict` (which is Manifest Digest) is an unhashable type. I'm not sure if duplicates are possible in this list - and if there is a nice way to remove those

twaugh: Can we make `ManifestDigest` a hashable type?

vrutkovs: Its a bit complicated, but possible, I think. I believe it would use tuple as a base and have init part reworked - I'll try to implement this in a new commit, stay tuned

vrutkovs: That was surprisingly easier that I thought.

Introduced a new `HashableDict` class, which calculates a hash by converting dict items into `frozenset`. This puts several limitations - `ManifestDigest` won't change the hash if any values were updated after it was put in the `set`, but luckily we don't need to do this.

lcarva:
Updated release notes draft:
> - `group_manifest` returns `ManifestDigest` object when `group=True`

PR #876, opened 2017-11-06
Remove manifest list from the registry after pulp_sync

vrutkovs:
This also changes the return value of GroupManifest plugin to a map of repos as
keys and digests as values, as exit_delete_from_registry plugin needs to know which repo to remove digest from. 
As a result, `ManifestDigest` is no longer required to be hashable, so `HashableDict` class was removed

### DIFF
--- a/atomic_reactor/plugins/exit_delete_from_registry.py
+++ b/atomic_reactor/plugins/exit_delete_from_registry.py
@@ -13,6 +13,7 @@ import requests
 
 from atomic_reactor.plugin import ExitPlugin, PluginFailedException
 from atomic_reactor.util import RegistrySession, registry_hostname
+from atomic_reactor.constants import PLUGIN_GROUP_MANIFESTS_KEY
 from requests.exceptions import HTTPError, RetryError, Timeout
 
 
@@ -95,6 +96,17 @@ class DeleteFromRegistryPlugin(ExitPlugin):
 
         return deleted
 
+    def delete_manifest_lists(self, session, registry_noschema, deleted_digests):
+        manifest_list_digests = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY)
+        if not manifest_list_digests:
+            return
+
+        for repo, digest in manifest_list_digests.items():
+            url = self.make_url(repo, digest.default)
+            manifest = self.make_manifest(registry_noschema, repo, digest.default)
+            self.request_delete(session, url, manifest)
+            deleted_digests.add(digest.default)
+
     def get_worker_digests(self):
         """
          If we are being called from an orchestrator build, collect the worker
@@ -125,6 +137,9 @@ class DeleteFromRegistryPlugin(ExitPlugin):
 
         if registry_noschema not in worker_digests:
             return False
+
+        # Remove manifest list first to avoid broken lists in case an error occurs
+        self.delete_manifest_lists(session, registry_noschema, deleted_digests)
 
         digests = worker_digests[registry_noschema]
         for digest in digests:
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -302,8 +302,8 @@ class KojiImportPlugin(ExitPlugin):
             if not registries:
                 registries = self.workflow.push_conf.all_registries
             for registry in registries:
-                pullspec = "{0}/{1}@{2}".format(registry.uri, repo,
-                                                manifest_list_digests[0].v2_list)
+                manifest_list_digest = manifest_list_digests[repo]
+                pullspec = "{0}/{1}@{2}".format(registry.uri, repo, manifest_list_digest.default)
                 index['pull'] = [pullspec]
                 pullspec = "{0}/{1}:{2}".format(registry.uri, repo,
                                                 version_release)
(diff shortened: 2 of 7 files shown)


atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -292,7 +292,7 @@ def set_group_manifest_info(self, extra, worker_metadatas):
         tags = [image.tag for image in primary_images]
 
         manifest_list_digests = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY)
-        if manifest_list_digests:
+        if manifest_list_digests and len(manifest_list_digests) > 0:
```
lcarva: Checking length is redundant, isn't it? Empty dict evaluates to falsy.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -302,23 +302,25 @@ def set_group_manifest_info(self, extra, worker_metadatas):
             if not registries:
                 registries = self.workflow.push_conf.all_registries
             for registry in registries:
-                pullspec = "{0}/{1}@{2}".format(registry.uri, repo,
-                                                manifest_list_digests[0].v2_list)
+                manifest_list_digest = manifest_list_digests.get(repo)
```
lcarva: Nice! Much easier to read this way.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -302,23 +302,25 @@ def set_group_manifest_info(self, extra, worker_metadatas):
...
+                if not manifest_list_digest:
```
lcarva: In which scenario would the key for the repo be defined in `manifest_list_digests` but not a value?

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -302,23 +302,25 @@ def set_group_manifest_info(self, extra, worker_metadatas):
...
                 # Store each digest with according media type
                 index['digests'] = {}
-                for version, digest in manifest_list_digests[0].items():
+                for version, digest in manifest_list_digest.items():
                     if digest:
                         media_type = get_manifest_media_type(version)
                         index['digests'][media_type] = digest
                 break
             extra['image']['index'] = index
         # group_manifests returns None if didn't run, [] if group=False
```
lcarva: Is this comment still valid?

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -258,7 +258,7 @@ def group_manifests_and_tag(self, session, worker_digests):
             push_conf_registry.digests[image.tag] = digest
 
         self.log.info("%s: Manifest list digest is %s", session.registry, digest_str)
-        return digest
+        return manifest['repository'], digest
```
lcarva: `manifest` is the last item of `manifests` list, which gets created during one of the for-loops up above.
Can we make this a little bit less brittle? There could be a case where `manifest` is not even set at all!

atomic_reactor/plugins/exit_koji_import.py
vrutkovs: Right, fixed

atomic_reactor/plugins/post_group_manifests.py
vrutkovs: Correct, repo name should be fetched from the `ImageName` instance we use to fetch manifest digest.

Fixed

atomic_reactor/plugins/exit_koji_import.py
vrutkovs: Can't imagine one, its just a safety measure. Removed it for now, it doesn't seem to break the tests

atomic_reactor/plugins/exit_koji_import.py
vrutkovs: Agreed, fixed

atomic_reactor/plugins/exit_delete_from_registry.py
```diff
@@ -173,6 +185,10 @@ def run(self):
             orchestrator_delete = self.handle_worker_digests(session, worker_digests,
                                                              deleted_digests)
 
+            # Remove manifest from the registry
+            if orchestrator_delete:
+                self.delete_manifest_lists(session, registry_noschema, deleted_digests)
```
twaugh: Is it possible to arrange for the manifest lists to be deleted *before* the image manifests? Deleting them after leaves a small window of opportunity in which requests for that manifest list can fail (for example, if the client does not indicate they accept manifest lists).

tests/plugins/test_delete_from_registry.py
```diff
@@ -114,12 +121,28 @@ def test_delete_from_registry_plugin(saved_digests, req_registries, tmpdir, orch
...
         setattr(workflow, 'build_result', Y)
         setattr(workflow.build_result, 'annotations', annotations)
 
+        # group_manifest digest should be added only
+        # if there are worker builds and images are pushed to one registry
+        if len(req_registries) == 1 and len(saved_digests.keys()) == 1 and \
+           all(saved_digests.values()):
+            workflow.postbuild_results[PLUGIN_GROUP_MANIFESTS_KEY] = manifest_list_digests
+            for ml_repo, ml_digest in manifest_list_digests.items():
+                for reg, _ in req_registries.items():
```
twaugh: `for reg in req_registries:`?

atomic_reactor/plugins/exit_delete_from_registry.py
vrutkovs: Right, that's a good idea. Perhaps this should happen in `handle_worker_digests`

tests/plugins/test_delete_from_registry.py
vrutkovs: Right, good catch, fixed

lcarva:
Updated release notes draft:
> - Remove manifest list from the distribution registry registry after `pulp_sync`

PR #877, opened 2017-11-06
Add owner and submitter metadata

bfontecc007:
This adds the owner and submitter information to the build's metadata.

### DIFF
--- a/atomic_reactor/koji_util.py
+++ b/atomic_reactor/koji_util.py
@@ -167,3 +167,15 @@ def tag_koji_build(session, build_id, target, poll_interval=5):
         raise RuntimeError('Task %s failed to tag koji build' % task_id)
 
     return build_tag
+
+
+def get_koji_task_owner(session, task_id, default=""):
+    if task_id:
+        try:
+            koji_task_info = session.getTaskInfo(task_id)
+            koji_task_owner = session.getUser(koji_task_info['owner'])
+        except:
+            koji_task_owner = default
+    else:
+        koji_task_owner = default
+    return koji_task_owner
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -45,7 +45,8 @@ from atomic_reactor.util import (get_build_json, get_preferred_label,
                                  df_parser, ImageName, get_checksums, get_primary_images,
                                  get_manifest_media_type,
                                  get_digests_map_from_annotations)
-from atomic_reactor.koji_util import (create_koji_session, Output, KojiUploadLogger)
+from atomic_reactor.koji_util import (create_koji_session, Output, KojiUploadLogger,
+                                      get_koji_task_owner)
 from osbs.conf import Configuration
 from osbs.api import OSBS
 from osbs.exceptions import OsbsException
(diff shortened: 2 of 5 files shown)


atomic_reactor/koji_util.py
```diff
@@ -167,3 +167,15 @@ def tag_koji_build(session, build_id, target, poll_interval=5):
         raise RuntimeError('Task %s failed to tag koji build' % task_id)
 
     return build_tag
+
+
+def get_koji_task_owner(session, task_id, default=""):
+    if task_id is not None:
```
lcarva: Maybe just `if task_id:` ?

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -420,6 +421,12 @@ def get_build(self, metadata, worker_metadatas):
         if flatpak_source_info is not None:
             extra['image'].update(flatpak_source_info.koji_metadata())
 
+        if koji_task_id:
```
lcarva: Looks like `get_koji_task_owner` should be able to handle falsy `koji_task_id`.
This check is redundant in that case.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -420,6 +421,12 @@ def get_build(self, metadata, worker_metadatas):
...
+            koji_task_owner = get_koji_task_owner(self.session, koji_task_id, default=None)['name']
+        else:
+            koji_task_owner = None
+        extra['submitter'] = self.session.getLoggedInUser()['name']
```
vrutkovs: I don't think its correct - `getLoggedInUser` most likely would return koji account OSBS uses (`osbs`), not the original submitter.

See `_get_koji_submitter` in `exit_sendmail.py`

lcarva: Actually, submitter is the one who submitted the build via content generator API (osbs), and owner is the user that started the task (me, for example). The owner is used to populate the "Built by" field in the UI, for example. The submitter field is simply added for tracking purposes, AFAIK it's not actually actively used.

@breillyr, please confirm.

atomic_reactor/plugins/exit_koji_import.py
bfontecc007: I added that guard when I added the index by 'name'. You can't index None.

atomic_reactor/plugins/exit_koji_import.py
twaugh: Ah! Yes, now I see. "owner" is described in the [documentation](https://docs.pagure.org/koji/content_generator_metadata/).

I don't see "submitter" described anywhere except for the description of [this issue](https://pagure.io/koji/issue/299#), but from the story description the code looks right.

bfontecc007:
I added a change to the docs.

lcarva:
Updated release notes draft:
> - Koji build metadata contains owner and submitter

atomic_reactor/plugins/exit_koji_import.py
sosiouxme: @bfontecc007 why put `default=None` then if you know you'll get an exception when you actually get `None` back? Is that desirable if your task id comes up with nothing?

bfontecc007: @sosiouxme I don't really remember in great detail but maybe I thought it was an easier to interpret exception than the one you'd get if you tried to index string '' by 'name.'

twaugh: @sosiouxme is right, that doesn't make any sense.

It looks to me like it should be `get_koji_task_owner()` that does the ['name'] look-up, not the caller of that function.

PR #878, opened 2017-11-08
tests: flexmock requests.Sessions functions instead of requests module functions

vrutkovs:
This fixes several yum inject tests on Fedora, as these are running with networking
disabled and were incorrectly mocked


lcarva:
Not updating release notes draft because this is bugfix to unit tests.

PR #879, opened 2017-11-08
tests: skip tests if dockpulp has no imgutils module

vrutkovs:
I don't fully understand why tests in test_pulp_tag fail in Fedora,
but it seems previous runs are importing a mocked dockpulp and Python 
remembers that import. This PR would ensure that tests will not be 
executed on mocked dockpulp, as it doesn't have imgutils module

### DIFF
--- a/tests/plugins/test_pulp_tag.py
+++ b/tests/plugins/test_pulp_tag.py
@@ -150,7 +150,7 @@ def prepare(v1_image_ids={}):
     return tasker, workflow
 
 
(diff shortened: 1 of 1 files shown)


twaugh:
Shouldn't we be filling in the tests/mock/dockpulp stub to provide this?

vrutkovs:
In the future - yes, the short term goal now is 'allow Fedora release a package without disabling all tests'

lcarva:
Not updating release notes draft because this is bugfix to unit tests.

PR #880, opened 2017-11-08
WIP: spec: add BuildConflicts due to missing osbs-client method and fix osbs-client name

vrutkovs:
Fedora builds can't complete with outdated osbs-client, as some are using get_orchestrator_logs,
thus, if check is enabled, the package should require osbs-client 0.44

This should not affect RHEL

### DIFF
--- a/atomic-reactor.spec
+++ b/atomic-reactor.spec
@@ -55,7 +55,7 @@ BuildRequires:  python-dockerfile-parse >= 0.0.5
 BuildRequires:  python-docker-py
 BuildRequires:  python-flexmock >= 0.10.2
 BuildRequires:  python-six
-BuildRequires:  python-osbs >= 0.15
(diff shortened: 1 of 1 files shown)


twaugh:
Any reason not to just bump the BuildRequires minimum version?

vrutkovs:
Oh, indeed, it would be easier. Initially I though `BuildRequires` was outside `with check`

vrutkovs:
And now tests fail, as osbs-client was not yet build. Lets keep this open for some time

lcarva:
As of today, latest osbs-client package is osbs-client-0.41-1.

lcarva:
0.45 is now available in centOS, but we're hitting a fedora issue. Can you rebase to fix that?

lcarva:
@cverna, is this still needed for fedora?

lcarva:
Sounds good to me.

PR #882, opened 2017-11-09
group_manifests: include namespace if its present in the results

vrutkovs:
The repo name is used in `exit_delete_from_registry` plugin to remove a manifest list - and the plugin can fail to find the digest in the results list if namespace was not added. 

This PR adds `ImageName.get_repo()` method, which returns namespace + repo (with `explicit_namespace` for V1 support) and updated plugins to use it.

### DIFF
--- a/atomic_reactor/plugins/post_group_manifests.py
+++ b/atomic_reactor/plugins/post_group_manifests.py
@@ -258,7 +258,7 @@ class GroupManifestsPlugin(PostBuildPlugin):
             push_conf_registry.digests[image.tag] = digest
 
         self.log.info("%s: Manifest list digest is %s", session.registry, digest_str)
-        return registry_image.repo, digest
+        return registry_image.get_repo(explicit_namespace=False), digest
 
     def tag_manifest_into_registry(self, session, worker_digest):
         """
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -86,7 +86,7 @@ class ImageName(object):
         if self.repo is None:
             raise RuntimeError('No image repository specified')
 
-        result = self.repo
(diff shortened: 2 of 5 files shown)


twaugh:
Great catch! Do the unit tests need updating?

If we were writing that class again without having to worry about V1 we'd probably make `repo` include a namespace.

vrutkovs:
Yep, writing tests for `group_manifests` now

lcarva:
Can we have explicit tests for the changes to `ImageName` ?

vrutkovs:
Tests were updated, additional tests for `ImageName.get_repo` were added

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -258,7 +258,7 @@ def group_manifests_and_tag(self, session, worker_digests):
             push_conf_registry.digests[image.tag] = digest
 
         self.log.info("%s: Manifest list digest is %s", session.registry, digest_str)
-        return registry_image.repo, digest
+        return registry_image.get_repo(explicit_namespace=False), digest
```
rcerven: why not just use : registry_image.get_repo(), since that explicit_namespace=False is default ?

vrutkovs: I've specified `explicit_namespace` here so that later on we could remove all traces of it (when v1 would be retired, as `library/` is v1-only thing). 

I don't mind removing it, but I'd rather keep it so that in the future the behaviour of this won't change if `get_repo` is changed

rcerven: ok

lcarva:
Updated release notes draft:
> - `group_manifests`: include namespace if its present in the results

PR #884, opened 2017-11-17
Speed up some unit tests

twaugh:
When mocking docker API failures, disable retries.


twaugh:
Before: e.g. Total time 43 min 8 sec
After: e.g. Total time 32 min 24 sec

lcarva:
32 minutes still seems super high!

twaugh:
That's the sum of runtimes for each environment, and they generally run in parallel. Perhaps a better comparison is minimum runtime per environment: 5m7s (before) vs 3m46s (after) for the runs linked above.

twaugh:

PR #886, opened 2017-11-21
Fix sending the OAuth2 token to ODCS

puiterwijk:
Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>

### DIFF
--- a/atomic_reactor/odcs_util.py
+++ b/atomic_reactor/odcs_util.py
@@ -17,8 +17,8 @@ logger = logging.getLogger(__name__)
 
 class ODCSClient(object):
 
-    OIDC_TOKEN_HEADER = 'OIDC_access_token'
-    OIDC_CLAIM_SCOPE_HEADER = 'OIDC_CLAIM_scope'
+    OIDC_TOKEN_HEADER = 'Authorization'
+    OIDC_TOKEN_TYPE = 'Bearer'
 
     def __init__(self, url, insecure=False, token=None):
         if url.endswith('/'):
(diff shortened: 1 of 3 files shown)


puiterwijk:
`OIDC_access_token` and `OIDC_CLAIM_scope` are environment variables set by mod_auth_openidc, when the client sends a token as a Bearer token.

lcarva:
@puiterwijk, thanks for this!

We have recently moved the ODCSClient class to a new file `odcs_utils.py`.
Think you could rebase?

puiterwijk:
@lcarva Done! And that has must been very recent then, I think :).

lcarva:
This looks good to me, but I have no way of validating these changes work. That was also the case for the previous implementation btw.

The project's README still mentions setting the scopes: [Using OpenIDC for authentication](https://pagure.io/odcs)

Is that just bad documentation?

puiterwijk:
@lcarva no, the documentation is correct, because they are passing that list of scopes to a python client that is used to actually get a token from the Identity Provider. That client does need the list of scopes so it knows what kind of token to request, but if you look at the way the ODCSClient is initialized, it does not get the list of scopes, because ODCS gets that list from the IdP:
````
odcs = ODCS("https://odcs.fedoraproject.org",
            auth_mech=AuthMech.OpenIDC,
            openidc_token="your_openidc_token")
````

lcarva:
@puiterwijk, we're having issues with osbs-bot, any chance you can amend and push the commit to have it run again?

lcarva:
I'm going to merge this because this is blocking my work in adding ssl auth support.

lcarva:
Updated release notes draft:
> - Fix sending the OAuth2 token to ODCS

PR #887, opened 2017-11-21
build_image for worker will be set same as on orchestrator

rcerven:

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -27,7 +27,7 @@ from atomic_reactor.build import BuildResult
 from atomic_reactor.plugin import BuildStepPlugin
 from atomic_reactor.plugins.pre_reactor_config import get_config
 from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
-from atomic_reactor.util import get_preferred_label, df_parser
+from atomic_reactor.util import get_preferred_label, df_parser, get_build_json
 from atomic_reactor.constants import PLUGIN_ADD_FILESYSTEM_KEY, PLUGIN_BUILD_ORCHESTRATE_KEY
 from osbs.api import OSBS
 from osbs.exceptions import OsbsException
@@ -252,7 +252,7 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
         :param build_kwargs: dict, keyword arguments for starting worker builds
         :param osbs_client_config: str, path to directory containing osbs.conf
         :param worker_build_image: str, the builder image to use for worker builds
-                                  (deprecated, use config_kwargs instead)
+                                  (not used, image is inherited from the orchestrator)
         :param config_kwargs: dict, keyword arguments to override worker configuration
         :param find_cluster_retry_delay: the delay in seconds to try again reaching a cluster
         :param failure_retry_delay: the delay in seconds to try again starting a build
@@ -272,8 +272,7 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
         self.release = self.get_release()
 
         if worker_build_image:
-            self.log.warning('worker_build_image is deprecated, use config_kwargs instead')
-            self.config_kwargs.setdefault('build_image', worker_build_image)
+            self.log.warning('worker_build_image is deprecated')
 
         self.worker_builds = []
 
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -512,7 +511,15 @@ def select_and_start_cluster(self, platform):
                     ctx.try_again_later(self.failure_retry_delay)
                     # this will put the cluster in retry-wait when get_clusters runs
 
+    def set_build_image(self):
+        """
+        Overrides build_image for worker, to be same as in orchestrator build
+        """
+        spec = get_build_json().get("spec", {})
+        self.config_kwargs['build_image'] = spec['strategy']['customStrategy']['from']['name']
```
lcarva: This seems to work as expected, but let's make this easier to debug if something goes wrong.
Verify 'kind' is 'DockerImage'. If it's not, it means OpenShift didn't convert the ImageStreamTag into a docker image reference as we expect. In this case, let's raise an error as I don't think there's a sane way to recover from it.

vrutkovs: `.get()` should be used to get items of other arrays as well, so that a nicer exception could be thrown

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -252,7 +252,7 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
         :param build_kwargs: dict, keyword arguments for starting worker builds
         :param osbs_client_config: str, path to directory containing osbs.conf
         :param worker_build_image: str, the builder image to use for worker builds
-                                  (deprecated, use config_kwargs instead)
+                                  (deprecated)
```
vrutkovs: A better description would be `not used, image is inherited from the orchestrator)`

tests/plugins/test_orchestrate_build.py
```diff
@@ -235,12 +238,11 @@ def test_orchestrate_build(tmpdir, caplog, config_kwargs, worker_build_image, lo
         'conf_section': 'worker_x86_64',
         'conf_file': tmpdir + '/osbs.conf',
     }
-    if worker_build_image:
```
vrutkovs: The test should have `worker_build_image` param removed - and instead verify that `config_kwargs['build_image']` is used as a worker image instead

tests/plugins/test_orchestrate_build.py
```diff
@@ -235,12 +238,11 @@ def test_orchestrate_build(tmpdir, caplog, config_kwargs, worker_build_image, lo
...
-        expected_kwargs['build_image'] = worker_build_image
     # Update with config_kwargs last to ensure that, when set
     # always has precedence over worker_build_image param.
     if config_kwargs is not None:
         expected_kwargs.update(config_kwargs)
+    expected_kwargs['build_image'] = 'fedora:latest'
```
vrutkovs: Hardcoding this is a bad idead

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: .get() will not throw anything, or are you suggesting to, use get on every key, and check if it isn't None?

rcerven: will do check for 'DockerImage'

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: ok

tests/plugins/test_orchestrate_build.py
rcerven: I can remove worke_build_image, but I am keeping it just because we have it in orchestrator_build, even though not used,
test is verifying that value from environ['BUILD'] (orchestrator image) is used as it should,
as config_kwargs['build_image"] is being now overwritten by that

tests/plugins/test_orchestrate_build.py
rcerven: so should I create constant and replace every occurrence in the file with that?

vrutkovs: I'm not sure why this needs to be set at all - `config_kwargs` should update this. A case where `config_kwargs` is empty is not valid for this change anymore and should be removed

tests/plugins/test_orchestrate_build.py
vrutkovs: I don't see much value in testing deprecated params, as it makes the test a bit more confusion, but I don't mind keeping it

atomic_reactor/plugins/build_orchestrate_build.py
vrutkovs: Since there doesn't seem to be a nice way of chaining `.get()` calls, lets wrap this function in `try..catch KeyError` and throw `RuntimeError` with a better description - e.g. `Build object is malformed, failed to fetch buildroot image`

tests/plugins/test_orchestrate_build.py
rcerven: at line 244 expected_kwargs are updated with config_kwargs, which might set build_image (from osbs which functionality should remain same as per story description), but this change will be using orchestrators build_image, which is mocked in mock_workflow, to this 'fedora:latest'

vrutkovs: Oh, I see, `fedora:latest` in `worker_build_image` has confused me.

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: @rcerven, do you plan to incorporate this suggestion?

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: This hasn't been addressed yet.

rcerven: it will be

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: yes

lcarva:
Updated release notes draft:
> - `build_image` for worker is set based on build image in orchestrator OpenShift build

PR #888, opened 2017-11-22
Support SSL Auth in ODCS from resolve_composes

lcarva:
As well as a more concise way of setting headers
and verify flag for requests.

### DIFF
--- a/tests/plugins/test_resolve_composes.py
+++ b/tests/plugins/test_resolve_composes.py
@@ -201,6 +201,44 @@ class TestResolveComposes(object):
         self.run_plugin_with_args(workflow, plugin_args,
                                   expect_error='cannot be used at the same time')
 
+    @pytest.mark.parametrize(('plugin_args', 'expected_kwargs'), (
+        ({
+            'odcs_insecure': True,
+        }, {'insecure': True}),
+
+        ({
+            'odcs_insecure': False,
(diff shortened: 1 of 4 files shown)


tests/plugins/test_resolve_composes.py
```diff
@@ -201,6 +201,43 @@ def test_signing_intent_and_compose_ids_mutex(self, workflow):
...
+            workflow._tmpdir.join('token').write('the-token')
+            plugin_args['odcs_openidc_secret_path'] = str(workflow._tmpdir)
+
+        if plugin_args.get('odcs_ssl_secret_path') is True:
+            workflow._tmpdir.join('cert').write('the-cert')
+            plugin_args['odcs_ssl_secret_path'] = str(workflow._tmpdir)
+            expected_kwargs['cert'] = str(workflow._tmpdir.join('cert'))
+
+        (flexmock(ODCSClient)
+            .should_call('__new__')
```
vrutkovs: This seems to cause the issue in Py3 - why not `__init__`?

lcarva: Yup, that was it. Changed it to `should_receive('__init__')`
Thank you!

twaugh:
Not updating release notes as this is a change to an unreleased feature.

PR #889, opened 2017-11-23
odcs related metadata in koji build

rcerven:

### DIFF
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -40,7 +40,8 @@ from atomic_reactor.constants import (PLUGIN_KOJI_IMPORT_PLUGIN_KEY,
                                       PLUGIN_PULP_SYNC_KEY,
                                       PLUGIN_FETCH_WORKER_METADATA_KEY,
                                       PLUGIN_GROUP_MANIFESTS_KEY,
-                                      PLUGIN_KOJI_PARENT_KEY)
+                                      PLUGIN_KOJI_PARENT_KEY,
+                                      PLUGIN_RESOLVE_COMPOSES_KEY)
 from atomic_reactor.util import (get_build_json, get_preferred_label,
                                  df_parser, ImageName, get_checksums, get_primary_images,
                                  get_manifest_media_type,
@@ -427,6 +428,14 @@ class KojiImportPlugin(ExitPlugin):
         else:
             koji_task_owner = None
         extra['submitter'] = self.session.getLoggedInUser()['name']
+
+        resolve_comp_result = self.workflow.prebuild_results.get(PLUGIN_RESOLVE_COMPOSES_KEY)
+        if resolve_comp_result:
+            extra['image']['odcs'] = {
+                'compose_ids': [item['id'] for item in resolve_comp_result['composes']],
+                'signing_intent': resolve_comp_result['signing_intent'],
+                'signing_intent_overridden': resolve_comp_result['signing_intent_overridden'],
+            }
 
         self.set_help(extra, worker_metadatas)
         self.set_media_types(extra, worker_metadatas)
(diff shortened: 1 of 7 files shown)


tests/plugins/test_koji_import.py
```diff
@@ -1841,3 +1842,55 @@ def test_koji_import_primary_images(self, tmpdir, os_env, add_tag_conf_primaries
...
+        if comp is not None and sign_int is not None and override is not None:
+            resolve_comp_entry = True
+
+            workflow.prebuild_results[PLUGIN_RESOLVE_COMPOSES_KEY] = {
+                'composes': comp,
+                'signing_intent': sign_int,
+                'signing_intent_overridden': override,
+            }
+        else:
+            workflow.prebuild_results[PLUGIN_RESOLVE_COMPOSES_KEY] = None
```
vrutkovs: Tests shouldn't set results to None in this case - as plugin might not run at all

rcerven: ahh right, I can remove that

tests/plugins/test_koji_import.py
```diff
@@ -1841,3 +1842,55 @@ def test_koji_import_primary_images(self, tmpdir, os_env, add_tag_conf_primaries
...
+    def test_odcs_metadata_koji(self, tmpdir, os_env, comp, sign_int, override):
+        session = MockedClientSession('')
+        tasker, workflow = mock_environment(tmpdir,
+                                            name='ns/name',
+                                            version='1.0',
+                                            release='1',
+                                            session=session)
+
+        resolve_comp_entry = False
+        if comp is not None and sign_int is not None and override is not None:
```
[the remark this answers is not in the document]
rcerven: that won't work, I want to check only if all are None

all will return true also if all are True

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -428,6 +429,14 @@ def get_build(self, metadata, worker_metadatas):
             koji_task_owner = None
         extra['submitter'] = self.session.getLoggedInUser()['name']
 
+        resolve_comp_result = self.workflow.prebuild_results.get(PLUGIN_RESOLVE_COMPOSES_KEY)
+        if resolve_comp_result:
+            extra['image']['odcs'] = {
+                'compose_ids': resolve_comp_result['composes'],
```
twaugh: The `composes` key has a value that is a list of dicts, the results from ODCS. The compose *ID*s are available via the `id` key in each of those dicts.

So this should probably be:
```
'compose_ids': [info['id'] for info in resolve_comp_result['composes']],
```

tests/plugins/test_koji_import.py
twaugh: We should test both cases.

atomic_reactor/plugins/exit_koji_import.py
rcerven: ahh ok will change

tests/plugins/test_koji_import.py
rcerven: ok

rcerven:
updated as well koji.md

tests/plugins/test_koji_import.py
```diff
@@ -1841,3 +1842,62 @@ def test_koji_import_primary_images(self, tmpdir, os_env, add_tag_conf_primaries
             return
 
         runner.run()
+
+    @pytest.mark.parametrize('resolve_run', [
+        True,
+        False,
```
twaugh: This works fine, but it's not very efficient because it ends up trying all of the 'comp' values, for instance, even when the plugin is not run in the scenario.

rcerven: made another test for that, so it is more efficient

rcerven:
rn updated:

- Koji build metadata now includes ODCS related information

docs/koji.md
```diff
@@ -50,6 +50,10 @@ Data which is placed here includes:
 - `build.extra.image.flatpak` (boolean): true if this image is a Flatpak
 - `build.extra.image.modules` (boolean): the [modules](https://docs.pagure.org/modularity/) that provide the packages for this image, resolved to `NAME-STREAM-VERSION`. This will include the resolved versions of the modules in `source_modules`, and the dependencies of those modules. (Currently only for Flatpaks)
 - `build.extra.image.source_modules` (boolean): the modules that were specified as input to the build process. (Currently only for Flatpaks)
+- `build.extra.image.odcs` (map): information about ODCS
+- `build.extra.image.odcs.compose_ids` (int list): List of each ODCS compose used
+- `build.extra.image.odcs.signing_intent` (str): Final signing intent of the ODCS composes after adjusting for CLI parameter
+- `build.extra.image.odcs.signing_intent.overridden` (boolean): Whether or not the signing intent used is different than the one defined in container.yaml
```
lcarva: Minor typo, should be `signing_intent_overridden`, not `signing_intent.overridden`.

rcerven: fix in https://github.com/projectatomic/atomic-reactor/pull/890https://github.com/projectatomic/atomic-reactor/pull/890

PR #891, opened 2017-11-30
tests: install specific versions of test package to avoid test packages update breaks

vrutkovs:

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -20,7 +20,6 @@ if [[ $OS == "fedora" ]]; then
   PKG_EXTRA="dnf-plugins-core desktop-file-utils flatpak ostree"
   BUILDDEP="dnf builddep"
   PYTHON="python$PYTHON_VERSION"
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -78,8 +76,7 @@ $RUN $PYTHON setup.py install
 
 # Install packages for tests
 $RUN $PIP install -r tests/requirements.txt
+if [[ $OS != "fedora" ]]; then $RUN $PIP install -U setuptools; fi
```
lcarva: Why this line?

vrutkovs: CentOS has some ancient setuptools installed, which [breaks](https://stackoverflow.com/questions/26589990/py-test-error-unrecognized-arguments-cov-ner-brands-cov-report-term-missi/41066450#41066450) `pytest` and `pytest-cov` integration.

lcarva: Let's add a comment then, please.

vrutkovs: Right, done

PR #892, opened 2017-12-04
pull_base_image: set original base image in annotations

### LINKED ISSUE #881: Race condition when pulling parent image
In order to avoid a race condition when pulling the parent image, we tag the image with the name of the OpenShift build:
https://github.com/projectatomic/atomic-reactor/blob/1.6.28/atomic_reactor/plugins/pre_pull_base_image.py#L88-L94

However,  we don't use this as the name of the base image in the workflow:
https://github.com/projectatomic/atomic-reactor/blob/1.6.28/atomic_reactor/plugins/pre_pull_base_image.py#L121

Because of this, inspecting the base image later may fail, for example here:
https://github.com/projectatomic/atomic-reactor/blob/1.6.28/atomic_reactor/plugins/pre_inject_yum_repo.py#L158

This is because, meanwhile, some other build using the same parent image may have deleted it:
https://github.com/projectatomic/atomic-reactor/blob/1.6.28/atomic_reactor/plugins/exit_remove_built_image.py#L49-L50

I think the fix is for the pull_base_image plugin to have this line changed:
https://github.com/projectatomic/atomic-reactor/blob/1.6.28/atomic_reactor/plugins/pre_pull_base_image.py#L121
to something like:
```
self.workflow.builder.set_base_image(buildid_base_image.to_str())
```

and to remove these unnecessary lines:
https://github.com/projectatomic/atomic-reactor/blob/1.6.28/atomic_reactor/plugins/pre_pull_base_image.py#L115-L119

vrutkovs:
Use unique ID in other cases to avoid race condition

Fixes #881

### DIFF
--- a/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
+++ b/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
@@ -201,7 +201,7 @@ class StoreMetadataInOSv3Plugin(ExitPlugin):
         except AttributeError:
             commit_id = ""
 
-        base_image = self.workflow.builder.base_image
+        base_image = self.workflow.builder.original_base_image
         if base_image is not None:
             base_image_name = base_image.to_str()
             try:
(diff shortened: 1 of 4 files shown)


vrutkovs:
Release notes updated

PR #893, opened 2017-12-07
store_metadata: handle base image builds correctly

vrutkovs:
Previos PR was failing on base image builds (as these don't set `original_base_image`) -
and the test was setting it incorrectly.


lcarva:

PR #894, opened 2017-12-12
force user for all_rpm_packages plugin

rcerven:


rcerven:
release note updated: 
- `all_rpm_packages`: forces root user

PR #897, opened 2017-12-13
missing required exit plugin will cause build failure

rcerven:


lcarva:
Can you provide an explanation of why this is needed?

rcerven:
added explanation into commit message

rcerven:
release note updated:
- exit plugins marked as required but is missing, will fail build

PR #899, opened 2018-01-09
Extend default odcs timeout to 30 minutes

lcarva:
A medium sized compose took about 7 minutes to complete.
After consulting with ODCS team, it's been suggested that
30 mins should be more than sufficient for OSBS builds.


twaugh:
There seems to be something up with the CentOS 7 test run.

lcarva:
Re-ran the centos build, passed. Failures are on Rawhide now.

lcarva:
@twaugh fixed centos run, ptal.

lcarva:

PR #900, opened 2018-01-10
Add support for module compose requests

lcarva:


lcarva:

PR #901, opened 2018-01-10
Disable rawhide testing

lcarva:
There's been some Fedora rawhide issues lately that
prent unit tests from starting in rawhide. Let's
disable testing there since it's not expected to
always be stable.


lcarva:

PR #902, opened 2018-01-11
Add digests for prod-type OSBS build to koji build archive extra metadata

rcerven:

### DIFF
--- a/atomic_reactor/plugins/exit_koji_promote.py
+++ b/atomic_reactor/plugins/exit_koji_promote.py
@@ -50,7 +50,7 @@ from atomic_reactor.util import (get_version_of_tools, get_checksums,
                                  get_docker_architecture, df_parser,
                                  are_plugins_in_order,
                                  get_image_upload_filename,
-                                 get_digests_map_from_annotations)
+                                 get_manifest_media_type)
 from atomic_reactor.koji_util import (create_koji_session, tag_koji_build,
                                       Output, KojiUploadLogger)
 from atomic_reactor.rpm_util import parse_rpm_output, rpm_qf_args
@@ -402,6 +402,21 @@ class KojiPromotePlugin(ExitPlugin):
 
         return output_images
 
+    def get_digests_with_types(self):
+        digests_with_types = {}
+        digests = {}
+        image = self.workflow.tag_conf.unique_images[0]
+        tag = image.to_str(registry=False)
+        for registry in self.workflow.push_conf.docker_registries:
+            if tag in registry.digests:
+                digests.update(registry.digests[tag])
+
+        for version in digests:
+            media_type = get_manifest_media_type(version)
+            digests_with_types[media_type] = digests[version]
+
+        return digests_with_types
+
     def get_output(self, buildroot_id):
         """
         Build the 'output' section of the metadata.
@@ -460,10 +475,10 @@ class KojiPromotePlugin(ExitPlugin):
                 },
             },
         })
-        annotations = self.workflow.build_result.annotations
-        if annotations and 'digests' in annotations:
-            digests = get_digests_map_from_annotations(annotations['digests'])
-            metadata['extra']['docker']['digests'] = digests
+
+        digests_with_types = self.get_digests_with_types()
+        if digests_with_types:
+            metadata['extra']['docker']['digests'] = digests_with_types
 
         if not config:
             del metadata['extra']['docker']['config']
--- a/tests/plugins/test_koji_promote.py
+++ b/tests/plugins/test_koji_promote.py
@@ -10,12 +10,12 @@ from __future__ import unicode_literals
 
 import json
 import os
+import sys
 
 try:
     import koji
 except ImportError:
     import inspect
-    import sys
 
     # Find our mocked koji module
     import tests.koji as koji
@@ -197,8 +197,7 @@ def mock_environment(tmpdir, session=None, name=None,
                      source=None, build_process_failed=False,
                      is_rebuild=True, docker_registry=True,
                      pulp_registries=0, pulp_supports_schema2=False,
-                     digest=None, digest_version=None,
-                     blocksize=None,
+                     registry_digests=None, blocksize=None,
                      task_states=None, additional_tags=None,
                      has_config=None,
                      logs_return_bytes=True):
@@ -267,13 +266,19 @@ def mock_environment(tmpdir, session=None, name=None,
         sync_result = {}
         for image in workflow.tag_conf.images:
             tag = image.to_str(registry=False)
-            digests = ManifestDigest(v1=fake_digest(image, 1),
-                                     v2=fake_digest(image, 2))
-            docker_reg.digests[tag] = digests
+            man_fake = ManifestDigest(v1=fake_digest(image, 1),
+                                      v2=fake_digest(image, 2))
+
+            if registry_digests:
+                man_digests = ManifestDigest(**registry_digests)
+            else:
+                man_digests = man_fake
+
+            docker_reg.digests[tag] = man_digests
             if pulp_registries:
-                sync_result[digests.v1] = {}
+                sync_result[man_fake.v1] = {}
                 if pulp_supports_schema2:
-                    sync_result[digests.v2] = {}
+                    sync_result[man_digests.v2] = {}
 
             if has_config:
                 docker_reg.config = {
@@ -293,11 +298,6 @@ def mock_environment(tmpdir, session=None, name=None,
                                                        'type': IMAGE_TYPE_DOCKER_ARCHIVE}])
 
     annotations = {}
-    if digest and digest_version:
-        annotations['digests'] = json.dumps([{
-            'digest': digest,
-            'version': digest_version
-        }])
 
     if build_process_failed:
         workflow.build_result = BuildResult(logs=["docker build log - \u2018 \u2017 \u2019 \n'"],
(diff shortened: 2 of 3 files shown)


twaugh:
Can we have a few more words in the commit message? ;-)

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -460,10 +460,20 @@ def add_log_type(output):
                 },
             },
         })
-        annotations = self.workflow.build_result.annotations
-        if annotations and 'digests' in annotations:
-            digests = get_digests_map_from_annotations(annotations['digests'])
```
lcarva: Right, because whenever `koji_prmote` runs it's either in worker build (arrangement >= 2), or in prod build.
In both cases, build_result.annotations is not set.

rcerven:
commit message updated :)

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -460,10 +460,20 @@ def add_log_type(output):
...
-            metadata['extra']['docker']['digests'] = digests
+
+        if self.workflow.push_conf.docker_registries:
+            if self.workflow.tag_conf.unique_images:
+                image = self.workflow.tag_conf.unique_images[0]
```
lcarva: I'm trying to work out how this works. 

For arrangement >= 3:
1. In worker build, the 'digests' annotation is added to the OpenShift worker build via `store_metadata_in_osv3` plugin.
2. In orchestrator build, the `orchestrate_build` plugin aggregates the annotation from each worker build under the `worker-builds` annotation.
3. `koji_import` uses the `digests` annotation of each worker build to create the aggregated desired digests map.

For prod builds, I think we want to iterate over digests the same way that `store_metadata_in_osv3` does. Shouldn't we then loop for each `docker_registry` and `each tag_conf.images`?

lcarva:
> commit message updated :)

I don't wanna split hairs or anything, but this patch is not necessarily target for ppc64le. It's simply adding digests metadata to koji build for prod-type OSBS builds.

atomic_reactor/plugins/exit_koji_promote.py
rcerven: right

atomic_reactor/plugins/exit_koji_promote.py
rcerven: ok I'll iterate through each docker_registry.
even though right now there is just one registry, it is possible to have more, and even though the digests for same tag would be same, it is possible that each registry will provide different set of digest versions, so to collect all versions I'll loop

rcerven: as for looping through images, as of now we have only 1 unique tag, and digests as of now contains digests for 1 tag only

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -460,10 +460,24 @@ def add_log_type(output):
...
+                tag = image.to_str(registry=False)
+                digests = {}
+
+                for registry in self.workflow.push_conf.docker_registries:
+                    if tag in registry.digests:
```
lcarva: Looking at `get_digests` in `store_metadata_in_osv3`, it seems that the keys in `registry.digests` contains the registry name.

```
        digests = {}  # repository -> digest
                image_str = image.to_str()
                    digest = registry.digests[image_str]
                    digests[image.to_str(registry=False)] = digest

        return digests
```

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -460,10 +460,24 @@ def add_log_type(output):
                 },
             },
         })
-        annotations = self.workflow.build_result.annotations
-        if annotations and 'digests' in annotations:
-            digests = get_digests_map_from_annotations(annotations['digests'])
-            metadata['extra']['docker']['digests'] = digests
+
+        if self.workflow.push_conf.docker_registries:
```
lcarva: I'd really like to see this whole block in its own method for enhanced readability.

atomic_reactor/plugins/exit_koji_promote.py
rcerven: looking at this they are tags:

class DockerRegistry(Registry):
    """ v1/v2 docker registry """
    def __init__(self, uri, insecure=False):
        """
        :param uri: str, uri for pushing/pulling
        :param insecure: bool
        """
        super(DockerRegistry, self).__init__(uri, insecure=insecure)
        self.digests = {} # maps tags (str) to their digest, if available

rcerven: name of registry is in self.name from 'Registry'

atomic_reactor/plugins/exit_koji_promote.py
rcerven: ok will do that

atomic_reactor/plugins/exit_koji_promote.py
twaugh: Yes, keys in `DockerRegistry.digests` are tags, i.e. pull specifications with no registry and without '@'; the result of Imagename.to_str(registry=False).

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -402,6 +402,25 @@ def get_repositories(self, digests):
...
+    def get_digests_with_types(self):
+        digests_with_types = {}
+
+        if self.workflow.push_conf.docker_registries:
+            if self.workflow.tag_conf.unique_images:
+                image = self.workflow.tag_conf.unique_images[0]
+                tag = image.to_str(registry=False)
+                digests = {}
+
+                for registry in self.workflow.push_conf.docker_registries:
```
twaugh: This line (in fact to the end of this block) looks to be indented too far. Don't we want this to run even if `self.workflow.tag_conf.unique_images` is falsy?

In fact at this point I think there will always be a unique image because tag_and_push makes sure of it.

Given that, I wonder whether this should be re-written as:
```
digests_with_types = {}
digests = {}
image = self.workflow.tag_conf.unique_images[0]
tag = image.to_str(registry=false)
    ....
....
```

atomic_reactor/plugins/exit_koji_promote.py
twaugh: Actually, what would be a great improvement would be to change that comment to make the code easier to follow:
```
-    self.digests = {} # maps tags (str) to their digest, if available
+    self.digests = {} # maps a tag (str) to a ManifestDigest instance, if available
```

tests/plugins/test_koji_promote.py
```diff
@@ -267,13 +266,19 @@ def mock_environment(tmpdir, session=None, name=None,
...
+
+            if digests:
+                man_digests = ManifestDigest(**digests)
+            else:
+                man_digests = man_fake
+
+            docker_reg.digests[tag] = man_digests
             if pulp_registries:
-                sync_result[digests.v1] = {}
+                sync_result[man_fake.v1] = {}
```
twaugh: Not sure why this isn't `sync_result[man_digests.v1] = {}`. Is it just that we don't test with pulp_registries as well as digests?

tests/plugins/test_koji_promote.py
```diff
@@ -267,13 +266,19 @@ def mock_environment(tmpdir, session=None, name=None,
...
                 if pulp_supports_schema2:
-                    sync_result[digests.v2] = {}
+                    sync_result[man_fake.v2] = {}
```
twaugh: Same comment as above, this looks to me like it should be `sync_result[man_digests.v2] = {}`.

atomic_reactor/plugins/exit_koji_promote.py
rcerven: ok will change that

atomic_reactor/plugins/exit_koji_promote.py
rcerven: ok will do that

tests/plugins/test_koji_promote.py
rcerven: well we want to feed pulp registry always with digests, but man_digests are only meant for docker registries, so it might be that man_digests doesn't have that type of digest, so I have to use the fake one

tests/plugins/test_koji_promote.py
rcerven: same as reply above

tests/plugins/test_koji_promote.py
twaugh: Ah, the distiction is that 'digests' is for the digests available in the docker registry, but 'sync_result' carries the digests available in Pulp after sync.

Can we make the naming reflect that more clearly? Maybe `registry_digests` instead of `digests` for the mock_environment parameter name?

Then I think we always want the v2 schema 2 digest to be identical in the registry and in Pulp. Only the v2 schema 1 digest might differ.

So I think *this* line is fine, just the variable names might be improved.

tests/plugins/test_koji_promote.py
twaugh: I think *this* line should be `man_digests.v2` (with improved variable naming), because the v2 schema 2 manifest digest in the registry will always match what gets synced to Pulp.

tests/plugins/test_koji_promote.py
rcerven: ok will rename that to 'registry_digests'

tests/plugins/test_koji_promote.py
rcerven: ok will do

rcerven:
rn updated:

Add digests for prod-type OSBS build to koji build archive extra metadata

PR #904, opened 2018-01-22
Update labels from check_and_set_rebuild

lcarva:
Using the method set_labels_in_build_config actually overwrites
all the labels. This is not desired. Use the update version
to avoid destroying other labels.


lcarva:
Updated release notes draft:
> Update labels from check_and_set_rebuild

PR #906, opened 2018-02-01
Add multi arch support for failing test

cverna:
The test_koji_upload_success is failling on non x86_64 arch, this commit add the local platform to the runner so that it supports all arch.


twaugh:
Thanks!

Hmm, wonder why the Travis CI tests are failing now. Seems unrelated to the code change.

twaugh:

PR #907, opened 2018-02-01
cleanup flatpaks

rcerven:

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_module_compose.py
+++ b/atomic_reactor/plugins/pre_resolve_module_compose.py
@@ -25,11 +25,14 @@ Example configuration:
 
 import os
 import re
+import yaml
 from modulemd import ModuleMetadata
 from pdc_client import PDCClient
 
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.odcs_util import ODCSClient
+from atomic_reactor.util import split_module_spec
+from atomic_reactor.constants import REPO_CONTAINER_CONFIG
 
 
 class ModuleInfo(object):
@@ -82,8 +85,7 @@ class ResolveModuleComposePlugin(PreBuildPlugin):
     is_allowed_to_fail = False
 
     def __init__(self, tasker, workflow,
-                 module_name, module_stream, module_version=None,
-                 compose_id=None,
+                 compose_ids=tuple(),
                  odcs_url=None, odcs_insecure=False,
                  odcs_openidc_secret_path=None,
                  pdc_url=None, pdc_insecure=False):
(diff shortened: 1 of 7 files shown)


atomic_reactor/plugins/pre_resolve_module_compose.py
```diff
@@ -137,9 +142,25 @@ def _resolve_compose(self):
...
+        self.read_configs_general()
+
+        modules = self.data.get('modules', [])
+        if modules:
+            source_spec = modules[0]
+            if len(modules) > 1:
+                self.log.info("compose config contains multiple modules,"
+                              "using first module %s", source_spec)
+        else:
+            raise RuntimeError('"compose" config is missing "modules", required for Flatpak')
```
twaugh: Could rephrase this like:
```
if not modules:
    raise RuntimeError(...)

source_spec = modules[0]
...
```

to avoid unnecessary indentation.

atomic_reactor/plugins/pre_resolve_module_compose.py
```diff
@@ -137,9 +142,25 @@ def _resolve_compose(self):
...
+
+        mod_name, mod_stream, mod_version = split_module_spec(source_spec)
```
twaugh: Can we call these `module_name`, `module_stream` etc for clarity (as in the original code)?

atomic_reactor/util.py
```diff
@@ -1125,3 +1125,14 @@ def get_primary_images(workflow):
...
             workflow.build_result.annotations['repositories']['primary']]
     return primary_images
+
+
+def split_module_spec(module):
+    pieces = module.split('-')
+    if len(pieces) == 2:
+        return pieces[0], pieces[1], None
+    elif len(pieces) == 3:
+        return pieces[0], pieces[1], pieces[2]
```
twaugh: Or just `return pieces`

atomic_reactor/plugins/pre_resolve_module_compose.py
rcerven: ok

atomic_reactor/plugins/pre_resolve_module_compose.py
rcerven: ok as discussed will rename these, and loop bellow with conflicting names, will put into its own method

atomic_reactor/util.py

atomic_reactor/util.py
```diff
@@ -1125,3 +1125,14 @@ def get_primary_images(workflow):
             ImageName.parse(primary) for primary in
             workflow.build_result.annotations['repositories']['primary']]
     return primary_images
+
+
+def split_module_spec(module):
+    pieces = module.split('-')
```
lcarva: The first portion of the module spec can actually contain dashes, right?
We should use rsplit, but! I don't think that's safe either... (my-module-3.0 will be incorrectly parsed as name=my, version=module, stream=3.0)

ODCS has recently changed to use ":" separators instead (https://pagure.io/odcs/pull-request/157). But that's not deployed yet. How about if we

1. If `module` contains `":"`, use that as the splitter
2. Otherwise, splitter is "-", but we log a warning that "module name with dashes may be incorrectly parsed"

Wdyt?

lcarva: Actually, we should combine what's already done in `_resolve_modules` by regex. We can leave the change from `"-"` to another story `":"`.

rcerven: ok will use that regex

rcerven:
rn updated:

- Removal of parameters: 'module_name', 'module_stream', 'module_version', 'compose_id'
  from plugin 'resolve_module_compose', added: 'compose_ids'

PR #908, opened 2018-02-02
Remove python-docker RPMs prior to installing docker-squash PyPI package

twaugh:
Mixing the old-style (1.x) and new-style (2.x) Python docker modules results in errors when collecting tests. To avoid this, make sure to remove the RPMs before installing docker-squash.

This leaves installing the Python docker module to the docker-squash requirement for it, avoiding RPM and PyPI versions overwriting each other.

In addition, on CentOS 7 allowing installing docker-squash to install the docker Python bindings will fail because of the way in which it specifies the package. To fix this,
update the setuptools package using pip first.


twaugh:

PR #909, opened 2018-02-05
Fix test URL for flatpak git repo

twaugh:
Use git: instead of http: for this test URL.

Ideally the tests would be self-contained and not require any network access, but until then let's at least keep them working.

```
$ curl http://pkgs.fedoraproject.org/modules/eog.git
curl: (7) Failed to connect to pkgs.fedoraproject.org port 80: No route to host
```


twaugh:

PR #912, opened 2018-02-07
logging: re-use existing logs in pulp

mlangsdorf:
Call setup_logger once, from pulp_util, and have all other plugins
that want to use the pulp logger import pulp_util.

Fixes the issue of repeated pulp log entries.

### DIFF
--- a/atomic_reactor/pulp_util.py
+++ b/atomic_reactor/pulp_util.py
@@ -10,17 +10,16 @@ from __future__ import print_function, unicode_literals
 
 import os
 import re
-import logging
 import warnings
 from collections import namedtuple
 
 try:
     import dockpulp
-    from dockpulp import setup_logger
+
 except (ImportError, SyntaxError):
     dockpulp = None
+    import logging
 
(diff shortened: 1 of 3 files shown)


atomic_reactor/pulp_util.py
```diff
@@ -10,19 +10,16 @@
...
 try:
     import dockpulp
-    from dockpulp import setup_logger
 except (ImportError, SyntaxError):
     dockpulp = None
 
-logger = logging.getLogger(__name__)
-
 PulpRepo = namedtuple('PulpRepo', ['registry_id', 'tags'])
+logger = dockpulp.setup_logger(dockpulp.log)
```
twaugh: We probably need to move this into the `try` block above to handle the Python 3 case (dockpulp is not yet ported to Python 3, so importing it fails there).

mlangsdorf: Fixed, though I had to move it somewhere else to get it to work.

twaugh:
Updated draft release notes.

PR #915, opened 2018-02-14
Improvements to get_manifest_digests utility function

twaugh:
This function now gives more debugging output to aid diagnosis of a
situation in which only one of the two expected media types is seen on
the Pulp server.

As part of this, the function has been improved to aid readability.
There are now two more functions:
* `guess_manifest_media_type()` takes a bytes object and attempts to
decode it and interpret the JSON in order to decide which manifest media
type it is
* `manifest_is_media_type()` takes a requests.Response object and an
expected media type and attempts to confirm or deny that the manifest
returned in the response is of the expected media type

The media type guessing has also been improved. Previously in the
absence of a Content-Type HTTP response header, API v1 manifests could
not be identified at all (they lack a mediaType key). Now, if there is
no mediaType key the schemaVersion key will be inspected. This allows us
to reasonably guess at the manifest being API v1.

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -9,6 +9,7 @@ of the BSD license. See the LICENSE file for details.
 from __future__ import print_function, unicode_literals
 
 import hashlib
+from itertools import chain
 import json
 import jsonschema
 import os
@@ -754,9 +755,67 @@ def query_registry(registry_session, image, digest=None, version='v1', is_blob=F
     logger.debug("query_registry: querying {}, headers: {}".format(url, headers))
 
     response = registry_session.get(url, headers=headers)
+    for r in chain(response.history, [response]):
+        logger.debug("query_registry: [%s] %s", r.status_code, r.url)
+
+    logger.debug("query_registry: response headers: %s", response.headers)
     response.raise_for_status()
 
     return response
+
+
+def guess_manifest_media_type(content):
+    """
+    Guess the media type for the given manifest content
+
+    :param content: JSON content of manifest (bytes)
+    :return: media type (str), or None if unable to guess
+    """
+    encoding = guess_json_utf(content)
+    try:
+        manifest = json.loads(content.decode(encoding))
+    except (ValueError,           # Not valid JSON
+            TypeError,            # Not an object
+            UnicodeDecodeError):  # Unable to decode the bytes
+        logger.exception("Unable to decode JSON")
+        logger.debug("response content (%s): %r", encoding, content)
+        return None
+
+    try:
+        return manifest['mediaType']
+    except KeyError:
+        # no mediaType key
+        if manifest.get('schemaVersion') == 1:
+            return get_manifest_media_type('v1')
+
+        logger.warning("no mediaType or schemaVersion=1 in manifest, keys: %s",
+                       manifest.keys())
+
+
+def manifest_is_media_type(response, media_type):
+    """
+    Attempt to confirm the returned manifest is of a given media type
+
+    :param response: a requests.Response
+    :param media_type: media_type (str), or None to confirm
+        the media type cannot be guessed
+    """
+    try:
+        received_media_type = response.headers['Content-Type']
+    except KeyError:
+        # Guess media type from content
+        logger.debug("No Content-Type header; inspecting content")
+        received_media_type = guess_manifest_media_type(response.content)
+        logger.debug("guessed media type: %s", received_media_type)
+
+    if received_media_type is None:
+        return media_type is None
+
+    # Only compare prefix as response may use +prettyjws suffix
+    # which is the case for signed manifest
+    response_h_prefix = received_media_type.rsplit('+', 1)[0]
+    request_h_prefix = media_type.rsplit('+', 1)[0]
+    return response_h_prefix == request_h_prefix
 
 
 def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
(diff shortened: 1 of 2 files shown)


twaugh:
Needs additional tests for the new functions.

atomic_reactor/util.py
```diff
@@ -867,41 +907,21 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
...
-            logger.debug('response headers: %s', response.headers)
-            logger.warning('Received media type %s mismatches the expected %s',
-                           received_media_type, media_type)
-            continue
+        if version != 'v1':
+            if not manifest_is_media_type(response, media_type):
+                logger.error("content does not match expected media type")
+                continue
+            else:
+                logger.debug("content matches expected media type")
```
mlangsdorf: Nit: assign the return of manifest_is_media_type to a variable and print out the expected and actual values in the error message for simpler debugging. and put the expected media type in the logger.debug message for consistency.

twaugh: We don't have the returned/guess media type by this point, and we want to log at different levels depending on the result (a match is not an error), so I'm not sure how you'd like this to look.

mlangsdorf: Sorry, for some reason I thought manifest_is_media_type returned something other than a Boolean. Ignore the request.

atomic_reactor/util.py
```diff
@@ -808,11 +809,55 @@ def query_registry(registry_session, image, digest=None, version='v1', is_blob=F
...
     response.raise_for_status()
 
     return response
 
 
+def guess_manifest_media_type(content):
+    encoding = guess_json_utf(content)
+    try:
+        manifest = json.loads(content.decode(encoding))
+        return manifest['mediaType']
```
rcerven: maybe move this to it's own 'try' ? just to be sure that decode won't raise any KeyError

twaugh: Line 830 catches any KeyError from this.

rcerven: yes, but if decode raises KeyError it should different than KeyError raised from "return manifest['mediaType']"

twaugh: Ah, bytes.decode() can raise a UnicodeDecodeError exception, yes, good point. Changed.

atomic_reactor/util.py
```diff
@@ -808,11 +809,59 @@ def query_registry(registry_session, image, digest=None, version='v1', is_blob=F
     logger.debug("query_registry: querying {}, headers: {}".format(url, headers))
 
     response = registry_session.get(url, headers=headers)
+    for r in chain(reversed(response.history), [response]):
```
lcarva: Since `history` "is sorted from the oldest to the most recent request", does it make sense to first look at response, then through the reversed history list? Alternatively, first look at non-reversed history list and then response. The current iteration order is not sequential.

atomic_reactor/util.py
```diff
@@ -808,11 +809,59 @@ def query_registry(registry_session, image, digest=None, version='v1', is_blob=F
...
+            UnicodeDecodeError):  # Unable to decode the bytes
+        logger.exception("Unable to decode JSON")
+        logger.debug("response content (%s): %r", encoding, content)
+        return None
+
+    try:
+        return manifest['mediaType']
+    except KeyError:
+        # no mediaType key
+        if 'schemaVersion' in manifest and manifest['schemaVersion'] == 1:
```
lcarva: more "pythonic": `if manifest.get('schemaVersion') == 1`

atomic_reactor/util.py
```diff
@@ -808,11 +809,59 @@ def query_registry(registry_session, image, digest=None, version='v1', is_blob=F
...
+        logger.debug("guessed media type: %s", received_media_type)
+
+    if received_media_type is None:
+        return False
+
+    # Only compare prefix as response may use +prettyjws suffix
+    # which is the case for signed manifest
+    response_h_prefix = received_media_type.rsplit('+', 1)[0]
+    request_h_prefix = media_type.rsplit('+', 1)[0]
+    if response_h_prefix != request_h_prefix:
```
lcarva: Easier to read? `return response_h_prefix == request_h_prefix`

atomic_reactor/util.py
```diff
@@ -867,41 +916,21 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
...
-        # which is the case for signed manifest
-        response_h_prefix = received_media_type.rsplit('+', 1)[0]
-        request_h_prefix = media_type.rsplit('+', 1)[0]
-        if response_h_prefix != request_h_prefix:
-            logger.debug('request headers: %s', headers)
-            logger.debug('response headers: %s', response.headers)
-            logger.warning('Received media type %s mismatches the expected %s',
-                           received_media_type, media_type)
-            continue
+        if version != 'v1':
```
lcarva: Why the `v1` check here?

atomic_reactor/util.py
```diff
@@ -808,11 +809,59 @@ def query_registry(registry_session, image, digest=None, version='v1', is_blob=F
...
+def manifest_is_media_type(response, media_type):
+    try:
+        received_media_type = response.headers['Content-Type']
+    except KeyError:
+        # Guess media type from content
+        logger.debug("No Content-Type header; inspecting content")
+        received_media_type = guess_manifest_media_type(response.content)
+        logger.debug("guessed media type: %s", received_media_type)
+
+    if received_media_type is None:
```
lcarva: What if `media_type` is None? This method will incorrectly return `False`. Not sure if it's worth handling that.

atomic_reactor/util.py
twaugh: Good catch! Don't know how I managed to read the documentation wrong there.

atomic_reactor/util.py
twaugh: I don't remember :-) Removed.

mlangsdorf:
added to rn:
- plugins dealing with manifest digests will log more details to improve the ability to diagnosis situations when only one of the two media types appears on a pulp server
under # Improvements

twaugh:
Yikes! That pulls OSBS-5099 into the current sprint then, unless we revert the merge.

PR #916, opened 2018-02-14
import_image will be as well exit plugin, and will run only if build …

rcerven:
…was successful

### DIFF
--- a/tests/plugins/test_import_image.py
+++ b/tests/plugins/test_import_image.py
@@ -50,7 +50,7 @@ DEFAULT_TAGS_AMOUNT = 6
 
 
 def prepare(insecure_registry=None, namespace=None, primary_images_tag_conf=DEFAULT_TAGS_AMOUNT,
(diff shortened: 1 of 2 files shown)


tests/plugins/test_import_image.py
```diff
@@ -59,6 +59,8 @@ def prepare(insecure_registry=None, namespace=None, primary_images_tag_conf=DEFA
     tasker = DockerTasker()
     workflow = DockerBuildWorkflow(SOURCE, "test-image")
     setattr(workflow, 'builder', X())
+    flexmock(workflow, build_process_failed=build_process_failed)
+#    setattr(workflow, 'build_process_failed', build_process_failed)
```
twaugh: Stale?

rcerven: right will delete it

rcerven:
deleted commented-out code

lcarva:
Updated release notes draft:
> - `import_image` is now also an `ExitPlugin`

PR #917, opened 2018-02-14
pre_bump_release: update all release labels

### LINKED ISSUE #601: bump release will change release only to 1st old label
bump release will change release only to 1st old label

in current case when we have 1 new and 1 old label it works, but once
we add another change of label = 2 old labels, it will change release value of only 1st old label

mlangsdorf:
pre_bump_release: only update the newest release label

pre_bump_release currently adjusts the release value for the 'release' and
'Release' values, but not correctly. add_label sets it correctly for all
variants of the release label. Remove the duplicate functionality so that
bump_release only bumps the release value.
    
Convert pre_bump_release to use Labels from osbs.utils, and then change the
tests case to test that it only sets 'release'.

Fixes #601

### DIFF
--- a/atomic_reactor/plugins/post_tag_from_config.py
+++ b/atomic_reactor/plugins/post_tag_from_config.py
@@ -11,7 +11,8 @@ import re
 
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.constants import INSPECT_CONFIG, TAG_NAME_REGEX
-from atomic_reactor.util import get_preferred_label_key, df_parser, LabelFormatter
+from atomic_reactor.util import df_parser, LabelFormatter
+from osbs.utils import Labels
 
 
 class TagFromConfigPlugin(PostBuildPlugin):
@@ -102,12 +103,11 @@ class TagFromConfigPlugin(PostBuildPlugin):
 
     def get_component_name(self):
         try:
-            name_label = str(get_preferred_label_key(self.labels, "name"))
-            name = self.labels[name_label]
+            labels = Labels(self.labels)
+            _, name = labels.get_name_and_value(Labels.LABEL_TYPE_NAME)
         except KeyError:
             self.log.error('Unable to determine component from "Labels"')
             raise
-
         return name
 
     def run(self):
--- a/atomic_reactor/plugins/pre_add_help.py
+++ b/atomic_reactor/plugins/pre_add_help.py
@@ -24,7 +24,7 @@ from datetime import datetime as dt
 from atomic_reactor import start_time as atomic_reactor_start_time
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.util import df_parser
-from atomic_reactor.util import get_preferred_label
+from osbs.utils import Labels
 
 DEFAULT_HELP_FILENAME = "help.md"
 
@@ -70,16 +70,19 @@ class AddHelpPlugin(PreBuildPlugin):
             return result
 
         dockerfile = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
-        labels = dockerfile.labels
+        labels = Labels(dockerfile.labels)
+        try:
+            _, name = labels.get_name_and_value(Labels.LABEL_TYPE_NAME)
+        except KeyError:
+            name = ''
+        maintainer = dockerfile.labels.get('maintainer', '')
 
         with open(help_path, 'r+') as help_file:
             lines = help_file.readlines()
 
             if not lines[0].startswith("% "):
-                lines.insert(0, "%% %s (1) Container Image Pages\n" %
-                             (get_preferred_label(labels, "name") or ""))
-                lines.insert(1, "%% %s\n" %
-                             (get_preferred_label(labels, "maintainer") or ""))
+                lines.insert(0, "%% %s (1) Container Image Pages\n" % name)
+                lines.insert(1, "%% %s\n" % maintainer)
                 lines.insert(2, "%% %s\n" % dt.fromtimestamp(atomic_reactor_start_time)
                              .strftime(format="%B %-d, %Y"))
 
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -9,9 +9,9 @@ of the BSD license. See the LICENSE file for details.
 from __future__ import unicode_literals
 
 from atomic_reactor.plugin import PreBuildPlugin
-from atomic_reactor.util import (get_all_label_keys, get_preferred_label_key,
-                                 get_preferred_label, df_parser)
+from atomic_reactor.util import df_parser
 from atomic_reactor.koji_util import create_koji_session
+from osbs.utils import Labels
 
 
 class BumpReleasePlugin(PreBuildPlugin):
@@ -109,20 +109,24 @@ class BumpReleasePlugin(PreBuildPlugin):
 
         parser = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
         dockerfile_labels = parser.labels
+        labels = Labels(dockerfile_labels)
 
-        release = get_preferred_label(dockerfile_labels, 'release')
-        if release is not None and not self.append:
-            self.log.debug("release set explicitly so not incrementing")
-            return
+        try:
+            _, release = labels.get_name_and_value(Labels.LABEL_TYPE_RELEASE)
+            if not self.append:
+                self.log.debug("release set explicitly so not incrementing")
+                return
+        except KeyError:
+            release = None
 
-        component_label = get_preferred_label_key(dockerfile_labels,
-                                                  'com.redhat.component')
+        component_label = labels.get_name(Labels.LABEL_TYPE_COMPONENT)
+
         try:
             component = dockerfile_labels[component_label]
         except KeyError:
             raise RuntimeError("missing label: {}".format(component_label))
 
-        version_label = get_preferred_label_key(dockerfile_labels, 'version')
+        version_label = labels.get_name(Labels.LABEL_TYPE_VERSION)
         try:
             version = dockerfile_labels[version_label]
         except KeyError:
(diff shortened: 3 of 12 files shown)


rcerven:
It would be actually nicer to use 'Labels' from osbs.utils, what do you think @twaugh ?

mlangsdorf:
I don't think Labels supports get_all_label_keys(name), which is needed for this code. I could add it to osbs.util, but then there's an unversioned dependency between the two.

My preference would be to commit this code as is, make the changes to osbs.util, and then come back in a month and unify the Labels interface between atomic-reactor and osbs-client.

rcerven:
Labels.LABEL_NAMES[Labels.LABEL_TYPE_RELEASE]

will give you all variants for 'release' label

mlangsdorf:
Okay, that's helpful, but that puts me back at my previous problem of not having a class for flexmock to grab onto.

rcerven:
you can just mock Labels.LABEL_NAMES to contain more values

mlangsdorf:
Right, because it's an object. Thanks.

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -135,10 +134,10 @@ def run(self):
...
         # Always set preferred release label - other will be set if old-style
         # label is present
-        release_labels = get_all_label_keys('release')
-        preferred_release_label = get_preferred_label_key(dockerfile_labels,
-                                                          'release')
-        old_style_label = get_all_label_keys('com.redhat.component')[1]
+        release_labels = LabelKeys.get_all_label_keys('release')
+        preferred_release_label = LabelKeys.get_preferred_label_key(dockerfile_labels,
+                                                                    'release')
+        old_style_label = LabelKeys.get_all_label_keys('com.redhat.component')[1]
```
rcerven: this seems wrong to me

it should read old_style 'release' labels instead

mlangsdorf: That's a straight copy of the code with the addition of the wrapper class for flexmock. There's no semantic change there - if it's wrong, it's wrong in the original code.

mlangsdorf: I'm just repeating the existing code and adding a call to my class wrapper so I can flexmock it later. There's no semantic change there, so if there's an error, it's in the original code.

rcerven: yes, I know, I am just pointing it out, that we might need to change it

mlangsdorf:
Current version uses osbs.utils::labels for this plugin only and passes the existing test cases.

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -9,9 +9,9 @@
 from __future__ import unicode_literals
 
 from atomic_reactor.plugin import PreBuildPlugin
-from atomic_reactor.util import (get_all_label_keys, get_preferred_label_key,
-                                 get_preferred_label, df_parser)
```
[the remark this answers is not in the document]
mlangsdorf: there are still 25 or so calls to get_preferred_label in other plugins.
I can change the focus of this task to clear the rest of them out.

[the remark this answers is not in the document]
twaugh: No, never mind, I've filed a technical-debt story to track that.

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -109,37 +109,40 @@ def run(self):
...
         parser = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
         dockerfile_labels = parser.labels
+        labels = Labels(dockerfile_labels)
 
-        release = get_preferred_label(dockerfile_labels, 'release')
-        if release is not None and not self.append:
-            self.log.debug("release set explicitly so not incrementing")
-            return
+        try:
+            release_val = labels.get_name_and_value(Labels.LABEL_TYPE_RELEASE)[1]
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -109,37 +109,40 @@ def run(self):
...
+            if not self.append:
+                self.log.debug("release set explicitly so not incrementing")
+                return
+        except:
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -109,37 +109,40 @@ def run(self):
...
+            release_val = None
+
+        release = labels.get_name(Labels.LABEL_TYPE_RELEASE)
```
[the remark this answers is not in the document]
rcerven: change this to : 

release_label = Labels.LABEL_NAMES[Labels.LABEL_TYPE_RELEASE][0]

to get really the newest label (what you will get now is most preferred label in Dockerfile)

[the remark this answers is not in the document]
mlangsdorf: Great idea. The original code was unfortunately very unclear about that.

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -109,37 +109,40 @@ def run(self):
...
 
         # Always set preferred release label - other will be set if old-style
         # label is present
-        release_labels = get_all_label_keys('release')
-        preferred_release_label = get_preferred_label_key(dockerfile_labels,
-                                                          'release')
-        old_style_label = get_all_label_keys('com.redhat.component')[1]
-        release_labels_to_be_set = [preferred_release_label]
+        release_labels = labels.LABEL_NAMES[Labels.LABEL_TYPE_RELEASE]
+        old_style_label = labels.LABEL_NAMES[Labels.LABEL_TYPE_COMPONENT][1]
```
rcerven: delete

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -109,37 +109,40 @@ def run(self):
...
+        release_labels_to_be_set = [release]
```
rcerven: delete this, until and including for loop

and leave only that
self.log.info
AND
dockerfile_labels[release_label] = next_release

mlangsdorf: The entire point of the PR is to be able to set an unspecified, but more than two, number of release labels. So removing the ability to set more than one release label seems like a bad solution.

mlangsdorf: Fixed to match the updated JIRA task.

atomic_reactor/plugins/post_tag_from_config.py
```diff
@@ -102,12 +103,12 @@ def get_and_add_tags(self):
 
     def get_component_name(self):
         try:
-            name_label = str(get_preferred_label_key(self.labels, "name"))
-            name = self.labels[name_label]
+            parsed_labels = self.labels.copy()
```
twaugh: Curious what this is for.`Labels` doesn't modify the mapping it is given (should that be clearer in its documentation?).

atomic_reactor/plugins/pre_add_help.py
```diff
@@ -70,16 +70,23 @@ def run(self):
...
+        parsed_labels = dockerfile.labels.copy()
+        labels = Labels(dockerfile.labels)
+        try:
+            _, name = labels.get_name_and_value(Labels.LABEL_TYPE_NAME)
+        except KeyError:
+            name = ""
+        try:
+            maintainer = parsed_labels.get('maintainer')
+        except KeyError:
+            maintainer = ""
```
twaugh: If `'maintainer'` is not present we'll end up with `None`. I think instead of the try/except block we want:
```
maintainer = parsed_labels.get('maintainer', '')
```

tests/plugins/test_bump_release.py
```diff
@@ -180,11 +180,8 @@ def ssl_login(self, cert=None, ca=None, serverca=None, proxyuser=None):
 
         parser = df_parser(plugin.workflow.builder.df_path, workflow=plugin.workflow)
         assert parser.labels['release'] == next_release['expected']
-        # Old-style spellings will be asserted only if other old-style labels are present
-        if 'BZComponent' not in parser.labels.keys():
-            assert 'Release' not in parser.labels
-        else:
-            assert parser.labels['Release'] == next_release['expected']
+        # Old-style spellings should not be asserted
+        assert 'Release' not in parser.labels
```
twaugh: Are we sure this change is correct? I'm not sure why it is needed. The test is slightly odd because (unless I'm reading it wrong) it doesn't set a release/Release label to begin with.

tests/plugins/test_tag_from_config.py
```diff
@@ -108,8 +108,8 @@ def test_tag_from_config_plugin_generated(tmpdir, docker_tasker, tags, name,
 
 
 @pytest.mark.parametrize(('inspect', 'error'), [  # noqa
-    ({'Labels': {}}, "KeyError('name'"),
-    ({}, "KeyError('Labels'"),
+    ({'Labels': {}}, "KeyError(<object"),
+    ({}, "KeyError('Labels"),
```
twaugh: Not sure I understand why either of these changes are needed.

mlangsdorf: Second one is a typo.
First one is because we're no longer looking up 'name', we're looking up LABEL_TYPE_NAME, which is an object. When it isn't found, the KeyError is against the object.

atomic_reactor/plugins/pre_add_help.py
mlangsdorf: Fixed

tests/plugins/test_bump_release.py
mlangsdorf: self.prepare returns a BumpReleasePlugin
plugin.run calls the run method of that plugin.
pre_bump_release.py::run() line 147 sets the release_label to next_release

atomic_reactor/plugins/post_tag_from_config.py
mlangsdorf: I was running into some weird errors when I ran that code, but they've resolved so I've removed the copy and simplified a bit.
Fixed.

tests/plugins/test_bump_release.py
twaugh: I see where it sets the next_release, i.e. the one we want to end up with. I don't see where actual_release, i.e. the one we started with in the Dockerfile, is set. @rcerven?

mlangsdorf:
RN: added
- release label is set correctly
to # Bug Fixes

PR #918, opened 2018-02-15
pre_pull_base_image: favor imageChangeBuild.buildID for parent

mlangsdorf:
if spec.triggeredBy[0].imageChangeBuild.imageID is set in the Build,
use that image as the parent image instead of the base image from
worflkow.builder.

### DIFF
--- a/atomic_reactor/plugins/pre_pull_base_image.py
+++ b/atomic_reactor/plugins/pre_pull_base_image.py
@@ -37,20 +37,30 @@ class PullBaseImagePlugin(PreBuildPlugin):
         self.parent_registry = parent_registry
         self.parent_registry_insecure = parent_registry_insecure
 
+    def resolve_base_image(self, build_json):
+        spec = build_json.get("spec")
+        try:
+            image_id = spec['triggeredBy'][0]['imageChangeBuild']['imageID']
+        except (TypeError, KeyError, IndexError):
+            base_image = self.workflow.builder.base_image
+        else:
+            base_image = ImageName.parse(image_id)  # any exceptions will propagate
+
+        return base_image
+
     def run(self):
         """
         pull base image
         """
-        base_image = self.workflow.builder.base_image
-        if self.parent_registry is not None:
-            self.log.info("pulling base image '%s' from registry '%s'",
-                          base_image, self.parent_registry)
-        else:
-            self.log.info("pulling base image '%s'", base_image)
+        build_json = get_build_json()
+
+        base_image = self.resolve_base_image(build_json)
 
         base_image_with_registry = base_image.copy()
 
         if self.parent_registry:
+            self.log.info("pulling base image '%s' from registry '%s'",
+                          base_image, self.parent_registry)
             # registry in dockerfile doesn't match provided source registry
             if base_image.registry and base_image.registry != self.parent_registry:
                 self.log.error("registry in dockerfile doesn't match provided source registry, "
@@ -62,6 +72,8 @@ class PullBaseImagePlugin(PreBuildPlugin):
                     % (base_image.registry, self.parent_registry))
 
             base_image_with_registry.registry = self.parent_registry
+        else:
+            self.log.info("pulling base image '%s'", base_image)
 
         try:
             self.tasker.pull_image(base_image_with_registry,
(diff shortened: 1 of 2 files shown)


tests/plugins/test_pull_base_image.py
```diff
@@ -30,6 +30,7 @@
 BASE_IMAGE_W_REGISTRY = LOCALHOST_REGISTRY + "/" + BASE_IMAGE
 BASE_IMAGE_W_LIB_REG = LOCALHOST_REGISTRY + "/" + BASE_IMAGE_W_LIBRARY
 UNIQUE_ID = 'build-name-123'
+CHANGE_IMAGE_W_LIB_REG = LOCALHOST_REGISTRY + "/library/busybox:122"
```
rcerven: this isn't used anywhere

mlangsdorf: Thanks. Fixed.

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -41,7 +41,13 @@ def run(self):
         """
         pull base image
         """
-        base_image = self.workflow.builder.base_image
+        build_json = get_build_json()
+        spec = build_json.get("spec")
+        try:
+            image_id = spec['triggeredBy'][0]['imageChangeBuild']['imageID']
+            base_image = ImageName.parse(image_id)
+        except:
```
twaugh: Better to be explicit about what we want to catch here, otherwise  a typo might always send us into the except block. Looks like we want TypeError (in case `spec` is `None`), KeyError (for the dict lookups), and IndexError (for the list lookup).

If we cannot parse the imageID provided in triggeredBy, that is probably a bug in `ImageName.parse()` (or in OpenShift) so it should probably be fatal rather than falling back to what the Dockerfile says the parent image is.

mlangsdorf: Fixed I think.
I'm not sure if the second paragraph means I should add nested trys to specifically blow up if the parse fails.

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -41,7 +41,13 @@ def run(self):
         """
         pull base image
         """
-        base_image = self.workflow.builder.base_image
+        build_json = get_build_json()
```
lcarva: It would be really great if this was encapsulated in a small helper method in this plugin: `base_image = self.resolve_base_image()`. This method is already large.

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -37,17 +37,29 @@ def __init__(self, tasker, workflow, parent_registry=None, parent_registry_insec
...
-    def run(self):
-        """
-        pull base image
-        """
-        base_image = self.workflow.builder.base_image
+    def resolve_base_image(self, build_json):
+        spec = build_json.get("spec")
+        try:
+            image_id = spec['triggeredBy'][0]['imageChangeBuild']['imageID']
+            base_image = ImageName.parse(image_id)
```
twaugh: I think we should move the `ImageName.parse(image_id)` call into an `else` block, to make sure we don't catch exceptions from it. This call should not raise exceptions; if it does it indicates a bug.
```
try:
    image_id = ...  # this might raise TypeError, KeyError, or IndexError
except (TypeError, KeyError, IndexError):
    base_image = ...
else:
    base_image = ImageName.parse(image_id)  # any exceptions will propagate
```

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -37,17 +37,29 @@ def __init__(self, tasker, workflow, parent_registry=None, parent_registry_insec
...
+        except (TypeError, KeyError, IndexError):
+            base_image = self.workflow.builder.base_image
         if self.parent_registry is not None:
             self.log.info("pulling base image '%s' from registry '%s'",
                           base_image, self.parent_registry)
         else:
             self.log.info("pulling base image '%s'", base_image)
```
twaugh: I feel like these log messages belong in `run()` after this method has returned. Perhaps these should be merged into the `if self.parent_registry: // else:` that's already there.

atomic_reactor/plugins/pre_pull_base_image.py
mlangsdorf: fixed.

atomic_reactor/plugins/pre_pull_base_image.py
mlangsdorf: fixed.

mlangsdorf:
added to RN:
- chain rebuilds will use latest built image if a parent image is rebuild during the chain rebuild
under # Bug Fixes

PR #920, opened 2018-02-15
util.py: make sure that the git commit ID is a string, not bytes

owtaylor:
subprocess.check_output() by default returns bytes for python3. Using
universal_newlines=True is a trick to get strings for both python2
and python3, since the encoding option is python3 only.

I hit this in testing atomic-reactor running with Python-3.6 - not sure why it hasn't previously been encountered since the code hasn't apparently changed for quite a while. The symptom that is seen is that the code in add_labels_in_df that escapes the environment values blows up if an environment value is bytes instead of string since b"Foo".translate() wants a different argument than "foo".translate().


twaugh:
Updated draft release notes

PR #921, opened 2018-02-15
pre_resolve_module_compose.py: Handle change to module separator

owtaylor:
It was officially decided that module versions are formatted as N:V:R not N-V-R.
ODCS on input accepts either, so keep on passing N-V-R, but we need to accept
N:V:R if we receive it.


owtaylor:
Will redo this against current master when I get a chance.

owtaylor:
Rebasing over the change introducing `container.yaml` to this plugin found some problems, fixed up here in a pre-patch.. The fixes aren't directly related to the separator, but included here to avoid having dependent pull requests.

twaugh:
Looks great!

I tried verifying the test changes in the first commit by undoing the change to the plugin, but it is still passing. I'm not sure it's testing it right.

atomic_reactor/util.py
```diff
@@ -1128,12 +1128,29 @@ def get_primary_images(workflow):
...
+        m = re.match(r'^([^:]+):([^:]+)$', module)
+    if not m:
+        m = re.match(r'^(.+)-([^-]+)-(\d{14})$', module)
+    if not m:
+        m = re.match(r'^(.+)-([^-]+)$', module)
+    if not m:
+        raise RuntimeError(
+            'Module specification should be NAME:STREAM or NAME:STREAM:VERSION. ' +
+            '(NAME-STREAM and NAME-STREAM-VERSION supported for compatibility.)'
+        )
```
mlangsdorf: This could be rewritten with a list of patterns and a for/break/else iterator. I don't think it would be much shorter, but it would be easier to extend in the future. 

The repeated if not m statements aren't wrong, but something about them bothers me.

owtaylor:
> I tried verifying the test changes in the first commit by undoing the change to the plugin, but it is still passing. I'm not sure it's testing it right.

When I revert the change to the plugin and just have the test changes from the first commit, I get 14/20 fails for 'pytest-3 tests/plugins/test_resolve_module_compose.py. Not sure what I'm doing differently.

owtaylor:
Pushed a new version that improves a commit message, fixes an error message ('s required Flatpak' => is required for Flapaks), and changes the construct that @mlangsdorf didn't like. I don't have a strong opinion as long as there's not indefinite nesting of ifs.

twaugh:
Oh, the tests were being skipped because I didn't have modulemd or pdc_client installed. Sorry for the noise.

twaugh:
Draft release notes updated.

PR #922, opened 2018-02-19
exit_sendmail: fetch and mail logs for failed autorebuilds

mlangsdorf:
When an autorebuild fails, fetch the logs from the obsbs build object
and attach them to the message in addition to providing a link to the
last koji job's log.

### DIFF
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -7,6 +7,9 @@ of the BSD license. See the LICENSE file for details.
 """
 
 from email.mime.text import MIMEText
+from email.mime.multipart import MIMEMultipart
+from email.mime.base import MIMEBase
+from email import encoders
 import smtplib
 import socket
 try:
@@ -19,7 +22,10 @@ from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
 from atomic_reactor.plugins.exit_koji_import import KojiImportPlugin
 from atomic_reactor.plugins.exit_koji_promote import KojiPromotePlugin
 from atomic_reactor.koji_util import create_koji_session, get_koji_task_owner
-from atomic_reactor.util import get_build_json
+from atomic_reactor.util import get_build_json, OSBSLogs
+
+from osbs.conf import Configuration
+from osbs.api import OSBS
 
 
 class SendMailPlugin(ExitPlugin):
@@ -69,7 +75,9 @@ class SendMailPlugin(ExitPlugin):
                  koji_krb_principal=None,
                  koji_krb_keytab=None,
                  to_koji_submitter=False,
-                 to_koji_pkgowner=False):
+                 to_koji_pkgowner=False,
+                 use_auth=None,
+                 verify_ssl=None):
         """
         constructor
 
@@ -114,6 +122,8 @@ class SendMailPlugin(ExitPlugin):
         self.to_koji_submitter = to_koji_submitter
         self.to_koji_pkgowner = to_koji_pkgowner
         self.submitter = self.DEFAULT_SUBMITTER
+        self.use_auth = use_auth
+        self.verify_ssl = verify_ssl
 
         try:
             metadata = get_build_json().get("metadata", {})
@@ -141,6 +151,20 @@ class SendMailPlugin(ExitPlugin):
             self.session = None
         else:
             self.log.info("Koji connection established")
+
+    def _fetch_log_files(self):
+        try:
+            namespace = get_build_json()['metadata']['namespace']
+        except KeyError:
+            namespace = None
+        osbs_conf = Configuration(conf_file=None, openshift_uri=self.url,
+                                  use_auth=self.use_auth, verify_ssl=self.verify_ssl,
+                                  namespace=namespace)
+        osbs = OSBS(osbs_conf, osbs_conf)
+        build_id = get_build_json()['metadata']['name'] or {}
+        osbs_logs = OSBSLogs(self.log)
+
+        return osbs_logs.get_log_files(osbs, build_id)
 
     def _should_send(self, rebuild, success, auto_canceled, manual_canceled):
         """Return True if any state in `self.send_on` meets given conditions, thus meaning
@@ -185,11 +209,30 @@ class SendMailPlugin(ExitPlugin):
             'user': '<autorebuild>' if rebuild else self.submitter,
             'logs': url
         }
-        return (subject_template % formatting_dict, body_template % formatting_dict)
 
-    def _send_mail(self, receivers_list, subject, body):
-        """Actually sends the mail with `subject` and `body` to all members of `receivers_list`."""
-        msg = MIMEText(body)
+        log_files = None
+        if rebuild and endstate == 'Failed':
+            log_files = self._fetch_log_files()
+
+        return (subject_template % formatting_dict, body_template % formatting_dict, log_files)
+
+    def _send_mail(self, receivers_list, subject, body, log_files=None):
+        """Actually sends the mail with `subject` and `body` and optionanl log_file attachements
+        to all members of `receivers_list`."""
+        if log_files:
+            msg = MIMEMultipart()
+            msg.attach(MIMEText(body))
+            for entry in log_files:
+                log_mime = MIMEBase('application', "octet-stream")
+                log_file = entry[0]  # Output.file
+                log_file.seek(0)
+                log_mime.set_payload(log_file.read())
+                encoders.encode_base64(log_mime)
+                log_mime.add_header('Content-Disposition',
+                                    'attachment; filename="{}"'.format(entry[1]['filename']))
+                msg.attach(log_mime)
+        else:
+            msg = MIMEText(body)
         msg['Subject'] = subject
         msg['From'] = self.from_address
         msg['To'] = ', '.join([x.strip() for x in receivers_list])
@@ -304,7 +347,7 @@ class SendMailPlugin(ExitPlugin):
             except RuntimeError as e:
                 self.log.error('couldn\'t get list of receivers, sending error message ...')
                 # Render the body although the receivers cannot be fetched for error message
-                _, expected_body = self._render_mail(
+                _, expected_body, _ = self._render_mail(
                     rebuild, success, auto_canceled, manual_canceled)
                 body = '\n'.join([
                     'Failed to get contact for %s, error: %s' % (str(self.workflow.image), str(e)),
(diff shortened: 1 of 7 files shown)


atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -142,6 +152,31 @@ def __init__(self, tasker, workflow,
...
+                                      use_auth=self.use_auth, verify_ssl=self.verify_ssl,
+                                      namespace=namespace)
+            self.osbs = OSBS(osbs_conf, osbs_conf)
+            res = self.workflow.prebuild_results.get(PLUGIN_KOJI_PARENT_KEY) or {}
+            build_info = res.get('parent-image-koji-build') or {}
+            self.build_id = build_info.get('id')
+
+        try:
+            logs = self.osbs.get_orchestrator_build_logs(self.build_id)
+        except OsbsException as ex:
```
rcerven: 168-176 are missing unit tests

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -185,14 +220,21 @@ def _render_mail(self, rebuild, success, auto_canceled, manual_canceled):
...
+        return (subject_template % formatting_dict, body_template % formatting_dict, full_logs)
+
+    def _send_mail(self, receivers_list, subject, body, full_logs=[]):
         """Actually sends the mail with `subject` and `body` to all members of `receivers_list`."""
         msg = MIMEText(body)
         msg['Subject'] = subject
         msg['From'] = self.from_address
         msg['To'] = ', '.join([x.strip() for x in receivers_list])
+        for entry in full_logs:
+            msg.attach(entry)
```
rcerven: not covered in tests

atomic_reactor/plugins/exit_sendmail.py
mlangsdorf: Fixed.

atomic_reactor/plugins/exit_sendmail.py
mlangsdorf: Fixed.

rcerven: still isn't covered line 237

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -114,6 +121,9 @@ def __init__(self, tasker, workflow,
         self.to_koji_submitter = to_koji_submitter
         self.to_koji_pkgowner = to_koji_pkgowner
         self.submitter = self.DEFAULT_SUBMITTER
+        self.osbs = None
```
twaugh: Is `osbs` used outside `_fetch_full_logs()`?

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -142,6 +152,31 @@ def __init__(self, tasker, workflow,
...
+    def _fetch_full_logs(self):
+        if not self.osbs:
+            namespace = get_build_json().get('metadata', {}).get('namespace', None)
+            osbs_conf = Configuration(conf_file=None, openshift_uri=self.url,
+                                      use_auth=self.use_auth, verify_ssl=self.verify_ssl,
+                                      namespace=namespace)
+            self.osbs = OSBS(osbs_conf, osbs_conf)
+            res = self.workflow.prebuild_results.get(PLUGIN_KOJI_PARENT_KEY) or {}
+            build_info = res.get('parent-image-koji-build') or {}
+            self.build_id = build_info.get('id')
```
twaugh: `build_id` doesn't seem to be used outside of this method. Make it just `build_id` instead of `self.build_id`?

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -142,6 +152,31 @@ def __init__(self, tasker, workflow,
...
+
+        try:
+            logs = self.osbs.get_orchestrator_build_logs(self.build_id)
+        except OsbsException:
+            msg = "unable to get build logs from OSBS"
+            self.log.error(msg)
```
twaugh: Maybe make this `self.log.exception`?

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -142,6 +152,31 @@ def __init__(self, tasker, workflow,
...
+            logs = [msg]
+        except TypeError:
+            # Older osbs-client has no get_orchestrator_build_logs
+            msg = "This OSBS client does not support get_orchestrator_build_logs; no logs available"
+            self.log.error(msg)
+            logs = [msg]
```
twaugh: Maybe just `[]` in that case? For friendlier emails. Don't really mind either way, just raising the question.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -185,14 +220,21 @@ def _render_mail(self, rebuild, success, auto_canceled, manual_canceled):
...
-        return (subject_template % formatting_dict, body_template % formatting_dict)
 
-    def _send_mail(self, receivers_list, subject, body):
+        full_logs = []
+        if rebuild and endstate == 'Failed':
+            full_logs = self._fetch_full_logs()
+
+        return (subject_template % formatting_dict, body_template % formatting_dict, full_logs)
+
+    def _send_mail(self, receivers_list, subject, body, full_logs=[]):
```
twaugh: Generally it's preferable to have None as the default value for parameters instead of `[]`, since lists are mutable. When using it later you can use `for entry in full_logs or []:`.

mlangsdorf: Fixed.

atomic_reactor/plugins/exit_sendmail.py
mlangsdorf: Fixed.

atomic_reactor/plugins/exit_sendmail.py
mlangsdorf: Fixed.

atomic_reactor/plugins/exit_sendmail.py
mlangsdorf: No, I think I was being overly cautious here. Fixed.

atomic_reactor/plugins/exit_sendmail.py
mlangsdorf: Okay, actually definitely fixed this time.

mlangsdorf:
Substantial change to the code so it now fetches the log files, writes them out to a temp file, and uses MIME stuff to read the temp files and attach them to the outgoing message.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -69,7 +76,9 @@ def __init__(self, tasker, workflow,
                  koji_krb_principal=None,
                  koji_krb_keytab=None,
                  to_koji_submitter=False,
-                 to_koji_pkgowner=False):
+                 to_koji_pkgowner=False,
+                 use_auth=None,
```
lcarva: Let's add a corresponding PR in osbs-client to set these new parameters.

Note: it's fine if osbs-client sets a plugin parameter that the plugin is not aware of. Atomic reactor is smart enough to discard those.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -142,6 +153,22 @@ def __init__(self, tasker, workflow,
         else:
             self.log.info("Koji connection established")
 
+    def _fetch_log_files(self):
+        try:
+            namespace = get_build_json().get('metadata', {}).get('namespace', None)
```
lcarva: The usage of `.get` won't cause `KeyError` exception. However, `get_build_json` could raise `KeyError`. It's probably cleaner to rely on `KeyError` for both cases and just do: `namespace = get_build_json()['metadata']['namespace']`

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -142,6 +153,22 @@ def __init__(self, tasker, workflow,
...
+        except KeyError:
+            namespace = None
+        osbs_conf = Configuration(conf_file=None, openshift_uri=self.url,
+                                  use_auth=self.use_auth, verify_ssl=self.verify_ssl,
+                                  namespace=namespace)
+        osbs = OSBS(osbs_conf, osbs_conf)
+        res = self.workflow.prebuild_results.get(PLUGIN_KOJI_PARENT_KEY) or {}
+        build_info = res.get('parent-image-koji-build') or {}
```
lcarva: We want the OpenShift build-id of the currently running build (`get_build_json()['metadata']['name']`), not the Koji build ID of parent image.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -185,11 +212,30 @@ def _render_mail(self, rebuild, success, auto_canceled, manual_canceled):
...
+            log_files = self._fetch_log_files()
+
+        return (subject_template % formatting_dict, body_template % formatting_dict, log_files)
+
+    def _send_mail(self, receivers_list, subject, body, log_files=None):
+        """Actually sends the mail with `subject` and `body` and optionanl log_file attachements
+        to all members of `receivers_list`."""
+        if log_files:
+            msg = MIMEMultipart()
+            msg.attach(MIMEText(body))
```
lcarva: We should probably use a different body message. Otherwise, it'll still contain link to Koji task. This can be handled by another story/PR.

atomic_reactor/util.py
```diff
@@ -1112,3 +1116,62 @@ def split_module_spec(module):
...
+
+        :return: list, of log files
+        """
+
+        logs = None
+        output = []
+
+        # Collect logs from server
+        try:
+            logs = osbs.get_orchestrator_build_logs(build_id)
```
lcarva: This will download logs once for `koji_import` and again for `sendmail`. I think that's fine. The main benefit of downloading again is that by the time `sendmail` runs there's actually more logs available; and the error may have happened between `koji_import` and `sendmail`, e.g. `koji_tag` fails due to changes in koji.

atomic_reactor/plugins/exit_sendmail.py
mlangsdorf: Fixed.

atomic_reactor/plugins/exit_sendmail.py
mlangsdorf: Fixed.

atomic_reactor/util.py
mlangsdorf: Okay.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -142,6 +152,21 @@ def __init__(self, tasker, workflow,
...
+    def _fetch_log_files(self):
+        try:
+            namespace = get_build_json()['metadata']['namespace']
+        except KeyError:
+            namespace = None
+        osbs_conf = Configuration(conf_file=None, openshift_uri=self.url,
+                                  use_auth=self.use_auth, verify_ssl=self.verify_ssl,
+                                  namespace=namespace)
+        osbs = OSBS(osbs_conf, osbs_conf)
+        build_info = get_build_json()['metadata']['name'] or {}
```
lcarva: Name is actually the build "ID".
See:
https://docs.openshift.com/container-platform/3.6/rest_api/openshift_v1.html#v1-build
https://docs.openshift.com/container-platform/3.6/rest_api/openshift_v1.html#v1-objectmeta

mlangsdorf:
added to release notes:
- failed chain rebuilds will attach all logs to the notification message, in addition to linking to the logs of the most recent build
under # Improvements

PR #923, opened 2018-02-20
koji_upload: create extra.docker.digests

mlangsdorf:
Create extra.docker.digests with the manifest media type and digest of
    each image.

### DIFF
--- a/atomic_reactor/plugins/post_koji_upload.py
+++ b/atomic_reactor/plugins/post_koji_upload.py
@@ -20,7 +20,8 @@ from atomic_reactor.plugins.post_rpmqa import PostBuildRPMqaPlugin
 from atomic_reactor.constants import PROG, PLUGIN_KOJI_UPLOAD_PLUGIN_KEY
 from atomic_reactor.util import (get_version_of_tools, get_checksums,
                                  get_build_json, get_docker_architecture,
-                                 get_image_upload_filename)
+                                 get_image_upload_filename,
+                                 get_manifest_media_type)
 from atomic_reactor.koji_util import create_koji_session
 from atomic_reactor.rpm_util import parse_rpm_output, rpm_qf_args
 from osbs.conf import Configuration
@@ -306,27 +307,6 @@ class KojiUploadPlugin(PostBuildPlugin):
 
         return metadata, output
 
-    def get_digests(self):
-        """
-        Returns a map of images to their digests
-        """
-
-        digests = {}  # repository -> digests
-        for registry in self.workflow.push_conf.docker_registries:
-            for image in self.workflow.tag_conf.images:
-                image_str = image.to_str()
-                if image_str in registry.digests:
-                    image_digests = registry.digests[image_str]
-                    if self.report_multiple_digests:
-                        digest_list = [digest for digest in (image_digests.v1,
-                                                             image_digests.v2)
-                                       if digest]
-                    else:
-                        digest_list = [self.select_digest(image_digests)]
-                    digests[image.to_str(registry=False)] = digest_list
-
-        return digests
-
     def select_digest(self, digests):
         digest = digests.default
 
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -324,8 +327,17 @@ def get_digests(self):
...
+                if image_str in digest_images:
+                    digest = digest_images[image_str]
+                    for digest_version in digest.content_type:
+                        if digest_version not in digest:
+                            continue
+                        digest_type = get_manifest_media_type(digest_version)
+                        docker_digests[digest_type] = digest[digest_version]
+
+        self.log.debug("docker_digests is -%s-", docker_digests)
+        return digests, docker_digests
```
twaugh: It would feel a little cleaner if this could return a dict mapping image names (`image.to_str(registry=False)`) to a tuple of (media_type, digest). Then get_repositories() could be fed a result based on that, e.g. `{key: [digest for _, digest in typed_digests] for key, typed_digests in digests.items()}`.

mlangsdorf: I honestly don't see how that's cleaner. Instead of returning the two dictionaries that we need, with different keys, we return one dictionary that needs to be parsed for each use case.

my code is extra.docker.repositories = get_repositories(digests)
extra.docker.digests = docker_digests

your code is extra.docker.repositories = get_repositories({key: [digest for _, digest in typed_digests] for key, typed_digests in digests.items()})
extra.docker.digests = {type_digests[1]: key for key, typed_digests in digests.items()}

I may have gotten the parsing for extra.docker.digests wrong there.

twaugh: Something like that, yes, although I'd write the second one something like `{media_type: digest for media_type, digest in digests.values()}`.

The thing that I'm trying to make clearer/cleaner is "return digests, docker_digests", which doesn't make it very clear what the difference is between the two items in that returned tuple.

To my mind, the dict comprehensions make it much clearer what the metadata represents. Just me?

mlangsdorf: Oh, the variable names are completely unclear. I agree that's a problem that I have often been struggling with during my experience with this project.
The digests return of get_digests is only ever fed to get_repositories. Would it help to merge get_digests and get_repositories into one function, get_docker_repositories_and_digests, that returns the dict that will be assigned to extras.docker.repositories and the dict that will be assigned to extra.docker.digests? That seems cleaner to me.

mlangsdorf:
Looking at test_koji_import, it already doesn't mock all of the keys in extra.docker, so I don't think it's necessary to mock in extra.docker.digests

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -308,10 +309,11 @@ def get_image_output(self):
 
     def get_digests(self):
         """
-        Returns a map of images to their digests
+        Returns a map of images to their digests and map of media types to their images
```
twaugh: How about:
`Returns a map of images to their digests and map of media types to each digest`

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -308,10 +309,11 @@ def get_image_output(self):
...
         """
 
         digests = {}  # repository -> digests
+        docker_digests = {}
```
twaugh: Can we call it `typed_digests` instead of `docker_digests` maybe? Also would be nice to have a comment, `# media type -> digest`. While here I notice that the line above should probably say "image -> digests" (not repository).

mlangsdorf: Fixed.

atomic_reactor/plugins/post_koji_upload.py
mlangsdorf: Fixed.
Also consolidated the functions and wrote a little bit more of an explanation why.

mlangsdorf:
RN: Added 
- extra.docker.digests 
to Bug Fixes

PR #924, opened 2018-02-20
recognize scratch builds

mlangsdorf:
Recognize the special status of scratch image. Adjust plugins to check for this to be able to succesfully utilize plugins even when basing on scratch. Changes included:
- Do not attempt to inspect if using scratch image as this will always fail
- Only perform ID check if we are not building on top of scratch image
- Do not fail on inspection check when building on top of scratch image
- Return empty inspect data when building on top of scratch image
- Set empty parent_id in case the base image was scratch
- Return empty component list for images based on scratch
- Do not include base image in squash when building on top of scratch image

Orignally coded by niko.kortstrom@gmail.com but apparently abandoned.

### DIFF
--- a/tests/plugins/test_pull_base_image.py
+++ b/tests/plugins/test_pull_base_image.py
@@ -57,7 +57,8 @@ def set_build_json(monkeypatch):
 
 
 @pytest.mark.parametrize(('parent_registry',
-                          'df_base',       # unique ID is always expected
+                          'df_base',      # the base image is always expected unless
+                                          # it's explicitly listed in 'not_expected'
                           'expected',      # additional expected images
                           'not_expected',  # additional images not expected
                           ), [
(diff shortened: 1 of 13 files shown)


tests/plugins/test_pull_base_image.py
```diff
@@ -245,3 +247,7 @@ class MockResponse(object):
         runner.run()
 
     assert error_message in exc.value.args[0]
+
+
+def test_pull_scratch_base():
+    test_pull_base_image_plugin('localhost:5000', 'scratch:latest', [], ['scratch:latest'])
```
rcerven: check at least the log message:
 self.log.debug('base image is scratch, do not attempt to pull')

rcerven:
looks good

goldmann:
This is a very nice contribution. Any chance to revisit it and resubmit?

lcarva:
> This is a very nice contribution. Any chance to revisit it and resubmit?

Sure. We just haven't gotten the bandwidth to do that right now.

PR #926, opened 2018-02-26
Set Repos to Autopublish False

bfontecc007:

### DIFF
--- a/atomic_reactor/pulp_util.py
+++ b/atomic_reactor/pulp_util.py
@@ -114,9 +114,12 @@ class PulpHandler(object):
             # Tell dockpulp.
             self.p.set_certs(cer, key)
 
-    def _create_missing_repos(self, pulp_repos, repo_prefix):
+    def _ensure_repos(self, pulp_repos, repo_prefix):
         repos = pulp_repos.keys()
-        found_repos = self.p.getRepos(repos, fields=["id"])
+        try:
+            found_repos = self.p.getRepos(repos, fields=["id"], distributors=True)
+        except TypeError:
+            found_repos = self.p.getRepos(repos, fields=["id"])
         found_repo_ids = [repo["id"] for repo in found_repos]
 
         missing_repos = set(repos) - set(found_repo_ids)
@@ -125,6 +128,16 @@ class PulpHandler(object):
             self.p.createRepo(repo, None,
                               registry_id=pulp_repos[repo].registry_id,
                               prefix_with=repo_prefix)
+
+        for repo in found_repos:
+            do_update = False
+            if 'distributors' in repo:
+                for d in repo["distributors"]:
+                    if d["auto_publish"]:
+                        do_update = True
+                        break
+            if do_update:
+                self.p.updateRepo(repo, {'auto_publish': False})
 
     def get_tar_metadata(self, tarfile):
         metadata = dockpulp.imgutils.get_metadata(tarfile)
@@ -163,7 +176,7 @@ class PulpHandler(object):
                     tags=[tag]
                 )
 
-        self._create_missing_repos(pulp_repos, repo_prefix)
+        self._ensure_repos(pulp_repos, repo_prefix)
 
         return pulp_repos
 
--- a/tests/test_pulp_util.py
+++ b/tests/test_pulp_util.py
@@ -17,6 +17,7 @@ from atomic_reactor.pulp_util import PulpHandler
 from atomic_reactor.util import ImageName
 
 import pytest
+import copy
 from flexmock import flexmock
 from tests.constants import INPUT_IMAGE, SOURCE, MOCK
 if MOCK:
@@ -80,7 +81,7 @@ def prepare(testfile="test-image", check_repo_retval=0):
      .with_args(object, object))
     (flexmock(dockpulp.Pulp)
      .should_receive('getRepos')
-     .with_args(list, fields=list)
+     .with_args(list, fields=list, distributors=True)
      .and_return([
          {"id": "redhat-image-name1"},
          {"id": "redhat-prefix-image-name2"}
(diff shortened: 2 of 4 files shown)


atomic_reactor/pulp_util.py
```diff
@@ -101,7 +101,7 @@ def _set_auth(self):
             # Tell dockpulp.
             self.p.set_certs(cer, key)
 
-    def _create_missing_repos(self, pulp_repos, repo_prefix):
+    def _ensure_repos(self, pulp_repos, repo_prefix):
         repos = pulp_repos.keys()
         found_repos = self.p.getRepos(repos, fields=["id"])
```
twaugh: We should use this opportunity to figure out whether we need to update auto_publish.

lcarva: Let's also make sure we keep backwards compatibility. For instance, if the version of dockpulp doesn't support checking repo's `auto_publish` status, there should be a fallback. This fallback could be be simply to update repo anyways, but... support for changing `auto_publish` is also not in the dockpulp version. So the fallback should be a noop. Thoughts?

twaugh: Fallback should definitely be a no-op.

lcarva:
Still needs tests.

atomic_reactor/pulp_util.py
```diff
@@ -103,7 +103,7 @@ def _set_auth(self):
 
     def _ensure_repos(self, pulp_repos, repo_prefix):
         repos = pulp_repos.keys()
-        found_repos = self.p.getRepos(repos, fields=["id"])
+        found_repos = self.p.getRepos(repos, fields=["id"], distributors=True)
```
twaugh: Well, hold on. This now won't work with versions of dockpulp that don't allow that parameter in getRepos(). :(

I think we need to catch `TypeError` and try again in that case, e.g.
```
kwargs = {'fields': ['id'], 'distributors': True}
try:
    found_repos = ...(..., **kwargs)
except TypeError:
    # dockpulp < x.y doesn't support getRepos(..., distributors=...)
    del kwargs['distributors']
    found_repos = ...(..., **kwargs)
```

tests/test_pulp_util.py
```diff
@@ -197,3 +197,53 @@ def test_get_pulp_instance():
...
+    pulp_registry_name = 'registry.example.com'
+    testfile = 'foo'
+
+    _, workflow = prepare(testfile)
+    handler = PulpHandler(workflow, pulp_registry_name, log)
+    handler.create_dockpulp_and_repos([])
+
+    (flexmock(dockpulp.Pulp)
+     .should_receive('getRepos')
+     .with_args(['redhat-myproject-hello-world'], fields=['id'], distributors=True)
```
twaugh: This is fine when dockpulp is new enough, but I think we also need a separate test in which this call raises `TypeError`. It's fine for that case not to try to update the auto_publish setting (it won't know it needs to).

atomic_reactor/pulp_util.py
```diff
@@ -113,6 +113,16 @@ def _create_missing_repos(self, pulp_repos, repo_prefix):
...
 
+        for repo in found_repos:
+            do_update = False
+            if 'distributors' in repo:
+                for d in repo["distributors"]:
+                    if d["auto_publish"]:
+                        do_update = True
+                        break
+            if do_update:
+                self.p.updateRepo(repo, {'auto_publish': False})
```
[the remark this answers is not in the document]
twaugh: (but I guess dockpulp is pretty inconsistent with how it handles other things like redirect-url anyway... :( )

tests/test_pulp_util.py
```diff
@@ -197,3 +198,69 @@ def test_get_pulp_instance():
...
     assert handler.get_pulp_instance() == pulp_registry_name
+
+
+@pytest.mark.skipif(dockpulp is None,
+                    reason='dockpulp module not available')
+@pytest.mark.parametrize(("auto_publish"), [
+    (True),
+    (False),
+])
+@pytest.mark.parametrize(("fail"), [
```
lcarva: Can we call this "unsupported" or something that is more descriptive?

tests/test_pulp_util.py
```diff
@@ -197,3 +198,69 @@ def test_get_pulp_instance():
...
+    dist_data = [
+          {
+            "repo_id": "redhat-myproject-hello-world",
+            "auto_publish": auto_publish,
+          }
+        ]
+    mock_get_data = [
+      {
+        "id": "redhat-myproject-hello-world",
+        "scratchpad": {
```
lcarva: Is `scratchpad`, `tags`, `display_name` and `description` relevant to the test? It seems to clutter the test case.

tests/test_pulp_util.py
```diff
@@ -197,3 +198,69 @@ def test_get_pulp_instance():
...
+
+    _, workflow = prepare(testfile)
+    handler = PulpHandler(workflow, pulp_registry_name, log)
+    handler.create_dockpulp_and_repos([])
+
+    if fail:
+        (flexmock(dockpulp.Pulp)
+         .should_receive('getRepos')
+         .with_args(['redhat-myproject-hello-world'], fields=['id'], distributors=True)
+         .and_raise(TypeError))
```
lcarva: Let's add `.once()` to ensure it's actually being called.

tests/test_pulp_util.py
```diff
@@ -197,3 +198,69 @@ def test_get_pulp_instance():
...
+        (flexmock(dockpulp.Pulp)
+         .should_receive('getRepos')
+         .with_args(['redhat-myproject-hello-world'], fields=['id'])
+         .and_return(mock_get_data))
```
lcarva: Let's add `.once()` to ensure it's actually being called.

tests/test_pulp_util.py
```diff
@@ -197,3 +198,69 @@ def test_get_pulp_instance():
...
+    else:
+        (flexmock(dockpulp.Pulp)
+         .should_receive('getRepos')
+         .with_args(['redhat-myproject-hello-world'], fields=['id'], distributors=True)
+         .and_return(data_with_dist))
```
lcarva: Let's add `.once()` to ensure it's actually being called.

tests/test_pulp_util.py
```diff
@@ -197,3 +198,69 @@ def test_get_pulp_instance():
...
+
+    if auto_publish and not fail:
+        (flexmock(dockpulp.Pulp)
+         .should_receive('updateRepo')
+         .with_args(data_with_dist[0], {'auto_publish': False}))
```
lcarva: Let's add `.once()` to ensure it's actually being called.

tests/test_pulp_util.py
```diff
@@ -197,3 +198,69 @@ def test_get_pulp_instance():
...
+         .should_receive('getRepos')
+         .with_args(['redhat-myproject-hello-world'], fields=['id'])
+         .and_return(mock_get_data))
+    else:
+        (flexmock(dockpulp.Pulp)
+         .should_receive('getRepos')
+         .with_args(['redhat-myproject-hello-world'], fields=['id'], distributors=True)
+         .and_return(data_with_dist))
+
+    if auto_publish and not fail:
```
lcarva: Let's add an else with `.never()` in the expectation to ensure `updateRepo` is not called.

tests/test_pulp_util.py
```diff
@@ -197,3 +198,69 @@ def test_get_pulp_instance():
...
+    data_with_dist = copy.deepcopy(mock_get_data)
+    data_with_dist[0]["distributors"] = dist_data
+
+    log = logging.getLogger("tests.test_pulp_util")
+    pulp_registry_name = 'registry.example.com'
+    testfile = 'foo'
+
+    _, workflow = prepare(testfile)
+    handler = PulpHandler(workflow, pulp_registry_name, log)
+    handler.create_dockpulp_and_repos([])
```
lcarva: Don't we want the expectations set before making this call?

bfontecc007: create_dockpulp_and_repos is called without any images so ensure is never called (no repos are found). I'd hate to have to mock images with repos so this seems like the easiest way.

lcarva: This seems to work:
```diff
diff --git a/tests/test_pulp_util.py b/tests/test_pulp_util.py
index 2dab182..da596f7 100644
--- a/tests/test_pulp_util.py
+++ b/tests/test_pulp_util.py
@@ -226,8 +226,6 @@ def test_ensure_repos(auto_publish, unsupported):
     testfile = 'foo'
 
     _, workflow = prepare(testfile)
-    handler = PulpHandler(workflow, pulp_registry_name, log)
-    handler.create_dockpulp_and_repos([])
 
     if unsupported:
         (flexmock(dockpulp.Pulp)
@@ -257,4 +255,6 @@ def test_ensure_repos(auto_publish, unsupported):
          .should_receive('updateRepo')
          .never())
 
-    handler._ensure_repos({'redhat-myproject-hello-world': None}, None)
+    image_names = [ImageName(repo="myproject-hello-world")]
+    handler = PulpHandler(workflow, pulp_registry_name, log)
+    handler.create_dockpulp_and_repos(image_names)

```

lcarva:
Updated release notes draft:
> - update pulp repos to not automatically publish

PR #927, opened 2018-03-02
sendmail: only include logs URL when appropriate (OSBS-5080)

twaugh:
We should only include the logs URL in certain circumstances.

This PR has three commits:
- clean up one of the sendmail test cases
- adjust the test case for intended change, to show the test case catches failure
- apply the intended change, to make the test case pass again

### DIFF
--- a/tests/plugins/test_sendmail.py
+++ b/tests/plugins/test_sendmail.py
@@ -168,6 +168,8 @@ class TestSendMailPlugin(object):
         p = SendMailPlugin(None, WF(), **kwargs)
         assert p._should_send(rebuild, success, auto_canceled, manual_canceled) == expected
 
+    @pytest.mark.parametrize('success', (True, False))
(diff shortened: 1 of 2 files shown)


tests/plugins/test_sendmail.py
```diff
@@ -243,33 +244,42 @@ class WF(object):
...
+                assert not logs
+            else:
+                assert logs
+
+        if auto_cancel or manual_cancel:
+            status = 'Canceled'
+        elif success:
+            status = 'Succeeded'
+        else:
+            status = 'Failed'
```
mlangsdorf: Merge these two if statements, ie:
       if auto_cancel or manual_cancel:
            status = 'Canceled'
            assert not logs
        elif success:
            status = 'Succeeded'
            assert not logs
        else:
            status = 'Failed'
            if autorebuild:
                assert logs
            else:
                assert not logs

Issue #866, opened 2017-10-18
Copy all Pulp docker_image units in a single task


twaugh:
Currently the "pulp_push" plugin copies all image layers separately, each in their own copy task:
```
             for layer in layers:
                 self.pulp_handler.copy(repo_id, layer)
```

Instead it should create a set of filters and use those in the task, e.g.
```
            pulp_filters = {
                'unit': {
                    '$or': [{'image_id': layer}
                            for layer in layers],
                },
            }
            pulp.copy_filters(repo_id, pulp_filters)
```

This details of this should probably go into the PulpHandler class.

lcarva:
This is done by the linked MR above.

PR #928, opened 2018-03-02
Copy all v1 layers in a single pulp task

lcarva:

This is not in this sprint, but this helps out a bit with performance when interacting with pulp.

Without this, a new pulp task is created for each layer in the built image (including layers in parent image).
Depending on the environment, each one of those tasks can take over 30 seconds. This change copies all v1 content in a single copy task.


twaugh:
Fixes #866

lcarva:
Updated release notes draft:
> - `pulp_push` plugin copies all v1 layers in a single pulp task

PR #935, opened 2018-03-08
reactor config env for worker builds

rcerven:

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -275,6 +275,7 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
 
         self.adjust_build_kwargs()
         self.adjust_config_kwargs()
+        self.reactor_config = get_config(self.workflow)
 
         self.find_cluster_retry_delay = find_cluster_retry_delay
         self.failure_retry_delay = failure_retry_delay
@@ -420,6 +421,8 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
 
     def get_cluster_info(self, cluster, platform):
         kwargs = deepcopy(self.config_kwargs)
+        if not self.reactor_config.is_default():
+            kwargs['reactor_config_override'] = self.reactor_config.conf
         kwargs['conf_section'] = cluster.name
         if self.osbs_client_config:
             kwargs['conf_file'] = os.path.join(self.osbs_client_config, 'osbs.conf')
@@ -590,8 +593,7 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
     def select_and_start_cluster(self, platform):
         ''' Choose a cluster and start a build on it '''
 
-        config = get_config(self.workflow)
-        clusters = config.get_enabled_clusters_for_platform(platform)
+        clusters = self.reactor_config.get_enabled_clusters_for_platform(platform)
 
         if not clusters:
             raise UnknownPlatformException('No clusters found for platform {}!'
--- a/atomic_reactor/plugins/pre_reactor_config.py
+++ b/atomic_reactor/plugins/pre_reactor_config.py
@@ -336,6 +336,9 @@ class ReactorConfig(object):
             odcs_config = ODCSConfig(**odcs_config_kwargs)
         return odcs_config
 
+    def is_default(self):
+        return self.conf == self.DEFAULT_CONFIG
+
 
 class ODCSConfig(object):
     """
--- a/tests/plugins/test_orchestrate_build.py
+++ b/tests/plugins/test_orchestrate_build.py
@@ -30,7 +30,6 @@ from osbs.conf import Configuration
 from osbs.build.build_response import BuildResponse
 from osbs.exceptions import OsbsException
 from tests.constants import MOCK_SOURCE, TEST_IMAGE, INPUT_IMAGE, SOURCE
-from tests.util import mocked_reactorconfig
 from tests.fixtures import reactor_config_map  # noqa
 from tests.docker_mock import mock_docker
 from textwrap import dedent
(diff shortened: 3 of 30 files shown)


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -420,6 +421,7 @@ def _get_openshift_session(self, kwargs):
 
     def get_cluster_info(self, cluster, platform):
         kwargs = deepcopy(self.config_kwargs)
+        kwargs['reactor_config_yaml'] = yaml.dump(self.reactor_config)
```
lcarva: Can we only set this if a non-empty `self.reactor_config` is returned by `get_config` ?

This might need a new method in `ReactorConfig` class to determine if config is empty. Essentially a comparison of `self.conf` with `DEFAULT_CONFIG`.

lcarva: Also, I don't think we need to serialize `self.reactor_config` here. That should probably be done by osbs-client when processing this value and stuffing it in an environment variable. Thoughts?

rcerven: fixed

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -336,6 +336,9 @@ def get_odcs_config(self):
             odcs_config = ODCSConfig(**odcs_config_kwargs)
         return odcs_config
 
+    def config_not_empty(self):
```
lcarva: This is an odd name for it, how about `is_empty` ?

rcerven: sure

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -420,6 +421,8 @@ def _get_openshift_session(self, kwargs):
 
     def get_cluster_info(self, cluster, platform):
         kwargs = deepcopy(self.config_kwargs)
+        if self.reactor_config.is_empty():
```
lcarva: `not`

rcerven: right :)

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -420,6 +421,8 @@ def _get_openshift_session(self, kwargs):
 
     def get_cluster_info(self, cluster, platform):
         kwargs = deepcopy(self.config_kwargs)
+        if not self.reactor_config.is_empty():
+            kwargs['reactor_config_override'] = self.reactor_config.conf
```
lcarva: The next step here is making this backwards compatible with osbs-client versions that don't support this keyword argument. I think checking that arrangement version is `>= 6` should be sufficient.

rcerven: if osbs-client doesn't support argument it will be in kwargs and nothing will happen

lcarva: You're absolutely right! Ignore my silly request.

tests/plugins/test_orchestrate_build.py
```diff
@@ -364,6 +364,7 @@ def test_orchestrate_build(tmpdir, caplog, config_kwargs,
     if config_kwargs is not None:
         expected_kwargs.update(config_kwargs)
     expected_kwargs['build_image'] = 'some_image:latest'
+    expected_kwargs['reactor_config_override'] = reactor_dict
```
lcarva: `reactor_config_override` is set in both cases?

rcerven: yes because even when we aren't using reactor config map we still "are using it for clusters"

lcarva: aha! So is the check for empty really not a valid use case because reactor_config will never be empty?

rcerven: if clusters will be None it will fail in reactor_config on NonType not having items,
so I've added use case in unkown_platform to provide empty clusters {}

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -336,6 +336,9 @@ def get_odcs_config(self):
             odcs_config = ODCSConfig(**odcs_config_kwargs)
         return odcs_config
 
+    def is_empty(self):
+        return self.conf != self.DEFAULT_CONFIG
+
```
mlangsdorf: Since the only caller of is_empty() is get_cluster_info() and it negates the return, could this be changed to is_default() and be 'return self.config == self.DEFAULT_CONFIG', and then get_cluster_info would have 'if self.reactor_config.is_default():'

The double negation is confusing.

rcerven: ok

rcerven: actually the logic was wrong there has to be 1 negation, because we wanna set override only if it is NOT default :)

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -336,6 +336,9 @@ def get_odcs_config(self):
             odcs_config = ODCSConfig(**odcs_config_kwargs)
         return odcs_config
 
+    def is_default(self):
```
lcarva: @rcerven is investigating if we can simply drop this mocked class since it's pretty much a copy of `ReactorConfig`.

rcerven: ok done, there was problem with that modification of conf in resolve_compose test, which actually was affecting values in new ReactorConfig instance ...

lcarva:
Updated release notes draft:
> - `orchestrate_build` uses 'reactor_config_override' when creating worker builds

PR #938, opened 2018-03-12
orchestrate_build plugin: override_build_kwarg() 'platform' parameter

mlangsdorf:
allow overrride_build_kwarg to take an optional platform parameter. When
it is supplied, the parameter key-value is assigned to a new platform
dictionary in the override dictionary. If it is not assigned, the key-value
pair is assigned to a 'None' dictionary instead.

When override parameters are processed, the 'None' dictionary updates the
kwargs first, then the platform dictionary for the platform, if either
exist.

please review.

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -69,15 +69,17 @@ def get_koji_upload_dir(workflow):
     return workspace[WORKSPACE_KEY_UPLOAD_DIR]
 
 
-def override_build_kwarg(workflow, k, v):
+def override_build_kwarg(workflow, k, v, platform=None):
     """
     Override a build-kwarg for all worker builds
     """
     key = OrchestrateBuildPlugin.key
+    # Use None to indicate an override for all platforms
 
     workspace = workflow.plugin_workspace.setdefault(key, {})
     override_kwargs = workspace.setdefault(WORKSPACE_KEY_OVERRIDE_KWARGS, {})
-    override_kwargs[k] = v
+    override_kwargs.setdefault(platform, {})
+    override_kwargs[platform][k] = v
 
 
 class UnknownPlatformException(Exception):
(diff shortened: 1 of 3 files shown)


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -69,15 +69,18 @@ def get_koji_upload_dir(workflow):
     return workspace[WORKSPACE_KEY_UPLOAD_DIR]
 
 
-def override_build_kwarg(workflow, k, v):
+def override_build_kwarg(workflow, k, v, platform=None):
```
rcerven: you can use default here directly 
platform='None'

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -69,15 +69,18 @@ def get_koji_upload_dir(workflow):
...
     """
     Override a build-kwarg for all worker builds
     """
     key = OrchestrateBuildPlugin.key
+    if not platform:
+        platform = 'None'
```
twaugh: It should be `None` not `'None'`.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -69,15 +69,18 @@ def get_koji_upload_dir(workflow):
...
 
     workspace = workflow.plugin_workspace.setdefault(key, {})
     override_kwargs = workspace.setdefault(WORKSPACE_KEY_OVERRIDE_KWARGS, {})
-    override_kwargs[k] = v
+    override_kwargs[platform] = override_kwargs.setdefault(platform, {})
```
twaugh: Don't think we need to assign here. Just the `setdefault()` call is enough to add the key.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -555,7 +558,12 @@ def do_worker_build(self, cluster_info):
         try:
             kwargs = self.get_worker_build_kwargs(self.release, cluster_info.platform,
                                                   self.koji_upload_dir, self.fs_task_id)
-            kwargs.update(override_kwargs)
+            if 'None' in override_kwargs:
+                kwargs.update(override_kwargs['None'])
```
twaugh: With `None` instead of `'None'`, yes. Can we also add a comment explaining that these parameters are overrides for all platforms?

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: ahhh ok so None type is for all of em, got it :)

atomic_reactor/plugins/build_orchestrate_build.py
mlangsdorf: Indexing things by None makes my head hurt, but it works.

Fixed.

atomic_reactor/plugins/build_orchestrate_build.py
mlangsdorf: Fixed.

atomic_reactor/plugins/build_orchestrate_build.py
mlangsdorf: Fixed.

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: Btw, this is ok because `None` is hashable to it can be used a dict key.

lcarva:
Updated release notes draft:
> - `orchestrate_build` allows platform specific overrides in override_build_kwarg()

PR #939, opened 2018-03-13
resolve_composes plugin: supply 'arches' key for ODCS 'tag' composes

mlangsdorf:
Add a function to get the platforms that build_orchestrate_build will
run on, and then pass those platforms to the OCDS tag composes in the
arches field.

ready for review

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -12,7 +12,7 @@ import os
 import yaml
 
 from atomic_reactor.constants import (PLUGIN_KOJI_PARENT_KEY, PLUGIN_RESOLVE_COMPOSES_KEY,
-                                      REPO_CONTAINER_CONFIG)
+                                      REPO_CONTAINER_CONFIG, PLUGIN_BUILD_ORCHESTRATE_KEY)
 
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.plugins.build_orchestrate_build import override_build_kwarg
@@ -131,6 +131,12 @@ class ResolveComposesPlugin(PreBuildPlugin):
             self.log.info('Autorebuild detected: Ignoring compose_ids plugin parameter')
             self.compose_ids = tuple()
 
+    def get_arches(self):
+        # This will be replaced by a plugin soon but this will do for now
+        for plugin in self.workflow.buildstep_plugins_conf:
+            if plugin['name'] == PLUGIN_BUILD_ORCHESTRATE_KEY:
+                return plugin['args']['platforms']
+
     def read_configs(self):
         self.odcs_config = get_config(self.workflow).get_odcs_config()
         if not self.odcs_config:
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -301,6 +310,8 @@ def render_request(self):
         request = None
         if self.packages:
             request = self.render_packages_request()
+            if self.arches:
+                request.update({'arches': self.arches})
```
rcerven: maybe just : request['arches'] = self.arches 
?

mlangsdorf: arches is an optional keyword. I'd like to leave it that way.
the handling for arches will change somewhat in osbs-5259 anyway.

rcerven: what I've meant, is to keep the 'if self.arches' but instead of update, just set that key directly, as update will do that anyway, but since we are updating only 1 key, we can just as well set it directly

mlangsdorf: Oh, okay.

Fixed.

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -131,6 +131,12 @@ def adjust_for_autorebuild(self):
             self.log.info('Autorebuild detected: Ignoring compose_ids plugin parameter')
             self.compose_ids = tuple()
 
+    def get_arches(self):
+        # This will be replaced by a plugin soon but this will do for now
+        for plugin in self.workflow.buildstep_plugins_conf:
+            if plugin['name'] == PLUGIN_BUILD_ORCHESTRATE_KEY:
+                return plugin.get('args').get('platforms', [])
```
twaugh: If 'args' isn't present, `.get('args')` will return `None` anyway leading to a traceback when we try to call `.get` on it, so we may as well use `plugin['args']`.
If the plugin isn't found, we'll return `None`, but if the plugin is found but doesn't have a 'platforms' argument we'll return `[]`. Probably better to fail in that second (invalid config) case.
So I'd suggest: `return plugin['args']['platforms']`

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -301,6 +323,8 @@ def render_request(self):
         request = None
         if self.packages:
             request = self.render_packages_request()
+            if self.arches:
+                request['arches'] = self.arches
```
twaugh: Do these two lines belong in `render_packages_request`?

atomic_reactor/plugins/pre_resolve_composes.py

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -250,8 +258,21 @@ def resolve_signing_intent(self):
...
+            for arch in compose_info.get('arches', []):
+                repos_by_arch.setdefault(arch, [])
+                repos_by_arch[arch].append(compose_info['result_repofile'])
+
+        if repos_by_arch:
+            for arch in repos_by_arch:
+                override_build_kwarg(self.workflow, 'yum_repourls',
+                                     repos_by_arch[arch], arch)
+        # set generic overrrides otherwise
+        else:
```
rcerven: missing coverage

mlangsdorf:
line 273 now has coverage.

mlangsdorf:
not adding anything to the release notes because these changes get overwritten in PR 942 and will be documented there.

PR #941, opened 2018-03-14
Fix manifest list example in documentation

lkolacek:


mlangsdorf:
no release notes change as this is solely a documentation change

PR #942, opened 2018-03-14
resolve_composes plugin: create 'pulp' ODCS composes and validate container sets

mlangsdorf:
When compose.pulp_repos is true, create 'pulp' ODCS composes for each architecture, using the content sets for each.

please review

### DIFF
--- a/atomic_reactor/constants.py
+++ b/atomic_reactor/constants.py
@@ -116,3 +116,4 @@ MEDIA_TYPE_OCI_V1 = "application/vnd.oci.image.manifest.v1+json"
 MEDIA_TYPE_OCI_V1_INDEX = "application/vnd.oci.image.index.v1+json"
 
 REPO_CONTAINER_CONFIG = 'container.yaml'
+REPO_CONTENT_SETS_CONFIG = 'content_sets.yaml'
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -10,9 +10,11 @@ from __future__ import unicode_literals
 from datetime import datetime, timedelta
 import os
 import yaml
+from collections import defaultdict
 
 from atomic_reactor.constants import (PLUGIN_KOJI_PARENT_KEY, PLUGIN_RESOLVE_COMPOSES_KEY,
-                                      REPO_CONTAINER_CONFIG, PLUGIN_BUILD_ORCHESTRATE_KEY)
+                                      REPO_CONTAINER_CONFIG, REPO_CONTENT_SETS_CONFIG,
+                                      PLUGIN_BUILD_ORCHESTRATE_KEY)
 
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.plugins.build_orchestrate_build import override_build_kwarg
@@ -152,9 +154,15 @@ class ResolveComposesPlugin(PreBuildPlugin):
         if not data and not self.compose_ids:
             raise SkipResolveComposesPlugin('"compose" config not set and compose_ids not given')
 
+        file_path = os.path.join(workdir, REPO_CONTENT_SETS_CONFIG)
+        pulp_data = None
+        if os.path.exists(file_path):
+            with open(file_path) as f:
+                pulp_data = yaml.safe_load(f) or {}
+
         arches = self.get_arches()
 
-        self.compose_config = ComposeConfig(data, self.odcs_config, arches=arches)
+        self.compose_config = ComposeConfig(data, pulp_data, self.odcs_config, arches=arches)
 
     def adjust_compose_config(self):
         if self.signing_intent:
@@ -204,9 +212,10 @@ class ResolveComposesPlugin(PreBuildPlugin):
 
         self.compose_config.validate_for_request()
 
-        compose_request = self.compose_config.render_request()
-        compose_info = self.odcs_client.start_compose(**compose_request)
-        self.compose_ids = [compose_info['id'], ]
+        self.compose_ids = []
+        for compose_request in self.compose_config.render_requests():
+            compose_info = self.odcs_client.start_compose(**compose_request)
+            self.compose_ids.append(compose_info['id'])
 
     def wait_for_composes(self):
         self.log.debug('Waiting for ODCS composes to be available: %s', self.compose_ids)
@@ -258,19 +267,32 @@ class ResolveComposesPlugin(PreBuildPlugin):
         self.compose_config.set_signing_intent(signing_intent['name'])
 
     def forward_composes(self):
-        repos_by_arch = {}
+        repos_by_arch = defaultdict(list)
         # set overrides by arch if arches are available
         for compose_info in self.composes_info:
-            for arch in compose_info.get('arches') or []:
-                repos_by_arch.setdefault(arch, [])
-                repos_by_arch[arch].append(compose_info['result_repofile'])
+            result_repofile = compose_info['result_repofile']
+            try:
+                arches = compose_info['arches']
+            except KeyError:
+                repos_by_arch[None].append(result_repofile)
+            else:
+                for arch in arches.split():
+                    repos_by_arch[arch].append(result_repofile)
 
-        for arch in repos_by_arch:
-            override_build_kwarg(self.workflow, 'yum_repourls', repos_by_arch[arch], arch)
-        # set generic overrrides otherwise
+        # we should almost never have a None entry, but if we do, we need to merge
+        # it with all other repos.
+        try:
+            noarch_repos = repos_by_arch.pop(None)
+        except KeyError:
+            pass
         else:
-            yum_repourls = [compose_info['result_repofile'] for compose_info in self.composes_info]
-            override_build_kwarg(self.workflow, 'yum_repourls', yum_repourls, None)
+            for repos in repos_by_arch.values():
+                repos.extend(noarch_repos)
+        for arch, repofiles in repos_by_arch.items():
+            override_build_kwarg(self.workflow, 'yum_repourls', repofiles, arch)
+        # Only set the None override if there are no other repos
+        if not repos_by_arch:
+            override_build_kwarg(self.workflow, 'yum_repourls', noarch_repos, None)
 
     def make_result(self):
         result = {
@@ -297,10 +319,13 @@ class ResolveComposesPlugin(PreBuildPlugin):
 
 class ComposeConfig(object):
 
-    def __init__(self, data, odcs_config, koji_tag=None, arches=None):
+    def __init__(self, data, pulp_data, odcs_config, koji_tag=None, arches=None):
         data = data or {}
         self.packages = data.get('packages', [])
         self.modules = data.get('modules', [])
+        self.pulp = {}
+        if data.get('pulp_repos'):
+            self.pulp = pulp_data or {}
         self.koji_tag = koji_tag
         self.odcs_config = odcs_config
         self.arches = arches
@@ -315,16 +340,19 @@ class ComposeConfig(object):
     def has_signing_intent_changed(self):
         return self.signing_intent['name'] != self._original_signing_intent_name
 
-    def render_request(self):
+    def render_requests(self):
         self.validate_for_request()
 
-        request = None
+        requests = []
         if self.packages:
-            request = self.render_packages_request()
+            requests.append(self.render_packages_request())
         else:
-            request = self.render_modules_request()
+            requests.append(self.render_modules_request())
 
-        return request
+        for arch in self.pulp:
+            requests.append(self.render_pulp_request(arch))
+
+        return requests
 
     def render_packages_request(self):
         request = {
@@ -342,6 +370,14 @@ class ComposeConfig(object):
             'source_type': 'module',
             'source': ' '.join(self.modules),
             'sigkeys': self.signing_intent['keys'],
+        }
+
+    def render_pulp_request(self, arch):
+        return {
+            'source_type': 'pulp',
+            'source': ' '.join(self.pulp.get(arch, [])),
+            'sigkeys': [],
+            'arches': [arch]
         }
 
     def validate_for_request(self):
--- a/tests/plugins/test_resolve_composes.py
+++ b/tests/plugins/test_resolve_composes.py
@@ -71,7 +71,7 @@ ODCS_COMPOSE_REPOFILE = ODCS_COMPOSE_REPO + '/odcs-1.repo'
 ODCS_COMPOSE_SECONDS_TO_LIVE = timedelta(hours=24)
 ODCS_COMPOSE_TIME_TO_EXPIRE = datetime.utcnow() + ODCS_COMPOSE_SECONDS_TO_LIVE
 ODCS_COMPOSE_DEFAULT_ARCH = 'x86_64'
-ODCS_COMPOSE_ARCHES = [ODCS_COMPOSE_DEFAULT_ARCH]
+ODCS_COMPOSE_DEFAULT_ARCH_LIST = [ODCS_COMPOSE_DEFAULT_ARCH]
 ODCS_COMPOSE = {
     'id': ODCS_COMPOSE_ID,
     'result_repo': ODCS_COMPOSE_REPO,
@@ -80,7 +80,7 @@ ODCS_COMPOSE = {
     'source_type': 'tag',
     'sigkeys': '',
     'state_name': 'done',
-    'arches': ODCS_COMPOSE_ARCHES,
+    'arches': ODCS_COMPOSE_DEFAULT_ARCH,
     'time_to_expire': ODCS_COMPOSE_TIME_TO_EXPIRE.strftime(ODCS_DATETIME_FORMAT),
 }
 
@@ -110,7 +110,7 @@ def workflow(tmpdir):
     buildstep_plugin = [{
         'name': OrchestrateBuildPlugin.key,
         'args': {
-            'platforms': ODCS_COMPOSE_ARCHES
+            'platforms': ODCS_COMPOSE_DEFAULT_ARCH_LIST
         },
     }]
     workflow = DockerBuildWorkflow(MOCK_SOURCE, 'test-image', buildstep_plugins=buildstep_plugin, )
@@ -178,13 +178,25 @@ def mock_repo_config(tmpdir, data=None, signing_intent=None):
     tmpdir.join('container.yaml').write(data)
 
 
+def mock_content_sets_config(tmpdir, data=None):
+    if data is None:
+        data = dedent("""\
+            x86_64:
+            - pulp-spam
+            - pulp-bacon
+            - pulp-eggs
+        """)
+
+    tmpdir.join('content_sets.yaml').write(data)
+
+
 def mock_odcs_request():
     (flexmock(ODCSClient)
         .should_receive('start_compose')
         .with_args(
             source_type='tag',
             source=KOJI_TAG_NAME,
-            arches=ODCS_COMPOSE_ARCHES,
+            arches=ODCS_COMPOSE_DEFAULT_ARCH_LIST,
             packages=['spam', 'bacon', 'eggs'],
             sigkeys=['R123'])
         .and_return(ODCS_COMPOSE))
@@ -239,6 +251,115 @@ class TestResolveComposes(object):
                 source='spam bacon eggs',
                 sigkeys=['R123'])
             .and_return(ODCS_COMPOSE))
+
+        self.run_plugin_with_args(workflow, reactor_config_map=reactor_config_map)
+
+    @pytest.mark.parametrize(('pulp_arches', 'arches'), (  # noqa
+        (None, None),
+        (['x86_64'], None),
+        (['x86_64'], ['x86_64']),
+        (['x86_64', 'ppce64le'], ['x86_64', 'ppce64le']),
+        (['x86_64', 'ppce64le', 'arm64'], ['x86_64', 'ppce64le', 'arm64']),
+        (['x86_64', 'ppce64le', 'arm64'], None),
+    ))
+    def test_request_pulp_and_multiarch(self, workflow, reactor_config_map, pulp_arches, arches):
+        content_set = ''
+        pulp_composes = {}
+        base_repos = ['spam', 'bacon', 'eggs']
+        pulp_id = ODCS_COMPOSE_ID
+        for arch in pulp_arches or []:
+            pulp_id += 1
+            pulp_repos = []
+            content_set += """\n    {0}:""".format(arch)
+            for repo in base_repos:
+                pulp_repo = repo + '-' + arch
+                pulp_repos.append(pulp_repo)
+                content_set += """\n    - {0}""".format(pulp_repo)
+            source = ' '.join(pulp_repos)
+
+            pulp_compose = {
+                'id': pulp_id,
+                'result_repo': ODCS_COMPOSE_REPO,
+                'result_repofile': ODCS_COMPOSE_REPO + '/pulp_compose-' + arch,
+                'source': source,
+                'source_type': 'pulp',
+                'sigkeys': '',
+                'state_name': 'done',
+                'arches': arch,
+                'time_to_expire': ODCS_COMPOSE_TIME_TO_EXPIRE.strftime(ODCS_DATETIME_FORMAT),
+            }
+            pulp_composes[arch] = pulp_compose
+            (flexmock(ODCSClient)
+                .should_receive('start_compose')
+                .with_args(source_type='pulp', source=source, arches=[arch], sigkeys=[])
+                .and_return(pulp_composes[arch]).once())
+            (flexmock(ODCSClient)
+                .should_receive('wait_for_compose')
+                .with_args(pulp_id)
+                .and_return(pulp_composes[arch]).once())
+
+        mock_content_sets_config(workflow._tmpdir, content_set)
+
+        repo_config = dedent("""\
+            compose:
+                pulp_repos: true
+                packages:
+                - spam
+                - bacon
+                - eggs
+            """)
+        mock_repo_config(workflow._tmpdir, repo_config)
+        workflow.buildstep_plugins_conf[0]['args']['platforms'] = arches
+        tag_compose = deepcopy(ODCS_COMPOSE)
+        if arches:
+            tag_compose['arches'] = ' '.join(arches)
+            (flexmock(ODCSClient)
+                .should_receive('start_compose')
+                .with_args(source_type='tag', source=KOJI_TAG_NAME, arches=arches,
+                           packages=['spam', 'bacon', 'eggs'], sigkeys=['R123'])
+                .and_return(tag_compose).once())
+        else:
+            tag_compose.pop('arches')
+            (flexmock(ODCSClient)
+                .should_receive('start_compose')
+                .with_args(source_type='tag', source=KOJI_TAG_NAME,
+                           packages=['spam', 'bacon', 'eggs'], sigkeys=['R123'])
+                .and_return(tag_compose).once())
+
+        (flexmock(ODCSClient)
+            .should_receive('wait_for_compose')
+            .with_args(ODCS_COMPOSE_ID)
+            .and_return(tag_compose).once())
+
+        self.run_plugin_with_args(workflow, reactor_config_map=reactor_config_map,
+                                  platforms=pulp_arches, is_pulp=pulp_arches)
+
+    def test_request_compose_for_pulp_no_container(self, workflow, reactor_config_map):  # noqa
+        (flexmock(ODCSClient)
+            .should_receive('start_compose')
+            .with_args(
+                source_type='pulp',
+                source='pulp is no good here',
+                arches=['x86_64'],
+                sigkeys=[])
+            .never())
+        (flexmock(ODCSClient)
+            .should_receive('wait_for_compose')
+            .with_args(85)
+            .never())
+
+        mock_content_sets_config(workflow._tmpdir, '')
+
+        repo_config = dedent("""\
+            compose:
+                pulp_repos: true
+                packages:
+                - spam
+                - bacon
+                - eggs
+            """)
+        mock_repo_config(workflow._tmpdir, repo_config)
+        mock_odcs_request()
 
         self.run_plugin_with_args(workflow, reactor_config_map=reactor_config_map)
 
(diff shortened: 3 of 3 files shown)


tests/plugins/test_resolve_composes.py
```diff
@@ -493,6 +540,26 @@ def test_invalid_compose_request(self, workflow, config, error_message,  # noqa
         self.run_plugin_with_args(workflow, expect_error=error_message,
                                   reactor_config_map=reactor_config_map)
 
+    @pytest.mark.parametrize(('container_sets', 'error_message'), (
```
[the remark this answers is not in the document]

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -250,8 +303,21 @@ def resolve_signing_intent(self):
         self.compose_config.set_signing_intent(signing_intent['name'])
 
     def forward_composes(self):
-        yum_repourls = [compose_info['result_repofile'] for compose_info in self.composes_info]
-        override_build_kwarg(self.workflow, 'yum_repourls', yum_repourls)
+        repos_by_arch = {}
+        # set overrides by arch if arches are available
+        for compose_info in self.composes_info:
+            for arch in compose_info.get('arches', []):
```
lcarva: What if "arches" is empty? Do we need to set `None` key in `repos_by_arch` ?

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -250,8 +303,21 @@ def resolve_signing_intent(self):
...
+                repos_by_arch.setdefault(arch, [])
+                repos_by_arch[arch].append(compose_info['result_repofile'])
+
+        if repos_by_arch:
```
lcarva: This is not necessary, you can just jump into the for-loop, it'l be a noop if dict is empty.

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -250,8 +303,21 @@ def resolve_signing_intent(self):
...
+            for arch in repos_by_arch:
+                override_build_kwarg(self.workflow, 'yum_repourls',
+                                     repos_by_arch[arch], arch)
+        # set generic overrrides otherwise
+        else:
+            yum_repourls = [compose_info['result_repofile'] for compose_info in self.composes_info]
+            override_build_kwarg(self.workflow, 'yum_repourls', yum_repourls, None)
```
lcarva: We're expecting some composes to apply to all arches, while other composes to be arch specific. This should probably be handled while iterating over `self.composes_info`.

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -131,6 +133,36 @@ def adjust_for_autorebuild(self):
...
+    def get_arches(self):
+        # This will be replaced by a plugin soon but this will do for now
+        for plugin in self.workflow.buildstep_plugins_conf:
+            if plugin['name'] == PLUGIN_BUILD_ORCHESTRATE_KEY:
+                return plugin.get('args').get('platforms', [])
+
+    def validate_container_sets(self, data, container_sets, arches):
+        fail_msg = 'failed container set validation: '
+        data_rpms = []
+        for data_type in ['pulp_repos', 'packages', 'modules']:
```
lcarva: I don't understand what this validation method is trying to accomplish. Can you help me out?

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -146,7 +178,20 @@ def read_configs(self):
...
 
-        self.compose_config = ComposeConfig(data, self.odcs_config)
+        file_path = os.path.join(workdir, REPO_CONTAINER_SETS_CONFIG)
+        container_sets = None
+        if os.path.exists(file_path):
+            with open(file_path) as f:
+                container_sets = yaml.safe_load(f)
+
+        if not container_sets:
+            raise ValueError('no container sets found')
```
lcarva: `container_sets.yaml` is not required.

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -198,7 +243,15 @@ def request_compose_if_needed(self):
...
         compose_info = self.odcs_client.start_compose(**compose_request)
-        self.compose_ids = [compose_info['id'], ]
+        if self.compose_config.pulp:
+            self.composes_info = []
+            for arch in self.compose_config.arches:
+                compose_request = self.compose_config.render_pulp_request(arch)
+                compose_info = self.odcs_client.start_compose(**compose_request)
+                self.composes_info.append(compose_info)
+            self.compose_ids = [item['id'] for item in self.composes_info]
+        else:
```
lcarva: You can have `pulp` and `tag` type composes on the same build.

atomic_reactor/plugins/pre_resolve_composes.py
mlangsdorf: do you expect a single container.yaml file to have both arches and non-arches composes in it? because I'm not sure that's even possible.
if it possible and I'm misreading the docs again, I'd like to see an sample container.yaml file so I have something to work against.

atomic_reactor/plugins/pre_resolve_composes.py
mlangsdorf: validate the container sets
confirm at all arches we intend to compose are in the container sets
confirm that for all RPMs we intend to compose, they are also in the container sets

if the purpose of container sets file is not stop composition when there are RPMs in the compose that aren't in the container sets, then I misread the docs again and would like an explanation of what's going on.

atomic_reactor/plugins/pre_resolve_composes.py

atomic_reactor/plugins/pre_resolve_composes.py
mlangsdorf: fixed.

atomic_reactor/plugins/pre_resolve_composes.py
mlangsdorf: for else: conditionals look weird to me.

atomic_reactor/plugins/pre_resolve_composes.py
mlangsdorf: The way I thought this was going to work:
if there are arches, then override is set specifically per arch
if there are not arches, then override is set generally (ie, None) for all repos in the compose.

mlangsdorf: Fixed I think.

atomic_reactor/plugins/pre_resolve_composes.py

atomic_reactor/plugins/pre_resolve_composes.py
mlangsdorf: Fixed after seeing the examples in the docs.

atomic_reactor/plugins/pre_resolve_composes.py
mlangsdorf: Fixed by deletion.

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -261,16 +277,14 @@ def forward_composes(self):
         repos_by_arch = {}
         # set overrides by arch if arches are available
         for compose_info in self.composes_info:
-            for arch in compose_info.get('arches') or []:
-                repos_by_arch.setdefault(arch, [])
-                repos_by_arch[arch].append(compose_info['result_repofile'])
-
+            if compose_info['source_type'] == 'pulp':
+                arch = compose_info['arches'][0]
```
lcarva: Instead of picking the first one, can we trust the value of `arches` in each `compose_info`?

```python
from collections import defaultdict


repos_by_arch = defaultdict(list)

for compose_info in self.composes_info:
  for arch in compose_info['arches']:
    repos_by_arch[arch].append(compose_info['result_repofile'])

for arch, repofiles in repos_by_arch.items():
  override_build_kwarg(self.workflow, 'yum_repourls', repofiles, arch)
```

tests/plugins/test_resolve_composes.py
```diff
@@ -178,13 +175,25 @@ def mock_repo_config(tmpdir, data=None, signing_intent=None):
     tmpdir.join('container.yaml').write(data)
 
 
+def mock_container_set_config(tmpdir, data=None):
+    if data is None:
+        data = dedent("""\
+            compose:
```
twaugh: As noted in an earlier comment, this line should be removed.

tests/plugins/test_resolve_composes.py
```diff
@@ -242,6 +251,119 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
 
         self.run_plugin_with_args(workflow, reactor_config_map=reactor_config_map)
 
+    @pytest.mark.parametrize('arches', (  # noqa
+        ['x86_64'],
+        ['x86_64', 'ppce64le'],
+        ['x86_64', 'ppce64le', 'arm64'],
+    ))
+    def test_request_compose_for_pulp(self, workflow, reactor_config_map, arches):  # noqa
+        container_set = """compose:"""
```
twaugh: Initialise as empty string.

tests/plugins/test_resolve_composes.py
```diff
@@ -242,6 +251,119 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
 
         self.run_plugin_with_args(workflow, reactor_config_map=reactor_config_map)
 
+    @pytest.mark.parametrize('arches', (  # noqa
+        ['x86_64'],
+        ['x86_64', 'ppce64le'],
+        ['x86_64', 'ppce64le', 'arm64'],
+    ))
+    def test_request_compose_for_pulp(self, workflow, reactor_config_map, arches):  # noqa
```
twaugh: Guessing '# noqa' is redundant here (it's already on L254)

tests/plugins/test_resolve_composes.py
```diff
@@ -242,6 +251,119 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
+        container_set = """compose:"""
+        pulp_composes = {}
+        base_repos = ['spam', 'bacon', 'eggs']
+        pulp_id = ODCS_COMPOSE_ID
+        for arch in arches:
+            pulp_id += 1
+            pulp_repos = []
+            container_set += """\n    {0}:""".format(arch)
```
twaugh: actually content_set

atomic_reactor/constants.py
```diff
@@ -116,3 +116,4 @@
 MEDIA_TYPE_OCI_V1_INDEX = "application/vnd.oci.image.index.v1+json"
 
 REPO_CONTAINER_CONFIG = 'container.yaml'
+REPO_CONTAINER_SETS_CONFIG = 'container_sets.yaml'
```
twaugh: It's content_sets.yaml

tests/plugins/test_resolve_composes.py

tests/plugins/test_resolve_composes.py

atomic_reactor/constants.py
mlangsdorf: Fixed.

tests/plugins/test_resolve_composes.py

atomic_reactor/plugins/pre_resolve_composes.py

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -206,7 +213,16 @@ def request_compose_if_needed(self):
 
         compose_request = self.compose_config.render_request()
         compose_info = self.odcs_client.start_compose(**compose_request)
-        self.compose_ids = [compose_info['id'], ]
+        if self.compose_config.pulp:
```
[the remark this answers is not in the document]

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -152,9 +153,15 @@ def read_configs(self):
         if not data and not self.compose_ids:
             raise SkipResolveComposesPlugin('"compose" config not set and compose_ids not given')
 
+        file_path = os.path.join(workdir, REPO_CONTAINER_SETS_CONFIG)
+        pulp_data = None
+        if os.path.exists(file_path):
+            with open(file_path) as f:
+                pulp_data = (yaml.safe_load(f) or {}).get('compose')
```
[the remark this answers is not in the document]

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -258,19 +267,16 @@ def resolve_signing_intent(self):
...
     def forward_composes(self):
-        repos_by_arch = {}
+        repos_by_arch = defaultdict(list)
         # set overrides by arch if arches are available
         for compose_info in self.composes_info:
             for arch in compose_info.get('arches') or []:
-                repos_by_arch.setdefault(arch, [])
                 repos_by_arch[arch].append(compose_info['result_repofile'])
+            else:
+                repos_by_arch[None].append(compose_info['result_repofile'])
```
twaugh: There's no `break` in the loop, so this `else:` block will *always* run, which doesn't look like what you mean.

mlangsdorf: Fixed. Now pulling arches for pulp from the pulp entries.

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -258,19 +267,18 @@ def resolve_signing_intent(self):
...
+            if not compose_info.get('arches'):
+                repos_by_arch[None].append(compose_info['result_repofile'])
 
-        for arch in repos_by_arch:
-            override_build_kwarg(self.workflow, 'yum_repourls', repos_by_arch[arch], arch)
-        # set generic overrrides otherwise
-        else:
-            yum_repourls = [compose_info['result_repofile'] for compose_info in self.composes_info]
-            override_build_kwarg(self.workflow, 'yum_repourls', yum_repourls, None)
+        print repos_by_arch
```
lcarva: :)

mlangsdorf:
Update to reflect that compose_info.arches is a string, not a list of strings.

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -258,19 +267,16 @@ def resolve_signing_intent(self):
         self.compose_config.set_signing_intent(signing_intent['name'])
 
     def forward_composes(self):
-        repos_by_arch = {}
+        repos_by_arch = defaultdict(list)
         # set overrides by arch if arches are available
         for compose_info in self.composes_info:
-            for arch in compose_info.get('arches') or []:
-                repos_by_arch.setdefault(arch, [])
+            for arch in compose_info.get('arches', '').split(" "):
```
twaugh: Minor:
If 'arches' is not present, we'll end up running through this loop once with `''` for arch (because `''.split(' ')` is `['']`). If we want to handle 'arches' not being present, I'd suggest:
```
result_repofile = compose_info['result_repofile']
try:
    arches = compose_info['arches']
except KeyError:
    repos_by_arch[None].append(result_repofile)
else:
    for arch in arches.split(' '):
        repos_by_arch[arch].append(result_repofile)
```

But first: do we really need to handle 'arches' not being present in the compose response?

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -258,19 +267,16 @@ def resolve_signing_intent(self):
...
+                repos_by_arch[None].append(compose_info['result_repofile'])
 
-        for arch in repos_by_arch:
-            override_build_kwarg(self.workflow, 'yum_repourls', repos_by_arch[arch], arch)
-        # set generic overrrides otherwise
-        else:
-            yum_repourls = [compose_info['result_repofile'] for compose_info in self.composes_info]
-            override_build_kwarg(self.workflow, 'yum_repourls', yum_repourls, None)
+        for arch, repofiles in repos_by_arch.items():
+            override_build_kwarg(self.workflow, 'yum_repourls', repofiles, arch)
```
twaugh: If we ended up setting `None`, those repos will get lost. We need to distribute overrides from `None` to all the actual architectures if we want them to get used:
```
try:
    noarch_repos = repos_by_arch.pop(None)
except KeyError:
    pass
else:
    for repos in repos_by_arch.values():
        repos.extend(noarch_repos)
for arch, repofiles in repos_by_arch.items():
    override_build_kwarg(...)
```

But first: do we really need to handle the case where a compose response does not tell us which arches it is for?

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -152,9 +154,15 @@ def read_configs(self):
         if not data and not self.compose_ids:
             raise SkipResolveComposesPlugin('"compose" config not set and compose_ids not given')
 
+        file_path = os.path.join(workdir, REPO_CONTENT_SETS_CONFIG)
+        pulp_data = None
+        if os.path.exists(file_path):
+            with open(file_path) as f:
+                pulp_data = (yaml.safe_load(f) or {})
```
twaugh: Tiny nit: outer parentheses not needed

tests/plugins/test_resolve_composes.py
```diff
@@ -242,6 +250,143 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
+
+        self.run_plugin_with_args(workflow, reactor_config_map=reactor_config_map,
+                                  pulp_platforms=pulp_arches, arches=arches)
+
+    def test_request_compose_for_pulp_no_container(self, workflow, reactor_config_map):  # noqa
+        pulp_compose = {
+            'id': 85,
+            'result_repo': ODCS_COMPOSE_REPO,
+            'result_repofile': ODCS_COMPOSE_REPO + '/pulp_compose-x86_64',
+            'source': KOJI_TAG_NAME,
```
twaugh: Looks confusing, see https://github.com/projectatomic/atomic-reactor/pull/942/files#r175706467

tests/plugins/test_resolve_composes.py
```diff
@@ -242,6 +250,143 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
+            'time_to_expire': ODCS_COMPOSE_TIME_TO_EXPIRE.strftime(ODCS_DATETIME_FORMAT),
+        }
+        (flexmock(ODCSClient)
+            .should_receive('start_compose')
+            .with_args(
+                source_type='pulp',
+                source='pulp is no good here',
+                arches=['x86_64'],
+                sigkeys=[])
+            .and_return(pulp_compose).never)
```
twaugh: If we don't ever want this to be called, we don't care what our mocked version should return. So we can remove `.and_return(pulp_compose)` and `pulp_compose = ...` above.

tests/plugins/test_resolve_composes.py
```diff
@@ -242,6 +250,143 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
+        (flexmock(ODCSClient)
+            .should_receive('wait_for_compose')
+            .with_args(85)
+            .and_return(pulp_compose).never)
```
twaugh: Also, don't need `.and_return(pulp_compose)`

atomic_reactor/plugins/pre_resolve_composes.py
mlangsdorf: What? No, I don't think that's right. override(None) sets an override for all arches, then override(arch) sets it for an arch.

we should do the pop to guarantee order, but we don't need to distribute because that's already handled by override_build_kwarg.

tests/plugins/test_resolve_composes.py
lcarva: Also, you probably want `.never()` instead of `.never` to actually set the expectation. Otherwise, you're just retrieving the function reference without invoking it.

atomic_reactor/plugins/pre_resolve_composes.py
lcarva: ODCS API is a little odd when it comes to `sigkeys` and `arches` values. Instead of returning a list of strings, it returns a white-space separated string. I'd suggest using the same approach that was used for handling `sigkeys`:

```python
keys = compose_info.get('sigkeys', [])
....
if isinstance(keys, six.text_type):
    keys = keys.split()            
keys = set(keys)                   
```

This has the benefit of being future ready if ODCS ever decides to return the correct data type.

Note that it's better to not pass a parameter to `split`:
```python
>>> ''.split(' ')
['']
>>> ''.split()
[]
```

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -258,19 +267,16 @@ def resolve_signing_intent(self):
...
     def forward_composes(self):
-        repos_by_arch = {}
+        repos_by_arch = defaultdict(list)
         # set overrides by arch if arches are available
         for compose_info in self.composes_info:
-            for arch in compose_info.get('arches') or []:
-                repos_by_arch.setdefault(arch, [])
+            for arch in compose_info.get('arches', '').split(" "):
                 repos_by_arch[arch].append(compose_info['result_repofile'])
+            if not compose_info.get('arches'):
```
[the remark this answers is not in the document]
lcarva: I think it does now, yes.

atomic_reactor/plugins/pre_resolve_composes.py
lcarva: The way `orchestrate_build` plugin reads this parameter it does not merge `None` and `<platform>` keys. If both are set, `<platform>` will override `None`. That's because the internal story for adding that behavior called explicitly to not merge values recursively.

That means they need to be merged here, and `override_build_kwarg(...., None)` should never be called from this `resolve_composes` plugin.

But yes, this may not be needed at all. Talked to Jan from ODCS and he thinks arches "will always be set to something".

mlangsdorf: okay.

mlangsdorf: I'm confused now. Why doesn't Tim's code fail to set any repos when None is the only entry in repos_by_arch?

twaugh: Perhaps it should be only be distributed if there are non-None repos_by_arch:
```
if any(arch != None for arch in repos_by_arch):
    try:
        noarch_repos = repos_by_arch.pop(None)
    except KeyError:
        pass
    else:
        for repos in repos_by_arch.values():
            repos.extend(noarch_repos)
        for arch, repofiles in repos_by_arch.items():
            override_build_kwarg(...)
```

lcarva: Because `pop` will remove the `None` key and its value from `repos_by_arch`, leaving it an empty dict which is still iterable:

```python
>>> {}.values()
[]
```

Oh! I think you're pointing out that the code will essentially be a noop in this case!

Maybe it needs this to always call `override_build_kwarg(...., noarch_repos, None)` then.

twaugh: @lcarva, in other words, you think it should not `pop()` it?
```
try:
    noarch_repos = repos_by_arch[None]
except KeyError:
    pass
else:
    # Copy the repos relevant for all architectures
    # into each arch-specific override
    for repos in repos_by_arch.values():
        repos.extend(noarch_repos)
for arch, repofiles in repos_by_arch.items():
    override_build_kwarg(...)
```

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -344,6 +357,14 @@ def render_modules_request(self):
             'sigkeys': self.signing_intent['keys'],
         }
 
+    def render_pulp_request(self, arch):
+        return {
+            'source_type': 'pulp',
+            'source': ' '.join(self.pulp.get(arch, [])),
```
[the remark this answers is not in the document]
mlangsdorf: Okay.

atomic_reactor/plugins/pre_resolve_composes.py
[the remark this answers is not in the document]
mlangsdorf: Handling it on the off chance.

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -315,16 +324,20 @@ def set_signing_intent(self, name):
...
+        requests = []
         if self.packages:
-            request = self.render_packages_request()
+            requests.append(self.render_packages_request())
         else:
-            request = self.render_modules_request()
+            requests.append(self.render_modules_request())
 
-        return request
+        if self.pulp:
```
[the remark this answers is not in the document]

tests/plugins/test_resolve_composes.py
```diff
@@ -242,6 +250,143 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
+        ['x86_64', 'ppce64le', 'arm64'],
+    ))
+    @pytest.mark.parametrize('arches', (
+        None,
+        ['amd64'],
+        ['amd64', 'power9'],
+        ['amd64', 'power9', 'armv8'],
+    ))
+    def test_request_pulp_and_multiarch(self, workflow, reactor_config_map, pulp_arches, arches):
+        content_set = """"""
```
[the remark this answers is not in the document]

tests/plugins/test_resolve_composes.py
```diff
@@ -242,6 +250,143 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
+            for repo in base_repos:
+                pulp_repo = repo + '-' + arch
+                pulp_repos.append(pulp_repo)
+                content_set += """\n    - {0}""".format(pulp_repo)
+
+            pulp_compose = {
+                'id': pulp_id,
+                'result_repo': ODCS_COMPOSE_REPO,
+                'result_repofile': ODCS_COMPOSE_REPO + '/pulp_compose-' + arch,
+                'source': KOJI_TAG_NAME,
```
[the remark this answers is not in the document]

tests/plugins/test_resolve_composes.py
```diff
@@ -242,6 +250,143 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
         self.run_plugin_with_args(workflow, reactor_config_map=reactor_config_map)
 
+    @pytest.mark.parametrize('pulp_arches', (  # noqa
+        None,
+        ['x86_64'],
+        ['x86_64', 'ppce64le'],
+        ['x86_64', 'ppce64le', 'arm64'],
+    ))
+    @pytest.mark.parametrize('arches', (
+        None,
```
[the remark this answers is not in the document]
mlangsdorf: Revised.

tests/plugins/test_resolve_composes.py
```diff
@@ -242,6 +250,143 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
+                .with_args(source_type='tag', source=KOJI_TAG_NAME,
+                           packages=['spam', 'bacon', 'eggs'], sigkeys=['R123'])
+                .and_return(ODCS_COMPOSE).once)
+            (flexmock(ODCSClient)
+                .should_receive('wait_for_compose')
+                .with_args(ODCS_COMPOSE_ID)
+                .and_return(ODCS_COMPOSE).once)
+
+        self.run_plugin_with_args(workflow, reactor_config_map=reactor_config_map,
+                                  pulp_platforms=pulp_arches, arches=arches)
```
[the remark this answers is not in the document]

tests/plugins/test_resolve_composes.py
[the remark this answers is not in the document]

tests/plugins/test_resolve_composes.py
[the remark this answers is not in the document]

tests/plugins/test_resolve_composes.py
```diff
@@ -242,6 +250,143 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
+            """)
+        mock_repo_config(workflow._tmpdir, repo_config)
+        (flexmock(ODCSClient)
+            .should_receive('start_compose')
+            .with_args(
+                source_type='tag',
+                source=KOJI_TAG_NAME,
+                packages=['spam', 'bacon', 'eggs'],
+                sigkeys=['R123'])
+            .and_return(ODCS_COMPOSE).once)
```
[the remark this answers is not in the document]

tests/plugins/test_resolve_composes.py
```diff
@@ -242,6 +250,143 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
+        (flexmock(ODCSClient)
+            .should_receive('wait_for_compose')
+            .with_args(ODCS_COMPOSE_ID)
+            .and_return(ODCS_COMPOSE).once)
```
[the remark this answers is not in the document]

tests/plugins/test_resolve_composes.py
```diff
@@ -242,6 +250,143 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
+            pulp_composes[arch] = pulp_compose
+            (flexmock(ODCSClient)
+                .should_receive('start_compose')
+                .with_args(source_type='pulp', source=' '.join(pulp_repos),
+                           arches=[arch], sigkeys=[])
+                .and_return(pulp_composes[arch]).once)
+            (flexmock(ODCSClient)
+                .should_receive('wait_for_compose')
+                .with_args(pulp_id)
+                .and_return(pulp_composes[arch]).once)
```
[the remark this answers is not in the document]

tests/plugins/test_resolve_composes.py
```diff
@@ -242,6 +250,143 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
+                'state_name': 'done',
+                'arches': arch,
+                'time_to_expire': ODCS_COMPOSE_TIME_TO_EXPIRE.strftime(ODCS_DATETIME_FORMAT),
+            }
+            pulp_composes[arch] = pulp_compose
+            (flexmock(ODCSClient)
+                .should_receive('start_compose')
+                .with_args(source_type='pulp', source=' '.join(pulp_repos),
+                           arches=[arch], sigkeys=[])
+                .and_return(pulp_composes[arch]).once)
```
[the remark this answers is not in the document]

atomic_reactor/plugins/pre_resolve_composes.py
mlangsdorf: Fixed for both.

atomic_reactor/plugins/pre_resolve_composes.py
mlangsdorf: We should pop it, distribute it we can, and set None if there's nothing left in repos_by_arch.

Fixed.

atomic_reactor/plugins/pre_resolve_composes.py

tests/plugins/test_resolve_composes.py

tests/plugins/test_resolve_composes.py

tests/plugins/test_resolve_composes.py
```diff
@@ -639,46 +760,9 @@ def test_parameters_ignored_for_autorebuild(self, caplog, workflow, plugin_args,
...
-
-        assert plugin_result['composes'] == composes
-        if platforms:
-            for compose in composes:
-                assert compose['arches'].sort() == platforms.sort()
-        else:
-            assert compose['arches'] is None
-
-    def run_plugin_with_args(self, workflow, plugin_args=None, expect_result=None,
+    def run_plugin_with_args(self, workflow, plugin_args=None,
```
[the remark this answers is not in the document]
mlangsdorf: you are the worst.

twaugh:
Just needs stickler-ci appeased I think.

mlangsdorf:
added to release notes
- support for content_sets.yaml and pulp repositories in `resolve_composes`
under # Improvements

PR #943, opened 2018-03-14
source_registry fallback has to be RegistryURI not string

rcerven:


lcarva:

PR #944, opened 2018-03-16
use correct manifest list digest for worker build

rcerven:

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -22,6 +22,7 @@ import logging
 from datetime import timedelta
 import datetime as dt
 import copy
+import platform
 
 from atomic_reactor.build import BuildResult
 from atomic_reactor.plugin import BuildStepPlugin
@@ -32,9 +33,10 @@ from atomic_reactor.plugins.pre_reactor_config import (get_config,
                                                        get_source_registry, get_sources_command,
                                                        get_artifacts_allowed_domains,
                                                        get_yum_proxy, get_image_equal_labels,
-                                                       get_content_versions)
+                                                       get_content_versions, get_openshift_session,
+                                                       get_platform_descriptors)
 from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
-from atomic_reactor.util import df_parser, get_build_json
+from atomic_reactor.util import df_parser, get_build_json, get_manifest_list, ImageName
 from atomic_reactor.constants import (PLUGIN_ADD_FILESYSTEM_KEY, PLUGIN_BUILD_ORCHESTRATE_KEY,
                                       REPO_CONTAINER_CONFIG)
 from osbs.api import OSBS
@@ -88,6 +90,10 @@ class UnknownPlatformException(Exception):
 
 class AllClustersFailedException(Exception):
     """ Each cluster has reached max_cluster_fails """
+
+
+class UnknownKindException(Exception):
+    """ Build image from contains unknown kind """
 
 
 class ClusterRetryContext(object):
@@ -252,7 +258,9 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
                  config_kwargs=None,
                  find_cluster_retry_delay=FIND_CLUSTER_RETRY_DELAY,
                  failure_retry_delay=FAILURE_RETRY_DELAY,
-                 max_cluster_fails=MAX_CLUSTER_FAILS):
+                 max_cluster_fails=MAX_CLUSTER_FAILS,
+                 url=None, verify_ssl=True, use_auth=True,
+                 goarch=None):
         """
         constructor
 
@@ -268,6 +276,7 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
         :param failure_retry_delay: the delay in seconds to try again starting a build
         :param max_cluster_fails: the maximum number of times a cluster can fail before being
                                   ignored
+        :param goarch: dict, keys are platform, values are go language platform names
         """
         super(OrchestrateBuildPlugin, self).__init__(tasker, workflow)
         self.platforms = set(platforms)
@@ -286,11 +295,25 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
         self.fs_task_id = self.get_fs_task_id()
         self.release = self.get_release()
 
+        self.plat_des_fallback = []
+        for plat, architecture in (goarch or {}).items():
+            plat_dic = {'platform': plat,
+                        'architecture': architecture}
+            self.plat_des_fallback.append(plat_dic)
+
+        self.openshift_fallback = {
+            'url': url,
+            'insecure': not verify_ssl,
+            'auth': {'enable': use_auth}
+        }
+
         if worker_build_image:
             self.log.warning('worker_build_image is deprecated')
 
         self.worker_builds = []
         self.namespace = get_build_json().get('metadata', {}).get('namespace', None)
+        self.build_image_digests = {}  # by platform
+        self._openshift_session = None
 
     def adjust_config_kwargs(self):
         koji_fallback = {
@@ -426,6 +449,11 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
         kwargs['conf_section'] = cluster.name
         if self.osbs_client_config:
             kwargs['conf_file'] = os.path.join(self.osbs_client_config, 'osbs.conf')
+
+        if platform in self.build_image_digests:
+            kwargs['build_image'] = self.build_image_digests[platform]
+        else:
+            raise RuntimeError("build_image for platform '%s' not available" % platform)
 
         osbs = self._get_openshift_session(kwargs)
 
@@ -639,10 +667,14 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
                     ctx.try_again_later(self.failure_retry_delay)
                     # this will put the cluster in retry-wait when get_clusters runs
 
-    def set_build_image(self):
-        """
-        Overrides build_image for worker, to be same as in orchestrator build
-        """
+    @property
+    def openshift_session(self):
+        if not self._openshift_session:
+            self._openshift_session = get_openshift_session(self.workflow, self.openshift_fallback)
+
+        return self._openshift_session
+
+    def get_current_buildimage(self):
         spec = get_build_json().get("spec")
         try:
             build_name = spec['strategy']['customStrategy']['from']['name']
@@ -651,13 +683,168 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
             raise RuntimeError("Build object is malformed, failed to fetch buildroot image")
 
         if build_kind == 'DockerImage':
-            self.config_kwargs['build_image'] = build_name
+            return build_name
         else:
             raise RuntimeError("Build kind isn't 'DockerImage' but %s" % build_kind)
 
+    def process_image_from(self, image_from):
+        build_image = None
+        imagestream = None
+
+        if image_from['kind'] == 'DockerImage':
+            build_image = image_from['name']
+        elif image_from['kind'] == 'ImageStreamTag':
+            imagestream = image_from['name']
+        else:
+            raise UnknownKindException
+
+        return build_image, imagestream
+
+    def check_manifest_list(self, build_image, orchestrator_platform, platforms,
+                            current_buildimage):
+        registry_name, image = build_image.split('/', 1)
+        repo, tag = image.rsplit(':', 1)
+
+        registry = ImageName(registry=registry_name, repo=repo, tag=tag)
+        manifest_list = get_manifest_list(registry, registry_name, insecure=True)
+
+        # we don't have manifest list, but we want to build on different platforms
+        if not manifest_list:
+            raise RuntimeError("Buildroot image tag isn't manifest list,"
+                               " which is needed for specified arch")
+        arch_digests = {}
+        image_name = build_image.rsplit(':', 1)[0]
+
+        for manifest in manifest_list.content['manifests']:
+            arch = manifest['platform']['architecture']
+            arch_digests[arch] = image_name + '@' + manifest['digest']
+
+        platform_descriptors = get_platform_descriptors(self.workflow,
+                                                        self.plat_des_fallback)
+        plat_by_arch = {plat['architecture']: plat['platform']
+                        for plat in platform_descriptors}
+        for arch in arch_digests:
+            if arch not in plat_by_arch:
+                plat_by_arch[arch] = arch
+
+        self.build_image_digests = {plat_by_arch[arch]: arch_digests[arch]
+                                    for arch in arch_digests}
+
+        # orchestrator platform is in manifest list
+        if orchestrator_platform in self.build_image_digests:
+            if self.build_image_digests[orchestrator_platform] != current_buildimage:
+                raise RuntimeError("Orchestrator is using image '%s' which isn't"
+                                   " in manifest list" % current_buildimage)
+
+        else:
+            raise RuntimeError("Platform for orchestrator '%s' isn't in manifest list"
+                               % orchestrator_platform)
+
+    def get_image_info_from_buildconfig(self):
+        status = get_build_json().get("status", {})
+
+        if 'config' not in status:
+            return None, None
+
+        config = status['config']
+        if config['kind'] == 'BuildConfig':
+            build_config_name = config['name']
+        else:
+            raise RuntimeError("Build config type isn't BuildConfig : %s" % config['kind'])
+
+        try:
+            build_config = self.openshift_session.os.get_build_config(build_config_name)
+        except OsbsException:
+            raise RuntimeError("Build config not found : %s" % build_config_name)
+
+        try:
+            build_from = build_config['spec']['strategy']['customStrategy']['from']
+        except KeyError:
+            raise RuntimeError("BuildConfig object is malformed")
+
+        try:
+            return self.process_image_from(build_from)
+        except UnknownKindException:
+            raise RuntimeError("BuildConfig object has unknown 'kind' %s" % build_from['kind'])
+
+    def get_image_info_from_annotations(self):
+        annotations = get_build_json().get("metadata", {}).get('annotations', {})
+        if 'from' in annotations:
+            scratch_from = json.loads(annotations['from'])
+
+            try:
+                return self.process_image_from(scratch_from)
+            except UnknownKindException:
+                raise RuntimeError("Build annotation has unknown 'kind' %s" %
+                                   scratch_from['kind'])
+        else:
+            raise RuntimeError("Build wasn't created from BuildConfig and neither"
+                               " has 'from' annotation, which is needed for specified arch")
+
+    def get_build_image_from_imagestream(self, imagestream):
+        try:
+            tag = self.openshift_session.get_image_stream_tag(imagestream).json()
+        except OsbsException:
+            raise RuntimeError("ImageStreamTag not found %s" % imagestream)
+
+        try:
+            tag_image = tag['image']['dockerImageReference']
+        except KeyError:
+            raise RuntimeError("ImageStreamTag is malformed %s" % imagestream)
+
+        if '@sha256:' in tag_image:
+            try:
+                labels = tag['image']['dockerImageMetadata']['Config']['Labels']
+            except KeyError:
+                raise RuntimeError("Image in imageStreamTag '%s' is missing Labels" %
+                                   imagestream)
+
+            release = labels['release']
+            version = labels['version']
+            docker_tag = "%s-%s" % (version, release)
+            return tag_image[:tag_image.find('@sha256')] + ':' + docker_tag
+        else:
+            return tag_image
+
+    def set_build_image(self, platforms):
+        """
+        Overrides build_image for worker, to be same as in orchestrator build
+        """
+        current_platform = platform.processor()
+        orchestrator_platform = current_platform or 'x86_64'
+        current_buildimage = self.get_current_buildimage()
+
+        # orchestrator platform is same as platform on which we want to built on,
+        # so we can use the same image
+        if platforms == set([orchestrator_platform]):
+            self.build_image_digests[orchestrator_platform] = current_buildimage
+            return
+
+        # BuildConfig exists
+        build_image, imagestream = self.get_image_info_from_buildconfig()
+        if not (build_image or imagestream):
+            # get image build from build metadata, which is set for direct builds
+            # this is explicitly set by osbs-client, it isn't default OpenShift behaviour
+            build_image, imagestream = self.get_image_info_from_annotations()
+
+        # if imageStream is used
+        if imagestream:
+            build_image = self.get_build_image_from_imagestream(imagestream)
+
+        # we have build_image with sha, but we need tag to check for manifest list,
+        # as we want to build on different platforms
+        if build_image and '@sha256:' in build_image:
+            raise RuntimeError("Buildroot image doesn't have tag,"
+                               " which is needed for specified arch")
+
+        # we have build_image with tag, so we can check for manifest list
+        if build_image:
+            self.check_manifest_list(build_image, orchestrator_platform,
+                                     platforms, current_buildimage)
+
     def run(self):
-        self.set_build_image()
         platforms = self.get_platforms()
+        self.set_build_image(platforms)
 
         thread_pool = ThreadPool(len(platforms))
         result = thread_pool.map_async(self.select_and_start_cluster, platforms)
--- a/atomic_reactor/plugins/post_group_manifests.py
+++ b/atomic_reactor/plugins/post_group_manifests.py
@@ -65,9 +65,9 @@ class GroupManifestsPlugin(PostBuildPlugin):
         self.group = get_group_manifests(self.workflow, group)
 
         plat_des_fallback = []
-        for platform in goarch:
+        for platform, architecture in (goarch or {}).items():
             plat_dic = {'platform': platform,
-                        'architecture': goarch[platform]}
+                        'architecture': architecture}
             plat_des_fallback.append(plat_dic)
 
         platform_descriptors = get_platform_descriptors(self.workflow, plat_des_fallback)
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -831,6 +831,41 @@ def manifest_is_media_type(response, media_type):
     return response_h_prefix == request_h_prefix
 
 
+def get_manifest(image, registry_session, version):
+    saved_not_found = None
+    media_type = get_manifest_media_type(version)
+    try:
+        response = query_registry(registry_session, image, digest=None, version=version)
+    except (HTTPError, RetryError, Timeout) as ex:
+        if ex.response.status_code == requests.codes.not_found:
+            saved_not_found = ex
+        # If the registry has a v2 manifest that can't be converted into a v1
+        # manifest, the registry fails with status=400 (BAD_REQUEST), and an error code of
+        # MANIFEST_INVALID. Note that if the registry has v2 manifest and
+        # you ask for an OCI manifest, the registry will try to convert the
+        # v2 manifest into a v1 manifest as the default type, so the same
+        # thing occurs.
+        if version != 'v2' and ex.response.status_code == requests.codes.bad_request:
+            logger.warning('Unable to fetch digest for %s, got error %s',
+                           media_type, ex.response.status_code)
+            return None, saved_not_found
+        # Returned if the manifest could not be retrieved for the given
+        # media type
+        elif (ex.response.status_code == requests.codes.not_found or
+              ex.response.status_code == requests.codes.not_acceptable):
+            logger.debug("skipping version %s due to status code %s",
+                         version, ex.response.status_code)
+            return None, saved_not_found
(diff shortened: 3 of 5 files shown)


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -650,13 +660,159 @@ def set_build_image(self):
...
+            raise RuntimeError("Buildroot image doesn't have tag, which is needed for specified arch")
+
+        # we have build_image with tag, so we can check for manifest list
+        if build_image:
+            registry_name, image = build_image.split('/', 1)
+            registry_name = registry_name.split(':')[0]
+            repo, tag = image.rsplit(':', 1)
+
+            if 'pulp' in registry_name:
+                registry_name += '/pulp/docker'
```
twaugh: Wait, what? That shouldn't be needed. Just use the Docker Registry HTTP API against the registry and port.

rcerven: yeah I'll change that, as we spoke on IRC

atomic_reactor/util.py
```diff
@@ -916,6 +916,41 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
     return ManifestDigest(**digests)
 
 
+def get_manifest_list(image, registry, insecure=False, dockercfg_path=None):
+    """Return manifest digest for image.
```
twaugh: s/digest/list/

atomic_reactor/util.py
```diff
@@ -916,6 +916,41 @@ def get_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
     return ManifestDigest(**digests)
 
 
+def get_manifest_list(image, registry, insecure=False, dockercfg_path=None):
```
twaugh: Any possibility this could take a `version` parameter and be used by `get_manifest_digests()`, to reduce duplication?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -650,13 +660,158 @@ def set_build_image(self):
...
-            self.config_kwargs['build_image'] = build_name
+            return build_name
         else:
             raise RuntimeError("Build kind isn't 'DockerImage' but %s" % build_kind)
 
+    def set_build_image(self, platforms):
+        """
+        Overrides build_image for worker, to be same as in orchestrator build
+        """
+        orchestrator_plat = 'x86_64'
```
twaugh: Let's say `platform` instead of `plat` to reduce eye friction when reading the code.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -650,13 +660,158 @@ def set_build_image(self):
...
+        status = get_build_json().get("status")
+        build_config_name = None
+        build_image = None
+        imagestream = None
+        need_manifest = not (orchestrator_plat in platforms and len(platforms) == 1)
```
twaugh: `need_manifest_list`

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -650,13 +660,158 @@ def set_build_image(self):
...
+        use_same_image = False
+        current_buildimage = self.get_current_buildimage()
+        self.plat_digests = {}
```
twaugh: `platform_digests`, and should we initialize this in `__init__`?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -650,13 +660,158 @@ def set_build_image(self):
...
+            try:
+                build_name = build_config['spec']['strategy']['customStrategy']['from']['name']
+                build_kind = build_config['spec']['strategy']['customStrategy']['from']['kind']
+            except KeyError:
+                raise RuntimeError("BuildConfig object is malformed")
+
+            if build_kind == 'DockerImage':
+                build_image = build_name
+            elif build_kind == 'ImageStreamTag':
+                imagestream = build_name
```
twaugh: Can some of this be factored out to avoid repeating it? e.g. `build_image, imagestream = process_from(...)`.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -650,13 +660,158 @@ def set_build_image(self):
...
+                build_image = tag_image
+
+        # we don't have build_image and we need it as we want to build on differnt platforms
+        if not build_image and need_manifest:
+            raise RuntimeError("Failed to obtain buidroot image,"
+                               " which is needed for specified arch")
+
+        # we have build_image with sha, but we need tag to check for manifest list,
+        # as we want to build on different platforms
+        if build_image and '@sha256:' in build_image and need_manifest:
```
twaugh: In the future it's possible the digest will actually be for a manifest list, but right now this is fine.

atomic_reactor/util.py

atomic_reactor/util.py
rcerven: mmm and how would I use it by get_manifest_digest() ? as that is collection versions: Docker-Content-Digest dict? and for manifest list we don't have it?

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: ok

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: ok

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: sure

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: ok

atomic_reactor/util.py
rcerven: made general get_manifest function, to reduce duplication

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -650,13 +673,166 @@ def set_build_image(self):
...
+        else:
+            raise UnknownKindException
+
+        return build_image, imagestream
+
+    def set_build_image(self, platforms):
+        """
+        Overrides build_image for worker, to be same as in orchestrator build
+        """
+        orchestrator_platform = 'x86_64'
```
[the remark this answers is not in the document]
rcerven: we agreed wit @lcarva that we will assume that, just so we keep backwards compatibility

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -650,13 +673,166 @@ def set_build_image(self):
...
+                release = labels['release']
+                version = labels['version']
+                docker_tag = "%s-%s" % version, release
+                build_image = tag_image[:tag_image.find('@sha256')] + ':' + docker_tag
+            else:
+                build_image = tag_image
+
+        # we don't have build_image and we need it as we want to build on differnt platforms
+        if not build_image and need_manifest_list:
+            raise RuntimeError("Failed to obtain buidroot image,"
```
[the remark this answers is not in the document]
rcerven: ok

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -286,11 +294,24 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
         self.fs_task_id = self.get_fs_task_id()
         self.release = self.get_release()
 
+        self.plat_des_fallback = []
+        for platform in goarch:
```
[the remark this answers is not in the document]
rcerven: sure

atomic_reactor/plugins/build_orchestrate_build.py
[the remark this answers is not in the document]
lcarva: Would it be saner to dynamically determine this?
```python
>>> import platform
>>> platform.processor()
'x86_64'
```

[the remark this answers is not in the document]
rcerven: ok using that, and fallback to 'x86_64'

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -286,11 +295,24 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
...
+            'insecure': not verify_ssl,
+            'auth': {'enable': use_auth}
+        }
+
         if worker_build_image:
             self.log.warning('worker_build_image is deprecated')
 
         self.worker_builds = []
         self.namespace = get_build_json().get('metadata', {}).get('namespace', None)
+        self.plat_digests = {}
```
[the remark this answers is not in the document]
rcerven: we can, but I wanted just to point out that it's keys are platforms

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -286,11 +295,24 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
         self.fs_task_id = self.get_fs_task_id()
         self.release = self.get_release()
 
+        self.plat_des_fallback = []
+        for plat, architecture in goarch.items():
+            plat_dic = {'platform': plat,
+                        'architecture': architecture}
+            self.plat_des_fallback.append(plat_dic)
```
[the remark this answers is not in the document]
rcerven: we are building fallback for get_platform_descriptors, which is list of dicts

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -286,11 +295,24 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
         self.fs_task_id = self.get_fs_task_id()
         self.release = self.get_release()
 
+        self.plat_des_fallback = []
+        for plat, architecture in goarch.items():
```
twaugh: Needs `if goarch:` in front, or else `for ... in (goarch or {}).items():`

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -429,6 +451,11 @@ def get_cluster_info(self, cluster, platform):
         if self.osbs_client_config:
             kwargs['conf_file'] = os.path.join(self.osbs_client_config, 'osbs.conf')
 
+        if platform in self.build_image_digests:
+            kwargs['build_image'] = self.build_image_digests[platform]
+        else:
+            raise RuntimeError("We don't have build_image for platform %s", platform)
```
lcarva: This error message sounds odd. Maybe: `build_image for platform '%s' not available`

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -650,13 +674,167 @@ def set_build_image(self):
...
+            raise UnknownKindException
+
+        return build_image, imagestream
+
+    def set_build_image(self, platforms):
+        """
+        Overrides build_image for worker, to be same as in orchestrator build
+        """
+        current_platform = platform.processor()
+        orchestrator_platform = current_platform if current_platform else 'x86_64'
```
lcarva: Why not simply `orchestrator_platform = platform.processor() or 'x86_64'` ?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -650,13 +674,167 @@ def set_build_image(self):
...
+        status = get_build_json().get("status")
+        build_config_name = None
+        build_image = None
+        imagestream = None
+        need_manifest_list = not (orchestrator_platform in platforms and len(platforms) == 1)
```
lcarva: Simplified alternative: `need_manifest_list = platforms != [orchestrator_platform]`

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -650,13 +674,167 @@ def set_build_image(self):
...
+
+        # BuildConfig exists
+        if 'config' in status:
+            config = status['config']
+            if config['kind'] == 'BuildConfig':
+                build_config_name = config['name']
+            else:
+                raise RuntimeError("Build config type isn't BuildConfig : %s" % config['kind'])
+
+        # get image build from build metadata, which is set for direct builds
```
lcarva: Let's clarify that this is "set by osbs-client" to avoid confusion with OpenShift behavior.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -650,13 +674,167 @@ def set_build_image(self):
...
+
+        return build_image, imagestream
+
+    def set_build_image(self, platforms):
+        """
+        Overrides build_image for worker, to be same as in orchestrator build
+        """
+        current_platform = platform.processor()
+        orchestrator_platform = current_platform if current_platform else 'x86_64'
+        status = get_build_json().get("status")
```
twaugh: If this is `None`, `if 'config' in status:` below will blow up.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -650,13 +674,167 @@ def set_build_image(self):
...
+
+            if 'from' in metadata['annotations']:
+                scratch_from = json.loads(metadata['annotations']['from'])
+
+                try:
+                    build_image, imagestream = self.process_image_from(scratch_from)
+                except UnknownKindException:
+                    raise RuntimeError("Build annotation has unknown 'kind' %s" %
+                                       scratch_from['kind'])
+            elif not need_manifest_list:
```
lcarva: Should this be the first check this function does? If it doesn't need manifest list, just use the same image and bail out?

This will also ensure that this plugin does not always require an OpenShift session, which is really important for backwards compatibility.

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: well not sure if it is possible for status of build to be None, but ok will make it default {}

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: ok

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: sure

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: ok

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: well not really, IF there is manifest list used, I'd like to actually verify that we are using correct digest from it

lcarva: @rcerven pointed out that the story for this (OSBS-5221) actually calls out for always making this check. I think this was an oversight. @twaugh, think we should only bother with this new behavior if manifest list is actually required (at least one worker build uses different arch than orchestrator build) ?

I'm concerned in making the new plugin parameters for building openshift session required for all cases.

twaugh: @lcarva yes, I agree. If we do not need the manifest list (because there is a single worker, for the same architecture as the orchestrator), let's skip the additional processing.

rcerven: ack

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -651,13 +675,158 @@ def set_build_image(self):
...
+            if config['kind'] == 'BuildConfig':
+                build_config_name = config['name']
+            else:
+                raise RuntimeError("Build config type isn't BuildConfig : %s" % config['kind'])
+
+        # get image build from build metadata, which is set for direct builds
+        # this is explicitly set by osbs-client, it isn't default OpenShift behaviour
+        else:
+            metadata = get_build_json().get("metadata")
+            if 'from' in metadata['annotations']:
```
twaugh: 'annotations' is not a required property of v1.ObjectMeta, so maybe:
```
annotations = get_build_json().get('metadata', {}).get('annotations', {})
if 'from' in annotations:
```

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -651,13 +675,158 @@ def set_build_image(self):
...
+                tag = osbs.get_image_stream_tag(imagestream).json()
+            except OsbsException:
+                raise RuntimeError("ImageStreamTag not found %s" % imagestream)
+
+            try:
+                tag_image = tag['image']['dockerImageReference']
+            except KeyError:
+                raise RuntimeError("ImageStreamTag is malformed %s" % imagestream)
+
+            if '@sha256:' in tag_image:
```
twaugh: Minor (no action): 'sha256:' is part of the digest and in the future a different algorithm may be used. However, let's hope ImageStreams gain manifest list support before then!

tests/plugins/test_orchestrate_build.py
```diff
@@ -11,7 +11,7 @@
 from atomic_reactor.core import DockerTasker
 from atomic_reactor.inner import DockerBuildWorkflow
 from atomic_reactor.plugin import BuildCanceledException, PluginFailedException
-from atomic_reactor.plugin import BuildStepPluginsRunner
+from atomic_reactor.plugin import BuildStepPluginsRunner, PluginsRunner, PreBuildPluginsRunner
```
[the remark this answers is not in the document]
twaugh: Seems like sticker is right.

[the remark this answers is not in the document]
rcerven: yeah tests are still in progress

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: ok

tests/plugins/test_orchestrate_build.py
```diff
@@ -351,6 +369,8 @@ def test_orchestrate_build(tmpdir, caplog, config_kwargs,
                     namespace = {name}_namespace
                     """.format(name=cluster['name'])))
 
+    goarch = {'x86_64': 'amd64', 'ppc64le': 'ppc64'}
```
twaugh: I think it doesn't matter because this is just test data, but the GOARCH for 'ppc64le' is 'ppc64le' so we wouldn't bother including that key. A likely-looking dict would be `{'x86_64': 'amd64', 'aarch64': 'arm64'}`.

tests/plugins/test_orchestrate_build.py
```diff
@@ -97,8 +107,10 @@ def mock_workflow(tmpdir):
     df = df_parser(df_path)
     setattr(workflow.builder, 'df_path', df.dockerfile_path)
 
-    build = '{"spec": {"strategy": {"customStrategy": {"from": \
-            {"name": "some_image:latest", "kind": "DockerImage"}}}}}'
+    build = '{"spec": {"strategy": {"customStrategy": {"from":' + \
+            '{"name": "registry/some_image@sha256:123456", "kind": "DockerImage"}}}},' + \
+            '"metadata": {"annotations": {"from": "{\\"name\\": \\"registry' + \
+            '/some_image:latest\\", \\"kind\\": \\"DockerImage\\"}"}}}'
```
twaugh: There will come a point where it might be easier to just use `json.dumps()` on a dict to get these strings.

tests/plugins/test_orchestrate_build.py
rcerven: well even if it is same it has to be in goarch, or we will need to change logic how we use it, that is used in group_manifest plugin as well

twaugh: No, definitely not:
https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/post_group_manifests.py#L236

This is particularly important, because we do not have a platform descriptor for ppc64le because it doesn't need one.

tests/plugins/test_orchestrate_build.py
```diff
@@ -1138,48 +1182,144 @@ def test_orchestrate_override_build_kwarg(tmpdir, overrides):
...
+
+    ('{"spec": {"strategy": {"customStrategy": {"from": \
+     {"name": "osbs-buildroot:latest", "kind": "DockerImage"}}}}, \
+     "status": {"config": {"kind": "BuildConfig", "name": "build config"}}}',
+     "Buildroot image doesn't have tag, which is needed for specified arch",
+     True, {"spec": {"strategy": {"customStrategy": {"from": {"kind": "ImageStreamTag",
+                                                              "name": "ims"}}}}},
+     True, {"image": {"dockerImageReference": "some@sha256:12345",
+                      "dockerImageMetadata": {
+                          "Config": {"Labels": {"release": "@sha256:123",
```
twaugh: The 'release' label won't look like that though, more like '4'.

tests/plugins/test_orchestrate_build.py
rcerven: ok will change how to use platform_descriptors and default to platform if not there

tests/plugins/test_orchestrate_build.py
rcerven: yes I know, but I need scenario where I have @256 and this was way how to get it

rcerven: I'll change it, to get it from annotations

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -65,9 +65,9 @@ def __init__(self, tasker, workflow, registries=None, group=True, goarch=None):
         self.group = get_group_manifests(self.workflow, group)
 
         plat_des_fallback = []
-        for platform in goarch:
+        for platform, architecture in (goarch or {}).items():
             plat_dic = {'platform': platform,
-                        'architecture': goarch[platform]}
+                        'architecture': architecture}
```
mlangsdorf: what's the reason for this change? if you want to handle the case where goarch=None, couldn't this change just as easily be 'for platform in goarch or {}:'
I don't really object, I'm just curious about why one form is preferred over the other.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -651,13 +675,161 @@ def set_build_image(self):
...
+            self.build_image_digests[orchestrator_platform] = current_buildimage
+            return
+
+        # BuildConfig exists
+        if 'config' in status:
+            config = status['config']
+            if config['kind'] == 'BuildConfig':
+                build_config_name = config['name']
+            else:
+                raise RuntimeError("Build config type isn't BuildConfig : %s" % config['kind'])
```
mlangsdorf: since set_build_image is still a huge function and status isn't used for anything other than getting the config, I think this can safely be turned into a utility function, ie:
`
try:
    build_config_name = self.get_build_config_name()
except KeyError:
   the current else block
`

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -651,13 +675,161 @@ def set_build_image(self):
...
+                try:
+                    build_image, imagestream = self.process_image_from(scratch_from)
+                except UnknownKindException:
+                    raise RuntimeError("Build annotation has unknown 'kind' %s" %
+                                       scratch_from['kind'])
+            else:
+                raise RuntimeError("Build wasn't created from BuildConfig and neither"
+                                   " has 'from' annotation, which is needed for specified arch")
+
+        osbs = get_openshift_session(self.workflow, self.openshift_fallback)
```
mlangsdorf: can this be moved up to be above the check for config in status? it doesn't look like it depends on the results of the if else block.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -651,13 +675,161 @@ def set_build_image(self):
...
+
+            try:
+                build_from = build_config['spec']['strategy']['customStrategy']['from']
+            except KeyError:
+                raise RuntimeError("BuildConfig object is malformed")
+
+            try:
+                build_image, imagestream = self.process_image_from(build_from)
+            except UnknownKindException:
+                raise RuntimeError("BuildConfig object has unknown 'kind' %s" % build_from['kind'])
```
mlangsdorf: Since you only get a build_config_name if config is in status, this block could be moved up to be part of the if config in status block, and then moved to the function I proposed earlier.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -651,13 +675,161 @@ def set_build_image(self):
...
+            else:
+                build_image = tag_image
+
+        # we have build_image with sha, but we need tag to check for manifest list,
+        # as we want to build on different platforms
+        if build_image and '@sha256:' in build_image:
+            raise RuntimeError("Buildroot image doesn't have tag,"
+                               " which is needed for specified arch")
+
+        # we have build_image with tag, so we can check for manifest list
```
mlangsdorf: is it worthwhile splitting this entire block off into its own function? it's clearly very self-contained.

tests/plugins/test_orchestrate_build.py
mlangsdorf: I think we've past that point.

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: I'm also not sure if we actually need to perform these checks. If 'config' is set to something other than `BuildConfig` object there's something seriously wrong in OpenShift. 

But anyways, having a utility method that encapsulates this would help in reducing the size of this method. I'd even go further, and just let `get_build_config_name` that Mark suggested raise the RuntimeError with the helpful message. In other words, this whole block turns into:
```python
build_config_name = self.get_build_config_name()  # also fetches the status
```

`get_build_config_name` returns `None` if `config not in status`; or the actual build config name; or raises RuntimeError in case of data error.

atomic_reactor/plugins/build_orchestrate_build.py
lcarva: Could turn into a property so it's only initialized when needed, similar to how `resolve_composes` plugin handles `odcs_client` and `koji_client`.

atomic_reactor/plugins/post_group_manifests.py
rcerven: the same change was done in orchestrator, as Tim wanted to use items() since we are using both key&value

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: at this point we certainly need that session, in other cases it would raise and we wouldn't be at this point

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: mmm yeah

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: yeah

atomic_reactor/plugins/build_orchestrate_build.py
rcerven: ok will make it property

atomic_reactor/plugins/build_orchestrate_build.py
mlangsdorf: I think you can tighten that up more:
build_image, imagestream = get_image_info_from_config()
has the current get_build_config_name and all of the code in the current if build_config_name: block

```
-       if build_config_name:
-       else:
+      if not (build_image or imagetream):
```

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -651,13 +683,169 @@ def set_build_image(self):
...
+        repo, tag = image.rsplit(':', 1)
+
+        registry = ImageName(registry=registry_name, repo=repo, tag=tag)
+        manifest_list = get_manifest_list(registry, registry_name, insecure=True)
+
+        # we don't have manifest list, but we want to build on different platforms
+        if not manifest_list:
+            raise RuntimeError("Buildroot image tag isn't manifest list,"
+                               " which is needed for specified arch")
+        else:
```
lcarva: Unnecessary `else`.

rcerven: ok

tests/plugins/test_orchestrate_build.py
lcarva: Can we also remove `ppc64le` item to ensure the code is not requiring it?

tests/plugins/test_orchestrate_build.py
```diff
@@ -448,7 +488,8 @@ def mock_wait_for_build_to_finish(build_name):
                 'build_kwargs': make_worker_build_kwargs(),
                 'osbs_client_config': str(tmpdir),
                 'max_cluster_fails': 2,
-                'unreachable_cluster_retry_delay': .1
+                'unreachable_cluster_retry_delay': .1,
+                'goarch': {'x86_64': 'amd64', 'ppc64le': 'ppc64'},
```
lcarva: Can we also remove `ppc64le` item to ensure the code is not requiring it? Same for other cases.

tests/plugins/test_orchestrate_build.py
rcerven: ok

lcarva:
Updated release notes draft:
> - use correct manifest list digest for worker build image

PR #945, opened 2018-03-19
pulp_pull: expect v2 schema 2 manifest list when appropriate (OSBS-5275)

twaugh:
If the group_manifests plugin ran and had a truthy result, this means a v2 schema 2 manifest list is expected. Wait for this.


mlangsdorf:
added to the release notes
- pulp_pull will retrieve v2 schema 2 manifest lists
under # Improvements

PR #946, opened 2018-03-21
fix odcs pulp composes

mlangsdorf:
Fix the broken pulp composes in pre_resolve_composes by changing the way that signing intent is determined.

Change ODCSConfig in pre_reactor_config to return the most restrictive signing intent that
includes all the requested keys, without requiring that that the requested keys match all
the keys in the signing intent.

Edit some tests casts in test_resolve_composes to handle pulp composes' signing keys.

### DIFF
--- a/tests/plugins/test_reactor_config.py
+++ b/tests/plugins/test_reactor_config.py
@@ -320,9 +320,9 @@ class TestReactorConfigPlugin(object):
                 odcs:
                    signing_intents:
                    - name: release
-                     keys: [R123]
+                     keys: [R123, R234]
                    - name: beta
-                     keys: [R123, B456]
+                     keys: [R123, B456, B457]
                    - name: unsigned
                      keys: []
                    default_signing_intent: {default}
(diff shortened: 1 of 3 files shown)


tests/plugins/test_reactor_config.py
```diff
@@ -320,9 +320,9 @@ def test_odcs_config(self, tmpdir, default):
                 odcs:
                    signing_intents:
                    - name: release
-                     keys: [R123]
+                     keys: [R123, R234]
                    - name: beta
-                     keys: [R123, B456]
+                     keys: [R123, B456, B457]
```
lcarva: Usually, beta includes all the keys from release. Let's add `R234` here as well.

mlangsdorf: they're kind of arbitrary strings at this point, and I liked having a test for a set of signing keys that weren't quite in release and weren't quite in beta.
but if that's not a useful thing to check, I will add R234 to beta.

lcarva: Ok

lcarva:
Waiting for tests results before merging this.

csomh:
Test results are ok.

lcarva:

PR #947, opened 2018-03-22
manifest list returns string, so we need json()

rcerven:


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -715,7 +715,8 @@ def check_manifest_list(self, build_image, orchestrator_platform, platforms,
         arch_digests = {}
         image_name = build_image.rsplit(':', 1)[0]
 
-        for manifest in manifest_list.content['manifests']:
+        manifest_list_dict = json.loads(manifest_list.content)
```
lcarva: Can't you just call: `manifest_list.json()` ?

rcerven:
not updating release note, as it is fix for unreleased feature

PR #948, opened 2018-03-22
Add clusters_client_config_dir to reactor-config

lcarva:
This new config is used to determine the location of osbs.conf file to
be used when communicating with worker clusters.

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -34,7 +34,8 @@ from atomic_reactor.plugins.pre_reactor_config import (get_config,
                                                        get_artifacts_allowed_domains,
                                                        get_yum_proxy, get_image_equal_labels,
                                                        get_content_versions, get_openshift_session,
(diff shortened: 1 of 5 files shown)


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -281,6 +282,7 @@ def __init__(self, tasker, workflow, platforms, build_kwargs,
         super(OrchestrateBuildPlugin, self).__init__(tasker, workflow)
         self.platforms = set(platforms)
         self.build_kwargs = build_kwargs
+        self.osbs_client_config_fallback = osbs_client_config
         self.osbs_client_config = osbs_client_config
```
mlangsdorf: self.osbs_client_config is now a dead parameter that never gets used. Let's delete it.

lcarva: Good catch, for forgot to remove it after debugging. Fixed

lcarva:
Updated release notes draft:
> - Add `clusters_client_config_dir` to reactor config

PR #949, opened 2018-03-22
don't check orchestrator image in manifest list, if it doesn't contai…

rcerven:
…n sha

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -732,14 +732,14 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
                                     for arch in arch_digests}
 
         # orchestrator platform is in manifest list
-        if orchestrator_platform in self.build_image_digests:
-            if self.build_image_digests[orchestrator_platform] != current_buildimage:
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -733,7 +733,8 @@ def check_manifest_list(self, build_image, orchestrator_platform, platforms,
 
         # orchestrator platform is in manifest list
         if orchestrator_platform in self.build_image_digests:
-            if self.build_image_digests[orchestrator_platform] != current_buildimage:
+            if '@sha256:' in current_buildimage and \
```
lcarva: Can we change these two checks around so it's a bit easier to understand?
```python
if orchestrator_platform not in self.build_image_digests:
  raise RuntimeError("Platform for orchestrator '%s' isn't in manifest list" 
                     % orchestrator_platform)

if ('@sha256:' in current_buildimag and 
    self.build_image_digests[orchestrator_platform] != current_buildimage):
  raise RuntimeError("Orchestrator is using image digest '%s' which isn't"
                     " in manifest list" % current_buildimage)
```

Note that we should probably change the error message to specify that orchestrator is using image digest.

rcerven:
not updating release note as it is fix for unreleased feature

PR #950, opened 2018-03-22
Atomic reactor was sending a repo object

bfontecc007:
Atomic reactor was sending a repo object to dockpulp's updateRepo method instead of sending an id.


mlangsdorf:
added to release notes under # Bug Fixes
- repo id, not the full repo, is now sent to dockpulp.updateRepo so repos are actually updated to turn off autopublish

PR #951, opened 2018-03-26
when now platforms enabled, fail with explicit message

rcerven:


lcarva:
What's up with stickler?

lcarva:
Updated release notes draft:
> - `orchestrate_build` raises a more descriptive error when no platforms are enabled

PR #952, opened 2018-03-27
Arrangement 6 fixes

mlangsdorf:
A handful of fixes to make Arrangement6 work correctly.

Necessary for https://github.com/projectatomic/osbs-client/pull/745 to work.

### DIFF
--- a/atomic_reactor/plugins/pre_add_yum_repo_by_url.py
+++ b/atomic_reactor/plugins/pre_add_yum_repo_by_url.py
@@ -94,7 +94,7 @@ class AddYumRepoByUrlPlugin(PreBuildPlugin):
     key = "add_yum_repo_by_url"
     is_allowed_to_fail = False
 
-    def __init__(self, tasker, workflow, repourls, inject_proxy=None):
+    def __init__(self, tasker, workflow, repourls=None, inject_proxy=None):
         """
         constructor
 
@@ -105,7 +105,7 @@ class AddYumRepoByUrlPlugin(PreBuildPlugin):
         """
         # call parent constructor
         super(AddYumRepoByUrlPlugin, self).__init__(tasker, workflow)
-        self.repourls = repourls
+        self.repourls = repourls or []
         self.inject_proxy = inject_proxy
 
     def run(self):
--- a/atomic_reactor/plugins/pre_reactor_config.py
+++ b/atomic_reactor/plugins/pre_reactor_config.py
@@ -6,6 +6,7 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
 
+from copy import deepcopy
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.util import read_yaml, read_yaml_from_file_path, get_build_json
 from osbs.utils import RegistryURI
@@ -39,7 +40,9 @@ def get_config(workflow):
 
 def get_value(workflow, name, fallback):
     try:
-        return get_config(workflow).conf[name]
+        # make a deep copy to prevent plugins from changing the value for other plugins
+        value = deepcopy(get_config(workflow).conf[name])
+        return value
     except KeyError:
         if fallback != NO_FALLBACK:
             return fallback
@@ -332,7 +335,7 @@ class ReactorConfig(object):
     DEFAULT_CONFIG = {ReactorConfigKeys.VERSION_KEY: 1}
 
     def __init__(self, config=None):
-        self.conf = config or self.DEFAULT_CONFIG
+        self.conf = deepcopy(config or self.DEFAULT_CONFIG)
 
         version = self.conf[ReactorConfigKeys.VERSION_KEY]
         if version != 1:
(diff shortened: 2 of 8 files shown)


atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -232,6 +232,28 @@ def get_registries(workflow, fallback=NO_FALLBACK):
...
+        if reguri.version == 'v2':
+            regdict = {
+                'url': registry,
+                'insecure': registry.get('insecure', False)
+            }
+            if registry.get('auth'):
+                regdict['secret'] = registry['auth']['cfg_path']
+            return regdict
+
+    return fallback
```
lcarva: We should probably raise an exception here instead. The `fallback` is only meant to be used if the key is not found in reactor config.

atomic_reactor/plugins/pre_add_yum_repo_by_url.py
```diff
@@ -94,7 +94,7 @@ class AddYumRepoByUrlPlugin(PreBuildPlugin):
     key = "add_yum_repo_by_url"
     is_allowed_to_fail = False
 
-    def __init__(self, tasker, workflow, repourls, inject_proxy=None):
+    def __init__(self, tasker, workflow, repourls=None, inject_proxy=None):
```
lcarva: Curious as to what error this was causing.

I'd think that calling this plugin without `repourls` is not correct.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -332,7 +335,7 @@ class ReactorConfig(object):
     DEFAULT_CONFIG = {ReactorConfigKeys.VERSION_KEY: 1}
 
     def __init__(self, config=None):
-        self.conf = config or self.DEFAULT_CONFIG
+        self.conf = deepcopy(config) or self.DEFAULT_CONFIG
```
lcarva: Should `deepcopy` include default config too? `self.conf = deepcopy(config or self.DEFAULT_CONFIG)`

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -232,6 +232,28 @@ def get_registries(workflow, fallback=NO_FALLBACK):
     return registries_cm
 
 
+def get_docker_registry(workflow, fallback=NO_FALLBACK):
```
lcarva: Missing unit tests

atomic_reactor/plugins/pre_add_yum_repo_by_url.py
mlangsdorf: PluginsConfiguration currently doesn't remove add_yum_repo_by_url if there are valid yum_repourls values. I guess this could be removed and PluginsConfiguration could remove add_yum_repo_by_url - either is good.

atomic_reactor/plugins/pre_reactor_config.py
mlangsdorf: I'm not sure it really matters, but sure.

atomic_reactor/plugins/pre_add_yum_repo_by_url.py
lcarva: This is probably fine.

atomic_reactor/plugins/pre_reactor_config.py
lcarva: It does! Because if you create two instances of `ReactorConfig` without providing a `config`, they'll both point to the same object since `self.DEFAULT_CONFIG` is a class attribute.

mlangsdorf: Fixed.

atomic_reactor/plugins/pre_reactor_config.py
mlangsdorf: Fixed.

atomic_reactor/plugins/pre_reactor_config.py
mlangsdorf: Fixed.

atomic_reactor/plugins/pre_add_yum_repo_by_url.py
mlangsdorf: Leaving alone for now.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -232,6 +235,28 @@ def get_registries(workflow, fallback=NO_FALLBACK):
...
+    except KeyError:
+        if fallback != NO_FALLBACK:
+            return fallback
+        raise
+
+    for registry in all_registries:
+        reguri = RegistryURI(registry.get('url'))
+        if reguri.version == 'v2':
+            regdict = {
+                'url': registry,
```
lcarva: Should this be `reguri`?

lcarva: Ah! It probably works as is because of RegistryURI.__repr__ method, which calls `.uri`. So this should probably be `registry.uri` to avoid confusion.

mlangsdorf: Fixed.

lcarva:

PR #953, opened 2018-03-28
create and use pre_check_and_set_platforms

mlangsdorf:
##add a new plugin that retrieves the koji session information and gets
the platforms from it. The new plugin also checks the repo configuration
file and removes any excluded platforms and limits the result to any only
platforms.

build_orchestrate_build will use the results of the new plugin over the
platforms passed in from USER_PARAMS if available. This will allow
multi-arch autorebuilds to respond to changes in the koji tag.

pre_pull_base_image will also use the results of the new plugin.

Fixes osbs-5230.

requires https://github.com/projectatomic/osbs-client/pull/747 to enable the plugin

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -39,7 +39,7 @@ from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
 from atomic_reactor.util import (df_parser, get_build_json, get_manifest_list, ImageName,
                                  get_platforms_in_limits)
 from atomic_reactor.constants import (PLUGIN_ADD_FILESYSTEM_KEY, PLUGIN_BUILD_ORCHESTRATE_KEY,
-                                      REPO_CONTAINER_CONFIG)
+                                      PLUGIN_CHECK_AND_SET_PLATFORMS_KEY)
 from osbs.api import OSBS
 from osbs.exceptions import OsbsException
 from osbs.conf import Configuration
@@ -416,6 +416,12 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
             get_arrangement_version(self.workflow, self.build_kwargs['arrangement_version'])
 
     def get_platforms(self):
+        koji_platforms = self.workflow.prebuild_results.get(PLUGIN_CHECK_AND_SET_PLATFORMS_KEY)
+        if koji_platforms:
+            return koji_platforms
+
+        # if check_and_set_platforms didn't run, or didn't get any platforms from koji
+        # determine platforms from USER_PARAMS platforms parameter
         return get_platforms_in_limits(self.workflow, self.platforms)
 
     def get_current_builds(self, osbs):
@@ -841,14 +847,13 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
 
     def run(self):
         self.platforms = self.get_platforms()
-        platforms = self.platforms
 
-        if not platforms:
+        if not self.platforms:
             raise RuntimeError("No enabled platform to build on")
-        self.set_build_image(platforms)
+        self.set_build_image(self.platforms)
 
-        thread_pool = ThreadPool(len(platforms))
-        result = thread_pool.map_async(self.select_and_start_cluster, platforms)
+        thread_pool = ThreadPool(len(self.platforms))
+        result = thread_pool.map_async(self.select_and_start_cluster, self.platforms)
 
         try:
             result.get()
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -11,7 +11,6 @@ from collections import namedtuple
 from copy import deepcopy
 from multiprocessing.pool import ThreadPool
 
-import yaml
 import json
 import os
 from operator import attrgetter
@@ -37,7 +36,8 @@ from atomic_reactor.plugins.pre_reactor_config import (get_config,
                                                        get_platform_descriptors,
                                                        get_clusters_client_config_path)
 from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
-from atomic_reactor.util import df_parser, get_build_json, get_manifest_list, ImageName
+from atomic_reactor.util import (df_parser, get_build_json, get_manifest_list, ImageName,
+                                 get_platforms_in_limits)
 from atomic_reactor.constants import (PLUGIN_ADD_FILESYSTEM_KEY, PLUGIN_BUILD_ORCHESTRATE_KEY,
                                       REPO_CONTAINER_CONFIG)
 from osbs.api import OSBS
@@ -415,25 +415,8 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
         self.build_kwargs['arrangement_version'] =\
             get_arrangement_version(self.workflow, self.build_kwargs['arrangement_version'])
 
-    def make_list(self, value):
-        if not isinstance(value, list):
-            value = [value]
-        return value
-
     def get_platforms(self):
-        build_file_dir = self.workflow.source.get_build_file_path()[1]
-        excluded_platforms = set()
-        container_path = os.path.join(build_file_dir, REPO_CONTAINER_CONFIG)
-        if os.path.exists(container_path):
-            with open(container_path) as f:
-                data = yaml.safe_load(f)
-                if data is None or 'platforms' not in data or data['platforms'] is None:
-                    return self.platforms
-                excluded_platforms = set(self.make_list(data['platforms'].get('not', [])))
-                only_platforms = set(self.make_list(data['platforms'].get('only', [])))
-                if only_platforms:
-                    self.platforms = self.platforms & only_platforms
-        return self.platforms - excluded_platforms
+        return get_platforms_in_limits(self.workflow, self.platforms)
 
     def get_current_builds(self, osbs):
         field_selector = ','.join(['status!={status}'.format(status=status.capitalize())
@@ -857,7 +840,8 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
                                      platforms, current_buildimage)
 
     def run(self):
-        platforms = self.get_platforms()
+        self.platforms = self.get_platforms()
+        platforms = self.platforms
 
         if not platforms:
             raise RuntimeError("No enabled platform to build on")
--- a/atomic_reactor/plugins/pre_check_and_set_platforms.py
+++ b/atomic_reactor/plugins/pre_check_and_set_platforms.py
@@ -1,0 +1,55 @@
+"""
+Copyright (c) 2018 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+
+
+Query the koji build target, if any, to find the enabled architectures. Remove any excluded
+architectures, and return the resulting list.
+
+build_orchestrate_build will prefer this list of architectures over the platforms supplied by
+USER_PARAMS, which is necessary to allow autobuilds to build on the correct architectures
+when koji build tags change.
+"""
+
+from atomic_reactor.plugin import PreBuildPlugin
+from atomic_reactor.util import get_platforms_in_limits
+from atomic_reactor.plugins.pre_reactor_config import get_koji_session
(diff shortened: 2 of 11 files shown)


tests/plugins/test_check_and_set_platforms.py
```diff
@@ -0,0 +1,151 @@
...
+
+    setattr(workflow.builder, 'image_id', "asd123")
+    setattr(workflow.builder, 'base_image', ImageName(repo='Fedora', tag='21'))
+    setattr(workflow.builder, 'source', source)
+    setattr(workflow, 'source', source)
+
+    return tasker, workflow
+
+
+@pytest.mark.parametrize(('platforms', 'platform_exclude', 'platform_only', 'result'), [
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

atomic_reactor/plugins/pre_check_and_set_platforms.py
```diff
@@ -0,0 +1,79 @@
...
+    def __init__(self, tasker, workflow, target, hub=None, root=None, proxy=None,
+                 koji_ssl_certs_dir=None):
+
+        """
+        constructor
+
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param target: string, koji target to use as a source
+        :param hub: string, koji hub (xmlrpc)
```
lcarva: I'm wondering if we should not bother with providing these environment plugin parameters. This plugin is meant to run starting on arrangement 6 which requires reactor config map.

atomic_reactor/plugins/pre_check_and_set_platforms.py
```diff
@@ -0,0 +1,79 @@
...
+
+        self.koji_session = get_koji_session(self.workflow, self.koji_fallback)
+
+    def run(self):
+        """
+        run the plugin
+        """
+
+        self.log.info("Checking koji target for platforms")
+        event_id = self.koji_session.getLastEvent()['id']
```
lcarva: Do we care about this?

atomic_reactor/plugins/pre_check_and_set_platforms.py
```diff
@@ -0,0 +1,79 @@
...
+        target_info = self.koji_session.getBuildTarget(self.target, event=event_id)
+        build_tag = target_info['build_tag']
+        koji_build_conf = self.koji_session.getBuildConfig(build_tag, event=event_id)
+        koji_platforms = koji_build_conf['arches']
```
lcarva: Looking at koji-containerbuild's code, it looks like `'arches'` is actually a string:
```
'arches': 'i686 ppc ppc64 ppc64le s390 s390x x86_64 aarch64'
```

koji-containerbuild also uses `koji.canonArch` to modify each arch. I'm not sure what that does.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -415,6 +416,12 @@ def adjust_build_kwargs(self):
             get_arrangement_version(self.workflow, self.build_kwargs['arrangement_version'])
 
     def get_platforms(self):
+        koji_platforms = self.workflow.prebuild_results.get(PLUGIN_CHECK_AND_SET_PLATFORMS_KEY)
+        if koji_platforms:
+            return koji_platforms
```
lcarva: Do we need to set `self.platforms`? Not sure if it's actually used somewhere else, but it does seem odd to have that variable pointing to incorrect data.

atomic_reactor/plugins/pre_check_and_set_platforms.py
mlangsdorf: Not sure, but I can't find the pydoc for koji_session so I would rather leave it in place.

atomic_reactor/plugins/pre_check_and_set_platforms.py
mlangsdorf: Fixed.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -415,24 +414,11 @@ def adjust_build_kwargs(self):
...
-        if os.path.exists(container_path):
-            with open(container_path) as f:
-                data = yaml.safe_load(f)
-                if data is None or 'platforms' not in data or data['platforms'] is None:
-                    return self.platforms
-                excluded_platforms = set(self.make_list(data['platforms'].get('not', [])))
-                only_platforms = set(self.make_list(data['platforms'].get('only', [])))
-                if only_platforms:
-                    self.platforms = self.platforms & only_platforms
+        excluded_platforms, only_platforms = get_platform_limits(build_file_dir)
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

atomic_reactor/plugins/pre_check_and_set_platforms.py
mlangsdorf: It checks to make sure that it's a known arch tag. I think it's safe to just leave them as they are and let some other part of the program spit out an error.

atomic_reactor/plugins/build_orchestrate_build.py
mlangsdorf: That's how the original code had it, but I've gone ahead and changed it.

tests/test_util.py
```diff
@@ -1146,3 +1147,50 @@ def request_callback(request, all_headers=True):
...
+    if valid and platforms:
+        final_platforms = get_platforms_in_limits(str(tmpdir), platforms)
+        assert final_platforms == set(result)
+    elif valid:
+        final_platforms = get_platforms_in_limits(str(tmpdir), platforms)
+        assert final_platforms is None
+    else:
+        final_platforms = get_platforms_in_limits('bad_dir', platforms)
+        assert final_platforms == set(platforms)
+
```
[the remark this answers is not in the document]
mlangsdorf: Fixed

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -415,24 +414,11 @@ def adjust_build_kwargs(self):
         self.build_kwargs['arrangement_version'] =\
             get_arrangement_version(self.workflow, self.build_kwargs['arrangement_version'])
 
-    def make_list(self, value):
-        if not isinstance(value, list):
-            value = [value]
-        return value
-
     def get_platforms(self):
         build_file_dir = self.workflow.source.get_build_file_path()[1]
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -845,7 +847,8 @@ def set_build_image(self, platforms):
                                      platforms, current_buildimage)
 
     def run(self):
-        platforms = self.get_platforms()
+        self.platforms = self.get_platforms()
+        platforms = self.platforms
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

atomic_reactor/plugins/build_orchestrate_build.py
mlangsdorf: Fixed.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -232,6 +235,28 @@ def get_registries(workflow, fallback=NO_FALLBACK):
...
+        if reguri.version == 'v2':
+            regdict = {
+                'url': registry,
+                'insecure': registry.get('insecure', False)
+            }
+            if registry.get('auth'):
+                regdict['secret'] = registry['auth']['cfg_path']
+            return regdict
+
+    return fallback
```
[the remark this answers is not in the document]
mlangsdorf: Removed.

atomic_reactor/plugins/pre_check_and_set_platforms.py
```diff
@@ -0,0 +1,56 @@
...
+        """
+        # call parent constructor
+        super(CheckAndSetPlatformsPlugin, self).__init__(tasker, workflow)
+
+    def run(self):
+        """
+        run the plugin
+        """
+        koji_session = get_koji_session(self.workflow)
+        koji_target = get_value(self.workflow, 'koji_target', NO_FALLBACK)
```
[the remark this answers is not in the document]
mlangsdorf: I originally missed that koji_target is part of the class that UserParams inherits.

Fixed.

atomic_reactor/plugins/build_orchestrate_build.py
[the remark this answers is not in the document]
mlangsdorf: Drat, missed this comment.

Fixed.

tests/plugins/test_pull_base_image.py
```diff
@@ -327,10 +327,24 @@ def test_manifest_list_verified(self, caplog):
                                     workflow_callback=self.prepare)
         assert log_message in caplog.text()
 
+    def test_manifest_list_verified_check_failed(self, caplog):
```
lcarva: The name of this test is misleading because the verification was actually successful, but it used the "fallback" mechanism of checking `orchestrate_build` plugin parameters. Maybe the name `test_fallback_to_orchestrate_build` is more descriptive?

mlangsdorf: Fixed.

lcarva:
> I wonder if we should make pre_pull_base_image skip the platforms check when it's running in the worker build. It seems overkill to have to check it n+1 times, plus it means it has to try to figure out the platforms.

+1
I hadn't thought of that, but it does seem essential to ignore it for worker builds.
Maybe a plugin boolean parameter controls this behavior? That way we can simply explicitly enable the check via `orchestrator_inner:6.json`.

twaugh:
> Maybe a plugin boolean parameter controls this behavior?

Yes, I like that. So it would default to True, and worker_inner:6 would set it to false?

lcarva:
> So it would default to True, and worker_inner:6 would set it to false?

It might make more sense to default to false since that's a new additional behavior. Prod builds wouldn't want this check either.

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -134,6 +135,10 @@ def adjust_for_autorebuild(self):
             self.compose_ids = tuple()
 
     def get_arches(self):
+        platforms = self.workflow.prebuild_results.get(PLUGIN_CHECK_AND_SET_PLATFORMS_KEY)
+        if platforms:
+            return [platform for platform in platforms]
+
         # This will be replaced by a plugin soon but this will do for now
```
lcarva: Let's update this comment.

lcarva:
@mlangsdorf, we also need to use `add_filesystem` to use this new plugin to populate `architectures` parameter. Keep in mind that `add_filesystem` runs in both orchestrator and worker builds, but we only want to set `architectures` for orchestrator builds. It might be best to have a follow up PR for that since this one is very close to being merged.

atomic_reactor/util.py
```diff
@@ -1279,3 +1280,30 @@ def get_log_files(self, osbs, build_id):
...
+    container_path = os.path.join(build_file_dir, REPO_CONTAINER_CONFIG)
+    if os.path.exists(container_path):
+        with open(container_path) as f:
+            data = yaml.safe_load(f)
+            if data and 'platforms' in data and data['platforms']:
+                excluded_platforms = set(make_list(data['platforms'].get('not', [])))
+                only_platforms = set(make_list(data['platforms'].get('only', [])))
+                if only_platforms:
+                    expected_platforms = expected_platforms & only_platforms
+    return expected_platforms - excluded_platforms
```
rcerven: maybe we could write warning, in case when in container.yaml is same platform in 'only' and 'not' ?
I am aware that it wasn't changed in this PR.

tests/plugins/test_check_and_set_platforms.py
```diff
@@ -0,0 +1,130 @@
...
+    (None, '', 'ppc64le', None),
```
rcerven: can we add test, where same platform is in 'only' and 'exclude' at the same time?

mlangsdorf: Fixed.

atomic_reactor/util.py

lcarva:
Updated release notes draft:
> - New plugin `pre_check_and_set_platforms` added and used for dynamically determining list of platforms to build

PR #954, opened 2018-03-29
Fix logic for adding logs to mailing

bfontecc007:
The logs were not added to the email in some cases where they should, and were added in some cases where they should not.


mlangsdorf:
added to release notes
"- logs will be included for all builds that aren't rebuilds"
under # Bug Fixes

PR #955, opened 2018-04-02
Verify base image is manifest list for arches

lcarva:

### DIFF
--- a/atomic_reactor/plugins/pre_pull_base_image.py
+++ b/atomic_reactor/plugins/pre_pull_base_image.py
@@ -14,9 +14,10 @@ from __future__ import unicode_literals
 import docker
 
 from atomic_reactor.plugin import PreBuildPlugin
-from atomic_reactor.util import get_build_json, ImageName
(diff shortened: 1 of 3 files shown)


atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -138,3 +141,55 @@ def run(self):
         self.workflow.builder.set_base_image(base_image.to_str())
         self.workflow.builder.original_base_image = ImageName.parse(original_base_image.to_str())
         self.log.debug("image '%s' is available", pulled_base)
+
+    def validate_base_image_platforms(self, base_image):
```
mlangsdorf: please change this to validate_platforms_in_image_manifest_list
the logic is fine, but I read it expecting to validate that the expected platforms had good values and was surprised when you jumped out early if there was only one expected platform.

it's a complicated enough concept that a line or two of commentary about what is being validated here would be helpful.

mlangsdorf:
added to release notes:
"- builds using base_image will fail with a meaningful error message if manifest lists are not available for all required platforms"
under # Improvements

PR #956, opened 2018-04-03
plugins/exit_pulp_publish: tolerate missing annotations

sosiouxme:
Fix for OSBS-5262 in which when a worker build disappears or monitoring
otherwise fails for some reason, pulp_publish gives a traceback when
attempting to find annotations from it.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -94,3 +94,6 @@ if [[ $OS != "fedora" ]]; then $RUN $PIP install -U setuptools; fi
 
 # Run tests
 $RUN pytest -vv tests --cov atomic_reactor "$@"
+
+echo "To run tests again:"
+echo "$RUN pytest -vv tests --cov atomic_reactor "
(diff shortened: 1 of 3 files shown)


test.sh
```diff
@@ -94,3 +94,6 @@ if [[ $OS != "fedora" ]]; then $RUN $PIP install -U setuptools; fi
 
 # Run tests
 $RUN pytest -vv tests --cov atomic_reactor "$@"
+
+echo "To run tests again:"
+echo "$RUN pytest -vv tests --cov atomic_reactor "
```
mlangsdorf: Seems unrelated to the pulp_publish changes.

lcarva:
Updated release notes draft:
> - `exit_pulp_publish` tolerates missing annotations

test.sh
sosiouxme: It is unrelated... just a convenience for the future

PR #957, opened 2018-04-05
Fix arches for ODCS compose request

lcarva:


twaugh:
Typo in commit message ("muli-arch")

lcarva:
> Typo in commit message ("muli-arch")

Fixed.

PR #958, opened 2018-04-05
exit_koji_promote: 'owner' and 'submitter' in metadata

sosiouxme:
Now provides an 'owner' key in metadata['build'] (the username of the
person that built the image), as well as a 'submitter' key in
metadata['extra'] (the technical submitter; osbs in this case). Brew
will accept these keys and display both who triggered the build and what
service built the container.

Ref. OSBS-4608


sosiouxme:
@twaugh @lcarva @mlangsdorf can someone add me to the necessary github group to add reviewers?

Also the unit tests pass but I'm trying to figure out the best way to actually exercise this plugin... pointers welcome. Presumably configuring osbs-client to invoke the build with a different set of plugins...

lcarva:
> @twaugh @lcarva @rcerven can someone add me to the necessary github group to add reviewers?

Tim is following up on that.

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -506,14 +506,21 @@ def get_build(self, metadata):
...
+        koji_task_owner = None
         koji_task_id = metadata.get('labels', {}).get('koji-task-id')
         if koji_task_id is not None:
             self.log.info("build configuration created by Koji Task ID %s",
                           koji_task_id)
             try:
-                extra['container_koji_task_id'] = int(koji_task_id)
+                extra['container_koji_task_id'] = koji_task_id = int(koji_task_id)
+                koji_task_owner = get_koji_task_owner(
+                    session, koji_task_id, default=dict(name=None)
```
lcarva: Oh interesting how `default` is very misleading in this function. Can we change `get_koji_task_owner` to return an empty dict if owner info can't be found?

```python
def get_koji_task_owner(session, task_id, default=None):
  default = {} if default is None else default
  ...
```

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -677,10 +685,10 @@ def run(self):
             self.log.info("Not promoting failed build to koji")
             return
 
-        koji_metadata, output_files = self.get_metadata()
+        session = get_koji_session(self.workflow, self.koji_fallback)
```
lcarva: It might be worth to assign `self.koji_session` so it doesn't have to be passed around.

tests/plugins/test_koji_promote.py
```diff
@@ -1570,6 +1572,33 @@ def test_koji_promote_no_parent_id(self, tmpdir, os_env, reactor_config_map):  #
         assert 'parent_build_id' not in image.keys()
         assert PLUGIN_KOJI_PARENT_KEY not in workflow.prebuild_results
 
+    def test_koji_promote_owner_submitter(self, tmpdir, monkeypatch):
+        session = MockedClientSession('')
+        session.getTaskInfo = lambda x: {'owner': 1234}
```
lcarva: You probably want to validate that this function was called, how about:
```python
flexmock(session).should_receive('getTaskInfo').with_args(1234).and_return({'owner': 456789})
```

*You probably want to use a different value for task ID and user ID.

tests/plugins/test_koji_promote.py
```diff
@@ -1570,6 +1572,33 @@ def test_koji_promote_no_parent_id(self, tmpdir, os_env, reactor_config_map):  #
...
+        session.getUser = lambda x: {'name': 'dev1'}
```
lcarva: Same here, let's validate that this method was called with the args we expect.

sosiouxme:
Finally getting around to updating this... I think I addressed all concerns.

tests/plugins/test_koji_promote.py
```diff
@@ -1570,6 +1573,33 @@ def test_koji_promote_no_parent_id(self, tmpdir, os_env, reactor_config_map):  #
         assert 'parent_build_id' not in image.keys()
         assert PLUGIN_KOJI_PARENT_KEY not in workflow.prebuild_results
 
+    def test_koji_promote_owner_submitter(self, tmpdir, monkeypatch):
+        session = flexmock(MockedClientSession(''))
+        session.should_receive('getTaskInfo').once().with_args(1234).and_return({'owner': 2345})
```
rcerven: use variables for 'taskinfo', 'userid', 'username'

atomic_reactor/plugins/exit_koji_promote.py
```diff
@@ -677,10 +683,10 @@ def run(self):
             self.log.info("Not promoting failed build to koji")
             return
 
+        session = self.koji_session = get_koji_session(self.workflow, self.koji_fallback)
```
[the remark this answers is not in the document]
sosiouxme: yes but `session` is shorter and it's in there several times... *shrug*

sosiouxme:
Minor updates per @rcerven review

atomic_reactor/koji_util.py
```diff
@@ -166,12 +166,14 @@ def tag_koji_build(session, build_id, target, poll_interval=5):
...
-def get_koji_task_owner(session, task_id, default=""):
+def get_koji_task_owner(session, task_id, default=None):
+    default = {} if default is None else default
     if task_id:
         try:
             koji_task_info = session.getTaskInfo(task_id)
             koji_task_owner = session.getUser(koji_task_info['owner'])
-        except:
+        except Exception as e:
+            logger.warn('Exception getting task owner: {} {}'.format(type(e), e))
```
lcarva: Thanks for adding this log message. Consider using `logger.exception` instead:
```python
logger.exception('Unable to get Koji task owner')
```

^ That will actually log the whole traceback as context and provide the useful error message to users.

sosiouxme: Sure. I didn't have the context as to what kinds of things might cause this to fail and how much anything would care, I just knew it was confusing when I had a bug in my unit test

sosiouxme:
Thanks also @lcarva I went ahead and did that.

PR #959, opened 2018-04-06
Make env plugin parameters in koji_upload optional

lcarva:


mlangsdorf:
in the release notes, added
- in koji_upload, the kojihub, url, build_json_dir parameters are optional and will be taken from REACTOR_CONFIG
under # Build Json API changes

PR #961, opened 2018-04-06
Add helper for creating koji.PathInfo object

lcarva:
Also ensures that pre_koji plugin reads koji root url from reactor
config when available.


lcarva:
Tests are failing because `koji` module is not found during tests in `test_reactor_config`. I'm a little confused as to why tests for `get_koji_session` don't hit this problem. I'll look into this Monday.

lcarva:
Updated release notes draft:
> - `pre_koji` plugin reads koji root url from reactor config
> - Add helper for creating koji.PathInfo object

PR #962, opened 2018-04-06
Read docker registry from reactor config in import_image

lcarva:


lcarva:
Updated release notes draft:
> - read docker registry from reactor config in import_image

PR #965, opened 2018-04-09
Change image name to repositories in exit_sendmail plugin

bfontecc007:
Change the email sent by atomic reactor to include a proper image name and a list of well formatted repositories.

### DIFF
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -12,6 +12,7 @@ from email.mime.base import MIMEBase
 from email import encoders
 import smtplib
 import socket
+import json
 try:
     from urlparse import urljoin
 except ImportError:
@@ -21,11 +22,12 @@ from atomic_reactor.plugin import ExitPlugin, PluginFailedException
 from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
 from atomic_reactor.plugins.exit_koji_import import KojiImportPlugin
 from atomic_reactor.plugins.exit_koji_promote import KojiPromotePlugin
+from atomic_reactor.plugins.exit_store_metadata_in_osv3 import StoreMetadataInOSv3Plugin
 from atomic_reactor.plugins.pre_reactor_config import (get_smtp_session, get_koji_session,
                                                        get_smtp, get_koji, get_openshift,
                                                        get_openshift_session, get_koji_path_info)
 from atomic_reactor.koji_util import get_koji_task_owner
-from atomic_reactor.util import get_build_json, OSBSLogs
+from atomic_reactor.util import get_build_json, OSBSLogs, ImageName
 
 
 class SendMailPlugin(ExitPlugin):
@@ -196,11 +198,32 @@ class SendMailPlugin(ExitPlugin):
             should_send |= should_send_mapping[state]
         return should_send
 
+    def _get_image_name_and_repos(self):
+
+        repos = []
+        stored_data = self.workflow.exit_results.get(StoreMetadataInOSv3Plugin.key)
+        if not stored_data or 'annotations' not in stored_data:
+            raise ValueError('Stored Metadata not found')
+
+        repo_data = json.loads(stored_data['annotations']['repositories'])
+
+        repos.extend(repo_data.get('unique', []))
+        repos.extend(repo_data.get('primary', []))
+
+        if not repos:
+            raise ValueError('Repositories not found in stored Metadata')
+
+        image_name_obj = ImageName.parse(repos[0])
+        image_name = image_name_obj.get_repo()
+
+        return (image_name, repos)
+
     def _render_mail(self, rebuild, success, auto_canceled, manual_canceled):
         """Render and return subject and body of the mail to send."""
-        subject_template = '%(endstate)s building image %(image)s'
+        subject_template = '%(endstate)s building image %(image_name)s'
         body_template = '\n'.join([
-            'Image: %(image)s',
+            'Image Name: %(image_name)s',
+            'Repositories: %(repositories)s',
             'Status: %(endstate)s',
             'Submitted by: %(user)s',
         ])
@@ -219,8 +242,15 @@ class SendMailPlugin(ExitPlugin):
 
         url = self._get_logs_url()
 
+        image_name, repos = self._get_image_name_and_repos()
+
+        repositories = ''
+        for repo in repos:
+            repositories += '\n    ' + repo
+
         formatting_dict = {
-            'image': self.workflow.image,
+            'repositories': repositories,
+            'image_name': image_name,
             'endstate': endstate,
             'user': '<autorebuild>' if rebuild else self.submitter,
             'logs': url
--- a/tests/plugins/test_sendmail.py
+++ b/tests/plugins/test_sendmail.py
@@ -1,3 +1,4 @@
+from __future__ import unicode_literals
 import os
 import smtplib
 from collections import namedtuple
@@ -27,6 +28,7 @@ except ImportError:
 from atomic_reactor.plugin import PluginFailedException
 from atomic_reactor.plugins.pre_check_and_set_rebuild import CheckAndSetRebuildPlugin
 from atomic_reactor.plugins.exit_sendmail import SendMailPlugin
+from atomic_reactor.plugins.exit_store_metadata_in_osv3 import StoreMetadataInOSv3Plugin
 from atomic_reactor.plugins.exit_koji_import import KojiImportPlugin
 from atomic_reactor.plugins.exit_koji_promote import KojiPromotePlugin
 from atomic_reactor.plugins.pre_reactor_config import (ReactorConfigPlugin,
@@ -122,6 +124,21 @@ class MockedPathInfo(object):
         return "tasks/%s" % task_id
 
 
+DEFAULT_ANNOTATIONS = {
+    'repositories': {
+        'unique': ['foo/bar:baz'],
+        'primary': ['foo/bar:spam'],
+    }
+}
+
+
+def mock_store_metadata_results(workflow, annotations=DEFAULT_ANNOTATIONS):
+    result = {}
+    if annotations:
+        result['annotations'] = {key: json.dumps(value) for key, value in annotations.items()}
+    workflow.exit_results[StoreMetadataInOSv3Plugin.key] = result
+
+
 class TestSendMailPlugin(object):
     def test_fails_with_unknown_states(self, reactor_config_map):  # noqa
         class WF(object):
@@ -197,6 +214,13 @@ class TestSendMailPlugin(object):
         assert p._should_send(rebuild, success, auto_canceled, manual_canceled) == expected
 
     @pytest.mark.parametrize('success', (True, False))
+    @pytest.mark.parametrize(('has_store_metadata_results', 'annotations', 'has_repositories',
+                              'expect_error'), [
+        (True, True, True, False),
+        (True, True, False, True),
+        (True, False, False, True),
+        (False, False, False, True)
+    ])
     @pytest.mark.parametrize('koji_integration', (True, False))
     @pytest.mark.parametrize(('autorebuild', 'auto_cancel', 'manual_cancel',
                               'to_koji_submitter', 'has_koji_logs'), [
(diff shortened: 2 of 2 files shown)


atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -224,8 +227,31 @@ def _render_mail(self, rebuild, success, auto_canceled, manual_canceled):
...
+
+        repo_data = json.loads(stored_data['annotations']['repositories'])
+
+        repos.extend(repo_data['unique'])
+        repos.extend(repo_data['primary'])
+
+        if not repos:
+            pass
+
+        image_name_obj = ImageName().parse(repos[0])
```
lcarva: `parse` is a class method, no need to instantiate the class to call it. This should do: `ImageName.parse(repos[0])`

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -224,8 +227,31 @@ def _render_mail(self, rebuild, success, auto_canceled, manual_canceled):
 
         url = self._get_logs_url()
 
+        repos = []
+        stored_data = self.workflow.exit_results.get(StoreMetadataInOSv3Plugin.key)
+        if not stored_data:
+            pass
```
lcarva: What's the purpose of these `pass` statements?

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -224,8 +227,31 @@ def _render_mail(self, rebuild, success, auto_canceled, manual_canceled):
...
+        repo_data = json.loads(stored_data['annotations']['repositories'])
+
+        repos.extend(repo_data['unique'])
+        repos.extend(repo_data['primary'])
+
+        if not repos:
+            pass
+
+        image_name_obj = ImageName().parse(repos[0])
+        image_name = image_name_obj.namespace + '/' + image_name_obj.repo
```
lcarva: Not all images have a namespace. The `get_repo` method will handle that for you.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -224,8 +227,31 @@ def _render_mail(self, rebuild, success, auto_canceled, manual_canceled):
 
         url = self._get_logs_url()
 
+        repos = []
```
lcarva: Let's turn this into a helper method:
```python
image_name, repos = self.get_image_name_and_repos()
```

atomic_reactor/plugins/exit_sendmail.py
twaugh: Agreed.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -201,11 +203,34 @@ def _should_send(self, rebuild, success, auto_canceled, manual_canceled):
             should_send |= should_send_mapping[state]
         return should_send
 
+    def _get_image_name_and_repos(self):
+
+        repos = []
+        stored_data = self.workflow.exit_results.get(StoreMetadataInOSv3Plugin.key)
+        if not stored_data:
+            raise ValueError
```
lcarva: Let's be descriptive in what the error is.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -201,11 +203,34 @@ def _should_send(self, rebuild, success, auto_canceled, manual_canceled):
             should_send |= should_send_mapping[state]
         return should_send
 
+    def _get_image_name_and_repos(self):
+
+        repos = []
+        stored_data = self.workflow.exit_results.get(StoreMetadataInOSv3Plugin.key)
+        if not stored_data:
```
lcarva: These two if-statements can be combined: `if not stored_data or 'annotations' not in stored_data`

lcarva: Alternatively, you could just use the value and catch exceptions:
```python
try:
  repo_data = json.loads(stored_data['annotations']['repositories'])
except (KeyError, TypeErrror):
  self.log.exception('Cannot determine repositories, invalid result for plugin %s', StoreMetadataInOSv3Plugin.key)
  raise
```

I'm indifferent to re-raising the same exception vs raising a new `ValueError` one.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -201,11 +203,34 @@ def _should_send(self, rebuild, success, auto_canceled, manual_canceled):
...
+        elif 'annotations' not in stored_data:
+            raise ValueError
+
+        repo_data = json.loads(stored_data['annotations']['repositories'])
+
+        repos.extend(repo_data['unique'])
+        repos.extend(repo_data['primary'])
+
+        if not repos:
+            raise ValueError
```
lcarva: Let's provide a meaningful exception reason here.

atomic_reactor/plugins/exit_sendmail.py
lcarva: Not addressed.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -201,11 +203,34 @@ def _should_send(self, rebuild, success, auto_canceled, manual_canceled):
...
         return should_send
 
+    def _get_image_name_and_repos(self):
+
+        repos = []
+        stored_data = self.workflow.exit_results.get(StoreMetadataInOSv3Plugin.key)
+        if not stored_data:
+            raise ValueError('Stored Metadata not found')
+        elif 'annotations' not in stored_data:
+            raise ValueError('Stored Metadat malformed')
```
lcarva: Typo.

tests/plugins/test_sendmail.py
```diff
@@ -800,6 +800,8 @@ class WF(object):
          .with_args(True, False, False, False).and_return(True))
         flexmock(p).should_receive('_get_receivers_list').and_return(receivers)
         flexmock(p).should_receive('_fetch_log_files').and_return(None)
+        flexmock(p).should_receive('_get_image_name_and_repos').and_return(('foobar', ['baz']))
```
lcarva: This actually replaces the method from the plugin. Do we have any explicit unit test for it?

lcarva:
@bfontecc007, pushed commit [8be52](https://github.com/projectatomic/atomic-reactor/pull/965/commits/8be5215b11b2cebf12fe4cb2f69a55a07b35caba) to this branch as an attempt to fix existing test cases.

atomic_reactor/plugins/exit_sendmail.py
bfontecc007: I ended up using the 'combined ifs' approach. If you'd like me to modify the exception in some way please let me know.

tests/plugins/test_sendmail.py
```diff
@@ -310,7 +317,10 @@ class WF(object):
         }
 
         workflow = WF()
-        mock_store_metadata_results(workflow)
+        if annotations and metadata_stored:
+            mock_store_metadata_results(workflow)
+        if not annotations and metadata_stored:
```
lcarva: I find the name `metadata_stored` is misleading here. It really means there are plugin results for store metadata plugin, right? How about 'has_store_metadata_results` ?

tests/plugins/test_sendmail.py
```diff
@@ -345,6 +355,12 @@ class WF(object):
             # recipients available
             assert str(ex) == 'No recipients found'
 
+        if not metadata_stored or not annotations:
+            with pytest.raises(ValueError):
+                subject, body, logs = p._render_mail(autorebuild, success,
```
lcarva: No need to store return values from a function that is expected to raise an exception.

tests/plugins/test_sendmail.py
```diff
@@ -213,6 +217,8 @@ class WF(object):
         assert p._should_send(rebuild, success, auto_canceled, manual_canceled) == expected
 
     @pytest.mark.parametrize('success', (True, False))
+    @pytest.mark.parametrize('annotations', (True, False))
```
lcarva: I think it would make more sense to combine `annotations` and `metadata_stored` into a single entry with an additional `expect_error` param. That way it's more explicit that certain combinations of `annotations` and `metadata_stored` values are expected to raise an error. This is preferable to adding this conditional in the test body.

tests/plugins/test_sendmail.py
```diff
@@ -294,6 +320,11 @@ class WF(object):
         }
 
         workflow = WF()
+        if annotations and metadata_stored:
+            mock_store_metadata_results(workflow)
+        if not annotations and metadata_stored:
+            mock_store_missing_metadata_results(workflow)
```
twaugh: Does `annotations` make any sense if `metadata_stored` evaluates as false? If not, I'd rather see the test the other way around, e.g.:
```
if metadata_stored:
    if annotations:
        mock_store_metadata_results(workflow)
    else:
        mock_store_missing_metadata_results(workflow)
```

tests/plugins/test_sendmail.py
```diff
@@ -122,6 +124,24 @@ def taskrelpath(self, task_id):
...
+
+
+def mock_store_metadata_results(workflow, annotations=DEFAULT_ANNOTATIONS):
+    result = {}
+    result['annotations'] = {key: json.dumps(value) for key, value in annotations.items()}
+    workflow.exit_results[StoreMetadataInOSv3Plugin.key] = result
+
+
+def mock_store_missing_metadata_results(workflow):
+    workflow.exit_results[StoreMetadataInOSv3Plugin.key] = {}
```
twaugh: If we had one function to generate the annotations and one function to store them in the results, it might be easier to read in the actual test:
```
if metadata_stored:
    result = {}
    if annotations:
        result.update(make_annotations(workflow))
    mock_store_metadata_result(workflow, result)
```

and the regular case could set annotations by default:
```
mock_store_metadata_result(workflow)
```

bfontecc007: @twaugh what do you think of commit afb383e?

twaugh: Yes, I like it. :)

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -196,11 +198,32 @@ def _should_send(self, rebuild, success, auto_canceled, manual_canceled):
...
+        if not stored_data or 'annotations' not in stored_data:
+            raise ValueError('Stored Metadata not found')
+
+        repo_data = json.loads(stored_data['annotations']['repositories'])
+
+        repos.extend(repo_data['unique'])
+        repos.extend(repo_data['primary'])
+
+        if not repos:
+            raise ValueError('Repositories not found in stored Metadata')
```
twaugh: I don't think this line gets covered by unit testing.

bfontecc007: @twaugh I just pushed a test case for it. It's rebased in, so you'll see that the parametrize params and the test logic are a little different now.

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -196,11 +198,32 @@ def _should_send(self, rebuild, success, auto_canceled, manual_canceled):
...
+
+        repos = []
+        stored_data = self.workflow.exit_results.get(StoreMetadataInOSv3Plugin.key)
+        if not stored_data or 'annotations' not in stored_data:
+            raise ValueError('Stored Metadata not found')
+
+        repo_data = json.loads(stored_data['annotations']['repositories'])
+
+        repos.extend(repo_data['unique'])
+        repos.extend(repo_data['primary'])
```
rcerven: why are we adding primary, if we not ever use it?

bfontecc007: what makes you say we don't use it?

rcerven: nvm I see it is used outside of method

bfontecc007: ok np

tests/plugins/test_sendmail.py
```diff
@@ -858,6 +905,7 @@ def quit(self):
             .with_args(True, False, False, False).and_return(True))
         flexmock(p).should_receive('_get_receivers_list').and_return(receivers)
         flexmock(OSBS).should_receive('get_orchestrator_build_logs').and_return(fake_logs)
+        flexmock(p).should_receive('_get_image_name_and_repos').and_return(('foobar', ['baz']))
```
rcerven: shouldn't this be returning something like : 
('foobar', ['foo/bar:baz', 'foo/bar:spam']) ??

bfontecc007: yeah that would be more accurate

bfontecc007: @rcerven I just pushed an update

lcarva:
@twaugh, waiting for your approval before merging since you had comments. Although, I think they have all been addressed.

twaugh:
Updated release notes draft:
> - The email sent by atomic reactor now includes a proper image name and a list of well-formatted repositories

PR #966, opened 2018-04-10
revert pulp_pull reads from reactor config

mlangsdorf:
We don't need this and the existing code is fine.


atomic_reactor/plugins/post_pulp_pull.py
```diff
@@ -62,17 +60,9 @@ def __init__(self, tasker, workflow,
...
-        pulp_config = get_pulp(workflow, False)
-        if pulp_config:
-            pulp_handler = get_pulp_session(workflow, self.log, NO_FALLBACK)
-            self.secret = pulp_handler.get_secret()
-            self.insecure = pulp_handler.get_insecure()
-        else:
-            self.insecure = insecure
-            self.secret = secret
-
-        self.expect_v2schema2 = not get_prefer_schema1_digest(workflow, not expect_v2schema2)
```
lcarva: We do still want to read `expect_v2schema2` from reactor config.

PR #967, opened 2018-04-10
koji_tag_build: make kojihub explicitly optional

mlangsdorf:
make kojihub explicitly optional and add a test case to handle
passing no arguments to the system.


lcarva:

PR #970, opened 2018-04-12
Resolve module compose test failure 2

owtaylor:
This PR fixes a problem where dependencies were missing during Travis testing causing Flatpak tests not being run because it looked like modulemd or pdc-client was missing, but actually the websockets dependency of docker-client was removed and not built subsequently.

Then two revealed problems are fixed:
 - disabling retries during mocking didn't work in all cases - see the commit message for details
 - A (harmless) backtrace was logged because of a minor logic error


lcarva:
Updated release notes draft:
> - fix error in a log() call in `resolve_module_compose` that was causing a backtrace

PR #972, opened 2018-04-12
Only remove pulp_push if v1 content is not on

lcarva:
The plugin pulp_push is meant to push v1 content to pulp repository. Use
the content_versions value from reactor config to determine whether or
not plugin should be removed.


lcarva:

PR #973, opened 2018-04-12
Make orchestrate_build's platforms param optional

lcarva:


lcarva:

PR #974, opened 2018-04-13
Use content_sets.yml instead of content_sets.yaml

lcarva:


lcarva:
Updated release notes draft:
> Use content_sets.yml instead of content_sets.yaml

PR #975, opened 2018-04-16
Allow pulp-only repos when requesting ODCS composes (OSBS-5462)

twaugh:
In the case where no packages are requested, no modules are requested, but pulp_repos is true *and* there are Pulp content sets defined, do the right thing.

Do not reject the build. Do not request package or module composes. Request the Pulp composes.


lcarva:
Updated release notes draft:
> - Allow pulp-only repos when requesting ODCS composes

PR #977, opened 2018-04-16
OSBS-3053: threaded filesystem poll

sosiouxme:
Have DBW use a background thread to track filesystem usage over time so it can be recorded in the build metadata.

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -14,6 +14,9 @@ import logging
 import tempfile
 import signal
 import docker
+import threading
+import os
+import time
 
 from atomic_reactor.build import InsideBuilder
 from atomic_reactor.plugin import (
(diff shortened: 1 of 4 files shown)


atomic_reactor/inner.py
```diff
@@ -252,6 +254,52 @@ def all_registries(self):
...
+        new_data = dict(
+            mb_free=st.f_bavail * st.f_frsize / mb,
+            mb_total=st.f_blocks * st.f_frsize / mb,
+            mb_used=(st.f_blocks - st.f_bfree) * st.f_frsize / mb,
+            inodes_free=st.f_ffree,
+            inodes_total=st.f_files,
+            inodes_used=st.f_files - st.f_ffree,
+        )
+        for key in ["mb_total", "mb_used", "inodes_total", "inodes_used"]:
+            if new_data[key] > data.get(key):
```
lcarva: `dict.get` returns `None` if key is set. In python2 `None` is evaluated as greater than for any integer. In python 3 it raises `TypeError`. Does it make sense to use `0` as the default value (`data.get(key, 0)`)? Or is checking for falsy value better?

sosiouxme: Ah, did not know that, thanks. I think `data.get(key, 0)` is better, can use it the same here and below.

sosiouxme: Do PR tests run under both python 2 and 3?

lcarva: I think if you use it below, it's going to mess up the "min" check.
I'm wondering if you need `sys.maxsize` for that one.
*Note that this is max **int** size, you can still get a larger value but it'll be of type **long**.

lcarva: Yes, they run for both 2 and 3. It looks like the tests for python 3 failed: https://travis-ci.org/projectatomic/atomic-reactor/builds/367111697?utm_source=github_status&utm_medium=notification

atomic_reactor/inner.py
```diff
@@ -252,6 +254,52 @@ def all_registries(self):
...
+class FSWatcher(threading.Thread):
+
+    def __init__(self, *args, **kwargs):
+        super(FSWatcher, self).__init__(*args, **kwargs)
+        self.daemon = True  # exits whenever the process exits
+        self._lock = threading.RLock()
+        self._data = {}
+
+    def run(self):
+        while True:
```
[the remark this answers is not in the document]
sosiouxme: Ah, I hadn't thought about dangling threads in testing (overhead would still be minimal, but it does feel untidy). The API doesn't provide an obvious way to outright kill a thread, but I guess it would be easy enough to have it exit.

sosiouxme:
Updated and hopefully complete -- please review.

atomic_reactor/inner.py
```diff
@@ -252,6 +255,62 @@ def all_registries(self):
...
+    def __init__(self, *args, **kwargs):
+        super(FSWatcher, self).__init__(*args, **kwargs)
+        self.daemon = True  # exits whenever the process exits
+        self._lock = threading.Lock()
+        self._done = False
+        self._data = {}
+
+    def run(self):
+        """ Overrides parent method to implement thread's functionality. """
+        while True:
```
lcarva: `while not self._done` ?

sosiouxme: I guess that's prettier :+1:

atomic_reactor/inner.py
```diff
@@ -452,6 +513,7 @@ def build_docker_image(self):
             finally:
                 self.source.remove_tmpdir()
 
+            self.fs_watcher.finish()
```
lcarva: Should this go under the `finally` block above? If an exit plugin fails, this line is not reached.

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -223,7 +233,8 @@ def run(self):
             "base-image-name": base_image_name,
             "image-id": self.workflow.builder.image_id or '',
             "digests": json.dumps(self.get_pullspecs(self.get_digests())),
-            "plugins-metadata": json.dumps(self.get_plugin_metadata())
+            "plugins-metadata": json.dumps(self.get_plugin_metadata()),
+            "filesystem": json.dumps(self.get_filesystem_metadata()),
```
lcarva: It would be great to add a unit test for ensuring this annotation is set properly.

requirements.txt
```diff
@@ -4,3 +4,4 @@ dockerfile-parse>=0.0.5
 jsonschema
 PyYAML
 six
+backports.ssl-match-hostname==3.4.0.2
```
lcarva: Why is this needed?

sosiouxme: It's not, I'm a doofus

atomic_reactor/inner.py
sosiouxme: That would be unlikely to cause much concern :) but sure, to be tidy.

sosiouxme:
Thanks, I think I've addressed everything again.

lcarva:
Updated release nodes draft:
> - New Build annotation `filesystem` will provide consumed storage stats for each build

PR #980, opened 2018-04-18
Require a platform for GOARCH=amd64 (OSBS-5375)

twaugh:
This is needed because otherwise the resulting tag will not be able to be pulled by clients without support for the 'manifest list' type.

In fact, the pulp_pull plugin will sit waiting forever for an amd64 image manifest to show up.

**Note**: this adds an additional parameter to the plugin, `goarch`, but I need advice on how to get that value filled in.

### DIFF
--- a/atomic_reactor/plugins/pre_check_and_set_platforms.py
+++ b/atomic_reactor/plugins/pre_check_and_set_platforms.py
@@ -16,8 +16,14 @@ when koji build tags change.
 
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.util import get_platforms_in_limits
-from atomic_reactor.plugins.pre_reactor_config import get_koji_session, NO_FALLBACK
+from atomic_reactor.plugins.pre_reactor_config import (get_koji_session,
+                                                       get_platform_to_goarch_mapping,
+                                                       NO_FALLBACK)
 from atomic_reactor.constants import PLUGIN_CHECK_AND_SET_PLATFORMS_KEY
+
+
+class MustBuildForAmd64(Exception):
+    """ Platforms must include one for GOARCH amd64 """
 
 
 class CheckAndSetPlatformsPlugin(PreBuildPlugin):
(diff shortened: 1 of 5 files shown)


atomic_reactor/plugins/pre_check_and_set_platforms.py
```diff
@@ -20,21 +20,41 @@
...
 
         :param tasker: DockerTasker instance
         :param workflow: DockerBuildWorkflow instance
+        :param koji_target: str, Koji build target name
+        :param goarch: Mapping, platform name to GOARCH platform name
         """
         # call parent constructor
         super(CheckAndSetPlatformsPlugin, self).__init__(tasker, workflow)
         self.koji_target = koji_target
+        self.goarch = goarch
```
lcarva: This is available in reactor config.
Since this plugin is new to arrangement 6 which requires reactor config map, I don't think it's worth adding a fallback option to plugin parameter.

twaugh: Done.

atomic_reactor/plugins/pre_check_and_set_platforms.py
```diff
@@ -31,10 +37,32 @@ def __init__(self, tasker, workflow, koji_target):
...
+            self.goarch = None
+
+    def validate_platforms(self, platforms):
+        """
+        Verify that a platform with GOARCH=amd64 is present. If it is not,
+        the tag will not be pullable by clients that do not have
+        support for the 'manifest list' type.
+        """
+        if self.goarch is None:
+            # Don't perform this check without a real reactor_config_map
```
lcarva: Maybe log that validation is being skipped?

atomic_reactor/plugins/pre_check_and_set_platforms.py
```diff
@@ -31,10 +37,32 @@ def __init__(self, tasker, workflow, koji_target):
...
+            return
+
+        for platform in platforms:
+            goarch = self.goarch.get(platform, platform)
+            if goarch == 'amd64':
+                return
+        else:
```
lcarva: Unnecessary `else`.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -306,6 +307,17 @@ def get_clusters_client_config_path(workflow, fallback=NO_FALLBACK):
     return os.path.join(client_config_dir, 'osbs.conf')
 
 
+def get_platform_to_goarch_mapping(workflow,
```
lcarva: It would be great to use this in `pull_base_image` too, if possible.

twaugh: Added.

atomic_reactor/plugins/pre_check_and_set_platforms.py
twaugh: Added.

twaugh:
Updated release notes:
- check_and_set_platform: check that there is a platform for GOARCH=amd64 and fail the build if there is not

PR #981, opened 2018-04-23
Use helper method to get image_label_info_url_format from ConfigMap (OSBS-5520)

twaugh:


twaugh:
Updated release notes draft:
- Add `image_label_info_url_format` to reactor config

PR #983, opened 2018-04-24
override content versions for each platform

rcerven:


atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -515,6 +516,19 @@ def get_worker_build_kwargs(self, release, platform, koji_upload_dir,
...
+            for descriptor in self.platform_descriptors:
+                if descriptor['platform'] == platform:
+                    enable_v1 = descriptor.get('enable_v1', False)
+                    break
+
+            if not enable_v1 and 'content_versions' in worker_reactor_conf:
+                if 'v2' in worker_reactor_conf['content_versions']:
+                    worker_reactor_conf['content_versions'] = ['v2']
+                else:
+                    raise RuntimeError("v2 only platform in v1 only instance")
```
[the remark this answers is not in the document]
rcerven: yop can do, error message was taken from that conf.py where this logic was for arr5

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -498,6 +499,22 @@ def get_koji_upload_dir():
...
+                break
+
+        # Remove v1 from content_versions if the platform descriptor doesn't enable v1
+        if not enable_v1 and 'content_versions' in worker_reactor_conf:
+            try:
+                worker_reactor_conf['content_versions'].remove('v1')
+            except ValueError:
+                pass
+            if not worker_reactor_conf['content_versions']:
+                raise RuntimeError("v1 not enabled in platform descriptor for v1-only platform")
```
twaugh: Need to test the new code paths here.

rcerven: yop I will work on unit tests :)

lcarva: This error message doesn't seem right to me. What's a "v1-only" platform? You probably mean that there's no v1 enabled, and this is a v1-only build request.

Would it be better to log up above that `'v1'` is being removed for this platform, then here we can simply say that `content_versions` is empty.

rcerven: ok will change that

PR #984, opened 2018-04-25
Add support for registry cert authentication.

cverna:
Open this for early review, before I start working on some tests

----

This commit allows to use the registry certificates to
authenticate. It makes use of the registry secret to either
store a .dockercfg file with username/password or
a registry.crt and registry.key files that will be used
by the requests session.

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -618,15 +618,18 @@ class Dockercfg(object):
 
         :param secret_path: str, dirname of .dockercfg location
         """
-
         self.json_secret_path = os.path.join(secret_path, '.dockercfg')
-        try:
-            with open(self.json_secret_path) as fp:
-                self.json_secret = json.load(fp)
-        except Exception:
-            msg = "failed to read registry secret"
-            logger.error(msg, exc_info=True)
-            raise RuntimeError(msg)
+        if not os.path.exists(self.json_secret_path):
+            self.json_secret_path = None
+            self.json_secret = {}
+        else:
+            try:
+                with open(self.json_secret_path) as fp:
+                    self.json_secret = json.load(fp)
+            except Exception:
+                msg = "failed to read registry dockercfg secret"
+                logger.error(msg, exc_info=True)
+                raise RuntimeError(msg)
 
     def get_credentials(self, docker_registry):
         # For maximal robustness we check the host:port of the passed in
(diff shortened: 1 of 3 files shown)


lcarva:
Looks reasonable.

atomic_reactor/util.py
```diff
@@ -652,6 +650,16 @@ def get_credentials(self, docker_registry):
...
             return {}
 
+    def _log_and_raise_error(self, msg):
+        """ Helper method that raise and log a
+        RuntimError exception
+
+        :param msg: message displayed in the log
+                    and in the exception
+        """
+        logger.error(msg, exc_info=True)
```
owtaylor: The place where you factored this out of was specifically handling catching an exception and rethrowing it in simplified form (while logging the original exception) - it doesn't make sense to me as a general way of raising an error. You can just raise RuntimeError("dockercfg secret file does not exist")

cverna: Good point

atomic_reactor/util.py
```diff
@@ -652,13 +660,25 @@ def __init__(self, registry, insecure=False, dockercfg_path=None):
...
+
+                username = dockercfg.get('username')
+                password = dockercfg.get('password')
+                if username and password:
+                    self.auth = requests.auth.HTTPBasicAuth(username, password)
+            else:
+                # Use the registry certificates to authenticate instead of
+                # instead of username/password.
+                self.cert_path = os.path.join(dockercfg_path, 'registry.crt')
+                self.key_path = os.path.join(dockercfg_path, 'registry.key')
```
twaugh: Oh! I had thought there was some way a .docker/config.json (= 'dockercfg' secret type) could hold a certificate and key, but it seems not. In that  case, this seems to be an ad-hoc method for storing a certificate and key. :(

Ideally we'd use the same mechanism that e.g. Kubernetes and/or Docker uses for this.

For Docker, it seems that /etc/docker/certs.d/ is used, and the filenames there (per registry) are client.cert, client.key, and ca.crt (yes, .crt in one filename and .cert in another).

At the very least I'd hope we'd use the same filenames Docker does.

Does Kubernetes do the same thing? I can't seem to find any documentation about it. In fact, there's very little documentation anyway I can find about securing docker registries with certificates instead of basic auth.

cverna: > At the very least I'd hope we'd use the same filenames Docker does.

Ok, I will update the code to use client.crt and client.key.

>Does Kubernetes do the same thing? I can't seem to find any documentation about it. In fact, there's very little documentation anyway I can find about securing docker registries with certificates instead of basic auth.

I  don't have much history on why we secured the registry that way in Fedora. Maybe @maxamillion or @puiterwijk could help with that

cverna:
Ok, so I am not sure how to proceed with this change. 

> In that case, this seems to be an ad-hoc method for storing a certificate and key. :(

I tend to agree here with @twaugh, and it seems that the setup we have in Fedora infra is not the norm so maybe keeping this change as a patch maintained only in the fedora rpm is the way to go.

Another solution would be to go with @owtaylor first proposal 

> skopeo reads /etc/docker/certs.d, so could require configuration of certificates in the builder image. But for that approach, it would be good if the atomic-reactor code would also just understand certificates in that location, to avoid having to configure the certificates in three places. (And secrets in the builder image are a ratherugly.)

And change atomic-reactor so that it will use the certs in /etc/docker/certs.d in the builder image.

I might go the patch in Fedora way while working on atomic-reactor to use certs present in /etc/docker/certs.d.

owtaylor:
I don't have a firm recommendation. The most straightforward way to just get things working would be to switch Fedora authentication to the registry to use passwords. I don't think the security of that is significantly lower, but it does require different configuration - you'd either have to hardcode services in the docker registry configuration. or do a bunch of LDAP configuration work. Feedback from Randy and Patrick about whether this is an acceptable route would be great.

The second most straightforward route would be, as you suggest, extending this patch to read  /etc/docker/certs.d and baking the certs into the buildroot image. This definitely violates best practices for container management - but might be secure enough if we check that there is no way for the buildroot image to be leaked out of Fedora infrastructure.

Or, you could just proceed with the approach here, but that's going to require skopeo work - I can help out with that if you want to proceed that way.

(From the Fedora perspective, the best thing would be OIDC authentication, but that would require seeing if the differences between docker-registry Oauth2 and Fedora OIDC can be papered over, AND adding support for Oauth2 to both this code and to skopeo ... an even bigger project.)

cverna:
> The most straightforward way to just get things working would be to switch Fedora authentication to the registry to use passwords

I think we are heading in that direction, that will make more sense, therefore I am closing this PR.

PR #985, opened 2018-04-26
Increase default pulp_pull timeout (OSBS-5544)

twaugh:


twaugh:
Will re-base once #987 is merged.

PR #986, opened 2018-04-26
test_fs_watcher: make sure the thread has time to start

twaugh:


sosiouxme:
So I don't want to sit around and bikeshed about this, but... I'd rather fix a race condition logically rather than sleeping. Consider: https://github.com/projectatomic/atomic-reactor/pull/987

twaugh:
I like that much better. Closing.

PR #992, opened 2018-05-04
Initial imagebuilder buildstep plugin

sosiouxme:
Introduces an alternate method for running the docker build with imagebuilder. Can be configured with the following in reactor-config-map:
`default_image_build_method: imagebuilder`

Or per repo in `container.yaml`:
`override_image_build: imagebuilder`

A scratch build of imagebuilder for testing in the buildroot is available at http://download.eng.bos.redhat.com/brewroot/work/tasks/7796/15977796/imagebuilder-0.0.1-1.git.0.71d05b5.el7.x86_64.rpm

Commits are structured to review individually.


tests/plugins/test_imagebuilder.py
```diff
@@ -0,0 +1,124 @@
...
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import unicode_literals
+
+import requests
```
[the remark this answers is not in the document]
sosiouxme: oh yeah, `pyflakes tests` for some real fun.

atomic_reactor/plugins/build_imagebuilder.py
```diff
@@ -0,0 +1,70 @@
...
+        # TODO: directly invoke go imagebuilder library in shared object via python module
+        ib_process = subprocess.Popen([
+            'imagebuilder', '-t', image, builder.df_dir,
+        ], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+        self.log.debug('build is submitted, waiting for it to finish')
+        (output, errors) = ([], [])
+        while True:
+            poll = ib_process.poll()
+            out = ib_process.stdout.readline().decode()
```
sosiouxme: python3 tests fail because `decode()` no longer applies to strings. I'm at a loss for how get this to pass for both python 2 and 3. I added the `decode` due to this comment [here](https://www.endpoint.com/blog/2015/01/28/getting-realtime-output-using-python):

> In Python 3.x the process will hang because 'output' is a byte array instead of a string.
> Make sure you decode readline() into a string thus:
>
> output = process.stdout.readline().decode()

This is fine for 2 but not 3. And the other suggestion, adding an encoding:

> process = subprocess.Popen(cmd, stdout=subprocess.PIPE, encoding='utf8')

Works for 3 but not 2. Suggestions please... Simplest solution might be to use `communicate()` and just not try to retrieve the logs until it completes, but that feels wrong too.

rcerven: you can do the same as docker does

if six.PY3:
   data = data.decode('utf-8')

atomic_reactor/plugins/build_imagebuilder.py
```diff
@@ -0,0 +1,76 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
```
twaugh: ... time ghost ...

atomic_reactor/plugins/build_imagebuilder.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # NOTE: it may make more sense to record this in the BuildResult
+        self.workflow.skip_layer_squash = True
+
+        image_id = builder.get_built_image_info()['Id']
+        if ':' not in image_id:
+            # Older versions of the daemon do not include the prefix
+            image_id = 'sha256:{}'.format(image_id)
+
+        return BuildResult(logs=output, image_id=image_id)
```
twaugh: If there is output on stderr but the exit code is 0, is that output lost?

atomic_reactor/plugins/build_imagebuilder.py
```diff
@@ -0,0 +1,76 @@
...
+                errors.append(err)
+            if out == '' and err == '':
+                if poll is not None:
+                    break
+                sleep(0.1)  # don't busy-wait when there's no output
+
+        if ib_process.returncode != 0:
+            return BuildResult(logs=output, fail_reason="image not built: " + "".join(errors))
+
+        # NOTE: it may make more sense to record this in the BuildResult
```
twaugh: Yes, that might be neater.

atomic_reactor/plugins/build_imagebuilder.py
twaugh: That sounds like a good approach, yes.

atomic_reactor/plugins/build_imagebuilder.py
sosiouxme: Thanks for that, it was simpler than I was thinking :)

atomic_reactor/plugins/build_imagebuilder.py
```diff
@@ -0,0 +1,77 @@
...
+                output.append(out)
+            err = sixdecode(ib_process.stderr.readline())
+            if err:
+                self.log.error(err.strip())
+                output.append(err)  # include stderr with stdout
+                errors.append(err)  # as well as separately
+            if out == '' and err == '':
+                if poll is not None:
+                    break
+                sleep(0.1)  # don't busy-wait when there's no output
```
twaugh: This line doesn't get covered during unit tests. Is it feasible to arrange for tests to hit this line? For example, maybe a test could `flexmock(time).should_receive('sleep').with_args(0.1).and_raise(NotImplementedError)`, and then run the plugin `with pytest.raises(NotImplementedError): ...`.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -337,6 +338,14 @@ def get_build_image_override(workflow, fallback=NO_FALLBACK):
     return get_value(workflow, 'build_image_override', fallback)
 
 
+def get_default_image_build_method(workflow, logger, fallback=CONTAINER_DEFAULT_BUILD_METHOD):
+    value = get_value(workflow, 'default_image_build_method', fallback)
+    if value not in CONTAINER_BUILD_METHODS:
+        logger.error("invalid default_image_build_method '{}'; using default".format(value))
+        return fallback
```
twaugh: These two lines are not covered by unit testing.

atomic_reactor/source.py
```diff
@@ -31,8 +32,29 @@
...
+                # read file and validate against schema
+                self.data = read_yaml_from_file_path(file_path, 'schemas/container.json') or {}
+            except Exception:
+                logger.exception("Failed to load and validate source config YAML from " + file_path)
+                # and proceed with default source config, that is, none
+
+        self.autorebuild = self.data.get('autorebuild', {})
+        self.override_image_build = oib = self.data.get('override_image_build')
+        if oib is not None and oib not in CONTAINER_BUILD_METHODS:
+            raise RuntimeError("unknown build method " + oib)
```
twaugh: This line is not covered by unit tests.

atomic_reactor/plugins/pre_reactor_config.py
sosiouxme: Should be now

atomic_reactor/source.py
sosiouxme: Should be now

sosiouxme:
One other thing to bring up about this implementation: as nothing was validating container.yaml with schema before, users could see new failures to read their existing container.yaml once this is rolled out. That is one reason only the new property is actually using the new SourceConfig reader; the others are continuing to roll their own yaml reading. A failure in the SourceConfig reader just means that the config is considered empty and there are errors logged, so only the new property is affected by the schema validation. I think it would be best if everything reading container.yaml did it via SourceConfig eventually but didn't want to make that change immediately. Feedback welcome however.

twaugh:
Yes, I'm in favour of validation everywhere.

atomic_reactor/schemas/container.json
```diff
@@ -16,6 +16,18 @@
...
+    "autorebuild": {
+      "type": "object"
+    },
+    "compose": {
+      "type": "object"
+    },
+    "flatpak": {
+      "type": "object"
+    },
+    "override_image_build": {
```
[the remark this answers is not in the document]
sosiouxme: Fine with me, makes sense to be consistent.

atomic_reactor/source.py
```diff
@@ -31,8 +32,29 @@
...
+    def __init__(self, build_path):
+        self.data = {}
+        file_path = os.path.join(build_path, REPO_CONTAINER_CONFIG)
+        if os.path.exists(file_path):
+            try:
+                # read file and validate against schema
+                self.data = read_yaml_from_file_path(file_path, 'schemas/container.json') or {}
+            except Exception:
+                logger.exception("Failed to load and validate source config YAML from " + file_path)
+                # and proceed with default source config, that is, none
```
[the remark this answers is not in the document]
sosiouxme: It would make sense to fail the build at this point (or is there another good alternative?). I am just thinking of those users who might already have a config that doesn't validate, but works given the existing lack of validation. They'll be forced to understand the schema and fix their config to do any kind of build. If that's an acceptable UX... sure, just propagate the exception. Now does seem like the least risky time to start enforcing validation.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -337,6 +338,14 @@ def get_build_image_override(workflow, fallback=NO_FALLBACK):
     return get_value(workflow, 'build_image_override', fallback)
 
 
+def get_default_image_build_method(workflow, logger, fallback=CONTAINER_DEFAULT_BUILD_METHOD):
```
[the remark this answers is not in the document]
sosiouxme: Would not need a logger at all if not swallowing the exception.

sosiouxme:
Added commits (currently last 4) to address feedback and rebased. Schema validation is applied everywhere and can fail the build. PTAL :)

lcarva:
Very well done!

lcarva:
Updated release notes draft:
> - New container build plugin using [imagebuilder](https://github.com/openshift/imagebuilder) with ability of choosing default approach per environment, as well as, per repository.

ben-alkov:
@sosiouxme Can you refresh the imagebuilder rpm link? I'm getting a 404...

PR #993, opened 2018-05-10
Don't log failures as errors in get_manifests()

twaugh:
Fixes OSBS-5453.

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -876,7 +876,7 @@ def get_manifest(image, registry_session, version):
             raise
 
     if not manifest_is_media_type(response, media_type):
-        logger.error("content does not match expected media type")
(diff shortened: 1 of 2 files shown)


mlangsdorf:
The comment "shouldn't log failures" in the unit tests is confusing. I've read the pydoc-catchlog docs twice and I'm still not 100% sure what is going on there.
I get that we want to fail if the warning messages are reported as errors, just not sure how you're testing that.

twaugh:
I added some comments to hopefully make things clearer. The overall approach is: set the log level for captured messages to ERROR, then afterwards verify that no log messages were captured -- then we know no ERROR+ messages were logged.

twaugh:
Will re-base after #995 is merged.

mlangsdorf:
Thanks, that's a lot less cryptic.

PR #994, opened 2018-05-10
Use catchlog instead of capturelog

twaugh:
In pytest-3.3, included in Fedora 28, the 'caplog' fixture is built-in, merged from the pytest-catchlog package. This has a different interface to pytest-capturelog:
http://blog.pytest.org/2017/whats-new-in-pytest-33/

Use this new interface, with pytest-catchlog installed to handle versions of pytest prior to the merge.

### DIFF
--- a/tests/plugins/test_add_dockerfile.py
+++ b/tests/plugins/test_add_dockerfile.py
@@ -183,7 +183,7 @@ CMD blabla"""
         }]
     )
     runner.run()
-    assert "plugin 'add_dockerfile' raised an exception: ValueError" in caplog.text()
+    assert "plugin 'add_dockerfile' raised an exception: ValueError" in caplog.text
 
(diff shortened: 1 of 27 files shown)


mlangsdorf:
travis is consistently failing with on '+docker exec -ti atomic-reactor-fedora-27-py3 pytest -vv tests --cov atomic_reactor' with the final message of "argparse.ArgumentError: argument --log-format: conflicting option string: --log-format".

I don't see any obvious reason why these changes are failing that way.

lcarva:
It's probably related to use catchlog instead of capturelog. BTW, we're still running tests on fedora 26.

twaugh:
Not required. Failures seen locally were due to system-install dependencies, whereas our tests install from pip and avoid the trouble.

PR #995, opened 2018-05-11
tests: don't let builddep enable *-source repos on CentOS

twaugh:
They give 404 errors and prevent the test run.


mlangsdorf:
This is an internal change to the Travis tests - I don't think it needs a release note.

PR #998, opened 2018-05-15
tag_from_config: de-duplicate tag values (OSBS-5638)

twaugh:


tests/plugins/test_tag_from_config.py
```diff
@@ -148,6 +148,9 @@ def test_bad_inspect_data(tmpdir, docker_tasker, inspect, error):
                                      'name_value:version_value']),
     (['foo', 'bar'], ['{version}-{release}'],
      ['name_value:foo', 'name_value:bar', 'name_value:version_value-7.4.1']),
+    (['foo', 'bar'], ['{version}', 'version_value', 'baz'],
```
mlangsdorf: I know this is a sufficient test, but setting it to:
['bar', 'foo', 'bar'], ['baz', '{version}', 'version_value', 'baz']
would make it explicit and obvious that we're testing for duplicate tags.

lcarva:
Updated release notes draft:
> - tag values are de-duplicated

PR #1000, opened 2018-05-15
If expecting only v2 manifest list, don't expect v2 manifest when pulling image

rcerven:

### DIFF
--- a/tests/plugins/test_pulp_pull.py
+++ b/tests/plugins/test_pulp_pull.py
@@ -501,6 +501,7 @@ class TestPostPulpPull(object):
                 plugin.run()
 
     @responses.activate
+    @pytest.mark.parametrize('expect_v2', [True, False])
     @pytest.mark.parametrize(('platforms', 'platform_descriptors', 'manifest_list_only'), [
         (['ppc64le'], True, True),
         (['ppc64le'], False, False),
@@ -511,7 +512,8 @@ class TestPostPulpPull(object):
         (['x86_64'], True, False),
         (['x86_64'], False, False),
     ])
-    def test_expect_v2schema2list_only(self, platforms, platform_descriptors, manifest_list_only):
+    def test_expect_v2schema2list_only(self, caplog, expect_v2, platforms, platform_descriptors,
+                                       manifest_list_only):
         def get_manifest(request):
             media_types = request.headers.get('Accept', '').split(',')
             content_type = media_types[0]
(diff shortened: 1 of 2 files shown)


twaugh:
I feel like lines 99-102 need to be updated:
```
if self.expect_v2schema2 and not digests.v2:
    self.log.warn("Expected schema 2 manifest")
elif self.expect_v2schema2list and not digests.v2_list:
    self.log.warn("Expected schema 2 manifest list")
```

to something like:
```
if self.expect_v2schema2list and not digests.v2_list:
    ...
elif not self.expect_v2schema2list_only and self.expect_v2schema2 and not digests.v2:
    ...
```

In other words, if expect_v2schema2list_only is truthy, it shouldn't really matter what expect_v2schema2 is.

rcerven:
@twaugh  yop makes more sense, will change

tests/plugins/test_pulp_pull.py
```diff
@@ -528,11 +530,121 @@ def get_manifest(request):
...
+
+        # (for v1, v2, list.v2, oci, and oci.index media types) before get_manifest_digests
+        # v1
+        if expect_v1:
+            expectation.and_return(self.config_response_config_v1)
+        else:
+            expectation.and_return(not_found)
+
+        # v2
+        if found_v2:
```
lcarva: It is confusing to use variable `found_v2` to set up the expectation.

tests/plugins/test_pulp_pull.py
```diff
@@ -528,11 +530,121 @@ def get_manifest(request):
...
+                    plugin.run()
+                if expect_v1 or found_v2:
+                    assert "Expected schema 2 manifest list" in caplog.text()
+        else:
+            if found_v2_list and found_v2:
+                media_types = plugin.run()
+            else:
+                with pytest.raises(Exception):
+                    plugin.run()
+                if not found_v2_list and (found_v2 or expect_v1):
```
lcarva: The different conditionals in this test makes it very difficult to actually determine the expected behavior.

tests/plugins/test_pulp_pull.py
lcarva: Actually the terminology used in this test is ambiguous to me. What's meant by `found_*` vs `expect_*` ?

I think something like `expect_*` vs `has_*` is more straightforward.

tests/plugins/test_pulp_pull.py
```diff
@@ -528,11 +530,121 @@ def get_manifest(request):
...
         expected_media_types = [MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST]
         if not manifest_list_only:
             expected_media_types.append(MEDIA_TYPE_DOCKER_V2_SCHEMA1)
             expected_media_types.append(MEDIA_TYPE_DOCKER_V2_SCHEMA2)
+        else:
+            assert "Only V2 schema 2 manifest list is expected, " in caplog.text()
         assert set(media_types) == set(expected_media_types)
+
+    @responses.activate
+    @pytest.mark.parametrize(('found_v2', 'found_v2_list'), [
```
lcarva: How about:
```
@pytest.mark.parametrize(('platforms', 'has_v1', 'has_v2', 'has_v2_list', 'expect_v1', 'expect_v2', 'expect_v2_list')
```

This will make it easier to tell what the different scenarios being tested are.

rcerven: sure can do that, I just wanted to cover all v2  & v2 list cases
as v1 isn't very useful variable in this test

rcerven: also expect_v1 is not exactly we care about at all in pulp_pull

PR #1002, opened 2018-05-18
schema validation fixes

sosiouxme:
Now that container.yaml is being validated everywhere we need the validation to be correct. Adding some properties in schema and fixing up the error handling when validation fails (OSBS-5173).


sosiouxme:
In addition to bug fix, release notes should probably reflect that container.yaml schema is being validated now.

lcarva:
Updated release notes draft:
> - `container.yaml` is validated against schema

PR #1003, opened 2018-05-18
plugins/post_compress.py: export image if needed

sosiouxme:
With squash optionally omitted depending on the build method now, the
built image may not be exported from docker once it's time for compress
to run. So in that case, attempt to export it directly to the
compression stream.

### DIFF
--- a/tests/plugins/test_compress.py
+++ b/tests/plugins/test_compress.py
@@ -34,12 +34,14 @@ class X(object):
 
 
 class TestCompress(object):
-    @pytest.mark.parametrize('method, load_exported_image, extension', [
-        ('gzip', False, 'gz'),
-        ('lzma', False, 'xz'),
-        ('gzip', True, 'gz'),
+    @pytest.mark.parametrize('method, load_exported_image, give_export, extension', [
+        ('gzip', False, True, 'gz'),
+        ('lzma', False, False, 'xz'),
+        ('gzip', True, True, 'gz'),
+        ('gzip', True, False, 'gz'),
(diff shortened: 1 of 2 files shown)


sosiouxme:
It's not completely clear to me if this will work for base images -- need to figure out how to test those.

tests/plugins/test_compress.py
```diff
@@ -34,12 +34,13 @@ class X(object):
...
 class TestCompress(object):
-    @pytest.mark.parametrize('method, load_exported_image, extension', [
-        ('gzip', False, 'gz'),
-        ('lzma', False, 'xz'),
-        ('gzip', True, 'gz'),
+    @pytest.mark.parametrize('method, load_exported_image, give_export, extension', [
+        ('gzip', False, True, 'gz'),
+        ('lzma', False, False, 'xz'),
+        ('gzip', True, True, 'gz'),
+        ('gzip', True, False, 'gz'),
```
mlangsdorf: Since you're editing this test, could you add a parameter for an unsupported compression format test here?

lcarva:
This looks fine to me. I don't expect base images would behave differently at this point, but definitely needs to be tested.

sosiouxme:
Stage build rhel-server-docker-private-lucarval-extras-rhel-713ec-18 ran the compress plugin with `load_exported_image: False` and it compressed a base image straight from docker just fine.

mlangsdorf:
Added to release notes under Improvements
- in the post_compress plugin, images will be exported even if the squash plugin did not run

PR #1004, opened 2018-05-22
pull base image searches for manifest_list names and uses that instea…

rcerven:
…d of v2 digest sha


atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -170,6 +172,24 @@ def validate_platforms_in_base_image(self, base_image):
...
+            base_image = base_image_result.copy()
+
+            try:
+                config_blob = get_config_from_registry(base_image, base_image.registry,
+                                                       base_image.tag,
+                                                       insecure=self.parent_registry_insecure)
+            except (HTTPError, RetryError, Timeout) as ex:
+                self.log.warning('Unable to fetch config for %s, got error %s',
+                                 base_image, ex.response.status_code)
+                raise RuntimeError('Unable to fetch config for base image')
```
mlangsdorf: Why are you raising an error here, instead of falling through to the get_manifest_list() path?
what's the advantage of searching for the manifest_list by the manifest_list_name instead of the v2 sha digest?

rcerven: and we can't search manifest list with v2 sha digest only

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -170,6 +172,24 @@ def validate_platforms_in_base_image(self, base_image):
                           'because platform descriptors are not defined')
             return
 
+        if '@sha256:' in str(base_image):
+            base_image_result = ImageName.parse(str(base_image))
+            base_image = base_image_result.copy()
```
twaugh: What's this for? I don't understand it. Isn't `base_image` an ImageName here? And why do we need to copy it?

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -170,6 +172,24 @@ def validate_platforms_in_base_image(self, base_image):
...
+
+            try:
+                config_blob = get_config_from_registry(base_image, base_image.registry,
+                                                       base_image.tag,
+                                                       insecure=self.parent_registry_insecure)
+            except (HTTPError, RetryError, Timeout) as ex:
```
twaugh: But does this retry? I think it should, but I can't see what would drive that.

rcerven: get_config_from_registry is using RegistrySession, which has retries implemented as any other method in utils using that session

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: yes base_image is indeed ImageName, but look at line 191, I want to change tag to manifest_list nvr, but JUST for get_manifest_list, IF I'd change directly base_image without copying it, it would try later to pull manifest_list instead of that specified sha, which I want to keep unchanged for pulling

twaugh: So should these two lines be replaced by one?:
```
base_image = base_image.copy()  # we want to adjust the tag only for manifest list fetching
```

rcerven: right yes, will do that

rcerven:
rn updated "pull base image searches for manifest_list names and uses that instead of v2 digest sha"

PR #1007, opened 2018-05-31
test.sh: add resurrection, EXTRA_MOUNT, htmlcov

sosiouxme:
Have this reuse an existing container that's been stopped.
Also make it easy to mount more volumes into the test container.
Generate HTML coverage report.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 *.py[co]
-*.swp
+*.sw[po]
 dist/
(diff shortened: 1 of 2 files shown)


lcarva:
Worth doing this for osbs-client too?

sosiouxme:
Also https://github.com/DBuildService/dockerfile-parse/pull/60

Issue #1008, opened 2018-06-03
Missing base image retries with "library" namespace


sosiouxme:
I came across this:

https://github.com/projectatomic/atomic-reactor/blob/d058ca2a6db25391f01f0c430c397b1891e558fa/atomic_reactor/plugins/pre_pull_base_image.py#L103-L108

I think the intent is to retry `example.com/spam` as `example.com/library/spam` but it looks to me like it will also retry `example.com/spam/eggs` as `example.com/library/eggs` which would be a bit surprising. Simple enough to fix, just checking if there's something I'm missing here.

sosiouxme:
@twaugh @rcerven

twaugh:
I think you're right.

rcerven:
yop it should be adding 'library/' in front of namespace

twaugh:
I think it should be leaving repository names alone if they already contain a 'namespace' part. The 'library' prefix is a bit of a historical thing really.
* [name] -> [list of names to try]
* example.com/spam -> example.com/spam, example.com/library/spam
* example.com/spam/eggs -> example.com/spam/eggs

PR #1009, opened 2018-06-04
pull_base_image: fix retry pull w/ library namespace

sosiouxme:
This retry should only occur for images that have no namespace already.

fix https://github.com/projectatomic/atomic-reactor/issues/1008

### DIFF
--- a/tests/plugins/test_pull_base_image.py
+++ b/tests/plugins/test_pull_base_image.py
@@ -190,6 +190,16 @@ def test_pull_base_wrong_registry(reactor_config_map):  # noqa
                                     reactor_config_map=reactor_config_map)
 
 
+# test previous issue https://github.com/projectatomic/atomic-reactor/issues/1008
+def test_pull_base_library(reactor_config_map, caplog):  # noqa
+    with pytest.raises(PluginFailedException) as exc:
+        test_pull_base_image_plugin(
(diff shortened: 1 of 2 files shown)


tests/plugins/test_pull_base_image.py
```diff
@@ -190,6 +190,15 @@ def test_pull_base_wrong_registry(reactor_config_map):  # noqa
                                     reactor_config_map=reactor_config_map)
 
 
+# test previous issue https://github.com/projectatomic/atomic-reactor/issues/1008
+def test_pull_base_library(reactor_config_map):  # noqa
+    with pytest.raises(PluginFailedException) as exc:
+        test_pull_base_image_plugin(
+            LOCALHOST_REGISTRY, "spam/library-only:latest", [], [], reactor_config_map
+        )
+    assert "not found" in str(exc.value)
```
rcerven: this should be 'not in' as it should raise error as namespace is defined

but the raise error actually raises : 
2018-06-04 21:20:21,354 platform:- - atomic_reactor.plugin - ERROR - plugin 'pull_base_image' raised an exception: RetryGeneratorException("Failed to mock_method image (u'localhost:5000/spam/library-only',): u'Error: image ***:latest not found'",)

sosiouxme:
Something to put in the release notes?

lcarva:
Updated release notes draft:
> - `pull_base_image` does not replace namespace with "library" string when retrying

PR #1010, opened 2018-06-07
always override auto labels

rcerven:

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -162,9 +162,6 @@ class AddLabelsPlugin(PreBuildPlugin):
             if lbl not in generated:
                 self.log.warning("requested automatic label %r is not available", lbl)
 
-            elif lbl in plugin_labels:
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -162,7 +166,7 @@ def generate_auto_labels(self, base_labels, df_labels, plugin_labels):
             if lbl not in generated:
                 self.log.warning("requested automatic label %r is not available", lbl)
 
-            elif lbl in plugin_labels:
+            elif lbl in plugin_labels and lbl not in self.auto_always_overwrite:
```
twaugh: Is there any particular reason you can think of not to just *always* overwrite the labels mentioned in `auto_labels`?

lcarva: Given the plugin defaults, always overwriting makes a lot of sense.

rcerven: ok, will always overwrite all auto_labels

lcarva:
@twaugh, waiting to merge this until you approve since you had comments, although, I think they have been addressed.

PR #1013, opened 2018-06-11
post_push_to_pulp: export image if not exported

sosiouxme:
With an imagebuilder build, the squash plugin is not invoked, thus the image is not exported from Docker. With the `pulp_push` plugin in sequence before `compress`, it finds no image to load and fails. So, this is an update, like the one made to `compress`, to export the image first in that case.

A more correct fix to work with more varied plugin sequences would be to have the imagebuilder plugin export the image itself. However I ran into problems trying to get this plugin to accept that exported image and thought this fix could stand on its own (it addressed the issue when tested in stage) until I figure that out.

### DIFF
--- a/tests/plugins/test_pulp.py
+++ b/tests/plugins/test_pulp.py
@@ -150,7 +150,7 @@ def prepare(check_repo_retval=0, existing_layers=[],
     return tasker, workflow
 
 
-@pytest.mark.skipif(dockpulp is None,
+@pytest.mark.skipif(dockpulp is None,  # noqa: F811; redefinition of 'reactor_config_map'
                     reason='dockpulp module not available')
 @pytest.mark.parametrize(("unsupported"), [
     (True),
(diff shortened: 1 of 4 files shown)


sosiouxme:
Belay that, I figured out what was going wrong with the export in imagebuilder.

tests/plugins/test_pulp.py
```diff
@@ -150,7 +150,7 @@ def prepare(check_repo_retval=0, existing_layers=[],
     return tasker, workflow
 
 
-@pytest.mark.skipif(dockpulp is None,
+@pytest.mark.skipif(dockpulp is None,  # noqa
```
twaugh: What's the diagnostic code we're waiving here? Better to be explicit if we can be, e.g. `# noqa:W811`

sosiouxme: Good point.

`F811 redefinition of unused 'reactor_config_map' from line 36`

I don't think stickler actually complains about this but it's noise in flake8

twaugh: I think using `noqa:F811` (with `:`) actually silences that particular diagnostic only.

sosiouxme: It turns out the specific silencing only works if you use just the right format in the comment. Ironic, I know.

`# noqa: F811` silences F811
`# noqa:F811` silences everything

twaugh: :open_mouth:

sosiouxme:
bugfix for unreleased bug, no need for release notes

PR #1014, opened 2018-06-14
Move some commonly-used stubs into tests/stubs.py

twaugh:
Inspired by this article:
https://testing.googleblog.com/2018/02/testing-on-toilet-cleanly-create-test.html


mlangsdorf:
Is the intent to convert all 29 test files that could use this format over now, or just have an example and convert as people are working in those files?

twaugh:
I'm happy with either approach. Converting over time is less likely to run into conflicts, and we don't have time put aside for working on this specifically.

lcarva:

PR #1017, opened 2018-06-15
fix reactor config test

rcerven:


sosiouxme:
python deps v2 versus v3 as I recall... might be worth a comment or something because it's stupid

rcerven:
https://github.com/projectatomic/atomic-reactor/pull/1020
fix for parsing message

PR #1021, opened 2018-06-18
Add support for dockerconfigjson secret file.

cverna:
The only different between the current .dockercfg and .dockerconfigjson format is an extra `auths` key in the later.
 Example : 
.dockercfg
```{"yourprivateregistry.com":{"username":"janedoe","password":"xxxxxxxxxxx","email":"jdoe@example.com","auth":"c3R...zE2"}}}```
.dockerconfigjson
```{"auths":{"yourprivateregistry.com":{"username":"janedoe","password":"xxxxxxxxxxx","email":"jdoe@example.com","auth":"c3R...zE2"}}}```

So this solution offers to support both file, and detect if we are using a dockerconfigjson if the `auths` key is present.

If this solution is accepted, I ll be adding tests and documentation.

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -613,16 +613,27 @@ def registry_hostname(registry):
 class Dockercfg(object):
     def __init__(self, secret_path):
         """
-        Create a new Dockercfg object from a .dockercfg file whose
+        Create a new Dockercfg object from a .dockercfg/.dockerconfigjson file whose
         containing directory is secret_path.
 
-        :param secret_path: str, dirname of .dockercfg location
+        :param secret_path: str, dirname of .dockercfg/.dockerconfigjson location
         """
 
-        self.json_secret_path = os.path.join(secret_path, '.dockercfg')
+        if os.path.exists(os.path.join(secret_path, '.dockercfg')):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by lcarva** [comment]
@twaugh, could you have a look at this one? You're more familiar with the format of these files.

File: atomic_reactor/util.py [hunk 1]
```diff
@@ -619,10 +619,21 @@ def __init__(self, secret_path):
...
+        else:
+            raise RuntimeError("The registry secret was not found on the filesystem, "
+                               "either .dockercfg or .dockerconfigjson are supported")
         try:
             with open(self.json_secret_path) as fp:
                 self.json_secret = json.load(fp)
+
+            # If the auths key exist then we have a dockerconfigjson secret.
+            if self.json_secret.get('auths') is not None:
+                self.json_secret = self.json_secret.get('auths')
```
twaugh: The comment doesn't technically describe what the code does ('if auths key exists' vs `if [value] is not None`). Maybe re-write as:
```
if 'auths' in self.json_secret:
    self.json_secret = self.json_secret['auths']
```

lcarva:
Updated release notes draft:
> - Add support for dockerconfigjson secret file.

PR #1022, opened 2018-06-18
handle ConnectionErrors from koji.ClientSession

mlangsdorf:
koji.ClientSession is supposed to retrying ConnectionErrors, but it doesn't
appear to do so correctly. Add a wrapper to koji.ClientSession that will catch
ConnectionErrors generated by methods and retry the method up to 10 times,
only raising if there is still a ConnectionError or if there is some other
error.

### DIFF
--- a/atomic_reactor/koji_util.py
+++ b/atomic_reactor/koji_util.py
@@ -10,11 +10,14 @@ from __future__ import print_function
 
 
 import koji
+from requests.exceptions import ConnectionError
+
 import logging
 import os
 import time
 
-from atomic_reactor.constants import DEFAULT_DOWNLOAD_BLOCK_SIZE
+from atomic_reactor.constants import (DEFAULT_DOWNLOAD_BLOCK_SIZE,
+                                      HTTP_BACKOFF_FACTOR, HTTP_MAX_RETRIES)
 
 logger = logging.getLogger(__name__)
 
@@ -38,6 +41,35 @@ class KojiUploadLogger(object):
             self.last_percent_done = percent_done
             self.logger.debug("upload: %d%% done (%.1f MiB/sec)",
                               percent_done, size / t1 / 1024 / 1024)
+
+
+class KojiSessionWrapper(object):
+    """
+    Wrap all calls to koji.ClientSession methods in a catch/exception block, so that
+    improperly handled ConnectionErrors from koji.ClientSession will get retried silently.
+
+
+    """
+    def __init__(self, session):
+        self._wrapped_session = session
+
+    def __getattr__(self, name):
+        session_attr = getattr(self._wrapped_session, name)
+        if callable(session_attr):
+            def call_with_catch(*a, **kw):
+                retry_delay = HTTP_BACKOFF_FACTOR
+                last_exc = None
+                for retry in range(HTTP_MAX_RETRIES):
+                    try:
+                        return session_attr(*a, **kw)
+                    except ConnectionError as exc:
+                        time.sleep(retry_delay * (2 ** retry))
+                        last_exc = exc
+                        continue
+                raise last_exc
+            return call_with_catch
+        else:
+            return session_attr
 
 
 def koji_login(session,
@@ -103,7 +135,7 @@ def create_koji_session(hub_url, auth_info=None):
     :param auth_info: dict, authentication parameters used for koji_login
     :return: koji.ClientSession instance
     """
-    session = koji.ClientSession(hub_url, opts={'krb_rdns': False})
+    session = KojiSessionWrapper(koji.ClientSession(hub_url, opts={'krb_rdns': False}))
 
     if auth_info is not None:
         koji_login(session, **auth_info)
--- a/tests/test_koji_util.py
+++ b/tests/test_koji_util.py
@@ -7,6 +7,9 @@ of the BSD license. See the LICENSE file for details.
 """
 
 from __future__ import absolute_import, print_function, unicode_literals
+import time
+
+from requests.exceptions import ConnectionError
 
 try:
     import koji
@@ -29,6 +32,7 @@ from atomic_reactor.koji_util import (koji_login, create_koji_session,
                                       TaskWatcher, tag_koji_build)
 from atomic_reactor import koji_util
 from atomic_reactor.plugin import BuildCanceledException
+from atomic_reactor.constants import HTTP_MAX_RETRIES
 import flexmock
 import pytest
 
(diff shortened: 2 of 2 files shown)


atomic_reactor/koji_util.py
```diff
@@ -40,6 +42,33 @@ def callback(self, offset, totalsize, size, t1, t2):  # pylint: disable=W0613
...
+
+    def __getattr__(self, name):
+        session_attr = getattr(self._wrapped_session, name)
+        if callable(session_attr):
+            def call_with_catch(*a, **kw):
+                for retry in range(10):
+                    try:
+                        return session_attr(*a, **kw)
+                    except ConnectionError as exc:
+                        continue
```
lcarva: Should we add some sort of back off logic here?

twaugh:
Yes, there is a `retry()` method in atomic_reactor/core.py that has the sort of logic we want. I don't think we can use it directly but we could make a new one for use with koji.

mlangsdorf:
Added the retry logic, fixed the weird exception error for Python 3.

tests/test_koji_util.py
```diff
@@ -101,16 +105,57 @@ def test_create_simple_session(self):
...
+        ('krb_login', {}),
+    ])
+    @pytest.mark.parametrize(('error_type', 'should_recover', 'attempts'), [
+        (ConnectionError, True, HTTP_MAX_RETRIES - 1),
+        (ConnectionError, False, HTTP_MAX_RETRIES),
+        (KeyError, False, 1),
+    ])
+    def test_create_session_failures(self, tmpdir, session_type, session_args,
+                                     error_type, should_recover, attempts):
+        url = 'https://koji-hub-url.com'
```
twaugh: Let's always use `.example.com` for domains in test cases.

mlangsdorf: Fixed.

mlangsdorf:
Relocated the sleep backoff so it actually executes, and removed the raise from koji_login since it was impossible to trigger since the underlying functions either return True or raise errors.

rcerven:
apart from mine 1 comment looks good

tests/test_koji_util.py
```diff
@@ -96,21 +100,64 @@ def test_koji_login_ssl(self, tmpdir, proxyuser, serverca):
...
+            session.should_receive('ssl_login').and_return(True)
+        else:
+            session.should_receive('krb_login').once().and_return(True)
 
         (flexmock(koji_util.koji).should_receive('ClientSession').with_args(
             url, opts={'krb_rdns': False}).and_return(session))
-        assert create_koji_session(url, {}) == session
+        assert create_koji_session(url, args)._wrapped_session == session
+
+    @pytest.mark.parametrize(('session_type', 'session_args'), [
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

atomic_reactor/koji_util.py
```diff
@@ -40,6 +43,38 @@ def callback(self, offset, totalsize, size, t1, t2):  # pylint: disable=W0613
...
+                retry_delay = HTTP_BACKOFF_FACTOR
+                last_exc = None
+                for retry in range(HTTP_MAX_RETRIES):
+                    try:
+                        return session_attr(*a, **kw)
+                    except ConnectionError as exc:
+                        time.sleep(retry_delay * (2 ** retry))
+                        last_exc = exc
+                        continue
+                    except Exception as exc:
```
sosiouxme: Why catch this at all? Should be the same result without it.

tests/test_koji_util.py
```diff
@@ -96,21 +100,64 @@ def test_koji_login_ssl(self, tmpdir, proxyuser, serverca):
...
+    ])
+    def test_create_authenticated_session(self, tmpdir, ssl_session):
+        url = 'https://example.com'
+        args = {}
 
-    def test_create_authenticated_session(self):
-        url = 'https://koji-hub-url.com'
         session = flexmock()
-        session.should_receive('krb_login').once().and_return(True)
+        if ssl_session:
```
sosiouxme: Seems like an unrelated change... is it just to improve code coverage?

tests/test_koji_util.py
```diff
@@ -96,21 +100,64 @@ def test_koji_login_ssl(self, tmpdir, proxyuser, serverca):
...
+            session.should_receive('ssl_login').and_return(True)
+        else:
+            session.should_receive('krb_login').once().and_return(True)
 
         (flexmock(koji_util.koji).should_receive('ClientSession').with_args(
             url, opts={'krb_rdns': False}).and_return(session))
-        assert create_koji_session(url, {}) == session
+        assert create_koji_session(url, args)._wrapped_session == session
+
+    @pytest.mark.parametrize(('auth_type', 'auth_args'), [
```
sosiouxme: Covers the same branch in auth logic as test case above; I'd rather this just focused on the one behavior of new interest, the exception retry.

mlangsdorf: The session wrapper code needs to handle 3 cases: genuine exception such as a KeyError, ConnectionError that times out, and momentary ConnectionError that clears on a retry. And it needs to do it for ssl and krb logins.

The clears on a retry eventually ends up in the same place as a successful session, but I think there's merit in testing both. Both to get 100% coverage and to show how the code is supposed to work.

atomic_reactor/koji_util.py
mlangsdorf: Fixed, thanks.

atomic_reactor/koji_util.py
```diff
@@ -88,9 +120,7 @@ def koji_login(session,
 
         result = session.krb_login(**kwargs)
 
-    if not result:
-        raise RuntimeError('Unable to perform Koji authentication')
-
+    # koji.ClientSession.*_login raise errors or return True, so do the same here.
```
lcarva: Why this change?

I think it's pretty bad that koji's lib doesn't raise an error if auth fails. This was a way to make its usage a little bit nicer. With this change, all callers must be modified to verify that auth actually succeeded.

mlangsdorf: Doh! I missed that *_login can return False under some circumstances. I'll fix the code and the test case.

mlangsdorf: Fixed.

atomic_reactor/koji_util.py
```diff
@@ -40,6 +43,35 @@ def callback(self, offset, totalsize, size, t1, t2):  # pylint: disable=W0613
...
+        session_attr = getattr(self._wrapped_session, name)
+        if callable(session_attr):
+            def call_with_catch(*a, **kw):
+                retry_delay = HTTP_BACKOFF_FACTOR
+                last_exc = None
+                for retry in range(HTTP_MAX_RETRIES):
+                    try:
+                        return session_attr(*a, **kw)
+                    except ConnectionError as exc:
+                        time.sleep(retry_delay * (2 ** retry))
```
twaugh: This will go:
* call -> exception
* time.sleep() ...
* call -> exception
* time.sleep() ......
* call -> exception
* time.sleep() ..................
* propagate exception

In other words, we are sleeping *after* each attempt, even the last. Don't we want to sleep *before* each attempt other than the first?

mlangsdorf:
added to release notes under # Bugfixes:
- koji connection errrors will be retried a few times instead of ending a build immediately

PR #1024, opened 2018-06-19
core.py:  Remove stream from build_image_from_path as it has been removed from the docker-py

jcajka:
Work around latest changes to the docker-py API breakage, removal of the stream parameter.

### DIFF
--- a/atomic_reactor/core.py
+++ b/atomic_reactor/core.py
@@ -323,7 +323,7 @@ class DockerTasker(LastLogger):
 
             return cmd_result
 
-    def build_image_from_path(self, path, image, stream=False, use_cache=False, remove_im=True):
+    def build_image_from_path(self, path, image, use_cache=False, remove_im=True):
         """
         build image from provided path and tag it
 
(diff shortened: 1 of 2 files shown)


atomic_reactor/core.py
```diff
@@ -344,9 +344,15 @@ def build_image_from_path(self, path, image, stream=False, use_cache=False, remo
...
-            response = self.d.build(path=path, tag=image.to_str(), stream=stream,
-                                    nocache=not use_cache, decode=True,
-                                    rm=remove_im, forcerm=True,)  # returns generator
+            try:
+                response = self.d.build(path=path, tag=image.to_str(), stream=stream,
+                                        nocache=not use_cache, decode=True,
+                                        rm=remove_im, forcerm=True,)  # returns generator
+            except TypeError:
+                # because changing api yet again is even more fun
+                response = self.d.build(path=path, tag=image.to_str(),
```
lcarva: So what happens if we want to stream? Is the API now returning a stream by default?

Do logs appear as each instruction is happening, or will they just show up when build operation is done?

jcajka: According to the upstream docs https://docker-py.readthedocs.io/en/stable/api.html#docker.api.build.BuildApiMixin.build it should.

cverna: In the docker-py version available in centos 7.5 stream is forced to true --> https://github.com/docker/docker-py/blob/1.10.6/docker/api/build.py#L59

jcajka: If we can agree on that I will remove the stream parameter from the usage of build_image_from_path.

Actually looking at the logs docs upstream the stream parameter is still there https://docker-py.readthedocs.io/en/stable/api.html#docker.api.container.ContainerApiMixin.logs, sorry for overlooking that.

jcajka: Oh and what about "pull" parameter?

cverna: > Oh and what about "pull" parameter?

So I might be wrong but it seems that the version of docker-py we are using all have the `pull` attribute for the build method.

Centos 7 docker-py 1.10.6 https://github.com/docker/docker-py/blob/1.10.6/docker/api/build.py#L18
Fedora 27 docker-py 2.6.1 https://github.com/docker/docker-py/blob/2.6.1/docker/api/build.py#L18
Fedora 28 docker-py 3.2.1 https://github.com/docker/docker-py/blob/3.2.1/docker/api/build.py#L18

So I guess calling with a `pull` attribute should just work.

lcarva:
Updated release notes draft:
> - Support for newer version of docker-py by removing `stream` parameter to `build_image_from_path` and `build` API methods.

PR #1025, opened 2018-06-21
enable multistage builds

sosiouxme:
Enhance atomic-reactor to correctly manage multi-stage build inputs and metadata.

Implementation of OSBS-5663. The commits are intended to be individually coherent for easier reviews.

### DIFF
--- a/atomic_reactor/build.py
+++ b/atomic_reactor/build.py
@@ -141,6 +141,8 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
         # arguments for build
         self.source = source
         self.base_image = None
+        self.original_base_image = None
+        self.parent_images = {}  # dockerfile image => locally available image
         self.image_id = None
         self.built_image_info = None
         self.image = ImageName.parse(image)
@@ -166,13 +168,21 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
 
     def set_df_path(self, path):
         self._df_path = path
-        self.set_base_image(df_parser(path).baseimage)
+        dfp = df_parser(path)
+        base = dfp.baseimage
+        if base is None:
+            raise RuntimeError("no base image specified in Dockerfile")
+        self.set_base_image(base)
         logger.debug("base image specified in dockerfile = '%s'", self.base_image)
-        if not self.base_image.tag:
-            self.base_image.tag = 'latest'
+        self.parent_images.clear()
+        for image in dfp.parent_images:
+            self.parent_images[image] = None
 
     def set_base_image(self, base_image):
+        logger.info("setting base image to '%s'", base_image)
         self.base_image = ImageName.parse(base_image)
+        self.original_base_image = self.original_base_image or self.base_image
+        self.parent_images[str(self.original_base_image)] = base_image
 
     def inspect_base_image(self):
         """
--- a/atomic_reactor/plugins/pre_change_from_in_df.py
+++ b/atomic_reactor/plugins/pre_change_from_in_df.py
@@ -6,46 +6,95 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 
 
-Pre build plugin which changes FROM instruction
+Pre-build plugin that changes the parent images used in FROM instructions
+to the more specific names given by the builder.
 """
-import fileinput
-import re
-import sys
 from atomic_reactor.plugin import PreBuildPlugin
-from atomic_reactor.util import ImageName
+from dockerfile_parse import DockerfileParser
+
+
+class BaseImageMismatch(RuntimeError):
+    pass
+
+
+class ParentImageUnresolved(RuntimeError):
+    pass
+
+
+class ParentImageMissing(RuntimeError):
+    pass
+
+
+class NoIdInspection(RuntimeError):
+    pass
 
 
 class ChangeFromPlugin(PreBuildPlugin):
     key = "change_from_in_dockerfile"
 
-    def __init__(self, tasker, workflow, base_image=None):
+    def __init__(self, tasker, workflow):
         """
         constructor
 
         :param tasker: DockerTasker instance
         :param workflow: DockerBuildWorkflow instance
-        :param base_image: str, change base image to ID of this image
         """
         # call parent constructor
         super(ChangeFromPlugin, self).__init__(tasker, workflow)
-        self.base_image = ImageName.parse(base_image) if base_image else None
 
     def run(self):
-        """
-        run the plugin
-        """
-        base_image = self.base_image or self.workflow.builder.base_image
-        try:
-            base_image_id = self.workflow.base_image_inspect['Id']
-        except KeyError:
-            self.log.error("Id is missing in inspection: '%s'", self.workflow.base_image_inspect)
-            raise
-        self.log.debug("using base image '%s', id '%s'", base_image, base_image_id)
-        for line in fileinput.input(self.workflow.builder.df_path, inplace=1):
-            re_match = re.match(r"^FROM .+$", line)
-            if re_match:
-                new_from = "FROM %s" % base_image_id
-                sys.stdout.write(new_from + '\n')
-                self.log.info("changed FROM: '%s' -> '%s'", re_match.group(0), new_from)
-            else:
-                sys.stdout.write(line)
+        builder = self.workflow.builder
+        dfp = DockerfileParser(builder.df_path)
+
+        df_base = dfp.baseimage
+        build_base = builder.base_image.to_str()
+
+        # do some sanity checks to defend against bugs and rogue plugins
+
+        if build_base != builder.parent_images[df_base]:
+            # something updated parent_images entry for base without updating
+            # the build's base_image; treat it as an error
+            raise BaseImageMismatch(
+                "Parent image '{}' does not match base_image '{}'"
+                .format(builder.parent_images[df_base], build_base)
+            )
+
+        unresolved = [key for key, val in builder.parent_images.items() if not val]
+        if unresolved:
+            # this would generally mean pull_base_image didn't run and/or
+            # custom plugins modified parent_images; treat it as an error.
+            raise ParentImageUnresolved("Parent image(s) unresolved: {}".format(unresolved))
+
+        missing = [df_img for df_img in dfp.parent_images if df_img not in builder.parent_images]
+        if missing:
+            # this would indicate another plugin modified parent_images out of sync
+            # with the Dockerfile or some other code bug
+            raise ParentImageMissing("Lost parent image(s) from Dockerfile: {}".format(missing))
+
+        # docker inspect all parent images so we can address them by Id
+        parent_image_ids = {}
+        for img, new_img in builder.parent_images.items():
+            inspection = builder.tasker.inspect_image(new_img)
+            try:
+                parent_image_ids[img] = inspection['Id']
+            except KeyError:  # unexpected code bugs or maybe docker weirdness
+                self.log.error(
+                    "Id for image %s is missing in inspection: '%s'",
+                    new_img, inspection)
+                raise NoIdInspection("Could not inspect Id for image " + new_img)
+
+        # update the parents in Dockerfile
+        new_parents = []
+        for parent in dfp.parent_images:
+            pid = parent_image_ids[parent]
+            self.log.info("changed FROM: '%s' -> '%s'", parent, pid)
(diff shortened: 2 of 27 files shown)


atomic_reactor/build.py
```diff
@@ -166,13 +168,20 @@ def df_path(self):
...
+        self.set_base_image(base)
         logger.debug("base image specified in dockerfile = '%s'", self.base_image)
-        if not self.base_image.tag:
-            self.base_image.tag = 'latest'
+        self.parent_images.clear()
+        for image in dfp.parent_images:
+            self.parent_images[image] = None
 
     def set_base_image(self, base_image):
+        logger.info("setting base image to '%s'", base_image)
```
mlangsdorf: Is this really necessary? We're already pretty noisy.

atomic_reactor/build.py
```diff
@@ -166,13 +168,20 @@ def df_path(self):
...
+        base = dfp.baseimage
+        if base is None:
+            raise RuntimeError("no base image specified in Dockerfile")
+        self.set_base_image(base)
         logger.debug("base image specified in dockerfile = '%s'", self.base_image)
-        if not self.base_image.tag:
-            self.base_image.tag = 'latest'
+        self.parent_images.clear()
+        for image in dfp.parent_images:
+            self.parent_images[image] = None
```
mlangsdorf: when is self.parent_images ever referenced? you set it here, but AFAIK never use it again.

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -65,92 +87,78 @@ def resolve_base_image(self, build_json):
...
-            self.log.info("'%s' not found", base_image_with_registry.to_str())
-            base_image_with_registry.namespace = 'library'
-            self.log.info("trying '%s'", base_image_with_registry.to_str())
+        return image_with_registry
 
+    def _pull_and_tag_image(self, image, build_json):
+        """Docker pull the image and tag it uniquely for use by this build"""
+        image = image.copy()
+        first_library_exc = None
+        for _ in range(20):
```
mlangsdorf: Please add documentation for the logic here.

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -65,92 +87,78 @@ def resolve_base_image(self, build_json):
...
-        # Attempt to tag it using a unique ID. We might have to retry
-        # if another build with the same parent image is finishing up
-        # and removing images it pulled.
-
-        # Use the OpenShift build name as the unique ID
-        unique_id = build_json['metadata']['name']
-        original_base_image = base_image.copy()
-        base_image = ImageName(repo=unique_id)
-
-        for _ in range(20):
```
mlangsdorf: any reason why we try this 20 times, as opposed to 10 or 5 or 100? I get that you're inheriting this code, mostly curious.

tests/plugins/test_change_from_in_df.py
```diff
@@ -14,80 +14,171 @@
...
     dfp = df_parser(str(tmpdir))
     dfp.content = df_content
 
-    workflow = mock_workflow(dfp.dockerfile_path, dict(no_id="here"))
-    result = runner(docker_tasker, workflow).run()
+    workflow = mock_workflow(dfp.dockerfile_path)
+    workflow.builder.tasker.inspect_image = lambda img: dict(no_id="here")
+
+    with pytest.raises(NoIdInspection):
+        ChangeFromPlugin(docker_tasker(), workflow).run()
```
mlangsdorf: run_plugin could take a second optional argument of "should_raise" that defaults to None and do pytest.raises(should_raise). I suggest that because after nicely encapsulating ChangeFromPlugin in your run_plugin() function, it's jarring to see it again being called directly in the tests cases.

atomic_reactor/build.py
sosiouxme: it's filled in later in pull_base_image to record what local tag or digest is used for those images -- used to update the dockerfile FROMs and recorded in the build metadata

atomic_reactor/build.py
sosiouxme: Not really, just was helpful in some of my later debugging of integration tests. To me, verbosity is good as logs are about all we have for debugging builds, and nobody looks at them unless something goes wrong. But I can take this out if you like :)

atomic_reactor/plugins/pre_pull_base_image.py
sosiouxme: I don't know, of course. You only go through the loop if the image pulls successfully but gets immediately deleted before being tagged. That ought to be an edge case race condition, so 20 seems excessive, but on the other hand there could be a *lot* of builds running with the same parents, and retries are cheap.

tests/plugins/test_change_from_in_df.py
sosiouxme: There are a number of situations where it makes more sense to go straight to the class to test a particular behavior, and there's a very specific reason for it here.

I like to raise named exceptions so I can test for exactly the thing that is expected to go wrong without relying on the text. But if you import the exception class from the plugin, and then run the plugin via a plugin-runner, the comparison of the exception you got to the expected class will fail. This is because the plugin runner reloads all the modules before running the plugin and what you get back is technically a different class, even though it came from the same code. So the level of indirection really gets in the way here.

atomic_reactor/plugins/pre_pull_base_image.py
sosiouxme: I'm all for docs. Are you talking about comments in the code, or docs in the repo, or on readthedocs....?

And "the logic here" being the behavior of applying a unique tag to the parent image for the duration of the build?

mlangsdorf: inline docs of whatever is going on in this for loop, please.

atomic_reactor/build.py
mlangsdorf: I'm not fond of it but I accept your justification.

atomic_reactor/build.py
mlangsdorf: Okay. You might want to add a note about that in the commit message.

mlangsdorf:
Minor nitpicks and questions but generally looks good. I would like to see the changes to test_change_from_in_df squashed to the original creation patch but no other major changes requested.

atomic_reactor/build.py
sosiouxme: Will do!

atomic_reactor/plugins/pre_pull_base_image.py
mlangsdorf: Yes, thanks.

sosiouxme:
poke @rcerven @twaugh @lcarva

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -65,15 +65,22 @@ def run(self):
         Pull parent images and retag them uniquely for this build.
         """
         build_json = get_build_json()
-        base_image = self._resolve_base_image(build_json)
-        base_image_with_registry = self._ensure_image_registry(base_image)
+        base_image_str = str(self.workflow.builder.base_image)
+        for nonce, parent in enumerate(sorted(self.workflow.builder.parent_images.keys())):
```
twaugh: Might `nonce` contain ":" here? If so, won't using that value as an image tag fail?

atomic_reactor/plugins/pre_change_from_in_df.py
```diff
@@ -29,21 +39,54 @@ def __init__(self, tasker, workflow):
...
-                sys.stdout.write(line)
+        builder = self.workflow.builder
+        dfp = DockerfileParser(builder.df_path)
+
+        df_base = dfp.baseimage
+        build_base = builder.base_image.to_str()
+        if build_base != df_base and build_base != builder.parent_images[df_base]:
+            # something updated the build's base_image without updating the corresponding
+            # parent_images entry; that should be OK, let base_image override.
+            builder.parent_images[df_base] = build_base
```
twaugh: Should `builder.base_image` be a property which handles updating `self.parent_images`?

sosiouxme: I debated that a few times and ultimately decided not to. However I'm reconsidering now because it rested on not having ready access everywhere to know which key in `parent_images` was from the Dockerfile baseimage. Since later I made `set_base_image` also set `original_base_image` with that info, that now seems like a lame excuse for not keeping the two in sync automatically.

atomic_reactor/util.py
```diff
@@ -1183,6 +1186,25 @@ def get_primary_images(workflow):
     return primary_images
 
 
+def get_parent_image_koji_data(workflow):
+    """Transform koji_parent plugin results into metadata dict."""
+    koji_parent = workflow.prebuild_results.get(PLUGIN_KOJI_PARENT_KEY) or {}
+
+    image_metadata = {}
+    image_metadata[PARENT_IMAGE_BUILDS_KEY] = koji_parent.get(PARENT_IMAGES_KOJI_BUILDS) or {}
```
lcarva: Let's be sure to document this additional piece of metadata.

tests/plugins/test_change_from_in_df.py
```diff
@@ -14,80 +14,173 @@
...
-    if MOCK:
-        mock_docker()
 
+def mock_workflow(df_path, base_image="mock:default"):
+    """
+    Provide just enough structure that workflow can be used to run a plugin.
+    Defaults below are solely to enable that purpose; tests where those values
+    matter should provide their own.
+    """
     builder = flexmock(
```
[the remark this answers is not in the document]
sosiouxme: Probably - I missed that stubs had been committed.

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -65,92 +87,86 @@ def resolve_base_image(self, build_json):
...
+            # if registry specified in Dockerfile image, ensure it's the one allowed by config
+            if image.registry and image.registry != self.parent_registry:
+                error = (
+                    "Registry specified in dockerfile image doesn't match configured one. "
+                    "Dockerfile: '%s'; expected registry: '%s'"
+                    % (image, self.parent_registry))
+                self.log.error("%s", error)
+                raise RuntimeError(error)
+
+            self.log.info(
```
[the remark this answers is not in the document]
lcarva: Yeah, probably a good idea.

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -83,6 +90,7 @@ def _resolve_base_image(self, build_json):
         except (TypeError, KeyError, IndexError):
             base_image = self.workflow.builder.base_image
         else:
+            self.log.info("rebuild triggered by image %s; using that as base image.", image_id)
```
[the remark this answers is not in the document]
sosiouxme: I see, I think... I was just trying to log some kind of justification as to why we're substituting this image... trying to think of something clear and correct. Is it maybe better to say nothing? The base image change will be logged regardless.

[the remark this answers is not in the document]
lcarva: I think it's worth mentioning that there was something in triggerBy for debugging purposes.

sosiouxme:
Addressed comments as extra commits - will squash if no one raises issues with them and tests pass.

sosiouxme:
Rebased, squashed, integration tests pass (modulo test issues). Should be ready for review again.

sosiouxme:
poke @twaugh @lcarva @rcerven for final review...

twaugh:
Please update docs/koji.md to reflect the new keys in 'extra'.

sosiouxme:
Done!

sosiouxme:
test.sh installs dockerfile-parse from source. travis sets up test environment and tries to install it from pypi, where 0.0.11 isn't published yet.

For now, I backed up the version in requirements.txt to 0.0.10 in the hopes of not breaking travis.

sosiouxme:
https://travis-ci.org/projectatomic/atomic-reactor/jobs/401014084 reveals travis is also checking the spec file and fails because dockerfile-parse 0.0.11 hasn't been published as a Fedora RPM.

At this point these reqs are likely wildly inaccurate anyway. I'm going to file an issue to publish dockerfile-parse 0.0.11 and then update the atomic-reactor reqs.

sosiouxme:
updated release notes with:

- correctly manage multi-stage build inputs and metadata

PR #1026, opened 2018-06-21
Port from modulemd to libmodulemd

owtaylor:
The Python modulemd library has been deprecated. The replacement is
accessing the libmodulemd file through gobject-introspection bindings.
    
Porting to libmodulemd enables support for version 2 of the modulemd
specification, which is being moved to for all modules.

[ Includes two small cleanups as separate commits ]


lcarva:
Updated release notes draft:
> - Port from modulemd to libmodulemd

PR #1027, opened 2018-06-22
if base image is specified with sha, first try if it is sha for manif…

rcerven:
…est list, and if not, try to get config from v2


twaugh:
Looks great. Nit: could you change the commit message so the first line is a short summary of the change, e.g. "Fix pull_base_image to handle manifest list digests". Then the rest of the commit message can explain how the fix works (i.e. what you currently have).

(Also needs tests of course.)

rcerven:
commit message updated, tests added

rcerven:
rn updated: when pull_base_image plugin is searching for manifest list, and parent image is specified by sha, check first if that is manifest list

PR #1028, opened 2018-06-25
add_filesystem: better error messages

mlangsdorf:
when a task fails, report the koji task_id as a full URL.

### DIFF
--- a/atomic_reactor/plugins/pre_add_filesystem.py
+++ b/atomic_reactor/plugins/pre_add_filesystem.py
@@ -30,7 +30,7 @@ from atomic_reactor.constants import (DEFAULT_DOWNLOAD_BLOCK_SIZE, PLUGIN_ADD_FI
                                       PLUGIN_CHECK_AND_SET_PLATFORMS_KEY)
 from atomic_reactor.plugin import PreBuildPlugin, BuildCanceledException
 from atomic_reactor.plugins.exit_remove_built_image import defer_removal
-from atomic_reactor.plugins.pre_reactor_config import get_koji_session
+from atomic_reactor.plugins.pre_reactor_config import get_koji_session, get_koji
 from atomic_reactor.koji_util import TaskWatcher, stream_task_output
 from atomic_reactor.util import get_retrying_requests_session
 from atomic_reactor import util
@@ -125,6 +125,8 @@ class AddFilesystemPlugin(PreBuildPlugin):
         self.is_orchestrator = True if self.architectures else False
         self.architecture = architecture
         self.scratch = util.is_scratch_build()
+        koji = get_koji(self.workflow, self.koji_fallback) or {}
+        self.koji_url = koji.get('hub_url')
 
     def get_arches(self, fallback):
         architectures = self.workflow.prebuild_results.get(PLUGIN_CHECK_AND_SET_PLATFORMS_KEY)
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -319,8 +321,9 @@ def run_image_task(self, image_build_conf):
                 task_result = self.session.getTaskResult(task_id)
             except Exception as exc:
                 task_result = repr(exc)
-            raise RuntimeError('image task, {}, failed: {}'
-                               .format(task_id, task_result))
+            raise RuntimeError('image task failed. see {}/taskinfo?taskID={}    \n'
+                               'underlying task failure: {}'
+                               .format(self.koji_url, task_id, task_result))
```
lcarva: I don't think koji_url (which is hub url) maps directly to brew web link. Can you verify that the link generated is valid?

mlangsdorf: [mlangsdorf@redhatnow rh-osbs]$ git grep koji_hub osbs-prod.conf
osbs-prod.conf:12:koji_hub = https://brewhub.stage.engineering.redhat.com/brewhub
osbs-prod.conf:37:koji_hub = http://brewhub.engineering.redhat.com/brewhub
osbs-prod.conf:58:koji_hub = https://brewhub.stage.engineering.redhat.com/brewhub
osbs-prod.conf:80:koji_hub = https://brewhub.stage.engineering.redhat.com/brewhub
osbs-prod.conf:104:koji_hub = http://brewhub.engineering.redhat.com/brewhub
osbs-prod.conf:132:# koji_hub = https://brewhub.stage.engineering.redhat.com/brewhub
osbs-prod.conf:164:# koji_hub = https://brewhub.stage.engineering.redhat.com/brewhub
osbs-prod.conf:198:# koji_hub = http://brewhub.engineering.redhat.com/brewhub
osbs-prod.conf:232:# koji_hub = http://brewhub.engineering.redhat.com/brewhub

it really looks to me like koji_hub is the brew web link.

lcarva: If I click on any of those links, I get:
```
Method Not Allowed
This is an XML-RPC server. Only POST requests are accepted.
```

If I append `/taskinfo?taskID=1` to it, I get the same error. The correct prefix is `https://brewweb.engineering.redhat.com/brew`

I'm not sure if we have the information in reactor-config-map to build the correct link.
We could either 1) not make this a link, or 2) enhance reactor-config-map to provide Koji web url which might be useful for other cases as well (email notification).

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -319,8 +321,9 @@ def run_image_task(self, image_build_conf):
                 task_result = self.session.getTaskResult(task_id)
             except Exception as exc:
                 task_result = repr(exc)
-            raise RuntimeError('image task, {}, failed: {}'
-                               .format(task_id, task_result))
+            raise RuntimeError('image task failed. see {}/taskinfo?taskID={}    \n'
```
lcarva: Consider using triple quotes to build a multi-line string:
```python
from textwrap import dedent

raise RuntimeError(dedent("""\
    image task failed
    see ....
    underlying task failure: ...""".format(...)
))
```

`dedent` will remove the leading spaces added. This makes the source code look properly indented while not affecting the string created.

atomic_reactor/plugins/pre_add_filesystem.py
mlangsdorf: Okay.
I thought the original request was to make it a link - the original error returned the taskID number but it wasn't obvious to the end user that it was a taskID.
Adding the necessary scaffolding for the koji web URL will be a pain. I'm going to ask in the original obs if just getting the taskID made obvious would be enough.

lcarva:
@mlangsdorf, should we close this PR until we have time to actively work on it again?

mlangsdorf:
Still waiting on feedback from the original requester.

PR #1029, opened 2018-06-26
Make v1 image tagging atomic (OSBS-5820)

twaugh:
This change introduces a new context manager, LockedPulpRepository, which locks operations on a Pulp repository by creating another repository, similar to the way a lock file works.

### DIFF
--- a/atomic_reactor/pulp_util.py
+++ b/atomic_reactor/pulp_util.py
@@ -210,7 +210,8 @@ class PulpHandler(object):
         self.p.copy_filters(repo_id, filters=pulp_filter, v1=True, v2=False)
 
     def update_repo(self, repo_id, tag):
-        self.p.updateRepo(repo_id, tag)
+        with LockedPulpRepository(self.p, repo_id):
+            self.p.updateRepo(repo_id, tag)
 
     def remove_image(self, repo_id, image):
         self.p.remove(repo_id, image)
--- a/atomic_reactor/pulp_util.py
+++ b/atomic_reactor/pulp_util.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2017 Red Hat, Inc
+Copyright (c) 2017, 2018 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
@@ -8,17 +8,20 @@ of the BSD license. See the LICENSE file for details.
 
 from __future__ import print_function, unicode_literals
 
+import logging
 import os
 import re
+import time
 import warnings
 from collections import namedtuple
 
 try:
     import dockpulp
-
 except (ImportError, SyntaxError):
     dockpulp = None
-    import logging
+
+from atomic_reactor.constants import (LOCKEDPULPREPOSITORY_RETRIES,
+                                      LOCKEDPULPREPOSITORY_BACKOFF)
 
 
 PulpRepo = namedtuple('PulpRepo', ['registry_id', 'tags'])
@@ -27,6 +30,8 @@ PulpRepo = namedtuple('PulpRepo', ['registry_id', 'tags'])
 # which may result in tenths of messages: very annoying
 # with "module", it just prints one warning -- this should balance security and UX
 warnings.filterwarnings("module")
+
+logger = logging.getLogger(__name__)
 
 
 class PulpLogWrapper(object):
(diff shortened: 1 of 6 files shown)


atomic_reactor/plugins/post_pulp_tag.py
```diff
@@ -53,12 +54,13 @@ def __init__(self, tasker, workflow, pulp_registry_name=None, pulp_secret_path=N
     def set_v1_tags(self, v1_image_id):
         image_names = self.workflow.tag_conf.images[:]
         handler = get_pulp_session(self.workflow, self.log, self.pulp_fallback)
-
-        pulp_repos = handler.create_dockpulp_and_repos(image_names)
+        handler.create_dockpulp()
+        pulp_repos = handler.create_repos_mapping(image_names)
```
mlangsdorf: If you don't want to run _ensure_repos, I'd rather pass an optional argument to create_dockpulp_and_repos called "ensure" that defaults to True, and runs  _ensure_repos if it is True. Just seems tidier than breaking it up like this.

atomic_reactor/pulp_util.py
```diff
@@ -224,3 +245,61 @@ def get_registry_hostname(self):
 
     def get_pulp_instance(self):
         return self.pulp_instance
+
+
+class LockedPulpRepository(object):
```
mlangsdorf: I think it makes more sense to pack a Lock argument to PulpHandler.update_repo that tells it to perform the locking, instead of requiring users of PulpHandler to do it themselves.

twaugh:
Yes, we're not going to use Pulp at some point in the future. In the mean time, we have to fix bugs in our integration with it. :(

atomic_reactor/pulp_util.py
twaugh: Fair point, will take a look at doing it this way.

atomic_reactor/pulp_util.py
```diff
@@ -224,3 +227,55 @@ def get_registry_hostname(self):
...
+                self.pulp.createRepo(self.repo_id, None,
+                                     registry_id=registry_id,
+                                     is_origin=True,  # avoid origin-lock-...
+                                     prefix_with=self.prefix)
+                break
+            except dockpulp.DockPulpError as exc:
+                if counter == self.retry_times:
+                    # This was the last chance; go ahead anyway. This
+                    # avoids a stale lock preventing builds.
+                    logging.info("got error, breaking lock: %s", exc)
```
twaugh: @lcarva, not sure if you spotted this, but I've made it more explicit that we will break the lock if we have to.

lcarva: I think that's fine. The update repo operation is pretty fast, unlike things like publish.

atomic_reactor/pulp_util.py
midnightercz: I don't want to meddle in your PRs, but consider implementing some watchdog mechanism to avoid stalled locked repos. Something like: Ah, this lock repo was created 10 hours ago, there's not way it's still relevant to any running process, so lets remove it and lock it again for new process.

atomic_reactor/pulp_util.py
```diff
@@ -224,3 +227,55 @@ def get_registry_hostname(self):
...
+        self.prefix = prefix
+
+        self.retry_times = 9
+        self.retry_delay = 5
+
+    def __enter__(self):
+        logger.info("creating lock repository %s%s", self.prefix, self.repo_id)
+        registry_id = 'lock/{}'.format(self.repo_id)
+        retry = 0
+        for counter in range(self.retry_times + 1):
```
mlangsdorf: This looked weird to me until I realize you're implementing a do while loop and need to loop one extra time for the hopefully successful initial try. Still, it's not an idiom we use consistently inside the code base so would you please add a comment explaining why the range is retry_times + 1?

atomic_reactor/pulp_util.py
```diff
@@ -224,3 +227,55 @@ def get_registry_hostname(self):
...
+        locked_repo.set_image_tags(v1_image_id, tags)
+    """
+
+    def __init__(self, pulp, repo_id, prefix='lock-'):
+        self.pulp = pulp
+        self.repo_id = repo_id
+        self.prefix = prefix
+
+        self.retry_times = 9
+        self.retry_delay = 5
```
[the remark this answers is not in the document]
twaugh: Yes, fixed.

atomic_reactor/pulp_util.py
twaugh: Added.

twaugh:
@lcarva, want to take another look to check it's still ok? I've updated tests and fixed a couple of small bugs in the code.

lcarva:
Updated release notes draft:
> - v1 image tagging operation is atomic

PR #1032, opened 2018-06-27
exit_sendmail: filter list of receivers for valid email addresses

mlangsdorf:
`filter the list of possible receivers for valid email addresses. Valid
in this is case is one or more alphanumerics, ., -, or _, followed by
an @ sign, followed by one or more tokens like the first, followed by
a ., followed by some alphanumerics. So m_-.@2_-.c is valid, but
me@example or not/me@example.com are not.`

This isn't strict RFC 2821 compliance, but matches the soft rules that
likely emails will follow.

### DIFF
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -6,6 +6,7 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
 
+import re
 from email.mime.text import MIMEText
 from email.mime.multipart import MIMEMultipart
 from email.mime.base import MIMEBase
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -30,6 +31,22 @@
...
 
+# an email address consisting of local name, an @ sign, and a domain name
+# the local name starts with a letter and then has one or more alphanumerics, _, +, and -
+# symbols in any order.
+# the domain name has one or more alphanumerics or - in any order
+# followed by a ., followed by one or more alphanumerics and - in any order.
+# RFC 2821 defines a valid email address much more expansively, but this definition matches
+# what most people loosely expect to be valid.
+# specifically, email addresses of the form id/name@domain.tla are not valid.
+VALID_EMAIL_REGEX = "^[a-zA-Z][a-zA-Z0-9-_.\+]+@[a-zA-Z0-9-\.]+\.[a-zA-Z0-9-]+$"
```
lcarva: Need to escape the `.` on the second range?

Consider using raw strings for defining regex: `r'....'`

mlangsdorf: Fixed.

lcarva:
Updated release notes draft:
> - `exit_sendmail`: filter list of receivers for valid email addresses

PR #1033, opened 2018-06-28
support for arch-override in scratch builds

mlangsdorf:
add support in pre_check_and_set_platforms to detect scratch and isolated builds. For those builds, if
the platform list from user_params is different than the list from koji_target, assume that the user_params list take priority.

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -40,7 +40,8 @@ from atomic_reactor.constants import (DOCKERFILE_FILENAME, REPO_CONTAINER_CONFIG
                                       HTTP_CLIENT_STATUS_RETRY, HTTP_REQUEST_TIMEOUT,
                                       MEDIA_TYPE_DOCKER_V2_SCHEMA1, MEDIA_TYPE_DOCKER_V2_SCHEMA2,
                                       MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST, MEDIA_TYPE_OCI_V1,
-                                      MEDIA_TYPE_OCI_V1_INDEX, GIT_MAX_RETRIES, GIT_BACKOFF_FACTOR)
+                                      MEDIA_TYPE_OCI_V1_INDEX, GIT_MAX_RETRIES, GIT_BACKOFF_FACTOR,
+                                      PLUGIN_BUILD_ORCHESTRATE_KEY)
 
 from dockerfile_parse import DockerfileParser
 from pkg_resources import resource_stream
@@ -597,6 +598,12 @@ def is_isolated_build():
     except KeyError:
         logger.error('metadata.labels not found in build json')
         raise
+
+
+def get_orchestrator_platforms(workflow):
+    for plugin in workflow.buildstep_plugins_conf or []:
+        if plugin['name'] == PLUGIN_BUILD_ORCHESTRATE_KEY:
+            return plugin['args']['platforms']
 
 
 # copypasted and slightly modified from
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -590,6 +590,15 @@ def is_scratch_build():
         raise
 
 
+def is_isolated_build():
+    build_json = get_build_json()
+    try:
+        return build_json['metadata']['labels'].get('isolated', False)
+    except KeyError:
+        logger.error('metadata.labels not found in build json')
+        raise
+
+
 # copypasted and slightly modified from
 # http://stackoverflow.com/questions/1094841/reusable-library-to-get-human-readable-version-of-file-size/1094933#1094933
 def human_size(num, suffix='B'):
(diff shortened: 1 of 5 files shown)


tests/plugins/test_check_and_set_platforms.py
```diff
@@ -129,3 +134,48 @@ def test_check_and_set_platforms(tmpdir, platforms, platform_exclude, platform_o
         assert plugin_result[PLUGIN_CHECK_AND_SET_PLATFORMS_KEY] == set(result)
     else:
         assert plugin_result[PLUGIN_CHECK_AND_SET_PLATFORMS_KEY] is None
+
+
+@pytest.mark.parametrize(('labels', 'platforms', 'orchestrator_platforms', 'result'), [
+    ({}, None, None, None),
+    ({}, 'x86_64 arm64', ['spam', 'bacon'], ['arm64', 'x86_64']),
+    ({'isolated': True}, 'spam bacon', ['x86_64', 'arm64'], ['arm64', 'x86_64']),
+    ({'isolated': True}, 'x86_64 arm64', None, ['arm64', 'x86_64']),
```
rcerven: I don't think this is valid use case, as orchestrator platforms will be always set with koji-c

atomic_reactor/util.py
```diff
@@ -590,6 +591,21 @@ def is_scratch_build():
...
+        return build_json['metadata']['labels'].get('isolated', False)
+    except KeyError:
+        logger.error('metadata.labels not found in build json')
+        raise
+
+
+def get_orchestrator_platforms(workflow):
+    for plugin in workflow.buildstep_plugins_conf or []:
+        if plugin['name'] == PLUGIN_BUILD_ORCHESTRATE_KEY:
+            return plugin['args']['platforms']
```
sosiouxme: I don't think it needs to derail this PR but... could this be in build metadata? It does not feel like a plugin arg is the place this should be coming from if it affects anything outside that plugin.

mlangsdorf: Quick and dirty fix, but yes, it's inelegant. One of the reasons I pushed it into its own function is so it would be encapsulated and could be changed to something sensible later.

owtaylor:
I don't think it can be right that with this patch:
 - If no koji_target is specified, then --arch-override will obey platforms: only and platforms: exclude from container.yaml
 - If a koji_target is specified (so check_and_set_platforms runs) then --arch-override overrides the settings from --container.yaml

I'm about to touch this code again to change the handling of platforms not in the cluster config, and unless people think otherwise, will change things so that the container.yaml settings uniformly win (and if they conflict, and you end up with an empty set, the build fails.)

lcarva:
> If no koji_target is specified, then --arch-override will obey platforms: only and platforms: exclude from container.yaml

`koji_target` is required if build was submitted via koji. `--arch-override` only exists in the context of koji. By the time control is handed to atomic-reactor, there's only a list of platforms, which may or may not have been set via koji's arch-override.

> If a koji_target is specified (so check_and_set_platforms runs) then --arch-override overrides the settings from --container.yaml

The comparison of arches in build tag and user_params.platforms is heuristics to determine if the list of platforms we got was submitted via arch-override.

In the case where arch-override was used, we want to ignore the platform filters in container.yaml. This allows users to try out a new arch without having to modify the koji build tag.

owtaylor:
> If no koji_target is specified, then --arch-override will obey platforms: only and platforms: exclude from container.yaml
> 
> koji_target is required if build was submitted via koji. --arch-override only exists in the context of koji. By the time control is handed to atomic-reactor, there's only a list of platforms, which may or may not have been set via koji's arch-override.

Well, yes, --arch-only is a koji-containerbuild thing but I was thinking of it as being equivalent to:

    osbs build --orchestrator ... --platforms <xx>

Which *is* limited by container.yaml.  (See OrchestrateBuildPlugin.get_plugins())  It's a bit odd to me if you can *only* override container.yaml when building in koji - but if it's useful, then that's more important...

Ignoring the platform filters in container.yaml doesn't allow you to try out a new arch without having to modify the koji build tag, it allows you to try out an arch you previously excluded without modifying the container.yaml ;-)

lcarva:
> Which is limited by container.yaml. (See OrchestrateBuildPlugin.get_plugins()) It's a bit odd to me if you can only override container.yaml when building in koji - but if it's useful, then that's more important...

We have a story to teach osbs-client and atomic-reactor about arch-override which would make all this more predictable and straight forward. The heuristics we added was simply a stop gap to make arch override work properly in the mean time.

Question, if you're not using koji, why would you want arch override? If you're submitting builds directly via osbs-client, you could just change `--platforms` value, no?

> Ignoring the platform filters in container.yaml doesn't allow you to try out a new arch without having to modify the koji build tag, it allows you to try out an arch you previously excluded without modifying the container.yaml ;-)

It depends. If you pass in a yum repo that has all the packages the koji build tag is not really used. So the build should be successful. :)

PR #1034, opened 2018-06-28
 Check output of rpmqa command and retry if needed

lcarva:


lcarva:
Updated release notes draft:
> - Output of rpmqa command is checked and retried if empty

PR #1036, opened 2018-07-03
sendmail: don't notify if no addresses are valid

mlangsdorf:
If all the addresses available for sendmail are invalid, log that but don't
send any emails.


lcarva:

PR #1037, opened 2018-07-05
Fix logger for LockedPulpRepository

lcarva:
Some of the log messages were using logging, instead of logger. This
doesn't cause a python error because logging is actually valid. But
there are no handlers attached to it so the messages go nowhere. By
using logger, the code uses the module level logger which contains the
correct handlers.

### DIFF
--- a/atomic_reactor/pulp_util.py
+++ b/atomic_reactor/pulp_util.py
@@ -272,9 +272,9 @@ class LockedPulpRepository(object):
                 if counter == self.retry_times:
                     # This was the last chance; go ahead anyway. This
                     # avoids a stale lock preventing builds.
-                    logging.info("got error, breaking lock: %s", exc)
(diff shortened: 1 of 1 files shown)


atomic_reactor/pulp_util.py
```diff
@@ -285,5 +285,5 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         try:
             self.pulp.deleteRepo(repo_id)
         except dockpulp.errors.DockPulpError as exc:
-            logging.info("ignoring error from lock repository deletion: %s",
+            logger.info("ignoring error from lock repository deletion: %s",
                          exc)
```
sosiouxme: nit: now it doesn't line up.

why not join with the previous line though...

lcarva: Good catch. Fixed it.

lcarva:

PR #1038, opened 2018-07-06
input_osv3: fix for user_params schema

mlangsdorf:
worker build pass the entire json of the new reactor config in the
reactor_config_override, and the schema broke on that.

Rewrite the schema so that it expects reactor_config_override to be
a string if the buildtype is orchestrator, and an object if the
buildtype is worker.

### DIFF
--- a/atomic_reactor/schemas/user_params.json
+++ b/atomic_reactor/schemas/user_params.json
@@ -38,7 +38,7 @@
       "items": {"type": "string"}
     },
     "reactor_config_map": {"type": "string"},
-    "reactor_config_override": {"type": "string"},
+    "reactor_config_override": {"type": "object"},
     "release": {"type": "string"},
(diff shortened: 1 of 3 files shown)


atomic_reactor/schemas/user_params.json
```diff
@@ -3,54 +3,83 @@
...
-    "yum_repourls": {
-      "type": "array",
-      "items": {"type": "string"}
+    "orchestrator": {
+      "allOf": [
+        {"$ref": "#/definitions/shared"},
+        {
+          "properties": {
+            "build_type": {"enum": ["orchestrator"]},
+            "reactor_config_override": {"type": "string"}
```
lcarva: Why is this a string?

mlangsdorf: what should it be?

lcarva:

PR #1039, opened 2018-07-06
Prevent repo name collision

lcarva:
When adding a yum repository by URL, save it to a unique filename.
Otherwise, the following two repo urls end up on the same file:
- http://yum.example.com/spam/compose.repo
- http://yum.example.com/bacon/compose.repo

NOTE: This PR includes a commit from https://github.com/projectatomic/atomic-reactor/pull/1038 because master is currently broken.


lcarva:
> lgtm once you figure out the travis build failures.

Fixed up now.

lcarva:
Updated release notes draft:
> - When using yum repo urls, repo files are saved with unique name to prevent repo name collision

PR #1041, opened 2018-07-09
user_params schema: fix signing_intent and filesystem_koji_task_id

mlangsdorf:
filesystem_koji_task_id is an integer, and signing intent is a boolean.

### DIFF
--- a/atomic_reactor/schemas/user_params.json
+++ b/atomic_reactor/schemas/user_params.json
@@ -18,7 +18,7 @@
       "items": {"type": "integer"}
     },
     "customize_conf": {"type": "string"},
-    "filesystem_koji_task_id": {"type": "string"},
+    "filesystem_koji_task_id": {"type": "integer"},
     "flatpak": {"type": "boolean"},
     "flatpak_base_image": {"type": "string"},
     "git_branch": {"type": "string"},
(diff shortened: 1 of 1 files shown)


atomic_reactor/schemas/user_params.json
```diff
@@ -41,10 +41,7 @@
     "reactor_config_override": {"type": "object"},
     "release": {"type": "string"},
     "scratch": {"type": "boolean"},
-    "signing_intent": {
-      "type": "array",
-      "items": {"type": "string"}
-    },
+    "signing_intent": {"type": "boolean"},
```
lcarva: Isn't this a string?

mlangsdorf: :param signing_intent: bool, True to sign the resulting image

per the docs for osbs-client/build/buildrequestv2, it's a boolean.

lcarva: Ok, that is also wrong. The value is used by PluginsConfiguration to set the `signing_intent` parameter of the `resolve_composes` plugin of atomic reactor. And that value is then used as a string.

mlangsdorf: Sure, fixed.

PR #1042, opened 2018-07-10
if x86_64 platform isn't present in manifest list, pull schema2 diges…

rcerven:
…t for different arch, used only for orchestrator

### DIFF
--- a/atomic_reactor/plugins/pre_pull_base_image.py
+++ b/atomic_reactor/plugins/pre_pull_base_image.py
@@ -17,6 +17,7 @@ this build so that it isn't removed by other builds doing clean-up.
 from __future__ import unicode_literals
 
 import docker
+import platform
 
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.util import (get_build_json, get_manifest_list,
@@ -25,7 +26,8 @@ from atomic_reactor.util import (get_build_json, get_manifest_list,
 from atomic_reactor.constants import PLUGIN_CHECK_AND_SET_PLATFORMS_KEY
 from atomic_reactor.core import RetryGeneratorException
 from atomic_reactor.plugins.pre_reactor_config import (get_source_registry,
-                                                       get_platform_to_goarch_mapping)
+                                                       get_platform_to_goarch_mapping,
+                                                       get_goarch_to_platform_mapping)
 from requests.exceptions import HTTPError, RetryError, Timeout
 from osbs.utils import RegistryURI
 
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -168,6 +192,31 @@ def _pull_and_tag_image(self, image, build_json, nonce):
         self.log.error("giving up trying to pull image")
         raise RuntimeError("too many attempts to pull and tag image")
 
+    def _get_manifest_list(self, image):
```
lcarva: We can make this a bit more efficient by remembering the manifest list for each image:
```python
def _get_manifest_list(self, image, cache={}):
  if image not in cache:
    ...
    cache[image] = manifest_list
  return cache[image]
```

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -75,6 +78,27 @@ def run(self):
             if self.check_platforms:
                 self._validate_platforms_in_image(image)
 
+                manifest_list = self._get_manifest_list(image)
+
+                if manifest_list:
```
lcarva: Let's throw this logic in its own method.

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -75,6 +78,27 @@ def run(self):
...
+                    manifest_list_dict = manifest_list.json()
+                    arch_digests = {}
+                    build_image_digests = {}
+                    image_name = image.to_str(tag=False)
+
+                    for manifest in manifest_list_dict['manifests']:
+                        arch = manifest['platform']['architecture']
+                        arch_digests[arch] = image_name + '@' + manifest['digest']
```
lcarva: And break here to avoid looping over unnecessarily?

lcarva: And break here to avoid looping over unnecessarily?

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -75,12 +79,44 @@ def run(self):
...
+            for manifest in manifest_list_dict['manifests']:
+                arch = manifest['platform']['architecture']
+                arch_digests[arch] = image_name + '@' + manifest['digest']
+
+            present_platform = None
+            try:
+                arch_to_platform = get_goarch_to_platform_mapping(self.workflow)
+                for arch in arch_digests:
+                    build_image_digests[arch_to_platform[arch]] = arch_digests[arch]
+                    present_platform = arch_to_platform[arch]
```
twaugh: Could be simplified:
```
for arch, digest in arch_digests.items():
    present_platform = arch_to_platform[arch]
    build_image_digests[present_platform] = digest
```

atomic_reactor/plugins/pre_pull_base_image.py
twaugh: In future we could consider decorating with @functools.lru_cache instead of handling caching explicitly, since atomic-reactor does not need to run on Python < 2.7.

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: changed

PR #1043, opened 2018-07-13
add support for include_unpublished_pulp_repos

mlangsdorf:
add support in odcs_utils for passing flags to a compose
add support in pre_resolve_composes for retrieving the include_unpublished_pulp_repos when doing a pulp_compose and passing that flag, if set, to odcs_utils

### DIFF
--- a/atomic_reactor/odcs_util.py
+++ b/atomic_reactor/odcs_util.py
@@ -41,7 +41,8 @@ class ODCSClient(object):
 
         self.session = session
 
-    def start_compose(self, source_type, source, packages=None, sigkeys=None, arches=None):
+    def start_compose(self, source_type, source, packages=None, sigkeys=None, arches=None,
+                      flags=None):
         """Start a new ODCS compose
 
         :param source_type: str, the type of compose to request (tag, module, pulp)
@@ -71,6 +72,9 @@ class ODCSClient(object):
 
         if sigkeys is not None:
             body['source']['sigkeys'] = sigkeys
+
+        if flags is not None:
+            body['flags'] = flags
 
         if arches is not None:
             body['arches'] = arches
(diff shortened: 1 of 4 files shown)


tests/plugins/test_resolve_composes.py
```diff
@@ -289,6 +291,11 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
 
         self.run_plugin_with_args(workflow, reactor_config_map=reactor_config_map)
 
+    @pytest.mark.parametrize(('flags'), [
```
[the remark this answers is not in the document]

tests/plugins/test_resolve_composes.py
```diff
@@ -298,12 +300,20 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
         (['x86_64', 'ppce64le', 'arm64'], ['x86_64', 'ppce64le', 'arm64'], 'unsigned', 'unsigned'),
         (['x86_64', 'ppce64le', 'arm64'], None, 'beta', 'beta'),
     ))
+    @pytest.mark.parametrize(('flags'), [
```
twaugh: Can we change this to 'flags,expected_flags', with e.g. `({UNPUBLISHED_REPOS: True}, [UNPUBLISHED_REPOS])` as an example test case? Then we can remove the logic in the test duplicating the logic in the code under test.

atomic_reactor/odcs_util.py
```diff
@@ -69,6 +70,9 @@ def start_compose(self, source_type, source, packages=None, sigkeys=None, arches
         if packages:
             body['source']['packages'] = packages
 
+        if flags is not None:
+            body['source']['flags'] = flags
```
lcarva: Looking at odcs source code, `'flags`' behave like `'arches'` and should be directly in `'body'`, not nested under `'source'`: https://pagure.io/odcs/blob/master/f/client/odcs/client/odcs.py#_246-257

mlangsdorf: The ODCS docs do not make that very clear:
https://pagure.io/odcs
will fix.

mlangsdorf: Fixed.

lcarva:
Updated release notes draft:
> - add support for include_unpublished_pulp_repos ODCS flag

PR #1044, opened 2018-07-17
imagebuilder: simplify reading output, decode utf-8

sosiouxme:
Previous attempts to readline() separate stderr and stdout streams from
the command were making things more confusing than necessary, and it
turns out there's basically no sane way to do this in a non-blocking
way, so now the output is all read as a single blocking stream.

The decode() statement was defaulting to 'ascii' decoding which didn't
work for anything that outputs utf-8. So now it decodes to utf-8. I
tried to exercise this in unit tests but couldn't figure out how, with
python2 apparently attempting ascii decoding in unexpected places.

(It seems like it should just work if I add unicode in the 3rd test but I tried with several variations and BytesIO -- `bytes(cmd_output + cmd_error)` gives me `UnicodeEncodeError: 'ascii' codec` even though it doesn't with cmdline interpreter.)

### DIFF
--- a/atomic_reactor/plugins/build_imagebuilder.py
+++ b/atomic_reactor/plugins/build_imagebuilder.py
@@ -8,7 +8,6 @@ of the BSD license. See the LICENSE file for details.
 from __future__ import print_function, unicode_literals
 
 import subprocess
-import time
 from six import PY2
 import os
 
@@ -17,10 +16,6 @@ from atomic_reactor.plugin import BuildStepPlugin
 from atomic_reactor.build import BuildResult
 from atomic_reactor.constants import CONTAINER_IMAGEBUILDER_BUILD_METHOD
 from atomic_reactor.constants import EXPORTED_SQUASHED_IMAGE_NAME, IMAGE_TYPE_DOCKER_ARCHIVE
-
-
-def sixdecode(data):
-    return data.decode() if PY2 else data
 
 
 class ImagebuilderPlugin(BuildStepPlugin):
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/build_imagebuilder.py
```diff
@@ -43,36 +38,28 @@ def run(self):
...
         self.log.debug('imagebuilder build has begun; waiting for it to finish')
-        (output, last_error) = ([], None)
+        output = []
         while True:
             poll = ib_process.poll()
-            # NOTE: imagebuilder writes both stdout and stderr in normal operation.
-            # Because the two streams are not always logged in the same order as they're
-            # produced, prefix logs with stderr/stdout to distinguish the streams.
-            out = sixdecode(ib_process.stdout.readline())
+            out = ib_process.stdout.readline()
```
lcarva: Wouldn't this hang if there's no output?

sosiouxme: it's supposed to block until there is output. and if the process ends / stream closes, return ''

it works, most of the time.

lcarva: Nope, it just returns an empty string:
```python
>>> import subprocess
>>> kwargs = dict(stdout=subprocess.PIPE, stderr=subprocess.STDOUT, universal_newlines=True)
>>> p = subprocess.Popen(['echo', 'yolo'], **kwargs)
>>> p.poll()
0
>>> p.stdout.readline()
'yolo\n'
>>> p.stdout.readline()
''
>>> p.stdout.readline()
''

```

lcarva:
Updated release notes draft:
> - simplify reading output for imagebuilder and force decode it as utf-8

PR #1046, opened 2018-07-19
Clean up platform handling and restrict platforms by enabled clusters

owtaylor:
The main point of this pull request is to drop (with a warning) platforms that aren't in the enabled set of clusters. This is useful because the koji_target is used for two things:

 * The package set as implemented by the 'koji' plugin
 * The list of architectures

If the koji_target for the package set you want to use has architectures that aren't enabled for OSBS, this prevents that from causing failures.

To implement this robustly, the PR consolidates handling of platforms to check_and_set_platforms, and assumes that it *will* be run even in the (unusual) case of no koji_target. I'll follow up with a PR to osbs-client to implement that.

A number of test cases are removed because they were testing the same things going through different code paths and don't make sense or don't work now that the handling is consolidated.

### DIFF
--- a/atomic_reactor/plugins/pre_check_and_set_platforms.py
+++ b/atomic_reactor/plugins/pre_check_and_set_platforms.py
@@ -56,7 +56,7 @@ class CheckAndSetPlatformsPlugin(PreBuildPlugin):
 
         if is_scratch_build() or is_isolated_build():
             override_platforms = get_orchestrator_platforms(self.workflow)
-            if override_platforms and set(override_platforms) != koji_platforms:
+            if override_platforms and set(override_platforms) != set(platforms):
                 # platforms from user params do not match platforms from koji target
                 # that almost certainly means they were overridden and should be used
                 return set(override_platforms)
--- a/atomic_reactor/plugins/pre_check_and_set_platforms.py
+++ b/atomic_reactor/plugins/pre_check_and_set_platforms.py
@@ -17,7 +17,7 @@ when koji build tags change.
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.util import (get_platforms_in_limits, is_scratch_build, is_isolated_build,
                                  get_orchestrator_platforms)
-from atomic_reactor.plugins.pre_reactor_config import get_koji_session, NO_FALLBACK
+from atomic_reactor.plugins.pre_reactor_config import get_config, get_koji_session, NO_FALLBACK
 from atomic_reactor.constants import PLUGIN_CHECK_AND_SET_PLATFORMS_KEY
 
 
@@ -37,6 +37,7 @@ class CheckAndSetPlatformsPlugin(PreBuildPlugin):
         # call parent constructor
         super(CheckAndSetPlatformsPlugin, self).__init__(tasker, workflow)
         self.koji_target = koji_target
+        self.reactor_config = get_config(self.workflow)
 
     def run(self):
         """
@@ -64,4 +65,16 @@ class CheckAndSetPlatformsPlugin(PreBuildPlugin):
         else:
             platforms = get_orchestrator_platforms(self.workflow)
 
-        return get_platforms_in_limits(self.workflow, platforms)
+        if not platforms:
+            raise RuntimeError("Cannot determine platforms; no koji target or platform list")
+
+        # Filter platforms based on clusters
+        enabled_platforms = []
+        for p in platforms:
+            if self.reactor_config.get_enabled_clusters_for_platform(p):
+                enabled_platforms.append(p)
+            else:
+                self.log.warning(
+                    "No cluster found for platform '%s' in reactor config map, skipping", p)
+
+        return get_platforms_in_limits(self.workflow, enabled_platforms)
--- a/atomic_reactor/plugins/pre_check_and_set_platforms.py
+++ b/atomic_reactor/plugins/pre_check_and_set_platforms.py
@@ -25,7 +25,7 @@ class CheckAndSetPlatformsPlugin(PreBuildPlugin):
     key = PLUGIN_CHECK_AND_SET_PLATFORMS_KEY
     is_allowed_to_fail = False
 
-    def __init__(self, tasker, workflow, koji_target):
+    def __init__(self, tasker, workflow, koji_target=None):
 
         """
         constructor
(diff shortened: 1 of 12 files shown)


atomic_reactor/util.py
```diff
@@ -615,9 +615,8 @@ def get_platforms(workflow):
     if koji_platforms:
         return koji_platforms
 
-    # if check_and_set_platforms didn't run, or didn't get any platforms from koji
-    # determine platforms from USER_PARAMS platforms parameter
-    return get_platforms_in_limits(self.workflow, get_orchestrator_platforms())
+    # Not an orchestrated build
+    return None
```
lcarva: What if we return a one-item list with the currently running platform instead?

atomic_reactor/plugins/pre_check_and_set_platforms.py
```diff
@@ -64,4 +65,17 @@ def run(self):
         else:
             platforms = get_orchestrator_platforms(self.workflow)
 
-        return get_platforms_in_limits(self.workflow, platforms)
+        if not platforms:
+            # Not an orchestrated build
```
lcarva: If koji target is used, this won't be None. I think this comment is misleading. We probably need different heuristics for determining whether or not this is an orchestrator build. `self.reactor_config.is_default()` could be used to determine if there is a reactor config at all which seems like a safer check.

tests/plugins/test_check_and_set_platforms.py
```diff
@@ -136,20 +136,28 @@ def test_check_and_set_platforms(tmpdir, platforms, platform_exclude, platform_o
...
+                          'result'), [
+    ({}, None, None, '', None),
+    ({}, 'x86_64 arm64', ['spam', 'bacon'], '', ['arm64', 'x86_64']),
+    ({'isolated': True}, 'spam bacon', ['x86_64', 'arm64'], '', ['arm64', 'x86_64']),
+    ({'isolated': True}, 'x86_64 arm64', None, '', ['arm64', 'x86_64']),
+    ({'isolated': True}, None, ['x86_64', 'arm64'], '', None),
+    ({'scratch': True}, 'spam bacon', ['x86_64', 'arm64'], '', ['arm64', 'x86_64']),
+    ({'scratch': True}, 'x86_64 arm64', None, '', ['arm64', 'x86_64']),
+    ({'scratch': True}, None, ['x86_64', 'arm64'], '', None),
+    ({'scratch': True}, 'x86_64 arm64', ['x86_64', 'arm64'], 'x86_64', ['x86_64']),
```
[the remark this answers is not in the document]
owtaylor: Sure, that makes sense for clarifying and testing all the logic.

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -161,9 +151,12 @@ def read_configs(self):
             with open(file_path) as f:
                 pulp_data = yaml.safe_load(f) or {}
 
-        arches = self.get_arches()
+        platforms = get_platforms(self.workflow)
+        if platforms:
+            platforms = sorted(platforms)
```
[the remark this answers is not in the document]
owtaylor: Yeah:

    (flexmock(ODCSClient)
        .should_receive('start_compose'
        .with_args(source_type='tag', source=KOJI_TAG_NAME, arches=sorted(arches),
                   packages=['spam', 'bacon', 'eggs'], sigkeys=sig_keys)
        .and_return(tag_compose).once())

used to end up with arches set directly from the plugin params for the test case, but now there's an intermediate list => set => list conversion. I don't know how to make flexmock say "this list parameter but in any order", so the alternative would have been to make ODCSClient.start_compose() take a set. I'll add a "# sorted to keep predictable for tests"

tests/plugins/test_add_filesystem.py
```diff
@@ -529,9 +489,9 @@ def test_image_build_overwrites(tmpdir, architectures, architecture, reactor_con
                     """))
     plugin = create_plugin_instance(tmpdir, {
         'repos': repos,
-        'architectures': architectures,
         'architecture': architecture
-    }, reactor_config_map=reactor_config_map)
+    }, reactor_config_map=reactor_config_map,
+                                    architectures=architectures)
```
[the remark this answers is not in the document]
owtaylor: I'm not sure how I'd indent it differently given that line breaking. I'll just change it to:

    plugin = create_plugin_instance(tmpdir, {
        'repos': repos,
        'architecture': architecture
    }, reactor_config_map=reactor_config_map, architectures=architectures)

atomic_reactor/plugins/pre_check_and_set_platforms.py
owtaylor: Well, yes, the comment is misleading because this plugin is only ever used for orchestrator builds (assuming that plugin configurations generated by osbs-client are the only valid ones!), so if this gets hit, it basically means that an orchestrator build was triggered without platforms and without a koji_target - which will cause a failed build. It probably should just raise a RuntimeError() here rather than waiting for OrchestratorBuildPlugin.run()

I propagated the comment after writing it in util.get_platforms() where a None return does imply 'not an orchestrat**or** build'. (So a prod or worker build.) I've looked through the plugins that use get_platforms(), and they seem to have reasonable looking handling for a None return from get_platforms().

I've added the RuntimeError here and changed orchestrated to orchestrator in get_platforms() in the version I'm about to push.

owtaylor:
> Is there any issue with running this version with arrangement 5?

Should be the same situation as with arrangement 6 generated by an old version of osbs-client - if no koji_target is set, then the build will fail. Since as far as we know, koji_target is *always* set for real uses of OSBS, this isn't a big problem. I didn't adjust the arrangement 5 code in osbs-client, but that could be done.

atomic_reactor/util.py
```diff
@@ -608,6 +610,15 @@ def get_orchestrator_platforms(workflow):
             return plugin['args']['platforms']
 
 
+def get_platforms(workflow):
+    koji_platforms = workflow.prebuild_results.get(PLUGIN_CHECK_AND_SET_PLATFORMS_KEY)
+    if koji_platforms:
+        return koji_platforms
+
+    # Not an orchestrator build
+    return None
```
lcarva: If it's not an orchestrator build, how about we return a one item list with the current platform?

owtaylor: I'm inclined to say no - in that in an orchestrator build the *same* thing isn't being done with the list of platforms as is done for the single platform in a worker build. [That would argue that utils.get_orchestrator_platforms() should be a private utility function and this function should be called get_orchestrator_platforms()]

Examining the 5 call sites, 4 wouldn't care one way or the other - the 5th would need some code changes:

**Only in the orchestrated case**:  plugins/build_orchestrate_build.py, plugins/post_pulp_pull.py, plugins/pre_resolve_composes.py
**Does nothing if the platform list is None *or* has a single element**: plugins/pre_pull_base_image.py
**Requires changes**:  plugins/pre_add_filesystem.py

add_filesystem does: 

        self.architectures = get_platforms(self.workflow)
        self.is_orchestrator = True if self.architectures else False

lcarva: Ok, good point about `add_filesystem`.

lcarva:
Updated release notes draft:
> - platform filters applied to scratch/isolated builds
> - Restrict platforms by enabled clusters

PR #1047, opened 2018-07-20
Explicitly enable modules to install into a Flatpak

owtaylor:
Versions of DNF in Fedora 28 and later understand modular content, and won't install packages from modules unless the modules are enabled. So to support (and require) using a newer base_image,
explicitly 'dnf module enable' the modules from the compose we generated before installing packages. It's necessary to exclude the 'platform' pseudo-module from this, so update test data so we can test this exclusion.


lcarva:
Updated release notes draft:
> - Modules to install into a Flatpak are enabled explicitly

PR #1048, opened 2018-07-20
flatpak_create_oci: Fix docker compatibility issues

owtaylor:
This patch fixes two issues found when testing Flatpak support in a Fedora 28 environment:

 * Errors from create_container() when no command is provided
 * ApiClient.export() returning a generator instead of a file-like object

I can't point to the exact upstream change for either. The first should be safely compatible with old versions as well, the second may not be, but should be fine in the places where we're actually going to use this  code for Flatpak creation.

### DIFF
--- a/tests/plugins/test_flatpak_create_oci.py
+++ b/tests/plugins/test_flatpak_create_oci.py
@@ -723,7 +723,7 @@ def test_flatpak_create_oci(tmpdir, docker_tasker, config_name, breakage, mock_f
 
     (flexmock(docker_tasker.d.wrapped)
      .should_receive('create_container')
-     .with_args(workflow.image)
+     .with_args(workflow.image, command=["/bin/bash"])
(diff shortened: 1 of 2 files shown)


tests/plugins/test_flatpak_create_oci.py
```diff
@@ -721,14 +721,25 @@ def test_flatpak_create_oci(tmpdir, docker_tasker, config_name, breakage, mock_f
 
     export_stream = open(filesystem_tar, "rb")
 
+    def stream_to_generator(s):
+        while True:
+            # Yield small chunks to test the StreamAdapter code better
+            buf = s.read(100)
+            if len(buf) == 0:
+                return
+            else:
```
twaugh: Could remove the `else:` here since the `if:` clause returns.

lcarva:
Updated release notes draft:
> - `flatpak_create_oci` docker compatibility issues fixed

PR #1050, opened 2018-07-25
post_group_manifests: make the plugin work on no amd64 platforms

jcajka:
I have been hitting this error path on ppc64le while running osbs-box. I don't think that restricting this code path to intel only is necessary, although I'm not 100% sure what other side effects this change might have.

For the record I have been using this way patched atomic-reactor for some time on ppc64le in conjunction with osbs-box and seen no issues.

### DIFF
--- a/atomic_reactor/plugins/post_group_manifests.py
+++ b/atomic_reactor/plugins/post_group_manifests.py
@@ -6,8 +6,8 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 
 get the image manifest lists from the worker builders. If possible, group them together
-and return them. if not, return the x86_64/amd64 image manifest instead after re-uploading
-it for all existing image tags.
+and return them. if not, return empty dict after re-uploading it for all existing image
+tags.
 """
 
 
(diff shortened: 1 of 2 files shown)


twaugh:
The amd64 platform is "special" for manifest lists, in that if the client does not support manifest lists then it gets an image for the amd64 platform (or 404 if there is none).

This is the logic that lead to the decision in the code for group_manifests=false to only pick the amd64 platform.

However, it doesn't really make sense to have multiple worker builds when group_manifests=false because only one is used as the result. Given that, perhaps that case (`len(source) != 1`) should be considered an error?

jcajka:
Ok, sounds right, but what should happen if there is no intel image only non intel ones?

lcarva:
> I think the len(source) == 1 check should be instead of the self.goarch.get(platform, platform) == 'amd64' check.

Yes that makes sense. And if it's not exactly 1 raise an error?

twaugh:
Yes, raise an error if there were multiple workers.

rcerven:
@twaugh  @lcarva 
well what about len(source) > 0 ?
why not just tag digests for all platforms?

also that comment needs change or we should be returning something in case group manifests are disabled, as of now it returns empty dict

jcajka:
Updated the PR with the suggested fix and changed the comment. I hope it is now correct.
Sorry for longer iteration time.

mlangsdorf:
It's still failing the group_manifest unit test cases.

jcajka:
Tests should be fixed now, question is if it is correct expected way.

jcajka:
Fixed py3 incompatibility of tests.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -372,9 +372,9 @@ def run(self):
             else:
                 found = False
                 for platform, digest in source.items():
-                    if self.goarch.get(platform, platform) == 'amd64':
+                    if len(source) == 1:
```
lcarva: Can we just move this check outside the for-loop and raise an exception there?
```python
if len(source) != 1:
  raise ValueError('Failed to find any platform')
for ....
```

lcarva:
Updated release notes draft:
> - `post_group_manifests` runs properly without grouping manifests on non-amd64 platform

PR #1052, opened 2018-07-31
Move to using flatpak-module-tools as a library

owtaylor:
[flatpak-module-tools](https://pagure.io/flatpak-module-tools) is a command line tool to 
 * Build Flatpak modules locally
 * **Convert Flatpak modules built locally or in Koji into Flatpaks**
 * Install the resulting Flatpaks (built locally or via OSBS)

The second function shares code with atomic-reactor to do the filesystem => Flatpak conversion - I have been cutting and pasting this code back and forth as the logic evolves, but that is getting cumbersome, and also interferes with having proper tests for the code in one place. I've made the code in flatpak-module-tools into a "library" that you can import and use and would like to make atomic-reactor depend on that when building flatpaks.

flatpak-module-tools is in PyPI and in Fedora - it could be added to EPEL as necessary - but since we already have the flatpak tests made to skip when libmodulemd, etc, aren't available, the code works fine when it isn't available at all.

If there's a strong distaste for having an extra dependency, I could keep the code reorganization to depend on an external flatpak_builder.py and just cut-and-paste the file back and forth and mark it with a THIS FILE IS SHARED, but that's a bit less convenient for me, especially when it comes to tests.

This patch leaves the tests in atomic-reactor exactly as before (except for minor changes needed to sync up with flatpak-module-tools) - this gives a good assurance that the refactoring isn't changing anything - but it may be desirable over time to simplify them a bit to make the tests less fragile to changes in flatpak-module-tools .

### DIFF
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -24,6 +24,7 @@ from atomic_reactor.plugins.pre_reactor_config import get_openshift_session
 
 try:
     from atomic_reactor.plugins.pre_flatpak_create_dockerfile import get_flatpak_source_info
+    from atomic_reactor.plugins.pre_resolve_module_compose import get_compose_info
 except ImportError:
     # modulemd and/or pdc_client not available
     def get_flatpak_source_info(_):
@@ -353,7 +354,10 @@ class KojiImportPlugin(ExitPlugin):
 
         flatpak_source_info = get_flatpak_source_info(self.workflow)
         if flatpak_source_info is not None:
-            extra['image'].update(flatpak_source_info.koji_metadata())
+            compose_info = get_compose_info(self.workflow)
+            koji_metadata = compose_info.koji_metadata()
+            koji_metadata['flatpak'] = True
+            extra['image'].update(koji_metadata)
 
         koji_task_owner = get_koji_task_owner(self.session, koji_task_id).get('name')
         extra['submitter'] = self.session.getLoggedInUser()['name']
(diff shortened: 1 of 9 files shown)


atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -353,7 +354,10 @@ def get_build(self, metadata, worker_metadatas):
 
         flatpak_source_info = get_flatpak_source_info(self.workflow)
         if flatpak_source_info is not None:
-            extra['image'].update(flatpak_source_info.koji_metadata())
+            compose_info = get_compose_info(self.workflow)
+            koji_metadata = compose_info.koji_metadata()
+            koji_metadata['flatpak'] = True
```
lcarva: Let's update the docs to list out this new koji metadata value.

owtaylor: > Let's update the docs to list out this new koji metadata value.

It's not new with this patch - it just got moved around. So it's already in there: 

     - `build.extra.image.flatpak` (boolean): true if this image is a Flatpak

owtaylor:
> I wonder if it's time to create a flatpak subpackage in the spec file, similarly to what we do for koji. Or maybe it's not worth the effort.

Conceptually, it's the right thing to do - flatpak-module-tools, libmodulemd could be listed as dependencies of the subpackage and not be implicitly relied upon. And the imports in exit_koji_import/exit_koji_promote are already guarded with try/catch ImportError for other reasons, so things should work fine without the subpackage. Practically speaking - atomic-reactor is deployed in so few places that implicit dependencies aren't a huge deal. And is the spec file in atomic_reactor git actually used anywhere? I'm happy to update it if that is desired.

lcarva:
> Practically speaking - atomic-reactor is deployed in so few places that implicit dependencies aren't a huge deal. And is the spec file in atomic_reactor git actually used anywhere? I'm happy to update it if that is desired.

Agree, let's leave it as is for now. The spec file btw is copied internally during releases, but I think fedora has its own copy of it.

lcarva:
Updated release notes draft:
> - `flatpak-module-tools` library used for flatpak builds

PR #1055, opened 2018-08-10
check_and_set_platforms: log platforms and results

mlangsdorf:
add log messages for the koji platforms, and for isolated builds that
override koji platforms, the user platforms.

add code to the test cases to check for the log messages.

### DIFF
--- a/atomic_reactor/plugins/pre_check_and_set_platforms.py
+++ b/atomic_reactor/plugins/pre_check_and_set_platforms.py
@@ -55,15 +55,20 @@ class CheckAndSetPlatformsPlugin(PreBuildPlugin):
                 self.log.info("No platforms found in koji target")
                 return None
             platforms = koji_platforms.split()
+            self.log.info("Koji platforms are %s", sorted(platforms))
 
             if is_scratch_build() or is_isolated_build():
                 override_platforms = get_orchestrator_platforms(self.workflow)
                 if override_platforms and set(override_platforms) != set(platforms):
+                    sort_platforms = sorted(override_platforms)
+                    self.log.info("Received user specified platforms %s", sort_platforms)
+                    self.log.info("Using them instead of koji platforms")
                     # platforms from user params do not match platforms from koji target
                     # that almost certainly means they were overridden and should be used
                     return set(override_platforms)
         else:
             platforms = get_orchestrator_platforms(self.workflow)
+            self.log.info("No koji platforms. User specified platforms are %s", sorted(platforms))
 
         if not platforms:
             raise RuntimeError("Cannot determine platforms; no koji target or platform list")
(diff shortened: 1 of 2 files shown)


tests/plugins/test_check_and_set_platforms.py
```diff
@@ -160,10 +161,14 @@ def test_check_and_set_platforms(tmpdir, platforms, platform_exclude, platform_o
 
     plugin_result = runner.run()
     if platforms:
+        assert "Koji platforms are" in caplog.text()
+        for platform in platforms:
+            assert platform in caplog.text()
```
owtaylor: If I understand correctly, this seems like a pretty weak check, since the platform string could appear anywhere in the log - not necessarily associated with "Koji platforms are" - if  plugin args are logged or something - maybe have a helper function that you use for all the added test: 

    def assert_platforms_message(message, caplog, expected_platforms):
         matched = re.match(message + r' (\[[^\]]*\])', caplog.text())
         if not matched:
            raise RuntimeError("'{} [....]' not found in log".format(message))
         if str(expected_platforms) != matched.group(1):
             raise RuntimeError("Expected {} {}, found {}".format(message, expected_platforms, matched.group(1))

(Untested - something like that)

mlangsdorf: Or I could just sorted() the platforms in the log messages and add the sorted platforms to the message I'm looking for in the assert.

owtaylor:
Yes, modifying the code under test a bit would avoid crazy regexp/split
parsing :-)

atomic_reactor/plugins/pre_check_and_set_platforms.py
```diff
@@ -55,15 +55,20 @@ def run(self):
...
                 if override_platforms and set(override_platforms) != set(platforms):
+                    sort_platforms = sorted(override_platforms)
+                    self.log.info("Received user specified platforms %s", sort_platforms)
+                    self.log.info("Using them instead of koji platforms")
                     # platforms from user params do not match platforms from koji target
                     # that almost certainly means they were overridden and should be used
                     return set(override_platforms)
         else:
             platforms = get_orchestrator_platforms(self.workflow)
+            self.log.info("User specified platforms are %s", sorted(platforms))
```
rcerven: maybe also mention that koji target doesn't exist ?

lcarva: I think from this log message it should be easy enough to conclude that a koji platform was not used.

twaugh:
What's up with the Travis CI failure? Just that the test infrastructure was running slowly? I'll restart that one failed job.

mlangsdorf:
Added to release notes under improvements:
- `pre_check_and_set_platforms` provides log messages about what platforms are being used

PR #1058, opened 2018-08-15
pre_resolve_composes: add support for multilib

mlangsdorf:
Add support for with_multilib to pre_resolve_composes.

add the content sets definition to the schema.
x86_64 builds will generate i686 builds when with_multilib is set to true.

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -10,6 +10,7 @@ from __future__ import unicode_literals
 from datetime import datetime, timedelta
 import os
 import yaml
+import copy
 from collections import defaultdict
 
 from atomic_reactor.constants import (PLUGIN_KOJI_PARENT_KEY, PLUGIN_RESOLVE_COMPOSES_KEY,
@@ -319,12 +320,13 @@ class ComposeConfig(object):
         self.modules = data.get('modules', [])
         self.pulp = {}
         self.arches = arches or []
+        self.multilib = data.get('with_multilib', False)
+        self.flags = None
 
         if data.get('pulp_repos'):
             for arch in pulp_data or {}:
                 if arch in self.arches:
                     self.pulp[arch] = pulp_data[arch]
-            self.flags = None
             if data.get(UNPUBLISHED_REPOS):
                 self.flags = [UNPUBLISHED_REPOS]
         self.koji_tag = koji_tag
@@ -362,7 +364,11 @@ class ComposeConfig(object):
             'sigkeys': self.signing_intent['keys'],
         }
         if self.arches:
-            request['arches'] = self.arches
+            arches = copy.copy(self.arches)
+            if self.multilib and 'x86_64' in arches and 'i686' not in arches:
+                arches.append('i686')
+            # sorted to keep flexmock from being confused
+            request['arches'] = sorted(arches)
         return request
 
     def render_modules_request(self):
@@ -373,12 +379,15 @@ class ComposeConfig(object):
         }
 
     def render_pulp_request(self, arch):
+        arches = [arch]
+        if self.multilib and arch == 'x86_64':
+            arches = ['i686', 'x86_64']
         return {
             'source_type': 'pulp',
             'source': ' '.join(self.pulp.get(arch, [])),
             'sigkeys': [],
             'flags': self.flags,
-            'arches': [arch]
+            'arches': arches
         }
 
     def validate_for_request(self):
--- a/atomic_reactor/schemas/container.json
+++ b/atomic_reactor/schemas/container.json
@@ -50,6 +50,10 @@
         },
         "signing_intent": {
           "type": "string"
+        },
+        "with_multilib": {
+          "description": "Whether to request multilib composes: if true, a packages compose including x86_64 will also include i686 and a pulp x86_64 compose will include i686",
+          "type": "boolean"
         }
       },
       "additionalProperties": true
--- a/atomic_reactor/schemas/container.json
+++ b/atomic_reactor/schemas/container.json
@@ -28,7 +28,31 @@
       "additionalProperties": false
     },
     "compose": {
-      "type": "object"
+      "type": "object",
+      "properties": {
+        "packages": {
+          "description": "names of RPMs to include",
+          "type": ["array", "null"],
+          "items": {
+              "type": "string"
+          }
+        },
+        "pulp_repos": {
+          "description": "whether to build pulp composes",
+          "type": "boolean"
+        },
+        "modules": {
+          "description": "names of modules to include",
+          "type": ["array", "null"],
+          "items": {
+              "type": "string"
+          }
+        },
+        "signing_intent": {
+          "type": "string"
+        }
+      },
+      "additionalProperties": true
     },
     "flatpak": {
       "type": "object"
--- a/tests/plugins/test_resolve_composes.py
+++ b/tests/plugins/test_resolve_composes.py
@@ -287,29 +287,112 @@ class TestResolveComposes(object):
 
         self.run_plugin_with_args(workflow, reactor_config_map=reactor_config_map)
 
-    @pytest.mark.parametrize(('pulp_arches', 'arches', 'signing_intent', 'expected_intent'), (  # noqa:F811
-        (None, None, 'unsigned', 'unsigned'),
-        # For the next test, since arches is none, no compose is performed even though pulp_arches
-        # has a value. Expected intent doesn't change when nothing is composed.
-        (['x86_64'], None, 'release', 'release'),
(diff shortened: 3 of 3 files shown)


atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -329,6 +329,10 @@ def __init__(self, data, pulp_data, odcs_config, koji_tag=None, arches=None):
                 self.flags = [UNPUBLISHED_REPOS]
         self.koji_tag = koji_tag
         self.odcs_config = odcs_config
+        self.arches = arches or []
+        if data.get('with_multilib', False):
```
twaugh: Please add this key to the schema.

tests/plugins/test_resolve_composes.py
```diff
@@ -289,24 +289,45 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
+                                        multilib):
         content_set = ''
         pulp_composes = {}
         base_repos = ['spam', 'bacon', 'eggs']
         pulp_id = ODCS_COMPOSE_ID
+        arches = arches or []
+        temp_arches = arches
+        pulp_arches = pulp_arches or []
+        if multilib:
+            if 'x86_64' in pulp_arches and 'i686' not in pulp_arches:
```
twaugh: Rather than repeating this logic, is there a way to include the full list of expected architectures in the parametrization?

mlangsdorf: It's not pretty, but sure.

atomic_reactor/plugins/pre_resolve_composes.py
mlangsdorf: from atomic_reactor/schemas/container.json
```
    "compose": {
      "type": "object"
    },
```
There's not much there to add to.

twaugh: It needs the properties definitions, like in https://github.com/projectatomic/atomic-reactor/pull/1056.

Needs definitions for packages (list of str), modules (list of str), signing_intent (str), and with_multilib (bool).

tests/plugins/test_resolve_composes.py
```diff
@@ -287,27 +287,62 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
+        # has a value. Expected intent doesn't change when nothing is composed.
+        (None, ['x86_64', 'ppce64le', 'arm64'], True, None, 'beta', 'beta'),
     ))
     @pytest.mark.parametrize(('flags', 'expected_flags'), [
         ({}, None,),
         ({UNPUBLISHED_REPOS: False}, None),
         ({UNPUBLISHED_REPOS: True, 'some_flag': True}, [UNPUBLISHED_REPOS])
     ])
-    def test_request_pulp_and_multiarch(self, workflow, reactor_config_map, pulp_arches, arches,
+    def test_request_pulp_and_multiarch(self, workflow, reactor_config_map, arches, pulp_arches, multilib, allowed_arches,
```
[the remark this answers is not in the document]

atomic_reactor/plugins/pre_resolve_composes.py

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -319,6 +319,9 @@ def __init__(self, data, pulp_data, odcs_config, koji_tag=None, arches=None):
         self.modules = data.get('modules', [])
         self.pulp = {}
         self.arches = arches or []
+        if data.get('with_multilib', False):
+            if 'x86_64' in self.arches and 'i686' not in self.arches:
+                self.arches.append('i686')
```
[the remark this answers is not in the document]
mlangsdorf: If i686 is added to self.arches now, it can be used to confirm the i686 that needs to be in pulp_repos during the check from 326-329.

If there isn't an i686 entry in pulp_repos, there's no way to do a multilib pulp build, because there wasn't an i686 entry in content_sets.yml and there's no way to guess what RPMs should be included in the i686 pulp compose. Currently we silently don't build the i686 pulp in that case, but I could change that to either log a warning or raise an error if you prefer.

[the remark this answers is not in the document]
lcarva: The `i686` packages should be available for building `x86_64` images. That's what the multilib story is all about. We're not building `i686` container images. 

If we handle adding `i686` in the render methods, the checks from 326-329 don't need changing.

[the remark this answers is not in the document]
mlangsdorf: Okay, I misread the request.

Fixed.

Drastically extended the test cases to nail down every variation I could think of, but in a hopefully clearer manner.

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -362,7 +363,11 @@ def render_packages_request(self):
             'sigkeys': self.signing_intent['keys'],
         }
         if self.arches:
-            request['arches'] = self.arches
+            arches = self.arches
```
lcarva: You probably wanna create a copy here.

lcarva: Also, we need similar change in `render_pulp_request`.

mlangsdorf: render pulp request doesn't assign anything that would need a copy afaict.
```
    def render_pulp_request(self, arch):
        return {
            'source_type': 'pulp',
            'source': ' '.join(self.pulp.get(arch, [])),
            'sigkeys': [],
            'flags': self.flags,
            'arches': [arch]
        }
```
copy what?

lcarva: When you do this: `arches = self.arches` it doesn't create a copy, so all the changes you're doing in arches, will propagate to `self.arches`. That's error prone. Let's not modify `self.arches` in this method.

In render_pulp_request, you need to insert `i686` in list of arches if arch is `x86_64`:
```python
    def render_pulp_request(self, arch):
        arches = [arch]
        if self.multilib and arch == 'x86_64':
          arches.append('i686')
        return {
            'source_type': 'pulp',
            'source': ' '.join(self.pulp.get(arch, [])),
            'sigkeys': [],
            'flags': self.flags,
            'arches': arches
        }
```

atomic_reactor/plugins/pre_resolve_composes.py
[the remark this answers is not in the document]
mlangsdorf: Fixed again, hopefully correctly.

mlangsdorf:
Fixed the JSON schema. Again. For real this time.

atomic_reactor/schemas/container.json
```diff
@@ -28,7 +28,34 @@
...
+        },
+        "modules": {
+          "description": "names of modules to include",
+          "type": ["array", "null"],
+          "items": {
+              "type": "string"
+          }
+        },
+        "signing_intent": {
+          "type": ["string", "null"]
```
twaugh: I think it has to be a string doesn't it?

tests/plugins/test_resolve_composes.py
```diff
@@ -287,29 +287,112 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):  # noq
...
-        # pulp composes have the beta signing intent but the unsigned intent overrides that
-        (['x86_64', 'ppce64le', 'arm64'], ['x86_64', 'ppce64le', 'arm64'], 'unsigned', 'unsigned'),
-        # For the next test, since arches is none, no compose is performed even though pulp_arches
-        # has a value. Expected intent doesn't change when nothing is composed.
-        (['x86_64', 'ppce64le', 'arm64'], None, 'beta', 'beta'),
+    @pytest.mark.parametrize(('arches', 'final_arches', 'multilib',   # noqa:F811
+                              'pulp_arches', 'is_pulp',
+                              'signing_intent', 'expected_intent'), (
+        # submit no arches and no pulp, final_arches and is_pulp are none, intent doesn't change
+        (None, None, False, [], [], 'unsigned', 'unsigned'),
```
twaugh: Does `None` have a special meaning other than `[]` here?

atomic_reactor/schemas/container.json
```diff
@@ -28,7 +28,34 @@
...
+        },
+        "with_multilib": {
+          "type": "boolean"
```
twaugh: Could add a description? e.g. "Whether to request multilib composes: a packages compose including x86_64 will also include i686; a pulp x86_64 compose will include i686"

atomic_reactor/schemas/container.json
mlangsdorf: Fixed.

mlangsdorf: Helps if I push the fix.

lcarva:
Updated release notes draft:
> - `compose` section of `container.yaml` validated against schema
> - Added support for optional multi-lib container images

PR #1059, opened 2018-08-16
dont pull base image in orchestrator

rcerven:


atomic_reactor/util.py
```diff
@@ -1017,6 +1102,8 @@ def get_config_from_registry(image, registry, digest, insecure=False,
     tag = image.tag or 'latest'
     logger.debug('Image %s:%s has config:\n%s', context, tag, blob_config)
 
+    if return_config_digest:
+        return blob_config, config_digest
```
twaugh: Ick. Can't we make a new function `get_config_and_id_from_registry()`, and implement `get_config_from_registry()` using that?

atomic_reactor/util.py
```diff
@@ -985,8 +985,93 @@ def get_manifest_list(image, registry, insecure=False, dockercfg_path=None):
     return response
 
 
+def get_all_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
+                             versions=('v1', 'v2', 'v2_list', 'oci', 'oci_index')):
```
twaugh: Let's not attempt to fetch oci and oci_index when we don't care about the results.

atomic_reactor/util.py
```diff
@@ -985,8 +985,93 @@ def get_manifest_list(image, registry, insecure=False, dockercfg_path=None):
...
+    blob_config = None
+    config_digest = None
+    image_inspect = {}
+
+    # we have manifest list (get digest for 1st platform)
+    if 'v2_list' in all_man_digests:
+        man_list_json = all_man_digests['v2_list'].json()
+        v2_digest = man_list_json['manifests'][0]['digest']
+        blob_config, config_digest = get_config_from_registry(image, registry, v2_digest,
+                                                              insecure=insecure, version='v2',
```
twaugh: This assumes the manifest list entry is for a v2 schema 2 image manifest. We should at least assert that, even if we aren't prepared to handle v2 schema 1 here.

atomic_reactor/util.py
```diff
@@ -985,8 +985,93 @@ def get_manifest_list(image, registry, insecure=False, dockercfg_path=None):
...
+    :param image: ImageName, the remote image to inspect
+    :param registry: str, URI for registry, if URI schema is not provided,
+                          https:// will be used
+    :param insecure: bool, when True registry's cert is not verified
+    :param dockercfg_path: str, dirname of .dockercfg location
+
+    :return: dict of inspected image
+    """
+    all_man_digests = get_all_manifest_digests(image, registry, insecure=insecure,
+                                               dockercfg_path=dockercfg_path)
```
twaugh: Maybe restrict versions here to those we can handle?

atomic_reactor/util.py
```diff
@@ -985,8 +985,93 @@ def get_manifest_list(image, registry, insecure=False, dockercfg_path=None):
...
+        blob_config, config_digest = get_config_from_registry(image, registry, image.tag,
+                                                              insecure=insecure, version='v2',
+                                                              return_config_digest=True,
+                                                              dockercfg_path=dockercfg_path)
+    # read config from v1
+    elif 'v1' in all_man_digests:
+        v1_json = all_man_digests['v1'].json()
+        blob_config = json.loads(v1_json['history'][0]['v1Compatibility'].decode('utf-8'))
+    else:
+        raise NotImplementedError("OCI isn't implemented")
```
twaugh: We don't really know this is an OCI image. Maybe something more generic like "No v2 schema 1 image, or v2 schema 2 image or list, found".

atomic_reactor/util.py
```diff
@@ -985,8 +985,93 @@ def get_manifest_list(image, registry, insecure=False, dockercfg_path=None):
...
+
+    # dictionary to convert config keys to inspect keys
+    config_2_inspect = {
+        'created': 'Created',
+        'os': 'Os',
+        'container_config': 'ContainerConfig',
+        'architecture': 'Architecture',
+        'docker_version': 'DockerVersion',
+        'config': 'Config'
```
twaugh: Nicer to end the last entry with a trailing comma to make single-line diffs easier on the eye.

atomic_reactor/build.py
```diff
@@ -12,8 +12,10 @@
 import json
 
 import logging
+import docker
```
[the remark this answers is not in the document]
rcerven: yop

atomic_reactor/build.py
```diff
@@ -178,21 +183,56 @@ def set_df_path(self, path):
...
+        return self._base_image_inspect
+
+    def parent_image_inspect(self, image):
+        """
+        inspect parent image
+
+        :return: dict
+        """
+        if not self._parent_images_inspect or image not in self._parent_images_inspect:
+            if not self._base_image_pulled:
```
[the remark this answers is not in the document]
twaugh: It means set_base_image(pulled=True) was called. Currently we only call that if all the parent images were called, but is that what pulled=True means? It's not very explicit.

atomic_reactor/build.py
```diff
@@ -178,21 +183,56 @@ def set_df_path(self, path):
...
         """
         inspect base image
 
         :return: dict
         """
-        logger.info("inspecting base image '%s'", self.base_image)
-        inspect_data = self.tasker.inspect_image(self.base_image)
-        return inspect_data
+        if self._base_image_inspect is None:
+            if not self._base_image_pulled:
```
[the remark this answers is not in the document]
rcerven: ok

atomic_reactor/build.py
```diff
@@ -178,21 +183,56 @@ def set_df_path(self, path):
...
+                self._base_image_inspect = get_inspect_for_image(self.base_image,
+                                                                 self.base_image.registry,
+                                                                 self._base_image_insecure)
+            else:
+                try:
+                    self._base_image_inspect = self.tasker.inspect_image(self.base_image)
+
+                except docker.errors.NotFound:
+                    # If the base image cannot be found throw KeyError -
```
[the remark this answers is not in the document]
rcerven: _base_image_pulled is explicitly set only in pull_base_image via set_base_image, that is also used in different plugins, I know for sure only in pull_base_image that I've pulled that image, so if in worker build by any chance some plugin after pull_base_image (change_from_in_df) would change base image to wrong image, it would fail here

atomic_reactor/build.py
```diff
@@ -178,21 +183,56 @@ def set_df_path(self, path):
...
+
+        return self._base_image_inspect
+
+    def parent_image_inspect(self, image):
+        """
+        inspect parent image
+
+        :return: dict
+        """
+        if not self._parent_images_inspect or image not in self._parent_images_inspect:
```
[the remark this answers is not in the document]
rcerven: yop it is, deleting

atomic_reactor/build.py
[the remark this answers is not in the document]
rcerven: yes as in pull_base_image, all parents + base are pulled or not, so not sure how else to call it

atomic_reactor/build.py
```diff
@@ -178,21 +183,56 @@ def set_df_path(self, path):
...
+            else:
+                try:
+                    self._base_image_inspect = self.tasker.inspect_image(self.base_image)
+
+                except docker.errors.NotFound:
+                    # If the base image cannot be found throw KeyError -
+                    # as this property should behave like a dict
+                    raise KeyError("Unprocessed base image Dockerfile cannot be inspected")
+
+            if str(self.base_image) not in self._parent_images_inspect:
```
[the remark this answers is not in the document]
rcerven: ok

atomic_reactor/util.py

atomic_reactor/util.py

atomic_reactor/build.py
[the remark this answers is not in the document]
twaugh: Maybe `parents_pulled`? The base image is just one of the parent images, after all.

atomic_reactor/util.py
rcerven: yeah will restring it in that function

atomic_reactor/util.py

atomic_reactor/util.py

atomic_reactor/build.py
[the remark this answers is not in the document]
rcerven: ok

atomic_reactor/util.py

atomic_reactor/util.py
```diff
@@ -985,7 +985,95 @@ def get_manifest_list(image, registry, insecure=False, dockercfg_path=None):
     return response
 
 
-def get_config_from_registry(image, registry, digest, insecure=False,
+def get_all_manifest_digests(image, registry, insecure=False, dockercfg_path=None,
```
twaugh: Would `get_all_manifests` be a better name for this function? I think it returns the manifests themselves, and not the digests.

tests/test_util.py
```diff
@@ -905,7 +905,14 @@ def test_df_parser_parent_env_wf(tmpdir, caplog, env_arg):
         """)
     env_conf = {INSPECT_CONFIG: {"Env": env_arg}}
     workflow = DockerBuildWorkflow(MOCK_SOURCE, 'test-image')
-    flexmock(workflow, base_image_inspect=env_conf)
+
+    class MockInsideBuilder(object):
```
twaugh: Maybe use `StubInsideBuilder` for this?

atomic_reactor/util.py
```diff
@@ -985,7 +985,95 @@ def get_manifest_list(image, registry, insecure=False, dockercfg_path=None):
...
+                             versions=('v1', 'v2', 'v2_list')):
+    """Return manifest digests for image.
+
+    :param image: ImageName, the remote image to inspect
+    :param registry: str, URI for registry, if URI schema is not provided,
+                          https:// will be used
+    :param insecure: bool, when True registry's cert is not verified
+    :param dockercfg_path: str, dirname of .dockercfg location
```
twaugh: Could add a description of `versions` here.

atomic_reactor/util.py

atomic_reactor/util.py
rcerven: will do

tests/test_util.py
rcerven: ok

atomic_reactor/build.py
```diff
@@ -12,8 +12,10 @@
 import json
 
 import logging
+import docker.errors
+import atomic_reactor.util
```
twaugh: This seems to be only for `get_inspect_for_image`. Why not add that to the list of symbols we `from`...`import` from that module (on line 18)? Could also help with line wrapping in places where we call that function.

rcerven: that is just for unit tests, as I am unable to mock that function which is in different namespace for testing : test_parent_image_inspect, test_base_image_inspect

PR #1060, opened 2018-08-20
flatpak_create_dockerfile: create a skeleton /dev in the installroot

owtaylor:
rpm scripts that are being run in the installroot to create the Flatpak
filesystem may access certain standard files in /dev - create a very
small /dev/ to make such scripts work reliably. See, for example:
https://bugzilla.redhat.com/show_bug.cgi?id=1614162#c3


owtaylor:
(Updated to have 'mkdir -p' instead of a plain 'mkdir')

lcarva:
Updated release notes draft:
> - `flatpak_create_dockerfile` creates a skeleton /dev in the installroot

PR #1061, opened 2018-08-21
tests: increase delays to avoid occasional timeout failures

twaugh:
Tests that are sensitive to delays of 0.4s occasionally fail. Increase the minimum time between pass and fail to more than 0.4s.


lcarva:

PR #1065, opened 2018-08-29
Allow setting the Flatpak base image in the reactor config map

owtaylor:
The Flatpak base image is not specific to a particular build, so should
be set in the reactor config map.

### DIFF
--- a/atomic_reactor/schemas/config.json
+++ b/atomic_reactor/schemas/config.json
@@ -485,6 +485,17 @@
     "default_image_build_method": {
         "description": "Specify different default buildstep plugin for worker builds",
         "enum": ["docker_api", "imagebuilder"]
+    },
+    "flatpak": {
+        "description": "Flatpak specific properties",
+        "type": "object",
+        "properties": {
+            "base_image": {
+                "description": "host image used to install packages when creating a Flatpak",
+                "type": "string"
+            }
+        },
+        "additionalProperties": false
     }
   },
   "definitions": {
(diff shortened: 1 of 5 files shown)


atomic_reactor/schemas/config.json
```diff
@@ -485,6 +485,17 @@
...
+    "flatpak": {
+        "description": "Flatpak specific properties",
+        "type": "object",
+        "properties": {
+            "base_image": {
+                "description": "host image used to install packages when creating a Flatpak",
+                "type": "string"
+            }
+        },
+        "additionalProperties": {"type": "string"}
```
twaugh: Why allow additional properties here? We don't in other places.

tests/plugins/test_flatpak_create_dockerfile.py
```diff
@@ -114,10 +118,18 @@ def test_flatpak_create_dockerfile(tmpdir, docker_tasker, config_name, breakage)
...
+    base_image = "registry.fedoraproject.org/fedora:latest"
+
     args = {
-        'base_image': "registry.fedoraproject.org/fedora:latest",
+        'base_image': base_image,
     }
 
+    if reactor_config_map:
+        workflow.plugin_workspace[ReactorConfigPlugin.key] = {}
+        workflow.plugin_workspace[ReactorConfigPlugin.key][WORKSPACE_CONF_KEY] =\
```
twaugh: Or could assign the dict, `= {`\n`WORKSPACE_CONF_KEY: ...` (also might make line wrapping easier)

owtaylor:
Updated to disallow other properties (I copied-and-modifed an outlier), and to assign the dict.

lcarva:
Updated release notes draft:
> - Flatpak base image can be set in reactor config map

PR #1066, opened 2018-08-30
add organization & expected_media_types for registries to schema

rcerven:

### DIFF
--- a/atomic_reactor/schemas/config.json
+++ b/atomic_reactor/schemas/config.json
@@ -418,6 +418,7 @@
                     "description": "Registry URI including version",
                     "type": "string"
                 },
+                "organization": {"$ref": "#/definitions/organization"},
                 "insecure": {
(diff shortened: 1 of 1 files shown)


atomic_reactor/schemas/config.json
```diff
@@ -451,6 +468,10 @@
                 "description": "Registry url",
                 "type": "string"
             },
+            "organization": {
```
twaugh: I wonder if we want a "definition" for a registry, so we can reference it from both of these places.

lcarva: Probably a good idea.

rcerven: ack will do

rcerven:
added organization to definitions

rcerven:
rn updated
- Added organization & expected_media_types for registries to schema

PR #1067, opened 2018-08-31
Add support for Bearer token auth registries

lcarva:
Still needs unit tests, but I'd really appreciate some early feedback.

I've tested this against multiple container registries: Pulp, OpenShift Registry, Quay.io, and Proxy to Quay.io.

### DIFF
--- a/atomic_reactor/auth.py
+++ b/atomic_reactor/auth.py
@@ -1,0 +1,172 @@
+"""
+Copyright (c) 2018 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+from __future__ import unicode_literals
+
+from requests.auth import AuthBase, HTTPBasicAuth
+from requests.cookies import extract_cookies_to_jar
+from requests.utils import parse_dict_header
+from six.moves.urllib.parse import urlparse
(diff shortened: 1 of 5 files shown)


atomic_reactor/auth.py
```diff
@@ -0,0 +1,171 @@
...
+            return self.handle_401(response, repo, **kwargs)
+
+        response.register_hook('response', handle_401_with_repo)
+        return response
+
+    def handle_401(self, response, repo, **kwargs):
+        """Fetch Bearer token and retry."""
+
+        if response.status_code != requests.codes.unauthorized:
+            return
```
rcerven: this should be : "return response"

lcarva: Good catch! Fixed it.

lcarva:
Not exactly sure why unit tests seems to pass only for python 2 on fedora 28. Will debug further in the morning.

lcarva:
Issue was with, old, version of responses being used. Updated `tests/requirements.txt` to use a newer one.
All tests are passing now.

lcarva:
> Please update the docs; the rest looks good.

Ooops! Forgot about it. All set now.

lcarva:
Updated release notes draft:
> - Support for registries that use Bearer token authentication

PR #1070, opened 2018-09-05
Add ImageName.enclose method

lcarva:
This allows an image reference to be modified so it can be nested under
a namespace/organization.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -113,6 +113,36 @@ def test_image_name_format():
         assert image_name.to_str() == expected
 
 
+@pytest.mark.parametrize(('repo', 'organization', 'enclosed_repo'), (
+    ('fedora', 'spam', 'spam/fedora'),
+    ('spam/fedora', 'spam', 'spam/fedora'),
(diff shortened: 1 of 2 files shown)


tests/test_util.py
```diff
@@ -115,6 +115,36 @@ def test_image_name_format():
...
+    reference = repo
+    if tag:
+        reference = '{}:{}'.format(repo, tag)
+    if registry:
+        reference = '{}/{}'.format(registry, reference)
+
+    image_name = ImageName.parse(reference)
+    assert image_name.get_repo() == repo
+    assert image_name.registry == registry
+    assert image_name.tag == tag
```
lcarva: This line needs changing once https://github.com/projectatomic/atomic-reactor/pull/1062 is merged:
```python
assert image_name.tag == (tag or 'latest')
```

lcarva: Updated.

lcarva:
Updated release notes draft:
> - ImageName has enclose method to nest repos in organizations

PR #1073, opened 2018-09-14
pre_change_from_in_df: adjust for ImageName having a default tag

mlangsdorf:
now that ImageNames supply a default tag of ':latest', pre_change_from needs to be updated to take into account the new logic.
    
in the plugin itself, this involves being explicit about when a value is a raw string, an ImageName, or an ImageName.to_str().
    
In the test cases, this requires providing ':latest' tags for parent_images keys.
    
One test case was changed use a literal expected value instead of a calculated expected value, both as good practice and because calculating the expected value became much more complicated.

### DIFF
--- a/atomic_reactor/build.py
+++ b/atomic_reactor/build.py
@@ -146,7 +146,7 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
         self.original_base_image = None
         self._base_image_inspect = None
         self._parents_pulled = False
-        self.parent_images = {}  # dockerfile image => locally available image
+        self.parent_images = {}  # dockerfile ImageName => locally available ImageName
         self._parent_images_inspect = {}  # locally available image => inspect
         self.image_id = None
         self.built_image_info = None
@@ -181,13 +181,13 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
         logger.debug("base image specified in dockerfile = '%s'", self.base_image)
         self.parent_images.clear()
         for image in dfp.parent_images:
-            self.parent_images[image] = None
+            self.parent_images[ImageName.parse(image)] = None
 
     def set_base_image(self, base_image, parents_pulled=True, insecure=False):
         logger.info("setting base image to '%s'", base_image)
         self.base_image = ImageName.parse(base_image)
         self.original_base_image = self.original_base_image or self.base_image
-        self.parent_images[str(self.original_base_image)] = base_image
+        self.parent_images[self.original_base_image] = self.base_image
         self._parents_pulled = parents_pulled
         self._base_image_insecure = insecure
 
@@ -226,17 +226,17 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
 
         :return: dict
         """
-        if image not in self._parent_images_inspect:
+        image_name = ImageName.parse(image)
+        if image_name not in self._parent_images_inspect:
             if self._parents_pulled:
-                self._parent_images_inspect[image] = self.tasker.inspect_image(image)
+                self._parent_images_inspect[image_name] = self.tasker.inspect_image(image)
             else:
-                image_obj = ImageName.parse(image)
-                self._parent_images_inspect[image] =\
-                    atomic_reactor.util.get_inspect_for_image(image_obj,
-                                                              image_obj.registry,
+                self._parent_images_inspect[image_name] =\
+                    atomic_reactor.util.get_inspect_for_image(image_name,
+                                                              image_name.registry,
                                                               self._base_image_insecure)
 
-        return self._parent_images_inspect[image]
+        return self._parent_images_inspect[image_name]
 
     def inspect_built_image(self):
         """
--- a/atomic_reactor/plugins/pre_change_from_in_df.py
+++ b/atomic_reactor/plugins/pre_change_from_in_df.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2015 Red Hat, Inc
+Copyright (c) 2015-2018 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
@@ -10,7 +10,7 @@ Pre-build plugin that changes the parent images used in FROM instructions
 to the more specific names given by the builder.
 """
 from atomic_reactor.plugin import PreBuildPlugin
-from dockerfile_parse import DockerfileParser
+from atomic_reactor.util import ImageName, df_parser
 
 
 class BaseImageMismatch(RuntimeError):
@@ -44,10 +44,10 @@ class ChangeFromPlugin(PreBuildPlugin):
 
     def run(self):
         builder = self.workflow.builder
-        dfp = DockerfileParser(builder.df_path)
+        dfp = df_parser(builder.df_path)
 
-        df_base = dfp.baseimage
-        build_base = builder.base_image.to_str()
+        df_base = ImageName.parse(dfp.baseimage)
+        build_base = builder.base_image
 
         # do some sanity checks to defend against bugs and rogue plugins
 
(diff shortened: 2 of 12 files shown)


atomic_reactor/plugins/pre_change_from_in_df.py
```diff
@@ -44,19 +44,29 @@ def __init__(self, tasker, workflow):
...
+        build_base_image = builder.base_image
+        build_base_str = builder.base_image.to_str()
 
         # do some sanity checks to defend against bugs and rogue plugins
-
-        if build_base != builder.parent_images[df_base]:
+        # builder.base_image may have a default tag of latest, and builder.parent_images[]
+        # won't have a default tag. So convert builder.parent_images[] to an ImageName
+        # so we're comparing apples to apples.
+        if build_base_str != str(ImageName.parse(builder.parent_images[df_base_str])):
```
lcarva: What are the changes that we're gonna mess this up again in the future? 

Wouldn't it make it much simpler if keys in parent_image were simply ImageName objects? We could even make it a custom dict to enforce that.

mlangsdorf: @sosiouxme this is your design element.

sosiouxme: I think with this behavior of ImageName modifying the image it gets, it is too confusing to try to keep straight which `foo_image` is a string versus an ImageName versus a string representation of one for purposes of comparison. I would suggest that anywhere we start thinking of something as an image name, we parse it into an ImageName immediately and store and compare that way.

So `parent_images` becomes a dict of ImageName => ImageName. `original_base_name` becomes an ImageName. `set_base_name` should probably accept either an ImageName or a string that is parsed into an ImageName like now. There are only a few places where it will need to become a string again -- I think just the metadata plugins and of course when writing to DF.

Does that sound good?

mlangsdorf: Works for me. I'll go implement lucarva's suggestion.

mlangsdorf: That was fun. Fixed I think.

tests/plugins/test_change_from_in_df.py
```diff
@@ -23,18 +23,23 @@
 from textwrap import dedent
 
 
-def mock_workflow():
+def mock_workflow(image_name=None, dockerfile_path=None, parent_name=None, parent_images=None):
```
sosiouxme: This change is reversing the [test data builder pattern](https://testing.googleblog.com/2018/02/testing-on-toilet-cleanly-create-test.html) that we wanted to follow.

atomic_reactor/build.py
```diff
@@ -146,7 +146,7 @@ def __init__(self, source, image, **kwargs):
         self.original_base_image = None
         self._base_image_inspect = None
         self._parents_pulled = False
-        self.parent_images = {}  # dockerfile image => locally available image
+        self.parent_images = {}  # dockerfile ImageName => locally available ImageName
```
sosiouxme: There are a few more places where this change has to be taken into account e.g.

https://github.com/projectatomic/atomic-reactor/blob/f4122f110b7b4d53dcf499cdecd381a21ab9b30b/atomic_reactor/plugins/exit_store_metadata_in_osv3.py#L236

ImageName is not JSON serializable, and we would want the string in the output anyway.

tests/plugins/test_change_from_in_df.py
mlangsdorf: Fixed.

atomic_reactor/build.py
mlangsdorf: So my first thought was to just make ImageName JSON serializable, but a couple hours of banging my head against Python inheritance and overriding made me rethink that.

I added a little function that converts a dict of parent images into a dict of parent image strings.

sosiouxme: https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/pre_koji_parent.py#L83-L84
pulls the image name out of `parent_images` and calls
https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/build.py#L223
which is expecting a string - think that needs to be changed.

atomic_reactor/build.py
```diff
@@ -187,7 +187,7 @@ def set_base_image(self, base_image, parents_pulled=True, insecure=False):
         logger.info("setting base image to '%s'", base_image)
         self.base_image = ImageName.parse(base_image)
         self.original_base_image = self.original_base_image or self.base_image
-        self.parent_images[str(self.original_base_image)] = base_image
+        self.parent_images[str(self.original_base_image)] = self.base_image
```
sosiouxme: confused. shouldn't this be without the `str()` ?

There seem to be some more bits in this file that need updates like
https://github.com/projectatomic/atomic-reactor/blob/6ae4ec329f7b81eb3ffcc62734d392c2165cc259/atomic_reactor/build.py#L183-L184
... will initialize it with strings won't it?

atomic_reactor/build.py
mlangsdorf: Fixed.

atomic_reactor/build.py
mlangsdorf: Fixed.

atomic_reactor/plugins/pre_change_from_in_df.py
```diff
@@ -96,5 +99,5 @@ def run(self):
         builder.set_base_image(parent_image_ids[df_base])
         self.log.debug(
             "for base image '%s' using local image '%s', id '%s'",
-            df_base, build_base, builder.base_image
+            df_base, build_base_str, build_base
```
lcarva: When using '%s' placeholder in log message, logging module will try to convert object to string automatically by calling the object's `__str__` method (over simplified statement). Do we need the `build_base_str` at all?

lcarva: I'm also not sure why this log message used to output got `build_base` and `builder.base_image`. I can't see how they would ever not be the same thing. @sosiouxme ?

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -69,11 +69,11 @@ def run(self):
         build_json = get_build_json()
         current_platform = platform.processor() or 'x86_64'
         self.manifest_list_cache = {}
-        for nonce, parent in enumerate(sorted(self.workflow.builder.parent_images.keys())):
-            image = ImageName.parse(parent)
+        for nonce, parent in enumerate(sorted(self.workflow.builder.parent_images.keys(),
+                                              key=lambda parent_image: str(parent_image))):
```
lcarva: Can be simplified: `key=str`

lcarva:
@mlangsdorf, I wonder if we need to make changes to `pre_koji_parent` as well. 

Also, in build.py, parent_image_inspect should probably convert image param to ImageName object.

mlangsdorf:
I adjusted the test cases for koji_parent and it looks like it's doing the right thing without any changes.

atomic_reactor/plugins/pre_pull_base_image.py

tests/test_util.py
```diff
@@ -113,6 +113,15 @@ def test_image_name_format():
         assert image_name.to_str() == expected
 
 
+def test_image_name_parse_image_name(caplog):
+    warning = 'Attempting to parse ImageName test:latest as an ImageName'
+    test = ImageName.parse("test")
+    assert warning not in caplog.text()
+    image_test = ImageName.parse(test)
+    assert warning in caplog.text()
+    assert test == image_test
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

atomic_reactor/util.py
```diff
@@ -73,6 +73,10 @@ def __init__(self, registry=None, namespace=None, repo=None, tag=None):
     def parse(cls, image_name):
         result = cls()
 
+        if isinstance(image_name, ImageName):
```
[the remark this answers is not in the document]

atomic_reactor/plugins/pre_pull_base_image.py
lcarva: I don't see this change pushed.

mlangsdorf: Apparently you have to commit before you push.
Fixed for real this time.

lcarva:
Not sure why py2.6 test is marked as failed. I didn't see any failures in output. Restarted it.

lcarva:

PR #1074, opened 2018-09-19
OSBS-6224 reject COPY --from <image>

sosiouxme:
Since pulling and updating the image in this construct is complicated, disallow doing so.


atomic_reactor/build.py
```diff
@@ -183,6 +185,29 @@ def set_df_path(self, path):
         for image in dfp.parent_images:
             self.parent_images[image] = None
 
+        # validate user has not specified COPY --from=image
+        builders = []
+        for stmt in dfp.structure:
+            if stmt['instruction'] == 'FROM':
+                match = re.match(r'\S+\s+as\s+(\S+)', stmt['value'], re.I)
+                builders.append(match.group(1) if match else str(len(builders)))
```
lcarva: This is so clever, it could be admittance to a secret club.

atomic_reactor/build.py
```diff
@@ -183,6 +185,29 @@ def set_df_path(self, path):
...
+            if stmt['instruction'] == 'COPY':
+                match = re.match(r'--from=(\S+)', stmt['value'], re.I)
```
lcarva: Could there be white spaces before `--from` ? Or will dockerfile-parse trim the value?

atomic_reactor/build.py
```diff
@@ -183,6 +185,29 @@ def set_df_path(self, path):
         for image in dfp.parent_images:
             self.parent_images[image] = None
 
+        # validate user has not specified COPY --from=image
+        builders = []
+        for stmt in dfp.structure:
+            if stmt['instruction'] == 'FROM':
+                match = re.match(r'\S+\s+as\s+(\S+)', stmt['value'], re.I)
+                builders.append(match.group(1) if match else str(len(builders)))
+            if stmt['instruction'] == 'COPY':
```
lcarva: `elif`?

atomic_reactor/build.py
```diff
@@ -183,6 +185,29 @@ def set_df_path(self, path):
...
+                match = re.match(r'--from=(\S+)', stmt['value'], re.I)
+                if not match:
+                    continue
+                stage = match.group(1)
+                if any(stage in [str(idx), builder] for idx, builder in enumerate(builders[:-1])):
```
lcarva: Probably worth a comment explaining why the last item in `builders` is ignored.

atomic_reactor/build.py
sosiouxme: Oh, that's not a compliment is it?

atomic_reactor/build.py
sosiouxme: rather than test and comment about that, probably just best to handle them by using `search` instead of `match`

atomic_reactor/build.py
sosiouxme: Yeah. Actually... while it would be weird, I wonder if it would actually work for the stage to refer to itself there...

tests/test_build.py
```diff
@@ -179,6 +181,33 @@ def test_no_base_image(tmpdir):
...
+    source = {'provider': 'path', 'uri': 'file://' + str(tmpdir), 'tmpdir': str(tmpdir)}
+
+    dfp.content = dedent("""\
+        FROM monty AS vikings
+        FROM python
+        COPY --from=vikings /spam/eggs /bin/eggs
+        COPY --from=0 /spam/eggs /bin/eggs
+        COPY src dest
+    """)
+    InsideBuilder(get_source_instance_for(source), 'built-img')  # calls set_df_path at init
```
mlangsdorf: stylistic nit: comments should proceed code, not caption them.
also, not sure the inline comment is necessary or even helpful.

atomic_reactor/build.py
lcarva: It's not an insult either. I thought it was cool :)

tests/test_build.py
```diff
@@ -179,6 +181,33 @@ def test_no_base_image(tmpdir):
...
+    # should not raise an error
+
+    dfp.content = dedent("""\
+        FROM monty as vikings
+        FROM python
+        COPY --from=notvikings /spam/eggs /bin/eggs
+    """)
+    with pytest.raises(RuntimeError) as exc_info:
+        InsideBuilder(get_source_instance_for(source), 'built-img')  # calls set_df_path at init
+    assert "COPY --from" in str(exc_info.value)
```
mlangsdorf: is it worth adding one more test case like this?:
```
dfp.content = dedent("""\
        FROM monty as vikings
        FROM python
        COPY --from=2 /spam/eggs /bin/eggs
    """)
```

It's the same logic as notvikings, I get that, but it might be useful for someone puzzling out the code in 6 months.

atomic_reactor/build.py
sosiouxme: Actually a bit of the cleverness wasn't needed at all. I hope it's clearer now.

atomic_reactor/build.py
sosiouxme: Turns out self-reference *does* work and I could even imagine where it would be useful. So, removed the extra noise.

tests/test_build.py
sosiouxme: yes, thanks!

sosiouxme:
Thanks for reviews; should be ready for another.

atomic_reactor/build.py
```diff
@@ -183,6 +185,31 @@ def set_df_path(self, path):
         for image in dfp.parent_images:
             self.parent_images[image] = None
 
+        # validate user has not specified COPY --from=image
+        builders = []
+        for stmt in dfp.structure:
+            if stmt['instruction'] == 'FROM':
+                # extract "bar" from "foo as bar" and record as build stage
+                match = re.search(r'\S+ \s+  as  \s+ (\S+)', stmt['value'], re.I | re.X)
```
lcarva: With `re.X` the literal whitespaces are chopped off?

lcarva: Yes, they are: "Whitespace within the pattern is ignored": https://docs.python.org/3/library/re.html#re.X

Neat!

PR #1075, opened 2018-09-21
verify_media_types: take advantage of the plugin

mlangsdorf:
Add code to exit_koji_import.py and exit_store_metadata_in_osv3.py to get the media type metadata 
from verify_media_types if that information isn't available from pulp_pull.

Add a warning in input_osv3 to detect if an arrangement 6 build has neither of pulp_pull or 
verify_media_types or if it has both of them.

### DIFF
--- a/atomic_reactor/plugins/input_osv3.py
+++ b/atomic_reactor/plugins/input_osv3.py
@@ -45,13 +45,18 @@ class OSv3InputPlugin(InputPlugin):
     def get_value(self, name, default=None):
         return self.reactor_env.get(name, default)
 
-    def remove_plugin(self, phase, target_plugin, reason):
+    def find_plugin(self, phase, target_plugin):
         if phase in self.plugins_json:
             for index, plugin in enumerate(self.plugins_json[phase]):
                 if plugin['name'] == target_plugin:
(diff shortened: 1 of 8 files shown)


atomic_reactor/plugins/input_osv3.py
```diff
@@ -45,13 +45,23 @@ def get_plugins_with_user_params(self, build_json, user_params):
...
+    def find_plugin(self, phase, target_plugin):
+        if phase in self.plugins_json:
+            for index, plugin in enumerate(self.plugins_json[phase]):
+                if plugin['name'] == target_plugin:
+                    return index
+        return -1
+
     def remove_plugin(self, phase, target_plugin, reason):
+        """
         if phase in self.plugins_json:
```
lcarva: Remove this too?

twaugh: Yes, let's remove this.

mlangsdorf: Fixed.

twaugh:
What's the test failure about?

mlangsdorf:
I have no idea. I can't reproduce it locally and I didn't modify pulp_pull, test_pull_pull, or util.

lcarva:
Existing entry in release notes already covers this change.

PR #1076, opened 2018-09-24
Apply organization to destination registry

lcarva:


lcarva:
Updated release notes draft:
> - Added expected_media_types for registries, and registries_organization to reactor config schema
> - Images are pushed to repository in organization, if one is defined

PR #1077, opened 2018-09-27
still more ImageName default tag fixes

mlangsdorf:
Add some more information to build.py and change_from_in_df to make it easier to determine why a
parent image isn't found on a BaseMismatchError.

Fix exit_store_metadata to not crash if a parent image is None.

### DIFF
--- a/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
+++ b/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
@@ -225,12 +225,6 @@ class StoreMetadataInOSv3Plugin(ExitPlugin):
         except AttributeError:
             dockerfile_contents = ""
 
-        def parent_images_to_str(parent_images):
-            results = {}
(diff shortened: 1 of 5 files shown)


atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -228,7 +228,7 @@ def run(self):
         def parent_images_to_str(parent_images):
             results = {}
             for base_image_name, parent_image_name in parent_images.items():
-                results[base_image_name.to_str()] = parent_image_name.to_str()
```
lcarva: What if we check if `base_image_name` and `parent_image_name` are truthy before adding to results? I'd rather not have `None` in the annotations.

sosiouxme: None *shouldn't* get in there if the build succeeded to the point where it's storing metadata, but no harm in sanity checking.

lcarva:

PR #1079, opened 2018-10-01
verify_media_types: correctly handle removing it in input_osv3

mlangsdorf:
exit_verify_media_types should be configured for arrangment 6 orchestrator
builds when pulp or koji is not available. Adjust the test for the
warning message and remove verify_media_types on arrangement 6 orchestrator
builds that do have pulp_pull enabled.


lcarva:

PR #1080, opened 2018-10-01
Provide mechanism to disable package comparison

lcarva:
In some cases, due to odd packaging practices, certain packages are
expected to have different versions across different arches. This change
allows admins to define a list of packages that should not be compared.

### DIFF
--- a/atomic_reactor/plugins/post_compare_components.py
+++ b/atomic_reactor/plugins/post_compare_components.py
@@ -6,6 +6,7 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
 from atomic_reactor.plugin import PostBuildPlugin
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/post_compare_components.py
```diff
@@ -91,6 +94,10 @@ def run(self):
                 t = component['type']
                 name = component['name']
 
+                if name in package_comparison_exceptions:
+                    self.log.warn("Ignoring comparison of package %s", name)
```
mlangsdorf: Wouldn't info be more appropriate? atomic reactor is ignoring the package that reactor config told it to ignore. Logging it is a good idea, but it's not a potential bug.

lcarva: Yes, you're right. Updated.

lcarva:
Updated release notes draft:
> - New mechanism added to bypass component comparison based on server configuration

PR #1082, opened 2018-10-02
Adjust data in parent_image_builds in koji import

lcarva:
When importing build into Koji, all dict keys are required to be of str
type.


lcarva:

PR #1083, opened 2018-10-03
get_koji_module_build: Handle change to the "archives" that MBS writes

owtaylor:
The module build service used to store a single archive "modulemd.txt",
in koji which had all the RPMs that were built attached to it. It was changed
to write that archive, plus one archive per-architecture, with only the
relevant RPMs for that architecture (modulemd.<arch>.txt).
(https://pagure.io/fm-orchestrator/pull-request/1007)
Handle this by picking out the all-architectures version. While we actually
only need the list of RPMs for a single architecture, getting the architecture we
need to the right place in the code would require considerable plumbing,
and having a bigger list of RPMs only causes (a hopefully small) amount
processing overhead when we set include_pkgs for DNF.

### DIFF
--- a/atomic_reactor/koji_util.py
+++ b/atomic_reactor/koji_util.py
@@ -257,6 +257,10 @@ def get_koji_module_build(session, module_spec):
             raise RuntimeError("No build found for {}".format(module_spec.to_str()))
 
     archives = session.listArchives(buildID=build['build_id'])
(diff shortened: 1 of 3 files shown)


lcarva:
Updated release notes draft:
> - `get_koji_module_build` handles change to the "archives" that MBS writes

PR #1085, opened 2018-10-04
verify media types: fix bug in input_osv3

mlangsdorf:
The wrong key name was used in input_osv3 and verify_media_types wasn't being removed from the plugins configuration when pulp was configured. Fix that particular bug, and consistently use the KEY literals from atomic_reactor.constants to prevent that kind of bug from happening again in the future.


mlangsdorf:
Not adding anything to release notes as this is another bugfix for an unreleased feature.

PR #1087, opened 2018-10-04
verify media types: don't break builds if it runs when it shouldn't

mlangsdorf:
Change verify_media_types to return nothing if it detects configured
pulp_registries, instead of raising an error.

Also, fix more of the logic in input_osv3 so that verify_media_types
won't run in pulp_pull is scheduled to run.


atomic_reactor/plugins/input_osv3.py
```diff
@@ -112,9 +115,9 @@ def has_plugin(self, phase, target_plugin):
                 self.remove_plugin(phase, PLUGIN_PULP_PULL_KEY, 'no pulp or koji available')
             else:
                 has_pulp_pull |= has_plugin(self, phase, PLUGIN_PULP_PULL_KEY)
-        arrangement_six = self.plugins_json.get('arrangement_version', 0) >= 6
-        orchestrator_build = self.plugins_json.get('build_type', None) == 'orchestrator'
-        if arrangement_six and orchestrator_build:
+
+        # if we're here at all, we're on arrangment 6+
+        if self.build_type == 'orchestrator':
```
lcarva: I'd rather we remove this block completely. It's really unnecessary. Let the plugin itself decide whether or not it should run.

atomic_reactor/plugins/exit_verify_media_types.py
```diff
@@ -55,7 +55,8 @@ def run(self):
         if not self.workflow.tag_conf.unique_images:
             raise ValueError("no unique image set, impossible to verify media types")
         if self.workflow.push_conf.pulp_registries:
-            raise RuntimeError("pulp registry configure, verify_media_types should not run")
+            self.log.warn("pulp registry configure, verify_media_types should not run")
```
lcarva: I think just info is enough here.

PR #1089, opened 2018-10-10
odcs_utils: add support for multilib composes

mlangsdorf:
Pass the multilib compose arguments, if any, into odcs.


atomic_reactor/odcs_util.py
```diff
@@ -81,6 +86,10 @@ def start_compose(self, source_type, source, packages=None, sigkeys=None, arches
         if arches is not None:
             body['arches'] = arches
 
+        if multilib_arches:
+            body['multilib_arches'] = multilib_arches
+            body['multilib_method'] = multilib_method
```
lcarva: Because the default value is `None` for this method, we might want to do something like this:
```python
body['multilib_method'] = multilib_method or []
```

I think `None`, which gets converted to json's `null` is not a valid value.

lcarva: Ah good catch! So maybe we need to handle default value here instead of `resolve_composes` plugin.

mlangsdorf: Fixed.

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -22,12 +22,12 @@
                                                        get_odcs_session,
                                                        get_koji_session, get_koji)
 from atomic_reactor.util import get_platforms
+from atomic_reactor.odcs_util import MULTILIB_METHOD_DEFAULT
```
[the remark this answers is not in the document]
twaugh: Either that, or odcs_util should verify that is it not None when multilib_arches is set.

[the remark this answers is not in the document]

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -334,10 +333,7 @@ def __init__(self, data, pulp_data, odcs_config, koji_tag=None, arches=None):
             if arch in arches:
                 self.multilib_arches.append(arch)
         if self.multilib_arches:
-            if data.get('multilib_method'):
-                self.multilib_method = data.get('multilib_method')
-            else:
-                self.multilib_method = MULTILIB_METHOD_DEFAULT
+            self.multilib_method = data.get('multilib_method', [])
```
twaugh: Nit: could even be `data.get('multilib_method')` without providing an if-not-there value, because `None` is falsy just like `[]` is. In other words, the `multilib_method or MULTILIB_METHOD_DEFAULT` expression will evaluate to the same as `MULTILIB_METHOD_DEFAULT` regardless of which falsy value we set `self.multilib_method` to here.

tests/test_odcs_util.py
```diff
@@ -94,7 +94,13 @@ def compose_json(state, state_name, source_type='module', source=MODULE_NSV,
     ['no_deps'],
     ['breakfast', 'lunch'],
 ))
-def test_create_compose(odcs_client, source, source_type, packages, sigkeys, arches, flags):
+@pytest.mark.parametrize(('multilib_arches', 'multilib_method'), (
+    (None, None),
+    (['x86_64'], ['all']),
+    (['breakfast', 'lunch'], ['some', 'random'])
```
twaugh: Can we have a test that sets multilib_arches to some non-empty list but multilib_method to something falsy?

tests/test_odcs_util.py
```diff
@@ -111,14 +117,17 @@ def handle_composes_post(request):
         assert body_json['source'].get('sigkeys') == sigkeys
         assert body_json.get('flags') == flags
         assert body_json.get('arches') == arches
+        assert body_json.get('multilib_arches') == multilib_arches
+        assert body_json.get('multilib_method') == multilib_method or MULTILIB_METHOD_DEFAULT
```
twaugh: This is repeating the logic of the code under test. Would prefer to see an expected value in the parametrization.

atomic_reactor/plugins/pre_resolve_composes.py
mlangsdorf: Need the [] to avoid triggering an error on sorted() later on, and need sorted() to keep the list ordering consistent between python2 and python3 test cases.

twaugh: Can't you use sorted() on both sides of the expression in the test, e.g. `assert sorted(body_json.get('multilib_method', [])) == sorted(expected_method)`?

atomic_reactor/odcs_util.py
```diff
@@ -81,6 +87,10 @@ def start_compose(self, source_type, source, packages=None, sigkeys=None, arches
         if arches is not None:
             body['arches'] = arches
 
+        if multilib_arches:
+            body['multilib_arches'] = multilib_arches
+            body['multilib_method'] = sorted(set(multilib_method or MULTILIB_METHOD_DEFAULT))
```
twaugh: Trying to make a sorted list with unique elements?

Do we care about element uniqueness, or list ordering? I feel like we ought to just accept what the user gave us.

lcarva: The sorted is to make testing more sane, I think.

@mlangsdorf, I think if we sort just on the tests it should be enough:
```python
# tests/test_odcs_util.py ~ line 124
assert sorted(body_json.get('multilib_method')) == sorted(expected_method)
```
Just accepting what was given seems reasonable to me.

atomic_reactor/plugins/pre_resolve_composes.py
lcarva: The way it's currently written, `[]` will get replaced with `MULTILIB_METHOD_DEFAULT` anyways in `odcs_utils.py`. Using None should be fine here, assuming, of course, that we remove the sorted calls later on in this file as mentioned in my previous comment.

There's a test in `test_resolve_plugins` that will fail because it expects `multilib_method` to be set to `[]` when it's `None`. IMO this is invalid at this layer. Applying a valid default value is done in `odcs_utils.py` and it should be tested in `test_odcs_utils.py`.

atomic_reactor/odcs_util.py
lcarva: There are a couple of places in `test_resolve_composes` that this should be done too. That way we can also remove the sorting for both `multilib_arches` and `multilib_method` in `pre_resolve_composes` plugin.

mlangsdorf: Fixed.

atomic_reactor/plugins/pre_resolve_composes.py

lcarva:
Let's wait for @twaugh to give some final feedback since he had quite a few comments in this PR.

atomic_reactor/odcs_util.py
```diff
@@ -60,6 +61,11 @@ def start_compose(self, source_type, source, packages=None, sigkeys=None, arches
                         these keys will be included in a compose.
         :param arches: list<str>, List of additional Koji arches to build this compose for.
                         By default, the compose is built only for "x86_64" arch.
+        :param multilib_arches: list<str>, List of Koji arches to build as multilib in this
+                        compose. By default, no arches are built as multilib.
+        :param multilib_method: list<str>, list of methods to determine which packages should
+                        be included in a multilib compose. By default, is none and must be
+                        set to non-default if multilib_arches is not empty.
```
twaugh: Is the wording about "must be set to non-default" still accurate?

mlangsdorf: Fixed.

PR #1091, opened 2018-10-12
jsonschema containers.json: allow empty sections

MartinBasti:
- [x] PoC with "platforms" attribute
- [x] Allow other attributes (sections) to be empty
- [x] Check if other attributes are handled well when set to `None`
- [x] Tests

Hello,
I've created a PoC implementation. Please check if I placed tests to the proper location, if tests are written in expected form, if I missed anything, please check commit messages, code style, etc.... I want to fix my newbie errors as soon as possible :-)

Also it looks that other attributes can be used as empty sections, I'd go ahead and fix it everywhere in `containers.json`, is that OK?

Martin

### DIFF
--- a/atomic_reactor/source.py
+++ b/atomic_reactor/source.py
@@ -37,13 +37,18 @@ class SourceConfig(object):
 
     def __init__(self, build_path):
         self.data = {}
-        file_path = os.path.join(build_path, REPO_CONTAINER_CONFIG)
-        if os.path.exists(file_path):
+        self.file_path = os.path.join(build_path, REPO_CONTAINER_CONFIG)
+        if os.path.exists(self.file_path):
             try:
                 # read file and validate against schema
-                self.data = read_yaml_from_file_path(file_path, 'schemas/container.json') or {}
+                self.data = read_yaml_from_file_path(
+                    self.file_path, 'schemas/container.json'
+                ) or {}
             except Exception:
-                logger.exception("Failed to load and validate source config YAML from " + file_path)
+                logger.exception(
+                    "Failed to load and validate source config YAML from %s",
+                    self.file_path
+                )
                 raise
 
         self.autorebuild = self.data.get('autorebuild') or {}
--- a/atomic_reactor/source.py
+++ b/atomic_reactor/source.py
@@ -46,7 +46,7 @@ class SourceConfig(object):
                 logger.exception("Failed to load and validate source config YAML from " + file_path)
                 raise
 
-        self.autorebuild = self.data.get('autorebuild', {})
+        self.autorebuild = self.data.get('autorebuild') or {}
         self.flatpak = self.data.get('flatpak')
         self.compose = self.data.get('compose')
         self.image_build_method = meth = self.data.get('image_build_method')
--- a/tests/test_source.py
+++ b/tests/test_source.py
@@ -146,6 +146,11 @@ class TestSourceConfigSchemaValidation(object):
                - s390x
             """,
             {'data': {'platforms': {'only': ['s390x'], 'not': ['s390x']}}}
+        ), (
+            """\
+            platforms:
+            """,
+            {'data': {'platforms': None}}
         ),
     ])
     def test_valid_source_config(self, tmpdir, yml_config, attrs_updated):
@@ -158,10 +163,6 @@ class TestSourceConfigSchemaValidation(object):
             assert getattr(source_config, attr_name) == value
 
     @pytest.mark.parametrize('yml_config', [
-        """\
-        platforms:
-        """,
-
         """\
         platforms: not_an_object
         """,
--- a/tests/test_source.py
+++ b/tests/test_source.py
@@ -1,8 +1,17 @@
+import copy
 import os
 
 import pytest
+import jsonschema
 
-from atomic_reactor.source import (Source, GitSource, PathSource, get_source_instance_for)
+from atomic_reactor.constants import REPO_CONTAINER_CONFIG
+from atomic_reactor.source import (
+    Source,
+    SourceConfig,
+    GitSource,
+    PathSource,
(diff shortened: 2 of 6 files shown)


tests/test_source.py
```diff
@@ -79,3 +87,58 @@ def test_broken_source_config_file(self):
...
+        platforms:
+            only: s390x
+        """,
+
+        """\
+        platforms:
+            not: s390x
+        """,
+
+        # TODO: schema probably should not allow to use both keywords
```
twaugh: Actually the schema deliberately allows this, but it should document that "only" is processed before "not". The reason is that this allows a temporary exclusion of an architecture, e.g.:
```
only: abc def ghi jkl mno
# not yet working on ghi, bug #.....
not: ghi
```

twaugh:
> Also it looks that other attributes can be used as empty sections, I'd go ahead and fix it everywhere in containers.json, is that OK?

Yes, we ought to be able to handle sections whose keys are all commented-out.

tests/test_source.py
MartinBasti: Thanks for clarification, TODO removed.

Test for `get_platforms_in_limits` added

MartinBasti:
Do we want also support for:

```
platforms:
   only:
   # - x86_64
   not: ppc64le
```

it needs some changes in `get_platforms_in_limits` ?

twaugh:
Agreed, I think we should not allow that.

MartinBasti:
I did what was requested in ticket, but so far it looks that it is not enough. If I use config values that violates schema, instead of schema error I got weird error (see traceback). It looks that config is not validated early enough by schema. (BTW: I have unit test for that particular case and schema works as expected)

```
cat container.yaml
---
platforms: not_an_object
```

orchestrator.log
```
2018-10-15 16:57:22,238 platform:- - atomic_reactor.plugin - DEBUG - Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/atomic_reactor/plugin.py", line 239, in run
  File "/usr/lib/python2.7/site-packages/atomic_reactor/plugins/pre_check_and_set_platforms.py", line 85, in run
    return get_platforms_in_limits(self.workflow, enabled_platforms)
  File "/usr/lib/python2.7/site-packages/atomic_reactor/util.py", line 1525, in get_platforms_in_limits
    excluded_platforms = set(make_list(data['platforms'].get('not', [])))
AttributeError: 'str' object has no attribute 'get'
2018-10-15 16:57:22,238 platform:- - atomic_reactor.plugin - ERROR - plugin 'check_and_set_platforms' raised an exception: AttributeError("'str' object has no attribute 'get'",)

-----

2018-10-15 16:57:22,344 platform:- - atomic_reactor.plugin - DEBUG - plugin 'remove_worker_metadata' finished in 0s
  File "/usr/bin/atomic-reactor", line 11, in <module>
    load_entry_point('atomic-reactor==1.6.35', 'console_scripts', 'atomic-reactor')()
  File "/usr/lib/python2.7/site-packages/atomic_reactor/cli/main.py", line 309, in run
    cli.run()
  File "/usr/lib/python2.7/site-packages/atomic_reactor/cli/main.py", line 292, in run
    args.func(args)
  File "/usr/lib/python2.7/site-packages/atomic_reactor/cli/main.py", line 95, in cli_inside_build
    substitutions=args.substitute)
  File "/usr/lib/python2.7/site-packages/atomic_reactor/inner.py", line 546, in build_inside
    build_result = dbw.build_docker_image()
  File "/usr/lib/python2.7/site-packages/atomic_reactor/inner.py", line 429, in build_docker_image
    prebuild_runner.run()
  File "/usr/lib/python2.7/site-packages/atomic_reactor/plugin.py", line 279, in run
    raise PluginFailedException(msg)
atomic_reactor.plugin.PluginFailedException: plugin 'check_and_set_platforms' raised an exception: AttributeError("'str' object has no attribute 'get'",)
```

The reason is that function `get_platforms_in_limits` loads `container.yaml` config without any validation. IMO this is not a correct behavior, there should be a single object responsible for parsing and loading container config (I assumed that this is `SourceConfig`, the only object that uses the particular schema validation).
Currently I don't know exactly how to glue it together, i.e. having only `SourceConfig` for container config data. I can use it as instance directly in `get_platforms_in_limits`, but it doesn't look like the best solution to me. Config files should be validated as early as possible.

MartinBasti:
After discussion with @rcerven this may be a solution to enforce validation
```diff
-    build_file_dir = workflow.source.get_build_file_path()[1]
-    container_path = os.path.join(build_file_dir, REPO_CONTAINER_CONFIG)
-    if os.path.exists(container_path):
-        with open(container_path) as f:
-            data = yaml.safe_load(f)
+  data = workflow.source.config
```

MartinBasti:
Ready for review :tada:

atomic_reactor/source.py
```diff
@@ -37,16 +37,21 @@ class SourceConfig(object):
...
             except Exception:
-                logger.exception("Failed to load and validate source config YAML from " + file_path)
+                logger.exception(
+                    "Failed to load and validate source config YAML from %s",
+                    self.file_path
+                )
                 raise
 
-        self.autorebuild = self.data.get('autorebuild', {})
+        self.autorebuild = self.data.get('autorebuild') or {}
```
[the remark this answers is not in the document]
MartinBasti: If autorebuild section in yaml is empty: 
```
---
autorebuild:
```
then the key `autorebuild` is defined with value `None` and `.get('autorebuild', {})` returns `None` instead `{}`. Functions later expect dict and they will fail with `None`.

```python
>>> print(dict(autorebuild=None).get('autorebuild', {}))
None
```

lcarva:
Updated release notes draft:
> - empty sections allowed in `container.yaml`

PR #1092, opened 2018-10-15
simplify import_image

rcerven:

### DIFF
--- a/atomic_reactor/plugins/post_import_image.py
+++ b/atomic_reactor/plugins/post_import_image.py
@@ -80,12 +80,14 @@ class ImportImagePlugin(ExitPlugin, PostBuildPlugin):
 
         self.osbs = get_openshift_session(self.workflow, self.openshift_fallback)
         self.get_or_create_imagestream()
-        self.process_tags()
+
         try:
+            self.osbs.import_image_tags(self.imagestream_name, self.get_trackable_tags(),
+                                        self.docker_image_repo, insecure=self.insecure_registry)
+        except AttributeError:
+            self.log.info('Falling back to calling import_image instead of import_image_tags')
+            self.process_tags()
             self.osbs.import_image(self.imagestream_name, tags=self.get_trackable_tags())
-        except TypeError:
-            self.log.info('Falling back to calling import_image without tags')
-            self.osbs.import_image(self.imagestream_name)
 
     def get_or_create_imagestream(self):
         try:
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/post_import_image.py
```diff
@@ -80,12 +80,14 @@ def run(self):
 
         self.osbs = get_openshift_session(self.workflow, self.openshift_fallback)
         self.get_or_create_imagestream()
-        self.process_tags()
+
         try:
-            self.osbs.import_image(self.imagestream_name, tags=self.get_trackable_tags())
+            self.osbs.import_image_tags(self.imagestream_name, self.get_trackable_tags(),
+                                        self.docker_image_repo, insecure=self.insecure_registry)
```
mlangsdorf: Why aren't you passing the source_registry information into osbs-client here, since atomic_reactor has the correct reactor_config and can easily access it?

rcerven: I am, with 'self.docker_image_repo'

atomic_reactor/plugins/post_import_image.py
```diff
@@ -80,12 +80,14 @@ def run(self):
...
         except TypeError:
```
lcarva: I think it would be an `AttruibuteError` if an older version of osbs-client is used.
It was `TypeError` because we were calling the same method but with newer parameters.

atomic_reactor/plugins/post_import_image.py
```diff
@@ -80,12 +80,14 @@ def run(self):
...
-            self.log.info('Falling back to calling import_image without tags')
-            self.osbs.import_image(self.imagestream_name)
+            self.log.info('Falling back to calling import_image instead of import_image_tags')
+            self.process_tags()
+            self.osbs.import_image(self.imagestream_name, tags=self.get_trackable_tags())
```
lcarva: Coveralls reporting is missing for some reason, but I don't think we're testing this whole block.

atomic_reactor/plugins/post_import_image.py
lcarva: Hmm I see. Would it be simpler overall to have two parameters? One for the registry and one for the repo?

rcerven:
rn updated:
import_image plugin uses now import_image_tags, which imports image and tags at once, instead of importing tags one by one, and import image after

PR #1093, opened 2018-10-16
Default to koji defined target for image tasks

athos-ribeiro:
By using the target forwarded from the koji task, one can omit it from
the image-build.conf, allowing the usage of different targets without
the need to perform changes in the configuration file.

### DIFF
--- a/docs/base_images.md
+++ b/docs/base_images.md
@@ -37,8 +37,13 @@ file used for Koji's CLI client image build subcommand:
 For convenience, defaults are provided to minimize the size of the configuration
 file. See the `DEFAULT_IMAGE_BUILD_CONF` variable in `add_filesystem` plugin's
 [source](/atomic_reactor/plugins/pre_add_filesystem.py) for all defaults.
-Most notably, the `ksurl` item is by default dynamically mapped to the current
-branch and commit.
+Most notably,
+
+- the `ksurl` item is by default dynamically mapped to the current branch and
+  commit;
+- the `target` item is by default set to the koji target parameter and should
+  be omitted from the config file. If explicitly set in the file, the default
+  value is overridden.
 
 It's important to note that currently in the `[factory-parameters]` section,
 the `create_docker_metadata` parameter must be set to `False`. Setting this
--- a/tests/plugins/test_add_filesystem.py
+++ b/tests/plugins/test_add_filesystem.py
@@ -50,6 +50,7 @@ if MOCK:
     from tests.retry_mock import mock_get_retry_session
 
 KOJI_HUB = 'https://koji-hub.com'
+KOJI_TARGET = 'guest-fedora-23-docker'
 FILESYSTEM_TASK_ID = 1234567
 
 DEFAULT_DOCKERFILE = dedent("""\
@@ -157,7 +158,6 @@ def mock_image_build_file(tmpdir, contents=None):
             [image-build]
             name = fedora-23
             version = 1.0
-            target = guest-fedora-23-docker
             install_tree = http://install-tree.com/$arch/fedora23/
 
             format = docker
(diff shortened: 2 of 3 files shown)


tests/plugins/test_add_filesystem.py
```diff
@@ -496,7 +494,7 @@ def test_image_build_overwrites(tmpdir, architectures, architecture, reactor_con
         name = my-name
```
lcarva: I'd like to know which take precedence: 1) target from build param, or 2) target from image-build.conf. I think the way it's coded, target from image-build.conf will win. This is fine, it just means that after we get these changes in, we should work with internal RCM to clean up their image-build.conf; not a problem, safer actually.

Let's just add a test here to verify the behavior. Above, when `create_plugin_instance` is called, let's call it with a dummy target to make sure it's indeed ignored.

athos-ribeiro: Thanks for the review!
You are right. The target from `image-buiild.conf` overwrites the target from the build param. This way we will not break/change existing setups with the change and users would have to manually change their image-build.conf to use the feature.

About the test, on line 214 we now have:

`return AddFilesystemPlugin(tasker, workflow, KOJI_HUB, koji_target=KOJI_TARGET, **kwargs)`

So all calls to `create_plugin_instance` already set the target as defined by KOJI_TARGET and it is indeed changed from docker-f23 to docker-f23-candidate in that test. Should I still change the `create_plugin_instance` to explicitly receive the target?

athos-ribeiro:
I also added a note in the docs to include these changes

docs/base_images.md
```diff
@@ -37,8 +37,13 @@ file used for Koji's CLI client image build subcommand:
...
 file. See the `DEFAULT_IMAGE_BUILD_CONF` variable in `add_filesystem` plugin's
 [source](/atomic_reactor/plugins/pre_add_filesystem.py) for all defaults.
-Most notably, the `ksurl` item is by default dynamically mapped to the current
-branch and commit.
+Most notably,
+
+- the `ksurl` item is by default dynamically mapped to the current branch and
+  commit;
+- the `target` item is by default set to the koji target parameter and should
+  be ommited from the config file. If explicitly set in the file, the default
```
twaugh: Typo: should be 'omitted'

tests/plugins/test_add_filesystem.py
lcarva: What happens if koji_target is not given? Will the target from image-build.conf get overridden with None?

athos-ribeiro: It is the opposite: the target value in the `image-build.conf`, when given, overwrites the koji target. If the koji_target is provided, it is always used.

If neither values are provided, we fail after parsing the config file, which was the behavior when no target was specified in `image-build.conf`.

If no koji_target is provided, but a target is specified in `image-build.conf`, the value in the file is used.

I added a few more tests for these cases.

tests/plugins/test_add_filesystem.py
```diff
@@ -620,3 +618,45 @@ def test_image_download(tmpdir, docker_tasker, architecture, architectures, down
     else:
         assert plugin_result['base-image-id'] is None
         assert plugin_result['filesystem-koji-task-id'] is None
+
+
+@pytest.mark.parametrize(('koji_target'), [None, '', 'guest-fedora-23-docker'])
```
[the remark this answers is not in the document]
twaugh: We usually add `  # noqa:F811` to the end of this line to avoid that diagnostic.

@lcarva did you say there was a better way?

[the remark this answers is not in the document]
lcarva: What @MartinBasti said :)
I think it would be a nice way to clean up our testing, but a much bigger initial change, of course.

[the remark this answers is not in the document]
athos-ribeiro: Proof of concept for conftest.py change: https://github.com/athos-ribeiro/atomic-reactor/pull/1

lcarva:
Updated release notes draft:
> - koji target build parameter is used as default value for createImage tasks

PR #1094, opened 2018-10-16
Improve component mismatch reporting

MartinBasti:
Jira: OSBS-5880

### DIFF
--- a/atomic_reactor/plugins/post_compare_components.py
+++ b/atomic_reactor/plugins/post_compare_components.py
@@ -5,13 +5,24 @@ All rights reserved.
 This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
+import logging
+
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.plugins.pre_reactor_config import get_package_comparison_exceptions
 from atomic_reactor.constants import (PLUGIN_COMPARE_COMPONENTS_KEY,
                                       PLUGIN_FETCH_WORKER_METADATA_KEY)
 
 
-SUPPORTED_TYPES = ("rpm",)
+T_RPM = "rpm"
+SUPPORTED_TYPES = (T_RPM,)
+
+
+def filter_components_by_name(name, components_list, type_=T_RPM):
+    """Generator filters components from components_list by name"""
+    for components in components_list:
+        for component in components:
+            if component['type'] == type_ and component['name'] == name:
+                yield component
 
 
 class CompareComponentsPlugin(PostBuildPlugin):
@@ -63,6 +74,15 @@ class CompareComponentsPlugin(PostBuildPlugin):
 
         return comp_list
 
+    def log_rpm_component(self, component, loglevel=logging.WARNING):
+        assert component['type'] == T_RPM
+        self.log.log(
+            loglevel,
+            "%s: %s-%s-%s (%s)",  # platform: name-version-release (signature)
+            component['arch'], component['name'], component['version'],
+            component['release'], component['signature']
+        )
+
     def run(self):
         """
         Run the plugin.
@@ -106,14 +126,19 @@ class CompareComponentsPlugin(PostBuildPlugin):
                     master_comp[identifier] = component
                     continue
 
-                try:
+                if t == T_RPM:
                     mc = master_comp[identifier]
-
-                    if t == 'rpm':
+                    try:
                         self.rpm_compare(mc, component)
-                except ValueError as ex:
-                    self.log.warn("Comparison mismatch for component %s: %s", name, ex)
-                    failed = True
+                    except ValueError as ex:
+                        self.log.debug("Mismatch details: %s", ex)
+                        self.log.warning(
+                            "Comparison mismatch for component %s:", name)
+
+                        # use all components to provide complete list
+                        for comp in filter_components_by_name(name, comp_list):
+                            self.log_rpm_component(comp)
+                        failed = True
 
         if failed:
             raise ValueError("Failed component comparison")
--- a/atomic_reactor/plugins/post_compare_components.py
+++ b/atomic_reactor/plugins/post_compare_components.py
@@ -108,7 +108,7 @@ class CompareComponentsPlugin(PostBuildPlugin):
         # list to compare with.
 
         # Keep everything separated by component type
-        failed = False
+        failed_components = set()
         for components in comp_list:
             for component in components:
                 t = component['type']
@@ -138,7 +138,12 @@ class CompareComponentsPlugin(PostBuildPlugin):
                         # use all components to provide complete list
                         for comp in filter_components_by_name(name, comp_list):
                             self.log_rpm_component(comp)
-                        failed = True
+                        failed_components.add(name)
 
-        if failed:
-            raise ValueError("Failed component comparison")
+        if failed_components:
+            raise ValueError(
+                "Failed component comparison for components: "
+                "{components}".format(
+                    components=', '.join(sorted(failed_components))
+                )
+            )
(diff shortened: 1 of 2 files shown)


tests/plugins/test_compare_components.py
```diff
@@ -21,6 +21,9 @@
 from atomic_reactor.plugins.pre_reactor_config import (ReactorConfigPlugin,
                                                        WORKSPACE_CONF_KEY,
                                                        ReactorConfig)
+from atomic_reactor.plugins.post_compare_components import (
+    filter_components_by_name
+)
```
mlangsdorf: single line, no parenthesis needed.

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -5,13 +5,24 @@
 This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
+import logging
```
mlangsdorf: Unnecessary since you're calling the log object in self.

MartinBasti: I used logging.WARNING in method as optional paramter

tests/plugins/test_compare_components.py
MartinBasti: I did it to pass PEP8 check due line lenght. I like more parenthesis than escaping. I can change it if escaping if it fits more current codestyle

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -106,14 +126,19 @@ def run(self):
...
-                except ValueError as ex:
-                    self.log.warn("Comparison mismatch for component %s: %s", name, ex)
-                    failed = True
+                    except ValueError as ex:
+                        self.log.debug("Mismatch details: %s", ex)
+                        self.log.warning(
+                            "Comparison mismatch for component %s:", name)
+
+                        # use all components to provide complete list
+                        for comp in filter_components_by_name(name, comp_list):
```
lcarva: I think it's a bit inefficient to loop through the list again. Can we just log `mc` and `component`?

The warning will be repeated if more than two arches are used regardless.

lcarva: Alternatively, you could maintain a list of all the failures and only print warning once at the end. You could also use the component names when raising "Failed component comparison" exception to make things clearer to users, since that's what they'll see as the failure. Everything else is buried in logs.

```python
mismatches = defaultdict(list)

# rpm_compare return bool instead of raising exception
if self.rpm_compare(mc, component):
  dict[identifier].append(component)

...

# Report warnings here once
for component_name, components in mismatches.items():
  ...

if mismatches:
  raise ValueError('Failed component comparison for components: {}'.format(mismatches.keys()))

```

MartinBasti: How often does this happen? I'd not do premature optimization, I don't think that searching a few lists and dicts may be somehow serious.

I wanted (and got agreement in ticket) to provide complete list of all architectures. This may not happen when I compare only `mc` and `component` in order: `good1, good2, bad` and it is also less user friendly.

I will incorporate your idea of reporting only once into patches. Great idea! Shame on me I didin't do that. But it will be different because as I mentioned previously, I cannot rely only on comparison `mc` and `component`.

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -63,6 +74,15 @@ def get_component_list_from_workers(self, worker_metadatas):
 
         return comp_list
 
+    def log_rpm_component(self, component, loglevel=logging.WARNING):
+        assert component['type'] == T_RPM
+        self.log.log(
+            loglevel,
+            "%s: %s.%s-%s (%s)",  # platform: name.version-release (signature)
```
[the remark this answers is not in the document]
MartinBasti: `n-v-r` I will fix that. In examples was `n.v-r` but Tim requested `n-v-r`

atomic_reactor/plugins/post_compare_components.py
MartinBasti: BTW: I cannot use sets, dict is mutable. (May work in py2, but it is still bad in way how is that implemented)

lcarva: Slight misunderstanding. The idea is that mismatches would look something like this:
```python
mismatches = {
  (rpm, libquadmath): [
    {'name': 'libquadmath', 'version': '1.2', 'arch': 'x86_64', ...},
    {'name': 'libquadmath', 'version': '1.2', 'arch': 'ppc64le', ...},
    {'name': 'libquadmath', 'version': '1.2', 'arch': 's390x', ...},
  ]
}
```
In other words, it would include *all* the mismatches. Including `mc` and `component`. The comment about `mc` in snippet is that you may want to use something like this:
```python
mismatches = defaultdict(set)
...
if self.rpm_compare(mc, component):
  mismatches[identifier].add(component)
  mismatches[identifier].add(mc)
```
Using a set would ensure that `mc` doesn't end up in there twice :)

EDIT: Since dict can be added to a set, you could always turn it into a string since that's all you need at this point anyways.

Just an idea. Feel free to ignore it :)

MartinBasti: I got your point, but it won't work reliably all the time (or I'm wrong, correct me)

Let's assume that we have following packages, checked in order:
```
(1) x86_64: libquadmath-1.2-abc (987654321)
(2) ppc64le: libquadmath-1.2-abc (987654321)
(3) s390x: libquadmath-1.3.1415-abc (987654321)  # mismatch
```

In the current implementation, first entry is permanently stored as `mc`
```
store mc = (1)
compare (mc) vs (2) --> ok
compare (mc) vs (3) --> mismatch
```
so (2) will be never added into `mismatches` dict. There is no comparison `(2) vs (3)`. This will lead into incomplete list of components in the log output.

I can do mapping `{name: [all_components], ...}` but creating/updating dicts and lists is less efficient in memory and speed, than simple iteration over yielded values (in case that mismatches are for few entries only and rare).

atomic_reactor/plugins/post_compare_components.py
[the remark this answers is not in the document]
MartinBasti: Fixed

tests/plugins/test_compare_components.py
```diff
@@ -175,3 +197,46 @@ def test_bad_component_type(tmpdir):
 
     with pytest.raises(PluginFailedException):
         runner.run()
+
+
+@pytest.mark.parametrize('mismatch', (True, False))
+def test_mismatch_reporting(tmpdir, caplog, mismatch):
+    """Test if expected log entries are reported when compoentns mismatch"""
```
twaugh: Typo: components

atomic_reactor/plugins/post_compare_components.py
```diff
@@ -5,13 +5,24 @@
...
                                       PLUGIN_FETCH_WORKER_METADATA_KEY)
 
 
-SUPPORTED_TYPES = ("rpm",)
+T_RPM = "rpm"
+SUPPORTED_TYPES = (T_RPM,)
+
+
+def filter_components_by_name(name, comp_list, type_=T_RPM):
+    """Generator filters components from coponents_list by name"""
```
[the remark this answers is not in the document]
MartinBasti: I'll use `components_list` if you agree.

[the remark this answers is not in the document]
twaugh: I prefer `components_list` for that parameter name, yes.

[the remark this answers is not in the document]
MartinBasti: Fixed

tests/plugins/test_compare_components.py
MartinBasti: Fixed.

I've changed settings in my IDE to mark typos in way I couldn't ignore :-)

MartinBasti:
@twaugh Done :-)

lcarva:
Updated release notes draft:
> - Components mismatch are reported more clearly

PR #1095, opened 2018-10-16
yum repo utils: detect if the repourl already has an md5sum

mlangsdorf:
Don't append md5sums to repourls that already have md5sums.

### DIFF
--- a/atomic_reactor/plugins/pre_add_yum_repo.py
+++ b/atomic_reactor/plugins/pre_add_yum_repo.py
@@ -33,7 +33,8 @@ class AddYumRepoPlugin(PreBuildPlugin):
         """
         # call parent constructor
         super(AddYumRepoPlugin, self).__init__(tasker, workflow)
-        self.repo = YumRepo(baseurl=baseurl, name=repo_name, enabled=True)
+        self.repo_name = repo_name
+        self.baseurl = baseurl
 
     def run(self):
         """
--- a/atomic_reactor/yum_util.py
+++ b/atomic_reactor/yum_util.py
@@ -35,9 +35,10 @@ REPO_SUFFIX = ".repo"
 
 
 class YumRepo(object):
-    def __init__(self, repourl, content='', dst_repos_dir=YUM_REPOS_DIR):
+    def __init__(self, repourl, content='', dst_repos_dir=YUM_REPOS_DIR, add_hash=True):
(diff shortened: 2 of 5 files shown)


atomic_reactor/yum_util.py
```diff
@@ -54,7 +54,12 @@ def filename(self):
         '''
         urlpath = unquote(urlsplit(self.repourl, allow_fragments=False).path)
         basename = os.path.basename(urlpath)
-        suffix = '-' + md5(self.repourl.encode('utf-8')).hexdigest()[:5]
+        poss_name = self.repourl[:-11] + REPO_SUFFIX
+        # check if this name is already encoded
+        if md5(poss_name.encode('utf-8')).hexdigest()[:5] == self.repourl[-10:-5]:
```
lcarva: Do we know what's causing it to be encoded twice?

twaugh:
We shouldn't have to be guessing what's in the filename. This looks like, to avoid problems when hashing twice, we hash a third time. Can we avoid the second hash instead?

mlangsdorf:
if I'm following the code:
* pre_add_yum_repo_by_url, pre_koji, and pre_flatpak_create_dockerfile all add yum repos to workflow.files.
* pre_inject_yum_repo consumes workflow.files by writing the repos' content out to a local directory and adjusting the Dockerfile.

I could add a flag to YumRepo that controls the hash, so pre_inject_yum_repo wouldn't rehash.

atomic_reactor/yum_util.py
mlangsdorf: another approach

atomic_reactor/plugins/pre_add_yum_repo.py
```diff
@@ -33,7 +33,8 @@ def __init__(self, tasker, workflow, repo_name, baseurl):
         """
         # call parent constructor
         super(AddYumRepoPlugin, self).__init__(tasker, workflow)
-        self.repo = YumRepo(baseurl=baseurl, name=repo_name, enabled=True)
+        self.repo_name = repo_name
```
rcerven: those 2 new lines are missing coverage

lcarva: This whole class is not tested :disappointed: 
Unrelated to this change though

lcarva:

PR #1096, opened 2018-10-19
Fix CI build F27-py3

MartinBasti:
Package *biscuits* from pip3 requires gcc+devel libs for successfull
build. This package is required only for python3.

It is dependency of responses package, but F27 provides older version
than is required (responses>=0.9.0), thus pip must build it locally.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -66,6 +66,15 @@ fi
 
 # Install package
 $RUN $PKG install -y $PIP_PKG
+
+# fix biscuits installation (dependency of responses)
+# pip installation requires newer version of responses than provided by F27
+# so it is build by pip locally and needs gcc+devel packages to build
+# successfully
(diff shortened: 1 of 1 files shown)


MartinBasti:
CI failures started yesterday (2018-10-18) because biscuits has been added as new dependency in 0.10 version.

https://pypi.org/project/responses/#history

athos-ribeiro:
Should we query the responses version instead of the distro version (this may be useful whenever RHEL 8 is out, depending on requests version there)?

Note that [Fedora 29 will be out by the end of the month](https://fedoraproject.org/wiki/Releases/29/Schedule). Another month after that, [f27 wil be at EOL](https://fedoraproject.org/wiki/Fedora_Release_Life_Cycle).

MartinBasti:
F28+ has requests version > 0.9, so it will work there, we may however freeze pip version to 0.9.x to be consistent

PR #1097, opened 2018-10-19
Move fixtures to conftest.py

athos-ribeiro:
To avoid PyFlakes F811 warnings regarding redefinition of unused names
when importing and using fixtures, we move our fixtures module to
conftest.py. See
https://docs.pytest.org/en/latest/fixture.html#conftest-py and the
discussion in #1093 for reference.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -22,7 +22,7 @@ from atomic_reactor.core import DockerTasker
 from atomic_reactor.plugin import InputPluginsRunner
 import atomic_reactor.cli.main
 
-from tests.fixtures import is_registry_running, temp_image_name, get_uuid  # noqa
+from tests.util import uuid_value
 from tests.constants import LOCALHOST_REGISTRY, DOCKERFILE_GIT, DOCKERFILE_OK_PATH, FILES, MOCK
 
 if MOCK:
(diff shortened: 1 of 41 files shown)


athos-ribeiro:
Builds fail due to issue pointed in #1096. Will rebase after it is merged.

tests/test_cli.py
```diff
@@ -44,6 +44,11 @@
     ('path', DOCKERFILE_OK_PATH),
 ])
 
+
+def get_uuid():
```
lcarva: Can we turn this into a proper fixture in conftest.py?
```python
@pytest.fixture()      
def uuid_value():                                                    
    return uuid.uuid4().hex                                        
                                                                   
                                                                   
@pytest.fixture()                                                  
def temp_image_name(uuid_value):                                             
    return ImageName(repo=("atomic-reactor-tests-%s" % uuid_value))
```

That way we don't need to redefine it here.

lcarva:
@athos-ribeiro, there are quite a few `# noqa` comments left. Can those be removed?
```
🐚 git grep -w noqa tests | wc -l
213
```
Some of it are unrelated to fixture redefinition though.

lcarva:
I think the coverage drop is because for some reason coveralls didn't pick up the python 3 test results.

mlangsdorf:
I am completely in favor of never having to write   # noqa:F811 again and needing to guess which decorator line that stickler-ci think needs it.

tests/test_cli.py
athos-ribeiro: Since we are only using it in the other fixture and in a setup_method, maybe we should either move it to `tests/util.py` or just go ahead and make the calls to uuid directly instead of having a function?

athos-ribeiro:
@lcarva, the other `# noqa` comments are not related to this issue. There are a few `# noqa: F811` left, but then I realized that the space after the colon makes it ignore everything, not only F811 warnings.

tests/test_cli.py
lcarva: WFM

athos-ribeiro: Done

MartinBasti:
Tests only, update of release notes is not needed

PR #1099, opened 2018-10-19
Enable pylint checks

MartinBasti:
All currently failing checks were disabled until they will be fixed, see `.pylintrc` for details.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -6,6 +6,9 @@ jobs=1
 # Pickle collected data for later comparisons.
 persistent=no
 
+# Activate the evaluation score.
+score=no
+
 [MESSAGES CONTROL]
 
 # Only show warnings with the listed confidence levels. Leave empty to show
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,101 @@
+[MASTER]
+
+# Use multiple processes to speed up Pylint.
+jobs=1
+
+# Pickle collected data for later comparisons.
+persistent=no
+
+[MESSAGES CONTROL]
+
+# Only show warnings with the listed confidence levels. Leave empty to show
+# all. Valid levels: HIGH, INFERENCE, INFERENCE_FAILURE, UNDEFINED
+confidence=
+
+# Disable the message, report, category or checker with the given id(s). You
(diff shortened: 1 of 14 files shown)


.pylintrc
```diff
@@ -0,0 +1,101 @@
...
+# --enable=similarities". If you want to run only the classes checker, but have
+# no Warning level messages displayed, use"--disable=all --enable=classes
+# --disable=W"
+enable=all,
+       python3
+disable=I,
+        C,  # code style checks
+        R,  # refactoring checks
+        abstract-method,  # nice to have
+        anomalous-backslash-in-string,  # should be enabled
```
lcarva: Is the goal to eventually change the code so we can enable some of these checks?

tests/plugins/test_pulp.py
```diff
@@ -38,6 +38,11 @@
     from tests.docker_mock import mock_docker
 
 
+# py2/3 compatibility
+if sys.version_info.major > 2:
+    unicode = str
```
lcarva: Can we use python six here?

integration-tests/test_pulp.py
```diff
@@ -28,7 +28,10 @@
 
 from atomic_reactor.inner import DockerBuildWorkflow
 from atomic_reactor.util import ImageName
+# TODO: PulpUploader doesn't exist, remove or fix this test?
```
lcarva: Oh my... we should probably just get rid of this folder completely. @twaugh ?

.pylintrc
MartinBasti: It may be long term goal. Now goal is to enable pylint to not introduce new type of errors

MartinBasti: At least we have to fix py2/3 compatibility errors

tests/plugins/test_pulp.py
MartinBasti: Done

integration-tests/test_pulp.py
twaugh: Yes, let's remove it.

lcarva: Feel free to open another PR to remove it.

lcarva:

PR #1100, opened 2018-10-19
Travis: add install section

MartinBasti:
Without install section, travis do default `pip install -r
requirements.txt` and we don't need it. It only slowdowns builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,10 @@ env:
     - OS=fedora
       OS_VERSION=28
       PYTHON_VERSION=3
+install:
+  - pip install coveralls
 script:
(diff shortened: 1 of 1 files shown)


lcarva:
Any benefit in making sure that by requirements.txt actually contains valid entries?

MartinBasti:
Not really, everything is tested in the container created by `test.sh`. So you don't find if anything is missing. You can basically check only if syntax is valid for pip and packages exists in pypi.

MartinBasti:
We can run it in container, it should't download new packages and if does, it means that specfile is incomplete, sounds it good?

lcarva:

PR #1105, opened 2018-10-23
WIP: better imagebuild config defaults

MartinBasti:


MartinBasti:
1. I've fixed some unrelated issues, should I open a new ticket and create a new PR, or can it stay here?

2. Please check the code that resolves this issue. I'm not sure if I used the right tools, and I'm also unsure about setting the `release` config key.

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -168,6 +169,27 @@ def get_default_image_build_conf(self):
...
+        Updates:
+          * set "name" from LABEL com.redhat.component (if exists)
+          * set "version" from LABEL version (if exists)
+          * set "release" from LABEL release (if exists)
+        """
+        labels = Labels(df_parser(self.workflow.builder.df_path).labels)
+        for config_key, label in (
+            ('name', Labels.LABEL_TYPE_COMPONENT),
+            ('version', Labels.LABEL_TYPE_VERSION),
+            ('release', Labels.LABEL_TYPE_RELEASE),  # TODO: does this key exist in config file?
```
lcarva: It may, but it likely won't. Hard coding the release value would require a new commit for every new build. When release value is not defined, the createImage task will query Koji for the next release value that should be used. This is similar to what atomic-reactor does via bump_release plugin.

However! I think we should just let the release value be dynamically determined always. Here's what I'm thinking. If we set `name`, `version`, and `release` for the createImage task, there's going to be conflict when the container build is imported into Koji because it will share the same NVR. Obviously, we don't want that. I think it still makes sense to use the same `name` and `version`. Any of this MUST be overwritten by values from image-build.conf.

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -180,6 +202,7 @@ def parse_image_build_config(self, config_file_name):
 
         config = ConfigParser()
         config.readfp(self.get_default_image_build_conf())
+        self.update_config_from_dockerfile(config)
```
lcarva: Can we call this method from within `get_default_image_build_conf` ?

MartinBasti: Sounds better :+1:,

it can also encapsulate 
```
config = ConfigParser()
config.readfp(self.get_default_image_build_conf())
... default stuff from dockerfile ....
```

atomic_reactor/plugins/pre_add_filesystem.py
MartinBasti: sounds reasonable, only `name` and `version` then, thanks.

lcarva:
Conflicts.

atomic_reactor/plugins/pre_add_filesystem.py
MartinBasti: updated

atomic_reactor/plugins/pre_add_filesystem.py
MartinBasti: updated

MartinBasti:
Conflicts resolved

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -165,8 +171,35 @@ def get_default_image_build_conf(self):
             'install_tree': install_tree,
             'repo': repo,
         }
+        config_fp = StringIO(self.DEFAULT_IMAGE_BUILD_CONF.format(**kwargs))
+
+        config = ConfigParser()
+        config.readfp(config_fp)
+
+        self.update_config_from_dockerfile(config)
```
lcarva: You *could* operate on `kwargs` instead of `config`, but I don't have a strong preference.

MartinBasti: I was thinking about that, but `version` is not in template (has no hardcoded default) and it is required so I cannot just add it into template. Updating ConfigParser directly looked more elegant to me than adding special logic to update default config only when it is defined in Dockerfile.

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -209,7 +241,7 @@ def parse_image_build_config(self, config_file_name):
         for option, value in config.items(section):
             if option in ('repo', 'format'):
                 value = [v for v in value.split(',') if v]
-            elif option in ('disk_size'):
+            elif option in ('disk_size',):
```
twaugh: Nice catch

MartinBasti:
Release notes:

> - `image-build.conf` default name and version are taken from Dockerfile labels (if defined)
>   - LABEL `com.redhat.component` is used as default value for `name`
>   - LABEL `version` is used as default value for `version`

PR #1106, opened 2018-10-23
koji_upload: include koji identifiers in buildroot

mlangsdorf:
Add some koji identification data to buildroot.extra.osbs.

### DIFF
--- a/atomic_reactor/plugins/post_koji_upload.py
+++ b/atomic_reactor/plugins/post_koji_upload.py
@@ -24,7 +24,7 @@ from atomic_reactor.constants import PROG, PLUGIN_KOJI_UPLOAD_PLUGIN_KEY
 from atomic_reactor.util import (get_version_of_tools, get_checksums,
                                  get_build_json, get_docker_architecture,
                                  get_image_upload_filename,
-                                 get_manifest_media_type)
+                                 get_manifest_media_type, ImageName)
 from atomic_reactor.rpm_util import parse_rpm_output, rpm_qf_args
 from osbs.exceptions import OsbsException
 
@@ -444,6 +444,25 @@ class KojiUploadPlugin(PostBuildPlugin):
 
         return output_files
 
+    def update_buildroot_koji(self, buildroot, output):
+        """
+        put the final koji information in the buildroot under extra.osbs
+        """
+        docker = output[1]['extra']['docker']
+
+        name = ''
+        for tag in docker['tags']:
+            for repo in docker['repositories']:
+                if tag in repo:
+                    iname = ImageName.parse(repo)
+                    name = iname.to_str(registry=False)
+                    break
+
+        buildroot['extra']['osbs']['koji'] = {
+            'build_name': name,
+            'builder_image_id': docker.get('digests', {})
+        }
+
     def get_metadata(self):
         """
         Build the metadata needed for importing the build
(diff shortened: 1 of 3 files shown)


tests/plugins/test_koji_upload.py
```diff
@@ -711,11 +711,27 @@ def validate_buildroot(self, buildroot):
...
             'builder_image_id',
+            'koji',
         ])
 
         assert is_string_type(osbs['build_id'])
         assert is_string_type(osbs['builder_image_id'])
 
+        koji = osbs['koji']
+        assert isinstance(koji, dict)
+        assert set(koji.keys()) == set([
```
MartinBasti: JFYI, you can create set directly `{'item1', 'item2'}` without usage of an extra list. (Except empty set, `empty_set = set()`, because `{}` is empty dict)

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -444,6 +444,25 @@ def add_log_type(output, arch):
 
         return output_files
 
+    def update_buildroot_koji(self, buildroot, output):
+        """
+        put the final koji information in the buildroot under extra.osbs
+        """
+        docker = output[1]['extra']['docker']
+
+        name = ''
```
MartinBasti: What causes  empty `name`? Failed build?

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -444,6 +444,25 @@ def add_log_type(output, arch):
...
+        for tag in docker['tags']:
+            for repo in docker['repositories']:
+                if tag in repo:
+                    iname = ImageName.parse(repo)
+                    name = iname.to_str(registry=False)
+                    break
+
+        buildroot['extra']['osbs']['koji'] = {
+            'build_name': name,
+            'builder_image_id': docker.get('digests', [])
```
twaugh: Should the default value be `{}` instead?

atomic_reactor/plugins/post_koji_upload.py
mlangsdorf: I don't think you can get to this stage of koji_upload and not have a valid tag in a repo.  But I'd rather not break builds in case I'm wrong, so I'll set a sensible default.

tests/plugins/test_koji_upload.py
mlangsdorf: Yes, but I'm following the idiom of the rest of the function.

mlangsdorf:
Documentation added.

atomic_reactor/plugins/post_koji_upload.py
mlangsdorf: Fixed.

atomic_reactor/plugins/post_koji_upload.py
MartinBasti: ok

tests/plugins/test_koji_upload.py
twaugh: I think it was originally done this way for Python 2.6 compatibility, which we no longer need.

IMHO we should follow the existing style in the commit making the change, and have a separate PR (not needed for this story) for using the "set literal" syntax throughout atomic-reactor where appropriate.

mlangsdorf:
Added to the release notes under Improvements:
- koji builds now display the koji name-nvr and digests with versions in the metadata under extra.osbs.koji

PR #1108, opened 2018-10-24
Drop integration-tests dir

MartinBasti:
The only integration test there is not working due big changes in codebase.


athos-ribeiro:
Maybe `pytest.ini` could go with this commit as well? The reason it was introduced was to skip integration tests, as shown in 016176a

MartinBasti:
Nice catch! Removed

MartinBasti:
No release notes needed, test change only

PR #1109, opened 2018-10-24
get_koji_module_build: fix handling of streams with dashes

owtaylor:
When the module build service stores information about a build into Koji,
it creates the version from the stream name by replacing '-' with '_'


lcarva:
Updated release notes draft:
> - `get_koji_module_build` handles streams with dashes

PR #1110, opened 2018-10-25
Check plugins availability before execution

athos-ribeiro:
Before running the first plugin, we want to verify if all requested
plugins exist. Former behavior would check availability for a plugin and
run it before checking the availability of the next plugin to be run.


MartinBasti:
For me it still looks quite much duplication of code, I'd propose different approach:
- split `run` into two parts: `preparation` and `execution` (run method is super long anyway)
- `preparation` part should process all requests, check plugins (failed if errors), extract required data, guarantee validity of data and return list of [named]tuples [(plugin_name, plugin_class, plugin_conf, allowed_to_fail)] (no yield, all plugins must be checked before executed)
- `execution` should just use values returned from the `preparation` and do the hard work.

MartinBasti:
Basically it can return just (plugin_name, plugin_instance, allowed_to_fail),  so preparation will check if instantiation is OK and wont fail later.

* maybe allowed_to_fail could be attribute of plugin

mlangsdorf:
Should any of this be moved to input_osv3?  We're already validating the plugins json against a schema there.

athos-ribeiro:
Do we want to also instantiate all plugins before running? Some plugins, such as `atomic_reactor/plugins/{pre_koji,prepub_squash}.py` do extra work when being instantiated.
The question here is if there is any plugin which instantiation would require that some other plugin had run. @MartinBasti, any thoughts?

athos-ribeiro:
> Should any of this be moved to input_osv3? We're already validating the plugins json against a schema there.

@mlangsdorf, thanks for the feedback!
Which checks do you believe should be osv3 plugin responsibility? It feels like loading plugins and handling possible loading errors should be part of the core application, not of a plugin. However osv3 does remove a few plugins based on some environment inspections.

in `docs/plugins.md` describes `input_osv3` plugin reads

> Loads build configuration from environment variables provided by OpenShift v3

mlangsdorf:
get_available_plugins does not, as far as I can tell, actually run or load plugins.  It just confirms that the name and phase are valid and errors out if there's an invalid value on a required phase or name.  But the plugins.json went through a yaml config parser to confirm that stuff in input_osv3, so I don't understand what's being done here and how it differs from the checks in input_osv3.

It feels like get_available_plugins() is redundant code that could be moved to input_osv3 or restructured somehow.

MartinBasti:
@athos-ribeiro so I'd rather not instantiate the plugins

athos-ribeiro:
Thanks @mlangsdorf.
Wouldn't moving the changes to the osv3 plugin mean we'd get different behaviors when using different input methods? Then, should I assume that the only input method we curently care about is osv3?

athos-ribeiro:
@mlangsdorf, I moved the redundant 'name' key verification into the osv3 plugin (we just needed to check the schema earlier). Does is still make sense to re-check the schema after removing plugins?

We are verifying if the plugin classes were actually loaded and are available. Would it make sense to also move those to the osv3 plugin (this would require to set the bindings, in the osv3 plugin, replicating some plugin.py code)?

mlangsdorf:
I think loading the plugins should stay inside plugins.py.  Verifying the JSON and the resulting python arrays and objects inside input_osv3 makes sense and prevents duplication of effort, but input_osv3 is not the right place to handle what happens if a plugin that exists in the schema doesn't actually exist on disk or fails to load or whatever.

atomic_reactor/plugin.py
```diff
@@ -181,36 +184,27 @@ def save_plugin_timestamp(self, plugin, timestamp):
...
-                raise exc
-
-            plugin_conf = plugin_request.get("args", {})
-            try:
-                plugin_class = self.plugin_classes[plugin_name]
-            except KeyError:
+            plugin_name = plugin_request['name']
+            plugin_class = self.get_plugin_class(plugin_name)
+            if not plugin_class:
                 if plugin_request.get('required', True):
```
mlangsdorf: Can this still fail?  The plugin name and class were already verified.

athos-ribeiro: This will fail if the plugins configuration contains a plugin that passes the schema but was not loaded during `atomic_reactor.plugin.load_plugins()` (e.g., a plugin file was removed or there is a typo in some plugin name)

atomic_reactor/plugin.py
```diff
@@ -227,6 +221,26 @@ def run(self, keep_going=False, buildstep_phase=False):
                 plugin_is_allowed_to_fail = plugin_request['is_allowed_to_fail']
             except (TypeError, KeyError):
                 plugin_is_allowed_to_fail = getattr(plugin_class, "is_allowed_to_fail", True)
+            plugin_conf = plugin_request.get("args", {})
+            plugins_data.append((plugin_name, plugin_class, plugin_conf, plugin_is_allowed_to_fail))
```
MartinBasti: I'm not a huge fan of having structured data in tuples. Could we use NamedTuple or dict?

tests/test_plugin.py
```diff
@@ -154,6 +154,77 @@ def test_required_plugin_failure(tmpdir, docker_tasker, runner_type, required):
...
+    runner = runner_type(docker_tasker, workflow,
+                         [{"name": MyPlugin.key, "required": False},
+                          {"name": "no_such_plugin", "required": required}])
+    expected_log_message = "running plugin '%s'" % MyPlugin.key
+
+    # build step plugins set "required" to False
+    if required and (runner_type != BuildStepPluginsRunner):
+        with pytest.raises(PluginFailedException):
+            runner.run()
+        assert expected_log_message not in [l.getMessage() for l in caplog.records()]
```
MartinBasti: Don't waste memory by creating lists :-)
`assert all(expected_log_message not in l.getMessage() for l in caplog.records())`

tests/test_plugin.py
```diff
@@ -154,6 +154,77 @@ def test_required_plugin_failure(tmpdir, docker_tasker, runner_type, required):
...
+    else:
+        runner.run()
+        assert expected_log_message in [l.getMessage() for l in caplog.records()]
```
MartinBasti: `assert any(expected_log_message in l.getMessage() for l in caplog.records())`

atomic_reactor/plugin.py
mlangsdorf: Ah, okay. I missed the distinction between self.plugin_classes and the plugins conf from osv3.

atomic_reactor/plugin.py
```diff
@@ -181,36 +184,27 @@ def save_plugin_timestamp(self, plugin, timestamp):
...
-                if keep_going:
-                    continue
-                raise exc
-
-            plugin_conf = plugin_request.get("args", {})
-            try:
-                plugin_class = self.plugin_classes[plugin_name]
-            except KeyError:
+            plugin_name = plugin_request['name']
+            plugin_class = self.get_plugin_class(plugin_name)
```
mlangsdorf: Can't this be simplified to:
```
plugin_class = self.plugin_classes.get(plugin_name)
```

atomic_reactor/plugin.py
```diff
@@ -227,6 +221,26 @@ def run(self, keep_going=False, buildstep_phase=False):
                 plugin_is_allowed_to_fail = plugin_request['is_allowed_to_fail']
```
mlangsdorf: can't this be simplified to:
```
plugin_is_allowed_to_fail = plugin_request.get('is_allowed_to_fail', True)
```

MartinBasti: Probably this, If I read correctly, it can be defined in two places
```python
plugin_is_allowed_to_fail = plugin_request.get('is_allowed_to_fail', getattr(plugin_class, "is_allowed_to_fail", True))
```

atomic_reactor/inner.py
```diff
@@ -422,11 +422,24 @@ def build_docker_image(self):
         try:
             self.fs_watcher.start()
             signal.signal(signal.SIGTERM, self.throw_canceled_build_exception)
-            # time to run pre-build plugins, so they can access cloned repo
-            logger.info("running pre-build plugins")
             prebuild_runner = PreBuildPluginsRunner(self.builder.tasker, self,
                                                     self.prebuild_plugins_conf,
                                                     plugin_files=self.plugin_files)
+            buildstep_runner = BuildStepPluginsRunner(self.builder.tasker, self,
```
mlangsdorf: Why was this changed?  What was wrong with creating each running and executing it?

atomic_reactor/inner.py
```diff
@@ -493,9 +497,13 @@ def build_docker_image(self):
         finally:
             # We need to make sure all exit plugins are executed
             signal.signal(signal.SIGTERM, lambda *args: None)
-            exit_runner = ExitPluginsRunner(self.builder.tasker, self,
-                                            self.exit_plugins_conf,
-                                            plugin_files=self.plugin_files)
+            try:
+                exit_runner
+            except NameError:
```
mlangsdorf: I don't understand what's going on here.  Try the exit_runner and if it generates a key_error, assign the exact same variable and run it again.  How is this an improvement over the old way?

atomic_reactor/inner.py
athos-ribeiro: Currently, atomic-reactor loads plugins on a PluginsRunner initialization. It does so by loading all plugins available in the plugins path. It then makes all the plugins of the called runner type available to that runner.

This patch takes advantage of that work on initialization to also check if the requested plugins are available. Since this is done on a per runner basis, we now initialize all runners before running the first one.

Since we are not running each runner right after initializing it, I am also proposing to not reload all plugins every time we create a runner instance (current behavior). This will reduce the size of our logs.

In order to perform these changes, I had to make a few changes around the whole test suite, since most inner/plugin/plugins tests depend on the PluginsRunner initialization.

atomic_reactor/inner.py
athos-ribeiro: We now create instances of each type of runner in the beginning of the build process. The runners may do some work upon initialization, which can lead to exceptions. The exit runner is also initialized with all the other runners. If a runner fail to initialize, the code will proceed to run the exception section and then move to the `finally` statement. If the exception occurs in any runner initialization other than the exit_runner, it needs to be initialized here.

athos-ribeiro:
Rebased on `master` latest changes.

atomic_reactor/inner.py
MartinBasti: Could be use `exit_runner = None` instead? I don't really like catching `NameError`

atomic_reactor/inner.py
```diff
@@ -419,14 +419,29 @@ def build_docker_image(self):
...
+                                                      plugin_files=self.plugin_files)
+            prepublish_runner = PrePublishPluginsRunner(self.builder.tasker, self,
+                                                        self.prepublish_plugins_conf,
+                                                        plugin_files=self.plugin_files)
+            postbuild_runner = PostBuildPluginsRunner(self.builder.tasker, self,
+                                                      self.postbuild_plugins_conf,
+                                                      plugin_files=self.plugin_files)
+            exit_runner = ExitPluginsRunner(self.builder.tasker, self,
+                                            self.exit_plugins_conf,
+                                            keep_going=True,
```
rcerven: not needed, keep_going is used in run() not init

atomic_reactor/inner.py
```diff
@@ -493,9 +499,11 @@ def build_docker_image(self):
         finally:
             # We need to make sure all exit plugins are executed
             signal.signal(signal.SIGTERM, lambda *args: None)
-            exit_runner = ExitPluginsRunner(self.builder.tasker, self,
-                                            self.exit_plugins_conf,
-                                            plugin_files=self.plugin_files)
+            if exit_runner is None:
+                exit_runner = ExitPluginsRunner(self.builder.tasker, self,
+                                                self.exit_plugins_conf,
+                                                keep_going=True,
```
rcerven: not needed, keep_going is used in run() not init

rcerven:
check for invalid plugin request was removed : https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugin.py#L201
check should in get_available_plugins, or either in new method

athos-ribeiro:
> check for invalid plugin request was removed : https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugin.py#L201
check should in get_available_plugins, or either in new method

The check was removed during the PR review (discussion starts [here](https://github.com/projectatomic/atomic-reactor/pull/1110#issuecomment-433883755)). The reason behind it is that the configuration is being validated with the json schema by the `input_osv3` plugin. @rcerven, thoughts?

rcerven:
ok I see, maybe we can add small test for input_osv3, with plugin without 'name' to just validate it

athos-ribeiro:
@rcerven, there is a test covering this case here: https://github.com/projectatomic/atomic-reactor/pull/1110/files#diff-d34fcce875d0948d957af0f6118c8377R391

athos-ribeiro:
Release Notes:

- Check requested plugins availability before build process (aborts if required plugin is missing)
- Do not reload plugins before each phase. Instead, load all plugins before entering build

PR #1111, opened 2018-10-26
remove flatpak_base_image from user params schema

rcerven:


MartinBasti:
Release notes:

> - remove flatpak_base_image from user params schema (Part of: flatpak base image is now properly loaded from a reactor-config-map)

PR #1112, opened 2018-10-26
flatpak_create_dockerfile: Initialize /etc/os-release in the installroot

owtaylor:
Recent versions of DNF look for a PLATFORM_ID in /etc/os-release to figure
out a module name for the base package set (e.g. platform:f29.) This is
looked up in the installroot, so we need to create it there. To get the
*correct* /etc/os-release, we install 'system-release' into a separate
installroot, then copy /etc/os-release over. Installing this package into
the final installroot as an early step doesn't work because once we do
that, the DNF configuration is no longer read from outside the installroot.

Some shell variables are added to reduce rampant duplication.


rcerven:
rn updated
- `flatpak_create_dockerfile` Initialize /etc/os-release in the installroot

PR #1114, opened 2018-10-26
Fix confusing error when parent image not found

MartinBasti:


atomic_reactor/util.py
```diff
@@ -1104,7 +1105,12 @@ def get_inspect_for_image(image, registry, insecure=False, dockercfg_path=None):
         else:
             blob_config = json.loads(v1_json['history'][0]['v1Compatibility'])
     else:
-        raise NotImplementedError("No v2 schema 1 image, or v2 schema 2 image or list, found")
+        raise RuntimeError(
+            "Image {image_name} not found: No v2 schema 1 image, "
+            "or v2 schema 2 image or list, found".format(
+                image_name=image
```
mlangsdorf: split this as:
```
        raise RuntimeError("Image {image_name} not found: No v2 schema 1 image, "
                           "or v2 schema 2 image or list, found".format(image_name=image))
```

atomic_reactor/util.py
```diff
@@ -1117,7 +1123,10 @@ def get_inspect_for_image(image, registry, insecure=False, dockercfg_path=None):
     }
 
     if not blob_config:
-        raise RuntimeError("Couldn't get inspect data from digest config")
+        raise RuntimeError(
+            "Image {image_name}: Couldn't get inspect data from digest config".format(
```
mlangsdorf: split this as
```
        raise RuntimeError("Image {image_name}: Couldn't get inspect data "
                           "from digest config".format(image_name))
```

atomic_reactor/util.py
```diff
@@ -1083,7 +1083,8 @@ def get_inspect_for_image(image, registry, insecure=False, dockercfg_path=None):
     if 'v2_list' in all_man_digests:
         man_list_json = all_man_digests['v2_list'].json()
         if man_list_json['manifests'][0]['mediaType'] != MEDIA_TYPE_DOCKER_V2_SCHEMA2:
-            raise NotImplementedError('v2 schema 1 in manifest list')
+            raise RuntimeError(
+                'Image {image_name}: v2 schema 1 in manifest list'.format(image_name=image))
```
mlangsdorf: split this as
```
            raise RuntimeError('Image {image_name}: v2 schema 1 ',
                               'in manifest list'.format(image_name=image))
```

atomic_reactor/util.py
MartinBasti: I'm used to use named substitutions because it makes it clean what is expected there and is readable better and it is required for translators (not the case here though).

> Also, consider not passing arguments by name into str.format(), since you're eating up 20+ characters each time you do that and the string formats are not that complicated.

Why is this so bad? It is still better to split long text/code to multiple lines for readability.

Also the indentation you requested makes big empty spaces when functions are long.
```
block:
    block:
         block:
             self.myobject.anotherobject.method("Lorem ipsum dolor sit amet, "
                                                "consectetur adipiscing elit, "
                                                "sed do eiusmod tempor "
                                                "incididunt ut labore et "
                                                "dolore magna aliqua.")
```

This way seems cleaner to me:
```
block:
    block:
         block:
             self.myobject.anotherobject.method(
                "Lorem ipsum dolor sit amet, consectetur adipiscing elit,  sed 
                "do eiusmod tempor incididunt ut labore et dolore magna aliqua."
             )
```

Makes git diffs bigger when change happen.

```diff
- logger.info(
+ logger.warning(
   "Something bad happened"
   "2nd line"
)
```
```diff
- logger.info("Something bad happened"
-             "2nd line")
+ logger.warning("Something bad happened"
+                "2nd line")
```

I will incorporate requested changes given they are PEP8 compatible too.

PR #1116, opened 2018-10-30
pulp_sync: remove unused delete_from_registry param

athos-ribeiro:
The delete_from_registry param was added in f83797f to allow v2 images
to be deleted from the registry after syncing. This feature was then
implemented later in a dedicated plugin in 7cd936c.

We are moving towards removal of the post-syncing removal feature, hence, there is no longer need for this non-implemented feature here.


MartinBasti:
No releaseno notes: removal of unimplemented functionality

PR #1117, opened 2018-10-30
WrappedDocker: Work around export inefficiency in older docker-python

owtaylor:
Exporting a container with older docker-python is very slow, because the
contents are returned in chunks of a few dozen bytes. Hack around this
by going right to the requests iter_content() API to pass in a better
chunk size (we use the same 2MB as the default as recent docker-python.)

This cuts the time to export a 70MB image from many minutes to 2 sends.

<hr>

Note: it might look more natural to do this in DockerTasker rather than in WrappedDocker, but I found it worked better in WrappedDocker when trying to get exception/retry handling right. In WrappedDocker it can also be dropped with minimal effort if the docker-python requirement is increased appropriately.

### DIFF
--- a/atomic_reactor/core.py
+++ b/atomic_reactor/core.py
@@ -262,6 +262,20 @@ class WrappedDocker(object):
         else:
             return orig_attr
 
+    def _export(self, container, chunk_size):
+        # Older versions of docker-python don't handle streaming responses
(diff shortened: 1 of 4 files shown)


rcerven:
rn updated:
- WrappedDocker: Work around export inefficiency in older docker-python

PR #1118, opened 2018-10-31
docs: update delete_from_registry plugin status

athos-ribeiro:
Depends on https://github.com/projectatomic/osbs-client/pull/813


athos-ribeiro:
Doc changes. Not adding to release notes (release note entry added in osbs-client for projectatomic/osbs-client#813)

PR #1119, opened 2018-10-31
flatpak_create_dockerfile: Avoid -'s in the version label

owtaylor:
The version label will be used for the Koji version - and Koji
versions can't have -'s. Convert - to _. (This is the same thing
that MBS does when converting streams to Koji versions.)


lkolacek:
LGTM :+1:

athos-ribeiro:
Internal change - Not documenting in release notes.

PR #1121, opened 2018-11-05
resolve_module_compose: Start the compose with the correct architectures

owtaylor:
We need to pass the correct set of architectures to ODCS for starting the compose
(the compose is always started by the orchestrator, and the compose_id passed
to worker builds.)

<hr>

Found in testing on Fedora now that Fedora is setting up multi-arch OSBS


MartinBasti:
Release notes:

> - ODCS: Start the compose with the correct architectures

PR #1124, opened 2018-11-06
flatpak: allow overriding the name and com.redhat.component labels

owtaylor:
atomic-reactor uses:
 name label => repository name when uploading to a registry
 com.redhat.component => name when uploading to Koji

Generally these names don't matter a lot for a Flatpak - the reverse-DNS ID is
the important things, but there are some reasons we might need to override them -
to match conventions for containers, for example, and, in particular, to distinguish
multiple Flatpaks being generated from different profiles of the same module.

Allow specifying name/component in the flatpak: section of container.yaml to
override the labels in the generated Dockerfile.
<hr>
Suggestions for alternative names for the keys in container.yaml or ways of representing this information accepted. I think this way is functional but a little obscure.


athos-ribeiro:
Release notes:
 - possibility to specify 'name' and 'component' for flatpaks to override labels in generated Dockerfile

PR #1126, opened 2018-11-09
Use logger for warnings

MartinBasti:
Koji doesn't collect stderr/stdout. Only logger is collected, so warning.warn() output
is not shown in collected logs.


MartinBasti:
No release notes, fixes unreleased feature

PR #1127, opened 2018-11-10
Flatpak: pass compose ids to the worker builds correctly

owtaylor:
When the module_compose_id build kwargs was replaced with compose_ids,
the override that is done by the orchestrator wasn't adjusted, meaning
that worker builds were individually triggering composes.

Move the override to pre_resolve_module_compose to make the connection
more obvious and harder to accidentally break.


MartinBasti:
Release notes:

> - Flatpak: pass compose ids to the worker builds correctly

PR #1128, opened 2018-11-21
squash image from imagebuilder when last parent is scratch, because i…

rcerven:
when using imagebuilder and last/only parent is 'scratch', it produces 2 layers, from which 1 of them is empty, so we have to actually run squash even for imagebuilder in that case

### DIFF
--- a/atomic_reactor/plugins/prepub_squash.py
+++ b/atomic_reactor/plugins/prepub_squash.py
@@ -89,7 +89,8 @@ class PrePublishSquashPlugin(PrePublishPlugin):
         self.save_archive = save_archive
 
     def run(self):
-        if self.workflow.build_result.skip_layer_squash:
+        if (self.workflow.build_result.skip_layer_squash and
+                not self.workflow.builder.base_from_scratch):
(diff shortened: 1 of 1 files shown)


ben-alkov:
@rcerven Tested. FWIW, LGTM. Need more acronyms :P

twaugh:
We end up with two layers after a custom base image too -- one for the imported filesystem, and another for the Dockerfile processing. I think this is might be a very similar case.

Do we really need to only end up with a single layer for FROM-scratch builds? Note that if we left this alone, so there were two layers, we could potentially drop squashing altogether at some point in the future. If we stick to wanting a single layer, we must always carry squashing support.

rcerven:
@twaugh 
Yeah I know we end up with 2 layers for base as well, in scratch case is only weird that the one layer is empty 0 size, but yeah I am aware of the fact we wanted to get rid of squashing, so this is basically for us to decide, if we want to have 2 layers for scratch (1 empty), and get rid of squash, or keep squash and have only 1 layer

rcerven:
closing this, we will go with 2 layers for scratch without need on squashing

PR #1129, opened 2018-11-22
Fix yum repo handling when URL does not end '.repo' (OSBS-6562)

twaugh:
The URL we use to retrieve the file should be used as-is. Only the filename we inject into the container image should end .repo if it does not already.


twaugh:
Draft release notes updated:
- Adding yum repo URLs which do not end '.repo' works correctly again

PR #1130, opened 2018-11-22
Prevent race conditions cause by floating tags

MartinBasti:
Image tags are floating so builds can end up in inconsistent state where
orchestrator/workers may end up in using different image.

To avoid such issue we must switch to image digests. Orchestrator
will provide digests to be used at workers to eliminite race condition.

To keep BW compatibility, if orchestrator doesn't provide information
about digests only warning is logged.

Jira: OSBS-5674

### DIFF
--- a/atomic_reactor/plugins/pre_pull_base_image.py
+++ b/atomic_reactor/plugins/pre_pull_base_image.py
@@ -181,6 +181,16 @@ class PullBaseImagePlugin(PreBuildPlugin):
                     )
 
     def _get_image_for_different_arch(self, image, platform):
+        """Get image from random arch
+
+        This is a workaround for aarch64 platform, because orchestrator cannot
+        get this arch from manifests lists so we have to provide digest of a
+        random platform to get image metadata for orchestrator.
+
+        For standard platforms like x86_64, ppc64le, ... this method returns
+        the corresponding digest
+
+        """
         new_image = None
         parents_digests = self.workflow.builder.parent_images_digests
         try:
--- a/atomic_reactor/plugins/pre_pull_base_image.py
+++ b/atomic_reactor/plugins/pre_pull_base_image.py
@@ -37,7 +37,7 @@ class PullBaseImagePlugin(PreBuildPlugin):
     is_allowed_to_fail = False
 
     def __init__(self, tasker, workflow, parent_registry=None, parent_registry_insecure=False,
-                 check_platforms=False, inspect_only=False):
+                 check_platforms=False, inspect_only=False, parent_images_digests=None):
         """
         constructor
 
@@ -62,6 +62,8 @@ class PullBaseImagePlugin(PreBuildPlugin):
         else:
             self.parent_registry = None
             self.parent_registry_insecure = False
+        if parent_images_digests:
+            self._load_parent_images_digests(parent_images_digests)
 
     def run(self):
         """
@@ -82,6 +84,19 @@ class PullBaseImagePlugin(PreBuildPlugin):
                 image = self._resolve_base_image(build_json)
 
             image = self._ensure_image_registry(image)
+
+            if self.check_platforms:
+                # run only at orchestrator
+                self._collect_image_digests(image)
+
+            # try to stay with digests
+            image_with_digest = self._get_image_with_digest(image, current_platform)
+            if image_with_digest is None:
+                self.log.warning("Cannot resolve platform '%s' specific digest for image '%s'",
+                                 current_platform, image)
+            else:
+                self.log.info("Replacing image '%s' with '%s'", image, image_with_digest)
+                image = image_with_digest
 
             if organization:
                 image.enclose(organization)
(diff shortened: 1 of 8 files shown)


atomic_reactor/schemas/user_params.json
```diff
@@ -31,6 +31,7 @@
     "koji_task_id": {"type": "integer"},
     "koji_upload_dir": {"type": "string"},
     "name": {"type": "string"},
+    "parent_images_digests": {"type": "object"},
```
mlangsdorf: type: object covers a lot of structures.  Can you provide more detail on the internals of this structure here?

MartinBasti: I can update schema, but It will be too generic anyway.

Example of data
```
{
  'registry/fedora:latest': {
    'x86_64': 'fedora@sha256:14654aaf....',
    'ppc64le': 'fedora@sha256:445431dea...'
  }
}
```

Race condition is eliminated because pull_base_image will use the same digests as orchestrator, workers are unaffected by image tag change during build.

MartinBasti:
Depends on: https://github.com/projectatomic/osbs-client/pull/826

atomic_reactor/schemas/user_params.json
twaugh: I feel like we only need a single manifest digest (the one for the manifest list) for each image name. All the workers can use the same manifest list and find their own arch-native image within it.

atomic_reactor/util.py
```diff
@@ -1567,3 +1567,95 @@ def dump_stacktraces(sig, frame):
...
+        image_name_tag = self._key(image)
+
+        image_name = image.to_str(tag=False)
+        name_digest = '{}@{}'.format(image_name, digest)
+
+        image_digests = self._images_digests.setdefault(image_name_tag, {})
+        image_digests[platform] = name_digest
+
+    def get_image_digests(self, image):
+        """Get digest of specified image and platform
```
twaugh: This doesn't seem to describe the method.

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -109,32 +125,76 @@ def run(self):
...
+                    parents_digests.update_image_digest(
+                        image, present_platform, digest
+                    )
 
-                if platform not in build_image_digests:
-                    new_image = ImageName.parse(build_image_digests[present_platform])
-            except KeyError:
-                self.log.info('Cannot validate available platforms for base image '
-                              'because platform descriptors are not defined')
+    def _get_image_for_different_arch(self, image, platform):
```
twaugh: I don't think I understand the purpose of this.

atomic_reactor/util.py
MartinBasti: fixed

atomic_reactor/plugins/pre_pull_base_image.py
MartinBasti: added docstring

atomic_reactor/schemas/user_params.json
MartinBasti: I've updated schema, please check if it's ok or if it should be more specific

atomic_reactor/schemas/user_params.json
```diff
@@ -31,6 +31,26 @@
...
     "koji_upload_dir": {"type": "string"},
     "name": {"type": "string"},
+    "parent_images_digests": {
+      "type": "object",
+      "patternProperties": {
+        "^.+:.+": {
+          "description": "Image with tags",
+          "type": "object",
+          "patternProperties": {
+            "^.+": {
```
twaugh: Is this equivalent to "."?

atomic_reactor/schemas/user_params.json
```diff
@@ -31,6 +31,26 @@
     "koji_task_id": {"type": "integer"},
     "koji_upload_dir": {"type": "string"},
     "name": {"type": "string"},
+    "parent_images_digests": {
+      "type": "object",
+      "patternProperties": {
+        "^.+:.+": {
```
twaugh: Maybe include `$` at the end of the regex?

atomic_reactor/schemas/user_params.json
```diff
@@ -31,6 +31,26 @@
...
+          "description": "Image with tags",
+          "type": "object",
+          "patternProperties": {
+            "^.+": {
+              "description": "Platform specific digests",
+              "type": "string",
+              "pattern": "@sha256:"
```
twaugh: Technically 'sha256' is the algorithm name, which will change in future.

atomic_reactor/schemas/user_params.json
MartinBasti: No, it is equivalent to `..*`, i.e. at least one character expected

MartinBasti: Yes, because I have omitted `$` at the end, so it is equivalent of `.`

atomic_reactor/schemas/user_params.json
MartinBasti: Added

atomic_reactor/schemas/user_params.json
MartinBasti: replaced with generic pattern

tests/plugins/test_pull_base_image.py
```diff
@@ -661,6 +669,66 @@ def workflow_callback(workflow):
...
+                # fail to provide platform specific digests
+                manifest_list = {
+                    'manifests': []
+                }
+
+                manifest_tag = 'registry.example.com' + '/' + BASE_IMAGE_W_SHA
+                base_image_result = ImageName.parse(manifest_tag)
+                manifest_image = base_image_result.copy()
+
+                (flexmock(atomic_reactor.util)
```
MartinBasti: TODO: I have to figure out, why in fail case I'm still getting manifest_list with digests

tests/plugins/test_pull_base_image.py
```diff
@@ -661,6 +669,66 @@ def workflow_callback(workflow):
         assert pulling_msg in caplog.text
         assert tagging_msg in caplog.text
 
+    @pytest.mark.parametrize('orchestrator, fail', tuple(product((True, False), repeat=2)))
+    def test_parent_images_digests(self, caplog, orchestrator, fail):
+        def workflow_callback(workflow):
+            workflow = self.prepare(workflow)
```
MartinBasti: TODO: at orchestrator check if `wokflow.builder.parent_images_digest` are filled properly

MartinBasti:
Release notes:
> - orchestrator collects images digests which are used by workers to prevent a race condtition when tags may change during worker builds.

PR #1131, opened 2018-11-23
Update test requirements

athos-ribeiro:
The test requirements have been frozen to avoid unexpected failures due
to test package upgrades. They should be manually upgraded from time to
time. The issues below had to be addressed for this upgrade.

* pytest 3.3 removed backwards compatibility with pytest-capturelog and
merged pytest-catchlog into the core.
* pytest 4.0.0 fails if fixtures are called implicitly.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,7 +1,7 @@
 flexmock
 ordereddict
 responses>=0.9.0
-pytest==3.2.5
-pytest-capturelog==0.7
-pytest-cov==2.5.1
+pytest==4.0.*
+pytest-cov==2.6.*
+six>=1.10.0 # required by pytest-cov
 flake8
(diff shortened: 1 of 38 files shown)


tests/requirements.txt
```diff
@@ -1,7 +1,7 @@
 flexmock
 ordereddict
 responses>=0.9.0
-pytest==3.2.5
-pytest-capturelog==0.7
-pytest-cov==2.5.1
+pytest==4.0.0
```
MartinBasti: do we really need exact version? How about `pytest==4.0.*` to keep compatibility but get bugfixes?

athos-ribeiro: That is a good idea :+1: 

Done.

athos-ribeiro:
Tests related change only. Not adding release notes.

PR #1133, opened 2018-11-27
set release for scratch build from build_id if release isn't set in D…

rcerven:
…ockerfile

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -13,6 +13,7 @@ from atomic_reactor.util import df_parser
 from osbs.utils import Labels
 from atomic_reactor.plugins.pre_reactor_config import get_koji_session
 from atomic_reactor.constants import PLUGIN_BUMP_RELEASE_KEY
+from atomic_reactor.util import get_build_json, is_scratch_build
 
 
 class BumpReleasePlugin(PreBuildPlugin):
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -137,6 +138,9 @@ def run(self):
 
         if self.append:
             next_release = self.get_next_release_append(component, version, release)
+        elif is_scratch_build():
```
MartinBasti: Shouldn't be scratch build first case to test to override other build options?

rcerven: that append will work even when "wrong" release is specified as it is just appending additional string to already existing release, so we can keep that functionality

MartinBasti: From my POW, scratch builds are isolated from any releases, so I assume that once --scratch is used, we shouldn't even touch koji for existing releases.

This makes results inconsistent.

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -137,6 +138,9 @@ def run(self):
...
+            metadata = get_build_json().get("metadata", {})
+            next_release = metadata.get("name", "1")
```
MartinBasti: Haven't we talked about using a timestamp as default value?

rcerven: well we are using Openshift build name, if that isn't defined something is really off, so we just default to 1, that is really just corner case

rcerven:
rn updated: 
- when release label isn't specified for scratch builds, openshift build id will be used as value

PR #1134, opened 2018-11-27
support empty package lists in pre_resolve_composes

mlangsdorf:
ODCS now has support for generating packages based on the koji tag if no package list is provided.

Add support for the same in atomic_reactor.
See also https://github.com/projectatomic/osbs-docs/pull/64

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -315,6 +315,7 @@ class ComposeConfig(object):
 
     def __init__(self, data, pulp_data, odcs_config, koji_tag=None, arches=None):
         data = data or {}
+        self.use_packages = 'packages' in data
         self.packages = data.get('packages', [])
         self.modules = data.get('modules', [])
         self.pulp = {}
@@ -352,7 +353,7 @@ class ComposeConfig(object):
         self.validate_for_request()
 
         requests = []
-        if self.packages:
+        if self.use_packages:
             requests.append(self.render_packages_request())
         elif self.modules:
             requests.append(self.render_modules_request())
@@ -398,7 +399,7 @@ class ComposeConfig(object):
 
     def validate_for_request(self):
         """Verify enough information is available for requesting compose."""
-        if not self.packages and not self.modules and not self.pulp:
+        if not self.use_packages and not self.modules and not self.pulp:
             raise ValueError("Nothing to compose (no packages, modules, or enabled pulp repos)")
 
         if self.packages and self.modules:
--- a/tests/plugins/test_resolve_composes.py
+++ b/tests/plugins/test_resolve_composes.py
@@ -290,6 +290,34 @@ class TestResolveComposes(object):
 
         self.run_plugin_with_args(workflow, reactor_config_map=reactor_config_map)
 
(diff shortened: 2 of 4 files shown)


atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -315,6 +315,8 @@ class ComposeConfig(object):
 
     def __init__(self, data, pulp_data, odcs_config, koji_tag=None, arches=None):
         data = data or {}
+        self.use_packages = 'packages' in data
+        print("use package is {0}".format(self.use_packages))
```
MartinBasti: Logger?

mlangsdorf: Errant debug line, removed.

athos-ribeiro:
Other than the possible test name typo, this LGTM

tests/plugins/test_resolve_composes.py
```diff
@@ -290,6 +290,34 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):
 
         self.run_plugin_with_args(workflow, reactor_config_map=reactor_config_map)
 
+    @pytest.mark.parametrize(('with_modules'), (True, False))
+    def test_request_compose_empty_packagese(self, workflow, reactor_config_map, with_modules):
```
twaugh: packagese → packages

tests/plugins/test_resolve_composes.py
```diff
@@ -290,6 +290,34 @@ def test_request_compose_for_modules(self, workflow, reactor_config_map):
...
+                packages:
+            """)
+        if with_modules:
+            repo_config = dedent("""\
+                compose:
+                    packages:
+                    modules:
+                    - spam_modules
+                    - bacon_modules
+                    - eggs_modules
```
twaugh: Oh, I see, and the modules will be ignored.

One thing we *could* do is adjust the schema so that either 'packages' or 'modules' is specified, but not both:
```
"compose": {
  ...
    "oneOf": [
    {
      "required": [
        "packages"
      ]
    },
    {
      "required": [
        "modules"
      ]
    },
  ]
}
```

Since providing both 'packages' and 'modules' satisfies both of these requirements, the "oneOf" specifier will fail:
```
<ValidationError: "{'packages': None, 'modules': ['spam_modules', 'bacon_modules', 'eggs_modules']} is valid under each of {u'required': [u'modules']}, {u'required': [u'packages']}"
```
What do you think about doing this (perhaps as a separate story)? @mlangsdorf @rcerven @MartinBasti

mlangsdorf: The docs specify "The 'packages' and 'modules' keys are mutually exclusive. If both are
provided, 'modules' will be ignored."  So that's what I implemented.

If the preference is to implement failure on multiple provision, I'm fine with that and it's probably a more sensible solution.

twaugh: How about if I file a separate story to make them mutually exclusive such that providing both is an error, including updating the docs to state that? I think it would involve updating some test cases as well.

tests/plugins/test_resolve_composes.py

mlangsdorf:
Added to RN under #Improvements
"- module composes can specify an empty packages key, and the compose will include all packages associated with the base image's koji tag.

PR #1135, opened 2018-11-28
custom base images: return early from plugins that can't handle them

mlangsdorf:
add a utility function to detect a custom base image and stash whether
the base image is a custom base in the Builder.

In pre_add_filesystem, check if the base image is a custom base and
return immediately if it isn't.  In the other 4 plugins, return
immediately if the base image is a custom base image.

### DIFF
--- a/atomic_reactor/build.py
+++ b/atomic_reactor/build.py
@@ -18,7 +18,7 @@ import docker.errors
 import atomic_reactor.util
 from atomic_reactor.core import DockerTasker, LastLogger
 from atomic_reactor.util import (ImageName, print_version_of_tools, df_parser,
-                                 base_image_is_scratch, DigestCollector)
+                                 base_image_is_scratch, DigestCollector, base_image_is_custom)
 from atomic_reactor.constants import DOCKERFILE_FILENAME
 
 logger = logging.getLogger(__name__)
@@ -157,6 +157,7 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
         self.built_image_info = None
         self.image = ImageName.parse(image)
         self.base_from_scratch = False
+        self.custom_base_image = False
 
         # get info about base image from dockerfile
         build_file_path, build_file_dir = self.source.get_build_file_path()
@@ -233,6 +234,8 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
 
     def set_base_image(self, base_image, parents_pulled=True, insecure=False):
         self.base_from_scratch = base_image_is_scratch(base_image)
+        if not self.custom_base_image:
+            self.custom_base_image = base_image_is_custom(base_image)
         self.base_image = ImageName.parse(base_image)
         self.original_base_image = self.original_base_image or self.base_image
         self.recreate_parent_images()
--- a/atomic_reactor/plugins/pre_add_filesystem.py
+++ b/atomic_reactor/plugins/pre_add_filesystem.py
@@ -366,7 +366,7 @@ class AddFilesystemPlugin(PreBuildPlugin):
 
     def run(self):
         base_image = self.workflow.builder.base_image
-        if base_image.namespace != 'koji' or base_image.repo != 'image-build':
+        if not self.workflow.builder.custom_base_image:
             self.log.info('Base image not supported: %s', base_image)
             return
 
--- a/atomic_reactor/plugins/pre_koji_parent.py
+++ b/atomic_reactor/plugins/pre_koji_parent.py
@@ -73,7 +73,7 @@ class KojiParentPlugin(PreBuildPlugin):
         self._poll_start = None
 
     def run(self):
-        if not self.workflow.builder.base_from_scratch:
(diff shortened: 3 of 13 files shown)


atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -366,7 +366,7 @@ def stream_filesystem(self, task_id, filesystem_regex):
 
     def run(self):
         base_image = self.workflow.builder.base_image
-        if base_image.namespace != 'koji' or base_image.repo != 'image-build':
+        if not self.workflow.builder.custom_base_image:
```
twaugh: I'd expect this to be
```
if not (self.workflow.builder.custom_base_image or self.workflow.builder.base_from_scratch):
```

mlangsdorf: I can make the change, but we're currently only running pre_add_filesystem for custom base image builds, not scratch builds.  And from what rcerven said when I asked about setting self.parent_images[] in InsideBuilder, I thought that scratch didn't get add_filesystem.

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -73,13 +73,15 @@ def __init__(self, tasker, workflow, koji_hub=None, koji_ssl_certs_dir=None,
...
     def run(self):
-        if not self.workflow.builder.base_from_scratch:
-            nvr = self._base_image_nvr = self.detect_parent_image_nvr(
-                self.workflow.builder.base_image,
-                inspect_data=self.workflow.builder.base_image_inspect,
-            )
-            if nvr:
-                self._base_image_build = self.wait_for_parent_image_build(nvr)
+        if self.workflow.builder.base_from_scratch or self.workflow.builder.custom_base_image:
+            return
```
MartinBasti: Should we log INFO here that this plugin was skipped?

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -67,6 +67,9 @@ def run(self):
         """
         Pull parent images and retag them uniquely for this build.
         """
+        if self.workflow.builder.custom_base_image or self.workflow.builder.base_from_scratch:
+            return
```
MartinBasti: Info about skipping plugin here as well?

atomic_reactor/plugins/pre_koji_parent.py
rcerven: we don't want to skip this plugin for base_from_scratch, the only thing we are skipping in that case is checking base_image since that is 'FROM scratch' but we still want to perform check for other parents

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: we don't want to skip for base_from_scratch, there can still be some pullable parents

atomic_reactor/build.py
```diff
@@ -232,6 +233,7 @@ def recreate_parent_images(self):
 
     def set_base_image(self, base_image, parents_pulled=True, insecure=False):
         self.base_from_scratch = base_image_is_scratch(base_image)
+        self.custom_base_image = base_image_is_custom(base_image)
```
rcerven: this check should be done just once, before you set_base_image again with add_filesystem plugin to newly created image, or you should check if it was set already to True, you won't override it

it works for base_from_scratch because we are not calling set_base_image if base image is 'scratch'

mlangsdorf: Fixed.

atomic_reactor/plugins/pre_pull_base_image.py
Not sure how to test it, though.

atomic_reactor/plugins/pre_koji_parent.py
mlangsdorf: Fixed.

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: you can just test that when you set workflow.builder.custom_base_image
you will get that log message, and maybe that it won't run set_base_image

atomic_reactor/plugins/pre_add_filesystem.py
rcerven: this is fine for now, I've created new story to actually support custom base image with multistage, as they aren't working now

atomic_reactor/plugins/pre_pull_base_image.py
mlangsdorf: Simple test case added.

atomic_reactor/plugins/pre_add_filesystem.py
mlangsdorf: Do you mean keep with the change I've proposed or change to Tim's suggestion?

rcerven: keep what you have

PR #1137, opened 2018-12-04
Reuse module validations from osbs-client

athos-ribeiro:


athos-ribeiro:
@rcerven, as requested in https://github.com/projectatomic/osbs-client/pull/832, we reuse the module validation code here.

This would require a bump on the minimum osbs-client version required for atomic-reactor (implies changes on RPM package spec file).

A better approach would be to have this check coming from an external source (e.g., ODCS) for both atomic-reactor and osbs-client.

thoughts?

twaugh:
Looks good, needs to bump requirements.txt and the spec file as you mentioned.

rcerven:
I think it is fine to bump minimal version, as for having check in ODCS, that would mean somebody would have to first do it, and another thing is it would be even slower to request additional things from ODCS than local check @twaugh what do you think?

athos-ribeiro:
BR osbs-client version updated in spec file.

There were no requirements  for osbs-client on requirements.txt nor on the python-atomic-reactor package (osbs-client was already required prior this PR).
Hence, those were not changed (I have seen other packager listing dependencies as optinal). If this is to be changed, I believe it would deserve another PR.

requremets-devel.txt does require osbs-client on master branch, so development environments should not be affected.

athos-ribeiro:
CI tests break due to latest osbs-client changes. https://github.com/projectatomic/atomic-reactor/pull/1138 fixes them

rcerven:
you should update as well osbs version at lines 149, 208

athos-ribeiro:
CI fails because `test.sh` tries to install packages in BuildRequires with `dnf builddep`. Since it cannot find the required version of osbs-client, it failt.

IMO, we should not rely on distributed versions of **our** packages for CI. Any ideas on how to work around that other than using sed to remove osbs-client?

@rcerven, the subpackages on those lines do not require the new osbs-client for this change. Should we change it anyway?

athos-ribeiro:
We are merging this without the spec file bumps until a new osbs-client version is released (January). We should update the spec file with new minumum osbs-client versions by then.

athos-ribeiro:
Internal changes only. No Release notes added.

PR #1139, opened 2018-12-05
Custom base images with multistage

rcerven:


atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -358,21 +358,47 @@ def run_image_task(self, image_build_conf):
...
 
-        new_base_image = self.import_base_image(filesystem)
-        self.workflow.builder.set_base_image(new_base_image)
-        defer_removal(self.workflow, new_base_image)
+        new_parent_image = self.import_base_image(filesystem)
+        new_imagename = ImageName.parse(new_parent_image)
 
-        return new_base_image
+        if self.workflow.builder.custom_base_image:
+            self.workflow.builder.set_base_image(new_imagename)
```
twaugh: Feels like we should know whether we need to call set_base_image without referring to custom_base_image.

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -358,21 +358,47 @@ def run_image_task(self, image_build_conf):
...
 
-    def run(self):
+        # setting the same new image for all custom base images
```
twaugh: Is that right though?

atomic_reactor/build.py
```diff
@@ -157,6 +157,8 @@ def __init__(self, source, image, **kwargs):
         self.built_image_info = None
         self.image = ImageName.parse(image)
         self.base_from_scratch = False
+        self.custom_base_image = False
+        self.custom_parent_image = False
```
twaugh: Would be great to have comments explaining what these mean -- it isn't immediately obvious to me from the names.

rcerven: yes I'll add comments

rcerven:
the added custom_parent_image, identifies whether any parent_image is custom base image,
custom_base_image only identifies if last parent is custom

but now when plugins have to decide whether they run or not, we have 2 cases,
1st plugins like inject_parent_image, is based on last parent only, therefor we check custom_base_image
2nd plugin like add_filesystem has to run when ANY parent is custom base image therefor we use custom_parent_image, also for this plugin based on those 2 options, we want to either set_base (change base image) OR only set parent_images

custom base image has feature that when you specify tag, that tag is used as configuration file for custom base image, but that would mean it would be possible to have multiple different custom base images in the same Dockerfile which would : 1st take a long time, 2nd there shouldn't be need for that, so I've initially wanted to use last configuration for custom base image specified, but I'll change that, so when there are multiple different custom base images specified in Dockerfile, it will raise error

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -358,21 +358,46 @@ def run_image_task(self, image_build_conf):
...
-        defer_removal(self.workflow, new_base_image)
+        new_parent_image = self.import_base_image(filesystem)
+        new_imagename = ImageName.parse(new_parent_image)
 
-        return new_base_image
+        if self.workflow.builder.custom_base_image:
+            self.workflow.builder.set_base_image(new_imagename)
 
-    def run(self):
+        for parent in self.workflow.builder.parent_images.keys():
```
mlangsdorf: So if this a custom base image, we remove any instances of the custom image name in the parent_images as well as the current base_image, but we don't reset custom_base_image?  I think that makes sense, because other plugins need to know that this was a a custom base image, I just want to make sure that is the intent.

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -358,21 +358,46 @@ def run_image_task(self, image_build_conf):
...
+                    image_build_conf = parent_image.tag
+                    break
+
         if not image_build_conf or image_build_conf == 'latest':
             image_build_conf = 'image-build.conf'
+        return image_build_conf
+
+    def run(self):
+        if not self.workflow.builder.custom_parent_image:
+            self.log.info('Non-Custom base images are not supported')
```
mlangsdorf: Change this to "Nothing to do for a non-custom base image" because "not supported" makes it seem like there is a problem, while "nothing to do" conveys that the plugin ran and returned immediately and successfully.

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -82,6 +83,9 @@ def run(self):
                 self._base_image_build = self.wait_for_parent_image_build(nvr)
 
         for img, local_tag in self.workflow.builder.parent_images.items():
+            if base_image_is_custom(img.to_str()):
```
mlangsdorf: Since add_filesystem runs before koji_parent, isn't this a null-op?  Or is it possible to have an non-custom base image and a custom base image for a parent?

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -69,17 +69,16 @@ def run(self):
...
-            return
-
         build_json = get_build_json()
         current_platform = platform.processor() or 'x86_64'
         self.manifest_list_cache = {}
         organization = get_registries_organization(self.workflow)
 
         for nonce, parent in enumerate(sorted(self.workflow.builder.parent_images.keys(),
                                               key=str)):
+            if base_image_is_custom(parent.to_str()):
```
mlangsdorf: Same question as above.

atomic_reactor/plugins/pre_add_filesystem.py
rcerven: we are not removing anything from parent_images, but updating value (new image)
for that custom base image, and yes we don't reset custom_base_image or custom_parent_image, as those are flags for other plugins to handle those custom base images

atomic_reactor/plugins/pre_add_filesystem.py
rcerven: ok

atomic_reactor/plugins/pre_koji_parent.py
rcerven: yes we can have for example
FROM koji/image-builder
FROM fedora

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: the same answer :)

atomic_reactor/plugins/pre_add_filesystem.py
rcerven: removed comment and changing just 1st, since it will raise error in build.py if there are 2+ different ones

atomic_reactor/plugins/pre_add_filesystem.py
rcerven: we have to call set_base_image only when custom base image is the last parent, in other cases when custom base image isn't last, we want to change just parent_images

rcerven:
ready for another review

atomic_reactor/build.py
```diff
@@ -187,15 +192,23 @@ def set_df_path(self, path):
...
         logger.debug("base image specified in dockerfile = '%s'", self.base_image)
         self.parent_images.clear()
+        custom_base_images = set()
         for image in dfp.parent_images:
             image_name = ImageName.parse(image)
             if base_image_is_scratch(image_name.get_repo()):
                 image_name.tag = None
                 self.parents_ordered.append(image_name.to_str())
                 continue
+            if base_image_is_custom(image_name.to_str()):
```
MartinBasti: just nitpick, because `image_name.to_str()` is used 4 times in the while cycle, couldn't be better to convert it once into extra var and use that?

rcerven:
rn upated: - Custom base images now properly work with multistage

PR #1123, opened 2018-11-06
Enable F29 testing

MartinBasti:
Dropping F27 as it is close to EOL and we don't expect another release
there.

Signed-off-by: Martin Bašti <mbasti@redhat.com>

PR #1142, opened 2018-12-12
Support python 3.7 BaseException repr

athos-ribeiro:
Python 3.7 changed BaseException repr to not include a trailing comma.

* See https://docs.python.org/3/whatsnew/3.7.html for reference

With the pytest 4 support changes, this should be the final piece to run CI in Fedora 29, as proposed in #1123 

### DIFF
--- a/tests/test_inner.py
+++ b/tests/test_inner.py
@@ -1238,19 +1238,22 @@ def test_cancel_build(request, fail_at):
                                                       'watcher': watch_exit
                                                   }}],
                                    plugin_files=[this_file])
+    # BaseException repr does not include trailing comma in Python >= 3.7
+    # we look for a partial match in log strings for Python < 3.7 compatibility
+    expected_entry = (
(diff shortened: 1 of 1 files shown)


tests/test_inner.py
```diff
@@ -1238,19 +1238,21 @@ def restore_logger():
                                                       'watcher': watch_exit
                                                   }}],
                                    plugin_files=[this_file])
+    # BaseException repr does not include trailing comma in Python >= 3.7
+    # we look for a partial match in log strings for Python < 3.7 compatibility
+    expected_entry = "plugin '{}_watched' raised an exception:" + \
```
MartinBasti: pep8 (and I) discourages using line escaping in favor of parentheses and concatenation of constant strings using + is not the nicest too.
Could be this rewritten to:
```python
expected_entry = (
    "plugin '{}_watched' raised an exception: "
    "BuildCanceledException('Build was canceled'"
)
```

PR #1144, opened 2018-12-12
import image should use registries and not source_registry for images…

rcerven:
…tream


twaugh:
This doesn't seem right. We pull images from the source registry, so that's what we should be monitoring.

rcerven:
import image is importing image which was just build, and that image was build into dest registry, and not source registry, so in Imagestream for just build image, has to be dest registry

twaugh:
But it is the image trigger for a build config which will pull it from the source registry.

I think part of the reason for the split between registries and source registry is that we can push to multiple places but must pull from one -- as well as Pulp handling being different.

PR #1145, opened 2018-12-13
 Request registry push access when grouping manifests

twaugh:
For OAuth2-aware registries we need to request push access when obtaining a bearer token we intend to use to create manifest list objects.


MartinBasti:
Release notes (bugfix section):

> - Request registry push access when grouping manifests

PR #1146, opened 2018-12-13
don't enclose base_image and parent_images when image is custom base …

rcerven:
…image


rcerven:
we can't enclose custom base image, since that is just placeholder for us, and builder.parent_images has unmodified koji/image-build, so if we enclose it in change_in_df plugin, it will fail with parentimagemissing, because koji/image-build != organization/koji-image-build

twaugh:
Would be nice to have a test case demonstrating it.

rcerven:
added test case for that

rcerven:
rn updated: - Custom base images work with organization defined

PR #1148, opened 2018-12-18
metadata for scratch

rcerven:

### DIFF
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -11,6 +11,8 @@ from __future__ import unicode_literals
 import json
 import os
 import time
+import logging
+from tempfile import NamedTemporaryFile
 
 from atomic_reactor import start_time as atomic_reactor_start_time
 from atomic_reactor.plugin import ExitPlugin
@@ -40,12 +42,11 @@ except ImportError:
 from atomic_reactor.constants import (
     PLUGIN_KOJI_IMPORT_PLUGIN_KEY, PLUGIN_PULP_PULL_KEY, PLUGIN_PULP_SYNC_KEY,
     PLUGIN_FETCH_WORKER_METADATA_KEY, PLUGIN_GROUP_MANIFESTS_KEY, PLUGIN_RESOLVE_COMPOSES_KEY,
-    PLUGIN_VERIFY_MEDIA_KEY
-)
+    PLUGIN_VERIFY_MEDIA_KEY, METADATA_TAG)
 from atomic_reactor.util import (Output, get_build_json,
                                  df_parser, ImageName, get_primary_images,
                                  get_manifest_media_type,
-                                 get_digests_map_from_annotations)
+                                 get_digests_map_from_annotations, is_scratch_build)
 from atomic_reactor.koji_util import (KojiUploadLogger, get_koji_task_owner)
 from atomic_reactor.plugins.pre_reactor_config import get_koji_session
 from osbs.utils import Labels
@@ -251,8 +252,12 @@ class KojiImportPlugin(ExitPlugin):
                 version_release = image.tag
                 break
 
-        assert version_release is not None, 'Unable to find version-release image'
-        tags = [image.tag for image in primary_images]
+        if is_scratch_build():
+            tags = [image.tag for image in self.workflow.tag_conf.images]
+            version_release = tags[0]
+        else:
+            assert version_release is not None, 'Unable to find version-release image'
+            tags = [image.tag for image in primary_images]
 
         manifest_list_digests = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY)
         if manifest_list_digests:
@@ -450,6 +455,22 @@ class KojiImportPlugin(ExitPlugin):
         self.log.debug("uploaded %r", path)
         return path
 
+    def upload_scratch_metadata(self, koji_metadata, koji_upload_dir, koji_session):
+        metadata_file = NamedTemporaryFile(prefix="metadata", suffix=".json", mode='wb')
+        metadata_file.write(json.dumps(koji_metadata, indent=2).encode('utf-8'))
+        metadata_file.flush()
+
+        filename = "metadata.json"
+        meta_output = Output(file=metadata_file, metadata={'filename': filename})
+
+        try:
+            self.upload_file(koji_session, meta_output, koji_upload_dir)
+            path = os.path.join(koji_upload_dir, filename)
+            log = logging.LoggerAdapter(self.log, {'arch': METADATA_TAG})
+            log.info(path)
+        finally:
+            meta_output.file.close()
+
     def run(self):
         """
         Run the plugin.
@@ -464,6 +485,10 @@ class KojiImportPlugin(ExitPlugin):
         server_dir = get_koji_upload_dir(self.workflow)
 
         koji_metadata, output_files = self.combine_metadata_fragments()
+
+        if is_scratch_build():
+            self.upload_scratch_metadata(koji_metadata, server_dir, self.session)
+            return
 
         try:
             for output in output_files:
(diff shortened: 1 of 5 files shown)


rcerven:
still missing tests

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -530,6 +532,26 @@ def upload_file(self, session, output, serverdir):
...
+                                           suffix=".json",
+                                           mode='wb')
+        metadata_file.write(json.dumps(koji_metadata))
+        metadata_file.flush()
+
+        filename = "{}-metadata.json".format(self.platform)
+        meta_output = Output(file=metadata_file,
+                             metadata=self.get_output_metadata(metadata_file.name, filename))
+
+        try:
```
MartinBasti: how about
```python
if meta_output.file:
    try:
        session = ...
        self.upload_file(....)
    finally:
        meta_output.file.close()
```

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -450,6 +456,40 @@ def upload_file(self, session, output, serverdir):
...
+        return metadata
+
+    def upload_scratch_metadata(self, koji_metadata, koji_upload_dir, koji_session):
+        metadata_file = NamedTemporaryFile(prefix="metadata", suffix=".json", mode='wb')
+        metadata_file.write(json.dumps(koji_metadata))
+        metadata_file.flush()
+
+        filename = "metadata.json"
+        meta_output = Output(file=metadata_file,
+                             metadata=self.get_output_metadata(metadata_file.name, filename))
```
twaugh: We don't need or use the checksum or filesize for this file, so perhaps we can just say `metadata={'filename': filename}` and drop `get_output_metadata`.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -450,6 +456,40 @@ def upload_file(self, session, output, serverdir):
...
+
+        if meta_output.file:
```
twaugh: It will also be truthy won't it? We just created the file.

atomic_reactor/plugins/exit_koji_import.py
rcerven: ok

atomic_reactor/plugins/exit_koji_import.py
rcerven: right, left over from loop :)

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -450,6 +456,40 @@ def upload_file(self, session, output, serverdir):
...
+            try:
+                self.upload_file(koji_session, meta_output, koji_upload_dir)
+                path = os.path.join(koji_upload_dir, filename)
+                self.log = logging.LoggerAdapter(self.log, {'arch': METADATA_TAG})
+                self.log.info(path)
+                self.log = logging.LoggerAdapter(self.log, {})
```
twaugh: Wonder if resetting the LoggerAdapter should move into the `finally` clause.

rcerven: yeah good idea

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -251,8 +252,12 @@ def set_group_manifest_info(self, extra, worker_metadatas):
                 version_release = image.tag
                 break
 
-        assert version_release is not None, 'Unable to find version-release image'
-        tags = [image.tag for image in primary_images]
+        if is_scratch_build():
+            tags = [image.tag for image in self.workflow.tag_conf.images]
+            version_release = tags[0]
+        else:
+            assert version_release is not None, 'Unable to find version-release image'
```
MartinBasti: Why this assert? (Should it be really assert)

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -450,6 +455,23 @@ def upload_file(self, session, output, serverdir):
...
+        metadata_file.write(json.dumps(koji_metadata).encode('utf-8'))
+        metadata_file.flush()
+
+        filename = "metadata.json"
+        meta_output = Output(file=metadata_file, metadata={'filename': filename})
+
+        try:
+            self.upload_file(koji_session, meta_output, koji_upload_dir)
+            path = os.path.join(koji_upload_dir, filename)
+            self.log = logging.LoggerAdapter(self.log, {'arch': METADATA_TAG})
```
MartinBasti: This is weird IMHO, with adapter there is no need to assign it to `self.log` and restoring it in finally:
```python
log = logging.LoggerAdapter(self.log, {'arch': METADATA_TAG})
log.info(path)
```

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -465,6 +487,10 @@ def run(self):
 
         koji_metadata, output_files = self.combine_metadata_fragments()
 
+        if is_scratch_build():
+            self.upload_scratch_metadata(koji_metadata, server_dir, self.session)
+            return
```
MartinBasti: Is this `return` ok here? (I don't think so)

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -541,15 +541,16 @@ def run(self):
...
-        try:
-            session = get_koji_session(self.workflow, self.koji_fallback)
-            for output in output_files:
-                if output.file:
-                    self.upload_file(session, output, self.koji_upload_dir)
-        finally:
-            for output in output_files:
-                if output.file:
-                    output.file.close()
+        if not is_scratch_build():
```
MartinBasti: Is this `if` needed? Why is not here uploading scratch metadata as in exit plugin? (This part is copy pasted, the same code)

atomic_reactor/plugins/exit_koji_import.py
MartinBasti: How is actually that adapter supposed working here?

MartinBasti: Found it, `ArchFormatter` uses that

rcerven: yeah we are setting different arch tag which is used by ArchFormatter, to identify metadata log line

atomic_reactor/plugins/exit_koji_import.py
rcerven: this plugin should do just 1 thing for scratch build and that is to upload that metadata file, so after that we are done for scratch

atomic_reactor/plugins/exit_koji_import.py
rcerven: that was there already before, well it shouldn't ever happen

MartinBasti: Ah sorry missed that

atomic_reactor/plugins/exit_koji_import.py
MartinBasti: How about that `build_info` later? Don't we want to still log it?

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -450,6 +455,23 @@ def upload_file(self, session, output, serverdir):
         self.log.debug("uploaded %r", path)
         return path
 
+    def upload_scratch_metadata(self, koji_metadata, koji_upload_dir, koji_session):
+        metadata_file = NamedTemporaryFile(prefix="metadata", suffix=".json", mode='wb')
+        metadata_file.write(json.dumps(koji_metadata).encode('utf-8'))
```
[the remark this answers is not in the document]
MartinBasti: changing this to must, because non-scratch metadata have indentation=2

atomic_reactor/plugins/exit_koji_import.py
rcerven: that is again for build and output from cgimport, but we are doing scratch and we don't have build neither are importing anything

MartinBasti: ok

atomic_reactor/plugins/exit_koji_import.py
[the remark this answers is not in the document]
rcerven: indent added

atomic_reactor/plugins/exit_koji_import.py
rcerven: changed

rcerven:
rn updated: - upload metadata for scratch builds

PR #1150, opened 2019-01-02
Handle 'auth' in dockercfg credentials

### LINKED ISSUE #1143: RegistrySession doesn't handle 'auth' in dockerconfigjson
The RegistrySession class expects 'username' and 'password' to be present for the registry in the Dockercfg credentials. However, 'auth' (instead of 'username' or 'password') is also valid.

@rcerven @MartinBasti

athos-ribeiro:
Allow RegistrySession to set the 'auth' key from the dockerconfig json
file directly in the authorization header, rather than supporting only
the username/password keys combination for authentication.
username/password format is still supported, but when all keys are
present, 'auth' value is used.

* Closes #1143

### DIFF
--- a/atomic_reactor/auth.py
+++ b/atomic_reactor/auth.py
@@ -22,6 +22,9 @@ class HTTPBearerAuth(AuthBase):
     when fetching the Bearer token from realm. Otherwise, Bearer token
     is retrivied with anonymous access.
 
+    auth_b64 may be provided for authentication (instead of username and
+    password).
+
     Once Bearer token is retrieved, it will be cached and used in subsequent
     requests. Since tokens are specific to repositories, the token cache may
     store multiple tokens.
@@ -31,7 +34,7 @@ class HTTPBearerAuth(AuthBase):
     BEARER_PATTERN = re.compile(r'bearer ', flags=re.IGNORECASE)
     V2_REPO_PATTERN = re.compile(r'^/v2/(.*)/(manifests|tags|blobs)/')
 
-    def __init__(self, username=None, password=None, verify=True, access=None):
+    def __init__(self, username=None, password=None, verify=True, access=None, auth_b64=None):
         """Initialize HTTPBearerAuth object.
 
         :param username: str, username to be used for authentication
@@ -41,9 +44,11 @@ class HTTPBearerAuth(AuthBase):
         :param access: iter<str>, iterable (list, tuple, etc) of access to be
             requested; possible values to be included are 'pull' and/or 'push';
             defaults to ('pull',)
+        :param auth_b64: str, base64 credendials as described in RFC 7617
         """
         self.username = username
         self.password = password
+        self.auth_b64 = auth_b64
         self.verify = verify
         self.access = access or ('pull',)
 
@@ -98,7 +103,9 @@ class HTTPBearerAuth(AuthBase):
         realm = bearer_info.pop('realm')
 
         realm_auth = None
-        if self.username and self.password:
+        if self.auth_b64:
+            realm_auth = HTTPBasicAuthWithB64(self.auth_b64)
+        elif self.username and self.password:
             realm_auth = HTTPBasicAuth(self.username, self.password)
 
         realm_response = requests.get(realm, params=bearer_info, verify=self.verify,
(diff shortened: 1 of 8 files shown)


atomic_reactor/auth.py
```diff
@@ -146,12 +173,15 @@ def __call__(self, request):
         url_parts = urlparse(request.url)
 
         if self.V1_URL.search(url_parts.path):
-            if not self.username or not self.password:
+            if (not self.username or not self.password) and not self.auth_b64:
                 # V1 API only supports basic auth which requires user/pass
```
MartinBasti: Is this comment still valid? IIUC we cannot use `auth` with v1

MartinBasti: Ahh nevermind, it is still basic auth

athos-ribeiro: I believe the comment is valid, although V1 only API support has been dropped in osbs-client. I believe that this whole code snippet will have to go at some point.

atomic_reactor/auth.py
```diff
@@ -146,12 +173,15 @@ def __call__(self, request):
...
                 return request
 
             if not self.v1_auth:
-                self.v1_auth = HTTPBasicAuth(self.username, self.password)
+                if self.auth_b64:
```
MartinBasti: Could we use this logic just one?

```python
if self.auth_b64:
    basic_auth = HTTPBasicAuthWithB64(self.auth_b64)
elif self.username and self.password:
    basic_auth =  HTTPBasicAuth(self.username, self.password)
else:
    basic_auth = None

...

### V1
if not self.v1_auth and basic_auth:
    self.v1_auth = basic_auth

### V2
....
if basic_auth:
    self.v2_auths.append(basic_auth)

```

athos-ribeiro: Nice catch. Thanks!
Done. Also rebased on master.

athos-ribeiro:
In an offline conversation, @MartinBasti pointed out that a plugin (tag_and_push) still requires 'password'.

atomic_reactor.core also requires 'username' to login to the registry.

While extracting those from the encoded auth is straightforward, I need to check if the supported versions of docker-py also accept 'auth'.

twaugh:
The tag_and_push plugin needs username and password in order to pass them to skopeo, if skopeo is used (i.e. for flatpaks). Think we'll need to decode auth to get those.

The core module uses docker.APIClient.login(), and that method allows setting a dockercfg_path parameter instead of username+password, which might be the way forward there.

tests/test_auth.py
```diff
@@ -262,12 +286,13 @@ def auth_callback(request):
...
 
-        auth = HTTPRegistryAuth(username=username, password=password)
+        auth = HTTPRegistryAuth(username=username, password=password, auth_b64=auth_b64)
         assert requests.get(repo_url, auth=auth).json() == 'success'
-        if username and password:
+        if (username and password) or auth_b64:
             assert len(auth.v2_auths) == 2
             assert isinstance(auth.v2_auths[0], HTTPBearerAuth)
-            assert isinstance(auth.v2_auths[1], HTTPBasicAuth)
+            assert isinstance(auth.v2_auths[1], HTTPBasicAuth) or isinstance(auth.v2_auths[1],
```
MartinBasti: Just pro tip:
```python
>>> isinstance(1, (int, None))
True
```

tests/test_tasker.py
```diff
@@ -15,13 +15,16 @@
...
+import os
+import json
 import docker
 import docker.errors
 import requests
 import sys
 import time
 import atomic_reactor
 from docker.errors import APIError
+from base64 import b64encode
```
MartinBasti: I know that we don't follow PEP8 module imports guidelines but I cannot resist :). This import should be first (and alphabetical order json before os)

```
# import stdlib modules

# import 3rd party modules

# import project modules
```

https://www.python.org/dev/peps/pep-0008/#imports

tests/test_tasker.py
```diff
@@ -483,3 +486,33 @@ def simplegen():
...
+        if 'auth' in dockerconfig_contents[LOCALHOST_REGISTRY]:
+            with pytest.raises(ValueError) as exc:
+                t.login(LOCALHOST_REGISTRY, tmpdir_path)
+                assert "Failed to parse 'auth'" in str(exc)
+        else:
+            with pytest.raises(RuntimeError) as exc:
+                t.login(LOCALHOST_REGISTRY, tmpdir_path)
+                assert "Failed to extract a username" in str(exc)
+    else:
+        t.login(LOCALHOST_REGISTRY, tmpdir_path)
```
MartinBasti: Could we test here if mocked object was called  with proper args?

tests/test_tasker.py
```diff
@@ -483,3 +486,38 @@ def simplegen():
...
+        if 'auth' in dockerconfig_contents[LOCALHOST_REGISTRY]:
+            with pytest.raises(ValueError) as exc:
+                t.login(LOCALHOST_REGISTRY, tmpdir_path)
+                assert "Failed to parse 'auth'" in str(exc)
+        else:
+            with pytest.raises(RuntimeError) as exc:
+                t.login(LOCALHOST_REGISTRY, tmpdir_path)
+                assert "Failed to extract a username" in str(exc)
+    else:
+        (fake_api
```
MartinBasti: If I read this correctly, `fake_api` may be undefined when `MOCK=False`

athos-ribeiro:
Release notes added:

- Allow dckercfg 'auth' to be used for authentication (instead of username and password)

PR #1151, opened 2019-01-07
[WIP] Add Buildah buildstep plugin

LorbusChris:
WIP


MartinBasti:
Thank you for contribution!

I just briefly read and tested this PR. I got following error:
```
2019-01-14 13:20:05,747 - atomic_reactor.inner - ERROR - buildstep plugin failed: image build failed (rc=1): 'overlay' is not supported over overlayfs: backing file system is unsupported for this graph driver
```

Related Issue: https://github.com/containers/buildah/issues/867

We have to check possibility of using something else than overlayfs for OSBS as a storage.

LorbusChris:
@MartinBasti I wasn't able to quickly figure out how to set up appropriate integration testing.
Afaict there is also still the hard dependency on docker, because the DockerTasker is used for all build plugins right now? Might not be required for this change, but would be a nice followup to use buildah for that, too.

atomic_reactor/constants.py
```diff
@@ -21,9 +21,14 @@
 CONTAINER_RESULTS_JSON_PATH = os.path.join(CONTAINER_SHARE_PATH, RESULTS_JSON)
 CONTAINER_DOCKERFILE_PATH = os.path.join(CONTAINER_SHARE_PATH, DOCKERFILE_FILENAME)
 
+CONTAINER_BUILDAH_BUILD_METHOD = 'buildah'
```
twaugh: Let's change this to `'buildah_bud'` to reflect the fact that only the 'buildah bud' functionality is available through this plugin.

atomic_reactor/plugins/build_buildah.py
```diff
@@ -0,0 +1,80 @@
...
+    Build image using buildah https://buildah.io/
+    This requires the buildah executable binary to be in $PATH.
+    """
+
+    key = CONTAINER_BUILDAH_BUILD_METHOD
+
+    def run(self):
+        """
+        Build image inside current environment using buildah;
+        It's expected this may run within (privileged) oci container.
```
twaugh: Does it need to be privileged?

atomic_reactor/plugins/build_buildah.py
```diff
@@ -0,0 +1,80 @@
...
+
+        Returns:
+            BuildResult
+        """
```
twaugh: There is quite a lot more work needed to make this functional, because 'buildah bud' creates an image available to podman but not docker.

We're planning to do the podman integration work soon, but for now can we make this build-step plugin raise `NotImplementedError` or something in `run()`? That way it won't get used accidentally.

LorbusChris:
@twaugh, thanks!

I just got back from PTO and we are pretty busy over at the MCO at the moment, so I don't really know when I'll be able to tackle this again.

If this is a priority now, do you think someone from CBS team can pick it up?

rcerven:
@LorbusChris will you have time to address those comments from @twaugh  ? if not I can create new PR, include your commit and address comments myself

LorbusChris:
@rcerven that would be great. Thank you!

rcerven:
@LorbusChris commit used in https://github.com/containerbuildsystem/atomic-reactor/pull/1246

PR #1153, opened 2019-01-10
Remove obsolete add_yum_repo plugin

athos-ribeiro:
The add_yum_repo plugin is not used anymore. Its code is no longer
functional due to changes in atomic_reactor's core and since it is
unmaintained, this commit removes it from the code base.


athos-ribeiro:
RN added:

- Removed obsolete add_yum_repo plugin

PR #1154, opened 2019-01-10
don't fail sendmail when all worker builds fail

rcerven:

### DIFF
--- a/tests/plugins/test_sendmail.py
+++ b/tests/plugins/test_sendmail.py
@@ -57,8 +57,21 @@ MOCK_KOJI_SUBMITTER_NAME = "baz"
 MOCK_KOJI_SUBMITTER_EMAIL = "baz@bar.com"
 MOCK_KOJI_SUBMITTER_GENERATED = "@".join([MOCK_KOJI_SUBMITTER_NAME, MOCK_EMAIL_DOMAIN])
 MOCK_ADDITIONAL_EMAIL = "spam@bar.com"
+MOCK_NAME_LABEL = 'foo/bar_in_df'
+MOCK_DOCKERFILE = ('FROM base\n'
+                   'LABEL Name={name}\n'
+                   .format(name=MOCK_NAME_LABEL))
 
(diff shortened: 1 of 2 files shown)


rcerven:
maybe I can add there in that case 'name' label from Dockerfile.

thoughts ? @twaugh  @MartinBasti  @athos-ribeiro

twaugh:
Yes, I'd rather we used the name label really. Ideally we'd include the git reference in that message, IMHO.

rcerven:
using name label for image, and also added git url & ref

tests/plugins/test_sendmail.py
```diff
@@ -490,28 +521,41 @@ class WF(object):
...
             'Status: ' + status,
             'Submitted by: ',
+            'Source url: ' + git_source_url,
+            'Source ref: ' + git_source_ref,
         ]
+        exp_body.extend(common_body)
+
         if autorebuild:
-            exp_body[-1] += '<autorebuild>'
+            exp_body[-3] += '<autorebuild>'
```
athos-ribeiro: It would be nice to refactor this to avoid using thise index here

tests/plugins/test_sendmail.py
```diff
@@ -490,28 +521,41 @@ class WF(object):
...
+            exp_body[-3] += SendMailPlugin.DEFAULT_SUBMITTER
 
         if log_url_cases[(koji_integration, autorebuild, success)]:
             if has_koji_logs:
-                exp_body.append("Logs: https://koji/work/tasks/12345")
+                exp_body.insert(-2, "Logs: https://koji/work/tasks/12345")
             else:
-                exp_body.append("Logs: "
-                                "https://something.com/builds/blablabla/log")
+                exp_body.insert(-2, "Logs: https://something.com/builds/blablabla/log")
```
athos-ribeiro: and here

rcerven:
rn updated: - don't fail sendmail plugin when all worker build fails

PR #1155, opened 2019-01-22
propagate logging level to osbs-client

rcerven:

### DIFF
--- a/atomic_reactor/cli/main.py
+++ b/atomic_reactor/cli/main.py
@@ -15,6 +15,7 @@ import pkg_resources
 import locale
 
 from atomic_reactor import set_logging
+from osbs import set_logging as set_logging_osbs
 from atomic_reactor.api import (build_image_here, build_image_in_privileged_container,
                                 build_image_using_hosts_docker)
 from atomic_reactor.constants import DESCRIPTION, PROG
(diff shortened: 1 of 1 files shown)


MartinBasti:
IMHO it fixes issue with logging level, but the whole logging setting looks fuzzy to me.

Why root logger is not configured? How config is propagated to the child loggers when they don't have  `atomic_reactor.*` prefix?

https://docs.python.org/3/howto/logging-cookbook.html

MartinBasti:
* logging is configured during import time of `set_logging method`, we shouldn't do that, loggin should be configured in main application.
* osbs client configures logging too, libraries should not configure logging, they even should have logging disabled by: https://docs.python.org/3/howto/logging.html#library-config (nto sure about py2)
* we don't configure root handler, every logger out of `osbs.*`/`atomic_reactor.*` patterns will pass log entries to root logger which will use default behavior (logs >=WARNING message to stderr), messages with lower loglevel will be discarded even if they come from atomic_reactor
* we don't enforce osbs/atomic_reactor prefix for sub-loggers, for example `atomic_reactor/util.py`. Log entries from that file may be lost, WARNING+ entries incorrectly formatted.
* nice example of configuring loggers: https://docs.python.org/3/howto/logging-cookbook.html#an-example-dictionary-based-configuration

MartinBasti:
I didn't want to block this PR, for logging changes we may new a separate issue. This PR works fine for now.

rcerven:
rn updated: - Propagate logging level to osbs-client

PR #1157, opened 2019-01-28
Check parent image koji build state

athos-ribeiro:
We perform this check to make sure the parent image koji build has not
been deleted due to garbage-collection deletions.


rcerven:
maybe we should use only 'COMPLETE' ?
and fail in other cases:
if build['state'] != koji.BUILD_STATES['COMPLETE']:

@twaugh @athos-ribeiro  what do you think?

twaugh:
That would probably be better actually, yes.

athos-ribeiro:
release note added
- Check if parent image koji build has not been removed

PR #1159, opened 2019-01-31
container.yaml schema: store top-level Go module name (OSBS-6848)

twaugh:
This will be used by image owners to declare the top-level Go package which will be built into the image from source.

### DIFF
--- a/atomic_reactor/schemas/container.json
+++ b/atomic_reactor/schemas/container.json
@@ -5,6 +5,34 @@
 
   "type": ["object", "null"],
   "properties": {
+    "go": {
+      "type": "object",
+      "properties": {
+        "modules": {
+          "type": ["array", "null"],
+          "items": {
+            "type": "object",
+            "properties": {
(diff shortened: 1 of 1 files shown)


twaugh:
I'd quite like to get that fixed, yes.

twaugh:
Marking as work-in-progress as there is investigation I need to carry out.

twaugh:
Re-reviews requested as the format has changed.

atomic_reactor/schemas/container.json
```diff
@@ -5,6 +5,33 @@
 
   "type": ["object", "null"],
   "properties": {
+    "go": {
```
athos-ribeiro: Any reason for not specifying the type for "go"?

twaugh: Oh, missed it. I'll add it.

twaugh: Fixed.

PR #1160, opened 2019-01-31
Fix flatpak tests

twaugh:
When not mocking the flatpak command, make sure its filetype-checking does not trip over our mocked data. In particular, PNG files must be valid.


athos-ribeiro:
Tests are still failing...
Just for the records, the whole test suite started failing after a flatpak upgrade in Fedora 29 [1] with a patch that validates exported icons [2].
There is also a follow-up update in f29-testing [3] to require librsvg2 to allow svg icons to be exported (does not seem to be related to our issue).

[1] https://bodhi.fedoraproject.org/updates/FEDORA-2019-fe463a4b9c
[2] https://github.com/flatpak/flatpak/commit/f52d00f3916199f92e8bf5f42274de97dd97ebf1
[3] https://bodhi.fedoraproject.org/updates/FEDORA-2019-809743b754

twaugh:
Yes, the updates-testing package doesn't help. I haven't been able to reproduce this locally yet, even though I'm running test.sh.

athos-ribeiro:
`/usr/libexec/flatpak-validate-icon --sandbox 512 512 /some_file.png` returns -1 in our container (it succeeds in my system).

From strace, I got `read(3, "bwrap: Creating new namespace failed: Operation not permitted\n", 4096)`

if we run the container with in privileged mode, we get `bwrap: Failed to mount tmpfs: Permission denied`

And then turning off selinux does the trick.

From flatpak code, I understood the proper way to deactivate the sandbox would be by setting the `DISABLE_SANDBOXED_TRIGGERS` macro [1].

[1] https://github.com/flatpak/flatpak/blob/master/app/flatpak-builtins-build-export.c#L377

athos-ribeiro:
I built [flatpak with sandbox disabled](https://copr.fedorainfracloud.org/coprs/athoscr/osbs/package/flatpak/) in COPR (athoscr/osbs). Installing flatpak from that repository will get the tests to pass.

```
[athos@eurus atomic-reactor]$ git diff
diff --git a/test.sh b/test.sh
old mode 100644
new mode 100755
index fed0de9..c86468a
--- a/test.sh
+++ b/test.sh
@@ -23,7 +23,7 @@ if [[ $OS == "fedora" ]]; then
   PIP="pip$PYTHON_VERSION"
   PKG="dnf"
   PKG_EXTRA="dnf-plugins-core desktop-file-utils flatpak libmodulemd ostree python$PYTHON_VERSION-gobject-base"
-  ENABLE_REPO="--enablerepo updates-testing"
+  ENABLE_REPO=
   BUILDDEP="dnf builddep"
   PYTHON="python$PYTHON_VERSION"
 else
@@ -49,6 +49,11 @@ if [[ $OS == "centos" ]]; then
   $RUN rm -f /etc/yum.repos.d/CentOS-Sources.repo
 fi
 
+# Enable osbs copr to fetch special flatpak package without sandboxing
+if [[ $OS == "fedora" && $OS_VERSION -ge 29 ]]; then
+  $RUN $PKG install -y dnf-plugins-core
+  $RUN $PKG copr enable -y athoscr/osbs
+fi
 # Install dependencies
 $RUN $PKG install -y $ENABLE_REPO $PKG_EXTRA
 [[ ${PYTHON_VERSION} == '3' ]] && WITH_PY3=1 || WITH_PY3=0
```

The problem we face is on a `flatpak build-export` run, which triggers a `flatpak-validate-icon` run. While the latter command does accept a `--sandbox` option to toggle sandboxing on/off, I understand that the former does not: it sets the `--sandbox` option unless flatpak was compiled with the `DISABLE_SANDBOXED_TRIGGERS` option.

It seems that the issue will also affect production environments, not only our test suite.

I am not sure about how we should approach this. Unless I am missing something about the flatpak runs, we can

* Advise users to use a custom flatpak package, compiled with the DISABLE_SANDBOXED_TRIGGERS macro defined.
* Change documentation to make it clear that the worker containers for flatpak builds must run in provileged mode and the environment must have certain selinux permissions on to allow sandboxing (needs more investigation on which permissions to set).
* Ask flatpak upstream if it would be feasible to allow the `build-export` command to accept a `--no-sandbox` option to be used for the flatpak-validate-icon run.

For now, I do not like any of these options. Thoughts?

athos-ribeiro:
The reason I did not like the last option (pathcing flatpak) was because I believed it would not be a feature wanted by flatpak upstream. Apparently, thay also needed taht for tests:

https://github.com/flatpak/flatpak/commit/ea227187d40a8194d389585b69a0120f782d1bde

For flatpak >= 1.2.1, we will be able to pass `--disable-sandbox` to `flatpak build-export` and the tests shall succeed.

twaugh:
What shall we do in the mean time? Would be nice to have the test suite passing so we can merge PRs.

* Disable F29 tests?
* Make those tests conditional on `--disable-sandbox` support?
* ...something else?

Also, is it within our control to pass the new option or is that up to another Python module?

athos-ribeiro:
For getting the test to pass, we should just pass the flag to the build-export command the relevant test.

For a production environment, we would need a change in flatpak_module_tools [1].

To fix out test suite, I would propose to install flatpak from github on top of https://github.com/flatpak/flatpak/commit/ea227187d40a8194d389585b69a0120f782d1bde and pass the proper flag to the build-export command. This way, the next atomicrreactor version will not be compatible with flatpak <= 1.2.0 and we will need the changes in flatpak_module_tools.

Ccing @owtaylor for visibility.

[1]  https://pagure.io/flatpak-module-tools/blob/master/f/flatpak_module_tools/flatpak_builder.py#_700

twaugh:
I don't really mind either way for updates-testing.

owtaylor:
Cc'ing @alexlarsson in case he has any insights, but basically, I think basically the diagnosis is right: we need to get a build of Flatpak into F29 with `flatpak build-export --disable-sandbox` and a build of flatpak-module-tools that uses that when available. And the F29 test can be disabled until then. I think only small modications to atomic-reactor are needed - just making sure that tests/plugins/test_flatpak_create_oci.py still works (there's some argument for removing most of that test - I left it there for the transition, but it really depends on exact details of flatpak-module-tools that atomic-reactor shouldn't need to care about.)

alexlarsson:
For the record, 1.2.1 is out now, so should be possible to disable the sandbox here now.

PR #1164, opened 2019-02-05
Don't run pytest in verbose mode

MartinBasti:
We cannot read verbose output nicely in travis output because it is trimmed.


twaugh:
Would be nice to rebase so we can see it pass first.

MartinBasti:
Rebased

MartinBasti:
no release notes, change in testing only

PR #1168, opened 2019-02-11
retry push when v2 s2 or oci is missing

rcerven:

### DIFF
--- a/atomic_reactor/constants.py
+++ b/atomic_reactor/constants.py
@@ -118,6 +118,10 @@ DOCKER_MAX_RETRIES = 3
 DOCKER_BACKOFF_FACTOR = 5
 # docker retries statuses
 DOCKER_CLIENT_STATUS_RETRY = (408, 500, 502, 503, 504)
+# max retries for docker push
+DOCKER_PUSH_MAX_RETRIES = 6
+# how many seconds should wait before another try of docker push
+DOCKER_PUSH_BACKOFF_FACTOR = 5
 # max retries for http requests
 HTTP_MAX_RETRIES = 3
 # how many seconds should wait before another try of http request
--- a/atomic_reactor/plugins/post_tag_and_push.py
+++ b/atomic_reactor/plugins/post_tag_and_push.py
@@ -9,11 +9,14 @@ of the BSD license. See the LICENSE file for details.
 from copy import deepcopy
 import re
 import subprocess
+import time
 
-from atomic_reactor.constants import IMAGE_TYPE_DOCKER_ARCHIVE, IMAGE_TYPE_OCI, IMAGE_TYPE_OCI_TAR
+from atomic_reactor.constants import (IMAGE_TYPE_DOCKER_ARCHIVE, IMAGE_TYPE_OCI, IMAGE_TYPE_OCI_TAR,
+                                      MEDIA_TYPE_DOCKER_V2_SCHEMA2, DOCKER_PUSH_MAX_RETRIES,
+                                      DOCKER_PUSH_BACKOFF_FACTOR)
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.plugins.exit_remove_built_image import defer_removal
-from atomic_reactor.plugins.pre_reactor_config import get_registries
+from atomic_reactor.plugins.pre_reactor_config import get_registries, get_group_manifests
 from atomic_reactor.util import (get_manifest_digests, get_config_from_registry, Dockercfg)
 
 
@@ -46,6 +49,8 @@ class TagAndPushPlugin(PostBuildPlugin):
         super(TagAndPushPlugin, self).__init__(tasker, workflow)
 
         self.registries = get_registries(self.workflow, deepcopy(registries or {}))
+        self.group = get_group_manifests(self.workflow, False)
+
 
     def need_skopeo_push(self):
         if len(self.workflow.exported_image_sequence) > 0:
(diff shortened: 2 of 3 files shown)


atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -136,16 +137,25 @@ def run(self):
                 registry_image.registry = registry
                 if self.need_skopeo_push():
                     self.push_with_skopeo(registry_image, insecure, docker_push_secret)
+                    digests = get_manifest_digests(registry_image, registry,
+                                                   insecure, docker_push_secret)
```
lcarva: Does it make sense to retry for skopeo push also?

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -136,16 +137,25 @@ def run(self):
...
-                    self.tasker.tag_and_push_image(self.workflow.builder.image_id,
-                                                   registry_image, insecure=insecure,
-                                                   force=True, dockercfg=docker_push_secret)
+                    for retry in range(DOCKER_MAX_RETRIES):
+                        self.tasker.tag_and_push_image(self.workflow.builder.image_id,
+                                                       registry_image, insecure=insecure,
+                                                       force=True, dockercfg=docker_push_secret)
+                        digests = get_manifest_digests(registry_image, registry,
+                                                       insecure, docker_push_secret)
+                        if (not (digests.v2 or digests.oci) and (retry + 1 != DOCKER_MAX_RETRIES)):
```
lcarva: Should we check if v2 schema 2 is desired based on configuration?

atomic_reactor/plugins/post_tag_and_push.py
rcerven: yeah as we don't know if skopeo will fail also, we can include it in retry

atomic_reactor/plugins/post_tag_and_push.py
rcerven: yeah I'll check if group_manifest are set and also if v2-2 is expected

tests/plugins/test_tag_and_push.py
```diff
@@ -258,11 +269,34 @@ def custom_get(method, url, headers, **kwargs):
         if url == config_blob_url:
             return config_blob_response
 
+    def custom_get_v2missing(method, url, headers, **kwargs):
+        if url == manifest_latest_url:
+            # For a manifest stored as v2 or v1, the docker registry defaults to
+            # returning a v1 manifest if a v2 manifest is not explicitly requested
+            if headers['Accept'] == 'application/vnd.docker.distribution.manifest.v2+json':
+                return manifest_unknown_response
```
athos-ribeiro: `manifest_unknown_response` is the only value that changes between `custom_get` and `custom_get_v2missing`. Could we just edit `custom_get` to return `manifest_unknown_response` when needed if missing_v2==True ? This way we avoid the function duplication and some extra code below (line 292 to 299).

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -134,18 +137,38 @@ def run(self):
...
                 registry_image = image.copy()
                 registry_image.registry = registry
-                if self.need_skopeo_push():
-                    self.push_with_skopeo(registry_image, insecure, docker_push_secret)
-                else:
-                    self.tasker.tag_and_push_image(self.workflow.builder.image_id,
-                                                   registry_image, insecure=insecure,
-                                                   force=True, dockercfg=docker_push_secret)
-                    defer_removal(self.workflow, registry_image)
+                max_retries = DOCKER_MAX_RETRIES
```
athos-ribeiro: This is actually the number of tries, right? For instance, if one decides to set it to 0, no pushes would happen, when one would most likely be expecting 1 push attempt to happen.

Maybe we could create a new constant for this case (instead of change the logic in this patch) since DOCKER_MAX_RETRIES is used for something else in the `core` module and tuning it for one purpose could result in unexpected behavior somewhere else. Another option would be to set `max_tries` to 1+ max_retries or something like that, but then it would be nice to evaluate the consequences of changing DOCKER_MAX_RETRIES value. What do you think?

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -134,18 +137,38 @@ def run(self):
...
+                        self.push_with_skopeo(registry_image, insecure, docker_push_secret)
+                    else:
+                        self.tasker.tag_and_push_image(self.workflow.builder.image_id,
+                                                       registry_image, insecure=insecure,
+                                                       force=True, dockercfg=docker_push_secret)
+
+                    digests = get_manifest_digests(registry_image, registry,
+                                                   insecure, docker_push_secret)
+
+                    if (not (digests.v2 or digests.oci) and (retry + 1 != DOCKER_MAX_RETRIES)):
```
athos-ribeiro: Do we need this last part after the `and`? Could it lead to logging a retry that would never happen if max_retries is set to 1?

athos-ribeiro:
Looking good :+1:

Added a few inline comments.

Also, what do you think about having the reference to OSBS story in the message body (instead of in the title)? Maybe we should work on some pattern for that...

Something like
```
Title

* OSBS-XXXX

Signed-off-by: ...
```

?

tests/plugins/test_tag_and_push.py
rcerven: yop will do that

atomic_reactor/plugins/post_tag_and_push.py
rcerven: that last part is to prevent logging retry message for last cycle

rcerven: ahh right will change it to != max_retries

atomic_reactor/plugins/post_tag_and_push.py
rcerven: created new constants

athos-ribeiro:
CI is failing due to some pytest dependency new release. See https://github.com/projectatomic/atomic-reactor/pull/1169

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -134,18 +139,40 @@ def run(self):
...
+                                                       force=True, dockercfg=docker_push_secret)
+
+                    digests = get_manifest_digests(registry_image, registry,
+                                                   insecure, docker_push_secret)
+
+                    if (not (digests.v2 or digests.oci) and (retry != max_retries)):
+                        self.log.info("Retrying push because V2 schema 2 or "
+                                      "OCI manifest not found")
+
+                        time.sleep(DOCKER_PUSH_BACKOFF_FACTOR * (2 ** retry))
```
MartinBasti: IMO would be nice to log info here how many seconds it will be waiting

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -134,18 +139,40 @@ def run(self):
...
+                        self.push_with_skopeo(registry_image, insecure, docker_push_secret)
+                    else:
+                        self.tasker.tag_and_push_image(self.workflow.builder.image_id,
+                                                       registry_image, insecure=insecure,
+                                                       force=True, dockercfg=docker_push_secret)
+
+                    digests = get_manifest_digests(registry_image, registry,
+                                                   insecure, docker_push_secret)
+
+                    if (not (digests.v2 or digests.oci) and (retry != max_retries)):
```
MartinBasti: IMO it is safer to use `retry < max_retries` in case of accidental increment

atomic_reactor/constants.py
```diff
@@ -118,6 +118,10 @@
 DOCKER_BACKOFF_FACTOR = 5
 # docker retries statuses
 DOCKER_CLIENT_STATUS_RETRY = (408, 500, 502, 503, 504)
+# max retries for docker push
+DOCKER_PUSH_MAX_RETRIES = 6
+# how many seconds should wait before another try of docker push
```
MartinBasti: It is factor which multiplies seconds, I'd not say "how many seconds" in that comments, this comment is valid only for first try.

rcerven: yeah, I just kept the comments consistent with other backoff factors

atomic_reactor/plugins/post_tag_and_push.py
rcerven: sure

atomic_reactor/plugins/post_tag_and_push.py
rcerven: ok

atomic_reactor/constants.py
MartinBasti: ok

rcerven:
rn updated: - Retry push when v2 s2 or oci is missing and required

PR #1171, opened 2019-02-13
Verify manifest is available

athos-ribeiro:
~~This is a work in progress. Before working on tests, I would appreciate some feedback here. Is there a better way to make sure the manifests are available through digests other than querying the registry like this?~~
**EDIT:** Based on IRC feedback, querying the registries seems to be a valid approach.

- [x] tests
- [x] forward dockercfg and insecure params
- [x] update/create docstrings

Before uploading archive metadata on image manifests for
extra.docker.digests to koji, we query the registry to make sure the
manifests can indeed be fetched through its digest.

* OSBS-6849

### DIFF
--- a/atomic_reactor/plugins/post_koji_upload.py
+++ b/atomic_reactor/plugins/post_koji_upload.py
@@ -19,11 +19,11 @@ from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.plugins.post_rpmqa import PostBuildRPMqaPlugin
 from atomic_reactor.plugins.pre_reactor_config import (get_openshift_session,
                                                        get_prefer_schema1_digest,
-                                                       get_koji_session)
+                                                       get_koji_session, get_registries)
 from atomic_reactor.constants import PROG, PLUGIN_KOJI_UPLOAD_PLUGIN_KEY
 from atomic_reactor.util import (get_version_of_tools, get_checksums,
                                  get_build_json, get_docker_architecture,
-                                 get_image_upload_filename,
+                                 get_image_upload_filename, check_digest_availability,
                                  get_manifest_media_type, ImageName, is_scratch_build)
 from atomic_reactor.rpm_util import parse_rpm_output, rpm_qf_args
 from osbs.exceptions import OsbsException
@@ -334,12 +334,18 @@ class KojiUploadPlugin(PostBuildPlugin):
         """
         digests = {}  # image -> digests
         typed_digests = {}  # media_type -> digests
+        registries_conf = copy.deepcopy(get_registries(self.workflow, {}))
         for registry in self.workflow.push_conf.docker_registries:
+            registry_conf = registries_conf.get(registry.uri, {})
+            secret = registry_conf.get('secret', None)
             for image in self.workflow.tag_conf.images:
                 image_str = image.to_str()
                 if image_str in registry.digests:
                     image_digests = registry.digests[image_str]
-                    if self.report_multiple_digests:
+                    if (self.report_multiple_digests and
+                        check_digest_availability(image, registry.uri, image_digests,
+                                                  insecure=registry.insecure,
+                                                  dockercfg_path=secret)):
                         digest_list = [digest for digest in (image_digests.v1,
                                                              image_digests.v2)
                                        if digest]
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -995,6 +995,35 @@ def get_manifest(image, registry_session, version):
     return response, saved_not_found
 
 
+def check_digest_availability(image, registry, digests, versions=('v1', 'v2'),
+                              insecure=False, dockercfg_path=None):
+    """Return whether image manifests can be fetched by digest for the given schema versions.
+
+    :param image: ImageName, the remote image to inspect
+    :param registry: str, URI for registry
+    :param digests: dict, maps a tag (str) to a ManifestDigest instance
+    :param versions: tuple, manifest schema versions to be tested
+    :param insecure: bool, when True registry's cert is not verified
+    :param dockercfg_path: str, dirname of .dockercfg location
+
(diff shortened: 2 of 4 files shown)


docs/koji.md
```diff
@@ -97,7 +97,7 @@ The docker map has these entries:
 - `config` (map): the [v2 schema 2 'config' object](https://docs.docker.com/registry/spec/manifest-v2-2/#image-manifest-field-descriptions) but with the 'container_config' entry removed
 - `tags` (string list): the image tags (i.e. the part after the ":") applied to this image when it was tagged and pushed
 - `layer_sizes` (map list): the image layer uncompressed sizes, the oldest layer first (the size information comes from docker history command)
-- `digests` (map): a map of media type (such as “application/vnd.docker.distribution.manifest.v2+json”) to manifest digest (a string usually starting “sha256:”), for each available media type for which a digest is available.
+- `digests` (map): a map of media type (such as “application/vnd.docker.distribution.manifest.v2+json”) to manifest digest (a string usually starting “sha256:”), for each available media type which can be retrieved by digest (media types only reachable by tag should not be included).
```
twaugh: Could say `are not included` at the end there to be a bit more definitive.

twaugh: ...and now I realise you used the wording I originally suggested, verbatim. :-) Anyway, I think "are not included" is better.

atomic_reactor/util.py
```diff
@@ -995,6 +995,36 @@ def get_manifest(image, registry_session, version):
     return response, saved_not_found
 
 
+def check_digest_availability(image, registry, digests, versions=('v1', 'v2'),
+                              insecure=False, dockercfg_path=None):
+    """Return weather image manifests can be fetched by digest for the given schema versions.
```
twaugh: s/weather/whether/

atomic_reactor/util.py
```diff
@@ -995,6 +995,36 @@ def get_manifest(image, registry_session, version):
...
+    registry_session = RegistrySession(registry, insecure=insecure, dockercfg_path=dockercfg_path)
+    for version in versions:
+        try:
+            query_registry(registry_session, image, getattr(digests, version), version=version)
+        except (HTTPError, RetryError, Timeout) as ex:
+            if ex.response.status_code == requests.codes.not_found:
+                logger.debug("%s manifest for %s cannot be fetched from registry by digest %s",
+                             version, image.to_str(), getattr(digests, version))
+                return False
+            else:
```
twaugh: No need for the `else:` here.

atomic_reactor/util.py
```diff
@@ -995,6 +995,34 @@ def get_manifest(image, registry_session, version):
...
+
+    registry_session = RegistrySession(registry, insecure=insecure, dockercfg_path=dockercfg_path)
+    for version in versions:
+        try:
+            query_registry(registry_session, image, getattr(digests, version), version=version)
+        except (HTTPError, RetryError, Timeout) as ex:
+            if ex.response.status_code == requests.codes.not_found:
+                logger.debug("%s manifest for %s cannot be fetched from registry by digest %s",
+                             version, image.to_str(), getattr(digests, version))
+                return False
```
twaugh: Do we want to raise for other status codes though?

athos-ribeiro: Oh, right. I got rid of the whole block. Should be fixed now.

tests/plugins/test_koji_upload.py
```diff
@@ -1128,10 +1142,20 @@ def test_koji_upload_logs(self, tmpdir, monkeypatch, os_env, is_scratch, logs_re
             assert images[0].endswith(platform + ".tar.xz")
 
     @pytest.mark.parametrize('multiple', [False, True])
-    def test_koji_upload_multiple_digests(self, tmpdir, os_env,
-                                          multiple, reactor_config_map):
+    @pytest.mark.parametrize('manifest_available_with_digest', [False, True])
```
mlangsdorf: Is there a test case for the time-out error that I'm missing?

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -335,11 +335,19 @@ def get_repositories_and_digests(self):
         digests = {}  # image -> digests
         typed_digests = {}  # media_type -> digests
         for registry in self.workflow.push_conf.docker_registries:
+            registries = copy.deepcopy(get_registries(self.workflow, {}))
```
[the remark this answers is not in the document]
athos-ribeiro: Fixed. I also renamed the variable since there is another `registries` in the end of the method.

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -335,11 +335,19 @@ def get_repositories_and_digests(self):
...
+            registry_conf = registries.get(registry.uri, {})
+            secret = registry_conf.get('secret', None)
             for image in self.workflow.tag_conf.images:
                 image_str = image.to_str()
                 if image_str in registry.digests:
                     image_digests = registry.digests[image_str]
-                    if self.report_multiple_digests:
+                    if (self.report_multiple_digests
+                            and check_digest_availability(image,
```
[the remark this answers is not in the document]
athos-ribeiro: Fixed

tests/plugins/test_koji_upload.py
athos-ribeiro: I removed the code snippet where we were capturing and re-raising it from the `util` module (we were not dealing with it at all).

In this case, we will let RegistrySession raise the timeout after its retry logic since we will not be able to determine whether the resources are available through the digest or not.

mlangsdorf: `check_digest_availibility` still captures an exception and raises it under certain conditions.

mlangsdorf:
Not adding anything to release notes since this is an internal code change.

athos-ribeiro:
Thanks for the reviews!

PR #1172, opened 2019-02-20
resolve_composes: allow package and module composess in the same yaml

mlangsdorf:
Fixes OSBS-6823

There is no technical reason why resolve_composes can't support
package and module composes in the same yaml, so remove the
logic that prevents them.


twaugh:
Also can you please include the JIRA story number in the commit message?

mlangsdorf:
Added
"- composes can contain packages, modules, and pulp components instead of being limited to pulp and either packages or modules components"
to RN under Improvements.

PR #1173, opened 2019-02-21
Set archive metadata based based on pulp presence

athos-ribeiro:
We cannot query pulp registries in build time to verify the presence of
v1 and v2 manifests. Therefore, we assemble the archive metadata,
deciding whether v1 digests should be included, based on the presence of
pulp registries. We do this by reverting the changes introduced in
5a70132 and introducing the approach described.

* Reverts commit 5a701322ac2025e88768ea09536931d287cf3a75
* OSBS-6849

### DIFF
--- a/atomic_reactor/plugins/post_koji_upload.py
+++ b/atomic_reactor/plugins/post_koji_upload.py
@@ -19,11 +19,11 @@ from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.plugins.post_rpmqa import PostBuildRPMqaPlugin
 from atomic_reactor.plugins.pre_reactor_config import (get_openshift_session,
                                                        get_prefer_schema1_digest,
-                                                       get_koji_session, get_registries)
+                                                       get_koji_session)
 from atomic_reactor.constants import PROG, PLUGIN_KOJI_UPLOAD_PLUGIN_KEY
 from atomic_reactor.util import (get_version_of_tools, get_checksums,
                                  get_build_json, get_docker_architecture,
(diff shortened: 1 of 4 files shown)


twaugh:
Could we split the 'revert' commit from the 're-implement' commit?

athos-ribeiro:
> Looks great but could we split the 'revert' commit from the 're-implement' commit?

Done. No changes in final diff (except for a rebese on master).

rcerven:
can we make also this change ? https://github.com/projectatomic/atomic-reactor/pull/1171/files#diff-0cfb85c777f2f0984c266781a49f8478
but omit : "(media types only reachable by tag are not included)."

athos-ribeiro:
> can we make also this change ? https://github.com/projectatomic/atomic-reactor/pull/1171/files#diff-0cfb85c777f2f0984c266781a49f8478
> but omit : "(media types only reachable by tag are not included)."

This change is actually included. It may get confusing if we look only at the github file changes, since the change was merged, reverted, and reintroduced (so it will not show up in the diff if you do not check for the individual commits diff).

About omitting "(media types only reachable by tag are not included)": This is still true with this change, isn't it?

twaugh:
It is. We want all the digests we list in 'digests' to be valid refs when fetching the image. The v2 schema 1 image manifest is only reachable by tag in Quay.io (i.e. when using a docker client with no v2 schema 2 support), so would be omitted.

athos-ribeiro:
RN:
removed:
-  confirm that an image can be fetched from a registry through its digest before attempting to upload the archive metadata
added:
- only include manifest digests in archive metadata if they can be fetched by digest

PR #1174, opened 2019-02-21
inherit yum repos and composes from baseimage

rcerven:

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -394,6 +394,11 @@ class DockerBuildWorkflow(object):
         # List of RPMs that go into the final result, as per rpm_util.parse_rpm_output
         self.image_components = None
 
+        # List of all yum repos. The provided repourls might be changed (by resolve_composes) when
+        # inheritance is enabled. This property holds the updated list of repos, allowing
+        # post-build plugins (such as koji_import) to record them.
+        self.all_yum_repourls = None
+
         if client_version:
             logger.debug("build json was built by osbs-client %s", client_version)
 
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -382,6 +382,8 @@ class KojiImportPlugin(ExitPlugin):
                 'signing_intent': resolve_comp_result['signing_intent'],
                 'signing_intent_overridden': resolve_comp_result['signing_intent_overridden'],
             }
+        if self.workflow.all_yum_repourls:
+            extra['image']['yum_repourls'] = self.workflow.all_yum_repourls
 
         self.set_help(extra, worker_metadatas)
         self.set_media_types(extra, worker_metadatas)
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -21,7 +21,7 @@ from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
 from atomic_reactor.plugins.pre_reactor_config import (get_config,
                                                        get_odcs_session,
                                                        get_koji_session, get_koji)
-from atomic_reactor.util import get_platforms
+from atomic_reactor.util import get_platforms, is_isolated_build, is_scratch_build
 
 ODCS_DATETIME_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
 MINIMUM_TIME_TO_EXPIRE = timedelta(hours=2).total_seconds()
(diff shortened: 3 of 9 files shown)


rcerven:
osbs-client part https://github.com/projectatomic/osbs-client/pull/848

atomic_reactor/schemas/container.json
```diff
@@ -82,6 +82,9 @@
       },
       "additionalProperties": true
     },
+    "inherit_from_parent": {
```
twaugh: Should this be part of "compose"? It feels like it's confined to that area.

rcerven: problem is that it can also be present even jut to inherit yum repo, when compose isn't present

rcerven: so I've made it not dependent to compose

atomic_reactor/inner.py
```diff
@@ -394,6 +394,10 @@ def __init__(self, source, image, prebuild_plugins=None, prepublish_plugins=None
         # List of RPMs that go into the final result, as per rpm_util.parse_rpm_output
         self.image_components = None
 
+        # List of all yum repos, given repourls maybe be changed in resolve_composes
+        # with inheritance enabled, so we have to pass updated repos to koji_import
```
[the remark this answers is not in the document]
rcerven: ok

atomic_reactor/schemas/container.json
twaugh: Sure, but none of the other properties inside 'compose' are required. It could literally be:
```
compose:
  inherit_from_parent: true
```

It just looks like a namespacing issue to me.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -382,6 +382,8 @@ def get_build(self, metadata, worker_metadatas):
                 'signing_intent': resolve_comp_result['signing_intent'],
                 'signing_intent_overridden': resolve_comp_result['signing_intent_overridden'],
             }
+        if self.workflow.all_yum_repourls:
+            extra['image']['yum_repourls'] = self.workflow.all_yum_repourls
```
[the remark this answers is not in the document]
rcerven: ok

atomic_reactor/schemas/container.json
```diff
@@ -82,6 +82,9 @@
...
+      "type": "boolean"
```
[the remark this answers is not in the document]
rcerven: ack

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -134,6 +145,33 @@ def adjust_for_autorebuild(self):
         if self.compose_ids:
             self.log.info('Autorebuild detected: Ignoring compose_ids plugin parameter')
             self.compose_ids = tuple()
+            self.all_compose_ids = []
+
+    def adjust_for_inherit(self):
+        build_info = self.plugin_result[BASE_IMAGE_KOJI_BUILD]
```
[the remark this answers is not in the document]
rcerven: actually I'll check if plugin_result is None, because if it runs and finishes that key will be always defined, but might have None value, which will handle following tries which are getting values

atomic_reactor/schemas/container.json
rcerven: right it should be fine in compose anyway, moving it there

docs/koji.md
```diff
@@ -55,6 +55,7 @@ Data which is placed here includes:
 - `build.extra.image.odcs.compose_ids` (int list): List of each ODCS compose used
 - `build.extra.image.odcs.signing_intent` (str): Final signing intent of the ODCS composes after adjusting for CLI parameter
 - `build.extra.image.odcs.signing_intent_overridden` (boolean): Whether or not the signing intent used is different than the one defined in container.yaml
+- `build.extra.image.yum_repourls` (str list): information about provided yum repourls
```
twaugh: Is this the provided yum repos for the build, or the full list after inheritance?

atomic_reactor/schemas/container.json
```diff
@@ -78,6 +78,10 @@
         },
         "signing_intent": {
           "type": "string"
+        },
+        "inherit_from_parent": {
```
twaugh: Would just "inherit" be a better name? It can only inherit from the parent, after all.

docs/koji.md
rcerven: full list after inheritance, will change it

atomic_reactor/schemas/container.json
rcerven: sure that should do it

rcerven:
rn updated: inherit yum repos and composes from baseimage

PR #1175, opened 2019-02-22
allow atomic-reactor-repo directory in .dockerignore file,

rcerven:
when it exists

* OSBS-6920

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -48,7 +48,8 @@ from atomic_reactor.constants import (DOCKERFILE_FILENAME, REPO_CONTAINER_CONFIG
                                       PLUGIN_KOJI_PARENT_KEY,
                                       PARENT_IMAGE_BUILDS_KEY, PARENT_IMAGES_KOJI_BUILDS,
                                       BASE_IMAGE_KOJI_BUILD, BASE_IMAGE_BUILD_ID_KEY,
-                                      PARENT_IMAGES_KEY, SCRATCH_FROM)
+                                      PARENT_IMAGES_KEY, SCRATCH_FROM, RELATIVE_REPOS_PATH,
+                                      DOCKERIGNORE)
 from atomic_reactor.auth import HTTPRegistryAuth
 
 from dockerfile_parse import DockerfileParser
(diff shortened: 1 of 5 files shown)


atomic_reactor/util.py
```diff
@@ -1291,6 +1292,15 @@ def read_yaml(yaml_data, schema):
     return data
 
 
+def allow_repo_dir_in_dockerignore(build_path):
+    docker_ignore = os.path.join(str(build_path), DOCKERIGNORE)
+
+    if os.path.isfile(docker_ignore):
+        with open(docker_ignore, "a") as f:
+            f.write("!%s\n" % RELATIVE_REPOS_PATH)
```
twaugh: What if .dockerignore does not already end in a newline character? I think we may want to add one ourselves before the `!`.

tests/test_util.py
```diff
@@ -1510,6 +1511,29 @@ def test_introspection_signal_handler(capfd):
...
+            f.write(ignore_content)
+
+    allow_repo_dir_in_dockerignore(tmpdir)
+
+    if dockerignore_exists:
+        with open(docker_ignore_file, "r") as f:
+            ignore_lines = f.readlines()
+
+        assert len(ignore_lines) == 3
+        assert ignore_lines[2] == added_line
```
twaugh: This seems too prescriptive. If all we need is for it to be the last line in the file, and for the existing content to still be in place, we should test those two things:
```
with open(docker_ignore_file, "r") as f:
    # Last line is the one we added.
    assert f.readlines()[-1] == added_line

with open(docker_ignore_file, "r") as f:
    # Previous content is still in place.
    assert f.read(len(ignore_content)) == ignore_content
```

atomic_reactor/util.py
rcerven: sure

tests/test_util.py
```diff
@@ -1510,6 +1511,29 @@ def test_introspection_signal_handler(capfd):
     assert log_msg in err
 
 
+@pytest.mark.parametrize('dockerignore_exists', [True, False])
+def test_allow_repo_dir_in_dockerignore(tmpdir, dockerignore_exists):
+    docker_ignore_file = os.path.join(str(tmpdir), DOCKERIGNORE)
+
+    ignore_content = ["# Ignore everything so we can just have a whitelist of things to copy\n",
+                      "** / *\n"]
+    added_lines = ["\n", "!%s\n" % RELATIVE_REPOS_PATH]
```
twaugh: Maybe don't include "\n" in `added_lines`. We don't care whether we get an additional newline... we only care that "!..." appears as its own line.

rcerven: ok

rcerven:
rn updated : - allow atomic-reactor-repo directory in .dockerignore file, when it exists

PR #1176, opened 2019-02-25
Infer pulp usage from reactor config

athos-ribeiro:
Inferring pulp registries usage from push configurations in worker
builds may lead to undesired behaviour since this is currently only set
for x86_64 arches. We infer pulp registries are used by fetching it from
reactor configuration instead.

* OSBS-6849


athos-ribeiro:
@twaugh, [1] led me to believe I could use the values stored in pre_reactor_config workspace in a worker build as well.

Builds in a staging environment also behaved as expected. Would that be because orchestrator and worker builds shared the same cluster?

Would forwarding that pulp registry value as a parameter to the plugin and setting it as a fallback (as it is done with the other plugins) do it?

[1] https://github.com/projectatomic/atomic-reactor/commit/3b41ef8e32c7c11fd833fee791238bcfb0168b15#diff-5e7041655266b243b2a76e9c928c99b5

athos-ribeiro:
N o RN added. This fixes a bug introduced in this release cycle.

PR #1177, opened 2019-02-26
Extract operators metadata from built image

athos-ribeiro:
Add new export_operator_manifests plugin to extract operators metadata and upload it to Koji.

* OSBS-6686

Depends on https://github.com/projectatomic/osbs-client/pull/849

### DIFF
--- a/atomic_reactor/plugins/post_export_operator_manifests.py
+++ b/atomic_reactor/plugins/post_export_operator_manifests.py
@@ -1,0 +1,141 @@
+"""
+Copyright (c) 2019 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+import os
+import shutil
+import tarfile
+import tempfile
+import zipfile
+
+from atomic_reactor.constants import (PLUGIN_EXPORT_OPERATOR_MANIFESTS_KEY,
+                                      OPERATOR_MANIFESTS_ARCHIVE)
+from atomic_reactor.plugin import PostBuildPlugin
+from atomic_reactor.util import df_parser, is_scratch_build, get_platforms
+from docker.errors import APIError
+from osbs.utils import Labels
+from platform import machine
+
+MANIFESTS_DIR_NAME = 'manifests'
+
+
+class ExportOperatorManifestsPlugin(PostBuildPlugin):
+    """
+    Export operator manifest files
+
+    Fetch and archive operator manifest files from image so they can be
+    uploaded to koji.
+    """
+
+    key = PLUGIN_EXPORT_OPERATOR_MANIFESTS_KEY
+    is_allowed_to_fail = False
+
+    def __init__(self, tasker, workflow, operator_manifests_extract_platform=None, platform=None):
+        super(ExportOperatorManifestsPlugin, self).__init__(tasker, workflow)
+
+        self.operator_manifests_extract_platform = operator_manifests_extract_platform
+        if platform:
+            self.platform = platform
+        else:
+            self.platform = machine()
+
+    def has_operator_manifest(self):
+        """
+        Check if Dockerfile sets the operator manifest label
+
+        :return: bool
+        """
+        dockerfile = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        labels = Labels(dockerfile.labels)
+        try:
+            _, operator_label = labels.get_name_and_value(Labels.LABEL_TYPE_OPERATOR_MANIFESTS)
+        except KeyError:
+            operator_label = 'false'
+        return operator_label.lower() == 'true'
+
+    def is_orchestrator(self):
+        """
+        Check if the plugin is running in orchestrator.
+
+        :return: bool
+        """
+        if get_platforms(self.workflow):
+            return True
+        return False
+
+    def should_run(self):
+        """
+        Check if the plugin should run or skip execution.
+
+        :return: bool, False if plugin should skip execution
+        """
+        if self.is_orchestrator():
+            self.log.warning("%s plugin set to run on orchestrator. Skipping", self.key)
+            return False
+        if self.operator_manifests_extract_platform != self.platform:
+            self.log.info("Only platform [%s] will upload operators metadata. Skipping",
+                          self.operator_manifests_extract_platform)
+            return False
+        if is_scratch_build():
+            self.log.info("Scratch build. Skipping")
+            return False
+        if not self.has_operator_manifest():
+            self.log.info("Operator manifests label not set in Dockerfile. Skipping")
+            return False
+        return True
+
+    def run(self):
+        """
+        Run the plugin.
+
+        This plugin extracts the operator manifest files from an image, saves
+        them as a zip archive, and returns its path
+
+        :return: str, path to operator manifests zip file
+        """
+        if not self.should_run():
+            return
+
+        manifests_archive_dir = tempfile.mkdtemp()
+        image = self.workflow.image
+        # As in flatpak_create_oci, we specify command to prevent possible docker daemon errors.
+        container_dict = self.tasker.d.create_container(image, command=['/bin/bash'])
+        container_id = container_dict['Id']
+        try:
+            bits, stat = self.tasker.d.get_archive(container_id,
+                                                   os.path.join('/', MANIFESTS_DIR_NAME))
+        except APIError as ex:
+            self.log.error('Could not extract operator manifest files: %s', ex)
+            raise RuntimeError('Could not extract operator manifest files: %s' % ex)
+        finally:
+            self.tasker.d.remove_container(container_id)
+
+        with tempfile.NamedTemporaryFile() as extracted_file:
+            for chunk in bits:
+                extracted_file.write(chunk)
+            extracted_file.flush()
+            tar_archive = tarfile.TarFile(extracted_file.name)
+
+        tar_archive.extractall(manifests_archive_dir)
+        manifests_path = os.path.join(manifests_archive_dir, MANIFESTS_DIR_NAME)
+
+        manifests_zipfile_path = os.path.join(manifests_archive_dir, OPERATOR_MANIFESTS_ARCHIVE)
+        with zipfile.ZipFile(manifests_zipfile_path, 'w') as archive:
+            for root, _, files in os.walk(manifests_path):
+                for f in files:
+                    filedir = os.path.relpath(root, manifests_path)
+                    filepath = os.path.join(filedir, f)
+                    archive.write(os.path.join(root, f), filepath)
+            manifest_files = archive.namelist()
+            if not manifest_files:
+                self.log.error('Empty operator manifests directory')
+                raise RuntimeError('Empty operator manifests directory')
+            self.log.debug("Archiving operator manifests: %s", manifest_files)
+
+        shutil.rmtree(manifests_path)
+
+        return manifests_zipfile_path
--- a/atomic_reactor/plugins/post_koji_upload.py
+++ b/atomic_reactor/plugins/post_koji_upload.py
@@ -20,7 +20,9 @@ from atomic_reactor.plugins.post_rpmqa import PostBuildRPMqaPlugin
 from atomic_reactor.plugins.pre_reactor_config import (get_openshift_session,
                                                        get_prefer_schema1_digest,
                                                        get_koji_session, get_pulp)
-from atomic_reactor.constants import PROG, PLUGIN_KOJI_UPLOAD_PLUGIN_KEY
+from atomic_reactor.constants import (PROG, PLUGIN_KOJI_UPLOAD_PLUGIN_KEY,
+                                      PLUGIN_EXPORT_OPERATOR_MANIFESTS_KEY,
+                                      OPERATOR_MANIFESTS_ARCHIVE)
 from atomic_reactor.util import (get_version_of_tools, get_checksums,
                                  get_build_json, get_docker_architecture,
                                  get_image_upload_filename,
@@ -447,6 +449,20 @@ class KojiUploadPlugin(PostBuildPlugin):
         # Add the 'docker save' image to the output
         image = add_buildroot_id(output)
         output_files.append(image)
+
+        # add operator manifests to output
+        operator_manifests_path = (self.workflow.postbuild_results
+                                   .get(PLUGIN_EXPORT_OPERATOR_MANIFESTS_KEY))
+        if operator_manifests_path:
+            operator_manifests_file = open(operator_manifests_path)
+            manifests_metadata = self.get_output_metadata(operator_manifests_path,
+                                                          OPERATOR_MANIFESTS_ARCHIVE)
+            operator_manifests_output = Output(file=operator_manifests_file,
+                                               metadata=manifests_metadata)
+            # We use log type here until a more appropriate type name is supported by koji
+            operator_manifests_output.metadata.update({'arch': arch, 'type': 'log'})
+            operator_manifests = add_buildroot_id(operator_manifests_output)
+            output_files.append(operator_manifests)
 
         return output_files
 
--- a/tests/plugins/test_export_operator_manifests.py
+++ b/tests/plugins/test_export_operator_manifests.py
@@ -1,0 +1,159 @@
+"""
+Copyright (c) 2019 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+
+import os
+import pytest
+import tarfile
+import zipfile
+from atomic_reactor import util
+from atomic_reactor.constants import (
+    PLUGIN_EXPORT_OPERATOR_MANIFESTS_KEY,
+    PLUGIN_CHECK_AND_SET_PLATFORMS_KEY)
+from atomic_reactor.inner import DockerBuildWorkflow
+from atomic_reactor.plugins.post_export_operator_manifests import ExportOperatorManifestsPlugin
+from atomic_reactor.plugin import PostBuildPluginsRunner, PluginFailedException
+from docker.errors import NotFound
+from flexmock import flexmock
+from functools import partial
+from platform import machine
+from tests.constants import TEST_IMAGE
+from tests.stubs import StubInsideBuilder, StubSource
+from requests import Response
+
+
+CONTAINER_ID = 'mocked'
+
+
+def mock_dockerfile(tmpdir, has_label=True, label=True):
+    base = 'From fedora:30'
+    cmd = 'CMD /bin/cowsay moo'
+    operator_label = ''
+    if has_label:
+        operator_label = 'LABEL com.redhat.delivery.appregistry={}'.format(str(label).lower())
+    data = '\n'.join([base, operator_label, cmd])
+    tmpdir.join('Dockerfile').write(data)
+
+
+def mock_workflow(tmpdir):
+    workflow = DockerBuildWorkflow({"provider": "git", "uri": "asd"}, TEST_IMAGE)
+    workflow.source = StubSource()
+    builder = StubInsideBuilder().for_workflow(workflow)
+    builder.set_df_path(str(tmpdir))
+    builder.tasker = flexmock()
+    workflow.builder = flexmock(builder)
+    return workflow
+
+
+def generate_archive(tmpdir, empty=False):
+    archive_path = os.path.join(str(tmpdir), 'temp.tar')
+    archive_tar = tarfile.open(archive_path, 'w')
+    manifests_dir = os.path.join(str(tmpdir), 'manifests')
+    os.mkdir(manifests_dir)
+    another_dir = os.path.join(manifests_dir, 'another_dir')
+    os.mkdir(another_dir)
+    if not empty:
+        open(os.path.join(manifests_dir, 'stub.yml'), 'w').close()
+        open(os.path.join(another_dir, 'yayml.yml'), 'w').close()
+    archive_tar.add(manifests_dir, arcname='manifests')
+    archive_tar.close()
+    f = open(archive_path, 'rb')
+    for block in iter(partial(f.read, 8), b''):
+        yield block
+    f.close()
+    os.unlink(archive_path)
+
+
+def mock_env(tmpdir, docker_tasker, has_label=True, label=True, has_archive=True,
+             scratch=False, orchestrator=False, selected_platform=True, empty_archive=False):
+    build_json = {'metadata': {'labels': {'scratch': scratch}}}
+    flexmock(util).should_receive('get_build_json').and_return(build_json)
+    mock_dockerfile(tmpdir, has_label, label)
+    workflow = mock_workflow(tmpdir)
+    workflow.prebuild_results[PLUGIN_CHECK_AND_SET_PLATFORMS_KEY] = orchestrator
+    mock_stream = generate_archive(tmpdir, empty_archive)
+    plugin_conf = [{'name': ExportOperatorManifestsPlugin.key}]
+    if selected_platform:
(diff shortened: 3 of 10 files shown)


atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,179 @@
...
+            return False
+        return True
+
+    def run(self):
+        """
+        Run the plugin.
+        """
+        if not self.should_run():
+            return
+        saved_image = self.workflow.exported_image_sequence[-1].get('path')
```
rcerven: you should also check if anything was actually exported, as of now imagebuilder plugin exports it always, and prepubsquash has save_archive=true, but it can be also fals

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,179 @@
...
+        super(ExportOperatorManifestsPlugin, self).__init__(tasker, workflow)
+
+    def has_operator_manifest(self):
+        """
+        Check if the plugin is running in orchestrator.
+
+        :return: bool
+        """
+        dockerfile = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        # TODO: set the label in util Labels
```
rcerven: I'd say we want to add it to Labels in osbs-client's utils, as we are trying to keep all label constants in one place

atomic_reactor/plugins/post_export_operator_manifests.py
athos-ribeiro: should it be exported here in case it has not been exported yet?

rcerven: I'd say yes to make it resilient, I think in our code we always check if image was exported, if not we get it

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,179 @@
...
+
+        for image in manifest:
+            # We rely on the order of the array in manifests.json
+            for layer_path in image['Layers']:
+                try:
+                    layer_archive_info = image_archive.getmember(layer_path)
+                except KeyError as ex:
+                    raise RuntimeError('Image archive format error. '
+                                       'Layer archive not found for %s: %s' % (layer_path, ex))
+                if not layer_archive_info.isfile():
```
rcerven: this is actually directory

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,179 @@
...
+                    raise RuntimeError('No layer available for %s' % layer_path)
+                # layers can get large, let's extract them in the filesystem instead of load them
+                if layer_path.startswith('/') or '..' in layer_path:
+                    raise RuntimeError('Unsafe path found in image archive')
+                self.log.debug("extracting %s", layer_archive_info.name)
```
rcerven: inside of that directory with name of layer sha, is layer.tar

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,179 @@
...
+    def get_archived_manifests(self, members):
+        for tarinfo in members:
+            if tarinfo.name.startswith(os.path.join(MANIFESTS_DIR_NAME, '')):
+                yield tarinfo
+
+    def extract_manifests(self, image_archive_path, workdir):
+        """
+        Extract operator manifest files from image layers
+
+        :param image_archive_path: str, path to docker image archive tar file
```
twaugh: Why use this instead of the equivalent of `docker export`? The current code ties us to the `docker save` format.

atomic_reactor/plugins/post_export_operator_manifests.py
athos-ribeiro: The strings in the Layers list point to the tar files inside each directory.

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -448,6 +449,19 @@ def add_log_type(output, arch):
...
+        # add operator manifests to output
+        operator_manifests_path = (self.workflow.postbuild_results
+                                   .get(PLUGIN_EXPORT_OPERATOR_MANIFESTS_KEY))
+        if operator_manifests_path:
+            operator_manifests_file = open(operator_manifests_path)
+            manifests_metadata = self.get_output_metadata(operator_manifests_path,
+                                                          'operator_manifests.zip')
+            operator_manifests_output = Output(file=operator_manifests_file,
+                                               metadata=manifests_metadata)
+            operator_manifests_output.metadata.update({'arch': arch, 'type': 'log'})
```
twaugh: Could put a comment here explaining that 'log' is used as a stop-gap until a more appropriate type name is supported by Koji.

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -448,6 +449,19 @@ def add_log_type(output, arch):
         image = add_buildroot_id(output)
         output_files.append(image)
 
+        # add operator manifests to output
+        operator_manifests_path = (self.workflow.postbuild_results
+                                   .get(PLUGIN_EXPORT_OPERATOR_MANIFESTS_KEY))
+        if operator_manifests_path:
+            operator_manifests_file = open(operator_manifests_path)
+            manifests_metadata = self.get_output_metadata(operator_manifests_path,
+                                                          'operator_manifests.zip')
```
twaugh: Should this import a constant instead of using a string literal?

atomic_reactor/plugins/post_export_operator_manifests.py
rcerven: ahh sorry I am blind, you are right

atomic_reactor/plugins/post_export_operator_manifests.py
rcerven: nvm, ic now that strings lin Layers list point to layer.tar file

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,179 @@
...
+                        self.log.debug("%s found. Marking %s for removal",
+                                       entry, entry[len(WHITEOUT_FILE_PREFIX):])
+                        remove_list.append(entry)
+                        remove_list.append(entry[len(WHITEOUT_FILE_PREFIX):])
+                for entry in remove_list:
+                    os.unlink(entry)
+
+        manifests_path = os.path.join(workdir, MANIFESTS_DIR_NAME)
+        if not os.path.isdir(manifests_path):
+            self.log.warning('No operator manifest files found. No archive will be created.')
```
MartinBasti: IMO this should be error, if operator image is declared by label, it should contain `/manifests`.

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,178 @@
...
+        """
+        Check if the plugin is running in orchestrator.
+
+        :return: bool
+        """
+        dockerfile = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        # TODO: set the label in Labels
+        # labels = Labels(dockerfile.labels)
+        # _, image_name = labels.get_name_and_value(Labels.LABEL_TYPE_NAME)
+        return dockerfile.labels.get(OPERATORS_LABEL, False)
```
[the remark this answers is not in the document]
MartinBasti: In specification we have that `"com.redhat.delivery.appregistry=true"` means it is operator image. This info was already passed to container owners. So probably it should be something like:
```python
return dockerfile.labels.get(OPERATORS_LABEL, "").lower() == "true"
```

atomic_reactor/plugins/post_export_operator_manifests.py
ralphbean: Agreed.  This should fail the build if the label is set but `manifests_path` is non-existent *or empty*.

atomic_reactor/plugins/post_export_operator_manifests.py
athos-ribeiro: As Tim suggested, we now avoid the `docker save` format, so we are not using the exported image anymore.

atomic_reactor/plugins/post_export_operator_manifests.py
athos-ribeiro: would it be reasonable to use `get_archive` [1]?

[1]  https://docker-py.readthedocs.io/en/stable/api.html?highlight=get_image#docker.api.container.ContainerApiMixin.get_archive

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,123 @@
...
+        :return: str, path to operator manifests zip file
+        """
+        image = self.workflow.image
+        container_dict = self.tasker.d.create_container(image, command=['/bin/bash'])
+        container_id = container_dict['Id']
+        try:
+            bits, stat = self.tasker.d.get_archive(container_id,
+                                                   os.path.join('/', MANIFESTS_DIR_NAME))
+        except APIError as ex:
+            self.log.error('Could not extract operator manifest files: %s' % ex)
```
twaugh: Maybe `self.log.error('...: %s', ex)` or even `self.log.exception('...')`

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,123 @@
...
+            raise RuntimeError('Could not extract operator manifest files: %s' % ex)
+        finally:
+            self.tasker.d.remove_container(container_id)
+
+        fd, extracted_file = tempfile.mkstemp()
+        os.close(fd)
+        with open(extracted_file, 'w') as f:
```
twaugh: No need for `'wb'`?

atomic_reactor/plugins/post_export_operator_manifests.py
twaugh: Perfect.

tests/plugins/test_export_operator_manifests.py
```diff
@@ -0,0 +1,108 @@
...
+    workflow.source = StubSource()
+    builder = StubInsideBuilder().for_workflow(workflow)
+    builder.set_df_path(str(tmpdir))
+    builder.tasker = flexmock()
+    workflow.builder = flexmock(builder)
+    workflow.exported_image_sequence.append({'path': image_path})
+    return workflow
+
+
+def generate_image(tmpdir, image_path):
```
twaugh: Is this used now?

tests/plugins/test_export_operator_manifests.py
```diff
@@ -0,0 +1,108 @@
...
+    @pytest.mark.parametrize('has_label', [True, False])
+    @pytest.mark.parametrize('label', [True, False])
+    @pytest.mark.parametrize('orchestrator', [True, False])
+    @pytest.mark.parametrize('selected_platform', [True, False])
+    def test_skip(self, docker_tasker, tmpdir, caplog, scratch, has_label,
+                  label, orchestrator, selected_platform):
+        runner = mock_env(tmpdir, docker_tasker, has_label=has_label, label=label,
+                          scratch=scratch, orchestrator=orchestrator)
+        result = runner.run()
+        if any([not has_label, not label, scratch, orchestrator, not selected_platform]):
```
twaugh: Don't need `[`/`]` here, `any(not has_label, not label, ...)` (it's implicitly a generator expression)

athos-ribeiro: It raises a TypeError in this case
```
any(iterable) -> bool
```

twaugh: Oh, I'm thinking of `any([... for ... in ...])` -> `any(... for ... in ...)`. Sorry for the noise.

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,123 @@
...
+    def extract_manifests(self, workdir):
+        """
+        Extract operator manifest files from image
+
+        :param workdir: str, path to work directory
+
+        :return: str, path to operator manifests zip file
+        """
+        image = self.workflow.image
+        container_dict = self.tasker.d.create_container(image, command=['/bin/bash'])
```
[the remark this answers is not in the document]
athos-ribeiro: Added a comment on that

tests/plugins/test_export_operator_manifests.py
```diff
@@ -0,0 +1,114 @@
...
+            assert 'Skipping' in caplog.text
+            assert result[PLUGIN_EXPORT_OPERATOR_MANIFESTS_KEY] is None
+
+    def test_export_archive(self, docker_tasker, tmpdir, caplog):
+        runner = mock_env(tmpdir, docker_tasker)
+        result = runner.run()
+        archive = result[PLUGIN_EXPORT_OPERATOR_MANIFESTS_KEY]
+        assert archive
+        assert archive.split('/')[-1] == 'operator_manifests.zip'
```
twaugh: Could we have a test-case that shows we get an error if there are no manifests to include in the archive? i.e. to test this: https://github.com/projectatomic/atomic-reactor/pull/1177#discussion_r260883597

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,138 @@
...
+
+        :return: bool
+        """
+        dockerfile = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        labels = Labels(dockerfile.labels)
+        try:
+            _, operator_label = labels.get_name_and_value(Labels.LABEL_TYPE_OPERATOR_MANIFESTS)
+        except KeyError:
+            operator_label = 'false'
+        return operator_label == "true"
```
twaugh: Why do we use two different quoting styles here?

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,138 @@
...
+            return False
+        return True
+
+    def run(self):
+        """
+        Run the plugin.
+        """
+        if not self.should_run():
+            return
+        temp_dir = tempfile.mkdtemp()
```
twaugh: Do we care about removing this afterwards?

tests/plugins/test_export_operator_manifests.py
athos-ribeiro: Done. Although, we had to mock the docker response.

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,138 @@
...
+    is_allowed_to_fail = False
+
+    def __init__(self, tasker, workflow, operator_uploader_platform=None, platform=None):
+        super(ExportOperatorManifestsPlugin, self).__init__(tasker, workflow)
+
+        self.operator_uploader_platform = operator_uploader_platform
+        if platform:
+            self.platform = platform
+        else:
+            self.platform = os.uname()[4]
```
MartinBasti: Isn't this equivalent to https://docs.python.org/2/library/platform.html#platform.machine

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,138 @@
...
+
+        :return: bool
+        """
+        dockerfile = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        labels = Labels(dockerfile.labels)
+        try:
+            _, operator_label = labels.get_name_and_value(Labels.LABEL_TYPE_OPERATOR_MANIFESTS)
+        except KeyError:
+            operator_label = 'false'
+        return operator_label == 'true'
```
MartinBasti: How about `True`, `TRUE`? Do we want be case sensitive here?

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def is_orchestrator(self):
+        """
+        Check if the plugin is running in orchestrator.
+
+        :return: bool
+        """
+        if get_platforms(self.workflow):
```
MartinBasti: [Off topic] I'd like to see to have some generic way how to get info if code runs as worker or orchestrator, this is not the first time we need something like this.

tests/plugins/test_export_operator_manifests.py
```diff
@@ -0,0 +1,132 @@
...
+    workflow.builder = flexmock(builder)
+    return workflow
+
+
+def generate_archive(tmpdir):
+    archive_path = os.path.join(str(tmpdir), 'temp.tar')
+    archive_tar = tarfile.open(archive_path, 'w')
+    manifests_dir = os.path.join(str(tmpdir), 'manifests')
+    os.mkdir(manifests_dir)
+    open(os.path.join(manifests_dir, 'stub.json'), 'w').close()
```
MartinBasti: `/manifests/` dir should contain `yml` files

tests/plugins/test_export_operator_manifests.py
twaugh: How about if the directory exists but is empty? Will that behave the some way?

atomic_reactor/plugins/post_export_operator_manifests.py
athos-ribeiro: It is. I did not know about platform.machine :)

Note that there are other parts of the code using os.uname to fetch the arch name. Maybe we should also change those for consistency if we are going with platform.machine here.

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        Run the plugin.
+        """
+        if not self.should_run():
+            return
+        temp_dir = tempfile.mkdtemp()
+        manifests_zipfile_path = self.extract_manifests(temp_dir)
+        return manifests_zipfile_path
+
+    def extract_manifests(self, workdir):
```
mlangsdorf: is there any reason this is a separate function instead of part of run?  It seems unnecessary to call a function here instead of just performing the same sequence of instructions.

atomic_reactor/plugins/post_export_operator_manifests.py
athos-ribeiro: Should we? I did not chose the best name for that directory (changed).

The final operator manifests zip file will be saved in this directory. If we want to remove it, then I understand we want to remove the manifest zip file after it is uploaded, is that correct?

atomic_reactor/plugins/post_export_operator_manifests.py
athos-ribeiro: I wanted to avoid having a bloated run method, but in the way it was written, it woudn't make much sense to extract the method, as you pointed. I merged it into the `run` method as suggested.

tests/plugins/test_export_operator_manifests.py
athos-ribeiro: Added code to handle empty /manifests dir + tests

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,139 @@
...
+        tar_archive.extractall(manifests_archive_dir)
+        manifests_path = os.path.join(manifests_archive_dir, MANIFESTS_DIR_NAME)
+        base_zipfile_name = os.path.join(manifests_archive_dir, OPERATORS_ARCHIVE_NAME)
+        manifest_files = os.listdir(manifests_path)
+        if not manifest_files:
+            self.log.error('Empty operator manifests directory')
+            raise RuntimeError('Empty operator manifests directory')
+
+        self.log.debug("Archiving operator manifests: %s", manifest_files)
+        manifests_zipfile_path = shutil.make_archive(base_zipfile_name,
```
MartinBasti: There is unexpected dir named `.` in the archive
```
$ unzip -l operator_manifests.zip 
Archive:  operator_manifests.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  03-05-2019 08:10   ./
        4  03-05-2019 08:10   empty.yaml
---------                     -------
        4                     2 files
```

athos-ribeiro: This is was a [known bug in shutil](https://bugs.python.org/issue28488) which [has been fixed](https://github.com/python/cpython/commit/666de7772708047b63125126b0147931571254a4) in python >= 3.5 and in newer python 2.7 releases.

It is present in python <= 2.7.11 though, which is the version we got in CentOS 7. I will include some code to deal with the issue.

atomic_reactor/plugins/post_export_operator_manifests.py
MartinBasti: I'd avoid changing it everywhere for now

atomic_reactor/plugins/post_export_operator_manifests.py
athos-ribeiro: Changed implementation to use `zipfile` instead of `shutil.make_archive`

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,140 @@
...
+        fd, extracted_file = tempfile.mkstemp()
+        os.close(fd)
+        with open(extracted_file, 'wb') as f:
+            for chunk in bits:
+                f.write(chunk)
+
+        tar_archive = tarfile.TarFile(extracted_file)
+        tar_archive.extractall(manifests_archive_dir)
+        manifests_path = os.path.join(manifests_archive_dir, MANIFESTS_DIR_NAME)
+        manifest_files = os.listdir(manifests_path)
```
MartinBasti: This supports only 1 level but there is some ongoing work that manifests may have nested directories. I'd like to just archive the whole dir.

athos-ribeiro: Fixed. Also added a test case to cover it.

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -0,0 +1,141 @@
...
+                                                   os.path.join('/', MANIFESTS_DIR_NAME))
+        except APIError as ex:
+            self.log.error('Could not extract operator manifest files: %s', ex)
+            raise RuntimeError('Could not extract operator manifest files: %s' % ex)
+        finally:
+            self.tasker.d.remove_container(container_id)
+
+        with tempfile.NamedTemporaryFile() as extracted_file:
+            for chunk in bits:
+                extracted_file.flush()
```
MartinBasti: this is not nice

atomic_reactor/plugins/post_export_operator_manifests.py
twaugh: Oh, I see. I guess that's fine.

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -448,6 +450,20 @@ def add_log_type(output, arch):
...
+        if operator_manifests_path:
+            operator_manifests_file = open(operator_manifests_path)
+            manifests_metadata = self.get_output_metadata(operator_manifests_path,
+                                                          OPERATOR_MANIFESTS_ARCHIVE)
+            operator_manifests_output = Output(file=operator_manifests_file,
+                                               metadata=manifests_metadata)
+            # We use log type here until a more appropriate type name is supported by koji
+            operator_manifests_output.metadata.update({'arch': arch, 'type': 'log'})
+            operator_manifests = add_buildroot_id(operator_manifests_output)
+            output_files.append(operator_manifests)
```
twaugh: I just spotted that we don't have test coverage for this path.

athos-ribeiro:
Release Notes:

- add new extract_operator_manifests plugin. It is now possible to set a `com.redhat.delivery.appregistry` label to "true" in Dockerfile to allow files under `/manifests` in the built image to be uploaded to koji in a `operator_manifests.zip` archive

PR #1180, opened 2019-03-04
remove inherit from compse dict, so it can be identified as empty

rcerven:

### DIFF
--- a/atomic_reactor/source.py
+++ b/atomic_reactor/source.py
@@ -57,6 +57,10 @@ class SourceConfig(object):
         self.go = self.data.get('go') or {}
         self.image_build_method = meth = self.data.get('image_build_method')
         self.inherit = self.compose.get('inherit', False) if self.compose else False
+        if self.compose:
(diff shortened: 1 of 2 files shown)


MartinBasti:
I miss description why we need this

atomic_reactor/source.py
```diff
@@ -57,6 +57,8 @@ def __init__(self, build_path):
         self.go = self.data.get('go') or {}
         self.image_build_method = meth = self.data.get('image_build_method')
         self.inherit = self.compose.get('inherit', False) if self.compose else False
+        if self.compose:
+            self.compose.pop('inherit', None)
```
twaugh: I'd like a comment here explaining why this is useful, i.e. it means that self.compose can be evaluated as a bool in order to decide whether any ODCS composes will be created.

rcerven: ok will add

rcerven:
no updating RN since it is fix for unreleased feature

PR #1181, opened 2019-03-05
  inherit parent composes logic fix

rcerven:


rcerven:
additional unit tests added

rcerven:
ok updated tests, made compose in container.yaml independent of inherit option

rcerven:
not updating rn, as this is fix for unreleased feature

PR #1183, opened 2019-03-06
resolve_composes: add support for modular_koji_tags

mlangsdorf:
fixes osbs6833

Add support to pre_resolve_composes to parse modular_koji_tags and pass that on to the ODCS engine.

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -368,6 +368,7 @@ class ComposeConfig(object):
         self.arches = arches or []
         self.multilib_arches = []
         self.multilib_method = None
+        self.modular_tags = data.get('modular_koji_tags', [])
 
         if data.get('pulp_repos'):
             for arch in pulp_data or {}:
@@ -431,6 +432,8 @@ class ComposeConfig(object):
         }
         if self.arches:
             request['arches'] = self.arches
+        if self.modular_tags:
+            request['modular_koji_tags'] = self.modular_tags
         return request
 
     def render_pulp_request(self, arch):
--- a/atomic_reactor/schemas/container.json
+++ b/atomic_reactor/schemas/container.json
@@ -83,6 +83,24 @@
           "description": "enable inheritance of yum repourls and composes from baseimage build (disabled by default)",
           "type": "boolean"
         },
+        "include_unpublished_pulp_repos": {
+          "description": "include unpublished repos in pulp input content-sets",
+          "type": "boolean"
+        },
+        "multilib_method": {
+          "description": "list of options to decide if a package is multilib",
+          "type": ["array", "null"],
+          "items": {
+              "enum": ["none", "iso", "runtime", "devel", "all"]
+          }
+        },
+        "multilib_arches": {
+          "description": "list of arches which will get multilib composes",
+          "type": ["array", "null"],
+          "items": {
+              "type": "string"
+          }
+        },
         "modular_koji_tags": {
           "description": "tags of modular Koji content to include in the build",
           "type": ["array", "null"],
@@ -91,7 +109,7 @@
           }
         }
       },
-      "additionalProperties": true
+      "additionalProperties": false
     },
     "flatpak": {
       "type": ["object", "null"]
--- a/atomic_reactor/schemas/container.json
+++ b/atomic_reactor/schemas/container.json
@@ -82,6 +82,13 @@
         "inherit": {
           "description": "enable inheritance of yum repourls and composes from baseimage build (disabled by default)",
           "type": "boolean"
+        },
+        "modular_koji_tags": {
+          "description": "tags of modular Koji content to include in the build",
+          "type": ["array", "null"],
+          "items": {
+              "type": "string"
+          }
         }
       },
       "additionalProperties": true
--- a/tests/test_odcs_util.py
+++ b/tests/test_odcs_util.py
@@ -82,13 +82,16 @@ def compose_json(state, state_name, source_type='module', source=MODULE_NSV,
     ['x86_64'],
     ['breakfast', 'lunch'],
 ))
-@pytest.mark.parametrize(('source', 'source_type', 'packages', 'expected_packages', 'sigkeys'), (
-    (MODULE_NSV, 'module', None, None, None),
-    ('my-tag', 'tag', None, [], None),
-    ('my-tag', 'tag', ['spam', 'bacon', 'eggs'], ['spam', 'bacon', 'eggs'], None),
-    ('my-tag', 'tag', ['spam', 'bacon', 'eggs'], ['spam', 'bacon', 'eggs'], ['B456', 'R123']),
-    ('my-tag', 'tag', ['spam', 'bacon', 'eggs'], ['spam', 'bacon', 'eggs'], ""),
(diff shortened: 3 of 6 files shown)


mlangsdorf:
just realized we don't have a schema for pre_resolve_compose's data files.  Not sure if that can be fixed, either.

atomic_reactor/schemas/container.json
```diff
@@ -82,9 +82,42 @@
...
+              "enum": ["none", "iso", "runtime", "devel", "all"]
+          }
+        },
+        "multilib_arches": {
+          "description": "tags of modular Koji content to include in the build",
+          "type": ["array", "null"],
+          "items": {
+              "type": "string"
+          }
+        },
```
twaugh: Given that we don't do anything with these properties, do we really want to allow them in container.yaml?

mlangsdorf: we support multlib_arches and multilib_methods.  See #1084.

Fixed the description, though.

twaugh: Ah, thanks. and I see include_unpublished_pulp_repos is also used.

What about no_deps and check_deps?

mlangsdorf: They're options for the flags argument same as include_unpublished_pulp_repos and use the same syntax.

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -431,6 +432,8 @@ def render_modules_request(self):
         }
         if self.arches:
             request['arches'] = self.arches
+        if self.modular_tags:
+            request['modular_koji_tags'] = ' '.join(self.modular_tags)
```
athos-ribeiro: In the previous commit, `start_compose` docstring says it is supposed to receive a list of strings (ODCS code also seems to be expecting a list). Is there a reason for having a string here?

mlangsdorf: I got confused. Fixed.

atomic_reactor/schemas/container.json
rcerven: but I don't see those 2 options used anywhere

rcerven:
code looks good

atomic_reactor/schemas/container.json
mlangsdorf: pre_resolve_composes scoops up everything that isn't used somewhere else into flags and passed them to ODCS as a flag.

athos-ribeiro:
LGTM

About the test suite:

one of the parameters passed to `test_valid_source_config` at `tests/test_source.py:190` is not valid anymore. The extra_key entry could be moved to some on the invalid tests and new entries with valid examples could be place in `test_valid_source_config`

atomic_reactor/schemas/container.json
rcerven: this is only place where any flags are set

https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/pre_resolve_composes.py#L378

mlangsdorf:
Still can't figure out why flatpak tests are failing.  I can't reproduce locally.

tests/plugins/test_flatpak_create_oci.py
```diff
@@ -475,7 +475,7 @@ def build_export(repo, directory, branch):
       '--oci', '@filename', '@name', '@branch'],
      MockFlatpak.build_bundle),
     # For build-export, we assume the latest flatpak version for flatpak_module_tools compatibility
-    (['flatpak', 'build-export', '@repo', '@directory', '@branch', '--disable-sandbox'],
```
athos-ribeiro: I did reproduce the failure locally and reverting this change will fix the broken test.

See discussion #1160 as reference for why we need this flag.

About reproducing the error locally: It seems this tests are being skipped even using the same script as the CI does. I needed to install krb5-devel and python-koji in the container to get it to run. Maybe some package got updated in Fedora and didn't reach CI mirrors yet?

mlangsdorf: I added that change to get my local tests to work.  That is so crazy.  Well, I deleted that commit and things are green again.

tests/test_odcs_util.py
```diff
@@ -82,13 +82,16 @@ def compose_json(state, state_name, source_type='module', source=MODULE_NSV,
...
-    ('my-tag', 'tag', None, [], None),
-    ('my-tag', 'tag', ['spam', 'bacon', 'eggs'], ['spam', 'bacon', 'eggs'], None),
-    ('my-tag', 'tag', ['spam', 'bacon', 'eggs'], ['spam', 'bacon', 'eggs'], ['B456', 'R123']),
-    ('my-tag', 'tag', ['spam', 'bacon', 'eggs'], ['spam', 'bacon', 'eggs'], ""),
-    ('my-tag', 'tag', ['spam', 'bacon', 'eggs'], ['spam', 'bacon', 'eggs'], []),
+@pytest.mark.parametrize(('source', 'source_type', 'packages', 'expected_packages', 'sigkeys',
+                          'modular_koji_tags', 'expected_koji_tags'), (
+    (MODULE_NSV, 'module', None, None, None, [], None),
+    (MODULE_NSV, 'module', None, None, None, ['release'], ['release']),
+    ('my-tag', 'tag', None, [], None, 'release', None),
```
twaugh: Should that be `['release']`?

tests/test_odcs_util.py
```diff
@@ -82,13 +82,16 @@ def compose_json(state, state_name, source_type='module', source=MODULE_NSV,
...
+    ('my-tag', 'tag', ['spam', 'bacon', 'eggs'], ['spam', 'bacon', 'eggs'], None, None, None),
+    ('my-tag', 'tag', ['spam', 'bacon', 'eggs'], ['spam', 'bacon', 'eggs'], ['B456', 'R123'],
+     'release', None),
+    ('my-tag', 'tag', ['spam', 'bacon', 'eggs'], ['spam', 'bacon', 'eggs'], "", None, None),
+    ('my-tag', 'tag', ['spam', 'bacon', 'eggs'], ['spam', 'bacon', 'eggs'], [], 'release', None),
```
twaugh: Should that be `['release']`?

tests/test_odcs_util.py
mlangsdorf: No, modular koji tags are not a valid parameter for 'tag' releases and get ignored.

tests/test_odcs_util.py
mlangsdorf: No, modular koji tags are not a valid parameter for 'tag' releases and get ignored.

twaugh: Yes, I know we expect `None` for the final value there, but the value before that is a string not a list. I don't think that would be allowed by the schema, so I don't think it's testing something the logic would actually see.

mlangsdorf: Ah, okay.  Yes, you're right.  Fixed.

atomic_reactor/schemas/container.json
twaugh: What's the resolution to this?

mlangsdorf: They're valid flags, I'd like to keep them in the schema even if we don't support them in pre_resolve_composes.

mlangsdorf: Removed the flags we don't support.

mlangsdorf:
Added to release notes under improvements
"- Add support for 'modular_koji_tags' in ODCS compose builds"

PR #1185, opened 2019-03-21
verify media types plugin reworked, returns list of found types,

rcerven:
or manifest list type only if x86_64 is missing,
also reports missing types expected for each registry

* OSBS-6961

### DIFF
--- a/atomic_reactor/plugins/exit_verify_media_types.py
+++ b/atomic_reactor/plugins/exit_verify_media_types.py
@@ -13,7 +13,9 @@ from __future__ import unicode_literals
 from atomic_reactor.constants import (PLUGIN_GROUP_MANIFESTS_KEY, PLUGIN_VERIFY_MEDIA_KEY,
                                       MEDIA_TYPE_DOCKER_V1, MEDIA_TYPE_DOCKER_V2_SCHEMA1,
                                       MEDIA_TYPE_DOCKER_V2_SCHEMA2,
-                                      MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST)
+                                      MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST,
+                                      MEDIA_TYPE_OCI_V1,
+                                      MEDIA_TYPE_OCI_V1_INDEX)
 
 from atomic_reactor.plugin import ExitPlugin
 from atomic_reactor.util import get_manifest_digests, get_platforms, RegistrySession
@@ -35,10 +37,11 @@ def verify_v1_image(image, registry, log, insecure=False, dockercfg_path=None):
         log.debug("verify_v1_image: [%s] %s", r.status_code, r.url)
 
     log.debug("verify_v1_image: response headers: %s", response.headers)
-    response.raise_for_status()
-
-    # if we returned ok, then everything is fine.
-    return True
+    try:
+        response.raise_for_status()
+        return True
+    except Exception:
+        return False
 
 
 class VerifyMediaTypesPlugin(ExitPlugin):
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/exit_verify_media_types.py
```diff
@@ -35,10 +37,11 @@ def verify_v1_image(image, registry, log, insecure=False, dockercfg_path=None):
...
 
     log.debug("verify_v1_image: response headers: %s", response.headers)
-    response.raise_for_status()
-
-    # if we returned ok, then everything is fine.
-    return True
+    try:
+        response.raise_for_status()
+        return True
+    except Exception:
```
mlangsdorf: Do we need to test the kind of exception for anything here?  expect_media originally assumed that if you were verifying a v1 image, you had a v1 image to verify and any failure was a problem.  In this case, do we need to distinguish between 404 or 503 errors (or whatever) errors?  Or can we assume that if we can't verify the V1 image for whatever reason, we can just continue the build and raise the KeyError if it was expected?

rcerven: mmm we are already using registrysession which is doing retries on (408, 500, 502, 503, 504), so I think at this point if it raises exception we can consider it as it doesn't exist

also we don't have to raise here, but just return False, as now, we are first checking what is available and after that we are actually comparing it with what is expected

logic before was calling this method ONLY when it was already expecting v1, so it made sense in raising error in that case

mlangsdorf: Okay.  I wanted to double check the logic but that seems sensible.

atomic_reactor/plugins/exit_verify_media_types.py
```diff
@@ -59,77 +62,85 @@ def run(self):
...
+            if verify_v1_image(pullspec, registry_name, self.log, insecure, secret):
+                media_types.add(MEDIA_TYPE_DOCKER_V1)
+
+            media_in_registry[registry_name]['found'] = media_types
+
+        should_raise = False
+        all_found = set()
+        for registry_name, manifests in media_in_registry.items():
+            all_found.update(manifests['found'])
+            if manifests['expected'] - manifests['found']:
```
athos-ribeiro: Is it possible that 'found' contains some unexpected entry? If so, should we also capture the occurrence and log it?

atomic_reactor/plugins/exit_verify_media_types.py
```diff
@@ -59,77 +62,85 @@ def run(self):
...
+                               registry_name)
+
+        if should_raise:
+            raise KeyError("expected media types were not found")
+
+        if expect_list_only:
+            return [MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST]
+        return sorted(all_found)
+
+    def set_manifest_list_only_expectation(self):
```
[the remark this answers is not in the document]
rcerven: ok

atomic_reactor/plugins/exit_verify_media_types.py
rcerven: in 'found' we story only media_types which can be in 'expected_media_types'
if 'found' has more entries than 'expected' they will be returned, unless manifest list expectation only is met

we are treating 'expected_media_types' more like required types

atomic_reactor/plugins/exit_verify_media_types.py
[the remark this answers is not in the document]
athos-ribeiro: This is actually not setting anything, right? Could we actually change the method name? I could not think of anything better though? thoughts?

Then, the docstring could explain what the method does/returns

atomic_reactor/plugins/exit_verify_media_types.py
athos-ribeiro: Thanks for the answer. I wanted to check if logging those would be useful at all!

atomic_reactor/plugins/exit_verify_media_types.py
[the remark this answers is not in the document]
rcerven: sure, will do

rcerven:
rn updated: Verify media types plugin returns list of found types or
manifest list type

PR #1194, opened 2019-04-10
Fix py2/3 compatibility issues

chmeliik:
OSBS-6456

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -9,6 +9,8 @@ of the BSD license. See the LICENSE file for details.
 Script for building docker image. This is expected to run inside container.
 """
 
+from __future__ import absolute_import
+
 import json
 import logging
 import tempfile
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -9,7 +9,7 @@ of the BSD license. See the LICENSE file for details.
 Script for building docker image. This is expected to run inside container.
 """
 
-from __future__ import absolute_import
+from __future__ import absolute_import, division
 
 import json
 import logging
@@ -298,9 +298,9 @@ class FSWatcher(threading.Thread):
 
         mb = 1000 ** 2  # sadly storage is generally expressed in decimal units
         new_data = dict(
-            mb_free=st.f_bfree * st.f_frsize / mb,
-            mb_total=st.f_blocks * st.f_frsize / mb,
-            mb_used=(st.f_blocks - st.f_bfree) * st.f_frsize / mb,
+            mb_free=st.f_bfree * st.f_frsize // mb,
+            mb_total=st.f_blocks * st.f_frsize // mb,
+            mb_used=(st.f_blocks - st.f_bfree) * st.f_frsize // mb,
             inodes_free=st.f_ffree,
             inodes_total=st.f_files,
             inodes_used=st.f_files - st.f_ffree,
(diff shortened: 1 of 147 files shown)


atomic_reactor/inner.py
```diff
@@ -9,7 +9,7 @@
 Script for building docker image. This is expected to run inside container.
 """
 
-from __future__ import absolute_import
+from __future__ import absolute_import, division
```
MartinBasti: In this module I see we are working with storage sizes:

```
mb_free=st.f_bfree * st.f_frsize / mb,
```
Shouldn't be this rewrited to

```python
mb_free=st.f_bfree * st.f_frsize // mb,
```
To keep integers instead of floats?

MartinBasti: Please check required types on all places where division is used

twaugh: Yes, for mb_free etc we want integers (i.e. same as we had before).

atomic_reactor/plugins/post_push_to_pulp.py
```diff
@@ -188,7 +188,7 @@ def push_tar(self, filename, image_names=None, repo_prefix="redhat-"):
 
         # Only publish if we don't the pulp_sync plugin also configured
         if self.publish:
-            self.pulp_handler.publish(pulp_repos.keys())
+            self.pulp_handler.publish(list(pulp_repos.keys()))
```
twaugh: What is the warning this fixes?

twaugh: Oh, it's this: `dict.keys referenced when not iterating`

OK, that's fair.

chmeliik:
Division related issues should be fixed now, also removed some no longer required casts to `float`

chmeliik:
Didn't realize I had to run the python2 version of pylint as well, now the issues should hopefully be fixed for real.
Also, I changed the fixes for `dict-keys-not-iterating` from `list(some_dict.keys())` to just `list(some_dict)`, which should be more efficient on python2, though I'm going to need an opinion on whether it is explicit enough.

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
```diff
@@ -56,7 +56,7 @@ def match(self, build_archive):
         return False
 
     def match_all(self, build_archives):
-        return filter(self.match, build_archives)
+        return list(filter(self.match, build_archives))
```
twaugh: Another formulation of this would be `[archive for archive in build_archives if self.match(archive)]`, which is at least more explicit.

atomic_reactor/plugins/post_push_to_pulp.py
```diff
@@ -188,7 +188,7 @@ def push_tar(self, filename, image_names=None, repo_prefix="redhat-"):
 
         # Only publish if we don't the pulp_sync plugin also configured
         if self.publish:
-            self.pulp_handler.publish(pulp_repos.keys())
+            self.pulp_handler.publish(list(pulp_repos))
```
twaugh: Yes, I like this better than `list(pulp_repos.keys())`.

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
chmeliik: Ok, I like that

atomic_reactor/plugins/exit_remove_worker_metadata.py
```diff
@@ -38,4 +38,4 @@ def run(self):
                 osbs.delete_config_map(cm_key)
                 self.log.debug("ConfigMap %s deleted", cm_key)
             except OsbsResponseException as ex:
-                self.log.warning("Failed to delete ConfigMap %s: %s", cm_key, ex.message)
+                self.log.warning("Failed to delete ConfigMap %s: %s", cm_key, str(ex))
```
MartinBasti: `str()` is not  needed, `%s` in format string will call `str()`

chmeliik: Ok, fixed

MartinBasti:
Release notes:
> # Bug Fixes
> - python 2/3 compatibility fixes

PR #1195, opened 2019-04-11
fail build when vr tag for image already exists

rcerven:
* OSBS-7167

### DIFF
--- a/atomic_reactor/plugins/post_group_manifests.py
+++ b/atomic_reactor/plugins/post_group_manifests.py
@@ -21,7 +21,8 @@ from atomic_reactor.plugins.pre_reactor_config import (get_group_manifests,
                                                        get_platform_descriptors,
                                                        get_registries)
 from atomic_reactor.util import (RegistrySession, registry_hostname, ManifestDigest,
-                                 get_manifest_media_type)
+                                 get_manifest_media_type, get_primary_images,
+                                 get_manifest_list)
 from atomic_reactor.constants import (PLUGIN_GROUP_MANIFESTS_KEY, MEDIA_TYPE_DOCKER_V2_SCHEMA2,
                                       MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST, MEDIA_TYPE_OCI_V1,
                                       MEDIA_TYPE_OCI_V1_INDEX)
@@ -208,6 +209,40 @@ class GroupManifestsPlugin(PostBuildPlugin):
                 ],
         }, indent=4)
 
+    def check_existing_vr_tag(self):
+        """
+        Checks if version-release tag (primary not floating tag) exists already,
+        and fails plugin if it does.
+        """
+        primary_images = get_primary_images(self.workflow)
+
+        if not primary_images:
+            return
+
+        vr_image = None
+        for image in primary_images:
+            if '-' in image.tag:
+                vr_image = image
+                break
+        if not vr_image:
+            return
+
+        should_fail = False
+        for registry_name, registry in self.registries.items():
+            pullspec = vr_image.copy()
+            pullspec.registry = registry_name
+            insecure = registry.get('insecure', False)
+            secret = registry.get('secret', None)
+
+            manifest_list = get_manifest_list(pullspec, registry_name, insecure, secret)
+
+            if manifest_list:
+                self.log.error("Primary tag already exists in registry: %s", pullspec)
+                should_fail = True
+
+        if should_fail:
+            raise RuntimeError("Primary tag already exists in registry")
+
     def group_manifests_and_tag(self, session, worker_digests):
         """
         Creates a manifest list or OCI image index that groups the different manifests
@@ -364,6 +399,7 @@ class GroupManifestsPlugin(PostBuildPlugin):
 
     def run(self):
         digests = dict()
+        self.check_existing_vr_tag()
         for registry, source in self.sort_annotations().items():
             session = self.get_registry_session(registry)
 
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -106,6 +106,14 @@ class BumpReleasePlugin(PreBuildPlugin):
 
             suffix += 1
 
+    def check_build_existence_for_explicit_release(self, component, version, release):
+        build_info = {'name': component, 'version': version, 'release': release}
+        self.log.debug('checking that the build does not exist: %s', build_info)
+        build = self.xmlrpc.getBuild(build_info)
+        if build:
+            raise RuntimeError('build already exists in Koji: {}-{}-{} ({})'
+                               .format(component, version, release, build.get('id')))
+
     def run(self):
         """
         run the plugin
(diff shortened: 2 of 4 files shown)


atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -208,6 +209,36 @@ def build_list(self, manifests):
                 ],
         }, indent=4)
 
+    def check_existing_vr_tag(self):
+        self.primary_images = get_primary_images(self.workflow)
```
athos-ribeiro: Is there any reason for setting this instance var here instead of keeping it within this method? i.e., why do we need `self` here?

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -208,6 +209,36 @@ def build_list(self, manifests):
...
+            secret = registry.get('secret', None)
+
+            manifest_list = get_manifest_list(pullspec, registry_name, insecure, secret)
+
+            if manifest_list:
+                self.log.error("Primary tag already exists in registry: %s", pullspec)
+                should_fail = True
+
+        if should_fail:
+            raise RuntimeError("Primary tag already exists in registry")
```
athos-ribeiro: Do we want to print a list with the registries containing the tag here?

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -364,6 +395,7 @@ def get_registry_session(self, registry):
 
     def run(self):
         digests = dict()
+        self.check_existing_vr_tag()
```
athos-ribeiro: The design related to this PR says this check should be performed in `tag_manifest_into_registry`, although it seems that with this approach we would avoid tagging something in a registry prior to raising an error (since `tag_manifest_into_registry` loops through registries) The downside is to connect to each registry here and back again below. Is this the reason you decided for this implementation?

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -106,6 +106,13 @@ def get_next_release_append(self, component, version, base_release):
 
             suffix += 1
 
+    def check_build_existence_for_explicit_release(self, component, version, release):
+        build_info = {'name': component, 'version': version, 'release': release}
+        self.log.debug('checking that the build does not exist: %s', build_info)
+        build = self.xmlrpc.getBuild(build_info)
+        if build:
+            raise RuntimeError('build already exists in Koji: {}'.format(build_info))
```
athos-ribeiro: Could we inform the build ID here?

atomic_reactor/plugins/post_group_manifests.py
rcerven: no, I can just use local var

atomic_reactor/plugins/post_group_manifests.py
rcerven: I print it in the log.error above, for each occurence

atomic_reactor/plugins/post_group_manifests.py
rcerven: I just wanted to separate logic, and 1st do this check
and if all is good, we will proceed with tagging 

keep in mind that "looping through registries", there is more likely just 1 registry

atomic_reactor/plugins/pre_bump_release.py
rcerven: I did this because it is reported like that also in other methods, but if we really want I can report string: component-version-release

athos-ribeiro: I mean the koji buildID (for the existing build), wouldn't that be nice for the users to have in case one reaches this code path?

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -208,6 +209,36 @@ def build_list(self, manifests):
...
 
+    def check_existing_vr_tag(self):
+        self.primary_images = get_primary_images(self.workflow)
+
+        if not self.primary_images:
+            return
+
+        vr_image = None
+        for image in self.primary_images:
+            if '-' in image.tag:
```
MartinBasti: is this detection reliable? Checking just `-` seems too generic to me

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -136,6 +135,16 @@ def run(self):
...
             raise RuntimeError('missing label: {}'.format(version_label))
 
+        try:
+            _, release = labels.get_name_and_value(Labels.LABEL_TYPE_RELEASE)
+            if not self.append:
+                self.log.debug("release set explicitly so not incrementing")
+                if not is_scratch_build():
+                    self.check_build_existence_for_explicit_release(component, version, release)
+                return
+        except KeyError:
```
MartinBasti: Which method raises` KeyError`? This try-except block seems to be too wide to me, it should be minimalized with using try-except-else

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -208,6 +209,36 @@ def build_list(self, manifests):
                 ],
         }, indent=4)
 
+    def check_existing_vr_tag(self):
```
MartinBasti: maybe docstring that even explains what vr_tag is could be useful here.

atomic_reactor/plugins/post_group_manifests.py
twaugh: It is reliable for the moment, yes, and this method is used elsewhere in atomic-reactor (see image_image, koji_upload, and koji_import).

See OSBS-7169 for a better way of handling this in the future.

atomic_reactor/plugins/pre_bump_release.py
rcerven: sure

atomic_reactor/plugins/pre_bump_release.py
rcerven: just the get release, I'll put other things outside of Try

atomic_reactor/plugins/post_group_manifests.py
rcerven: sure

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -106,6 +106,14 @@ def get_next_release_append(self, component, version, base_release):
 
             suffix += 1
 
+    def check_build_existence_for_explicit_release(self, component, version, release):
+        build_info = {'name': component, 'version': version, 'release': release}
+        self.log.debug('checking that the build does not exist: %s', build_info)
+        build = self.xmlrpc.getBuild(build_info)
+        if build:
+            raise RuntimeError('build already exists in Koji: {}-{}-{} ({})'
```
athos-ribeiro: Thanks!

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -208,6 +209,40 @@ def build_list(self, manifests):
...
+
+        if not primary_images:
+            return
+
+        vr_image = None
+        for image in primary_images:
+            if '-' in image.tag:
+                vr_image = image
+                break
+        if not vr_image:
```
twaugh: could be `else:`, as in `for ... else ...`, and then you also wouldn't need to initialise it to `None` first. (But not everyone likes that construct...)

rcerven:
rn updated:
- Fail build when vr tag for image already exists
- Fail build when explicit release is set in Dockerfile and koji build already exists (for non-scratch builds)

PR #1198, opened 2019-04-18
Verify parent images manifest digests against manifests stored in koji

athos-ribeiro:


atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -342,10 +275,6 @@ def _validate_platforms_in_image(self, image):
             self.log.info('Skipping validation of available platforms '
                           'because expected platforms are unknown')
             return
-        if len(expected_platforms) == 1:
```
athos-ribeiro: We were not trying to fetch a manifest list for single platform builds before. If we want to keep this behavior, we would want to change the design to compare the v2s2 manifest digest with the v2s2 manifest digest in the koji build metadata. Do we ever save the v2s2 digest in the koji build metadata?

Is there any case where we do not generate the manifest list?

The other approach here would be to save the media type (as in the design). **If** the manifest list is not available, we would then try to compare the digest with the digest of a v2s2 manifest in koji and fail if such digest is either missing or different from what we calculated.

thouhgs?

athos-ribeiro: While talking to @rcerven offline, we agreed that, given there is a codepath which can allow images being pushed without a manifest list, we should try to compare the digests of the v2s2 manifests (in koji and in the registry).

I just pushed an update for that.

atomic_reactor/schemas/user_params.json
```diff
@@ -36,25 +36,9 @@
...
-              "pattern": "@.+:"
-            }
-          },
-          "uniqueItems": true,
-          "additionalProperties": false
-        }
-      },
-      "uniqueItems": true,
-      "additionalProperties": false
+      "additionalProperties": true
```
MartinBasti: Marking as TODO

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -83,18 +83,49 @@ def run(self):
...
-
+            else:
+                self._parent_builds[img] = self.wait_for_parent_image_build(nvr) if nvr else None
+
+            if self._parent_builds[img]:
+                mismatch = self.check_manifest_list_digest(local_tag, self._parent_builds[img])
+                if mismatch:
+                    manifest_list_mismatches.append(mismatch)
+            else:
+                self.log.warning('Could not get koji build info for parent image {}. '
```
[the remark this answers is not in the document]
MartinBasti: Note: Change to error instead warning will break osbs-box. But this question is still valid, do we want limit parent images to koji only?

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -97,122 +102,65 @@ def run(self):
...
-        if image in parents_digests:
-            # we already have recorded digests for the image, keep it the same
-            # to prevent race conditions at workers
-            return
+        image_name = image.to_str(tag=False)
+        new_image = '{}@sha256:{}'.format(image_name, digest)
+        return ImageName.parse(new_image)
 
+    def _store_manifest_digest(self, image):
+        """Store manifest list or v2 schema 2 manifest digest """
```
MartinBasti: I'm not very comfortable with mixing 2 kinds of information into the same data structure. Could we separate v2 digest and list digest at some level?

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -97,122 +102,65 @@ def run(self):
...
-        if not digests:
-            return None
-        platform_digest = digests.get(platform)
-        if platform_digest is None:
-            # exact match is not found, get random platform
-            platform_digest = tuple(digests.values())[0]
-
-        new_image = ImageName.parse(platform_digest)
-        return new_image
+        parent_metadata = {'digests': manifest_digest, 'media_type': media_type}
```
MartinBasti: May happen a case when `v2_list` digest will be overwritten by `v2` manifest digest

MartinBasti: Maybe we should error here if value already exists and we are trying to replace it with something else

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -83,18 +83,49 @@ def run(self):
...
             if nvr == self._base_image_nvr:  # don't look up base image a second time
                 self._parent_builds[img] = self._base_image_build
-                continue
-            self._parent_builds[img] = self.wait_for_parent_image_build(nvr) if nvr else None
-
+            else:
+                self._parent_builds[img] = self.wait_for_parent_image_build(nvr) if nvr else None
+
+            if self._parent_builds[img]:
+                mismatch = self.check_manifest_list_digest(local_tag, self._parent_builds[img])
```
MartinBasti: Wouldn't be better to consistently raise and catch `ValueError`/`ValidationError` exception? Instead of returning emsg? Seems more pythonic to me

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -83,18 +83,49 @@ def run(self):
...
+
+        if manifest_list_mismatches:
+            raise RuntimeError(manifest_list_mismatches)
         return self.make_result()
 
+    def check_manifest_list_digest(self, image, build_info):
+        media_type = self.workflow.builder.parent_images_metadata[image.to_str()]['media_type']
+        try:
+            expected_digest = build_info['extra']['image']['index']['digests'][media_type]
+        except KeyError:
```
MartinBasti: Missing coverage

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -97,122 +102,65 @@ def run(self):
...
-                             'because platform descriptors are not defined')
-            return
-
-        for manifest in manifest_list_dict['manifests']:
-            arch = manifest['platform']['architecture']
+        if manifest_list:
+            manifest_digest = get_checksums(BytesIO(manifest_list.content), ['sha256'])
+            media_type = get_manifest_media_type('v2_list')
+        else:
+            digests_dict, _ = get_all_manifests(image, image.registry,
```
MartinBasti: Missing coverage

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -362,7 +306,12 @@ def _validate_platforms_in_image(self, image):
         manifest_list = self._get_manifest_list(image)
 
         if not manifest_list:
-            raise RuntimeError('Unable to fetch manifest list for base image')
+            if len(expected_platforms) == 1:
+                self.log.warning('Skipping validation of available platforms for base image: '
```
MartinBasti: Missing coverage

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -97,122 +102,65 @@ def run(self):
...
-        For standard platforms like x86_64, ppc64le, ... this method returns
-        the corresponding digest
-
-        """
-        parents_digests = self.workflow.builder.parent_images_digests
-        try:
-            digests = parents_digests.get_image_digests(image)
-        except KeyError:
-            return None
+                raise RuntimeError('Could not extract manifest list or '
```
rcerven: instead of raising exception in this method, it should collect all errors for all missing parents, and raise exception in run() at the end

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: mmmm how it could be overwritten?
it either has manifest_list or it doesn't

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: we aren't exactly missing different kind of information as we have dict :
{'digests': manifest_digest, 'media_type': media_type}

but we should probably reflect that in that comment

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -83,18 +83,49 @@ def run(self):
...
-            self._parent_builds[img] = self.wait_for_parent_image_build(nvr) if nvr else None
-
+            else:
+                self._parent_builds[img] = self.wait_for_parent_image_build(nvr) if nvr else None
+
+            if self._parent_builds[img]:
+                mismatch = self.check_manifest_list_digest(local_tag, self._parent_builds[img])
+                if mismatch:
+                    manifest_list_mismatches.append(mismatch)
+            else:
```
rcerven: this can happen only when base image (last parent image) is FROM scratch or custom_base_image

IF build can't be find in koji, wait_for_parent_image_build will raise error

so I'd just remove that else section with error message (or write there that base image is scratch or custom, and we don't expect that exist in koji)

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -83,18 +83,49 @@ def run(self):
...
+        media_type = self.workflow.builder.parent_images_metadata[image.to_str()]['media_type']
+        try:
+            expected_digest = build_info['extra']['image']['index']['digests'][media_type]
+        except KeyError:
+            self.log.error('Koji build for parent image {} does not contain manifest digest '
+                           'information'.format(image.to_str()))
+            raise RuntimeError('Koji build for parent image {} does not contain manifest digest '
+                               'information'.format(image.to_str()))
+
+        digest = image.tag
```
rcerven: digest = self.workflow.builder.parent_images_metadata[image.to_str()]['digests']

atomic_reactor/plugins/pre_pull_base_image.py
MartinBasti: we are mixing, once it stores manifest list, then it may store v2 manifest digest. If we would need to store both, this method must be rewritten completely. But can stay as is it has at least identification what kind is stored there

atomic_reactor/plugins/pre_pull_base_image.py
MartinBasti: accidents happen, server may return different type on second call. Or optimization can be done here, if record exists, don't call registry again.

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: mmmm, when we will use images build with group manifests enabled,
we will always find v2_list for all parents (which were build with group manifests enabled)

so unless somebody is mixing some very old build, which had only v2s2, with builds where group manifests were enabled, we will always have there either only all manifest_lists or only v2s2 digests

atomic_reactor/plugins/pre_pull_base_image.py
MartinBasti: for multistage builds

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: but that comment could be something like "store digest and media types for v2s2 or v2_list"

atomic_reactor/plugins/pre_koji_parent.py
MartinBasti: How are parent images from other stages checked against existence of koji build? IIUC, only base_image is checked for  being in koji (i.e. last FROM statement). How about other stages? IIRC then parents from other stages may pass even if they aren't in koji.

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: parent_images contains unique images btw

atomic_reactor/plugins/pre_koji_parent.py
rcerven: 1st it checks for base image (if not scratch or custom) with wait_for_parent_image_build line 84

then it checks for each other parent with wait_for_parent_image_build line 95

atomic_reactor/plugins/pre_pull_base_image.py
MartinBasti: ok, then

rcerven:
also would be nice if we could actually use ImageName for keys of parent_images_metadata

atomic_reactor/plugins/pre_koji_parent.py
MartinBasti: * line 91: `detect_parent_image_nvr` returns `None` when NVR is not found, because of this:
```
        if len(label_values) != len(label_names):  # don't have all the necessary labels
            self.log.info("Image '%s' NVR missing; not searching for Koji build.", image_name)
            return None
``` 
* line 95: `self._parent_builds[img] = self.wait_for_parent_image_build(nvr) if nvr else None`
`nvr is None`, no `wait_for_parent_image_build` is executed, `None` is assigned to `_parent_builds[img]`

* line 97: mismatch check is not executed because of `None`

MartinBasti: Oh, but the same applies for `_base_image_build`

rcerven: yeah true, but osbs build will fail when any of those labels are missing, so you can't have in koji build which is missing them :)

MartinBasti: OMG, that is just relying on side effects from other plugins

rcerven: mmm actually what I am thinking is, that we should report IF nvr is None,
and also consider it as not found in koji (as that is some weird image, not build by osbs)

cc @athos-ribeiro

MartinBasti: AFAIK, if `we don't allow non-koji images` this is the best place for killing that build when NVR is not found.

atomic_reactor/plugins/pre_koji_parent.py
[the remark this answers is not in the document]
athos-ribeiro: Do we want this change?

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -83,18 +83,49 @@ def run(self):
...
+            if self._parent_builds[img]:
+                mismatch = self.check_manifest_list_digest(local_tag, self._parent_builds[img])
+                if mismatch:
+                    manifest_list_mismatches.append(mismatch)
+            else:
+                self.log.warning('Could not get koji build info for parent image {}. '
+                                 'Digest check will not be performed'.format(img.to_str()))
+
+        if manifest_list_mismatches:
+            raise RuntimeError(manifest_list_mismatches)
```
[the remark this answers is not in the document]
athos-ribeiro: done

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -83,18 +83,49 @@ def run(self):
...
         return self.make_result()
 
+    def check_manifest_list_digest(self, image, build_info):
+        media_type = self.workflow.builder.parent_images_metadata[image.to_str()]['media_type']
+        try:
+            expected_digest = build_info['extra']['image']['index']['digests'][media_type]
+        except KeyError:
+            self.log.error('Koji build for parent image {} does not contain manifest digest '
```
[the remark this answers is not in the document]
athos-ribeiro: done

atomic_reactor/schemas/user_params.json
athos-ribeiro: done

atomic_reactor/plugins/pre_pull_base_image.py
athos-ribeiro: I changed the structure here.

* re-use the parent_images_digests param to avoid having osbs-client changes (it is also a better name)
* each parent image dict will have the media_type as a key and digest as value.

Thoughts? @MartinBasti @rcerven

atomic_reactor/plugins/pre_koji_parent.py
athos-ribeiro: done

atomic_reactor/plugins/pre_pull_base_image.py
athos-ribeiro: done

atomic_reactor/plugins/pre_koji_parent.py
athos-ribeiro: done

athos-ribeiro:
I added a new commit to avoid having you going through the whole patch again (I will squash it after the review is done).

atomic_reactor/plugins/pre_pull_base_image.py
MartinBasti: Sounds good

atomic_reactor/plugins/pre_koji_parent.py
athos-ribeiro: > mmm actually what I am thinking is, that we should report IF nvr is None,
> and also consider it as not found in koji (as that is some weird image, not build by osbs)
> 
> cc @athos-ribeiro

If the NVR is not found, we are missing proper Labels. When will the build fail? This does pass in osbs-box when using a Fedora image without the propar labels (and as consequence, we cannot check for digests).

rcerven: I meant that build of such parent would fail with osbs

athos-ribeiro: Ok, so we should raise RuntimeError if nvr is None and the build is neither custom base nor from scrach. is that it?

@rcerven @MartinBasti.

Warning: this will break osbs-box, but seems to be the safe approach

atomic_reactor/plugins/pre_pull_base_image.py
twaugh: It's not really mixed. They're all manifest digests. Manifests come in different flavours. It's the same sort of idea as the platform specifier: a digest and a media type.

tests/plugins/test_pull_base_image.py
```diff
@@ -533,15 +533,16 @@ def workflow_callback(workflow):
                                     check_platforms=True)
         assert log_message in caplog.text
 
-    def test_single_platform_build(self, caplog):
+    @pytest.mark.parametrize('fetch_manifest_list', (True, False))
```
rcerven: this isn't used

tests/plugins/test_pull_base_image.py
```diff
@@ -533,15 +533,16 @@ def workflow_callback(workflow):
...
+    def test_single_platform_build(self, caplog, fetch_manifest_list):
 
         def workflow_callback(workflow):
             workflow = self.prepare(workflow)
             workflow.prebuild_results[PLUGIN_CHECK_AND_SET_PLATFORMS_KEY] = set(['x86_64'])
             return workflow
 
-        log_message = 'single platform build'
-        test_pull_base_image_plugin(LOCALHOST_REGISTRY, BASE_IMAGE,
+        log_message = 'Base image is a manifest list for all required platforms'
```
rcerven: this should be based on 'fetch_manifest_list' and if that is False
log_message should be : 
'Skipping validation of available platforms for base image: '
'this is a single platform build and base image has no manifest '
'list'

tests/plugins/test_pull_base_image.py
```diff
@@ -851,22 +865,64 @@ def workflow_callback(workflow):
...
-                "Cannot resolve platform 'x86_64' specific digest for image "
+                "Cannot resolve manifest digest for image "
                 "'registry.example.com/{}'".format(BASE_IMAGE))
         else:
             replacing_msg = ("Replacing image 'registry.example.com/{}' with "
                              "'{}@sha256:123456'".format(
                                 BASE_IMAGE, reg_image_no_tag))
         assert replacing_msg in caplog.text
 
+    def test_single_platform_build_no_manifest_list(self, caplog):
```
rcerven: this should be combined with 'test_single_platform_build' test

tests/plugins/test_pull_base_image.py
```diff
@@ -851,22 +865,64 @@ def workflow_callback(workflow):
...
+
+        log_message = 'base image has no manifest list'
+        test_pull_base_image_plugin(LOCALHOST_REGISTRY, BASE_IMAGE,
+                                    [], [], reactor_config_map=True,
+                                    inspect_only=False,
+                                    workflow_callback=workflow_callback,
+                                    check_platforms=True)
+        assert log_message in caplog.text
+
+    def test_single_platform_build_no_manifests(self, caplog):
```
rcerven: maybe even this could be combined with 'test_single_platform_build'

atomic_reactor/plugins/pre_koji_parent.py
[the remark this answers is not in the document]
athos-ribeiro: done

atomic_reactor/plugins/pre_pull_base_image.py
athos-ribeiro: I did change the docstring here. The method is slightly different from when the comment was added

atomic_reactor/plugins/pre_pull_base_image.py
athos-ribeiro: done

atomic_reactor/plugins/pre_koji_parent.py
athos-ribeiro: This is done now. @cverna would this break Fedora workflow?

atomic_reactor/util.py
```diff
@@ -466,41 +466,48 @@ def _process_plugin_substitution(mapping, key_parts, value):
...
+    """
+    Compute file checksums in given hash objects.
+
+    :param fd: file-like object
+    :param hash_objs: dict, maps algorithms to its hash objects from hashlib
+    :param blocksize: block size used to read fd
+    """
+    buf = fd.read(blocksize)
+    while len(buf) > 0:
+        for _, hash_object in hash_objs.items():
```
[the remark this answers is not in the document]
athos-ribeiro: done

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -99,122 +102,65 @@ def run(self):
             if self.check_platforms:
                 # run only at orchestrator
                 self._validate_platforms_in_image(image)
-                self._collect_image_digests(image)
+                self._store_manifest_digest(image)
 
             # try to stay with digests
```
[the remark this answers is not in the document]
athos-ribeiro: Removed

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -99,122 +102,65 @@ def run(self):
...
-            image_with_digest = self._get_image_with_digest(image, current_platform)
+            image_with_digest = self._get_image_with_digest(image)
             if image_with_digest is None:
-                self.log.warning("Cannot resolve platform '%s' specific digest for image '%s'",
+                self.log.warning("Cannot resolve platform '%s' manifest digest for image '%s'",
```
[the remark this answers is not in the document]
athos-ribeiro: The stirng was changed

atomic_reactor/plugins/pre_koji_parent.py
cverna: Yes this will most likely break osbs in Fedora, since we are not building the Fedora base image in OSBS for the following reasons
 - lack of OSBS cluster for non x86_64 arches
 - cannot publish images to Docker Hub official library using OSBS.

We are currently working on getting an aarch64 cluster but we are low on hardware resources, and I think it will take a while for us to have a ppc64le, armv7 and s390x cluster.

athos-ribeiro: Thanks for the feedback, @cverna.

I will follow-up with a PR adding an option to enable builds with external (non-osbs-built) base images.

athos-ribeiro:
Release notes added:

- Verify parent images digests stored in koji builds against the digests of the manifests stored in the internal registry. Whenever they differ, the build will fail.
- Do not allow builds to use images that were not built with OSBS (not in koji).

PR #1199, opened 2019-04-22
New yum proxies plugin

lcarva:


lcarva:
Still need to address the TODO comments. If you have any suggestion on how to address those, please comment there :)

atomic_reactor/plugins/pre_yum_proxy.py
```diff
@@ -0,0 +1,162 @@
...
+    def __init__(self, tasker, workflow):
+        """
+        Plugin initializer
+
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        """
+        super(YumProxyPlugin, self).__init__(tasker, workflow)
+
+        # TODO: Is this dir accessbile in the build context?
```
twaugh: Yes, not removed until after exit plugins have run.
https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/inner.py#L525

atomic_reactor/plugins/pre_yum_proxy.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        try:
+            yum_proxies = get_yum_proxies(self.workflow)
+        except KeyError:
+            return
+        # Let's make it easier to search which hosts should be proxied.
+        self.yum_proxies = {info['proxied']: info['proxy'] for info in yum_proxies}
+
+        # Iterate through each parent image to find any hard coded yum repo file.
+        for parent_image_id in self.workflow.builder.parent_images.keys():
```
twaugh: Don't need `.keys()` here

atomic_reactor/plugins/pre_yum_proxy.py
```diff
@@ -0,0 +1,162 @@
...
+        # TODO: This needs a retry
+        dest_path = os.path.join(self.workdir, 'original', image_id)
+        try:
+            os.makedirs(dest_path)
+        except OSError:
+            # Ignore if dirs have already been created. Python 2 doesn't suport exists_ok param
+            pass
+        container_id = self.tasker.copy_file(
+            image_id,
+            src_path='/etc/yum.repos.d',
```
twaugh: Wonder whether a trailing slash ('/etc/yum.repos.d/') would avoid us needing to search within ./yum.repos.d in _parse_yum_repos?

atomic_reactor/plugins/pre_yum_proxy.py
```diff
@@ -0,0 +1,162 @@
...
+        )
+        self._container_ids.append(container_id)
+        return self._parse_yum_repos(dest_path)
+
+    def _parse_yum_repos(self, dest_path):
+        dest_path = os.path.join(dest_path, 'yum.repos.d')
+        yum_repo_files = {}
+        for repo_file_path in os.listdir(dest_path):
+            repo_file_content = ConfigParser()
+            repo_file_content.read(os.path.join(dest_path, repo_file_path))
```
twaugh: Maybe we should ignore parse errors?

atomic_reactor/plugins/pre_yum_proxy.py
```diff
@@ -0,0 +1,162 @@
...
+        for parent_image_id, yum_repo_files in self.all_yum_repo_files.items():
+            for yum_repo_file_name, yum_repo_file_content in yum_repo_files.items():
+                rewrite_file = False
+                for yum_repo_name in yum_repo_file_content.sections():
+                    baseurl = yum_repo_file_content.get(yum_repo_name, 'baseurl')
+                    if not baseurl:
+                        continue
+                    parts = urlsplit(baseurl)
+                    if parts.netloc in self.yum_proxies:
+                        # TODO: Do we also need to replace URL schema?
```
twaugh: You mean the scheme, as in 'https', 'http' etc? Hmm, not sure.

atomic_reactor/plugins/pre_yum_proxy.py
```diff
@@ -0,0 +1,162 @@
...
+                        proxied_baseurl = urlunsplit(
+                            (parts[0], self.yum_proxies[parts.netloc]) + parts[2:])
+                        yum_repo_file_content.set(yum_repo_name, 'baseurl', proxied_baseurl)
+                        rewrite_file = True
+
+                if rewrite_file:
+                    dest_path = os.path.join(self.workdir, 'updated', parent_image_id,
```
twaugh: Could put 'updated' in a variable both places can use.

atomic_reactor/plugins/pre_yum_proxy.py
```diff
@@ -0,0 +1,162 @@
...
+            # TODO: will this ^ ignore SCRATCH_FROM ? Needs testing, but it looks like
+            # SCRATCH_FROM are exluded from parent_images .
+            self.all_yum_repo_files[parent_image_id] = self._get_yum_repos(parent_image_id)
+
+        self._generate_new_yum_repo_files()
+        self._update_dockerfile()
+
+    def _get_yum_repos(self, image_id):
+        # TODO: This needs a retry
+        dest_path = os.path.join(self.workdir, 'original', image_id)
```
twaugh: Could put 'original' in a variable both places can use.

atomic_reactor/plugins/pre_yum_proxy.py
```diff
@@ -0,0 +1,163 @@
...
+        try:
+            yum_proxies = get_yum_proxies(self.workflow)
+        except KeyError:
+            return
+        # Let's make it easier to search which hosts should be proxied.
+        self.yum_proxies = {info['proxied']: info['proxy'] for info in yum_proxies}
+
+        # Iterate through each parent image to find any hard coded yum repo file.
+        for parent_image_id in self.workflow.builder.parent_images.keys():
+            # TODO: will this ^ ignore SCRATCH_FROM ? Needs testing, but it looks like
```
rcerven: yeah parent_images doesn't contain FROM SCRATCH parent

atomic_reactor/plugins/pre_yum_proxy.py
```diff
@@ -0,0 +1,162 @@
...
+                    self.updated_yum_repo_files.setdefault(parent_image_id, [])
+                    self.updated_yum_repo_files[parent_image_id].append(yum_repo_file_name)
+
+    def _write_yum_repo_file(self, dest_path, content):
+        try:
+            os.makedirs(os.path.dirname(dest_path))
+        except OSError:
+            # Ignore if dirs have already been created. Python 2 doesn't suport exists_ok param
+            pass
+        with open(dest_path, 'w') as f:
```
twaugh: Do we need to take special action to preserve the file mode of the original file?

atomic_reactor/plugins/pre_yum_proxy.py
```diff
@@ -0,0 +1,162 @@
...
+        for structure in dfp.structure:
+            if structure['instruction'] != 'FROM':
+                continue
+            found_parent_image_id = structure['value'].split(' ')[0]
+            if found_parent_image_id == parent_image_id:
+                return structure
+
+        raise RuntimeError('Unable to find parent image instruction')
+
+    def _find_last_stage_structure(self, dfp, parent_structure):
```
twaugh: Could use a docstring. I original thought this method was intending to find the start of the final stage, but it seems to actually be looking for the end of the current stage.

atomic_reactor/plugins/pre_yum_proxy.py
lcarva: Yes. Otherwise we'll have to have the internal proxied host to use an identity cert that has been issued by an external entity. If we don't do that, we'll have to bake the internal root CA into the image being built?

lcarva:
Closing this as we'll have a different approach to the problem.

PR #1202, opened 2019-04-24
 pre_hide_files: iterate over stages, not parent images

twaugh:


atomic_reactor/plugins/pre_hide_files.py
```diff
@@ -43,8 +43,28 @@ def run(self):
         self._populate_start_file_lines()
         self._populate_end_file_lines()
 
-        for parent_image_id in self.workflow.builder.parent_images.values():
-            self._update_dockerfile(parent_image_id)
+        # Find limits of each Dockerfile stage
+        self.dfp = df_parser(self.workflow.builder.df_path)
```
MartinBasti: I may be wrong, but it seems that `self.dfp` is initialized out of `__init__`.  Is actually dfp needed as object attribute, could be just variable?

atomic_reactor/plugins/pre_hide_files.py
```diff
@@ -43,8 +43,28 @@ def run(self):
         self._populate_start_file_lines()
         self._populate_end_file_lines()
 
-        for parent_image_id in self.workflow.builder.parent_images.values():
-            self._update_dockerfile(parent_image_id)
+        # Find limits of each Dockerfile stage
```
MartinBasti: Could be this whole block extracted into method like `_update_build_stages` to have `run` written as steps on the same level of abstraction

atomic_reactor/plugins/pre_hide_files.py
MartinBasti: oh, found it. But still it should be defined in `__init__` as None at least

atomic_reactor/plugins/pre_hide_files.py
twaugh: Good idea. Pushed a change, see what you think.

atomic_reactor/plugins/pre_hide_files.py
MartinBasti: (lazy evaluated property would be the best)

twaugh: Thanks, updated.
It has to be None in `__init__` because an earlier plugin modifies the file.

twaugh: I think we want to explicitly parse during `run()` to make sure we don't parse too early.

MartinBasti: lazy property should be parsed at first usage, that should happen somewhere inside `run()`, but I'm fine with `None`

twaugh:
No release notes update needed.

PR #1204, opened 2019-05-02
Disallow usage of arrangements <= 5

chmeliik:
* OSBS-6491

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -435,9 +435,7 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
             return
 
         if arrangement_version <= 5:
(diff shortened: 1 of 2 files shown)


rcerven:
IF this is suppose to also remove prod builds we should also remove checking build_result (result from build plugin but in this case result from docker plugin)
https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/util.py#L1361
https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/util.py#L1371

as those won't have any more any info because docker plugin didn't run in orchestrator, but orchestrator plugin did, and in result is only annotations['worker-build']

chmeliik:
Rebased on master

MartinBasti:
Release notes:

> # Build JSON API changes
> - `arrangement_version` <= 5 support has been removed (in version 1.6.40)

PR #1206, opened 2019-05-06
Use travis_wait to avoid CI build time out

athos-ribeiro:
Travis builds have been frequently timing out while fetching packages
from Fedora or Pypi. This commit makes travis wait 20 minutes before
starting the timeout countdown. See
https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
for reference.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ env:
 install:
   - pip install coveralls
 script:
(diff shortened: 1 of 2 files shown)


athos-ribeiro:
No release notes added: CI change only

PR #1208, opened 2019-05-06
skip import image plugin when no floating tags are present

rcerven:

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -1369,9 +1369,9 @@ def get_primary_images(workflow):
 def get_floating_images(workflow):
     floating_images = workflow.tag_conf.floating_images
     if not floating_images:
-        floating_images = [
-            ImageName.parse(floating) for floating in
-            workflow.build_result.annotations['repositories']['floating']]
+        floating_results = workflow.build_result.annotations['repositories'].get('floating', [])
+        if floating_results:
+            floating_images = [ImageName.parse(floating) for floating in floating_results]
 
     return floating_images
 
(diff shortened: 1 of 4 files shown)


atomic_reactor/util.py
```diff
@@ -1369,9 +1369,11 @@ def get_primary_images(workflow):
...
     floating_images = workflow.tag_conf.floating_images
     if not floating_images:
-        floating_images = [
-            ImageName.parse(floating) for floating in
-            workflow.build_result.annotations['repositories']['floating']]
+        floating_results = workflow.build_result.annotations['repositories'].get('floating')
+        if floating_results:
+            floating_images = [
+                ImageName.parse(floating) for floating in
+                workflow.build_result.annotations['repositories']['floating']]
```
athos-ribeiro: this could be `floating_results` (and maybe it would all fit in a single line)

atomic_reactor/util.py
```diff
@@ -1369,9 +1369,9 @@ def get_primary_images(workflow):
 def get_floating_images(workflow):
     floating_images = workflow.tag_conf.floating_images
     if not floating_images:
-        floating_images = [
-            ImageName.parse(floating) for floating in
-            workflow.build_result.annotations['repositories']['floating']]
+        floating_results = workflow.build_result.annotations['repositories'].get('floating')
```
twaugh: I think this should be `.get('floating', [])`?

(If so, can we make sure one of the test cases catches it?)

rcerven: added

rcerven:
rn not updated: fix for unreleased feature

PR #1211, opened 2019-05-07
Revert "Verify parent images manifest digests against koji"

athos-ribeiro:
This reverts commit 05570c4937646e675a9893eb62abb87827b46d60.


MartinBasti:
Release notes reverted:

> - Verify parent images digests stored in koji builds against the digests of the manifests stored in the internal registry. Whenever they differ, the build will fail.
> - Do not allow builds to use images that were not built with OSBS (not in koji).

PR #1207, opened 2019-05-06
Add option to allow non-OSBS build parent images

athos-ribeiro:
With the allow_external_base_image option set to True, do not stop
(fail) builds when a parent image is not found in Koji. Instead, raise a
working, skip digests checking and continue.

Signed-off-by: Athos Ribeiro <athos@redhat.com>

PR #1212, opened 2019-05-13
 Verify parent images manifest digests against koji 

athos-ribeiro:
#1198 was reverted due to errors introduced when a build would be triggered by a change in a parent image. This PR re-introduces those changes, including a patch for handling autorebuilds.


MartinBasti:
your second commit misses the issue ID

rcerven:
looks good, maybe we should also add test case for that specific case

athos-ribeiro:
Documentation will follow with #1207 

Release notes added:

- Verify parent images digests stored in koji builds against the digests of the manifests stored in the internal registry. Whenever they differ, the build will fail.
- Do not allow builds to use images that were not built with OSBS (not in koji).

PR #1213, opened 2019-05-16
Build error messages from str(), not repr()

chmeliik:
OSBS-7183

In places where the exception handling was generic (`except Exception`), I decided to keep the class of exception in the error message (it can be useful, and is also tested in some tests).

### DIFF
--- a/atomic_reactor/plugins/pre_add_filesystem.py
+++ b/atomic_reactor/plugins/pre_add_filesystem.py
@@ -342,14 +342,15 @@ class AddFilesystemPlugin(PreBuildPlugin):
                 self.session.cancelTask(task_id)
                 self.log.info('task %s canceled', task_id)
             except Exception as exc:
-                self.log.info("Exception while canceling a task (ignored): %r", exc)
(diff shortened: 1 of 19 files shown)


atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -342,14 +342,15 @@ def run_image_task(self, image_build_conf):
...
+                self.log.info("Exception while canceling a task (ignored): %s",
+                              util.exception_message(exc))
 
         if task.failed():
             try:
                 # Koji may re-raise the error that caused task to fail
                 task_result = self.session.getTaskResult(task_id)
             except Exception as exc:
-                task_result = repr(exc)
+                task_result = str(exc)
```
rcerven: use exception_message(exc) here?

chmeliik: thanks, missed that one

MartinBasti:
No release notes, internal change only

PR #1217, opened 2019-05-31
Fix pylint warnings, part 1

chmeliik:
These are the easy fixes, part 2 (and probably 3) is going to be worse.
* OSBS-7206

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -56,7 +56,7 @@ class BuildResults(object):
 
 
 class BuildResultsEncoder(json.JSONEncoder):
-    def default(self, obj):
+    def default(self, obj):  # pylint: disable=method-hidden
         if isinstance(obj, BuildResults):
             return {
                 'build_logs': obj.build_logs,
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -297,14 +297,14 @@ class SendMailPlugin(ExitPlugin):
         return (subject_template % formatting_dict, body_template % formatting_dict, log_files)
 
     def _send_mail(self, receivers_list, subject, body, log_files=None):
+        """Sends a mail with `subject` and `body` and optional log_file attachments
+        to all members of `receivers_list`."""
         if not receivers_list:
             self.log.info('no valid addresses in requested addresses. Doing nothing')
             return
 
         self.log.info('sending notification to %s ...', receivers_list)
 
-        """Actually sends the mail with `subject` and `body` and optionanl log_file attachements
-        to all members of `receivers_list`."""
         if log_files:
             msg = MIMEMultipart()
             msg.attach(MIMEText(body))
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -43,7 +43,7 @@ from osbs.utils import Labels
 # RFC 2821 defines a valid email address much more expansively, but this definition matches
 # what most people loosely expect to be valid.
 # specifically, email addresses of the form id/name@domain.tla are not valid.
-VALID_EMAIL_REGEX = '^[a-zA-Z][a-zA-Z0-9-_\.\+]+@[a-zA-Z0-9-\.]+\.[a-zA-Z0-9-]+$'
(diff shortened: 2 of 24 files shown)


chmeliik:
Turns out that python2 method was actually a python3 method

tests/test_koji_util.py
```diff
@@ -8,8 +8,7 @@
 
 from __future__ import absolute_import, print_function, unicode_literals
 import time
-
-from requests.exceptions import ConnectionError
+import requests
```
MartinBasti: newline between stlib imports and 3rd party lib imports was intentional

atomic_reactor/inner.py
```diff
@@ -56,7 +56,11 @@ class BuildResults(object):
 
 
 class BuildResultsEncoder(json.JSONEncoder):
-    def default(self, obj):
+    def __init__(self, **kwargs):
+        kwargs['default'] = self._default
+        super(BuildResultsEncoder, self).__init__(**kwargs)
```
MartinBasti: I'd probably just disable pylint check here, JSONEncoder states clearly that `default` should be implemented in subclass.

Doesn't this fix the warning?
```diff
- def _default(self, obj):
+ def _default(self, o):
```

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -43,7 +43,7 @@
 # RFC 2821 defines a valid email address much more expansively, but this definition matches
 # what most people loosely expect to be valid.
 # specifically, email addresses of the form id/name@domain.tla are not valid.
-VALID_EMAIL_REGEX = '^[a-zA-Z][a-zA-Z0-9-_\.\+]+@[a-zA-Z0-9-\.]+\.[a-zA-Z0-9-]+$'
```
MartinBasti: Using raw string for regexp is ok. However escaping `.` and `+` inside `[ ]` is not needed and can be removed in favor of readability

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -236,7 +236,7 @@ def _pull_and_tag_image(self, image, build_json, nonce):
                 # need to retry with that.
                 if first_library_exc:
```
MartinBasti: Maybe
```python
if first_library_exc is not None:
```
may make pylint happy

MartinBasti: If it won't make pylint happy, definitely it will make me happy to avoid a condition when exception implements `__bool__` returning `False` :D

atomic_reactor/inner.py
chmeliik: Ok (changing the signature did not fix it)

atomic_reactor/plugins/pre_pull_base_image.py
chmeliik: I even tried
```python
if isinstance(first_library_exc, Exception):
```
Looks like pylint is not _that_ smart :D

chmeliik:
Fixed remaining raising-bad-type warnings (they don't get reported by pylint 2.3.1) and other details

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -297,14 +297,14 @@ def _render_mail(self, rebuild, success, auto_canceled, manual_canceled):
         return (subject_template % formatting_dict, body_template % formatting_dict, log_files)
 
     def _send_mail(self, receivers_list, subject, body, log_files=None):
+        """Actually sends the mail with `subject` and `body` and optional log_file attachments
+        to all members of `receivers_list`."""
```
athos-ribeiro: This could be rephrased since it will be the docstring for the method

chmeliik: I think it is a pretty decent docstring already, how would you like it rephrased? Maybe drop the "Actually"?

athos-ribeiro: > I think it is a pretty decent docstring already, how would you like it rephrased? Maybe drop the "Actually"?

Something in those lines (the 'actually' seems to be added there to refer to what was following next). Just a nitpick though.

chmeliik: Ok, changed it

athos-ribeiro:
@chmeliik, could you push to this branch again so we get CI to run?

chmeliik:
Rebased on master and fixed a logging-not-lazy warning

MartinBasti:
no Release notes needed

PR #1218, opened 2019-05-31
Catch all exceptions for get_archive and remove_container

rcerven:
*OSBS-7560

### DIFF
--- a/atomic_reactor/plugins/post_export_operator_manifests.py
+++ b/atomic_reactor/plugins/post_export_operator_manifests.py
@@ -118,8 +118,15 @@ class ExportOperatorManifestsPlugin(PostBuildPlugin):
                            IMG_MANIFESTS_PATH, ex)
             self.log.error(msg)
             raise RuntimeError('%s %s' % (msg, ex))
+
+        except Exception as ex:
+            raise RuntimeError('%s' % ex)
+
         finally:
-            self.tasker.d.remove_container(container_id)
+            try:
+                self.tasker.d.remove_container(container_id)
+            except Exception as ex:
+                self.log.warning('Failed to remove container %s: %s' % (container_id, ex))
 
         with tempfile.NamedTemporaryFile() as extracted_file:
             for chunk in bits:
--- a/tests/plugins/test_export_operator_manifests.py
+++ b/tests/plugins/test_export_operator_manifests.py
@@ -72,7 +72,8 @@ def generate_archive(tmpdir, empty=False):
 
 
 def mock_env(tmpdir, docker_tasker, has_label=True, label=True, has_archive=True,
-             scratch=False, orchestrator=False, selected_platform=True, empty_archive=False):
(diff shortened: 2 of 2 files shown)


atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -118,8 +118,15 @@ def run(self):
                            IMG_MANIFESTS_PATH, ex)
             self.log.error(msg)
             raise RuntimeError('%s %s' % (msg, ex))
+
+        except Exception as ex:
+            raise RuntimeError('%s ' % (ex))
```
MartinBasti: Do we need to catch exception here?

Also:
* extra trailing whitespace in `'%s '`
* unneeded parenthesis in `(ex)`

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -118,8 +118,15 @@ def run(self):
...
+
         finally:
-            self.tasker.d.remove_container(container_id)
+            try:
+                self.tasker.d.remove_container(container_id)
+            except Exception as ex:
+                self.log.error(ex)
```
MartinBasti: This error log could me more descriptive, like `"Failed to remove container %s: %s", container_id, ex`

MartinBasti: IMHO it should be just warning as it doesn't affect build results

* warning: something happened, shouldn't affect results but may cause issues later
* error: something happened and affect results/tasks
* critical: something happened and system is not able to work properly anymore

atomic_reactor/plugins/post_export_operator_manifests.py
rcerven: I'd say it doesn't hurt

atomic_reactor/plugins/post_export_operator_manifests.py
rcerven: yop

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -118,8 +118,15 @@ def run(self):
...
             raise RuntimeError('%s %s' % (msg, ex))
+
+        except Exception as ex:
+            raise RuntimeError('%s' % ex)
+
         finally:
-            self.tasker.d.remove_container(container_id)
+            try:
+                self.tasker.d.remove_container(container_id)
+            except Exception as ex:
```
athos-ribeiro: Do we want to proceed execution here after **any** sort of exceptions (instead of APIError and ReadTimeoutError only)?

rcerven: well it actually isn't that critical that remove_container fails, as long as it was able to get that archive

I can change commit message, but still we are now catching Exception for get_archive as well

rcerven:
rn updated: Catch all exceptions for get_archive and remove_container

PR #1219, opened 2019-06-03
Fix pylint warnings, part 2

chmeliik:
Part 2 of pylint issue fixes.

For attributes defined outside `__init__`, I usually just defined them as `None`, though in some cases it seemed possible to turn them into basic local variables so I did that.

For global statements, I just made slight workarounds to get rid of the `global` statement, but `docker_mock.py` still heavily relies on global variables.

Also, `test_tasker.py` uses `docker.Client`, which no longer exists. I don't know where (or if) these tests are used with NOMOCK, so I just added a TODO about it.

### DIFF
--- a/atomic_reactor/plugins/pre_pull_base_image.py
+++ b/atomic_reactor/plugins/pre_pull_base_image.py
@@ -71,12 +71,15 @@ class PullBaseImagePlugin(PreBuildPlugin):
             metadata = self.workflow.builder.parent_images_digests
             metadata.update(parent_images_digests)
 
+        self.manifest_list_cache = {}
+
     def run(self):
         """
         Pull parent images and retag them uniquely for this build.
         """
+        self.manifest_list_cache.clear()
+
         build_json = get_build_json()
-        self.manifest_list_cache = {}
         organization = get_registries_organization(self.workflow)
         digest_fetching_exceptions = []
         for nonce, parent in enumerate(sorted(self.workflow.builder.parent_images.keys(),
--- a/tests/docker_mock.py
+++ b/tests/docker_mock.py
@@ -43,7 +43,7 @@ mock_image = \
      'Size': 0,
      'VirtualSize': 856564160}
 
-mock_images = None
+mock_images = []
 
 mock_logs = 'uid=0(root) gid=0(root) groups=10(wheel)'
 
@@ -194,8 +194,6 @@ mock_inspect_container = {
 
 
 def _find_image(img, ignore_registry=False):
-    global mock_images
-
     tagged_img = ImageName.parse(img).to_str(explicit_tag=True)
     for im in mock_images:
         im_name = im['RepoTags'][0]
@@ -387,7 +385,7 @@ def mock_docker(build_should_fail=False,
              .and_raise(docker.errors.APIError, "xyz", response))
 
     if remember_images:
-        global mock_images
+        global mock_images  # pylint: disable=global-statement
         mock_images = [mock_image]
 
         flexmock(docker.APIClient, inspect_image=_mock_inspect)
--- a/tests/docker_mock.py
+++ b/tests/docker_mock.py
@@ -266,7 +266,7 @@ def mock_docker(build_should_fail=False,
                 inspect_should_fail=False,
                 wait_should_fail=False,
                 provided_image_repotags=None,
-                should_raise_error={},
+                should_raise_error=None,
                 remember_images=False,
                 push_should_fail=False):
     """
@@ -276,7 +276,7 @@ def mock_docker(build_should_fail=False,
     :param inspect_should_fail: True == inspect_image() will raise docker.errors.NotFound
     :param wait_should_fail: True == wait() will return 1 instead of 0
     :param provided_image_repotags: images() will contain provided image
-    :param should_raise_error: methods (with args) to raise docker.errors.APIError
+    :param should_raise_error: dict, methods (with args) to raise docker.errors.APIError
     :param remember_images: keep track of available image tags
     """
     if provided_image_repotags:
@@ -372,6 +372,7 @@ def mock_docker(build_should_fail=False,
 
     flexmock(docker.APIClient, remove_volume=lambda iid, **kwargs: remove_volume(iid))
 
+    should_raise_error = should_raise_error or {}
     for method, args in should_raise_error.items():
         response = flexmock(content="abc", status_code=123)
         if args:
--- a/tests/plugins/test_add_help.py
+++ b/tests/plugins/test_add_help.py
@@ -26,14 +26,14 @@ from flexmock import flexmock
 
 
 class Y(object):
-    pass
+    def __init__(self):
+        self.dockerfile_path = None
+        self.path = None
 
 
 class X(object):
     image_id = "xxx"
     source = Y()
-    source.dockerfile_path = None
-    source.path = None
     base_image = ImageName(repo="qwe", tag="asd")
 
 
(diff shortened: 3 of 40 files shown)


atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -71,12 +71,13 @@ def __init__(self, tasker, workflow, parent_registry=None, parent_registry_insec
...
             metadata.update(parent_images_digests)
 
+        self.manifest_list_cache = {}
+
     def run(self):
         """
         Pull parent images and retag them uniquely for this build.
         """
         build_json = get_build_json()
-        self.manifest_list_cache = {}
```
MartinBasti: Removing such initialization from run, may cause issues when we use the same instance for multiple runs. However I don't think that this is atomic-reactor case. Hopefully somebody will correct me if I'm wrong.

MartinBasti:
Needs rebase

atomic_reactor/plugins/pre_pull_base_image.py
MartinBasti: @twaugh @rcerven @athos-ribeiro

rcerven: yeah we always have just 1 plugin instance running

tests/docker_mock.py
```diff
@@ -387,8 +385,7 @@ def remove_volume(volume_name):
              .and_raise(docker.errors.APIError, "xyz", response))
 
     if remember_images:
-        global mock_images
```
MartinBasti: effectively this still works like global variable, it just bypassed warning and obfuscated code. I'm not sure if this worth fixing.

tests/test_tasker.py
```diff
@@ -40,12 +40,12 @@
 def setup_module(module):
     if MOCK:
         return
-    d = docker.Client()
+    d = docker.Client()  # TODO: current version of python-docker does not have Client anymore
```
MartinBasti: IMHO this needs investigation, we may need to freeze python-docker version in requirements.txt

tests/plugins/test_pulp.py
```diff
@@ -45,7 +45,7 @@ class X(object):
     base_image = ImageName(repo="qwe", tag="asd")
 
 
-def prepare(check_repo_retval=0, existing_layers=[],
```
MartinBasti: IMHO this changes behavior. `None` and `[]` have different meaning here.

You can use

```
_undefined = object()

def prepare(.., existing_layers=_undefined, ...):
    if existing_layers is _undefined:
       existing_layers = []
```

another way is, that all tests must be updated accordingly (preferred given the small number of tests using this function)

tests/plugins/test_rpmqa.py
```diff
@@ -49,16 +49,20 @@ def mock_logs_empty(cid, **kwargs):
     return ''
 
 
-def mock_logs_retry(cid, cache={}, **kwargs):
```
MartinBasti: Why this change in API? If needed it should be in separate commit. I don't see relation between fixing pylint warning and this change.

chmeliik: IMO this was one of the rare cases where pylint actually reported a relevant warning. This method is not really usable more than once, because there is no good way to reset the cache. You would have to do something like this:
```python
retry_cache = {}
mock_logs_retry = partial(mock_logs_retry, cache=retry_cache)
# run a test
retry_cache.clear()
# run another test
retry_cache.clear()
# ...
```
The change essentially forces you to use clean cache every time you want to mock that method.
But if you think it should be in a separate commit that's ok with me.

tests/plugins/test_pulp.py
chmeliik: Ok, updated tests

tests/docker_mock.py
chmeliik: Ok, reverted and disabled the warning

chmeliik:
Rebased on master

atomic_reactor/plugins/pre_pull_base_image.py
chmeliik: Maybe the cache could just be cleared in `run()`, just to be sure?

tests/plugins/test_rpmqa.py
MartinBasti: Yes please, use separate commit with description you provided here :)

That makes git archaeology much easier.

atomic_reactor/plugins/pre_pull_base_image.py
MartinBasti: Yes, cache invalidation is one of the 4 most common programming issues. Please just to be clear it in run.

MartinBasti:
@chmeliik Any idea why coverage decreased here? https://coveralls.io/builds/23800932/source?filename=atomic_reactor/plugins/prepub_squash.py

chmeliik:
@MartinBasti I found out it was because pylint doesn't know how to run test classes that have an `__init__` method. I also now clear cache when that plugin is run, and added a separate commit for the `mock_logs_retry` change.

tests/plugins/test_add_help.py
```diff
@@ -26,14 +26,14 @@
 
 
 class Y(object):
-    pass
+    def __init__(self):
+        self.dockerfile_path = None
+        self.path = None
```
twaugh: Alternatively, we could use `stubs.StubSource` instead of `Y`, and `stubs.StubInsideBuilder` instead of `X`.

tests/plugins/test_delete_from_registry.py
```diff
@@ -39,14 +39,14 @@
 
 
 class Y(object):
-    pass
+    def __init__(self):
+        self.dockerfile_path = None
+        self.path = None
```
twaugh: Alternatively, we could use `stubs.StubSource` instead of `Y`, and `stubs.StubInsideBuilder` instead of `X`.

tests/plugins/test_group_manifests.py
```diff
@@ -302,14 +302,14 @@ def mock_registries(registries, config, schema_version='v2', foreign_layers=Fals
 
 
 class Y(object):
-    pass
+    def __init__(self):
+        self.dockerfile_path = None
+        self.path = None
```
twaugh: Alternatively, we could use `stubs.StubSource` instead of `Y`, and `stubs.StubInsideBuilder` instead of `X`.

tests/plugins/test_tag_and_push.py
```diff
@@ -83,14 +83,14 @@
 
 
 class Y(object):
-    pass
+    def __init__(self):
+        self.dockerfile_path = None
+        self.path = None
```
twaugh: Alternatively, we could use `stubs.StubSource` instead of `Y`, and `stubs.StubInsideBuilder` instead of `X`.

tests/plugins/test_tag_by_labels.py
```diff
@@ -22,14 +22,14 @@
 
 
 class Y(object):
-    pass
+    def __init__(self):
+        self.dockerfile_path = None
+        self.path = None
```
twaugh: Alternatively, we could use `stubs.StubSource` instead of `Y`, and `stubs.StubInsideBuilder` instead of `X`.

tests/test_tasker.py
twaugh: Should we try `APIClient` and fall back to `Client`, like we've done elsewhere?

chmeliik:
@twaugh thanks, using `stubs.StubInsideBuilder` now. I felt like this change deserved a separate commit, as it is not really about fixing attributes defined outside init.

twaugh:
Not sure what's up with Travis CI. Does it need a prod?

chmeliik:
Pushed again to wake Travis up. Should I replace all those `X` classes with `StubInsideBuilder` as part of this PR? There is around 20 of them in various places.

twaugh:
No need to scramble to replace `X`, just if we happen to need to touch them I'd rather we phased them out.

twaugh:
The coverage decrease is only due to the `except ImportError` clauses handling modulemd not being available. I think we can ignore that.

MartinBasti:
No release notes needed

PR #1220, opened 2019-06-03
Add the posibility to use koji's fast upload API.

cverna:
This commit adds the possibility to set a flag in the configmap
to use or not Koji's fast upload API.
This solves https://pagure.io/fedora-infrastructure/issue/7846

osbs-docs PR : https://github.com/containerbuildsystem/osbs-docs/pull/78

### DIFF
--- a/atomic_reactor/schemas/config.json
+++ b/atomic_reactor/schemas/config.json
@@ -99,6 +99,11 @@
                         "additionalProperties": false
                     }
                 ]
+            },
+            "use_fast_upload": {
+                "description": "Use Koji's fast upload API",
+                "type": "boolean",
+                "default": true
             }
         },
         "additionalProperties": false,
--- a/atomic_reactor/koji_util.py
+++ b/atomic_reactor/koji_util.py
@@ -126,16 +126,21 @@ def koji_login(session,
     return result
 
 
-def create_koji_session(hub_url, auth_info=None):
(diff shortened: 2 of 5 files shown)


twaugh:
I think we could safely default to using fast upload, especially as the koji client uses it by default.

What do other approvers think?

MartinBasti:
@twaugh works for me

cverna:
Works for me too :+1:

MartinBasti:
@cverna it seems that there are no objections against having fast upload enabled by default. Could you please change the default?

cverna:
> @cverna it seems that there are no objections against having fast upload enabled by default. Could you please change the default?

Sure, I just updated the PR

atomic_reactor/schemas/config.json
```diff
@@ -99,6 +99,10 @@
                         "additionalProperties": false
                     }
                 ]
+            },
+            "use_fast_upload": {
+                "description": "Use Koji's fast upload API",
```
twaugh: Do we want to update this text to say that this defaults to true?

cverna: Done

athos-ribeiro:
Release notes added:

- Use Koji fast upload by default
- Add a `use_fast_upload` option to allow disabling koji fast uploads

PR #1226, opened 2019-06-13
initialize BuildStepPluginsRunner after prebuild plugins run,

rcerven:
because reactor_config plugin is changing build method

* OSBS-7530

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -471,9 +471,6 @@ class DockerBuildWorkflow(object):
             prebuild_runner = PreBuildPluginsRunner(self.builder.tasker, self,
                                                     self.prebuild_plugins_conf,
                                                     plugin_files=self.plugin_files)
-            buildstep_runner = BuildStepPluginsRunner(self.builder.tasker, self,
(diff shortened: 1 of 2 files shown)


atomic_reactor/inner.py
```diff
@@ -496,6 +493,10 @@ def build_docker_image(self):
                 self.autorebuild_canceled = True
                 raise
 
+            buildstep_runner = BuildStepPluginsRunner(self.builder.tasker, self,
```
twaugh: Can we add a comment to explain why we delay creating this?

twaugh:
Can we set up a unit test for buildstep_alias that covers this aspect of it?

rcerven:
@athos-ribeiro after initialization it has already set plugin_conf, and I would rather don't touch that from reactor_config

it makes sense to run buildstep plugin runner init, after pre build plugins, since it is dependent on them

rcerven:
unit test added

atomic_reactor/inner.py
rcerven: added

rcerven:
no rn update, fix for unreleased feature

PR #1232, opened 2019-07-03
omps_util: handle requests errors itself

MartinBasti:
OMPS class will parse requests results to provide better error details.
Don't raise exception when retries fails due wrong status.

* OSBS-7542

### DIFF
--- a/atomic_reactor/omps_util.py
+++ b/atomic_reactor/omps_util.py
@@ -49,7 +49,9 @@ class OMPS(object):
         self._token = token
         self._insecure = insecure
         self.log = logging.getLogger(self.__class__.__name__)
-        self.req_session = get_retrying_requests_session()
+
+        # this class handle status errors itself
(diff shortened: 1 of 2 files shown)


MartinBasti:
This patch doesn't work on RHEL7 due the old python-requests/python-urlib3 package

```
>>> Retry(raise_on_status=False)
TypeError: __init__() got an unexpected keyword argument 'raise_on_status'

```

athos-ribeiro:
Could we have some kind of fallback for old requests/urllib3 versions? Otherwise we should either change the approach or wait until we migrate deployments to python 3

twaugh:
Let's fall back to using the old Retries object without `raise_for_status`, and work on getting the dependency in place separately.

MartinBasti:
Fall back added

MartinBasti:
No release notes, fix of unreleased changes

PR #1236, opened 2019-07-11
Address bandit warnings: B602

athos-ribeiro:
To avoid the shell=True in the rpm query command, we query the RPM DB
with the RPM python bidings.  Note that the python 3 code path would
crash, since no stderr variable was being defined at all. This patch
consolidates the codepaths for both python 2 and 3, addressing the
related Bandit warning.

* https://bandit.readthedocs.io/en/latest/plugins/b605_start_process_with_a_shell.html
* OSBS-7453

### DIFF
--- a/tests/plugins/test_koji_upload.py
+++ b/tests/plugins/test_koji_upload.py
@@ -11,6 +11,7 @@ from __future__ import unicode_literals, absolute_import
 import json
 import os
 import platform
+import rpm
 import sys
 import tempfile
 import zipfile
@@ -62,10 +63,6 @@ NAMESPACE = 'mynamespace'
 BUILD_ID = 'build-1'
 KOJI_UPLOAD_DIR = 'upload'
 LOCAL_ARCH = platform.processor()
-if sys.version_info[0] == 2:
-    B_LOCAL_ARCH = bytes(LOCAL_ARCH)
-elif sys.version_info[0] == 3:
-    B_LOCAL_ARCH = LOCAL_ARCH.encode()
 
 
 def noop(*args, **kwargs): return None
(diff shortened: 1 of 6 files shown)


tests/plugins/test_koji_upload.py
```diff
@@ -187,27 +184,41 @@ def taskFinished(self, task_id):
         return self.tag_task_state in ['CLOSED', 'FAILED', 'CANCELED', None]
 
 
-FAKE_SIGMD5 = b'0' * 32
-FAKE_RPM_OUTPUT = (
-    b'name1;1.0;1;' + B_LOCAL_ARCH + b';0;' + FAKE_SIGMD5 + b';(none);'
-    b'RSA/SHA256, Mon 29 Jun 2015 13:58:22 BST, Key ID abcdef01234567\n'
+class MockedRpmHeader(object):
```
MartinBasti: This code seems duplicated

tests/plugins/test_koji_upload.py
```diff
@@ -187,27 +184,41 @@ def taskFinished(self, task_id):
...
+                    v = '(none)'
+                return v
 
-    b'gpg-pubkey-doc;01234567;01234567;noarch;(none);' + FAKE_SIGMD5 +
-    b';(none);(none)\n'
 
-    b'name2;2.0;2;' + B_LOCAL_ARCH + b';0;' + FAKE_SIGMD5 + b';' +
-    b'RSA/SHA256, Mon 29 Jun 2015 13:58:22 BST, Key ID bcdef012345678;(none)\n'
-    b'\n')
+class MockedTS(object):
```
MartinBasti: This seems duplicated too

tests/plugins/test_koji_upload.py
athos-ribeiro: Do you mean, duplication between koji_promote and koji_upload? I believe the duplication is intentional, while we still maintain the koji_promote plugin (which, IIRC, should be dropped in the future).

athos-ribeiro:
- Re-based on master
- added python-rpm dependency on the spec file. Not adding to requirements.txt; see https://github.com/rpm-software-management/rpm/issues/273

tests/plugins/test_koji_upload.py
MartinBasti: Yes, I haven't realized that it is koji_promote. ack

athos-ribeiro:
Rebased on master to get the  latest spec file changes in here and realized the test suite is broken due to a `flatpak_module_tools` update which requires a new version of `module-build-service`.

See https://pagure.io/flatpak-module-tools/c/52f878a91e4d5c42f25d8ed464461f32a538f538

MartinBasti:
Waiving test suite failure, unrelated to this patch.

MartinBasti:
Internal change only, no release notes needed

PR #1243, opened 2019-07-22
Libmodulemd v2 api

owtaylor:
This pull request:
 - Updates the code to use latest flatpak-module-tools and the libmodulemd v2 API
 - Enables testing against real flatpak-module-tools, flatpak, ostree on RHEL 7 (this is used in production)
 - Fixes a problem with the Flatpak tests with PyPNG that started happening recently

Since we now test against flatpak on all platforms, it would be possible to remove the mock_flatpak code paths, but this PR doesn't do that.

### DIFF
--- a/tests/plugins/test_flatpak_create_oci.py
+++ b/tests/plugins/test_flatpak_create_oci.py
@@ -486,7 +486,7 @@ COMMAND_PATTERNS = [
      MockFlatpak.build_init),
     (['ostree', 'commit',
       '--repo', '@repo',
-      '--owner-uid=0', '--owner-gid=0', '--no-xattrs',
+      '--owner-uid=0', '--owner-gid=0', '--no-xattrs', '--canonical-permissions',
       '--branch', '@branch', '-s', '@subject', '--tree=tar=@tar_tree', '--tree=dir=@dir_tree'],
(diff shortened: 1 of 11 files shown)


tests/plugins/test_flatpak_create_oci.py
```diff
@@ -67,9 +67,9 @@
 
 ICON = BytesIO()
 # create minimal 256x256 RGBA PNG
-png.Writer(256, 256, alpha=True).write(ICON,
-                                       [[0 for _ in range(4 * 256)]
-                                        for _ in range(256)])
+png.Writer(256, 256, greyscale=False, alpha=True).write(ICON,
```
athos-ribeiro: Does this fix the issue with the new pypng? If so, could you also remove the freeze on pypng 0.0.19 in tests/requirements.txt?

owtaylor: Done

owtaylor:
Pushed a new version with the stray files removed (I remember typo'ing 'git add tests/' and thinking "oops, but should be OK - sorry about that :-)"

athos-ribeiro:
Release notes added:

- Flatpak builds updated to use bindings for libmodulemd v2

PR #1246, opened 2019-08-06
Add Buildah buildstep plugin

rcerven:


atomic_reactor/plugins/build_buildah.py
```diff
@@ -0,0 +1,80 @@
...
+            # include in the failure summary.
+            err = output[-1] if output else "<buildah had bad exit code but no output>"
+            return BuildResult(
+                logs=output,
+                fail_reason="image build failed (rc={}): {}".format(ib_process.returncode, err),
+            )
+
+        image_id = builder.get_built_image_info()['Id']
+        if ':' not in image_id:
+            # Older versions of the daemon do not include the prefix
```
athos-ribeiro: Does this block still make sense here? This seems to be a docker daemon specific issue, but this image_id would have to be fetched by podman, right?

atomic_reactor/plugins/build_buildah.py
```diff
@@ -0,0 +1,80 @@
...
+            image_id = 'sha256:{}'.format(image_id)
+
+        # since we need no squash, export the image for local operations like squash would have
+        self.log.info("fetching image %s from docker", image)
```
athos-ribeiro: Same idea here, wouldn't we need to push the built image to the docker daemon storage first in orther to fetch it here?

atomic_reactor/plugins/build_buildah.py
rcerven: seems like docker daemon specific, but I have no way to verify that

atomic_reactor/plugins/build_buildah.py
rcerven: mmm? keep in mind that this plugin will use PodmanTasker which isn't implemented yet

athos-ribeiro:
LGTM, just a couple inline questions.

Should we start documenting this feature right away? I mean, should this already include a PR for osbs-docs?

Also, the 1st commit does not have a signed-off-by line. I wouldn't oppose to waiving that one for this PR though.

twaugh:
I would leave out the documentation for the time being, until we have a plugin that can actually work.

We should have Signed-Off-By though.

athos-ribeiro:
Waiving third party signed off by entry

rcerven:
rn updated: - new Buildah buildstep plugin

PR #1247, opened 2019-08-07
Deep manifest inspection improvements

athos-ribeiro:
- Fix bug which would lead a build to failure if deep manifest inspections were triggered for an operator build
- Use registry configuration to fetch manifest list

### DIFF
--- a/atomic_reactor/plugins/pre_koji_parent.py
+++ b/atomic_reactor/plugins/pre_koji_parent.py
@@ -9,7 +9,8 @@ from __future__ import print_function, unicode_literals, absolute_import
 
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.constants import (
-    INSPECT_CONFIG, PLUGIN_KOJI_PARENT_KEY, BASE_IMAGE_KOJI_BUILD, PARENT_IMAGES_KOJI_BUILDS
+    INSPECT_CONFIG, PLUGIN_KOJI_PARENT_KEY, BASE_IMAGE_KOJI_BUILD, PARENT_IMAGES_KOJI_BUILDS,
+    KOJI_BTYPE_IMAGE
 )
 from atomic_reactor.plugins.pre_reactor_config import (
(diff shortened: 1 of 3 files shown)


atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -79,6 +80,7 @@ def __init__(self, tasker, workflow, koji_hub=None, koji_ssl_certs_dir=None,
         self._base_image_build = None
         self._parent_builds = {}
         self._poll_start = None
+        self._registries = get_registries(workflow, {})
```
rcerven: use get_source_registry instead like in pull_base_image

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -209,7 +215,7 @@ def manifest_list_entries_match(self, image, build_id):
 
         archives = self.koji_session.listArchives(build_id)
         koji_archives_data = {}
-        for archive in archives:
+        for archive in [a for a in archives if a['btype'] == 'image']:
```
MartinBasti: We could use a generator expression here?

MartinBasti: And also can we use constant? Something like `KOJI_BTYPE_IMAGE` as we have for operator-manifests

atomic_reactor/plugins/pre_koji_parent.py
athos-ribeiro: Done

atomic_reactor/plugins/pre_koji_parent.py
athos-ribeiro: Done

twaugh: Still says `'image'` I think.

athos-ribeiro: my page was outdated when I commented this. I did not realize there was a second request :)

On it

PR #1254, opened 2019-08-20
Add method to check if code is running in orchestrator

chmeliik:
* OSBS-7005

The method assumes that a build is an orchestrator build if (and only if) it includes the `orchestrate_build` buildstep plugin. It is defined in `DockerBuildWorkflow`, which seems like the proper place for it given that it contains most of the build configuration. Build plugins also get an `is_in_orchestrator()` method which simply calls the `is_orchestrator_build()` method of their workflow (for convenience).

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -33,7 +33,10 @@ from atomic_reactor.plugin import (
 )
 from atomic_reactor.source import get_source_instance_for
 from atomic_reactor.constants import INSPECT_ROOTFS, INSPECT_ROOTFS_LAYERS
-from atomic_reactor.constants import CONTAINER_DEFAULT_BUILD_METHOD
+from atomic_reactor.constants import (
+    CONTAINER_DEFAULT_BUILD_METHOD,
+    PLUGIN_BUILD_ORCHESTRATE_KEY
+)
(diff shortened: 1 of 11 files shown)


atomic_reactor/inner.py
```diff
@@ -448,6 +451,28 @@ def __init__(self, source, image, prebuild_plugins=None, prepublish_plugins=None
...
+        Get the orchestrate_build plugin configuration for this workflow
+        if present (will be present for orchestrator, not for worker).
+
+        :return: orchestrate_build plugin configuration dict or None
+        """
+        for plugin in self.buildstep_plugins_conf or []:
+            if plugin['name'] == PLUGIN_BUILD_ORCHESTRATE_KEY:
+                return plugin
+        # Not an orchestrator build
+        return None
```
twaugh: Would it be more Pythonic for this to raise KeyError or something?

chmeliik: KeyError felt a bit weird, I made it a ValueError, see if you like it :-) .

twaugh: Looks great, thanks.

MartinBasti:
Release notes:
-  method `is_in_orchestrator()` has been added to `Plugin` class as a common way how to determine whether code is running in orchestrator or worker.

PR #1255, opened 2019-08-21
Fix tests for pytest version 5

chmeliik:
* OSBS-7708

This script right here is the MVP:
```python
#!/usr/bin/env python3
import re
import sys
from pathlib import Path

PATTERN = (r'with pytest\.raises\(.*?\) as (\w+):'
           r'.*?str\(\1\b(?!\.value)')
FLAGS = re.DOTALL

PAT = re.compile(PATTERN, flags=FLAGS)
REPLACE_FN = lambda match: match.group() + '.value'

file = Path(sys.argv[1])

content = file.read_text()
fixed = PAT.sub(REPLACE_FN, content)

# for cases like these:
#   with pytest.raises(Exception) as e:
#       do_stuff()
#   assert 'a' in str(e)
#   assert 'b' in str(e)
while fixed != content:
    content = fixed
    fixed = PAT.sub(REPLACE_FN, content)

file.write_text(fixed)

```
It's ugly, kinda slow, and sometimes catches false positives, but it gets the job done :D

Note that [pytest no longer supports python 2 starting with version 5](https://docs.pytest.org/en/latest/py27-py34-deprecation.html), which is why I added a `tests/requirements-py2.txt` with an older pytest version.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,10 +1,9 @@
-flexmock
+flexmock>=0.10.3
 ordereddict
 responses>=0.9.0
 pypng
-pytest==4.0.*
-pytest-cov==2.6.*
-six>=1.10.0 # required by pytest-cov
+pytest>=4.1.0
+pytest-cov
 flake8
 koji
 requests-mock
--- a/test.sh
+++ b/test.sh
@@ -78,13 +78,13 @@ fi
 $RUN $PKG install -y $PIP_PKG
 
 if [[ $OS == centos && $OS_VERSION == 7 ]]; then
+  # Get a less ancient version of pip to avoid installing py3-only packages
+  $RUN $PIP install "pip>=9.0.0,<10.0.0"
+  # ...but ancient enough to allow uninstalling packages installed by distutils
+
   # Older versions of setuptools don't understand the environment
   # markers used by docker-squash's requirements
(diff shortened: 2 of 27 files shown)


tests/requirements.txt
```diff
@@ -1,9 +1,9 @@
-flexmock
+flexmock>=0.10.3
 ordereddict
 responses>=0.9.0
 pypng
-pytest==4.0.*
-pytest-cov==2.6.*
+pytest>=4.1.0
```
chmeliik: Should we straight-up require `pytest>=5.0.0`? I believe pip installs the latest version anyway. 4.1.0 is the version which introduces an API change that breaks flexmock (which is why we need `flexmock>=0.10.3`)

chmeliik:
Turns out `pip` is not stupid enough to install a version of pytest that would not be compatible with its python version, removed specific requirements for py2 tests.

tests/requirements.txt
chmeliik: That would require having separate requirements for py2, as pytest>=5.0.0 is not available there.

chmeliik:
update: Since we are now installing a newer version of pip on centos, it is no longer necessary to install an exact version of `more-itertools`

MartinBasti:
No release notes needed, internal change in tests

PR #1257, opened 2019-08-28
Additional check for release and version labels in bump_release,

rcerven:
check that version label isn't empty,
check that release label isn't empty when specified

* OSBS-7697

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -186,24 +186,39 @@ class BumpReleasePlugin(PreBuildPlugin):
         parser = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
         dockerfile_labels = parser.labels
         labels = Labels(dockerfile_labels)
+        missing_labels = {}
+        missing_value = 'missing'
+        empty_value = 'empty'
 
         component_label = labels.get_name(Labels.LABEL_TYPE_COMPONENT)
-
         try:
             component = dockerfile_labels[component_label]
         except KeyError:
-            raise RuntimeError("missing label: {}".format(component_label))
+            self.log.error("%s label: %s", missing_value, component_label)
+            missing_labels[component_label] = missing_value
 
         version_label = labels.get_name(Labels.LABEL_TYPE_VERSION)
         try:
             version = dockerfile_labels[version_label]
+            if not version:
+                self.log.error('%s label: %s', empty_value, version_label)
+                missing_labels[version_label] = empty_value
         except KeyError:
-            raise RuntimeError('missing label: {}'.format(version_label))
+            self.log.error('%s label: %s', missing_value, version_label)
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -186,25 +186,39 @@ def run(self):
...
+            self.log.error('missing label: %s', version_label)
+            missing_labels.append(version_label)
 
+        release_label = labels.get_name(Labels.LABEL_TYPE_RELEASE)
         try:
             _, release = labels.get_name_and_value(Labels.LABEL_TYPE_RELEASE)
         except KeyError:
             release = None
 
+        if release is not None and not release:
```
athos-ribeiro: this could go in the try/except `else` clause instead. This way you could get `release_label` from the `get_name_and_value` call (no need for the `get_name` call above).

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -186,25 +186,39 @@ def run(self):
...
+            missing_labels.append(release_label)
+            self.log.error('label is empty: %s', release_label)
+
+        if missing_labels:
+            raise RuntimeError('Required labels are missing or empty or using'
+                               ' undefined variables: {}'.format(missing_labels))
```
athos-ribeiro: Could the error be more specific here (so users won't have to scroll up through the logs to actually know what the issue was)? For example, `missing_labels` could be a tuple with (label, reason). Whenever the reason is 'empty', we could let them know they should check for undefined env variables.

athos-ribeiro:
Nice!

Just added some nitpicks there

athos-ribeiro:
Do we need to update osbs-docs for the set of PRs open for this new feature?

Also, shoyuld we waive the "toggle feature"  configuration for them? I could not think of what it would mean to activate/deactivate this.

atomic_reactor/plugins/pre_bump_release.py
rcerven: yeah will do that

rcerven:
this isn't exactly feature, it is how it should be working :)

athos-ribeiro:
> this isn't exactly feature, it is how it should be working :)

I mean, do we let users know they can define some of the labels (and which) as env variables somewhere? This would be nice to have.

I am +1 on dropping the feature toggling (makes no sense here)

rcerven:
yes we will certainly let them know about that they can use variables for version & release

atomic_reactor/plugins/pre_bump_release.py
rcerven: printing there dict now with all problems

rcerven:
osbs-docs related PR: https://github.com/containerbuildsystem/osbs-docs/pull/92

rcerven:
rn updated: Additional check for release and version labels in bump_release

PR #1262, opened 2019-09-23
Remove code to mock Flatpak commands for tests

owtaylor:
Sufficiently new Flatpak is available and installed for the full
test matrix, so there is no reason to carry the mock-flatpak code
paths, which need quite a bit of maintenance as flatpak-module-tools
evolves.


rcerven:
not updating release note because it is only test change

PR #1266, opened 2019-10-01
inject parent image has to enclose original_base_image

rcerven:

### DIFF
--- a/atomic_reactor/build.py
+++ b/atomic_reactor/build.py
@@ -204,6 +204,15 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
                 custom_base_images.add(image_str)
                 self.custom_parent_image = True
             self.parents_ordered.append(image_str)
+
+            # we are setting values None, because we know parent but we don't
+            # have local copy yet
+            # if image is base image we want to keep image instance in key,
+            # so it is same as original_base_image instance
+            if self.original_base_image.to_str() == image_str:
+                self.parent_images[self.original_base_image] = None
+                continue
+
             self.parent_images[image_name] = None
 
         if len(custom_base_images) > 1:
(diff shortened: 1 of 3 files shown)


atomic_reactor/build.py
```diff
@@ -204,6 +204,11 @@ def set_df_path(self, path):
                 custom_base_images.add(image_str)
                 self.custom_parent_image = True
             self.parents_ordered.append(image_str)
+
+            if self.original_base_image.to_str() == image:
```
MartinBasti: Shouldn't we compare with `image_str` instead of `image`

Also this part of code is pretty obfuscated to me, there are no comments and I don't know why we are setting `None`

athos-ribeiro: Is `self.original_base_image` different from `image_name` in this case? If not, isn't this already being set to `None` below?

athos-ribeiro:
Waiving documentation requirements and feature activation through configuration. This is a bug fix.

atomic_reactor/build.py
twaugh: > Is `self.original_base_image` different from `image_name` in this case? If not, isn't this already being set to `None` below?

The problem is that they are two different *instances*, and so they have different keys.

twaugh: Oh, `image` is a string, I get it now.

twaugh: @MartinBasti so, `None` here means that there is a parent image we know about, and we have no local copy of it yet.

MartinBasti: /me would give a treasure for that comment in code

rcerven:
rn updated: fixed isolated builds, inject parent image plugin should enclose also original_base_image

PR #1269, opened 2019-10-14
Add pre-build fetch_sources plugin

athos-ribeiro:

### DIFF
--- a/atomic_reactor/plugins/pre_fetch_sources.py
+++ b/atomic_reactor/plugins/pre_fetch_sources.py
@@ -1,0 +1,224 @@
+"""
+Copyright (c) 2019 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+from __future__ import absolute_import
+
+import os
+import tempfile
+import time
+
+import requests
+from six.moves.urllib.parse import urlparse
+
+from atomic_reactor.constants import (
+    DEFAULT_DOWNLOAD_BLOCK_SIZE,
+    HTTP_BACKOFF_FACTOR,
+    HTTP_MAX_RETRIES,
+    PLUGIN_FETCH_SOURCES_KEY
+)
+from atomic_reactor.plugin import PreBuildPlugin
+from atomic_reactor.plugins.pre_reactor_config import (
+    get_koji,
+    get_koji_path_info,
+    get_koji_session,
+    get_config,
+    NO_FALLBACK
+)
+from atomic_reactor.util import get_retrying_requests_session
+
+
+class FetchSourcesPlugin(PreBuildPlugin):
+    """Download sources that may be used in further steps to compose Source Containers"""
+    key = PLUGIN_FETCH_SOURCES_KEY
+    is_allowed_to_fail = False
+    DOWNLOAD_DIR = 'image_sources'
+
+    def __init__(
+        self, tasker, workflow, koji_build_id=None, koji_build_nvr=None, signing_intent=None,
+    ):
+        """
+        :param tasker: ContainerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param koji_build_id: int, container image koji build id
+        :param koji_build_nvr: str, container image koji build NVR
+        :param signing_intent: str, ODCS signing intent name
+        """
+        if not koji_build_id and not koji_build_nvr:
+            err_msg = ('{} expects either koji_build_id or koji_build_nvr to be defined'
+                       .format(self.__class__.__name__))
+            raise TypeError(err_msg)
+        type_errors = []
+        if koji_build_id is not None and not isinstance(koji_build_id, int):
+            type_errors.append('koji_build_id must be an int. Got {}'.format(type(koji_build_id)))
+        if koji_build_nvr is not None and not isinstance(koji_build_nvr, str):
+            type_errors.append('koji_build_nvr must be a str. Got {}'.format(type(koji_build_nvr)))
+        if type_errors:
+            raise TypeError(type_errors)
+
+        super(FetchSourcesPlugin, self).__init__(tasker, workflow)
+        self.koji_build = None
+        self.koji_build_id = koji_build_id
+        self.koji_build_nvr = koji_build_nvr
+        self.signing_intent = signing_intent
+        self.session = get_koji_session(self.workflow, NO_FALLBACK)
+
+    def run(self):
+        """
+        :return: dict, binary image koji build id and nvr, and path to directory with
+        downloaded sources
+        """
+        self.set_koji_image_build_data()
+        signing_intent = self.get_signing_intent()
+        koji_config = get_koji(self.workflow, {})
+        insecure = koji_config.get('insecure_download', False)
+        urls = self.get_srpm_urls(signing_intent['keys'], insecure=insecure)
+        sources_dir = self.download_sources(urls, insecure=insecure)
+        return {
+                'sources_for_koji_build_id': self.koji_build_id,
+                'sources_for_nvr': self.koji_build_nvr,
+                'image_sources_dir': sources_dir,
+        }
+
+    def download_sources(self, urls, insecure=False):
+        """Download sources content
+
+        Download content in the given URLs into a new temporary directory and
+        return a list with each downloaded artifact's path.
+
+        :param urls: int, Koji build id of the container image we want SRPMs for
+        :param insecure: bool, whether to perform TLS checks of urls
+        :return: str, paths to directory with downloaded sources
+        """
+        workdir = tempfile.mkdtemp()
+        dest_dir = os.path.join(workdir, self.DOWNLOAD_DIR)
+        if not os.path.exists(dest_dir):
+            os.makedirs(dest_dir)
+
+        req_session = get_retrying_requests_session()
+        for url in urls:
+            parsed_url = urlparse(url)
+            dest_filename = os.path.basename(parsed_url.path)
+            dest_path = os.path.join(dest_dir, dest_filename)
+            self.log.debug('Downloading %s', url)
+
+            for attempt in range(HTTP_MAX_RETRIES + 1):
+                request = req_session.get(url, stream=True, verify=not insecure)
+                request.raise_for_status()
+                try:
+                    with open(dest_path, 'wb') as f:
+                        for chunk in request.iter_content(chunk_size=DEFAULT_DOWNLOAD_BLOCK_SIZE):
+                            f.write(chunk)
+                    break
+                except (requests.exceptions.RequestException):
+                    if attempt < HTTP_MAX_RETRIES:
+                        time.sleep(HTTP_BACKOFF_FACTOR * (2 ** attempt))
+                    else:
+                        raise
+
+            self.log.debug('Download finished: %s', dest_path)
+
+        return dest_dir
+
+    def set_koji_image_build_data(self):
+        build_identifier = self.koji_build_nvr or self.koji_build_id
+
+        # strict means this raises a koji.GenericError informing no matching build was found in
+        # case the build does not exist
+        self.koji_build = self.session.getBuild(build_identifier, strict=True)
+
+        if self.koji_build_id and (self.koji_build_id != self.koji_build['build_id']):
+            err_msg = (
+                'koji_build_id {} does not match koji_build_nvr {} with id {}. '
+                'When specifying both an id and an nvr, they should point to the same image build'
+                .format(self.koji_build_id, self.koji_build_nvr, self.koji_build['build_id'])
+                )
+            raise ValueError(err_msg)
+
+        if not self.koji_build_id:
+            self.koji_build_id = self.koji_build['build_id']
+        if not self.koji_build_nvr:
+            self.koji_build_nvr = self.koji_build['nvr']
+
+    def assemble_srpm_url(self, base_url, srpm_filename, sign_key=None):
+        """Assemble the URL used to fetch an SRPM file
+
+        :param base_url: str, Koji root base URL with the given build artifacts
+        :param srpm_filename: str, name of the SRPM file
+        :param sign_key: str, key used to sign the SRPM, as listed in the signing intent
+        :return: list, strings with URLs pointing to SRPM files
+        """
+        url_components = [base_url, 'src', srpm_filename]
+        if sign_key:
+            url_components[1:1] = ['data', 'signed', sign_key]
+        return '/'.join(url_components)
+
+    def get_srpm_urls(self, sigkeys=None, insecure=False):
+        """Fetch SRPM download URLs for each image generated by a build
+
+        Build each possible SRPM URL and check if the URL is available,
+        respecting the signing intent preference order.
+
+        :param sigkeys: list, strings for keys which signed the srpms to be fetched
+        :return: list, strings with URLs pointing to SRPM files
+        """
+        if not sigkeys:
+            sigkeys = ['']
+
+        archives = self.session.listArchives(self.koji_build_id, type='image')
+        rpm_build_ids = {rpm['id']: rpm['build_id'] for archive in archives
+                         for rpm in self.session.listRPMs(imageID=archive['id'])}
+
+        srpm_build_paths = {}
+        path_info = get_koji_path_info(self.workflow, NO_FALLBACK)
+        for rpm_id, rpm_build_id in rpm_build_ids.items():
+            rpm_hdr = self.session.getRPMHeaders(rpm_id, headers=['SOURCERPM'])
+            srpm_filename = rpm_hdr['SOURCERPM']
+            if srpm_filename in srpm_build_paths:
+                continue
+            rpm_build = self.session.getBuild(rpm_build_id, strict=True)
+            base_url = path_info.build(rpm_build)
+            srpm_build_paths[srpm_filename] = base_url
+
+        srpm_urls = []
+        missing_srpms = []
+        req_session = get_retrying_requests_session()
+        for srpm_filename, base_url in srpm_build_paths.items():
+            for sigkey in sigkeys:
+                # koji uses lowercase for paths. We make sure the sigkey is in lower case
+                url_candidate = self.assemble_srpm_url(base_url, srpm_filename, sigkey.lower())
+                request = req_session.head(url_candidate, verify=not insecure)
+                if request.ok:
+                    srpm_urls.append(url_candidate)
+                    break
+                self.log.debug('%s not found for signing key "%s" at %s (returned %s)',
+                               srpm_filename, sigkey, url_candidate, request.status_code)
+
+            else:
+                missing_srpms.append(srpm_filename)
+
+        if missing_srpms:
+            raise RuntimeError('Could not find files signed by any of {} for these SRPMS: {}'
+                               .format(sigkeys, srpm_filename))
+
+        return srpm_urls
+
+    def get_signing_intent(self):
+        """Get the signing intent to be used to fetch files from Koji
+
+        :return: dict, signing intent object as per atomic_reactor/schemas/config.json
+        """
+        odcs_config = get_config(self.workflow).get_odcs_config()
+        if not self.signing_intent:
+            try:
+                self.signing_intent = self.koji_build['extra']['image']['odcs']['signing_intent']
+            except (KeyError, TypeError):
+                self.log.debug('Image koji build, %s(%s), does not define signing_intent.',
+                               self.koji_build_nvr, self.koji_build_id)
+                self.signing_intent = odcs_config.default_signing_intent
+
+        signing_intent = odcs_config.get_signing_intent_by_name(self.signing_intent)
+        return signing_intent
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,150 @@
...
+        self.koji_build_nvr = koji_build_nvr
+        self.signing_intent = signing_intent
+        self.koji_fallback = {
+            'hub_url': kojihub,
+            'root_url': kojiroot,
+            'auth': {
+                'proxyuser': koji_proxy_user,
+                'ssl_certs_dir': koji_ssl_certs,
+                'krb_principal': str(koji_principal),
+                'krb_keytab_path': str(koji_keytab)
```
twaugh: Should `koji_principal` and `koji_keytab` default to the empty string instead of `None`?

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,150 @@
...
+        for archive in archives:
+            rpms = self.session.listRPMs(imageID=archive['id'])
+            for rpm in rpms:
+                rpm_hdr = self.session.getRPMHeaders(rpm['id'], headers=['SOURCERPM'])
+                rpm_build = self.session.getBuild(rpm['build_id'])
+                path_info = get_koji_path_info(self.workflow, self.koji_fallback)
+                srpm_url = '/'.join([path_info.build(rpm_build), 'src', rpm_hdr['SOURCERPM']])
+                # TODO: this should actually use a sigkey from the signing_intent, as:
+                # srpm_url = '/'.join([path_info.build(rpm_build), 'data', 'signed', sigkey, 'src', rpm_hdr['SOURCERPM']])
+                srpm_urls.append(srpm_url)
```
twaugh: Will we end up with duplicates when multiple architectures installed RPMs from the same SRPM?

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,208 @@
...
+class FetchSourcesPlugin(PreBuildPlugin):
+    """Download sources that may be used in further steps to compose Source Containers"""
+    key = PLUGIN_FETCH_SOURCES_KEY
+    is_allowed_to_fail = False
+    DOWNLOAD_DIR = 'image_sources'
+
+    def __init__(
+        self, tasker, workflow, koji_build_id=None, koji_build_nvr=None, signing_intent=None,
+        kojihub=None, kojiroot=None, koji_proxy_user=None, koji_ssl_certs=None, koji_principal='',
+        koji_keytab=''
```
twaugh: Maybe we should just take the koji information from the reactor config. Is there any benefit in allowing the plugin to override them?

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def get_signing_intent(self, build_id):
+        """Get the signing intent to be used to fetch files from Koji
+
+        :param build_id: int, Koji build id of the container image which owns the signing intent
+        :return: dict, signing intent object as per atomic_reactor/schemas/config.json
+        """
+        odcs_config = get_config(self.workflow).get_odcs_config()
+        if not self.signing_intent:
+            image_build = self.session.getBuild(build_id)
```
twaugh: Can we avoid this second call to `getBuild()` by caching the result of the earlier call (in `get_image_build_id()`)?

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,208 @@
...
+                    url_candidate = self.assemble_srpm_url(base_url, srpm_filename, sigkey)
+                    req_session = get_retrying_requests_session()
+                    request = req_session.head(url_candidate, verify=not insecure)
+                    if request.status_code == requests.codes.ok:
+                        srpm_url = url_candidate
+                        break
+                try:
+                    srpm_urls.add(srpm_url)
+                except NameError:
+                    raise RuntimeError('Could not satisfy signing intent for {}'.format(rpm['nvr']))
```
twaugh: This doesn't look quite right. Can `set.add` fail? I wouldn't expect to see `NameError` here unless there's a typo in the source code.

Maybe `except NameError:` should really be `else:`, as in for...else...?

I think we're missing something like 'if we already found source for this n-v-r, move on to the next one', since most of the RPMs in each archive will be the same but for different architectures.

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,208 @@
...
+                rpm_hdr = self.session.getRPMHeaders(rpm['id'], headers=['SOURCERPM'])
+                srpm_filename = rpm_hdr['SOURCERPM']
+                rpm_build = self.session.getBuild(rpm['build_id'])
+                path_info = get_koji_path_info(self.workflow, self.koji_fallback)
+                base_url = path_info.build(rpm_build)
+                for sigkey in sigkeys:
+                    url_candidate = self.assemble_srpm_url(base_url, srpm_filename, sigkey.lower())
+                    req_session = get_retrying_requests_session()
+                    request = req_session.head(url_candidate, verify=not insecure)
+                    if request.status_code == requests.codes.ok:
```
twaugh: `if request.ok:`

atomic_reactor/plugins/pre_fetch_sources.py
athos-ribeiro: I allowed overriding koji info to be consistent with other plugins getting a koji session. TBH I do not see much benefit on allowing it. @rcerven do you know of any any currently used use case for this in other plugins?

twaugh: I think they just do that for compatibility, since that's how it was done before the reactor config provided it.

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,211 @@
...
+            if not os.path.exists(dest_dir):
+                os.makedirs(dest_dir)
+
+            req_session = get_retrying_requests_session()
+            request = req_session.get(url, stream=True,
+                                      verify=not insecure)
+            request.raise_for_status()
+
+            with open(dest_path, 'wb') as f:
+                for chunk in request.iter_content(chunk_size=DEFAULT_DOWNLOAD_BLOCK_SIZE):
```
twaugh: If this streaming fails (e.g. Koji goes down briefly), should we retry the connection?

atomic_reactor/plugins/pre_fetch_sources.py
twaugh: Yes, relying on NameError here is no good: the next time round the rpms loop srpm_url will still be defined from the previous rpm.

MartinBasti: I agree this looks fragile, any future changes can break this assumption

tests/plugins/test_fetch_sources.py
```diff
@@ -0,0 +1,247 @@
...
+
+    config = {}
+    if data:
+        tmpdir.join('cert').write('')
+        config = util.read_yaml(data, 'schemas/config.json')
+
+    workflow.plugin_workspace[ReactorConfigPlugin.key][WORKSPACE_CONF_KEY] = ReactorConfig(config)
+
+
+def mock_dockerfile(tmpdir):
```
MartinBasti: I'm curious about reason why this plugin needs Dockerfile

athos-ribeiro: Thanks for spotting this. This remained after I removed the Dockerfile dependency at some point.

atomic_reactor/plugins/pre_fetch_sources.py
athos-ribeiro: I added a dict to store srpm names and base urls to avoid making extra requests to koji for duplicated sources (multiple arches). Not relying on name error anymore and we will collect all sources that could not be fetched before halting to report all missing files.

atomic_reactor/plugins/pre_fetch_sources.py
athos-ribeiro: The first call to getBuild is only necessary if a koji_nvr is provided. If a signing intent is provided, the second call is not necessary. I added a new `self.koji_build` to store the build in case the first call is needed.

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,226 @@
...
+    def get_signing_intent(self, build_id):
+        """Get the signing intent to be used to fetch files from Koji
+
+        :param build_id: int, Koji build id of the container image which owns the signing intent
+        :return: dict, signing intent object as per atomic_reactor/schemas/config.json
+        """
+        odcs_config = get_config(self.workflow).get_odcs_config()
+        if not self.signing_intent:
+            if not self.koji_build:
+                self.koji_build = self.session.getBuild(build_id, strict=True)
```
twaugh: Will this line ever get called?

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,226 @@
...
+                req_session = get_retrying_requests_session()
+                request = req_session.head(url_candidate, verify=not insecure)
+                if request.ok:
+                    srpm_url = url_candidate
+                    break
+                self.log.debug('%s not found for signing key "%s" at %s (returned %s)',
+                               srpm_filename, sigkey, url_candidate, request.status_code)
+
+            if not srpm_url:
+                missing_srpms.append(srpm_filename)
```
twaugh: Could replace
```
srpm_url = None
for sigkey in sigkeys:
    ...
if not srpm_url:
    ...
```

with:

```
for sigkey in sigkeys:
    ...
else:
    ...
```

...but I'm not sure if we like for...else...

atomic_reactor/plugins/pre_fetch_sources.py
athos-ribeiro: If an NVR for the image is not passed(let's say just a build id was provided), getBuild will not be called in `get_image_build_id`. Then if a signing intent from koji is needed, this will be called.

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,223 @@
...
+        self.session = get_koji_session(self.workflow, self.koji_fallback)
+
+    def run(self):
+        build_id = self.get_image_build_id()
+        signing_intent = self.get_signing_intent(build_id)
+        koji_config = get_koji(self.workflow, {})
+        insecure = koji_config.get('insecure_download', False)
+        urls = self.get_srpm_urls(build_id, signing_intent['keys'], insecure=insecure)
+        sources_list = self.download_sources(urls, insecure=insecure)
+        return sources_list
```
twaugh: Given how the build-step plugin wants to run, would it be nicer for this to return the output directory instead of all the files inside it? @MartinBasti wdyt?

atomic_reactor/plugins/pre_fetch_sources.py
twaugh: see e.g. #1280 which uses Koji but just gets config with no fallback.

atomic_reactor/plugins/pre_fetch_sources.py
twaugh: Also about the return value: we're going to want to include the original Koji build ID in the new Koji build metadata, so maybe we should return that as well.

MartinBasti: Oh yes, build_step plugin only needs path to directory with SRPMs

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: it was just for backwards compatibility during transition to reactor_config_map

for new plugins we should use just reactor_config_map

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,223 @@
...
+            nvr_build_id = self.koji_build['build_id']
+
+        if self.koji_build_id and nvr_build_id and (self.koji_build_id != nvr_build_id):
+            err_msg = (
+                'koji_build_id {} does not match koji_build_nvr {} with id {}. '
+                'When specifying both an id and an nvr, they should point to the same image build'
+                .format(self.koji_build_id, self.koji_build_nvr, nvr_build_id)
+                )
+            raise ValueError(err_msg)
+        return self.koji_build_id or nvr_build_id
```
rcerven: if self.koji_build_id or nvr_build_id:
   return self.koji_build_id or nvr_build_id
else:
  raise ValueError("either koji_build_id or koji_build_nvr has to be provided")

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,223 @@
...
+
+        return srpm_urls
+
+    def get_signing_intent(self, build_id):
+        """Get the signing intent to be used to fetch files from Koji
+
+        :param build_id: int, Koji build id of the container image which owns the signing intent
+        :return: dict, signing intent object as per atomic_reactor/schemas/config.json
+        """
+        odcs_config = get_config(self.workflow).get_odcs_config()
```
rcerven: use get_odcs_config directly like we do in plugins?

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,223 @@
...
+        return downloaded_sources
+
+    def get_image_build_id(self):
+        nvr_build_id = None
+        if self.koji_build_nvr:
+            # strict means this raises a koji.GenericError informing no matching build was found in
+            # case the build does not exist
+            self.koji_build = self.session.getBuild(self.koji_build_nvr, strict=True)
+            nvr_build_id = self.koji_build['build_id']
+
```
rcerven: if self.koji_build_id
   check for build and getBuild

and this method can actually return tuple of (build_id, getBuild_result)
that way you don't have to getBuild again in other methods

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,223 @@
...
+        archives = self.session.listArchives(build_id, type='image')
+        for archive in archives:
+            rpms = self.session.listRPMs(imageID=archive['id'])
+            for rpm in rpms:
+                rpm_hdr = self.session.getRPMHeaders(rpm['id'], headers=['SOURCERPM'])
+                srpm_filename = rpm_hdr['SOURCERPM']
+                if srpm_filename in srpm_build_paths:
+                    continue
+                rpm_build = self.session.getBuild(rpm['build_id'], strict=True)
+                path_info = get_koji_path_info(self.workflow, self.koji_fallback)
```
rcerven: move it outside of loop

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,223 @@
...
+        for url in urls:
+            parsed_url = urlparse(url)
+            dest_filename = os.path.basename(parsed_url.path)
+            dest_path = os.path.join(dest_dir, dest_filename)
+            self.log.debug('Downloading %s', url)
+
+            if not os.path.exists(dest_dir):
+                os.makedirs(dest_dir)
+
+            req_session = get_retrying_requests_session()
```
rcerven: move it outside of loop

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,223 @@
...
+                base_url = path_info.build(rpm_build)
+                srpm_build_paths[srpm_filename] = base_url
+
+        srpm_urls = []
+        missing_srpms = []
+        for srpm_filename, base_url in srpm_build_paths.items():
+            for sigkey in sigkeys:
+                # koji uses lowercase for paths. We make sure the sigkey is in lower case
+                url_candidate = self.assemble_srpm_url(base_url, srpm_filename, sigkey.lower())
+                req_session = get_retrying_requests_session()
```
rcerven: move it outside of loop

atomic_reactor/plugins/pre_fetch_sources.py
athos-ribeiro: I am doing it like this to get an ODCSConfig object, like in the resolve_composes plugin. If you prefer, I could get the odcs config and import the ODCSConfig class as well, then create the object passing the info from the config. WDYT?

atomic_reactor/plugins/pre_fetch_sources.py
athos-ribeiro: Would the else case ever be executed though? There is a check for that in the initialization. In case the build_id is not provided and the nvr given is invalid, getBuild will fail.

atomic_reactor/plugins/pre_fetch_sources.py
athos-ribeiro: removed koji config overriding capabilities

atomic_reactor/plugins/pre_fetch_sources.py
twaugh: Not right now. Without a scheme defined for what that should look like let's just stick to having everything at the top-level in this first iteration.

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,210 @@
...
+        downloaded sources
+        """
+        build_id = self.get_image_build_id()
+        signing_intent = self.get_signing_intent(build_id)
+        koji_config = get_koji(self.workflow, {})
+        insecure = koji_config.get('insecure_download', False)
+        urls = self.get_srpm_urls(build_id, signing_intent['keys'], insecure=insecure)
+        sources_dir = self.download_sources(urls, insecure=insecure)
+        return {
+                'bin_image_koji_build_id': build_id,
```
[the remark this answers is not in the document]
athos-ribeiro: Done, returning both id and nvr. It also solved the other comment on performing getBuild in multiple places.

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: yeah I forgot that get_odcs_config isn't the public method

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,227 @@
...
+
+            self.log.debug('Download finished: %s', dest_path)
+
+        return dest_dir
+
+    def set_koji_image_build_data(self):
+        if self.koji_build_nvr:
+            build_identifier = self.koji_build_nvr
+        else:
+            build_identifier = self.koji_build_id
```
twaugh: Minor: could write as `build_identifier = self.koji_build_nvr or self.koji_build_id`

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,227 @@
...
+        self.set_koji_image_build_data()
+        signing_intent = self.get_signing_intent()
+        koji_config = get_koji(self.workflow, {})
+        insecure = koji_config.get('insecure_download', False)
+        urls = self.get_srpm_urls(signing_intent['keys'], insecure=insecure)
+        sources_dir = self.download_sources(urls, insecure=insecure)
+        return {
+                'sources_for_koji_build_id': self.koji_build_id,
+                'sources_for_nvr': self.koji_build_nvr,
+                'image_sources_dir': sources_dir
```
twaugh: Minor: could add a trailing comma here for easier-to-read diffs in future

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,227 @@
...
+               }
```
twaugh: Minor: wouldn't this want to line up with `return`?

tests/plugins/test_fetch_sources.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+class MockSource(StubSource):
+
+    def __init__(self, workdir):
+        super(MockSource, self).__init__()
+        self.workdir = workdir
+
+    def get_build_file_path(self):
+        return os.path.join(self.workdir, 'Dockerfile'), self.workdir
```
twaugh: Is this used?

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -0,0 +1,227 @@
...
+        self, tasker, workflow, koji_build_id=None, koji_build_nvr=None, signing_intent=None,
+    ):
+        """
+        :param tasker: ContainerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param koji_build_id: int, container image koji build id
+        :param koji_build_nvr: str, container image koji build NVR
+        :param signing_intent: str, ODCS signing intent name
+        """
+        if not koji_build_id and not koji_build_nvr:
```
rcerven: and what if both are provided?
in set_koji_image_build_data
self.koji_build_nvr is favored 

but I think we should raise error if both are provided

athos-ribeiro: That's what we are doing in `set_koji_image_build_data`: if both are provided, they should refer to the same build (we are checking this through the `build_id`).

rcerven: right I see it now

athos-ribeiro:
Thanks for the reviews!

PR #1282, opened 2019-10-24
add floating tags push plugin

mlangsdorf:
Add a plugin that runs after koji_import and adds floating tags to the existing manifest lists
produced by group_manifests.

- (n/a) New feature can be disabled from a configuration file  -- would be too invasive

### DIFF
--- a/atomic_reactor/constants.py
+++ b/atomic_reactor/constants.py
@@ -106,6 +106,7 @@ PLUGIN_PUSH_OPERATOR_MANIFESTS_KEY = 'push_operator_manifests'
 PLUGIN_SOURCE_CONTAINER_KEY = 'source_container'
 PLUGIN_FETCH_SOURCES_KEY = 'fetch_sources'
 PLUGIN_KOJI_DELEGATE_KEY = 'koji_delegate'
+PLUGIN_PUSH_FLOATING_TAGS_KEY = 'push_floating_tags'
 
 # some shared dict keys for build metadata that gets recorded with koji.
 # for consistency of metadata in historical builds, these values basically cannot change.
--- a/atomic_reactor/manifest_util.py
+++ b/atomic_reactor/manifest_util.py
@@ -1,0 +1,232 @@
+"""
+Copyright (c) 2019 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import unicode_literals, absolute_import
+import json
+import requests
+from copy import deepcopy
+
+from atomic_reactor.plugin import PluginFailedException
+from atomic_reactor.plugins.pre_reactor_config import get_registries
+from atomic_reactor.util import (registry_hostname, RegistrySession, ManifestDigest,
+                                 get_manifest_media_type)
+from atomic_reactor.constants import (MEDIA_TYPE_DOCKER_V2_SCHEMA2,
+                                      MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST, MEDIA_TYPE_OCI_V1,
+                                      MEDIA_TYPE_OCI_V1_INDEX)
+
+
+class ManifestUtil(object):
+    manifest_media_types = (
+        MEDIA_TYPE_DOCKER_V2_SCHEMA2,
+        MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST,
+        MEDIA_TYPE_OCI_V1,
+        MEDIA_TYPE_OCI_V1_INDEX
+    )
+
+    def __init__(self, workflow, registries, log):
+        self.push_conf = workflow.push_conf
+        self.registries = get_registries(workflow, deepcopy(registries or {}))
+        self.worker_registries = {}
+        self.log = log
+
+    def valid_media_type(self, media_type):
+        return media_type in self.manifest_media_types
+
+    def sort_annotations(self, all_annotations):
+        sorted_digests = {}
+        all_platforms = set(all_annotations)
+
+        for plat, annotation in all_annotations.items():
+            for digest in annotation['digests']:
+                hostname = registry_hostname(digest['registry'])
+                media_type = get_manifest_media_type(digest['version'])
+                if not self.valid_media_type(media_type):
+                    continue
+
+                platforms = sorted_digests.setdefault(hostname, {})
+                repos = platforms.setdefault(plat, [])
+                repos.append(digest)
+
+        sources = {}
+        for registry in self.registries:
+            hostname = registry_hostname(registry)
+            platforms = sorted_digests.get(hostname, {})
+
+            if set(platforms) != all_platforms:
+                raise RuntimeError("Missing platforms for registry {}: found {}, expected {}"
+                                   .format(registry, sorted(platforms), sorted(all_platforms)))
+
+            selected_digests = {}
+            for p, repos in platforms.items():
+                selected_digests[p] = sorted(repos, key=lambda d: d['repository'])[0]
+
+            sources[registry] = selected_digests
+
+        return sources
+
+    def get_manifest(self, session, repository, ref):
+        """
+        Downloads a manifest from a registry. ref can be a digest, or a tag.
+        """
+        self.log.debug("%s: Retrieving manifest for %s:%s", session.registry, repository, ref)
+
+        headers = {
+            'Accept': ', '.join(self.manifest_media_types)
+        }
+
+        url = '/v2/{}/manifests/{}'.format(repository, ref)
+        response = session.get(url, headers=headers)
+        response.raise_for_status()
+        return (response.content,
+                response.headers['Docker-Content-Digest'],
+                response.headers['Content-Type'],
+                int(response.headers['Content-Length']))
+
+    def link_blob_into_repository(self, session, digest, source_repo, target_repo):
+        """
+        Links ("mounts" in Docker Registry terminology) a blob from one repository in a
+        registry into another repository in the same registry.
+        """
+        self.log.debug("%s: Linking blob %s from %s to %s",
+                       session.registry, digest, source_repo, target_repo)
+
+        # Check that it exists in the source repository
+        url = "/v2/{}/blobs/{}".format(source_repo, digest)
+        result = session.head(url)
+        if result.status_code == requests.codes.NOT_FOUND:
+            self.log.debug("%s: blob %s, not present in %s, skipping",
+                           session.registry, digest, source_repo)
+            # Assume we don't need to copy it - maybe it's a foreign layer
+            return
+        result.raise_for_status()
+
+        url = "/v2/{}/blobs/uploads/?mount={}&from={}".format(target_repo, digest, source_repo)
+        result = session.post(url, data='')
+        result.raise_for_status()
+
+        if result.status_code != requests.codes.CREATED:
+            # A 202-Accepted would mean that the source blob didn't exist and
+            # we're starting an upload - but we've checked that above
+            raise RuntimeError("Blob mount had unexpected status {}".format(result.status_code))
+
+    def link_manifest_references_into_repository(self, session, manifest, media_type,
+                                                 source_repo, target_repo):
+        """
+        Links all the blobs referenced by the manifest from source_repo into target_repo.
+        """
+
+        if source_repo == target_repo:
+            return
+
+        parsed = json.loads(manifest.decode('utf-8'))
+
+        references = []
+        if media_type in (MEDIA_TYPE_DOCKER_V2_SCHEMA2, MEDIA_TYPE_OCI_V1):
+            references.append(parsed['config']['digest'])
+            for l in parsed['layers']:
+                references.append(l['digest'])
+        else:
+            # manifest list support could be added here, but isn't needed currently, since
+            # we never copy a manifest list as a whole between repositories
+            raise RuntimeError("Unhandled media-type {}".format(media_type))
+
+        for digest in references:
+            self.link_blob_into_repository(session, digest, source_repo, target_repo)
+
+    def store_manifest_in_repository(self, session, manifest, media_type,
+                                     source_repo, target_repo, ref=None):
+        """
+        Stores the manifest into target_repo, possibly tagging it. This may involve
+        copying referenced blobs from source_repo.
+        """
+
+        if not ref:
+            raise RuntimeError("Either a digest or tag must be specified as ref")
+
+        self.link_manifest_references_into_repository(session, manifest, media_type,
+                                                      source_repo, target_repo)
+
+        url = '/v2/{}/manifests/{}'.format(target_repo, ref)
+        headers = {'Content-Type': media_type}
+        response = session.put(url, data=manifest, headers=headers)
+        response.raise_for_status()
+
+    def get_registry_session(self, registry):
+        registry_conf = self.registries[registry]
+
+        insecure = registry_conf.get('insecure', False)
+        secret_path = registry_conf.get('secret')
+
+        return RegistrySession(registry, insecure=insecure,
+                               dockercfg_path=secret_path,
+                               access=('pull', 'push'))
+
+    def add_tag_and_manifest(self, session, image_manifest, media_type, manifest_digest,
+                             source_repo, configured_tags):
+        push_conf_registry = self.push_conf.add_docker_registry(session.registry,
+                                                                insecure=session.insecure)
+        for image in configured_tags:
+            target_repo = image.to_str(registry=False, tag=False)
+            self.store_manifest_in_repository(session, image_manifest, media_type,
+                                              source_repo, target_repo, ref=image.tag)
+
+            # add a tag for any plugins running later that expect it
+            push_conf_registry.digests[image.tag] = manifest_digest
+
+    def tag_manifest_into_registry(self, session, digest, source_repo, configured_tags):
+        """
+        Tags the manifest identified by worker_digest into session.registry with all the
+        configured_tags
+        """
+        self.log.info("%s: Tagging manifest", session.registry)
+
+        image_manifest, _, media_type, _ = self.get_manifest(session, source_repo, digest)
+        if media_type == MEDIA_TYPE_DOCKER_V2_SCHEMA2:
+            digests = ManifestDigest(v1=digest)
+        elif media_type == MEDIA_TYPE_OCI_V1:
+            digests = ManifestDigest(oci=digest)
+        else:
+            raise RuntimeError("Unexpected media type {} found in source_repo: {}"
+                               .format(media_type, source_repo))
+
+        self.add_tag_and_manifest(session, image_manifest, media_type, digests, source_repo,
+                                  configured_tags)
+
+    def build_list(self, manifests):
+        """
+        Builds a manifest list or OCI image out of the given manifests
+        """
+
+        media_type = manifests[0]['media_type']
+        if (not all(m['media_type'] == media_type for m in manifests)):
+            raise PluginFailedException('worker manifests have inconsistent types: {}'
+                                        .format(manifests))
+
+        if media_type == MEDIA_TYPE_DOCKER_V2_SCHEMA2:
+            list_type = MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST
+        elif media_type == MEDIA_TYPE_OCI_V1:
+            list_type = MEDIA_TYPE_OCI_V1_INDEX
+        else:
+            raise PluginFailedException('worker manifests have unsupported type: {}'
+                                        .format(media_type))
+
+        return list_type, json.dumps({
+                "schemaVersion": 2,
+                "mediaType": list_type,
+                "manifests": sorted([
+                    {
+                        "mediaType": media_type,
+                        "size": m['size'],
+                        "digest": m['digest'],
+                        "platform": {
+                            "architecture": m['architecture'],
+                            "os": "linux"
+                        }
+                    } for m in manifests
+                ], key=lambda entry: entry['platform']['architecture']),
+        }, indent=4, sort_keys=True, separators=(',', ': '))
--- a/atomic_reactor/manifest_util.py
+++ b/atomic_reactor/manifest_util.py
@@ -196,6 +196,7 @@ class ManifestUtil(object):
 
         self.add_tag_and_manifest(session, image_manifest, media_type, digests, source_repo,
                                   configured_tags)
+        return image_manifest, media_type, digests
 
     def build_list(self, manifests):
         """
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -21,7 +21,8 @@ from atomic_reactor.plugins.build_orchestrate_build import (get_worker_build_inf
                                                             get_koji_upload_dir)
 from atomic_reactor.plugins.pre_add_filesystem import AddFilesystemPlugin
 from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
-from atomic_reactor.util import OSBSLogs, get_parent_image_koji_data
+from atomic_reactor.util import (OSBSLogs, get_parent_image_koji_data, get_manifest_media_version,
+                                 ManifestDigest)
 from atomic_reactor.plugins.pre_reactor_config import get_openshift_session
 
 try:
@@ -233,8 +234,9 @@ class KojiImportPlugin(ExitPlugin):
         floating_tags = [image.tag for image in floating_images]
         unique_tags = [image.tag for image in unique_images]
 
-        manifest_list_digests = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY)
-        if manifest_list_digests:
+        manifest_list_data = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY, {})
+        manifest_digest = manifest_list_data.get("manifest_digest")
+        if manifest_digest and isinstance(manifest_digest, ManifestDigest):
             index = {}
             index['tags'] = tags
             index['floating_tags'] = floating_tags
@@ -243,9 +245,12 @@ class KojiImportPlugin(ExitPlugin):
             repo = ImageName.parse(repositories[0]).to_str(registry=False, tag=False)
             # group_manifests added the registry, so this should be valid
             registries = self.workflow.push_conf.all_registries
+
+            digest_version = get_manifest_media_version(manifest_digest)
+            digest = manifest_digest.default
+
             for registry in registries:
-                manifest_list_digest = manifest_list_digests[repo]
-                pullspec = "{0}/{1}@{2}".format(registry.uri, repo, manifest_list_digest.default)
+                pullspec = "{0}/{1}@{2}".format(registry.uri, repo, digest)
                 index['pull'] = [pullspec]
                 pullspec = "{0}/{1}:{2}".format(registry.uri, repo,
                                                 version_release)
@@ -253,10 +258,9 @@ class KojiImportPlugin(ExitPlugin):
 
                 # Store each digest with according media type
                 index['digests'] = {}
-                for version, digest in manifest_list_digest.items():
-                    if digest:
-                        media_type = get_manifest_media_type(version)
-                        index['digests'][media_type] = digest
+                media_type = get_manifest_media_type(digest_version)
+                index['digests'][media_type] = digest
+
                 break
             extra['image']['index'] = index
         # group_manifests returns None if didn't run, {} if group=False
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -22,7 +22,7 @@ from atomic_reactor.plugins.build_orchestrate_build import (get_worker_build_inf
 from atomic_reactor.plugins.pre_add_filesystem import AddFilesystemPlugin
 from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
 from atomic_reactor.util import (OSBSLogs, get_parent_image_koji_data, get_manifest_media_version,
-                                 ManifestDigest)
+                                 is_manifest_list)
 from atomic_reactor.plugins.pre_reactor_config import get_openshift_session
 
 try:
@@ -234,9 +234,9 @@ class KojiImportPlugin(ExitPlugin):
         floating_tags = [image.tag for image in floating_images]
         unique_tags = [image.tag for image in unique_images]
 
-        manifest_list_data = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY, {})
-        manifest_digest = manifest_list_data.get("manifest_digest")
-        if manifest_digest and isinstance(manifest_digest, ManifestDigest):
+        manifest_data = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY, {})
+        if manifest_data and is_manifest_list(manifest_data.get("media_type")):
+            manifest_digest = manifest_data.get("manifest_digest")
             index = {}
             index['tags'] = tags
             index['floating_tags'] = floating_tags
--- a/atomic_reactor/plugins/exit_push_floating_tags.py
+++ b/atomic_reactor/plugins/exit_push_floating_tags.py
@@ -1,0 +1,94 @@
+"""
+Copyright (c) 2019 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import unicode_literals, absolute_import
+
+import json
+
+from atomic_reactor.constants import PLUGIN_PUSH_FLOATING_TAGS_KEY, PLUGIN_GROUP_MANIFESTS_KEY
+from atomic_reactor.manifest_util import ManifestUtil
+from atomic_reactor.plugin import ExitPlugin
+from atomic_reactor.util import get_floating_images, get_unique_images
+
+
+class PushFloatingTagsPlugin(ExitPlugin):
+    """
+    Push floating tags to registry
+    """
+
+    key = PLUGIN_PUSH_FLOATING_TAGS_KEY
+    is_allowed_to_fail = False
+
+    def __init__(self, tasker, workflow, registries=None):
+        """
+        constructor
+
+        :param tasker: DockerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param registries: dict, keys are docker registries, values are dicts containing
+                           per-registry parameters.
+                           Params:
+                            * "secret" optional string - path to the secret, which stores
+                              login and password for remote registry
+        """
+        super(PushFloatingTagsPlugin, self).__init__(tasker, workflow)
+        self.manifest_util = ManifestUtil(workflow, registries, self.log)
+
+    def add_floating_tags(self, session, manifest_list_data, floating_images):
+        manifest_json = manifest_list_data.get("manifest_list", {})
+        manifest_digest = manifest_list_data.get("manifest_digest", {})
+        list_type = json.loads(manifest_json)["mediaType"]
+
+        for image in floating_images:
+            target_repo = image.to_str(registry=False, tag=False)
+            # We have to call store_manifest_in_repository directly for each
+            # referenced manifest, since each one should be a new tag that requires uploading
+            # the manifest again
+            self.log.debug("storing %s as %s", target_repo, image.tag)
+            self.manifest_util.store_manifest_in_repository(session, manifest_json, list_type,
+                                                            target_repo, target_repo, ref=image.tag)
+        # And store the manifest list in the push_conf
+        push_conf_registry = self.workflow.push_conf.add_docker_registry(session.registry,
+                                                                         insecure=session.insecure)
+        for image in floating_images:
+            push_conf_registry.digests[image.tag] = manifest_digest
+        registry_image = get_unique_images(self.workflow)[0]
+
+        return registry_image.get_repo(explicit_namespace=False), manifest_digest
+
+    def run(self):
+        """
+        Run the plugin.
+        """
+        if self.workflow.build_process_failed:
+            self.log.info('Build failed, skipping %s', PLUGIN_PUSH_FLOATING_TAGS_KEY)
+            return
+
+        floating_tags = get_floating_images(self.workflow)
+        if not floating_tags:
+            self.log.info('No floating images to tag, skippping %s', PLUGIN_PUSH_FLOATING_TAGS_KEY)
+            return
+
+        #  can't run in the worker build
+        if not self.workflow.is_orchestrator_build():
+            self.log.warning('%s cannot be used by a worker builder', PLUGIN_PUSH_FLOATING_TAGS_KEY)
+            return
+
+        manifest_data = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY)
+        if not manifest_data or not manifest_data.get("manifest_digest"):
+            self.log.info('No manifest digest available, skipping %s',
+                          PLUGIN_PUSH_FLOATING_TAGS_KEY)
+            return
+
+        digests = dict()
+
+        for registry in self.manifest_util.registries:
+            session = self.manifest_util.get_registry_session(registry)
+            repo, digest = self.add_floating_tags(session, manifest_list_data, floating_tags)
+            digests[repo] = digest
+        return digests
--- a/atomic_reactor/plugins/exit_push_floating_tags.py
+++ b/atomic_reactor/plugins/exit_push_floating_tags.py
@@ -8,8 +8,6 @@ of the BSD license. See the LICENSE file for details.
 
 from __future__ import unicode_literals, absolute_import
 
-import json
-
 from atomic_reactor.constants import PLUGIN_PUSH_FLOATING_TAGS_KEY, PLUGIN_GROUP_MANIFESTS_KEY
 from atomic_reactor.manifest_util import ManifestUtil
 from atomic_reactor.plugin import ExitPlugin
@@ -40,9 +38,9 @@ class PushFloatingTagsPlugin(ExitPlugin):
         self.manifest_util = ManifestUtil(workflow, registries, self.log)
 
     def add_floating_tags(self, session, manifest_list_data, floating_images):
-        manifest_json = manifest_list_data.get("manifest_list", {})
-        manifest_digest = manifest_list_data.get("manifest_digest", {})
-        list_type = json.loads(manifest_json)["mediaType"]
+        list_type = manifest_list_data.get("media_type")
+        manifest = manifest_list_data.get("manifest")
+        manifest_digest = manifest_list_data.get("manifest_digest")
 
         for image in floating_images:
             target_repo = image.to_str(registry=False, tag=False)
@@ -50,7 +48,7 @@ class PushFloatingTagsPlugin(ExitPlugin):
             # referenced manifest, since each one should be a new tag that requires uploading
             # the manifest again
             self.log.debug("storing %s as %s", target_repo, image.tag)
-            self.manifest_util.store_manifest_in_repository(session, manifest_json, list_type,
+            self.manifest_util.store_manifest_in_repository(session, manifest, list_type,
                                                             target_repo, target_repo, ref=image.tag)
         # And store the manifest list in the push_conf
         push_conf_registry = self.workflow.push_conf.add_docker_registry(session.registry,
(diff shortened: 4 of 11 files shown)


rcerven:
in  post_group_manifests
104: for image in self.workflow.tag_conf.images:

we don't want to cycle through all images but only through

util.get_primary_images(self.workflow)  + util.get_unique_images(self.workflow)

atomic_reactor/plugins/exit_floating_tags.py
```diff
@@ -0,0 +1,87 @@
...
+            self.log.info('No manifest list digest available, skipping floating tags push')
+            return
+
+        # if this isn't present, we're running in the worker build and there's an error
+        worker_builds = self.workflow.build_result.annotations['worker-builds']
+        if not worker_builds:
+            self.log.warning('%s cannot be used by a worker builder', PLUGIN_FLOATING_TAGS_KEY)
+            return
+
+        floating_images = self.workflow.tag_conf.floating_images
```
rcerven: use:
util.get_floating_images(self.workflow)

rcerven:
in exit_floating_tags

instead of recreating manifest list

in post_group_manifests return tuple (digests, list_json) 
list_json is created from manifest_util.build_list

and use that in floating_tags plugin

atomic_reactor/manifest_util.py
```diff
@@ -0,0 +1,205 @@
...
+from atomic_reactor.plugin import PluginFailedException
+from atomic_reactor.plugins.pre_reactor_config import get_registries
+from atomic_reactor.util import RegistrySession, ManifestDigest
+from atomic_reactor.constants import (MEDIA_TYPE_DOCKER_V2_SCHEMA2,
+                                      MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST, MEDIA_TYPE_OCI_V1,
+                                      MEDIA_TYPE_OCI_V1_INDEX)
+
+
+class ManifestUtil(object):
+    manifest_media_types = [
```
twaugh: Better to make this class attribute a tuple so that it's immutable.

I realised this is pre-existing code, but since we're moving it around...

atomic_reactor/plugins/exit_floating_tags.py
```diff
@@ -0,0 +1,87 @@
...
+            return
+
+        manifest_list_digests = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY)
+        if not manifest_list_digests:
+            self.log.info('No manifest list digest available, skipping floating tags push')
+            return
+
+        # if this isn't present, we're running in the worker build and there's an error
+        worker_builds = self.workflow.build_result.annotations['worker-builds']
+        if not worker_builds:
```
twaugh: The new way of doing this is:
```
if not self.workflow.is_orchestrator_build():
```

mlangsdorf:
Revised with rcerven's suggestions.  I still need to go back and fix the failing unit tests for other plugins now that group_manifest's return value has changed.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -270,40 +135,11 @@ def group_manifests_and_tag(self, session, worker_digests):
...
-        push_conf_registry = self.workflow.push_conf.add_docker_registry(session.registry,
-                                                                         insecure=session.insecure)
-        for image in self.workflow.tag_conf.images:
-            target_repo = image.to_str(registry=False, tag=False)
-            self.store_manifest_in_repository(session, image_manifest, media_type,
-                                              source_repo, target_repo, tag=image.tag)
-
-            # add a tag for any plugins running later that expect it
-            push_conf_registry.digests[image.tag] = digests
+        return registry_image.get_repo(explicit_namespace=False), digest, json.loads(list_json)
```
twaugh: We definitely want to return the *bytes* of the manifest list, not the JSON representation, i.e. `list_json` not `json.loads(list_json)`. Once we've stored it in the registry once, the representation in bytes has been fixed. Any change from that and the digest will be wrong.

atomic_reactor/plugins/exit_floating_tags.py
```diff
@@ -0,0 +1,98 @@
...
+        self.manifest_util = ManifestUtil(workflow, registries, self.log)
+
+    def make_manifest(self, registry, repo, digest):
+        return "{registry}/{repo}@{digest}".format(**vars())
+
+    def add_floating_tags(self, session, manifest_list_data, floating_images):
+        manifest_list = manifest_list_data.get("list_json", {})
+        manifest_digests = manifest_list_data.get("digests", {})
+        list_type = manifest_list["mediaType"]
+        manifest_json = json.dumps(manifest_list, indent=4)
```
twaugh: We don't want to be re-creating the bytes here, we need them to be exactly as they were when we originally stored it.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -318,62 +154,27 @@ def sort_annotations(self):
...
                     raise RuntimeError('Without grouping only one source is expected')
                 for digest in source.values():
-                    self.tag_manifest_into_registry(session, digest)
+                    self.manifest_util.tag_manifest_into_registry(session, digest,
+                                                                  self.workflow.tag_conf.images)
                     found = True
                 if not found:
                     raise ValueError('Failed to find any platform')
-        return digests
+        return {"digests": digests, "list_json": list_json}
```
twaugh: The digest will be the same for every repository. Maybe we should really be returning "digest" and "repos" instead of "digests".

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
mlangsdorf: Fixed.

atomic_reactor/manifest_util.py
```diff
@@ -0,0 +1,243 @@
...
+            ref = tag
+        elif digest:
+            self.log.debug("%s: Storing manifest (or list) %s from %s in %s",
+                           session.registry, digest, source_repo, target_repo)
+            ref = digest
+        else:
+            raise RuntimeError("Either digest or tag must be specified")
+
+        self.link_manifest_references_into_repository(session, manifest, media_type,
+                                                      source_repo, target_repo)
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

atomic_reactor/constants.py
```diff
@@ -105,7 +105,7 @@
 PLUGIN_PUSH_OPERATOR_MANIFESTS_KEY = 'push_operator_manifests'
 PLUGIN_SOURCE_CONTAINER_KEY = 'source_container'
 PLUGIN_FETCH_SOURCES_KEY = 'fetch_sources'
-
+PLUGIN_FLOATING_TAGS_KEY = 'floating_tags_push'
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

atomic_reactor/plugins/exit_floating_tags.py
```diff
@@ -0,0 +1,95 @@
...
+    def add_floating_tags(self, session, manifest_list_data, floating_images):
+        manifest_json = manifest_list_data.get("repos", {})
+        manifest_digest = manifest_list_data.get("digest", {})
+        manifest_lists = json.loads(manifest_json)
+        list_type = manifest_lists["mediaType"]
+
+        for image in floating_images:
+            target_repo = image.to_str(registry=False, tag=False)
+            # We have to call store_manifest_in_repository directly for each
+            # referenced manifest, since they potentially come from different repos
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

atomic_reactor/plugins/exit_floating_tags.py
```diff
@@ -0,0 +1,95 @@
...
+            self.log.debug("storing {} as {}".format(target_repo, image.tag))
+            self.manifest_util.store_manifest_in_repository(session, manifest_json, list_type,
+                                                            target_repo, target_repo, tag=image.tag)
+        for digest, value in manifest_digest.items():
+            break
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -318,62 +152,27 @@ def sort_annotations(self):
...
                     raise RuntimeError('Without grouping only one source is expected')
                 for digest in source.values():
-                    self.tag_manifest_into_registry(session, digest)
+                    self.manifest_util.tag_manifest_into_registry(session, digest,
+                                                                  self.workflow.tag_conf.images)
                     found = True
                 if not found:
                     raise ValueError('Failed to find any platform')
-        return digests
+        return {"digest": digests, "repos": list_json}
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -318,62 +152,32 @@ def sort_annotations(self):
...
+                    source_digest = digest['digest']
+                    source_repo = digest['repository']
+                    self.manifest_util.tag_manifest_into_registry(session, source_digest,
+                                                                  source_repo,
+                                                                  self.workflow.tag_conf.images)
                     found = True
                 if not found:
                     raise ValueError('Failed to find any platform')
-        return digests
+        return {'digest': digest, 'repos': repos, 'manifest_list': list_json}
```
twaugh: Sorry, just spotted that `digest` here is a dict, I'd thought it was a string (it gets reused with different types). Can we arrange for `digest` to always be a string, so that the 'digest' in our return dict is always a string?

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -318,62 +152,32 @@ def sort_annotations(self):
...
                 found = False
                 if len(source) != 1:
                     raise RuntimeError('Without grouping only one source is expected')
                 for digest in source.values():
-                    self.tag_manifest_into_registry(session, digest)
+                    source_digest = digest['digest']
+                    source_repo = digest['repository']
+                    self.manifest_util.tag_manifest_into_registry(session, source_digest,
+                                                                  source_repo,
+                                                                  self.workflow.tag_conf.images)
```
rcerven: should be also just all non-floating tags :
util.get_primary_images(self.workflow) + util.get_unique_images(self.workflow)

atomic_reactor/plugins/post_group_manifests.py
rcerven: why do we need repos here?

we should return digest, digest_type (or both of them in some dict), and list_json

if we won't return digest_type ( we would need to get_manifest if list_json wasn't specified, because if list_json is specified we know it is list.v2)

mlangsdorf: there's supposed to be a comment that it was testing purposes, but it turns out to be not be very useful.

rcerven: we are looping through floating tag images in push_floating_images anyway so we know target_repo

rcerven: ic :)

mlangsdorf: it's actually a ManifestDigest object.  returning a manifest digest was kind of the point of group_manifests, or so I thought.

is it okay to return it as `{'manifest_digest': digest, 'manifest_list': list_json}` ?

twaugh: Oh! I had missed that it's a ManifestDigest object.

Yes, that looks fine to return.

twaugh: So we'll return a ManifestDigest (which includes the type as well as the digest), and the manifest list text.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -243,20 +244,30 @@ def set_group_manifest_info(self, extra, worker_metadatas):
...
+            digest = None
+            # try to find the actual str digest value in the manifest digest
+            try:
+                for digest_version, digest in manifest_digest.default.items():
+                    break
+            # the manifest digest isn't a manifest digest, but a simple dict
+            except AttributeError:
+                for digest_version, digest in manifest_digest.items():
+                    if digest:
+                        break
```
twaugh: This doesn't look right. `manifest_digest.default` is a method.

It's a shame ManifestDigest.default() only returns the digest, where we also want the type. Ideally there would be a .default_type_and_digest() which would tell us both.

@MartinBasti @rcerven @athos-ribeiro thoughts?

atomic_reactor/manifest_util.py
```diff
@@ -0,0 +1,232 @@
...
+        self.log.debug("%s: Retrieving manifest for %s:%s", session.registry, repository, ref)
+
+        headers = {
+            'Accept': ', '.join(self.manifest_media_types)
+        }
+
+        url = '/v2/{}/manifests/{}'.format(repository, ref)
+        response = session.get(url, headers=headers)
+        response.raise_for_status()
+        return (response.content,
```
MartinBasti: I'd prefer `namedtuple` for such long rval

atomic_reactor/plugins/exit_koji_import.py
rcerven: @twaugh  group_manifest plugin should always return that digest as ManifestDigest even for non-v2list
then we don't need this logic at all

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -243,20 +244,30 @@ def set_group_manifest_info(self, extra, worker_metadatas):
             repo = ImageName.parse(repositories[0]).to_str(registry=False, tag=False)
             # group_manifests added the registry, so this should be valid
             registries = self.workflow.push_conf.all_registries
+            digest_version = None
+            digest = None
+            # try to find the actual str digest value in the manifest digest
+            try:
+                for digest_version, digest in manifest_digest.default.items():
```
MartinBasti: Could we check it by `if hasattr(manifest_digest, 'default'):` instead catching `AttributeError`?

MartinBasti: I'm quite lost here. What are you trying to achieve here?

atomic_reactor/plugins/exit_koji_import.py
twaugh: Yes, it should always be a ManifestDigest instance. But what's a nice way to get the digest and its content type from one of those? This 'for digest_version, digest in manifest_digest.items()' ok? It makes the assumption there is a single digest -- are we happy with that?

Alternative is to use manifest_digest.default() for the digest, then match that with a type.

MartinBasti: I don't understand issue here.  `default` returns default schema, so I can get digest from that:
```python
digest = manifest_digest[manifest_digest.default]
```

twaugh: Oh, `.default` is a *property*... sorry, I'd misread it as a method.

But it returns the digest, not the content type.

MartinBasti: Ah nevermind, I read that code wrongly

MartinBasti: Maybe we can add property `default_media_type`

atomic_reactor/manifest_util.py
MartinBasti: Not material for this PR, I noticed that this code was just moved

atomic_reactor/plugins/exit_koji_import.py
mlangsdorf: the __getattr__ logic in ManifestDigest frustrated my attempts to make this less ugly.

atomic_reactor/plugins/exit_koji_import.py
mlangsdorf: Under certain unit test conditions, manifest_digest exists but is not a ManifestDigest.  In that case, this is a fallback to retrieve the digest and digest_version.

twaugh: Can we just make the tests use a ManifestDigest here?

atomic_reactor/plugins/exit_koji_import.py
twaugh: @mlangsdorf so could we add a `default_media_type` property to ManifestDigest, which returns the abbreviated type name (e.g. 'v2_list') of the first digest value in the preference order list?

Both `default` and `default_media_type` should share the same preference order list, so it might make sense to have it defined in a class attribute.

Then this code can simply be
```
digest_version = manifest_digest.default_media_type
digest = manifest_digest[digest_version]
```

mlangsdorf: Fixed, and the errant test case changed.

atomic_reactor/plugins/exit_koji_import.py
mlangsdorf: It was using the wrong Manifest digest version or something; I just removed it from the test.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -243,20 +245,22 @@ def set_group_manifest_info(self, extra, worker_metadatas):
             repo = ImageName.parse(repositories[0]).to_str(registry=False, tag=False)
             # group_manifests added the registry, so this should be valid
             registries = self.workflow.push_conf.all_registries
+
+            digest_version = get_manifest_media_version(manifest_digest)
+            digest = manifest_digest.default[digest_version]
```
MartinBasti: Does this actually work?

```python
>>> md = ManifestDigest({'v1': 'digest'})
>>> ver = get_manifest_media_version(md)
>>> ver
'v1'
>>> md.default[ver]
TypeError: string indices must be integers
```

tests/plugins/test_koji_import.py
```diff
@@ -1702,7 +1702,9 @@ def test_koji_import_set_manifest_list_info(self, caplog, tmpdir, monkeypatch, o
                                             session=session,
                                             docker_registry=True,
                                             add_tag_conf_primaries=not is_scratch)
-        group_manifest_result = {'myproject/hello-world': digest} if digest else {}
+        group_manifest_result = {}
+        if digest:
+            group_manifest_result = {'manifest_digest': ManifestDigest(v2_list=digest)}
```
twaugh:
```suggestion
            group_manifest_result = {'manifest_digest': digest}
```

`digest` is already a `ManifestDigest` instance.

atomic_reactor/plugins/exit_koji_import.py
twaugh:
```suggestion
            digest = manifest_digest.default
```

Test is wrong, see further down.

twaugh:
The first commit is missing a Signed-Off-By line.

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,610 @@
...
+
+    def add_blob(self, name, blob):
+        repo = self.get_repo(name)
+        digest = make_digest(blob)
+        repo['blobs'][digest] = blob
+        return digest
+
+    def get_blob(self, name, digest):
+        return self.get_repo(name)['blobs'][digest]
+
```
twaugh: Not needed.

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,610 @@
...
+                'Location': '/v2/{}/blobs/{}'.format(target_name, digest),
+                'Docker-Content-Digest': digest,
+            }
+            return (201, headers, '')
+        except KeyError:
+            headers = {
+                'Location': '/v2/{}/blobs/uploads/some-uuid'.format(target_name),
+                'Docker-Upload-UUID': 'some-uuid',
+            }
+            return (202, headers, '')
```
twaugh: Not needed.

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,610 @@
...
+        self._add_pattern(responses.HEAD, r'/v2/(.*)/manifests/([^/]+)',
+                          self._get_manifest)
+        self._add_pattern(responses.PUT, r'/v2/(.*)/manifests/([^/]+)',
+                          self._put_manifest)
+        self._add_pattern(responses.GET, r'/v2/(.*)/blobs/([^/]+)',
+                          self._get_blob)
+        self._add_pattern(responses.HEAD, r'/v2/(.*)/blobs/([^/]+)',
+                          self._get_blob)
+        self._add_pattern(responses.POST, r'/v2/(.*)/blobs/uploads/\?mount=([^&]+)&from=(.+)',
+                          self._mount_blob)
```
twaugh: Blob API mocks not needed.

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,610 @@
...
+    """
+    This class mocks a subset of the v2 Docker Registry protocol
+    """
+    def __init__(self, registry):
+        self.hostname = registry_hostname(registry)
+        self.repos = {}
+        self._add_pattern(responses.GET, r'/v2/(.*)/manifests/([^/]+)',
+                          self._get_manifest)
+        self._add_pattern(responses.HEAD, r'/v2/(.*)/manifests/([^/]+)',
+                          self._get_manifest)
```
twaugh: GET/HEAD mocks not needed.

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,610 @@
...
+
+    worker_builds = {}
+
+    for platform, regs in config.items():
+        digests = []
+
+        for reg, tags in regs.items():
+            registry = reg_map[reg]
+            layer_digest = make_digest('layer-' + platform)
+            config_digest = make_digest('config-' + platform)
```
twaugh: Don't need config or layer for manifest lists (which are the only object we're manipulating).

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,610 @@
...
+
+            if schema_version == 'v2':
```
twaugh: Only v2 is supported for push_floating_tags right now. We're not creating any OCI image index object.
```suggestion
            assert schema_version == 'v2'
```

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,610 @@
...
+         'ppc64le': {
+             REGISTRY_V2: ['namespace/httpd:worker-build-ppc64le-latest'],
+             OTHER_V2: ['namespace/httpd:worker-build-ppc64le-latest'],
+         },
+         'x86_64': {
+             REGISTRY_V2: ['namespace/httpd:worker-build-x86_64-latest'],
+             OTHER_V2: ['namespace/httpd:worker-build-x86_64-latest'],
+         }
+     },
+     'No manifest list digest available, skipping push_floating_tags'),
```
twaugh: We're missing the only test we care about: when the schema version is "v2_list": that's the only thing group_manifests will ever return.

Looking at ManifestUtil, I think we'll hit an exception in the link method, which we'll need to sort out (i.e. skip trying to link references for v2_list).

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,610 @@
...
+
+
+REGISTRY_V2 = 'registry_v2.example.com'
+OTHER_V2 = 'registry.example.com:5001'
+
+
+V2_RESULTS = {
+    "digest": {
+        "v2_list": "sha256:11c3ecdbfa"
+    },
```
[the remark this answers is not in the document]
mlangsdorf: Fixed, thanks.

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,610 @@
...
+    "repos": json.dumps({
```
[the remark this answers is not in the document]
mlangsdorf: Fixed, thanks.

atomic_reactor/plugins/exit_push_floating_tags.py
```diff
@@ -0,0 +1,95 @@
...
+                           per-registry parameters.
+                           Params:
+                            * "secret" optional string - path to the secret, which stores
+                              login and password for remote registry
+        """
+        super(PushFloatingTagsPlugin, self).__init__(tasker, workflow)
+        self.manifest_util = ManifestUtil(workflow, registries, self.log)
+
+    def add_floating_tags(self, session, manifest_list_data, floating_images):
+        manifest_json = manifest_list_data.get("repos", {})
```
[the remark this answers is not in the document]
mlangsdorf: Fixed, thanks.

atomic_reactor/plugins/exit_push_floating_tags.py
```diff
@@ -0,0 +1,95 @@
...
+        manifest_digest = manifest_list_data.get("digest", {})
+        manifest_lists = json.loads(manifest_json)
```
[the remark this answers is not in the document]
mlangsdorf: Fixed, thanks.

tests/plugins/test_push_floating_tags.py
mlangsdorf: so this is the actual value of `target_registry.get_repo(name)` when name/tag is '2.4' from running `test_group_manifests.py`:

```
 'manifests': {
  'sha256:cd619643ae': b'{"schemaVersion": 2, "mediaType": "application/vnd.docker.distribution.manifest.v2+json", "config": {"mediaType": "application/vnd.docker.container.image.v1+json", "digest": "sha256:0efa9f4e8f"}, "layers": [{"mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip", "digest": "sha256:aca5f3af1d"}]}', 
  'sha256:9dc3bbcd6c': b'{"schemaVersion": 2, "mediaType": "application/vnd.docker.distribution.manifest.v2+json", "config": {"mediaType": "application/vnd.docker.container.image.v1+json", "digest": "sha256:2e1ad9b568"}, "layers": [{"mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip", "digest": "sha256:4bffe81499"}]}', 
 'sha256:11c3ecdbfa': '{\n    "manifests": [\n        {\n            "digest": "sha256:9dc3bbcd6c",\n            "mediaType": "application/vnd.docker.distribution.manifest.v2+json",\n            "platform": {\n                "architecture": "amd64",\n                "os": "linux"\n            },\n            "size": 306\n        },\n        {\n            "digest": "sha256:cd619643ae",\n            "mediaType": "application/vnd.docker.distribution.manifest.v2+json",\n            "platform": {\n                "architecture": "powerpc",\n                "os": "linux"\n            },\n            "size": 306\n        }\n    ],\n    "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",\n    "schemaVersion": 2\n}'}, 
  'tags': {
    'worker-build-ppc64le-latest': 'sha256:cd619643ae', 
    'worker-build-x86_64-latest': 'sha256:9dc3bbcd6c', 
    '2.4': 'sha256:11c3ecdbfa'
  }
}
```

Is that right?  It looks wrong to me.

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,497 @@
...
+
+        source_builds = {}
+        source_manifests = {}
+
+        for platform in workers:
+            build = annotations['worker-builds'][platform]['digests'][0]
+            source_builds[platform] = build
+            source_registry = mocked_registries[build['registry']]
+            source_manifests[platform] = source_registry.get_manifest(build['repository'],
+                                                                      build['digest'])
```
twaugh: I don't think this is needed.

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,497 @@
...
+                assert all(d['platform']['os'] == 'linux' for d in manifests)
+
+                for platform in annotations['worker-builds']:
+                    descs = [d for d in manifests
+                             if d['platform']['architecture'] == goarch[platform]]
+                    assert len(descs) == 1
+                    assert descs[0]['digest'] == source_builds[platform]['digest']
+
+                    verify_manifest_in_repository(target_registry, name,
+                                                  source_manifests[platform], platform)
```
twaugh: I don't think this is needed.

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,497 @@
...
+        for platform in workers:
+            build = annotations['worker-builds'][platform]['digests'][0]
+            source_builds[platform] = build
+            source_registry = mocked_registries[build['registry']]
+            source_manifests[platform] = source_registry.get_manifest(build['repository'],
+                                                                      build['digest'])
+
+        for registry, _ in registry_conf.items():
+            target_registry = mocked_registries[registry]
+            for image in test_images:
```
twaugh:
```suggestion
            for image in floating_tags:
```

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,497 @@
...
+                name, tag = image.split(':')
+
+                if tag not in target_registry.get_repo(name)['tags']:
+                    continue
+
+                raw_manifest_list = to_text(target_registry.get_manifest(name, tag))
```
twaugh: and then we want to verify that this is the same exact thing as the manifest list available from the `test_images`.

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,497 @@
...
+            source_manifests[platform] = source_registry.get_manifest(build['repository'],
+                                                                      build['digest'])
+
+        for registry, _ in registry_conf.items():
+            target_registry = mocked_registries[registry]
+            for image in test_images:
+                name, tag = image.split(':')
+
+                if tag not in target_registry.get_repo(name)['tags']:
+                    continue
```
twaugh: Should be:
```
assert tag in target_registry.get_repo(name)['tags']
```

i.e. that each floating tag has been added to the repository.

tests/plugins/test_push_floating_tags.py
twaugh: That looks fine for test_group_manifests. It's not what we get in test_push_floating_tags though: the parametrized floating tags are put there, and *they* are what we want to be checking.

I added comments further down explaining what's up.

tests/plugins/test_push_floating_tags.py
twaugh: i.e. loop over the floating tags, because that's what we care about.

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,464 @@
...
+            ReactorConfig({'version': 1,
+                           'registries': registries_list,
+                           'platform_descriptors': platform_descriptors_list})
+
+    runner = ExitPluginsRunner(tasker, workflow, plugins_conf)
+    results = runner.run()
+    plugin_result = results[PushFloatingTagsPlugin.key]
+
+    if expected_exception is None:
+        manifest_type, list_type = {
```
twaugh: `manifest_type` isn't used so pylint complains.

tests/plugins/test_koji_import.py
twaugh: Still needs fixing.

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,464 @@
...
+     'No manifest list digest available, skipping push_floating_tags'),
+])
+@responses.activate  # noqa
+def test_floating_tags_push(tmpdir, test_name, registries, manifest_results, schema_version,
+                            floating_tags, workers, expected_exception, reactor_config_map,
+                            caplog):
+    if MOCK:
+        mock_docker()
+
+    test_images = ['namespace/httpd:2.4', 'namespace/httpd:latest']
```
twaugh: These are intended to be images with 'primary' tags. To make the test less confusing to read can we use something like 2.4-1 instead of 2.4 and latest (which are both floating tags)?
```suggestion
    test_images = ['namespace/httpd:2.4-1']
```

tests/plugins/test_push_floating_tags.py
twaugh: We could do something like this before the loop:
```
primary_tag = test_images[0].split(':', 1)[1]
primary_manifest_list = target_registry.get_manifest(name, primary_tag)
```

Then here, in the loop, check it's the same value:
```
assert target_registry.get_manfiest(name, tag) == primary_manifest_list
```

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,464 @@
...
+                assert tag in target_registry.get_repo(name)['tags']
+
+                raw_manifest_list = to_text(target_registry.get_manifest(name, tag))
+                manifest_list = json.loads(raw_manifest_list, object_pairs_hook=OrderedDict)
+
+                # Check if the manifest list is sorted
+                assert json.dumps(manifest_list, indent=4, sort_keys=True,
+                                  separators=(',', ': ')) == raw_manifest_list
+                arch_list = [m['platform']['architecture'] for m in manifest_list['manifests']]
+                assert arch_list == sorted(arch_list)
```
twaugh: This is a valid test for group_manifests, but I don't think it's valid for push_floating_tags. When *creating* the manifest list, yes, we want to make sure the keys are sorted. But when pushing new tags for the existing manifest list, we have to keep it exactly as it was, however that was.

tests/plugins/test_push_floating_tags.py
mlangsdorf: I'm renaming it to `primary_tags` ; the actual test tags are passed as arguments to the test function.

twaugh: Thanks!

tests/plugins/test_push_floating_tags.py
mlangsdorf: Fixed

tests/plugins/test_push_floating_tags.py
mlangsdorf: Fixed

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,477 @@
...
+                           'platform_descriptors': platform_descriptors_list})
+
+    runner = ExitPluginsRunner(tasker, workflow, plugins_conf)
+    results = runner.run()
+    plugin_result = results[PushFloatingTagsPlugin.key]
+
+    if expected_exception is None:
+        _, list_type = {
+            'v2': (
+                'application/vnd.docker.distribution.manifest.v2+json',
```
twaugh: We're not using this value.

We probably want:
```
list_type = {
  'v2': 'application/vnd.docker.distribution.manifest.list.v2+json',
  'oci': 'application/vnd.oci.image.index.v1+json',
}[schema_version]
```

tests/plugins/test_koji_import.py
twaugh: @mlangsdorf, please take a look at this.

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,477 @@
...
+                'application/vnd.docker.distribution.manifest.list.v2+json',
+            ),
+            'oci': (
+                'application/vnd.oci.image.manifest.v1+json',
+                'application/vnd.oci.image.index.v1+json',
+            ),
+        }[schema_version]
+
+        primary_name, primary_tag = primary_images[0].split(':')
+        for registry, _ in registry_conf.items():
```
twaugh:
```suggestion
        for registry in registry_conf:
```

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,477 @@
...
+                assert target_registry.get_manifest(name, tag) == primary_manifest_list
+
+                raw_manifest_list = to_text(target_registry.get_manifest(name, tag))
+                manifest_list = json.loads(raw_manifest_list, object_pairs_hook=OrderedDict)
+
+                arch_list = [m['platform']['architecture'] for m in manifest_list['manifests']]
+                assert arch_list == sorted(arch_list)
+
+                assert manifest_list['mediaType'] == list_type
+                assert manifest_list['schemaVersion'] == 2
```
twaugh: IMHO lines 459 – 467 are not needed.

tests/plugins/test_push_floating_tags.py
mlangsdorf: Fixed.

tests/plugins/test_push_floating_tags.py
mlangsdorf: if we remove lines 459-467, this needs to go away entirely.

Fixed.

tests/plugins/test_push_floating_tags.py
mlangsdorf: Fixed, thanks.

tests/plugins/test_koji_import.py
mlangsdorf: Fixed, finally.

atomic_reactor/plugins/exit_koji_import.py
mlangsdorf: Fixed.

rcerven:
when we don't use manifest_list / group

we aren't tagging that v2s2 image with floating tags:
https://github.com/containerbuildsystem/atomic-reactor/pull/1282/files#diff-f763f089e65a59d18f50dfefa0f60bc9R178

so we have to tag those as well in new push_floating_tags plugin
as of now we are skipping that plugin:
https://github.com/containerbuildsystem/atomic-reactor/pull/1282/files#diff-b93396a2865903deb32f9c6b1e7555d9R83

twaugh:
Regarding @rcerven comment: 

1. tag_manifest_into_registry could registry the image manifest, media type, and digest
2. group_manifests's run() would return {"digest": the digest, "media_type": the media type, and "manifest": the actual manifest list or image manifest content}

mlangsdorf:
added a 4th commit to address rcevern's concerns.  I'll still need to address the fallout of the change i the other unit tests downstream from group_manifest but I want to be sure this is the correct approach first.

rcerven:
ok this looks better but can we do this? :

group_manifests plugin will return 
dict with : manifest_digest, media_type, image_manifest

we aren't using 'tags' anyway, and we don't need them

and it would be nice if group_manifests plugin would return always same dict

also we could omit media_type if we wanted, since we have manifest_list / image_manifest in 'image_manifest' from where we could get it

what do you think ? @twaugh @mlangsdorf

twaugh:
Yes, I'd rather group_manifests always returned the same keys. I described a set that would work for both group=True and group=False:

> group_manifests's run() would return {"digest": the digest, "media_type": the media type, and "manifest": the actual manifest list or image manifest content}

i.e. in the group=False case, the 'manifest' key in the return value will be the image manifest.
In the group=True case, 'manifest' will be for the manifest list.

The push_floating_tags plugin doesn't care about that content of these manifests. It just has to tag that object, whatever it is, with the floating tags. We want:
```
def add_floating_tags(self, session, manifest_list_data, floating_images):
    manifest = group_manifests["manifest"]
    manifest_digest = group_manifests["manifest_digest"]
    media_type = group_manifests["media_type"]
    ...
        self.manifest_util.store_manifest_in_repository(session, manifest_json, media_type,
                                                            target_repo, target_repo, ref=image.tag)
```

mlangsdorf:
How does this all this work with verify_media_types if group_manifests is always returning the same type of values whether or not group is true?

rcerven:
> How does this all this work with verify_media_types if group_manifests is always returning the same type of values whether or not group is true?

verify_media_types uses output of group_manifest only for expectation of manifest_list only

so you can make there change as well that IF group_manifest returns something and it isn't manifest_list type return that False here : https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/plugins/exit_verify_media_types.py#L101

mlangsdorf:
Changes made.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -234,9 +234,10 @@ def set_group_manifest_info(self, extra, worker_metadatas):
         floating_tags = [image.tag for image in floating_images]
         unique_tags = [image.tag for image in unique_images]
 
-        manifest_list_data = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY, {})
-        manifest_digest = manifest_list_data.get("manifest_digest")
-        if manifest_digest and isinstance(manifest_digest, ManifestDigest):
+        manifest_data = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY, {})
+        if manifest_data and is_manifest_list(manifest_data.get("media_type")):
+            print("media type is {}".format(manifest_data.get("media_type")))
```
twaugh: Please remove the print().

atomic_reactor/plugins/exit_verify_media_types.py
```diff
@@ -98,7 +98,9 @@ def get_manifest_list_only_expectation(self):
 
         :return: bool, expect manifest list only?
         """
-        if not self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY):
+        group_types = [MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST, MEDIA_TYPE_OCI_V1_INDEX]
+        manifest_results = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY)
+        if not manifest_results or manifest_results.get("media_type") not in group_types:
```
twaugh: Do we want to use the new `is_manifest_list` function here?

mlangsdorf: Yes, fixed.

atomic_reactor/plugins/exit_koji_import.py
mlangsdorf: oops.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -318,62 +154,32 @@ def sort_annotations(self):
...
                 if len(source) != 1:
                     raise RuntimeError('Without grouping only one source is expected')
-                for digest in source.values():
-                    self.tag_manifest_into_registry(session, digest)
-                    found = True
-                if not found:
-                    raise ValueError('Failed to find any platform')
-        return digests
+                for orig_digest in source.values():
+                    break
```
twaugh: pylint is complaining about using `orig_digest` after this loop. Maybe `orig_digest = source.values()[0]` is a more explicit way of saying that we need there to be values, and we'll just pick the first one we find.

mlangsdorf: source.values() is a `dict_values` object, not a list, and can't be indexed that way.

Made what's happening more explicit so the linter will stop complaining.

atomic_reactor/plugins/post_group_manifests.py
```diff
@@ -318,62 +154,34 @@ def sort_annotations(self):
...
-                for digest in source.values():
-                    self.tag_manifest_into_registry(session, digest)
-                    found = True
-                if not found:
-                    raise ValueError('Failed to find any platform')
-        return digests
+                # source.values() isn't a list and can't be indexed, so this clumsy workaround
+                for orig_registry in source:
+                    orig_digest = source[orig_registry]
+                    break
```
twaugh: Another way of doing this:
```
_, orig_digest = source.popitem()  # any will do
```

atomic_reactor/plugins/exit_push_floating_tags.py
```diff
@@ -0,0 +1,92 @@
...
+            return
+
+        #  can't run in the worker build
+        if not self.workflow.is_orchestrator_build():
+            self.log.warning('%s cannot be used by a worker builder', PLUGIN_PUSH_FLOATING_TAGS_KEY)
+            return
+
+        manifest_data = self.workflow.postbuild_results.get(PLUGIN_GROUP_MANIFESTS_KEY)
+        if not manifest_data or not manifest_data.get("manifest_digest"):
+            self.log.info('No manifest list digest available, skipping %s',
```
rcerven: 'No manifest list digest available' => 'No manifest digest available'

tests/plugins/test_push_floating_tags.py
```diff
@@ -0,0 +1,456 @@
...
+     'No manifest digest available, skipping push_floating_tags'),
+])
+@responses.activate  # noqa
+def test_floating_tags_push(tmpdir, test_name, registries, manifest_results, schema_version,
+                            floating_tags, workers, expected_exception, reactor_config_map,
+                            caplog):
+    if MOCK:
+        mock_docker()
+
+    primary_images = ['namespace/httpd:2.4', 'namespace/httpd:primary']
```
athos-ribeiro: nitpick: were the concepts of primary tags and floating tags swapped in the test suite on purpose? A floating tag is the one containing just the `version` while the primary one would be the one with `version-release`.

twaugh: @athos-ribeiro is right, those are examples of floating tags not primary tags. Previous discussion about this line: https://github.com/containerbuildsystem/atomic-reactor/pull/1282#discussion_r345284524

mlangsdorf: The tags are arbitrary for testing purposes.

mlangsdorf:
added to Release Notes under Improvements:
- new push_floating_tags plugins that will run after koji_import completes and retag the image with any floating tags

PR #1284, opened 2019-11-01
koji delegate fix

rcerven:
* OSBS-8150

### DIFF
--- a/atomic_reactor/plugins/pre_koji_delegate.py
+++ b/atomic_reactor/plugins/pre_koji_delegate.py
@@ -95,9 +95,16 @@ class KojiDelegatePlugin(PreBuildPlugin):
             # will be used in koji_import
             self.workflow.triggered_after_koji_task = self.triggered_after_koji_task
 
+        task_running = False
         koji_task_id = self.metadata.get('labels', {}).get('koji-task-id')
-        task_info = self.kojisession.getTaskInfo(koji_task_id, request=True)
-        task_running = koji.TASK_STATES[task_info['state']] == 'OPEN'
+        if koji_task_id:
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_koji_delegate.py
```diff
@@ -95,9 +95,12 @@ def run(self):
             # will be used in koji_import
             self.workflow.triggered_after_koji_task = self.triggered_after_koji_task
 
+        task_running = False
         koji_task_id = self.metadata.get('labels', {}).get('koji-task-id')
-        task_info = self.kojisession.getTaskInfo(koji_task_id, request=True)
-        task_running = koji.TASK_STATES[task_info['state']] == 'OPEN'
+        if koji_task_id:
+            task_info = self.kojisession.getTaskInfo(koji_task_id, request=True)
+            if task_info:
```
athos-ribeiro: What would having a taskID and no task_info mean here?

rcerven: for example if stage brew was replaced with prod data, so build_config would have stage task which doesn't exist in prod

athos-ribeiro: It would be nice to log a warning in this case then. Or we could have the build to fail and add some configuration to allow the build to proceed (which we could use in test environments).

athos-ribeiro:
It would be nice to describe the issue being fixed in the commit message as well

rcerven:
commit message updated

atomic_reactor/plugins/pre_koji_delegate.py
rcerven: added

rcerven:
rn not updated: fix for unreleased feature

PR #1288, opened 2019-11-06
tag and push source container image

rcerven:
* OSBS-8016

### DIFF
--- a/atomic_reactor/plugins/post_tag_and_push.py
+++ b/atomic_reactor/plugins/post_tag_and_push.py
@@ -11,14 +11,22 @@ from __future__ import absolute_import
 from copy import deepcopy
 import subprocess
 import time
+import platform
+import random
 
 from atomic_reactor.constants import (IMAGE_TYPE_DOCKER_ARCHIVE, IMAGE_TYPE_OCI, IMAGE_TYPE_OCI_TAR,
                                       MEDIA_TYPE_DOCKER_V2_SCHEMA2, DOCKER_PUSH_MAX_RETRIES,
                                       DOCKER_PUSH_BACKOFF_FACTOR)
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.plugins.exit_remove_built_image import defer_removal
-from atomic_reactor.plugins.pre_reactor_config import get_registries, get_group_manifests
-from atomic_reactor.util import (get_manifest_digests, get_config_from_registry, Dockercfg)
+from atomic_reactor.plugins.pre_reactor_config import (get_registries, get_group_manifests,
+                                                       get_koji_session, NO_FALLBACK,
+                                                       get_registries_organization)
+from atomic_reactor.plugins.pre_fetch_sources import PLUGIN_FETCH_SOURCES_KEY
+from atomic_reactor.util import (get_manifest_digests, get_config_from_registry, Dockercfg,
+                                 ImageName)
+import osbs.utils
+from osbs.constants import RAND_DIGITS
 
 
 __all__ = ('TagAndPushPlugin', )
@@ -32,7 +40,7 @@ class TagAndPushPlugin(PostBuildPlugin):
     key = "tag_and_push"
     is_allowed_to_fail = False
 
-    def __init__(self, tasker, workflow, registries=None):
+    def __init__(self, tasker, workflow, registries=None, koji_target=None):
         """
         constructor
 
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -95,11 +108,48 @@ def push_with_skopeo(self, registry_image, insecure, docker_push_secret):
...
             raise
 
+    def source_get_unique_image(self):
+        source_result = self.workflow.prebuild_results.get(PLUGIN_FETCH_SOURCES_KEY)
+        koji_build_id = source_result['sources_for_koji_build_id']
+
+        kojisession = get_koji_session(self.workflow, NO_FALLBACK)
+        user_params = os.environ['USER_PARAMS']
+        user_data = json.loads(user_params)
+        container_target = user_data.get('koji_target')
```
twaugh: Is that the only way to get to it? It's a shame we don't store it somewhere after validating it in the input plugin.

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -95,11 +108,48 @@ def push_with_skopeo(self, registry_image, insecure, docker_push_secret):
             self.log.error("push failed with output:\n%s", e.output)
             raise
 
+    def source_get_unique_image(self):
```
twaugh: It feels like tag_and_push should just apply the tags already decided on for this image. Maybe The source container build plugin should set tag_conf?

rcerven: I don't know, then again source container build should just build it :)
I'd rather keep logic here

atomic_reactor/plugins/post_tag_and_push.py
rcerven: we don't store it in workflow, we are just using it plugins_configurations

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -95,11 +108,48 @@ def push_with_skopeo(self, registry_image, insecure, docker_push_secret):
...
+        source_result = self.workflow.prebuild_results.get(PLUGIN_FETCH_SOURCES_KEY)
```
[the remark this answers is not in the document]
rcerven: sure

atomic_reactor/plugins/post_tag_and_push.py
twaugh: The way this normally seems to work is that we have the plugin take a `koji_target` parameter in its `__init__()` method, which osbs-client populates.

atomic_reactor/plugins/post_tag_and_push.py
twaugh: Can we run the `tag_from_config` plugin? That ought to set tag_conf I think.

May need a change in `render_tag_from_config` to make sure we only put the unique tag in its config.

atomic_reactor/plugins/post_tag_and_push.py
rcerven: sure I can do that and also create that osbs-client counterpart

MartinBasti: https://github.com/containerbuildsystem/osbs-client/pull/898

Work started here, I;d rather wait with this effort until that ^ PR is merged

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -95,11 +108,48 @@ def push_with_skopeo(self, registry_image, insecure, docker_push_secret):
...
+            timestamp,
+            current_platform
+        ]
+
+        tag = '-'.join(tag_segments)
+
+        get_build_meta = kojisession.getBuild(koji_build_id)
+        pull_specs = get_build_meta['extra']['image']['index']['pull']
+
+        source_image_spec = ImageName.parse(pull_specs[0])
```
twaugh: Do we need to `.enclose()` it first? Thinking about pre-Quay.io builds.

atomic_reactor/plugins/post_tag_and_push.py
twaugh: Seems like the team preference is to figure out the unique tag here rather than in osbs-client.

atomic_reactor/plugins/post_tag_and_push.py
twaugh: Maybe let's note this change for later then.

atomic_reactor/plugins/post_tag_and_push.py
rcerven: added

atomic_reactor/plugins/post_tag_and_push.py
rcerven: added note

rcerven: using koji_target from __init__

rcerven:
rn updated: support for source container builds in tag_and_push plugin

PR #1293, opened 2019-11-14
Source containers: workaround inner workflow

MartinBasti:

### DIFF
--- a/atomic_reactor/source.py
+++ b/atomic_reactor/source.py
@@ -16,6 +16,7 @@ import copy
 import os
 import shutil
 import tempfile
+from textwrap import dedent
 import collections
 try:
     from urlparse import urlparse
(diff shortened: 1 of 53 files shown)


MartinBasti:
Please check diff and let me know what do you think about this approach.

PS: tests are completely broken but source container and regular builds are passing
/me fixing

twaugh:
Yes, looks fine to me.

atomic_reactor/source.py
```diff
@@ -172,6 +173,38 @@ def get(self):
...
+        self, provider, uri, dockerfile_path=None,
+        provider_params=None, tmpdir=None
+    ):
+        # intentionally not calling `super()`
+        self._config = None
+        self.provider = provider
+        self.uri = uri
+        self.dockerfile_path = dockerfile_path
+        self.provider_params = provider_params or {}
+        # TODO: do we want to delete tmpdir when destroying the object?
```
twaugh: Let's not bother.

atomic_reactor/source.py
```diff
@@ -172,6 +173,38 @@ def get(self):
...
+        self.tmpdir = tmpdir or tempfile.mkdtemp()
+        logger.debug("workdir is %r", self.tmpdir)
+        self.source_path = tempfile.mkdtemp()
```
twaugh: Could be self.tmpdir?

athos-ribeiro:
The approach is fine

atomic_reactor/source.py
MartinBasti: I didn't want to mess `work dir` with `source dir` to prevent writing temp data into source directory (not an issue now but who knows in future)

MartinBasti: I can use:
```
self.source_path = tempfile.mkdtemp(dir=self.tmpdir)
```
This seems better

MartinBasti:
no release notes, part of unreleased feature

PR #1296, opened 2019-11-14
Get input files from relative path when testing

athos-ribeiro:
When looking for osbs-client input files, do so through a relative path.
This way developers can run the test suite locally without relying on
the test.sh script and without installing the input files in their
system.

### DIFF
--- a/tests/plugins/test_input_osv3.py
+++ b/tests/plugins/test_input_osv3.py
@@ -8,6 +8,7 @@ of the BSD license. See the LICENSE file for details.
 
 from __future__ import print_function, unicode_literals, absolute_import
 import os
+import pkgutil
 import json
 import types
 from textwrap import dedent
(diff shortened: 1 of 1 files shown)


twaugh:
I'm finding it hard to understand how the code does this.

athos-ribeiro:
> I'm finding it hard to understand how the code does this.

I wanted to avoid doing [this](https://github.com/containerbuildsystem/atomic-reactor/blob/50079079a2bb1b9c1823b7dc6cf79df9c0562f0a/test.sh#L20) in a local development environment while still being able to run the test suite without the need of using the `test.sh` file (just running pytest locally).

The idea here is to have the test file to look for the osbs input files by navigating the directories from a child dir (where the osbs packge is installed) to a parent dir. This will work if the osbs file is installed using pip (see https://github.com/containerbuildsystem/osbs-client/blob/868a8f453083a4cb5efcd5f6ef9cb3e703371c51/setup.py#L17).

For instance, I have osbs in a virtual env, in `/home/athos/.virtualenvs/atomic-reactor-python-2.7/lib/python2.7/site-packages/osbs` and osbs-client installs the input files in `/home/athos/.virtualenvs/atomic-reactor-python-2.7/share/osbs`

If, for some other reason, OSBS is being loaded from the system, we'd get the files from `/usr/share/osbs` if the installation is performed at `/usr/lib/pythonX.Y/site-packages/osbs` (this is the case for our upstream osbs-client RPM package).

tests/plugins/test_input_osv3.py
```diff
@@ -340,8 +341,14 @@ class TestOSv3SourceContainerInputPlugin(object):
 
     @property
     def user_params(self):
+        osbs_loader = pkgutil.find_loader('osbs')
+        osbs_pkg_file = osbs_loader.get_filename()
+        osbs_pkg_path = os.path.dirname(osbs_pkg_file)
```
twaugh: Is `osbs_pkg_path = os.path.dirname(osbs_loader.path)` what we want here?

twaugh:
I see. Can we have a comment there showing those example paths we're converting between?

Feels a bit weird converting a libdir to a datadir using relative paths but I don't see another way of doing it... other than explicitly making the input files 'resources'

tests/plugins/test_input_osv3.py
athos-ribeiro: ImpLoader does not have a `path` attribute in python 2.7

twaugh: Ah!

PR #1302, opened 2019-11-20
exit_koji_import: refund reserve builds that fail

mlangsdorf:
Fixes OSBS-7832

If a reserved build fails, call CGRefundBuild inside of koji_import before returning early from the plugin.

This change requires fetching the CG build id and token id earlier in exit_koji_import in order to have it when the early return from failure occurs.

### DIFF
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -15,6 +15,7 @@ import logging
 from tempfile import NamedTemporaryFile
 
 from atomic_reactor import start_time as atomic_reactor_start_time
+from atomic_reactor.constants import PROG
 from atomic_reactor.plugin import ExitPlugin
 from atomic_reactor.source import GitSource
 from atomic_reactor.plugins.build_orchestrate_build import (get_worker_build_info,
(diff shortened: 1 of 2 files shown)


tests/plugins/test_koji_import.py
```diff
@@ -130,6 +131,9 @@ def CGImport(self, metadata, server_dir, token=None):
         self.server_dir = server_dir
         return {"id": "123"}
 
+    def CGRefundBuild(self, cg, id, token):
```
twaugh: pylint is complaining about using `id` as a variable name here, since it's a Python builtin.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -462,13 +464,19 @@ def run(self):
...
+        # get the session and token information in case we need to refund a failed build
+        self.session = get_koji_session(self.workflow, self.koji_fallback)
+        build_token = self.workflow.reserved_token
+        build_id = self.workflow.reserved_build_id
+
         # Only run if the build was successful
         if self.workflow.build_process_failed:
             self.log.info("Not importing failed build to koji")
+            if self.reserve_build and build_token is not None:
+                build_info = self.session.CGRefundBuild(PROG, build_id, build_token)
```
twaugh: The call returns `None` on success, and anyway we don't use the return value, so we should just remove `build_info = ` at the start of this line.

mlangsdorf:
Added to release notes under Improvements:
- reserved builds that fail to build correctly will be refunded during exit_koji_import

PR #1311, opened 2019-11-25
pre_bump_release: optionally set a release environment variable

mlangsdorf:
Fixes osbs-7244

Add code in pre_bump_release that checks if set_release_env is set in the container.yml, and if it is, adds an ENV variable of the same name with the value of the release to the Dockerfile.

documentation PR: https://github.com/containerbuildsystem/osbs-docs/pull/117

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -179,6 +179,14 @@ class BumpReleasePlugin(PreBuildPlugin):
             raise RuntimeError('build already exists in Koji: {}-{}-{} ({})'
                                .format(component, version, release, build.get('id')))
 
+    def add_release_env_var(self, parser):
+        release_env_var = self.workflow.source.config.release_env_var
+        final_labels = Labels(parser.labels)
+        _, final_release = final_labels.get_name_and_value(Labels.LABEL_TYPE_RELEASE)
+        if release_env_var:
+            release_line = "ENV {}={}".format(release_env_var, final_release)
+            parser.add_lines(release_line, at_start=True, all_stages=True)
+
     def run(self):
         """
         run the plugin
@@ -251,3 +259,5 @@ class BumpReleasePlugin(PreBuildPlugin):
         if self.reserve_build and not is_scratch_build():
             self.reserve_build_in_koji(component, version, release, release_label,
                                        dockerfile_labels)
+
+        self.add_release_env_var(parser)
--- a/atomic_reactor/schemas/container.json
+++ b/atomic_reactor/schemas/container.json
@@ -165,6 +165,15 @@
         "demo"
       ]
     },
+    "set_release_env": {
+      "type": "string",
+      "description": "Name of an environment variable to set in the Dockerfile file with the release number",
+      "examples": [
+        "CONTAINER_RELEASE_NVR",
+        "IMAGE_RELEASE",
+        "OS_BUILD"
+      ]
+    },
     "version": {
       "type": "integer",
       "description": "Deprecated option kept for backwards compatibility. Should not be used",
--- a/atomic_reactor/source.py
+++ b/atomic_reactor/source.py
@@ -54,6 +54,7 @@ class SourceConfig(object):
                 )
                 raise
 
+        self.release_env_var = self.data.get('set_release_env')
         self.autorebuild = self.data.get('autorebuild') or {}
         self.flatpak = self.data.get('flatpak')
         self.compose = self.data.get('compose')
--- a/tests/plugins/test_bump_release.py
+++ b/tests/plugins/test_bump_release.py
@@ -52,7 +52,7 @@ class MockedClientSessionGeneral(object):
 
 
 class MockSource(object):
-    def __init__(self, tmpdir, add_timestamp=None):
+    def __init__(self, tmpdir, add_timestamp=None, release_env=None):
         self.dockerfile_path = str(tmpdir.join('Dockerfile'))
         self.path = str(tmpdir)
         self.commit_id = None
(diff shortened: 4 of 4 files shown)


atomic_reactor/inner.py
```diff
@@ -562,6 +562,12 @@ def build_docker_image(self):
 
             signal.signal(signal.SIGTERM, signal.SIG_DFL)
 
+    def add_release_var(self, build_id):
+        dockerfile = df_parser(self.builder.df_path, workflow=self)
+        if self.source.config.release_env_var:
+            release_line = "ENV {}={}".format(self.source.config.release_env_var, build_id)
+            dockerfile.add_lines(release_line, all_stages=True, at_start=True)
```
twaugh: Wondering about 'all_stages'. I think it makes sense to add this to all stages, but @rcerven I think you'd originally planned to add this only to the last one. Have a preference?

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -214,6 +214,7 @@ def run(self):
 
         try:
             release_label, release = labels.get_name_and_value(Labels.LABEL_TYPE_RELEASE)
+
```
twaugh: Was this intentional?

atomic_reactor/inner.py
```diff
@@ -562,6 +562,12 @@ def build_docker_image(self):
 
             signal.signal(signal.SIGTERM, signal.SIG_DFL)
 
+    def add_release_var(self, build_id):
```
twaugh: What's the thinking behind adding this to the DockerBuildWorkflow class, rather than having it as part of the bump_release plugin? I'm thinking this is something that only a single plugin would want to do.

atomic_reactor/schemas/container.json
```diff
@@ -165,6 +165,10 @@
         "demo"
       ]
     },
+    "set_release_env": {
+      "type": "string",
+      "description": "Name of an environmental variable to set in the Dockerfile file with the release number"
```
twaugh: environmental variable → environment variable

tests/test_inner.py
```diff
@@ -1411,3 +1412,54 @@ def test_fs_watcher(monkeypatch):
...
+        def __init__(self, df_path):
+            self.df_path = df_path
+
+    tmpdir_str = str(tmpdir)
+
+    fake_image_uri = os.path.join(tmpdir_str, 'fake_imag_uri')
+    os.mkdir(fake_image_uri)
+    if os.path.exists('fake_imag_uri'):
+        os.removedirs('fake_imag_uri')
+    os.mkdir('fake_imag_uri')
```
twaugh: Are these three lines needed?

tests/test_inner.py
```diff
@@ -1411,3 +1412,54 @@ def test_fs_watcher(monkeypatch):
...
+
+    yaml_path = os.path.join(fake_image_uri, REPO_CONTAINER_CONFIG)
+    with open(yaml_path, 'w') as yf:
+        yf.write(yml_config)
+        yf.flush()
```
twaugh: I don't think flush should be needed here as we're closing the file as soon as we leave the 'with' context.

atomic_reactor/schemas/container.json
```diff
@@ -165,6 +165,10 @@
         "demo"
       ]
     },
+    "set_release_env": {
```
twaugh: Could add `"examples": [ ... ]`

atomic_reactor/inner.py
mlangsdorf: I initially wasn't sure where it was going to end up, so I put it in DockerBuildWorkflow so I could call it from whatever plugin would need it.

I'll go ahead and move it to pre_bump_release.

atomic_reactor/plugins/pre_bump_release.py
mlangsdorf: No, thanks for catching it.

atomic_reactor/inner.py
rcerven: mmm I mean we can add it to all stages, that doesn't hurt

tests/test_inner.py
mlangsdorf: Not any longer.

atomic_reactor/inner.py
mlangsdorf: I removed all_stages.

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -179,6 +179,14 @@ def check_build_existence_for_explicit_release(self, component, version, release
             raise RuntimeError('build already exists in Koji: {}-{}-{} ({})'
                                .format(component, version, release, build.get('id')))
 
+    def add_release_env_var(self, parser):
+        release_env_var = self.workflow.source.config.release_env_var
+        final_labels = Labels(parser.labels)
+        _, final_release = final_labels.get_name_and_value(Labels.LABEL_TYPE_RELEASE)
+        if release_env_var:
+            release_line = "ENV {}={}".format(release_env_var, final_release)
+            parser.add_lines(release_line, at_start=True)
```
rcerven: let's add it to all stages, it won't hurt all_stages=True

atomic_reactor/schemas/container.json
mlangsdorf: Fixed and examples added.

twaugh:
Thanks!

Needs re-base.

twaugh:
Merged manually without the merge commit.

mlangsdorf:
Added to RN under Improvements:
- support set_releave_env value in container.yml that sets the build release as an environment variable inside the container

twaugh:
Oh, knew I forgot something. Sorry!

PR #1321, opened 2019-12-03
Use koji.pathinfo.signed() to get signed SRPM path

chmeliik:
* OSBS-8321

The gist of the change is that the SRPM filename is no longer taken from RPM headers, it is instead constructed from the NVR using `koji.pathinfo.signed()` (or `koji.pathinfo.rpm()` if there is no key).


twaugh:
Yes, for real builds there will be 'name'/'version'/'release' keys as well as 'nvr'.

chmeliik:
@twaugh thanks, squashed commits

twaugh:
The listRPMs call will return all the binary RPMs, but we need to map those to their SRPMs. Multiple installed RPMs could come from a single SRPM.

An example: listRPMs might return:
```
[
    "build_id": 745979,
    "name": "libacl",
    "extra": null,
    "arch": "x86_64",
    "buildtime": 1534063074,
    "id": 6156897,
    "epoch": null,
    "version": "2.2.53",
    "metadata_only": false,
    "external_repo_id": 0,
    "release": "1",
    "payloadhash": "523a222cdd868826a0cee07f1d169aa2",
    "size": 34536,
    "buildroot_id": 4369557,
    "external_repo_name": "INTERNAL",
    "nvr": "libacl-2.2.53-1"
  },
  ...
]
```
but the getRPMHeaders call will tell us what the SRPM name is:
```
{
  "SOURCERPM": "acl-2.2.53-1.el8.src.rpm"
}
```

So we want to use 'acl', not 'libacl', when constructing the 'src' path.

chmeliik:
@twaugh, I see. Is the `SOURCERPM` header the only way to map an RPM to its SRPM? The `pathinfo.signed()` method requires a dict with 'name', 'version', 'release', 'arch' keys, this way I would have to parse the NVR back out of the  filename.

twaugh:
I don't know another way. There is {{koji.parse_NVRA}} for parsing the filename.

twaugh:
Would be nice if the tests had a different name for the RPM than the SRPM, it would have picked this up.

chmeliik:
Reworked the PR, the change is now overall a lot simpler. Also added the test @twaugh suggested.

PR #1327, opened 2019-12-06
Add `annotation` decorator to utils

chmeliik:
* OSBS-7828

Will be used to give names to function results, particularly the `run()`
methods of plugins.

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -1596,6 +1596,9 @@ def annotation(*keys):
     The `run()` method of the plugin being annotated has to return a dict
     containing all of the specified keys.
 
+    The `store_metadata_in_osv3` plugin will automatically collect these
+    annotations and upload them to OpenShift.
+
     Example:
     >>> from atomic_reactor.plugin import BuildPlugin
 
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2015, 2018 Red Hat, Inc
+Copyright (c) 2015, 2018, 2019 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
@@ -9,6 +9,7 @@ of the BSD license. See the LICENSE file for details.
 from __future__ import print_function, unicode_literals, absolute_import
 
 import hashlib
+import functools
 from itertools import chain
 import json
 import jsonschema
@@ -1585,3 +1586,56 @@ def get_platform_config(platform, build_annotations):
     cm_key = cm_key_tmp[cmlen:]
 
     return cm_key, cm_frag_key
+
+
+def annotation(*keys):
+    """
+    Annotate a `BuildPlugin` subclass. The `run()` method of this plugin will
+    store its results in the plugin's workflow as annotations.
+
+    The `run()` method of the plugin being annotated has to return a dict
+    containing all of the specified keys.
+
+    Example:
+    >>> from atomic_reactor.plugin import BuildPlugin
+
+    >>> @annotation('foo', 'bar')
+    >>> class MyBuildPlugin(BuildPlugin):
+    >>>     key = 'my_build_plugin'
+    >>>
+    >>>     def run(self):
+    >>>         return {'foo': 1, 'bar': 2, 'baz': 3}
+
+    >>> p = MyBuildPlugin(...)
+    >>> p.run()
+    >>> p.workflow.annotations  # {'foo': 1, 'bar': 2}
+
+    :param keys: Keys to annotate the plugin with
+    :return: Decorator that will turn the plugin into an annotated one
+    """
+    from atomic_reactor.plugin import BuildPlugin
+
+    def annotate(cls):
+        if not issubclass(cls, BuildPlugin):
+            raise TypeError('Only subclasses of BuildPlugin can be annotated')
+
+        run = cls.run
+
+        @functools.wraps(run)
+        def annotated_run(self):
+            result = run(self)
+            if not isinstance(result, dict):
+                raise TypeError('Annotated run() method did not return a dict')
+            annotations = self.workflow.annotations
+            for key in keys:
+                if key not in result:
+                    raise RuntimeError('Annotation not found in result: {!r}'.format(key))
+                if key in annotations:
+                    raise RuntimeError('Annotation already set: {!r}'.format(key))
+                annotations[key] = result[key]
+            return result
+
+        cls.run = annotated_run
+        return cls
+
+    return annotate
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -420,6 +420,9 @@ class DockerBuildWorkflow(object):
         self.koji_source_source_url = None
         self.koji_source_manifest = None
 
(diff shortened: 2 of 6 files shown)


twaugh:
The ultimate aim of this is to be able to have the store_metadata_in_osv3 plugin collect together the annotated results and add them to the OpenShift Build annotations.

So the results should be returned unchanged, but the decorator should register the keys and values ready for the store_metadata_in_osv3 plugin to collect.

It's very possible we'll want multiple keys and values returned from a single function, so it might make sense to take a list of key names instead of a single key name.

chmeliik:
@twaugh, so if I am understanding this correctly, we want some sort of global-ish structure in which the results of functions are registered under the given keys?

Like, for example:
```python
@annotation('foo')
def f(x):
    return x

@annotation('foo')
def g(x):
    return x

@annotation('bar')
def h(x):
    return x


f(1)
f(2)
g(3)
h(4)

<all annotations> == {'foo': [1, 2, 3], 'bar': [4]}
```

If so, I have a few questions:
* Is it possible that one annotation will be used more than once, as in the example above?
* Do I need to worry about re-initializing and/or cleaning up this annotation structure? I believe an atomic-reactor process only runs one plugin configuration and exits, so this might not be an issue, not sure
* How would we get multiple values from a single function? I can imagine annotating a function with multiple keys, but I don't see how anything other than its return value could be associated with those keys

I based this implementation on the description of OSBS-7828, but it is possible that I am missing something.

chmeliik:
After inspecting the `store_metadata_in_osv3` plugin, I got an idea - how about storing the annotations in `DockerBuildWorkflow`? The plugin already gets build result annotations from there: https://github.com/containerbuildsystem/atomic-reactor/blob/eb56ba9cc9e1d0b828d998e442fd233482e020ac/atomic_reactor/plugins/exit_store_metadata_in_osv3.py#L190

The `@annotation` decorator would then work kind of like this:
```python
def annotation(key):
    def annotate(method):
        def annotated(self, *args, **kwargs):
            result = method(self, *args, **kwargs)
            self.workflow.annotations[key].append(result)
            return result
        return annotated
    return annotate
```
This would only work for instance methods of plugins which have the `workflow` attribute (`BuildPlugin` and subclasses) but maybe that is good enough?

twaugh:
Yes, that sounds like a great direction.

chmeliik:
Reworked implementation

atomic_reactor/util.py
```diff
@@ -1557,3 +1558,39 @@ def has_operator_manifest(workflow):
     except KeyError:
         operator_label = 'false'
     return operator_label.lower() == 'true'
+
+
+def annotation(key):
```
chmeliik: Do we want to allow multiple keys here?

twaugh: I think it would be useful, yes.

chmeliik: ok, changed

atomic_reactor/util.py
```diff
@@ -1557,3 +1558,40 @@ def has_operator_manifest(workflow):
...
+    :param keys: Keys to annotate the method with
+    :return: Decorator that will turn the method into an annotated one
+    """
+
+    def annotate(build_plugin_method):
+        @functools.wraps(build_plugin_method)
+        def annotated(self, *args, **kwargs):
+            result = build_plugin_method(self, *args, **kwargs)
+            for key in keys:
+                prev_results = self.workflow.annotations.setdefault(key, [])
```
MartinBasti: Do we want to allow multiple plugins to append to the same annotation? Shouldn't this rather fail when updating existing annotation?

chmeliik: That's what I don't know, but your guess is probably better than mine

chmeliik: @twaugh what do you think?

twaugh: Let's make it fail. If we decide later it would be useful to combine multiple results we can change it then.

chmeliik: ok, done

MartinBasti:
Do we have a different use case than only decorating `run` method here? Shall it be a class decorator in that case?

twaugh:
Class decorator: that's a nice idea actually.

chmeliik:
Agreed, it is now a class decorator

atomic_reactor/util.py
```diff
@@ -1557,3 +1558,49 @@ def has_operator_manifest(workflow):
...
+        run = cls.run
+
+        @functools.wraps(run)
+        def annotated_run(self):
+            result = run(self)
+            annotations = self.workflow.annotations
+            for key in keys:
+                if key in annotations:
+                    raise RuntimeError('Annotation {!r} is already set'.format(key))
+                annotations[key] = result
```
twaugh: The reason for wanting to specify multiple annotation keys is to be able to return multiple *different* values and assign them to different keys. For example, we might want to set both 'pullspecs':'example.registry.com/foo/bar:baz' and 'digests':{'v1':'spam','v2':'maps'} from the same plugin, and we'd want the annotations to be:
* 'pullspecs', with the string value 'example.registry.com/foo/bar:baz'
* 'digests', with the string value being the JSON encoding of the dict, i.e. '{"v1": "spam", "v2": "maps": }'

So there are several ways to get the different values assigned to the keys:
* maybe the run() method has to return a tuple, and we map each tuple item to the keys in the order they were given (raise an error if the numbers don't match?)
* or maybe run() has to return a dict, and the values are fetched from that dict by the annotation key names provided
* or maybe you can think of a nicer scheme for it

chmeliik:
Now requiring `run()` methods of annotated plugins to return dicts.

atomic_reactor/util.py
```diff
@@ -1557,3 +1558,64 @@ def has_operator_manifest(workflow):
...
+    :param keys: Keys to annotate the plugin with
+    :return: Decorator that will turn the plugin into an annotated one
+    """
+    from atomic_reactor.plugin import BuildPlugin
+
+    def update_annotations(annotations, result):
+        if set(keys) != set(result):
+            msg = ('Expected the following annotations: {}, got: {}'
+                   .format(sorted(keys), sorted(result)))
+            raise TypeError(msg)
```
chmeliik: Would we want to allow the result to have keys other than those expected by the annotation? If not, then specifying the keys in the decorator is a bit redundant but it might serve as an extra bit of "self-documentation".

twaugh: I can *definitely* see it being useful to return other keys than those we want an annotation for. Some plugins need to communicate to others, but we don't necessarily want to clutter up the Build annotations with that chatter (annotation size is limited).

chmeliik: Right, I suspected that

chmeliik: Also updated the example in docstring to reflect this

twaugh:
> 29 of 29 new or added lines in 3 files covered. (100.0%)

Waiving coverage review item.

MartinBasti:
Needs rebase  :(

chmeliik:
Rebased on master

chmeliik:
No idea why, but `responses 0.10.8` breaks tests (in master too)

chmeliik:
Now requiring responses < 0.10.8

MartinBasti:
No release notes: internal API change

PR #1338, opened 2019-12-13
Cachito integration plugin

lcarva:

### DIFF
--- a/atomic_reactor/plugins/pre_reactor_config.py
+++ b/atomic_reactor/plugins/pre_reactor_config.py
@@ -9,6 +9,7 @@ of the BSD license. See the LICENSE file for details.
 from __future__ import absolute_import
 
 from copy import deepcopy
+from atomic_reactor.cachito_util import CachitoAPI
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.constants import (CONTAINER_BUILD_METHODS, CONTAINER_DEFAULT_BUILD_METHOD,
                                       CONTAINER_BUILDAH_BUILD_METHOD)
@@ -129,6 +130,26 @@ def get_smtp_session(workflow, fallback):
 
     import smtplib
     return smtplib.SMTP(config['host'])
+
+
+def get_cachito(workflow):
+    return get_value(workflow, 'cachito', NO_FALLBACK)
+
+
+def get_cachito_session(workflow):
+    config = get_cachito(workflow)
+
+    api_kwargs = {'insecure': config.get('insecure', False)}
+
+    ssl_certs_dir = config['auth'].get('ssl_certs_dir')
+    if ssl_certs_dir:
+        cert_path = os.path.join(ssl_certs_dir, 'cert')
+        if os.path.exists(cert_path):
+            api_kwargs['cert'] = cert_path
+        else:
+            raise RuntimeError("Cachito ssl_certs_dir doesn't exist")
+
+    return CachitoAPI(config['api_url'], **api_kwargs)
 
 
 def get_arrangement_version(workflow, fallback=NO_FALLBACK):
--- a/atomic_reactor/schemas/container.json
+++ b/atomic_reactor/schemas/container.json
@@ -156,6 +156,51 @@
       "description": "build-step plugin to be used",
       "enum": ["docker_api", "imagebuilder", "buildah_bud"]
     },
+    "remote_source": {
+      "type": ["object", "null"],
+      "properties": {
+        "repo": {
+          "type": "string",
+          "description": "URL to the SCM repository",
+          "examples": [
+            "https://git.example.com/team/repo.git",
+            "https://git.example.com/another-repo.git"
+          ]
+        },
+        "ref": {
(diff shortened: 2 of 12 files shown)


atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -131,6 +131,27 @@ def get_smtp_session(workflow, fallback):
     return smtplib.SMTP(config['host'])
 
 
+def get_cachito(workflow):
+    return get_value(workflow, 'cachito', NO_FALLBACK)
+
+
+def get_cachito_session(workflow):
+    config = get_cachito(workflow)
+    from atomic_reactor.cachito_util import CachitoAPI
```
MartinBasti: Why is this import here in function? Cyclic imports?

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -131,6 +131,27 @@ def get_smtp_session(workflow, fallback):
...
+
+    api_kwargs = {'insecure': config.get('insecure', False)}
+
+    ssl_certs_dir = config['auth'].get('ssl_certs_dir')
+    if ssl_certs_dir:
+        cert_path = os.path.join(ssl_certs_dir, 'cert')
+        if os.path.exists(cert_path):
+            api_kwargs['cert'] = cert_path
+        else:
+            raise KeyError("Cachito ssl_certs_dir doesn't exist")
```
MartinBasti: IMHO it should be `ValueError`

atomic_reactor/schemas/container.json
```diff
@@ -156,6 +156,51 @@
       "description": "build-step plugin to be used",
       "enum": ["docker_api", "imagebuilder", "buildah_bud"]
     },
+    "remote_source": {
```
MartinBasti: Wouldn't be better to having it in this format for future extension:

```
remote_sources: [
    {
      "repo": "git.example.com/proj-go",
      "ref": "b55c00f45ec3dfee0c766cea3d395d6e21cc2e5a",
      "pkg_manager": "gomod",
      "flags": [...cachito flags...],
   }, {
      "repo": "another.git.com/pyproject",
      "ref": "651abc41170e52866148cc0c99c3671a2ace7002",
      "pkg_manager": "pip",
      "flags": [...cachito flags...],
   }
]
```

atomic_reactor/plugins/pre_reactor_config.py
MartinBasti: Or `RuntimeError`

atomic_reactor/plugins/pre_reactor_config.py
lcarva: I followed the pattern in `get_odcs_session`. IIRC, this is so we can split up the plugin into subpackages. If that's not needed, I can move it to the top.

atomic_reactor/plugins/pre_reactor_config.py
lcarva: It's `KeyError` because that's what `get_value` raises if a config for `cachito` doesn't exist at all. I suppose `get_odcs_session` does this so it's easier to use by only catching a single exception.

atomic_reactor/schemas/container.json
lcarva: During the design phase, we decided to only support a single remote source. But a single remote source could have multiple package managers. This is inline with upstream projects.

Having multiple remote sources is challenging because if you have, for example, two goloang apps, the value for GOCACHE/GOPATCH is ambiguous. 

Given the complexities in implementing this, we decided that this could be added later on if there was need.

lcarva: During the design phase we decided to only support a single remote source, but allow multiple package managers for it. 

Supporting multiple remote sources can be challenging. For instance, if there are two remote sources using gomod, it's unclear what the value of GOCACHE/GOPATH should be set to. Should it point to the cache of the first remote source, or the second?

Given the complexities that can arise from this scenario, we decided to only support a single remote source until a real use case arises to makes us reconsider.

atomic_reactor/plugins/pre_reactor_config.py
MartinBasti: Please put it on top of the module

atomic_reactor/schemas/container.json
MartinBasti: ok

atomic_reactor/plugins/pre_reactor_config.py
lcarva: Done!

atomic_reactor/plugins/pre_reactor_config.py
lcarva: You're right though. `KeyError` doesn't make any sense. I changed it to `RuntimeError`.

lcarva:
Added issue ID to each commit. I think I addressed all the comments.

MartinBasti:
no release notes, when feature is done a beautiful release notes will be added

PR #1339, opened 2019-12-13
Use annotation mechanism for suitable plugins

chmeliik:
* OSBS-7828

`[WIP]` commits contain TODOs explaining what is still WIP.

Also, if anyone has a better idea for the `annotation_map` and `label_map` function names, that would be welcome.


chmeliik:
It appears that returning `'help_file': None` from `add_help` plugin is indeed OK

chmeliik:
I'm assuming labels will always be simple values (numbers, strings) which is why did not `json.dumps` them but simply convert to `str`. Dropping the remaining TODOs.

chmeliik:
Rebased on master

twaugh:
Nice!

Seeing it in action I wonder whether the decorators should live in a new `metadata.py` module,
```
from atomic_reactor import metadata

@metadata.label(...)
class ...
```

What do you think?

chmeliik:
> Looks really nice! Think you'll edit the commits to add things in logical order, or keep them as they are?

Can do, would the logical order be `update decorators -> move to new module -> update plugins`?

twaugh:
Or move, update decorators, update plugins.

atomic_reactor/util.py
```diff
@@ -1591,33 +1591,47 @@ def get_platform_config(platform, build_annotations):
...
     annotations and upload them to OpenShift.
 
     Example:
-    >>> from atomic_reactor.plugin import BuildPlugin
-
     >>> @annotation('foo', 'bar')
     >>> class MyBuildPlugin(BuildPlugin):
     >>>     key = 'my_build_plugin'
     >>>
+    >>>     # sets annotations: {'foo': 1, 'bar': 1}
```
twaugh: Not sure about this. It sets the same value on both annotations? It doesn't seem like useful behaviour to duplicate it.

Should it just take a single argument (the annotation name)? If it takes multiple arguments I'd expect it to unpack a tuple from the return value.

chmeliik: Fair point, I don't think I've seen any plugins returning tuples so I will probably just make it take only a single argument.

chmeliik:
Reordered commits, I went with update first, then move (which was easier), hope you don't mind

chmeliik:
Made `annotation` and `label` accept only 1 key

atomic_reactor/util.py
```diff
@@ -1616,29 +1616,53 @@ def annotation(*keys):
...
+    """
+    label_plugin = _decorate_metadata('labels', keys)
+    return label_plugin
+
+
+def _decorate_metadata(metadata_type, keys):
     from atomic_reactor.plugin import BuildPlugin
 
-    def annotate(cls):
+    wrong_class = '[{}] Not a subclass of BuildPlugin'.format(metadata_type)
```
MartinBasti: I'd prefer lazy formatting here

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -33,10 +33,12 @@
 from atomic_reactor.koji_util import TaskWatcher, stream_task_output
 from atomic_reactor.yum_util import YumRepo
 from atomic_reactor.util import get_platforms, df_parser, base_image_is_custom, ImageName
+from atomic_reactor.metadata import label_map
 from atomic_reactor import util
 from osbs.utils import Labels
 
 
+@label_map('filesystem-koji-task-id')
```
MartinBasti: Should we have constants for label and annotations keys?

chmeliik:
Added test and changed where error messages are created.

As for constants, this actually reduces the need (because the `store_metadata_in_osv3` plugin no longer needs to know what the annotations and labels are called) but I can add them if they are needed anyway.

chmeliik:
Force-pushed for coveralls results

MartinBasti:
Internal change, no release notes needed

PR #1340, opened 2019-12-13
fail source build if no srpms found already in fetch_sources

rcerven:
* OSBS-8372

### DIFF
--- a/atomic_reactor/plugins/pre_fetch_sources.py
+++ b/atomic_reactor/plugins/pre_fetch_sources.py
@@ -73,6 +73,12 @@ class FetchSourcesPlugin(PreBuildPlugin):
         koji_config = get_koji(self.workflow, {})
         insecure = koji_config.get('insecure_download', False)
         urls = self.get_srpm_urls(signing_intent['keys'], insecure=insecure)
+
+        if not urls:
+            msg = "No srpms found for source container, would produce empty source container image"
+            self.log.error(msg)
+            raise RuntimeError(msg)
+
         sources_dir = self.download_sources(urls, insecure=insecure)
         return {
                 'sources_for_koji_build_id': self.koji_build_id,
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -73,6 +73,10 @@ def run(self):
         koji_config = get_koji(self.workflow, {})
         insecure = koji_config.get('insecure_download', False)
         urls = self.get_srpm_urls(signing_intent['keys'], insecure=insecure)
+
+        if not urls:
+            raise RuntimeError('No srpms found for source container')
```
athos-ribeiro: Should we inform the build is being aborted because this would generate an empty source container image? Could either be in the Error message or just a comment to be explicit on why we are aborting here.

rcerven: added error  msg

MartinBasti: IMHO what athos meant was reporting to user via exception that source container image is empty and using log.error() to print reason why it is empty.

rcerven: in this plugin we don't have yet container image, so the reason why plugin failed is because there aren't any srpms,
also in error message explaining that it would produce empty image

athos-ribeiro: I was worrying about a developer seeing this a while from now and not getting the full picture of why not proceeding with an empty `urls` list, or what and empty `urls` list would mean. It would also be nice to have the same message in the error log and in the exception since (from my experience) OSBS users may read just the exception message (due to the size of our logs) before opening bugs.

This LGTM and I will leave the "having the same message in both the error log and the exception"  as a (non-blocker) nitpick

rcerven: added same msg to both

rcerven:
rn updated: fail source build if no srpms found already in fetch_sources

PR #1341, opened 2019-12-16
send email for autorebuilds to submitter from original task

rcerven:
* OSBS-8395

### DIFF
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -173,6 +173,15 @@ class SendMailPlugin(ExitPlugin):
         else:
             self.log.info("Koji task ID: %s", self.koji_task_id)
 
+        try:
+            metadata = get_build_json().get("metadata", {})
+            self.original_koji_task_id = int(metadata['labels']['original-koji-task-id'])
+        except Exception:
+            self.log.exception("Failed to fetch original koji task ID")
+            self.original_koji_task_id = None
+        else:
+            self.log.info("original Koji task ID: %s", self.original_koji_task_id)
+
         self.koji_build_id = self.workflow.exit_results.get(KojiImportPlugin.key)
         if not self.koji_build_id:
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -343,7 +352,10 @@ def _get_email_from_koji_obj(self, obj):
             return '@'.join([obj['name'], self.email_domain])
 
     def _get_koji_submitter(self):
-        koji_task_owner = get_koji_task_owner(self.session, self.koji_task_id)
+        if self.original_koji_task_id:
+            koji_task_owner = get_koji_task_owner(self.session, self.original_koji_task_id)
+        else:
+            koji_task_owner = get_koji_task_owner(self.session, self.koji_task_id)
```
twaugh: Minor: could be simplified with
```
koji_task_owner = get_koji_task_owner(self.session, self.original_koji_task_id or self.koji_task_id)
```

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -173,6 +173,15 @@ def __init__(self, tasker, workflow,
         else:
             self.log.info("Koji task ID: %s", self.koji_task_id)
 
+        try:
+            metadata = get_build_json().get("metadata", {})
+            self.original_koji_task_id = int(metadata['labels']['original-koji-task-id'])
+        except Exception:
```
athos-ribeiro: What other exceptions are we expecting here other than TypeError or KeyError? Could we be explicit on those here?

atomic_reactor/plugins/exit_sendmail.py
rcerven: ok

atomic_reactor/plugins/exit_sendmail.py
rcerven: using the same as in getting koji-task-id above, we don't really care that much in this case how it fails, we either get value or not

rcerven:
rn updated: send email for autorebuilds to submitter from original task

PR #1345, opened 2020-01-05
Add Koji support for remote sources

athos-ribeiro:
* Use new remote-sources koji btype
* Upload remote sources artifacts (tarball and json) as new btype
* Update build extra.image.remote_source_url metadata

* Relates to Cachito integration
* OSBS-8137

### DIFF
--- a/atomic_reactor/constants.py
+++ b/atomic_reactor/constants.py
@@ -163,6 +163,10 @@ OPERATOR_MANIFESTS_ARCHIVE = 'operator_manifests.zip'
 
 KOJI_BTYPE_IMAGE = 'image'
 KOJI_BTYPE_OPERATOR_MANIFESTS = 'operator-manifests'
+KOJI_BTYPE_CONTAINER_FIRST = 'remote-sources'
 
 # Path to where the remote source bundle is copied to during the build process
 REMOTE_SOURCE_DIR = '/remote-source'
+
+# Name of downloaded remote sources tarball
+REMOTE_SOURCES_FILENAME = 'remote-sources.tar.gz'
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -17,7 +17,6 @@ import logging
 from tempfile import NamedTemporaryFile
 
 from atomic_reactor import start_time as atomic_reactor_start_time
-from atomic_reactor.constants import PROG
 from atomic_reactor.plugin import ExitPlugin
 from atomic_reactor.source import GitSource
 from atomic_reactor.plugins.build_orchestrate_build import (get_worker_build_info,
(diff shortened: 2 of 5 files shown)


atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -255,10 +260,24 @@ def set_operators_metadata(self, extra, worker_metadatas):
...
+        if remote_source_result:
+            url = remote_source_result['annotations']['remote_source_url']
+            remote_source_typeinfo = {
+                KOJI_BTYPE_CONTAINER_FIRST: {
+                    'remote_source_url': url,
+                },
+            }
+            extra.setdefault('typeinfo', {}).update(remote_source_typeinfo)
+
+            # TODO: is setting it in the image metadata also needed?
```
athos-ribeiro: We do need the container-first metadata in the typeinfo to set the btype here. Should we use it to hold the remote_source_url metadata as above? In this case, should we keep the duplication below?

atomic_reactor/constants.py
```diff
@@ -163,6 +163,10 @@
 
 KOJI_BTYPE_IMAGE = 'image'
 KOJI_BTYPE_OPERATOR_MANIFESTS = 'operator-manifests'
+KOJI_BTYPE_CONTAINER_FIRST = 'container-first'
```
athos-ribeiro: Would anyone find this btype name confusing for holding the remote sources archives?

twaugh: We've talked about other names for this:
* upstream-sources (implies a downstream)
* container-sources
* ...

What about 'remote-sources', for consistency with what we already call the tarball?

atomic_reactor/plugins/exit_koji_import.py
twaugh: I'm not sure that's right.

'typeinfo' is about the build, not about a specific output of the build:

> [build.extra.]typeinfo: A map whose **single entry** is the name of the content generator type, which is a map containing type-specific information for this build.

(from [docs](https://docs.pagure.org/koji/content_generator_metadata/), emphasis mine)

whereas outputs have:

> type: The type of the file. Log files should use “log”.

and no other types are specified -- seems like it's a freeform string. There's no typeinfo in the output's 'extra' map either.

athos-ribeiro: I am not sure how accurate is the documentation though...

As per [1], it seems several entries are processed in the typeinfo map. With [2] it seems a build type is set based on the information in the typeinfo field.

By looking at [3], it also seems that we need the btype in the typeinfo so koji will build (and display) URLs for the archives.

The reason we do not have typeinfo in most of OSBS build metadata is because most builds use only the `image` legacy type. Operator builds do have the typeinfo field as well since they have their own koji btype.

[1] https://pagure.io/koji/blob/master/f/hub/kojihub.py#_6092
[2] https://pagure.io/koji/blob/master/f/hub/kojihub.py#_4300
[3] https://pagure.io/koji/blob/master/f/www/kojiweb/index.py#_1124

twaugh: OK, thanks for finding the facts :-)

atomic_reactor/constants.py
athos-ribeiro: Thinking about it, it seems to be the best option here.

2 files will be presented with this type for now: The archive with the remote sources and some metadata with the information used to fetch the remote sources.

atomic_reactor/plugins/exit_koji_import.py
athos-ribeiro: I opened https://pagure.io/koji/pull-request/1917 to update those docs.

athos-ribeiro:
Docs PR: https://github.com/containerbuildsystem/osbs-docs/pull/123

athos-ribeiro:
Release Notes added:

- upload cachito fetched sources and metadata to koji as a new remote-sources BType

PR #1347, opened 2020-01-07
Add additional remote sources build args

athos-ribeiro:
We add build arguments to allow remote sources to be copied into the
build image. Since the ADD command can extract tar archives directly,
there is no more need to extract the sources in the workdir.

To accomplish this, a Dockerfile must contain the following
instructions:

```
RUN mkdir $REMOTE_SOURCE_DIR
ADD $REMOTE_SOURCE $REMOTE_SOURCE_DIR
```

* OSBS-8473
* OSBS-8135

### DIFF
--- a/atomic_reactor/plugins/pre_download_remote_source.py
+++ b/atomic_reactor/plugins/pre_download_remote_source.py
@@ -10,8 +10,10 @@ Downloads and unpacks the source code archive from Cachito and sets appropriate 
 
 from __future__ import absolute_import
 
+import os
 import tarfile
 
+from atomic_reactor.constants import REMOTE_SOURCE_DIR
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.download import download_url
 
@@ -19,6 +21,7 @@ from atomic_reactor.download import download_url
 class DownloadRemoteSourcePlugin(PreBuildPlugin):
     key = 'download_remote_source'
     is_allowed_to_fail = False
+    REMOTE_SOURCE = 'unpacked_remote_sources'
 
     def __init__(self, tasker, workflow, remote_source_url,
                  remote_source_build_args=None):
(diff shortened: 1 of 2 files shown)


athos-ribeiro:
Docs PR: https://github.com/athos-ribeiro/osbs-docs/pull/new/add-remote-source

atomic_reactor/plugins/pre_download_remote_source.py
```diff
@@ -40,11 +41,19 @@ def run(self):
         # Download the source code archive
         archive = download_url(self.url, self.workflow.source.workdir)
 
-        # Unpack the source code archive into the workdir
```
athos-ribeiro: Read $COMMIT_MESSAGE for an explanation for removing this.

@rcerven brought an interesting question on IRC though: if we are to revert this and go back to extracting the sources in the workdir, would it be possible for files, such as the Dockerfile, to be overwritten? In this case, is this desirable?

twaugh:
Do we need to watch out for https://github.com/openshift/imagebuilder/issues/99 and/or https://github.com/openshift/imagebuilder/issues/139 though?

atomic_reactor/plugins/pre_download_remote_source.py
MartinBasti: I'd say that dockerfile shouldn't be replaced ever.

atomic_reactor/plugins/pre_download_remote_source.py
```diff
@@ -40,11 +41,19 @@ def run(self):
...
-            tf.extractall(self.workflow.source.workdir)
+        if not tarfile.is_tarfile(archive):
+            raise RuntimeError('Remote source archive {} is not a tar file'.format(archive))
 
         # Set build args
         self.workflow.builder.buildargs.update(self.buildargs)
 
+        # To extract the sources into the build image, Dockerfile should contain
+        # RUN mkdir $REMOTE_SOURCE_DIR
+        # ADD $REMOTE_SOURCE $REMOTE_SOURCE_DIR
```
MartinBasti: https://github.com/openshift/imagebuilder/issues/99

Is this issue related?

athos-ribeiro:
@twaugh, @MartinBasti: I did not account for those two imagebuilder issues. How about

- extract the sources in a temporary directory
- set $REMOTE_SOURCE to that directory (without a slash as the first char if in the workdir)
- advise COPY instead of ADD

?

athos-ribeiro:
Release notes:

- when fetching cachito sources, set build args to allow users to copy sources to the build image

PR #1351, opened 2020-01-09
container.yaml go and remote_source keys must be mutually exclusive

rcerven:
* OSBS-8482

### DIFF
--- a/atomic_reactor/schemas/container.json
+++ b/atomic_reactor/schemas/container.json
@@ -4,46 +4,142 @@
   "title": "container image build configuration",
 
   "type": ["object", "null"],
-  "properties": {
-    "go": {
-      "type": "object",
-      "properties": {
-        "modules": {
-          "type": ["array", "null"],
-          "items": {
-            "type": "object",
-            "properties": {
-              "module": {
-                "type": "string",
-                "description": "Top-level Go module (package) name which will be built",
-                "examples": [
-                  "example.com/go/packagename",
-                  "example.com/go/anotherpackage"
-                ]
+  "anyOf": [
+      {
+         "properties": {
+            "go": {
+              "type": "object",
+              "properties": {
+                "modules": {
+                  "type": ["array", "null"],
+                  "items": {
+                    "type": "object",
+                    "properties": {
+                      "module": {
+                        "type": "string",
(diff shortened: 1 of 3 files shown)


atomic_reactor/schemas/container.json
```diff
@@ -4,229 +4,376 @@
...
-              "archive": {
-                "type": "string",
-                "description": "Possibly-compressed archive containing full source code including dependencies",
-                "examples": [
-                  "anotherpackage.tar.gz"
-                ]
+              "additionalProperties": false
+            },
+            "platforms": {
+              "type": ["object", "null"],
```
twaugh: Would be nicer to put these other property schema into definitions and reference them in each alternative.

atomic_reactor/schemas/container.json
```diff
@@ -4,229 +4,376 @@
...
-            "type": "object",
-            "properties": {
-              "module": {
-                "type": "string",
-                "description": "Top-level Go module (package) name which will be built",
-                "examples": [
-                  "example.com/go/packagename",
-                  "example.com/go/anotherpackage"
-                ]
+  "anyOf": [
```
twaugh: Seems fine, I was thinking oneOf but it makes little difference in this instance.

atomic_reactor/schemas/container.json
rcerven: sure

atomic_reactor/schemas/container.json
rcerven: I can't use oneOf, because IF you have for example only Platforms with oneOf it isn't able to decide which schema variant to use, so we need anyOf

athos-ribeiro:
Would it be worth it to test all the cases here?

- both keys are invalid (already tested)
- go key only
- remote_source only

athos-ribeiro:
> Would it be worth it to test all the cases here?
> 
>     * both keys are invalid (already tested)
> 
>     * go key only
> 
>     * remote_source only

NVM, They are already tested

athos-ribeiro:
Waiving feature toggle requirement since it would not make much sense here.

osbs-docs do point to the schema file, so this is self-documented

rcerven:
rn updated: container.yaml go and remote_source keys must be mutually exclusive

PR #1354, opened 2020-01-10
log dockerfile only if builder isn't None

rcerven:
* OSBS-8397

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -619,7 +619,8 @@ def build_inside(input_method, input_args=None, substitutions=None):
     try:
         build_result = dbw.build_docker_image()
     except Exception as e:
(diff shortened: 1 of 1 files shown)


athos-ribeiro:
```
0 of 2 new or added lines in 1 file covered
```

Do we want to cover the new case?

Waiving feature toggling and docs PR since this change relates to logging data only.

rcerven:
> ```
> 0 of 2 new or added lines in 1 file covered
> ```
> 
> Do we want to cover the new case?
> 
> Waiving feature toggling and docs PR since this change relates to logging data only.

I don't think it is worth, because everything is mocked in there anyway, and this is special case when it fails in set_df_path

rcerven:
rn updated: - log Dockerfile only if builder isn't None

PR #1355, opened 2020-01-14
Allow cachito integration without koji

athos-ribeiro:
When using osbs-client directly, there will be no koji user to forward
to cachito. We allow the build to proceed by setting the cachito request
user param as 'unknown_user', which can be configured through the
reactor configmap.

* OSBS-8503
* OSBS-8375

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_remote_source.py
+++ b/atomic_reactor/plugins/pre_resolve_remote_source.py
@@ -51,6 +51,7 @@ class ResolveRemoteSourcePlugin(PreBuildPlugin):
             return
 
         user = self.get_koji_user()
+        self.log.info('Using user "%s" for cachito request', user)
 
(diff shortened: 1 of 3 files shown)


athos-ribeiro:
Marked this as a draft while we discuss if we want to allow the option to skip the user param (as presented in this PR) or if we want to get rid of that param completely.

athos-ribeiro:
`osbs-docs` already has an entry in cachito admin guide pointing to the config.json schema, which seems to be enough for this feature.

athos-ribeiro:
RN added:

- resolve_remote_source plugin allows builds to proceed without koji by passing an "unknown_user" to cachito requests

athos-ribeiro:
> LGTM
> 
> I quite don't like name `unknown_user` key, but I couldn't figure anything better

I could not think of a better name either... Let's merge it with `unknown_user` and we could, at least, change the name in the reactor config before going to prod, if we come up with something better!

PR #1356, opened 2020-01-14
Do not use cachito user representativeness feature

athos-ribeiro:
While cachito makes it possible for OSBS to query the koji user, there
is no need to set the user param to the koji build user from an OSBS
perspective. Moreover, it would have the drawback of generating two
different code paths, one for builds triggered through koji and another
one for builds triggered through osbs-client directly.

* OSBS-8375


athos-ribeiro:
This PR and #1355 are conflicting. One of the approaches should be picked.

The drawback for this approach would be losing direct tracking of the relation between koji build requester and a cachito request. Its benefit, as described in the commit message, would be to keep a single code path for osbs-client and koji builds.

athos-ribeiro:
Closing in favor of #1355

PR #1363, opened 2020-01-22
Whitelist annotation must be set as string

athos-ribeiro:
The OpenShift build annotation must be set as a string and then
converted to a list, if needed.

* OSBS-8138


twaugh:
Could this be done with an `@annotation(...)` decorator on the `ReactorConfigPlugin` class, and having its `run` method return the list? The store_metadata plugin knows to convert those things to strings.

rcerven:
> Could this be done with an `@annotation(...)` decorator on the `ReactorConfigPlugin` class, and having its `run` method return the list? The store_metadata plugin knows to convert those things to strings.

I would rather keep annotation away from reactor config map plugin

rcerven:
rn updated: Whitelist annotation must be set as string

PR #1366, opened 2020-01-31
add logging for yum repos and compose ids

mlangsdorf:
Fixes OSBS-8381

Add explicit logging of injected yum repos and inherited yum repos and compose ids.

### DIFF
--- a/atomic_reactor/plugins/pre_inject_yum_repo.py
+++ b/atomic_reactor/plugins/pre_inject_yum_repo.py
@@ -92,3 +92,5 @@ class InjectYumRepoPlugin(PreBuildPlugin):
         yum_repos = list(self.workflow.files)
         add_yum_repos_to_dockerfile(yum_repos, df, inherited_user,
                                     self.workflow.builder.base_from_scratch)
+        for repo in yum_repos:
+            self.log.info("injected yum repo: %s", repo)
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -7,6 +7,7 @@ of the BSD license. See the LICENSE file for details.
 """
 from __future__ import unicode_literals, absolute_import
 
+from copy import deepcopy
 from datetime import datetime, timedelta
 import os
 import yaml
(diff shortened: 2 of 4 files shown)


tests/plugins/test_resolve_composes.py
```diff
@@ -447,11 +447,18 @@ def test_inherit_parents(self, workflow, reactor_config_map, parent_compose, par
         else:
             assert set(nonearch_repourls) == set(expected_yum_repourls)
 
+        if allow_inherit and parent_compose:
+            print(caplog.text)
```
athos-ribeiro: Could we remove this print here (this is the reason CI if failing)?

athos-ribeiro:
Waiving docs and toggle requirements: changes to logging output only.

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -168,12 +169,20 @@ def adjust_for_inherit(self):
                            build_info['nvr'], build_info['id'])
 
         all_compose_ids = set(self.compose_ids)
+        original_compose_ids = deepcopy(all_compose_ids)
         all_compose_ids.update(parent_compose_ids)
         self.all_compose_ids = list(all_compose_ids)
+        for compose_id in all_compose_ids:
```
MartinBasti: Wouldn't be sufficient to list `parent_compose_ids` ?

atomic_reactor/plugins/pre_inject_yum_repo.py
```diff
@@ -92,3 +92,5 @@ def run(self):
         yum_repos = list(self.workflow.files)
         add_yum_repos_to_dockerfile(yum_repos, df, inherited_user,
                                     self.workflow.builder.base_from_scratch)
+        for repo in yum_repos:
+            self.log.info("injected yum repo: %s", repo)
```
MartinBasti: Do we need multi-line log entry for this?

mlangsdorf: I'm not sure, that's why I put it in a separate commit.

atomic_reactor/plugins/pre_resolve_composes.py
mlangsdorf: The parent_compose_ids might be the same as the existing ids, in which case no new compose ids are being added.

atomic_reactor/plugins/pre_inject_yum_repo.py
mlangsdorf: More specifically, I wasn't sure if injected repos should be logged at all, and if there were logged, whether it was better to log them one per line or as a group.  If they need to be logged, I suggest 1 per line for easier grepping.

tests/plugins/test_resolve_composes.py

PR #1369, opened 2020-02-05
flatpak_create_dockerfile: Improve method of disabling system repos

owtaylor:
The previous way we disabled repositories from the base image required
knowing a set of patterns for repository names injected by atomic-reactor;
but when repositories are added with the yum_repourls option, this isn't
possible - the repository names come from the source repository files.

So, instead remove all repositories from the system, and then re-add the
repositories injected by atomic-reactor via pre_inject_yum_repo.

### DIFF
--- a/atomic_reactor/plugins/pre_flatpak_create_dockerfile.py
+++ b/atomic_reactor/plugins/pre_flatpak_create_dockerfile.py
@@ -25,7 +25,7 @@ import os
 from flatpak_module_tools.flatpak_builder import FlatpakSourceInfo, FlatpakBuilder
 from osbs.repo_utils import ModuleSpec
 
-from atomic_reactor.constants import DOCKERFILE_FILENAME, YUM_REPOS_DIR
+from atomic_reactor.constants import DOCKERFILE_FILENAME, RELATIVE_REPOS_PATH, YUM_REPOS_DIR
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.plugins.pre_reactor_config import get_flatpak_base_image
 from atomic_reactor.plugins.pre_resolve_module_compose import get_compose_info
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/pre_flatpak_create_dockerfile.py
```diff
@@ -39,13 +39,20 @@
...
+# /only/ the yum repos from atomic_reactor, and nothing else.
 DOCKERFILE_TEMPLATE = '''FROM {base_image}
 
 LABEL name="{name}"
 LABEL com.redhat.component="{component}"
 LABEL version="{stream}"
 LABEL release="{version}"
 
+RUN rm -f {yum_repos_dir}*
+ADD {relative_repos_path}* '{yum_repos_dir}'
```
twaugh: Curious why only one of these gets quotes around it. Don't we want e.g. ``ADD '{relative_repos_path}'* '{yum_repos_dir}'`` etc?

owtaylor: The second line is verbatim from `pre_inject_yum_repo.py`

```python
    # Insert the ADD line at the beginning of each stage
    df.add_lines(
        "ADD %s* '%s'" % (RELATIVE_REPOS_PATH, YUM_REPOS_DIR),
        all_stages=True, at_start=True, skip_scratch=True
    )
```

And then when I added the preceding line, I just did it the way that made sense to me - not pretending that `yum_repos_dir` was ever going to have spaces in it (adding the quotes that way doesn't help for anything but the "spaces" case - you'd need to use shlex.quote() or something.) But yes, the result is a inconsistent - I'd be happy to add or remove quotes for consistency.

twaugh: I see. Let's remove them then.

owtaylor:
OK, removed the quotes, and also removed the corresponding quotes from inject_yum_repo to keep things consistent.

owtaylor:
[ repushed adding missing Signed-off-by ]

athos-ribeiro:
Release notes added:

- Flatpak builds: remove all repositories and then re-add the repositories injected through pre_inject_yum_repo to allow better handling of yum_repo_urls

PR #1370, opened 2020-02-05
flatpak_create_dockerfile: allow overriding base image

owtaylor:
The flatpak base_image ends up being used in two different ways:
 - To provide an environment to run dnf in to create the final filesystem
 - As the recorded parent image in openshift and koji

While the identity of the base image doesn't really matter at all for
the first usage, recording a parent that correctly matches a particular
branch of the distribution can help make things work correctly
automatic rebuilds, etc. For this reason, allow setting the base
image in container.yaml, instead of just in the global configuration.


owtaylor:
> It seems like https://osbs.readthedocs.io/en/latest/users.html#flatpak would also need an update.

Thanks - created: https://github.com/containerbuildsystem/osbs-docs/pull/130

athos-ribeiro:
Release notes added:

- add option to override flatpak base images in container.yaml. This is useful when working with autorebuilds

PR #1371, opened 2020-02-06
Allow specifying additional labels for Flatpaks in container.yaml

owtaylor:
If a user wants to add an additional labels on the resulting Flatpak image, like a `maintainer` label, then they need a way to do it without modifying the Dockerfile (since the Dockerfile for Flaptaks is autogenerated). 

This PR adds a new plugin `add_flatpak_labels` that reads the `labels` key from the `flatpak` section of `container.yaml` and adds a LABEL line to the resulting Dockerfile with the given labels.

Some relevant code from `add_labels_in_df` is moved into a utility function.

Corresponding patches to osbs-client and osbs-docs will be needed.

### DIFF
--- a/atomic_reactor/plugins/pre_add_flatpak_labels.py
+++ b/atomic_reactor/plugins/pre_add_flatpak_labels.py
@@ -1,0 +1,55 @@
+"""
+Copyright (c) 2020 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+
+
+Pre build plugin which adds additional labels to the Dockerfile automatically
+created for a flatpak, based on the flatpak: labels key in container.yaml.
+"""
+
+from __future__ import unicode_literals, absolute_import
+
+from atomic_reactor.plugin import PreBuildPlugin
+from atomic_reactor.util import df_parser, label_to_string
+
+
+class AddFlatpakLabelsPlugin(PreBuildPlugin):
+    key = "add_flatpak_labels"
+    is_allowed_to_fail = False
+
+    def __init__(self, tasker, workflow):
+        """
+        constructor
+
+        :param tasker: ContainerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        """
+        # call parent constructor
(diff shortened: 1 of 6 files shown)


twaugh:
I wonder if this is a good opportunity to define what properties are allowed within 'flatpak' in the schema?
https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/schemas/container.json#L247-L250

owtaylor:
> I wonder if this is a good opportunity to define what properties are allowed within 'flatpak' in the schema?

I think I avoided this earlier, because some of the parsing of keys from this section is implemented in flatpak-module-tools and I didn't want an update to flatpak-module-tools to require an update to atomic-reactor to work. But thinking "well, will give better error message", I went ahead and did the work to add all the properties supported by atomic-reactor and flatpak-module-tools to the schema. And realized some work is needed on the error printing code to realize my dream of better error messages - the anyOf in container.json results in "<contents of container.json> are not valid under any of the given schemas", and a dump of the schemas. Should be possible to do better - https://python-jsonschema.readthedocs.io/en/v1.2.0/errors.html - but I'm going to defer all of this to a separate PR.

owtaylor:
[ Repushed to fix a 'Flaptak' typo in the commit Subject ]

atomic_reactor/plugins/pre_add_flatpak_labels.py
```diff
@@ -0,0 +1,55 @@
...
+        run the plugin
+        """
+        flatpak_yaml = self.workflow.source.config.flatpak
+        if flatpak_yaml is None:
+            return
+        labels = flatpak_yaml.get('labels')
+        if not labels:
+            return
+
+        dockerfile = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
```
mlangsdorf: It seems weird to keep recreating the df_parser for all these plugins that need to manipulate the Dockerfile, and we should probably at some point just store the df_parser in tasker or workflow.

but that's beyond the scope of this PR, I just wanted to make that observation.

MartinBasti:
Release notes:
> - Additional labels can now be specified for a Flatpak image in container.yaml

PR #1373, opened 2020-02-10
Flatpak: replace resolve_module_compose with resolve_composes

owtaylor:
The separate resolve_module_compose plugin didn't support all features of the resolve_composes plugin - in particular it didn't support `compose: inherit`. Using resolve_composes for Flatpaks as well resolves this.
    
Because resolve_composes needs to run later in the plugin order, flatpak_create_dockerfile is split into two parts - flatpak_create_dockerfile to create an initial version of the Dockerfile and flatpak_update_dockerfile to fill in the details from the resolved compose.
    
In a worker build where resolve_composes doesn't run, the compose information is retrieved from ODCS - to make this work, the resolve_composes plugin is changed to override the compose_ids argument that is used when starting worker builds. (A technique taken from resolve_module_compose)
    
**NOTE**: This has not been tested in a live environment due to the complexity of the required environment, so it needs either to be merged early enough in release cycle to have time for follow-up fixes, or needs to get pre-merge testing from a branch. It's basically just refactoring current code, and there are complete pytest tests, so it should be *close* to working, but I wouldn't be surprised if there are a few issues when testing all the modules together with real data.


owtaylor:
OSBS client piece of this: https://github.com/containerbuildsystem/osbs-client/pull/934

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -402,10 +401,9 @@ def get_build(self, metadata, worker_metadatas):
 
             extra['image'].update(get_parent_image_koji_data(self.workflow))
 
-            flatpak_source_info = get_flatpak_source_info(self.workflow)
-            if flatpak_source_info is not None:
-                compose_info = get_compose_info(self.workflow)
-                koji_metadata = compose_info.koji_metadata()
+            flatpak_compose_info = get_flatpak_compose_info(self.workflow)
+            if flatpak_compose_info is not None:
```
mlangsdorf: any reason this isn't just `if flatpak_compose_info:` ?  the way you have it, `flakpak_compose_info = False` would pass the test, which it obviously shouldn't.  It probably can't be set to False anyway, but why the extra characters for a less clear form of the test?

atomic_reactor/plugins/pre_flatpak_update_dockerfile.py
```diff
@@ -0,0 +1,257 @@
...
+    def run(self):
+        """
+        run the plugin
+        """
+
+        compose_info = self._load_compose_info()
+        set_flatpak_compose_info(self.workflow, compose_info)
+
+        source = self._load_source()
+        set_flatpak_source_info(self.workflow, source)
```
mlangsdorf: since this and set_flatpak_compose_info are done unconditionally, is there any reason to not set them in `_load_*()` ?

atomic_reactor/plugins/pre_flatpak_update_dockerfile.py
```diff
@@ -0,0 +1,257 @@
...
+        # Create a file describing which packages from the base yum repositories are included
+
+        includepkgs = builder.get_includepkgs()
+        includepkgs_path = os.path.join(self.workflow.builder.df_dir, 'atomic-reactor-includepkgs')
+        with open(includepkgs_path, 'w') as f:
+            f.write('includepkgs = ' + ','.join(includepkgs) + '\n')
+
+        # Create the cleanup script
+
+        cleanupscript = os.path.join(self.workflow.builder.df_dir, "cleanup.sh")
```
mlangsdorf: Can cleanup.sh be a constant instead of a repeated magic variable string?

atomic_reactor/plugins/exit_koji_import.py
owtaylor: My personal thinking is that if you are expecting a variable to hold None or an object then 'if object is not None: return object.member' catches a violation of your expectation when a variable ends up with 0/''/[]/{}/False, rather than silently letting it go. But yes, it is a fair bit of verbosity to catch a rare mistake and it does seem like   disproportionate number of the uses of 'is not None' in the atomic-reactor codebase are mine :-)

atomic_reactor/plugins/pre_flatpak_update_dockerfile.py
owtaylor: You could - but then I don't like having _load_source() both set the source info and have a return value, so I switched to
```python
        self._load_source()
        source = get_flatpak_source_info(self.workflow, source)
```
(which is more like the other places in the code where the source_info is retrieved)

atomic_reactor/plugins/pre_flatpak_update_dockerfile.py
owtaylor: Sure, added:

owtaylor:
Will push a new version shortly - didn't try to comprehensively evaluate my usage of 'is not None' in the code, but removed the one that @mlangsdorf pointed out, and the same one in `exit_koji_promote`.

MartinBasti:
code looks good

rcerven:
apart from 'is not None' occurrences looks good

MartinBasti:
Release notes:
> - Flatpak builds properly inherit parent composes

PR #1374, opened 2020-02-12
Parse raw dependency replacements string

athos-ribeiro:
The dependency_replacements param should be passed to the
resolve_remote_source plugin as a string containing the cachito data
separated by ':'

* OSBS-8139

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_remote_source.py
+++ b/atomic_reactor/plugins/pre_resolve_remote_source.py
@@ -33,11 +33,37 @@ class ResolveRemoteSourcePlugin(PreBuildPlugin):
         """
         :param tasker: ContainerTasker instance
         :param workflow: DockerBuildWorkflow instance
+        :param dependency_replacements: list<str>, dependencies for the cachito fetched artifact to
+        be replaced. Must be of the form pkg_manager:name:version[:new_name]
         """
         super(ResolveRemoteSourcePlugin, self).__init__(tasker, workflow)
         self._cachito_session = None
         self._osbs = None
-        self._dependency_replacements = dependency_replacements
+        self._dependency_replacements = self.parse_dependency_replacements(dependency_replacements)
+
+    def parse_dependency_replacements(self, replacement_strings):
+        """Parse dependency_replacements param and return cachito-reaady dependency replacement dict
+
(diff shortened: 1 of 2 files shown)


athos-ribeiro:
Docs PR: https://github.com/containerbuildsystem/osbs-docs/pull/131

atomic_reactor/plugins/pre_resolve_remote_source.py
```diff
@@ -33,11 +33,37 @@ def __init__(self, tasker, workflow, dependency_replacements=None):
...
+
+        param replacement_strings: list<str>, pkg_manager:name:version[:new_name]
+        return: list<dict>, cachito formated dependency replacements param
+        """
+        if not replacement_strings:
+            return
+
+        dependency_replacements = []
+        for dr_str in replacement_strings:
+            pkg_manager, name, version, new_name = (dr_str.split(':', 3) + [None] * 4)[:4]
```
MartinBasti: Couldn't just osbs-client pass this as JSON?

ben-alkov: I (think I) understand what this does, but I don't understand why...

athos-ribeiro: I did think about dealing with it in osbs-client (the benefit would be earlier feedback), although it doesn't feel like this should be osbs-client's responsibility. In this case, I'd rather keep the current format.

If the idea would be to pass a JSON string to the CLI, it would not be much user friendly. If we go down this path, we'd need to check if rpkg people would feel OK with that as well. Unless OSBS client does handle the code above, translating the string to JSON.

athos-ribeiro: The string that comes from osbs-client could be any string. For instance 'foo'. Cachito requires at least these 3 params above to be passed to the dependency replacement option.

MartinBasti: I find osbs-client the best place to replace custom CLI passed representation to JSON/dict. So we have to deal with that only in one place and everywhere else it works with standard formats. However if this may cause complications to koji-c, rhpkg/fedpkg I'm fine with parsing it in atomic-reactor.

OK json as param may be too much :)

PR #1379, opened 2020-02-21
Change tests & travis for correct working with Fedora 31

SalatskySal:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,13 @@ services:
 env:
   matrix:
     - ACTION=bandit
-      OS=fedora
-      OS_VERSION=29
+      OS=centos
+      OS_VERSION=7
       PYTHON_VERSION=2
       ENGINE=docker
     - ACTION=bandit
       OS=fedora
-      OS_VERSION=29
+      OS_VERSION=31
       PYTHON_VERSION=3
       ENGINE=docker
     - ACTION=pylint
@@ -24,7 +24,7 @@ env:
       ENGINE=docker
     - ACTION=pylint
       OS=fedora
-      OS_VERSION=29
+      OS_VERSION=31
       PYTHON_VERSION=3
       ENGINE=docker
     - OS=centos
@@ -32,15 +32,11 @@ env:
       PYTHON_VERSION=2
       ENGINE=docker
     - OS=fedora
-      OS_VERSION=29
-      PYTHON_VERSION=2
-      ENGINE=docker
-    - OS=fedora
-      OS_VERSION=29
+      OS_VERSION=30
       PYTHON_VERSION=3
       ENGINE=docker
     - OS=fedora
-      OS_VERSION=30
+      OS_VERSION=31
       PYTHON_VERSION=3
       ENGINE=docker
 install:
--- a/atomic_reactor/plugins/pre_koji_parent.py
+++ b/atomic_reactor/plugins/pre_koji_parent.py
@@ -33,8 +33,6 @@ DEFAULT_POLL_INTERVAL = 10  # 10 seconds
 
 class KojiParentBuildMissing(ValueError):
     """Expected to find a build for the parent image in koji, did not find it within timeout."""
-    pass
-
 
 class KojiParentPlugin(PreBuildPlugin):
     """Wait for Koji build of parent images to be available
--- a/test.sh
+++ b/test.sh
@@ -12,9 +12,10 @@ IMAGE="$OS:$OS_VERSION"
 if [[ $OS == "fedora" ]]; then
   IMAGE="registry.fedoraproject.org/$IMAGE"
 fi
-engine_mounts="-v $PWD:$PWD:z"
+# Use arrays to prevent globbing and word splitting
+engine_mounts=(-v "$PWD":"$PWD":z)
 for dir in ${EXTRA_MOUNT:-}; do
-  engine_mounts="${engine_mounts} -v $dir:$dir:z"
+  engine_mounts=("${engine_mounts[@]}" -v "$dir":"$dir":z)
 done
 
 CONTAINER_NAME="atomic-reactor-$OS-$OS_VERSION-py$PYTHON_VERSION"
@@ -40,11 +41,11 @@ else
 fi
 
 # Create or resurrect container if needed
-if [[ $($ENGINE ps -qa -f name=$CONTAINER_NAME | wc -l) -eq 0 ]]; then
-  $ENGINE run --name $CONTAINER_NAME -d $engine_mounts -w $PWD -ti $IMAGE sleep infinity
-elif [[ $($ENGINE ps -q -f name=$CONTAINER_NAME | wc -l) -eq 0 ]]; then
+if [[ $($ENGINE ps -qa -f name="$CONTAINER_NAME" | wc -l) -eq 0 ]]; then
+  $ENGINE run --name "$CONTAINER_NAME" -d "${engine_mounts[@]}" -w "$PWD" -ti "$IMAGE" sleep infinity
+elif [[ $($ENGINE ps -q -f name="$CONTAINER_NAME" | wc -l) -eq 0 ]]; then
   echo found stopped existing container, restarting. volume mounts cannot be updated.
-  $ENGINE container start $CONTAINER_NAME
+  $ENGINE container start "$CONTAINER_NAME"
 fi
 
 if [[ $OS == "centos" ]]; then
--- a/tests/plugins/test_build_source_container.py
+++ b/tests/plugins/test_build_source_container.py
@@ -15,6 +15,7 @@ from flexmock import flexmock
 import pytest
 import json
 import tarfile
+import re
 
 from atomic_reactor.constants import PLUGIN_FETCH_SOURCES_KEY
 from atomic_reactor.inner import DockerBuildWorkflow
@@ -187,7 +188,8 @@ def test_running_build(tmpdir, caplog, sources_dir, sources_dir_exists, sources_
         build_result = runner.run()
         err_msg = "No SRPMs directory '{}' available".format(sources_dir_path)
         err_msg += "\nNo Remote source directory '{}' available".format(remote_dir_path)
-        assert err_msg in caplog.text
+        # Since Python 3.7 logger adds additional whitespaces by default -> checking without them
+        assert re.sub(r'\s+', " ", err_msg) in re.sub(r'\s+', " ", caplog.text)
         assert build_result.is_failed()
 
     elif export_failed:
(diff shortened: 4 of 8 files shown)


.travis.yml
```diff
@@ -9,40 +9,36 @@ env:
   matrix:
     - ACTION=bandit
       OS=fedora
-      OS_VERSION=29
-      PYTHON_VERSION=2
-      ENGINE=docker
-    - ACTION=bandit
```
MartinBasti: @athos-ribeiro I'm not sure about this, is it worth to keep bandit for py2 as well? I vote for dropping bandit on py2

.travis.yml
```diff
@@ -9,40 +9,36 @@ env:
...
-    - OS=fedora
-      OS_VERSION=29
+      OS_VERSION=31
       PYTHON_VERSION=3
       ENGINE=docker
     - OS=fedora
       OS_VERSION=30
       PYTHON_VERSION=3
       ENGINE=docker
+    - ACTION=bandit
```
MartinBasti: Please move this to original position at the top of document

tests/test_util.py
```diff
@@ -104,13 +104,6 @@
 }
 
 
-class CustomTestException(Exception):
```
MartinBasti: Nice!

tests/test_cachito_util.py
```diff
@@ -63,8 +63,8 @@ def handle_request_sources(http_request):
     response = api.request_sources(CACHITO_REQUEST_REPO, CACHITO_REQUEST_REF, **additional_params)
     assert response['id'] == CACHITO_REQUEST_ID
 
-    response_json = json.dumps(response_data, indent=4)
-    assert 'Cachito response:\n{}'.format(response_json) in caplog.text
+    response_json = 'Cachito response:\n{}'.format(json.dumps(response_data, indent=4))
+    assert "".join(response_json.split()) in "".join(caplog.text.split())
```
MartinBasti: Could be this written in more readable way?
```
>>> "something with many          whitespaces".replace(" ", "")                                                                                                                                                                            
'somethingwithmanywhitespaces'
```

Please put a comment above, why this change is needed.

MartinBasti:
I don't think that this PR deserves 13 commits. Could you please rebase and merge commits into 1 or 2 ?

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -33,8 +33,8 @@
 
 class KojiParentBuildMissing(ValueError):
     """Expected to find a build for the parent image in koji, did not find it within timeout."""
-    pass
-
+    def __init__(self, error):
```
ben-alkov: I don't understand how this change relates to the rest of this PR...

MartinBasti: New pylint screams. However I'm not sure if this fix is correct.

why not just remove `pass` without adding `__init__`?

tests/test_cachito_util.py
SalatskySal: Do I need to put it above each changing?

MartinBasti: IMHO Yes

ben-alkov: IMHO, **definitely**; I didn't even realize what was going on here until I saw the comment :slightly_smiling_face:

tests/test_util.py
```diff
@@ -58,10 +58,7 @@
                                  read_yaml, read_yaml_from_file_path, OSBSLogs,
                                  get_platforms_in_limits, get_orchestrator_platforms,
                                  dump_stacktraces, setup_introspection_signal_handler,
-                                 allow_repo_dir_in_dockerignore,
```
MartinBasti: :eyes:

tests/test_cachito_util.py
```diff
@@ -192,10 +196,11 @@ def handle_wait_for_request(http_request):
     expect_in_logs = dedent(
         """\
         Request {} is in "{}" state: Unknown
-        Details: {}
+         Details: {}
```
MartinBasti: Is this change needed?

tests/test_util.py
```diff
@@ -1645,49 +1635,3 @@ def test_allow_repo_dir_in_dockerignore(tmpdir, dockerignore_exists):
...
-    ),
-    (
-        ['com.redhat.delivery.operator.bundle=meh',
-         'com.redhat.delivery.appregistry=meh'],
-        [],
-        [has_operator_bundle_manifest, has_operator_appregistry_manifest],
-    )
-
-])
-def test_has_operator_manifest(tmpdir, workflow, labels, f_true, f_false):
```
MartinBasti: This shouldn't be removed

chmeliik:
Just realized, that symlink command should have been `$RUN ln -sf /usr/bin/python3 /usr/bin/python`, my bad (otherwise the command is executed on your PC instead of inside the container)

But if the current solution fixes it that would be better

tests/test_util.py
```diff
@@ -1646,7 +1639,6 @@ def test_allow_repo_dir_in_dockerignore(tmpdir, dockerignore_exists):
         assert ignore_lines[0:len(ignore_content)] == ignore_content
         assert ignore_lines[-1] == added_lines
 
-
```
MartinBasti: Unwanted change

tests/test_util.py
```diff
@@ -1670,7 +1662,6 @@ def test_allow_repo_dir_in_dockerignore(tmpdir, dockerignore_exists):
         [],
         [has_operator_bundle_manifest, has_operator_appregistry_manifest],
     )
-
```
MartinBasti: Unwanted change

MartinBasti:
Could we just run this as to avoid such error:
```python
py_bin = "python3" if six.PY3 else "python"
subprocess.check_call([python_bin, "setup.py", "sdist", "--dist-dir", tmpdir])
```

tests/test_cachito_util.py
SalatskySal: Done

tests/test_util.py
SalatskySal: Fixed

tests/test_util.py
SalatskySal: Fixed

SalatskySal:
Thanks you all for your help. It was important for me

.travis.yml
```diff
@@ -14,35 +14,31 @@ env:
...
       ENGINE=docker
     - OS=fedora
       OS_VERSION=30
       PYTHON_VERSION=3
       ENGINE=docker
+    - ACTION=pylint
+      OS=fedora
+      OS_VERSION=29
+      PYTHON_VERSION=2
+      ENGINE=docker
```
chmeliik: Could this be moved next to the other `pylint` action? It would be nice to have them grouped together.

test.sh
```diff
@@ -67,6 +67,10 @@ if [[ $OS == "fedora" ]]; then
   if [[ $PYTHON_VERSION == 2* ]]; then
     $RUN $PKG $ENABLE_REPO install -y python-backports-lzma
   fi
+
+  if [[ $PYTHON_VERSION == 3* ]]; then
+    $RUN ln -sf /usr/bin/python3 /usr/bin/python
+  fi
```
chmeliik: @SalatskySal have you tried the solution @MartinBasti suggested? That could work too, and you wouldn't need the symlink in the container.

See https://github.com/containerbuildsystem/atomic-reactor/blob/e669a655fde97e213fe72c426fec5390bb7ccce8/atomic_reactor/buildimage.py#L117

If that doesn't work, at least add a comment like `'python' command is expected to be available in container`

.travis.yml
athos-ribeiro: I am wondering if we really want this here. Fedora 29 Has reached its EOL in the end of 2019. IMHO, we should just stop running the CI in F29. If this means no more py2 tests in Fedora, so be it. We can keep running py2 tests in CentoOS7. What do you think?

.travis.yml
athos-ribeiro: Giving this a second thought,  we should drop F29 completely. Would it be possible to run bandit for py2 somewhere else? e.g., CentOS7? We could try to push the package to EPEL if it is not there :)

athos-ribeiro:
Could you add a reference for an OSBS ticket in the commit message? See other commit messages in this project for reference.

tests/plugins/test_build_source_container.py
```diff
@@ -187,7 +187,8 @@ def check_check_output(args, **kwargs):
         build_result = runner.run()
         err_msg = "No SRPMs directory '{}' available".format(sources_dir_path)
         err_msg += "\nNo Remote source directory '{}' available".format(remote_dir_path)
-        assert err_msg in caplog.text
+        # Since Python 3.7 logger adds additional whitespaces by default -> checking without them
+        assert err_msg.replace(" ", "") in caplog.text.replace(" ", "")
```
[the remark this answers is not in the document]
SalatskySal: The error message remains the same, but we create new strings without white spaces and check them, so it shouldn't make misunderstandings

test.sh
SalatskySal: @chmeliik Oh, sorry, I didn`t see his solution. @MartinBasti Thanks =)

.travis.yml
MartinBasti: I'd keep it there until we defer py2 support. It's just linting and newer versions of pylint stopped supporting py2. So F29 is the most recent we can use for py2 linting

tests/plugins/test_build_source_container.py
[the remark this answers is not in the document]
athos-ribeiro: What I mean here is that, if the assertion fails, the error message will be harder to read for the reader.

For instance, we would see something like this:

```
AssertionError: assert 'NoSRPMsdirectory/sourcesavailable\nNoRemotesourcedirectory/remotesourcesavailable' in NoSRPMsdirectory/sourcesavailable\nNoRemotesourcedirectory/remotesourceavailable'
```

This is also error prone since it is verifying a modified version of the stings and not the actual strings as you said. We are creating new strings and checking the newly created strings instead of checking whatever output comes from the plugin being tested.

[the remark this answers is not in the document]
SalatskySal: @athos-ribeiro You`re right, I didn`t understand you :) 
It will be better, if I change those messages by regex.

.travis.yml
chmeliik: Agreed, the package is installed using pip anyway so it should work for CentOS too

MartinBasti:
no release notes needed, CI change only

PR #1380, opened 2020-02-21
Operator utils module

chmeliik:
* OSBS-8636

TODO:
- [x] Some TODOs in code and specfile

### DIFF
--- a/atomic_reactor/operator_util.py
+++ b/atomic_reactor/operator_util.py
@@ -8,6 +8,7 @@ of the BSD license. See the LICENSE file for details.
 
 from __future__ import absolute_import, unicode_literals
 
+import os
 import logging
 
 from ruamel.yaml import YAML
@@ -20,6 +21,15 @@ yaml = YAML()
 log = logging.getLogger(__name__)
 
 
+OPERATOR_CSV_KIND = "ClusterServiceVersion"
+
+
+class NotOperatorCSV(Exception):
+    """
+    Data is not from a valid ClusterServiceVersion document
+    """
+
+
 class OperatorCSV(object):
     """
     A single ClusterServiceVersion file in an operator manifest.
@@ -28,6 +38,14 @@ class OperatorCSV(object):
     """
 
     def __init__(self, path, data):
+        """
+        Initialize an OperatorCSV
+
+        :param path: Path where data was found or where it should be written
+        :param data: ClusterServiceVersion yaml data
+        """
+        if data.get("kind") != OPERATOR_CSV_KIND:
+            raise NotOperatorCSV("Not a ClusterServiceVersion")
         self.path = path
         self.data = data
 
@@ -173,3 +191,47 @@ class OperatorCSV(object):
             # Doesn't matter if string was not a pullspec, it will simply not match anything
             # in replacement_pullspecs and no replacement will be done
             self._replace_pullspec(obj, k_or_i, replacement_pullspecs, log_template)
+
+
+class OperatorManifest(object):
+    """
+    A collection of operator files.
+
+    Currently, only ClusterServiceVersion files are considered relevant.
+    """
+
+    def __init__(self, files):
+        """
+        Initialize an OperatorManifest
+
+        :param files: list of OperatorCSVs
+        """
+        self.files = files
+
+    @classmethod
+    def from_directory(cls, path):
+        """
+        Make an OperatorManifest from all the relevant files found in
+        a directory (or its subdirectories)
+
+        :param path: Path to directory
+        :return: OperatorManifest
+        """
+        yaml_files = cls._get_yaml_files(path)
+        operator_csvs = list(cls._get_csvs(yaml_files))
+        return cls(operator_csvs)
+
+    @classmethod
+    def _get_yaml_files(cls, dirpath):
+        for d, _, files in os.walk(dirpath):
+            for f in files:
+                if f.endswith(".yaml") or f.endswith(".yml"):
+                    yield os.path.join(d, f)
+
+    @classmethod
+    def _get_csvs(cls, yaml_files):
+        for f in yaml_files:
+            try:
+                yield OperatorCSV.from_file(f)
+            except NotOperatorCSV:
+                pass
--- a/atomic_reactor/operator_util.py
+++ b/atomic_reactor/operator_util.py
@@ -1,0 +1,175 @@
+"""
+Copyright (c) 2020 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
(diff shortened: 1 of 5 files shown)


atomic_reactor/operator_util.py
```diff
@@ -0,0 +1,233 @@
...
+        Make an OperatorCSV from a file
+
+        :param path: Path to file
+        :return: OperatorCSV
+        """
+        with open(path) as f:
+            data = yaml.load(f)
+            return cls(path, data)
+
+    def write(self):
```
MartinBasti: maybe `dump` may express it better

atomic_reactor/operator_util.py
```diff
@@ -0,0 +1,233 @@
...
+            for f in files:
+                if f.endswith(".yaml") or f.endswith(".yml"):
+                    yield os.path.join(d, f)
+
+    @classmethod
+    def _get_csvs(cls, yaml_files):
+        for f in yaml_files:
+            try:
+                yield OperatorCSV.from_file(f)
+            except ValueError:
```
MartinBasti: Can we raise specific error, like `NotOperatorCSVFile`

atomic_reactor/operator_util.py
```diff
@@ -0,0 +1,233 @@
...
+        self._replace_env_pullspecs(replacement_pullspecs)
+        self._replace_container_image_pullspecs(replacement_pullspecs)
+
+    def replace_pullspecs_everywhere(self, replacement_pullspecs):
+        """
+        Replace all pullspecs found anywhere in data
+
+        :param replacement_pullspecs: mapping of pullspec -> replacment
+        """
+        log_template = "{path} - Replaced pullspec: {old} -> {new}"
```
MartinBasti: Why? Can we have just constant `LOG_TEMPLATE = "%s - Replaced pullspec: %s --> %s"` without interpolation adn conditions?

atomic_reactor/operator_util.py
```diff
@@ -0,0 +1,233 @@
...
+            for e in c.get("env", []) if e["name"].startswith("RELATED_IMAGE_")
+        ]
+        for env in envs:
+            if "valueFrom" in env:
+                msg = '{}: "valueFrom" references are not supported'.format(env["name"])
+                raise RuntimeError(msg)
+        return envs
+
+    def _get_pullspec(self, obj, key, log_template=None):
+        # TODO: better name than obj?
```
MartinBasti: fine by me

atomic_reactor/operator_util.py
chmeliik: I thought it would be nice to log where exactly a pullspec was found/replaced, but I did manage to get rid of string interpolation FWIW

atomic_reactor/operator_util.py
```diff
@@ -0,0 +1,238 @@
...
+        deployments_path = ("spec", "install", "spec", "deployments")
+        deployments = chain_get(self.data, deployments_path, default=[])
+
+        containers_path = ("spec", "template", "spec", "containers")
+        containers = [
+            c for d in deployments for c in chain_get(d, containers_path, default=[])
+        ]
+        return containers
+
+    def _get_envs(self):
```
MartinBasti: maybe `get_related_image_envs` ?

chmeliik:
Fixed specfile (ruamel package on RHEL)

atomic_reactor/operator_util.py
chmeliik: Renamed, also realized test for "valueFrom" in env definition was missing, added

chmeliik:
Removed resolved TODO

chmeliik:
Added more details in commit messages

atomic_reactor/operator_util.py
```diff
@@ -0,0 +1,237 @@
...
+        return containers
+
+    def _get_related_image_envs(self):
+        containers = self._get_containers()
+        envs = [
+            e for c in containers
+            for e in c.get("env", []) if e["name"].startswith("RELATED_IMAGE_")
+        ]
+        for env in envs:
+            if "valueFrom" in env:
```
rcerven: I think we should just ignore those, but not failing

@MartinBasti  wdyt ?

MartinBasti: Hmm, yup we should just ignore it. Good catch

atomic_reactor/util.py
```diff
@@ -1611,3 +1611,12 @@ def get_platform_config(platform, build_annotations):
     cm_key = cm_key_tmp[cmlen:]
 
     return cm_key, cm_frag_key
+
+
+def chain_get(d, keys, default=None):
+    for k in keys:
+        try:
+            d = d[k]
```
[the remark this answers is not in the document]
chmeliik: @ben-alkov; This function is shamelessly stolen from osbs-client (with minor changes). I think you have the right idea about what it does, but I'll try to give an example just to be sure:

Consider the following yaml file:
```yaml
spec:
  relatedImages:
    - name: foo
      image: registry/foo:1
```
Using `jq`, you would query for the image pullspec like this (after converting from yaml to json)
```bash
$ cat file.json | jq '.spec.relatedImages[0].image'
"registry/foo:1"
```
With this function, you can achieve the same thing for a python dict:
```python
>>> chain_get(data, ["spec", "relatedImages", 0, "image"])
"registry/foo:1"

# basically the same as
data["spec"]["relatedImages"][0]["image"]
# but returns None if any of the lookups would fail
```

> Also, single letter variable names :rage:

`d` as a variable name for "this is a dictionary with futher unspecified content" is pretty common, but it's true that the meaning may change inside the loop (sometimes, it may be a `list`, at the end, it will usually be a `str`, `int`, `float` or `bool`). I could rename to something that better reflects that, but best I can think of is `obj` for object

atomic_reactor/operator_util.py
chmeliik: Should we? It is still a `RELATED_IMAGE_*` environment variable with some pullspec, we just do not know what that pullspec is. Is it ok for that to be in the final operator manifest?

Alternatively, I could try to implement the `valueFrom` logic, I haven't looked at the specification (if there is any) but it shouldn't be impossibly difficult.

chmeliik:
Added a docstring to `chain_get` and improved variable naming

atomic_reactor/operator_util.py
MartinBasti: Well, it's reference we cannot catch everything. But on other hand we can set this as requirement to not use references for that in OSBS and play more safe.

atomic_reactor/util.py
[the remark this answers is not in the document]
ben-alkov: @chmeliik; Very clear, thanks!

PR #1384, opened 2020-02-27
when getNextRelease from koji fails, find next release

rcerven:
* OSBS-8652

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -102,7 +102,12 @@ class BumpReleasePlugin(PreBuildPlugin):
     def get_next_release_standard(self, component, version):
         build_info = {'name': component, 'version': version}
         self.log.debug('getting next release from build info: %s', build_info)
-        next_release = self.get_patched_release(self.xmlrpc.getNextRelease(build_info))
+        try:
+            next_release = self.get_patched_release(self.xmlrpc.getNextRelease(build_info))
+        # when release can't be bumped via koji's getNextRelease (unsupported format,
+        # eg. multiple dots, strings etc) it will raise exception
+        except koji.BuildError:
+            next_release = self.get_next_release(build_info)
 
         # getNextRelease will return the release of the last successful build
         # but next_release might be a failed build. Koji's CGImport doesn't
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -102,7 +102,10 @@ def next_release_general(self, component, version, release, release_label,
     def get_next_release_standard(self, component, version):
         build_info = {'name': component, 'version': version}
         self.log.debug('getting next release from build info: %s', build_info)
-        next_release = self.get_patched_release(self.xmlrpc.getNextRelease(build_info))
+        try:
+            next_release = self.get_patched_release(self.xmlrpc.getNextRelease(build_info))
+        except koji.BuildError:
```
athos-ribeiro: Could we add a comment on why we would reach a koji buildError here?

athos-ribeiro:
LGTM

It would be nice to add more explanation to the reasoning behind the patch. For instance, we could add some description of how isolated builds could be an issue in the commit message

atomic_reactor/plugins/pre_bump_release.py
rcerven: added

rcerven:
added explanation to the commit message

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -138,6 +143,21 @@ def get_next_release_append(self, component, version, base_release, base_suffix=
                     return next_release
             suffix += 1
 
+    def get_next_release(self, build_info):
+        queryopts = {'order': '-build.id', 'limit': 1}
+        # release is either single number or with decimal point, so we can easily bump next release
+        # pylint: disable=W1401
+        search_str = '^{}-{}-\d+(\.\d+)?$'.format(build_info['name'], build_info['version'])
```
chmeliik: Is there a reason not to use an r-string here? That should get rid of the pylint warning.

rcerven: added 'r'

ktdreyer:
If we update koji's `getNextRelease` to incorporate the changes that you have here, can we drop this from OSBS?

PR #1385, opened 2020-03-02
add_filesystem plugin uses also repos from composes

rcerven:
* OSBS-8719

### DIFF
--- a/atomic_reactor/plugins/pre_add_filesystem.py
+++ b/atomic_reactor/plugins/pre_add_filesystem.py
@@ -26,7 +26,8 @@ import json
 import re
 import os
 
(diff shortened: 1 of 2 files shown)


athos-ribeiro:
Waiving the feature toggle requiremtent: this is a new default behavior and disabling it would not make much sense IMO.

Do we want to make it explicit in the docs that we do use both the yum repos and the composes when building base images? It seems we did not have the former behavior documented, so there's nothing to update unless we want to add these details in the docs.

rcerven:
doc pr:
https://github.com/containerbuildsystem/osbs-docs/pull/133

rcerven:
rn updated: add_filesystem plugin uses also repos from composes

PR #1389, opened 2020-03-05
operators: Add base of pin_operator_digest plugin

chmeliik:
* OSBS-8637

The remainder of this plugin will be implemented in the future, once the
supporting code in osbs-client is ready.


atomic_reactor/schemas/container.json
```diff
@@ -109,6 +109,7 @@
               "additionalProperties": false,
               "required": ["repo", "ref"]
             },
+            "operator_manifest": {"$ref": "#/definitions/operator_manifest"},
```
chmeliik: Why does `container.json` have 2 anyOf schemas? Is this the right place for `operator_manifests`?

twaugh: Because either 'go' or 'remote_source' may be used, but not both.

chmeliik: I see, thanks

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        Initialize pin_operator_digests plugin
+
+        :param tasker: ContainerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param replacement_pullspecs: Dict[str, str], computed in orchestrator,
+                                      provided to workers by osbs-client
+        """
+        super(PinOperatorDigestsPlugin, self).__init__(tasker, workflow)
+        # TODO: container.yaml IS available in worker builds, right?
```
MartinBasti: yes

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -0,0 +1,73 @@
...
+            return True
+        else:
+            self.log.info("Not an operator manifest bundle build, skipping plugin")
+            return False
+
+    def run_in_orchestrator(self):
+        try:
+            self.site_config = get_operator_manifests(self.workflow)
+        except KeyError:
+            raise RuntimeError("operator_manifests configuration missing in reactor config map")
```
MartinBasti: For BW compatibility, we should not fail if config is missing and build is not an operator bundle build (i.e. let it fail with no allowed_registries when building operator bundle)

chmeliik: If build is not an operator bundle build this plugin should not run at all (see `should_run()`). Is that what you meant?

chmeliik: Updated tests to make it a bit more obvious

MartinBasti: oh right, blind me

chmeliik:
Added code to set param for worker build

chmeliik:
Also added worker param to user_params schema

MartinBasti:
`plugins.md` needs also updates

chmeliik:
Oops, name of plugin was supposed to be singular

chmeliik:
Release notes will be added when plugin is finished

PR #1393, opened 2020-03-11
Implement pin_operator_digest

chmeliik:
* OSBS-8637

Add the code that does what pin_operator_digest claims to do

Also add a minor fix for OperatorManifest (fail if path is not a directory)

### DIFF
--- a/atomic_reactor/plugins/pre_pin_operator_digest.py
+++ b/atomic_reactor/plugins/pre_pin_operator_digest.py
@@ -8,11 +8,14 @@ of the BSD license. See the LICENSE file for details.
 
 from __future__ import absolute_import, print_function
 
+import os.path
+
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.constants import PLUGIN_PIN_OPERATOR_DIGESTS_KEY
-from atomic_reactor.util import has_operator_bundle_manifest
+from atomic_reactor.util import has_operator_bundle_manifest, get_manifest_digests, ImageName
 from atomic_reactor.plugins.pre_reactor_config import get_operator_manifests
 from atomic_reactor.plugins.build_orchestrate_build import override_build_kwarg
+from atomic_reactor.operator_util import OperatorManifest
 
 
 class PinOperatorDigestsPlugin(PreBuildPlugin):
@@ -50,6 +53,9 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
         self.replacement_pullspecs = replacement_pullspecs
 
     def run(self):
+        """
+        Run pin_operator_digest plugin
+        """
         if self.should_run():
             if self.is_in_orchestrator():
                 return self.run_in_orchestrator()
@@ -57,6 +63,11 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
                 return self.run_in_worker()
 
     def should_run(self):
+        """
+        Determine if this is an operator manifest bundle build
+
+        :return: bool, should plugin run?
+        """
         if has_operator_bundle_manifest(self.workflow):
             return True
         else:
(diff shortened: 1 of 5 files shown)


chmeliik:
Still a bunch of TODOs to resolve, most of them are not super significant except for the one about registry auth/insecure configuration

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -50,25 +53,188 @@ def __init__(self, tasker, workflow, replacement_pullspecs=None):
...
+        """
+        Replace image tag with manifest list digest
+
+        :param image: ImageName
+        :return: ImageName
+        """
+        if image.tag.startswith("sha256:"):
+            return image
+        # TODO: authentication? insecure?
+        digests = get_manifest_digests(image, image.registry)
```
chmeliik: Tested with example operator bundle build with registry-proxy pullspecs, fails TLS handshake

chmeliik:
yaml formats things as it pleases, instead of checking raw file content first load it back and check the data

chmeliik:
Nevermind, apparently, with an old enough PyYAML, `yaml.safe_load(yaml.dump(data))` is not guaranteed to work ¯\\\_(ツ)\_/¯

Maybe ruamel.yaml will work

chmeliik:
Ah, the mythical extra whitespace on a new enough Python. Turns out you can get the log text without extra indentation (also without the stuff added by the logger) from `caplog.records`

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -50,25 +53,188 @@ def __init__(self, tasker, workflow, replacement_pullspecs=None):
...
+        if self.user_config is None:
+            raise RuntimeError("operator_manifest configuration missing in container.yaml")
+
+        manifests_dir = os.path.join(self.workflow.source.path, self.user_config["manifests_dir"])
+        self.log.info("Looking for operator CSV files in %s", manifests_dir)
+
+        operator_manifest = OperatorManifest.from_directory(manifests_dir)
+
+        if operator_manifest.files:
+            # TODO: log paths in a particular order? make sure OperatorManifest orders them?
```
MartinBasti: Does order matter? If you want to make them easy to read, feel free to sort it.

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -50,25 +53,188 @@ def __init__(self, tasker, workflow, replacement_pullspecs=None):
...
+        Initialize a PullspecReplacer
+
+        :param user_config: container.yaml operator_manifest configuration
+        :param site_config: reactor-config-map operator_manifests configuration
+        """
+        self.allowed_registries = site_config["allowed_registries"]
+        if self.allowed_registries is not None:
+            self.allowed_registries = set(self.allowed_registries)
+
+        # TODO: what if a replacement registry is not in allowed_registries?
```
MartinBasti: It's fine then (but weird). There may be use case to prevent maintainer use some target registries in manifests.

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -50,25 +53,187 @@ def __init__(self, tasker, workflow, replacement_pullspecs=None):
...
+    def pin_digest(self, image):
+        """
+        Replace image tag with manifest list digest
+
+        :param image: ImageName
+        :return: ImageName
+        """
+        if image.tag.startswith("sha256:"):
+            return image
+        # TODO: authentication? insecure?
```
MartinBasti: Ignore this for one now, we will resolve auth together with support for mutliple registries

atomic_reactor/plugins/pre_pin_operator_digest.py
chmeliik: It was mostly just to make log output deterministic, but I got around that in unit tests so it's probably fine

atomic_reactor/plugins/pre_pin_operator_digest.py
chmeliik: Sounds reasonable :+1:

chmeliik:
Addressed comments, also made plugin log a warning and abort instead of failing when reactor-config-map configuration is missing

atomic_reactor/plugins/pre_pin_operator_digest.py
chmeliik: Probably just missing cert in osbs-box as @MartinBasti pointed out

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -50,25 +53,186 @@ def __init__(self, tasker, workflow, replacement_pullspecs=None):
...
+    def pin_digest(self, image):
+        """
+        Replace image tag with manifest list digest
+
+        :param image: ImageName
+        :return: ImageName
+        """
+        if image.tag.startswith("sha256:"):
+            return image
+        digests = get_manifest_digests(image, image.registry)
```
rcerven: we are sure that those image tags we are replacing aren't manifest_lists ??
because as of now this would always use just x86_64 v2 digest

MartinBasti: Oooh, good catch, we need `'v2_list'`

chmeliik: Fixed, thank you!

chmeliik:
Release notes: Implemented first version of pin_operator_digest (supports digest pinning, registry replacement)

PR #1402, opened 2020-03-17
Implement repo name replacements

chmeliik:
* OSBS-8637

### DIFF
--- a/atomic_reactor/plugins/pre_pin_operator_digest.py
+++ b/atomic_reactor/plugins/pre_pin_operator_digest.py
@@ -10,9 +10,15 @@ from __future__ import absolute_import, print_function
 
 import os.path
 
+from osbs.utils import Labels
+
+from atomic_reactor import util
 from atomic_reactor.plugin import PreBuildPlugin
-from atomic_reactor.constants import PLUGIN_PIN_OPERATOR_DIGESTS_KEY
-from atomic_reactor.util import has_operator_bundle_manifest, get_manifest_digests, ImageName
+from atomic_reactor.constants import PLUGIN_PIN_OPERATOR_DIGESTS_KEY, INSPECT_CONFIG
+from atomic_reactor.util import (has_operator_bundle_manifest,
+                                 get_manifest_digests,
+                                 read_yaml_from_file_path,
+                                 ImageName)
 from atomic_reactor.plugins.pre_reactor_config import get_operator_manifests
 from atomic_reactor.plugins.build_orchestrate_build import override_build_kwarg
 from atomic_reactor.utils.operator import OperatorManifest
@@ -26,6 +32,8 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
     - finds container pullspecs in operator ClusterServiceVersion files
     - computes replacement pullspecs:
         - replaces tags with manifest list digests
+        - replaces repos (and namespaces) based on operator_manifests.repo_replacements
+          configuration in container.yaml and r-c-m*
         - replaces registries based on operator_manifests.registry_post_replace in r-c-m*
 
     When running in a worker:
@@ -155,9 +163,11 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
 
         for original in pullspecs:
             pinned = replacer.pin_digest(original)
-            replaced = replacer.replace_registry(pinned)
-            if replaced != original:
-                replacements[original] = replaced
+            repo_replaced = replacer.replace_repo(pinned)
+            registry_replaced = replacer.replace_registry(repo_replaced)
+
+            if registry_replaced != original:
+                replacements[original] = registry_replaced
 
         replacement_lines = "\n".join(
             "{} -> {}".format(p, replacements[p]) if p in replacements
@@ -197,6 +207,22 @@ class PullspecReplacer(object):
             for registry in site_config.get("registry_post_replace", [])
         }
 
+        self.package_mapping_files = {
+            mapping["registry"]: mapping["package_mappings_file"]
+            for mapping in site_config.get("repo_replacements", [])
+        }
+        # Mapping of [file path => package mapping]
+        # Loaded when needed, see _get_site_mapping
+        self.file_package_mappings = {}
+
+        self.user_package_mappings = {
+            mapping["registry"]: mapping["package_mappings"]
+            for mapping in user_config.get("repo_replacements", [])
+        }
+        # Final package mapping that you get by combining site mapping with user mapping
+        # Loaded when needed, see _get_final_mapping
+        self.final_package_mappings = {}
+
     def registry_is_allowed(self, image):
         """
         Is image registry allowed in OSBS config?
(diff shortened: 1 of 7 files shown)


chmeliik:
@MartinBasti some issues identified with repo_replacements, PTAL (tests will fail because I don't know what behaviour is expected)

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -229,10 +249,70 @@ def replace_registry(self, image):
...
+
+        if mapping is not None or user_mapping is not None:
+            mapping = mapping or {}
+            user_mapping = user_mapping or {}
+
+            # Mapping file is [package => list of repos], user mapping is [package => repo],
+            # convert to [package => repo] where possible
+            for repo, replacements in mapping.items():
+                if len(replacements) == 1:
+                    mapping[repo] = replacements[0]
```
MartinBasti: I'm not a fan of mixing various types in dict

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -229,10 +249,70 @@ def replace_registry(self, image):
...
+            raise RuntimeError(
+                "repo_replacements configured for registry but repo is missing: {}".format(image)
+            )
+        elif isinstance(replacement, list):
+            replacements = ", ".join(replacement)
+            raise RuntimeError(
+                "Multiple options found when replacing repo in {}: {}".format(image, replacements)
+            )
+
+        # TODO: replacement can (probably should) be namespace/repo
```
MartinBasti: Replacement is package name --> namespace/repo. We should strip everything from original url.
Package name must be taken from LABELs in image (com.redhat.bzcomponent or so), so registry has to be queried for that.

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -229,10 +249,70 @@ def replace_registry(self, image):
...
 
+    # TODO: rename, because it can also replace namespace
+    def replace_repo(self, image):
+        """
+        Replace image repo based on OSBS site/user configuration for image registry
+
+        :param image: ImageName
+        :return: ImageName
+        """
+        package_mapping = self._get_package_mapping(image.registry)
```
MartinBasti: We should avoid constructing this mapping in every call. This should be just one time operation.

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -229,10 +249,70 @@ def replace_registry(self, image):
...
+        if package_mapping is None:  # repo_replacements not configured for registry
+            return image
+
+        replacement = package_mapping.get(image.repo)
+
+        if replacement is None:
+            raise RuntimeError(
+                "repo_replacements configured for registry but repo is missing: {}".format(image)
+            )
+        elif isinstance(replacement, list):
```
MartinBasti: can we stick with one type?

atomic_reactor/schemas/container.json
```diff
@@ -254,6 +254,31 @@
...
+                "type": "object",
+                "patternProperties": {
+                  ".*": {"type": "string"}
+                },
+                "examples": [
+                  {"foo-bar": "foo/bar", "spam-eggs": "spam/eggs"}
+                ]
+              }
+            },
+            "required": ["registry", "package_mappings"],
```
MartinBasti: It shouldn't be required

tests/plugins/test_pin_operator_digests.py
```diff
@@ -294,15 +329,15 @@ def test_orchestrator(self, docker_tasker, tmpdir, caplog):
             'To be replaced:\n'
             'keep-registry/ns/foo:latest -> keep-registry/ns/foo@sha256:abcdef\n'
             'keep-registry/ns/spam@sha256:123456 - no change\n'
-            'replace-registry/ns/bar:1 -> new-registry/ns/bar@sha256:fedcba\n'
-            'replace-registry/ns/eggs@sha256:654321 -> new-registry/ns/eggs@sha256:654321'
+            'replace-registry/ns/bar:1 -> new-registry/ns/baz@sha256:fedcba\n'
```
MartinBasti: Maybe `replace-registry/ns/bar:1 -> new-registry/baz@sha256:fedcba` should be included too

atomic_reactor/plugins/pre_pin_operator_digest.py
chmeliik: That should already be the case, `_get_package_mapping()` saves the constructed mapping in `self.package_mapping_by_registry` and takes it from there on subsequent calls

atomic_reactor/schemas/container.json
chmeliik: I think it should. The `repo_replacements` key itself is not required, or it can be `repo_replacements: []`, but if you put an item in this list, that item should have both `registry` and `package_mappings`, or it does not make much sense

atomic_reactor/plugins/pre_pin_operator_digest.py
chmeliik: ok, will use list everywhere

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -229,10 +249,70 @@ def replace_registry(self, image):
             return image
         return self._replace(image, registry=self.registry_replace[image.registry])
 
+    # TODO: rename, because it can also replace namespace
```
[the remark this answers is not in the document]
chmeliik: Then it's fine I guess

chmeliik:
Fixed having more than one type in package mapping dict, more fixes coming soon

atomic_reactor/schemas/container.json
MartinBasti: ah right

chmeliik:
TODO:
* [ ] Inspect image, find package name, get replacement for that (not the repo part of pullspec)
* [ ] When user overrides replacement for a package that is also configured in config-map, check that the replacement is one of the options in config-map
* [ ] Add logging to PullspecReplacer (it does a lot of stuff now)

chmeliik:
Rebased on master

chmeliik:
Post-rebase fix

chmeliik:
Added check that when user overrides replacement, it is one of the options in config-map

TODO: logging

atomic_reactor/schemas/container.json
```diff
@@ -254,6 +254,31 @@
           "description": "Relative path to directory containing operator manifests",
           "type": "string",
           "pattern": "^[^/].*$"
+        },
+        "repo_replacements": {
```
athos-ribeiro: After https://github.com/containerbuildsystem/osbs-client/pull/938, it seems we would also need a change in osbs-client here. TBH, I was actually expecting we would stop checking this schema in atomic reactor.

chmeliik: Thanks! I even had the branch ready, just forgot to open the PR.

> TBH, I was actually expecting we would stop checking this schema in atomic reactor.

Yeah same, I think @MartinBasti  already has plans to do something about that.

chmeliik:
osbs-client copy of container.json changes: https://github.com/containerbuildsystem/osbs-client/pull/958

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -226,10 +267,108 @@ def replace_registry(self, image):
...
+        if mapping_file is None:
+            return None
+        elif mapping_file in self.file_package_mappings:
+            return self.file_package_mappings[mapping_file]
+
+        mapping = read_yaml_from_file_path(mapping_file, "schemas/package_mapping.json")
+        self.file_package_mappings[mapping_file] = mapping
+        return mapping
+
+    def _get_package(self, image):
```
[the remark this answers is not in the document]
chmeliik: That sounds better, thanks!

chmeliik:
> Are we updating osbs-docs later, when releasing the feature as a whole?

Precisely!

chmeliik:
Release notes: pin_operator_digest plugin now supports repo name replacements

PR #1405, opened 2020-03-19
don't suppress the 'koji' plugin for module-only-composes

owtaylor:
If resolve_composes was run to only compose modules, it would
still suppress the 'koji' plugin which handles adding the koji target
build tag repo to container builds. But modules assume that there's a base
package set coming from somewhere else. So, if we have a compose that only
involves modules, with no packages, pulp repos, inherited repourls, or
user-specified repourls, set the new force_tag_repo(*) parameter on the
client builds, to avoid suppressing the 'koji' plugin.

(*) https://github.com/containerbuildsystem/osbs-client/pull/956

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -109,6 +109,7 @@ class ResolveComposesPlugin(PreBuildPlugin):
         self.composes_info = None
         self._parent_signing_intent = None
         self.repourls = repourls or []
+        self.has_complete_repos = len(self.repourls) > 0
         self.plugin_result = self.workflow.prebuild_results.get(PLUGIN_KOJI_PARENT_KEY)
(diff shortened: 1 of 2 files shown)


MartinBasti:
So far LGTM, please address CI errors

rcerven:
LGTM, but tests need fixing

owtaylor:
> No action is needed for the the failing tests (see [containerbuildsystem/osbs-client#957](https://github.com/containerbuildsystem/osbs-client/pull/957)). They should pass when the fixes for the linter warnings are pushed

Thanks! - was scratching my head how my changes could have anything to do with that failure.

MartinBasti:
Stickler linter errors seems unrelated to osbs-client, right?

MartinBasti:
Release notes:
> * don't skip `koji` plugin when no repos are defined/inherited for flatpak builds to work with modules which assume that there is a package set coming from somewhere else

PR #1407, opened 2020-03-27
 operator manifests: Read mapping files from URLs 

chmeliik:
* OSBS-8639

### DIFF
--- a/atomic_reactor/plugins/pre_pin_operator_digest.py
+++ b/atomic_reactor/plugins/pre_pin_operator_digest.py
@@ -229,13 +229,13 @@ class PullspecReplacer(object):
             for registry in site_config.get("registry_post_replace", [])
         }
 
-        self.package_mapping_files = {
-            mapping["registry"]: mapping["package_mappings_file"]
+        self.package_mapping_urls = {
+            mapping["registry"]: mapping["package_mappings_url"]
             for mapping in site_config.get("repo_replacements", [])
         }
-        # Mapping of [file path => package mapping]
+        # Mapping of [url => package mapping]
         # Loaded when needed, see _get_site_mapping
-        self.file_package_mappings = {}
+        self.url_package_mappings = {}
 
         self.user_package_mappings = {
             mapping["registry"]: mapping["package_mappings"]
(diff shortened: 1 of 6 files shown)


atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -313,7 +313,8 @@ def _get_site_mapping(self, registry):
         elif mapping_file in self.file_package_mappings:
             return self.file_package_mappings[mapping_file]
 
-        mapping = read_yaml_from_file_path(mapping_file, "schemas/package_mapping.json")
+        self.log.debug("Downloading mapping file for %s", registry)  # TODO: safe to log URL?
```
chmeliik: Not sure about this, WDYT?

MartinBasti: It should be fine, we have nothing to hide :D  (unless URL contains password, but we are good engineers and we don't do that)

chmeliik:
Added URL to log message

Also hopefully fixed python2 failure, and removed no longer used test param

atomic_reactor/util.py
```diff
@@ -1269,6 +1270,18 @@ def read_yaml_from_file_path(file_path, schema):
...
+def read_yaml_from_url(url, schema):
+    session = get_retrying_requests_session()
+    resp = session.get(url, stream=True)
+
+    f = io.BytesIO()
+    for chunk in resp.iter_content(chunk_size=DEFAULT_DOWNLOAD_BLOCK_SIZE):
+        f.write(chunk)
+
+    f.seek(0)
+    return read_yaml(f.read().decode(), schema)
```
MartinBasti: To be consistent py2/3 shouldn't we enforce decode to use utf8?

tests/plugins/test_pin_operator_digests.py
```diff
@@ -130,16 +132,20 @@ def mock_operator_csv(tmpdir, filename, pullspecs):
...
     for registry, mapping in repo_replacements.items():
-        filename = 'mapping-{}.yaml'.format(registry)
-        path = tmpdir.join(filename)
-        with open(str(path), 'w') as f:
-            yaml.dump(mapping, f)
-        repo_replacements[registry] = str(path)
+        url = 'https://somewhere.net/mapping-{}.yaml'.format(registry)
+
+        # ruamel.yaml does not support dumping to str, use an io stream
+        f = io.BytesIO()
```
MartinBasti: Why BytesIO and not StringIO?

atomic_reactor/util.py
chmeliik: Huh, utf8 is not the default on python2? TIL

chmeliik:
Rebased on master, will address comments next

tests/plugins/test_pin_operator_digests.py
chmeliik: Originally because the `iter_content()` of a response is bytes, but turns out you can make it unicode with the `decode_unicode` param

chmeliik:
Now using StringIO instead of BytesIO for the util method. In test, BytesIO is still neede because ruamel.yaml refuses to write to StringIO on python2.

chmeliik:
Also added missing `raise_for_status()` in util method

chmeliik:
Release notes not needed, fix to make https://github.com/containerbuildsystem/atomic-reactor/pull/1402 (not yet released) usable

PR #1409, opened 2020-03-27
operator manifests: Handle pullspecs in order

chmeliik:
It is expected that users will have to update their configuration in
response to OSBS errors relatively frequently. If the order of pullspec
replacement is not deterministic, that will be a major pain.

### DIFF
--- a/atomic_reactor/plugins/pre_pin_operator_digest.py
+++ b/atomic_reactor/plugins/pre_pin_operator_digest.py
@@ -140,12 +140,16 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
         return operator_manifest
 
     def _get_pullspecs(self, operator_manifest):
-        pullspecs = set()
+        pullspec_set = set()
         for operator_csv in operator_manifest.files:
-            pullspecs.update(operator_csv.get_pullspecs())
+            pullspec_set.update(operator_csv.get_pullspecs())
+
+        # Make sure pullspecs are handled in a deterministic order
+        # ImageName does not implement ordering, use str() as key for sorting
+        pullspecs = sorted(pullspec_set, key=str)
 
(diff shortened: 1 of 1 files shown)


chmeliik:
Not related to a ticket, discovered while checking logs from unit tests

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -140,12 +140,16 @@ def _get_operator_manifest(self):
...
-            pullspecs.update(operator_csv.get_pullspecs())
+            pullspec_set.update(operator_csv.get_pullspecs())
+
+        # Make sure pullspecs are handled in a deterministic order
+        # ImageName does not implement ordering, use str() as key for sorting
+        pullspecs = sorted(pullspec_set, key=str)
 
         if pullspecs:
-            pullspec_lines = "\n".join(sorted(map(str, pullspecs)))
+            pullspec_lines = "\n".join(map(str, pullspecs))
```
rcerven: why are we converting to str? shouldn't pullspec be always string anyway?

chmeliik: They are `ImageName` because that's what the `utils.operator` module works with - it might make more sense to work with strings and only convert to `ImageName` when needed, but at this point it's probably not worth changing

rcerven: ok ic, wondering if

"\n".join((imname.to_str() for imname in pullspecs))
wouldn't be more explaining

chmeliik: Good point, updated

chmeliik:
Release notes not needed, expected behaviour of to-be-released plugin

PR #1410, opened 2020-03-30
improve read_yaml between osbs-client and atomic-reactor

mlangsdorf:
OSBS-8786

Remove most of the logic in atomic-reactor's read_yaml() and instead call the osbs-client version of the function with 'atomic_reactor' as the package argument.  Keep the existing read_yaml() as an alias to avoid making a bunch of minor changes in the files.

Remove container.json from atomic-reactor so that only one copy will exist in osbs-client.

Requires https://github.com/containerbuildsystem/osbs-client/pull/960


atomic_reactor/util.py
```diff
@@ -1280,48 +1278,15 @@ def read_yaml_from_url(url, schema):
...
     f.seek(0)
-    return read_yaml(f.read(), schema)
+    return osbs_read_yaml(f.read(), schema, package)
 
 
-def read_yaml(yaml_data, schema):
+def read_yaml(yaml_data, schema, package='atomic_reactor'):
     """
     :param yaml_data: string, yaml content
+    :param schema: string, path name to the schema file within the atomic-reactor package
```
athos-ribeiro: Could we also decribe 'package' here?

mlangsdorf:
We can't have osbs-client use a schema in atomic-reactor, because that will create a circular dependency.  We don't want to have container.json in two places.  I don't see any other solution than to have it osbs-client and require that changes to container.json get merged there.

atomic_reactor/util.py

mlangsdorf:
Not adding anything to release notes as this is an internal organization change.

PR #1411, opened 2020-03-31
operator manifests: Add OCP 4.4 workaround

chmeliik:
* OSBS-8640

Replace pullspecs everywhere, not just in locations in which they are
expected to be found.

### DIFF
--- a/atomic_reactor/plugins/pre_pin_operator_digest.py
+++ b/atomic_reactor/plugins/pre_pin_operator_digest.py
@@ -119,7 +119,9 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
 
         for operator_csv in operator_manifest.files:
             self.log.info("Replacing pullspecs in %s", operator_csv.path)
-            operator_csv.replace_pullspecs(replacement_pullspecs)
+            # Replace pullspecs everywhere, not just in locations in which they
+            # are expected to be found - OCP 4.4 workaround
+            operator_csv.replace_pullspecs_everywhere(replacement_pullspecs)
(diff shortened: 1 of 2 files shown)


chmeliik:
It is theoretically possible to configure OSBS in a way such that
`replacement_pullspecs` will have `{A: B, B: C}`.

Doing both `replace_pullspecs()` and `replace_pullspecs_everywhere()`
would then result in `A -> C`.

Example:
```yaml
operator_manifests:
  allowed_registries:
    - r1
    - r2
  registry_post_replace:
    - old: r1
      new: r2
    - old: r2
      new: r3
```
```yaml
kind: ClusterServiceVersion
spec:
  relatedImages:
    - name: foo1
      image: r1/foo@sha256:123456  # A
    - name: foo2
      image: r2/foo@sha256:123456  # B
```

I would consider this user / OSBS maintainer error, but in the interest
of keeping the same behaviour, use only `replace_pullspecs_everywhere()`
for now. The only downside should be less informative DEBUG messages.

Not sure if it would be worth checking such cases in code.

MartinBasti:
> I would consider this user / OSBS maintainer error, but in the interest
> of keeping the same behaviour, use only `replace_pullspecs_everywhere()`
> for now. The only downside should be less informative DEBUG messages.
> 
> Not sure if it would be worth checking such cases in code.

Not worth checking. That workaround will be removed when OCP4.4 goes EOL

chmeliik:
Release notes: pin_operator_digest will now replace pullspecs in entire CSV file (OCP 4.4 workaroud)

PR #1415, opened 2020-04-01
operator manifests: Allow pinning opt-out

chmeliik:
* OSBS-8789

Users can configure whether to enable digest pinning and/or repo
replacement during operator manifest bundle builds in container.yaml.

### DIFF
--- a/atomic_reactor/plugins/pre_pin_operator_digest.py
+++ b/atomic_reactor/plugins/pre_pin_operator_digest.py
@@ -165,6 +165,12 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
 
     def _get_replacement_pullspecs(self, pullspecs):
         self.log.info("Computing replacement pullspecs")
+
+        pin_digest, replace_repo, replace_registry = self._are_features_enabled()
+        if not any([pin_digest, replace_repo, replace_registry]):
+            self.log.warning("All replacement features disabled, skipping")
+            return {}
+
         replacer = PullspecReplacer(user_config=self.user_config, site_config=self.site_config)
 
         for p in pullspecs:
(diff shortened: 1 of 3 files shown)


chmeliik:
Depends on https://github.com/containerbuildsystem/osbs-client/pull/962

chmeliik:
Rebased

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -171,24 +171,35 @@ def _get_replacement_pullspecs(self, pullspecs):
...
 
-            self.log.debug("Replacing namespace/repo")
-            repo_replaced = replacer.replace_repo(pinned)
+            if replace_repo:
+                self.log.debug("Replacing namespace/repo")
+                replaced = replacer.replace_repo(replaced)
 
             self.log.debug("Replacing registry")
-            registry_replaced = replacer.replace_registry(repo_replaced)
+            replaced = replacer.replace_registry(replaced)
```
MartinBasti, line 197: I'm thinking if we should also add option for disabling registry replacements too

chmeliik: Would make sense. If all the replacements are disabled, should the plugin still check allowed_registries or just be skipped completely?

MartinBasti: it makes sense to skip it then, yup

chmeliik:
Added option to disable registry replacements. If no replacements are enabled, the part where plugin computes replacements will be skipped. Technically, the plugin wouldn't have to run at all, but it can still serve as a sort of sanity check (is `manifests_dir` set properly, are all the CSV files and their pullspecs found etc.)

rcerven:
code looks good, just test need fix

chmeliik:
Restarted Travis build, the tests were dependent on the osbs-client PR (which is now merged)

chmeliik:
> It would be nice to have the feature and tests in the same commit though (not a blocker for me).

Tests related to the atomic-reactor change are in the same commit. The 2nd commit is tests for new container.yaml entries added in an osbs-client PR, I'd rather keep it separate.

chmeliik:
Release notes: pin_operator_digest now allows pullspec replacement opt-out via the `enable_{digest_pinning,repo_replacements,registry_replacements}` options in container.yaml

PR #1420, opened 2020-04-03
Handle deprecated signing intents

athos-ribeiro:
To handle deprecated signing intents, we add a new deprecated_keys to the odcs configmap option. This deprecated keys will be used to infer the name of a signing intent based on its former keys. Given we are able to obtain a signing intent name, we can proceed to request a compose to be renewed with (and only with) the current signing keys in that signing intent (without the deprecated keys).

For the signing intent substitution to be respected, ODCS to be deployed with https://pagure.io/odcs/pull-request/362. Builds will not fail and the sigkeys passed to ODCS will just be ignored otherwise.

### DIFF
--- a/atomic_reactor/plugins/pre_reactor_config.py
+++ b/atomic_reactor/plugins/pre_reactor_config.py
@@ -17,6 +17,7 @@ from atomic_reactor.util import (read_yaml, read_yaml_from_file_path,
                                  get_build_json, DefaultKeyDict)
 from osbs.utils import RegistryURI
 
+import logging
 import os
 import six
 
@@ -494,6 +495,8 @@ class ODCSConfig(object):
     """
 
     def __init__(self, signing_intents, default_signing_intent):
+        self.log = logging.getLogger(self.__class__.__name__)
+
         self.default_signing_intent = default_signing_intent
 
         self.signing_intents = []
@@ -521,12 +524,25 @@ class ODCSConfig(object):
         if isinstance(keys, six.text_type):
             keys = keys.split()
         keys = set(keys)
+        intents_matching_deprecated_keys = []
         for entry in reversed(self.signing_intents):
             keys_set = set(entry['keys'])
             if (keys and keys_set >= keys) or keys == keys_set:
                 return entry
 
-        raise ValueError('unknown signing intent keys "{}"'.format(keys))
+            permissive_keys_set = set(entry['keys'] + entry.get('deprecated_keys', []))
+            if keys and permissive_keys_set >= keys:
+                intents_matching_deprecated_keys.append(entry)
+
+        if not intents_matching_deprecated_keys:
+            raise ValueError('unknown signing intent keys "{}"'.format(keys))
+
+        self.log.warning(
+            'signing intent keys "%s" contain deprecated entries in the "%s" signing intent',
+            keys,
+            intents_matching_deprecated_keys[0]['name']
+         )
+        return intents_matching_deprecated_keys[0]
 
 
 class ReactorConfigPlugin(PreBuildPlugin):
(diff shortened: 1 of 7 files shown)


rcerven:
code looks good

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -521,12 +524,25 @@ def get_signing_intent_by_keys(self, keys):
         if isinstance(keys, six.text_type):
             keys = keys.split()
         keys = set(keys)
+        deprecated_intent_matches = []
```
rcerven, line 527: rename it to something else, since it basically contains intents with correct keys, which was based on keys including deprecated

maybe something like

intents_for_deprecated_keys

athos-ribeiro: Done

atomic_reactor/utils/odcs.py
```diff
@@ -120,7 +120,8 @@ def renew_compose(self, compose_id):
         :return: dict, status of compose being renewed.
         """
         logger.info("Renewing compose %d", compose_id)
-        response = self.session.patch('{}composes/{}'.format(self.url, compose_id))
+        response = self.session.patch('{}composes/{}'.format(self.url, compose_id),
+                                      {"sigkeys": sigkeys})
```
athos-ribeiro, line 124: Since this is optional in ODCS, I will change the behavior to only pass sigkeys in case they are changed to avoid ODCS to assume "unsigned"

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -287,7 +287,19 @@ def wait_for_composes(self):
...
+                    self.log.info('Updating signing keys in "%s" from "%s", to "%s" in compose '
+                                  '"%s" due to sigkeys deprecation',
+                                  updated_signing_intent['name'],
+                                  sigkeys,
+                                  updated_signing_intent['keys'],
+                                  compose_info['id']
+                                  )
+                    sigkeys = updated_signing_intent['keys']
+
+                compose_info = self.odcs_client.renew_compose(compose_id, sigkeys)
```
rcerven, line 302: this should also handle old odcs which doesn't accept sigkeys

athos-ribeiro: It does already. ODCS will just ignore the sigkey param if it is provided to old ODCS instances.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -521,12 +524,25 @@ def get_signing_intent_by_keys(self, keys):
         if isinstance(keys, six.text_type):
             keys = keys.split()
         keys = set(keys)
+        intents_matching_deprecated_keys = []
         for entry in reversed(self.signing_intents):
             keys_set = set(entry['keys'])
             if (keys and keys_set >= keys) or keys == keys_set:
```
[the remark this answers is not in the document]
athos-ribeiro: This was written this way to handle the case where `keys` is the empty set (`unsigned` signing intent). Otherwise, any signing intent would match.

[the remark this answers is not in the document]
ben-alkov: Derp!

atomic_reactor/plugins/pre_resolve_composes.py
rcerven: ahh ok

athos-ribeiro:
Docs PR: https://github.com/containerbuildsystem/osbs-docs/pull/137

athos-ribeiro:
Release notes added:

- Allow signing intent keys deprecation to support ODCS composes renewal with different key sets

PR #1421, opened 2020-04-05
Set proper image_id for flatpak builds

athos-ribeiro:
The builder image_id is used to set extra.docker.id in the koji archive
metadata. For regular builds, the image id is the built image config
digest. For flatpak builds, this is set as the image config id for the
file system built by the docker_api builder. This is not the final image
config digest since flatpak will build a new OCI image using the
aforementioned file system. This patch sets the image_id to the OCI
image config built by flatpak.

The actual image_id should not change in the case the image is converted
from OCI to docker upon a registry push (given the JSON file order is
kept). The OCI image config and docker image config specs are
compatible.

* OSBS-8780

### DIFF
--- a/atomic_reactor/plugins/prepub_flatpak_create_oci.py
+++ b/atomic_reactor/plugins/prepub_flatpak_create_oci.py
@@ -12,6 +12,9 @@ and turns it into a Flatpak application or runtime.
 
 from __future__ import absolute_import
 
+import json
+import subprocess
+
 from flatpak_module_tools.flatpak_builder import FlatpakBuilder, FLATPAK_METADATA_ANNOTATIONS
 
 from atomic_reactor.constants import IMAGE_TYPE_OCI, IMAGE_TYPE_OCI_TAR
@@ -93,6 +96,12 @@ class FlatpakCreateOciPlugin(PrePublishPlugin):
             self.log.info("Cleaning up docker container")
             self.tasker.remove_container(container_id)
 
+    def _get_oci_image_id(self, oci_path):
+        cmd = ['skopeo', 'inspect', '--raw', 'oci:{}'.format(oci_path)]
+        raw_manifest = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+        oci_image_manifest = json.loads(raw_manifest)
+        return oci_image_manifest['config']['digest']
+
     def run(self):
         source = get_flatpak_source_info(self.workflow)
         if source is None:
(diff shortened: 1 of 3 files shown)


atomic_reactor/plugins/prepub_flatpak_create_oci.py
```diff
@@ -93,6 +96,12 @@ def _export_filesystem(self):
             self.log.info("Cleaning up docker container")
             self.tasker.remove_container(container_id)
 
+    def _get_oci_image_id(self, oci_path):
+        cmd = ['skopeo', 'inspect', '--raw', 'oci:{}'.format(oci_path)]
+        raw_manifest = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+        oci_image_manifest = json.loads(raw_manifest)
```
owtaylor, line 102: It wasn't obvious to me initially that this was encoding safe for both python2 and python3, but it is - the output of check_output is a str/bytes, and json.loads() handles that with a default-encoding of UTF-8 in both cases.

It wouldn't be hard to do this without skopeo: you have to load the <oci_path>/index.json and find the manifest id in there, then find that under blobs, but certainly using skopeo simpifies, and since we're already depending on skopeo for pushing, the extra dependency isn't really a problem.

tests/plugins/test_flatpak_create_oci.py
```diff
@@ -364,9 +364,10 @@ class MockSource(object):
 
 
 class X(object):
-    image_id = "xxx"
-    source = MockSource()
-    base_image = ImageName(repo="qwe", tag="asd")
+    def __init__(self):
```
owtaylor, line 367: If you are changing this, maybe rename X to MockBuilder - I originally copied this from other tests, but I think the use of X for a classname is pretty confusing and unreadable.

tests/plugins/test_flatpak_create_oci.py
```diff
@@ -595,7 +602,9 @@ def stream_to_generator(s):
             runner.run()
         assert expected_exception in str(ex.value)
     else:
+        assert workflow.builder.image_id == 'xxx'
         runner.run()
+        assert re.match(r'^sha256:\w{64}$', workflow.builder.image_id) is not None
```
owtaylor, line 607: I got the impression from a review discussion a few weeks ago that the consensus was `is not None` should be avoided in cases that it isn't necessary?

tests/plugins/test_flatpak_create_oci.py
athos-ribeiro: Fixed

tests/plugins/test_flatpak_create_oci.py
athos-ribeiro: You are right. It is definitely not necessary here. At first, I thought this would improve test failures output, but it ends up I was wrong.

atomic_reactor/plugins/prepub_flatpak_create_oci.py
athos-ribeiro: The reason I went with skopeo here instead of checking the oci_path contents was that `index.json` is an OCI image index (like a docker manifest list). It could have multiple entries and it does not have a `platform` key I could query to fetch the right image manifest. Since I lack understanding on the flatpak side (could more entries be added to the index for some reason?), I decided to go with skopeo.

athos-ribeiro:
Release notes added:

- Fix Flatpak koji archives image ID. These image IDs were being set as the digest for a temporary image used to build the final OCI. We now set the ID to be the digest of the final OCI config.

Not including a docs PR since this is a bug fix.

PR #1428, opened 2020-04-15
Fix built Flatpak OCI workflow

athos-ribeiro:
After a flatpak OCI is built, it should be pushed into the internal container storage for further analysis (built images are inspected for metadata). This fixes an issue introduced in 917335d, where atomic-reactor would try to inspect the image and fail since the image would not be available.
    
Since the filesystem image used to build the OCI is not being tracked anymore, this image would never get removed in the end of the build process. Therefore, we align the image for removal and proper clean up.

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -32,9 +32,11 @@ from atomic_reactor.plugin import (
     PrePublishPluginsRunner,
 )
 from atomic_reactor.source import get_source_instance_for, DummySource
-from atomic_reactor.constants import INSPECT_ROOTFS, INSPECT_ROOTFS_LAYERS
 from atomic_reactor.constants import (
     CONTAINER_DEFAULT_BUILD_METHOD,
+    DOCKER_STORAGE_TRANSPORT_NAME,
+    INSPECT_ROOTFS,
+    INSPECT_ROOTFS_LAYERS,
     PLUGIN_BUILD_ORCHESTRATE_KEY
 )
 from atomic_reactor.util import ImageName, exception_message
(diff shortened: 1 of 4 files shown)


atomic_reactor/inner.py
```diff
@@ -383,6 +385,7 @@ def __init__(self, image, source=None, prebuild_plugins=None, prepublish_plugins
         self.built_image_inspect = None
         self.layer_sizes = []
         self.default_image_build_method = CONTAINER_DEFAULT_BUILD_METHOD
+        self.storage_transport = DOCKER_STORAGE_TRANSPORT_NAME
```
rcerven, line 388: I don't think we need this, just use constant directly in that plugin

athos-ribeiro: I thought having it in the Docker workflow would serve as a reminder to change it whenever we start porting things to use buildah. Otherwise, this could easily be overlooked and we would only realize the issue after flatpak builds start failing.

rcerven: ok, let's leave it here then

atomic_reactor/plugins/prepub_flatpak_create_oci.py
```diff
@@ -124,10 +138,22 @@ def run(self):
...
         self.workflow.builder.image_id = image_id
 
+        labels = Labels(df_labels)
+        _, image_name = labels.get_name_and_value(Labels.LABEL_TYPE_NAME)
+        _, image_version = labels.get_name_and_value(Labels.LABEL_TYPE_VERSION)
+        _, image_release = labels.get_name_and_value(Labels.LABEL_TYPE_RELEASE)
+
+        name = '{}-{}'.format(self.key, image_name)
+        tag = '{}-{}'.format(image_version, image_release)
+        self._copy_oci_to_local_storage(outfile, name, tag)
```
lcarva, line 155: It would be nice to have a comment explaining why we don't need to also `defer_removal` of the image in docker storage.

athos-ribeiro: Done. I also added a docstring to the new method

athos-ribeiro:
Release notes:

Edited the entry about fixing the flatpak image ids to say:

- Fix Flatpak koji archives metadata. The metadata was being set with data from a temporary image used to build the final OCI. We now set the metadata to reference of the final OCI image. This includes the Image ID and the layer sizes.

PR #1431, opened 2020-04-17
Remove ODCS fallback config

tkdchen:

### DIFF
--- a/atomic_reactor/plugins/build_orchestrate_build.py
+++ b/atomic_reactor/plugins/build_orchestrate_build.py
@@ -329,12 +329,8 @@ class OrchestrateBuildPlugin(BuildStepPlugin):
         self.config_kwargs['koji_hub'] = koji_map['hub_url']
         self.config_kwargs['koji_root'] = koji_map['root_url']
 
-        odcs_fallback = {
-            'api_url': self.config_kwargs.get('odcs_url'),
-            'insecure': self.config_kwargs.get('odcs_insecure')
-        }
-        odcs_map = get_odcs(self.workflow, odcs_fallback)
-        self.config_kwargs['odcs_url'] = odcs_map['api_url']
+        odcs_map = get_odcs(self.workflow, {})
+        self.config_kwargs['odcs_url'] = odcs_map.get('api_url')
         self.config_kwargs['odcs_insecure'] = odcs_map.get('insecure', False)
 
         smtp_fallback = {
--- a/tests/plugins/test_orchestrate_build.py
+++ b/tests/plugins/test_orchestrate_build.py
@@ -369,7 +369,6 @@ def test_orchestrate_build(tmpdir, caplog, config_kwargs,
             ReactorConfig(reactor_dict)
 
         expected_kwargs['smtp_host'] = None
-        expected_kwargs['odcs_insecure'] = None
         expected_kwargs['koji_hub'] = None
         expected_kwargs['smtp_email_domain'] = None
         expected_kwargs['smtp_from'] = None
@@ -377,6 +376,7 @@ def test_orchestrate_build(tmpdir, caplog, config_kwargs,
         expected_kwargs['source_registry_uri'] = None
         expected_kwargs['yum_proxy'] = None
         expected_kwargs['odcs_url'] = None
+        expected_kwargs['odcs_insecure'] = False
         expected_kwargs['smtp_additional_addresses'] = ''
         expected_kwargs['koji_root'] = None
         expected_kwargs['smtp_error_addresses'] = ''
@@ -416,6 +416,7 @@ def test_orchestrate_build(tmpdir, caplog, config_kwargs,
      .and_return(None))
 
     (flexmock(Configuration).should_call('__init__').with_args(**expected_kwargs).once())
+
     build_result = runner.run()
     assert not build_result.is_failed()
 
--- a/atomic_reactor/plugins/pre_flatpak_update_dockerfile.py
+++ b/atomic_reactor/plugins/pre_flatpak_update_dockerfile.py
@@ -116,7 +116,7 @@ class FlatpakUpdateDockerfilePlugin(PreBuildPlugin):
         self.compose_ids = compose_ids
 
     def _load_composes(self):
-        odcs_client = get_odcs_session(self.workflow, fallback=NO_FALLBACK)
+        odcs_client = get_odcs_session(self.workflow)
         self.log.info(odcs_client)
 
         composes = []
--- a/atomic_reactor/plugins/pre_reactor_config.py
+++ b/atomic_reactor/plugins/pre_reactor_config.py
@@ -99,8 +99,8 @@ def get_odcs(workflow, fallback=NO_FALLBACK):
     return get_value(workflow, 'odcs', fallback)
 
 
-def get_odcs_session(workflow, fallback):
-    config = get_odcs(workflow, fallback)
+def get_odcs_session(workflow):
+    config = get_odcs(workflow)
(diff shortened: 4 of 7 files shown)


tkdchen:
Could anyone please review if these change is on the correct direction? I'm trying to understand the tests and will make changes accordingly.

athos-ribeiro:
Yes, this seems to be what is expected here so far.

Do not forget to add a reference to a CLOUDBLD in the commit message.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -329,11 +329,7 @@ def adjust_config_kwargs(self):
         self.config_kwargs['koji_hub'] = koji_map['hub_url']
         self.config_kwargs['koji_root'] = koji_map['root_url']
 
-        odcs_fallback = {
-            'api_url': self.config_kwargs.get('odcs_url'),
-            'insecure': self.config_kwargs.get('odcs_insecure')
-        }
-        odcs_map = get_odcs(self.workflow, odcs_fallback)
+        odcs_map = get_odcs(self.workflow)
         self.config_kwargs['odcs_url'] = odcs_map['api_url']
```
athos-ribeiro, line 333: This now implies an ODCS entry is mandatory in the configmap.

We should handle the cases where ODCS integration is not enabled (this is the case for osbs-box). It would also be nice to add some unit tests to cover this case along the changed plugins.

tkdchen: Good catch. Thanks. Perhaps, this piece of code shouldn't be changed. The CLOUDBLD-100 should only affect the `pre_resolve_composes`. @MartinBasti, is my understand correct?

BTW, what is the use case of  `build_orchestrate_build`, when the `config_kwargs` contains odcs config to override the corresponding odcs config defined in reactor_map_config?

athos-ribeiro: I believe that ATM, the only use case for these fallbacks is to have this chunk of code not failing when no ODCS config is found.

Since we are not using a fallback anymore, get_odcs will raise a KeyError in case odcs integration is not configured. We could set the fallback here eother to be None or to return an empty dict. Then we'd have to make sure nothing is trying to directly access the dict key. As an alternative, those fallback values could be directly set to None so we make sure these values are not being fetched from anywhere else. @rcerven WDYT?

rcerven: call get_odcs with {} as fallback
and set config_kwargs['odcs_url'] & config_kwargs['odcs_insecure'] only if they are defined

tkdchen:
PTAL.

tests/plugins/test_orchestrate_build.py
```diff
@@ -164,7 +164,24 @@ def mock_reactor_config(tmpdir, clusters=None, empty=False, add_config=None):
...
+                    'deprecated_keys': ['R122']
+                }
+            ],
+            'default_signing_intent': 'release',
+            'api_url': 'http://odcs.example.com',
+            'auth': {
+                'ssl_certs_dir': '/var/run/secrets/atomic-reactor/odcssecret'
+            }
+        }
+    }
```
chmeliik, line 184: Is the ODCS config required in all (or most) tests? Could it only be passed as the `add_config` param where necesary?

tests/plugins/test_orchestrate_build.py
```diff
@@ -363,20 +380,24 @@ def test_orchestrate_build(tmpdir, caplog, config_kwargs,
         workflow.plugin_workspace[ReactorConfigPlugin.key][WORKSPACE_CONF_KEY] =\
             ReactorConfig(reactor_dict)
     else:
-        reactor_dict = {'version': 1, 'clusters': clusters}
+        reactor_dict = {
+            'version': 1,
+            'clusters': clusters,
+            'odcs': {'api_url': None, 'insecure': False},
+        }
```
chmeliik, line 387: I think rather than adding odcs config here, it might be better to remove the odcs params from `expected_kwargs` to test what happens when odcs is not configured (which is a valid case)

tests/plugins/test_orchestrate_build.py
```diff
@@ -1618,6 +1640,9 @@ def test_set_build_image_with_override(tmpdir, platforms, override):
         'platform_descriptors': [{'platform': 'x86_64', 'architecture': 'amd64'}],
         'build_image_override': {plat: 'registry/osbs-buildroot-{}:latest'.format(plat)
                                  for plat in override},
+        'odcs': {
+            'api_url': 'http://odcs.example.com',
+        },
```
chmeliik, line 1645: Is the odcs config necessary here?

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -329,13 +329,10 @@ def adjust_config_kwargs(self):
...
-        odcs_fallback = {
-            'api_url': self.config_kwargs.get('odcs_url'),
-            'insecure': self.config_kwargs.get('odcs_insecure')
-        }
-        odcs_map = get_odcs(self.workflow, odcs_fallback)
-        self.config_kwargs['odcs_url'] = odcs_map['api_url']
-        self.config_kwargs['odcs_insecure'] = odcs_map.get('insecure', False)
+        odcs_map = get_odcs(self.workflow, {})
+        if odcs_map:
+            self.config_kwargs['odcs_url'] = odcs_map['api_url']
```
rcerven, line 334: odcs_map.get('api_url')

it won't be defined if it isn't defined in config => get_odcs will return that fallback {}

rcerven: when you do this, you have to modify just 1 line in test_orchestrate_build.py

372: from : expected_kwargs['odcs_insecure'] = None
to: expected_kwargs['odcs_insecure'] = False

tkdchen:
Thanks for your review. Updated, PTAL.

atomic_reactor/plugins/build_orchestrate_build.py
```diff
@@ -329,12 +329,8 @@ def adjust_config_kwargs(self):
...
 
-        odcs_fallback = {
-            'api_url': self.config_kwargs.get('odcs_url'),
-            'insecure': self.config_kwargs.get('odcs_insecure')
-        }
-        odcs_map = get_odcs(self.workflow, odcs_fallback)
-        self.config_kwargs['odcs_url'] = odcs_map['api_url']
+        odcs_map = get_odcs(self.workflow, {})
+        self.config_kwargs['odcs_url'] = odcs_map.get('api_url')
         self.config_kwargs['odcs_insecure'] = odcs_map.get('insecure', False)
```
chmeliik, line 334: This may not be necessary anymore, I think `odcs_url` and `odcs_insecure` are no longer used by osbs-client after  https://github.com/containerbuildsystem/osbs-client/pull/927/. However, there may be more parameters that are no longer used and I'd say removing them is out of scope for this PR.

chmeliik:
> Travis build failed, does not seem to be related to the changes here, looking into it

Ah, so @MartinBasti, @SalatskySal and I (@chmeliik) collectively broke atomic-reactor unit-testing with https://github.com/containerbuildsystem/osbs-client/pull/959/.

A while back, atomic-reactor's `read_yaml*` methods were changed to use the osbs-client method, which has now been updated to raise a different exception with a different error message.

I will revert the changes for now, we should re-do them later but also make sure they do not break atomic-reactor (or update unit-tests accordingly)

chmeliik:
Release notes not needed, removal of dead code

PR #1432, opened 2020-04-17
Allow specify test target to test.sh

tkdchen:

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -142,7 +142,9 @@ function setup_osbs() {
 case ${ACTION} in
 "test")
   setup_osbs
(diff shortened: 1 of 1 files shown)


athos-ribeiro:
LGTM

The shellcheck is complaining. Could you silent it? I would not mind if you'd just tell it to skip that check there. See https://github.com/koalaman/shellcheck/wiki/SC2086 for reference here.

tkdchen:
The tests could run directly from inside the created container. So, it does not make sense to make this change. Thanks for your review. I'll close this issue.

PR #1433, opened 2020-04-20
operators: Special handling for relatedImages

chmeliik:
* CLOUDBLD-493

Add `set_related_images()` method to OperatorCSV, refactor code to use the same underlying mechanism for other methods.

Call this method in `pin_operator_digest` after pinning pullspecs.

Add `has_related_image[_env]s` methods to OperatorCSV, use them to exclude CSVs that already have a relatedImages section from pinning etc.

### DIFF
--- a/atomic_reactor/plugins/pre_pin_operator_digest.py
+++ b/atomic_reactor/plugins/pre_pin_operator_digest.py
@@ -44,6 +44,9 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
     When running in a worker:
     - receives replacement pullspec mapping computed by orchestrator
     - replaces pullspecs in ClusterServiceVersion files based on said mapping
+    - creates relatedImages sections in ClusterServiceVersion files
+
+    Files that already have a relatedImages section are excluded.
 
     * reactor-config-map
     """
@@ -63,7 +66,7 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
         super(PinOperatorDigestsPlugin, self).__init__(tasker, workflow)
         self.user_config = workflow.source.config.operator_manifests
         self.site_config = None  # Only relevant (and available) in orchestrator
-        self.replacement_pullspecs = replacement_pullspecs
+        self.replacement_pullspecs = replacement_pullspecs or {}
 
     def run(self):
         """
@@ -90,7 +93,9 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
     def run_in_orchestrator(self):
         """
         Run plugin in orchestrator. Find all image pullspecs,
-        compute their replacements and set build arg for worker
+        compute their replacements and set build arg for worker.
+
+        Exclude CSVs which already have a relatedImages section.
         """
         try:
             self.site_config = get_operator_manifests(self.workflow)
@@ -108,25 +113,32 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
 
     def run_in_worker(self):
         """
-        Run plugin in worker. Replace image pullspecs
-        based on replacements computed in orchestrator.
-        """
-        if not self.replacement_pullspecs:
-            self.log.info("No pullspecs need to be replaced")
-            return
+        Run plugin in worker. Replace image pullspecs based on replacements
+        computed in orchestrator, then create relatedImages sections in CSVs.
 
+        Exclude CSVs which already have a relatedImages section.
+        """
         operator_manifest = self._get_operator_manifest()
         replacement_pullspecs = {
             ImageName.parse(old): ImageName.parse(new)
             for old, new in self.replacement_pullspecs.items()
         }
 
+        self.log.info("Updating operator CSV files")
+
         for operator_csv in operator_manifest.files:
-            self.log.info("Replacing pullspecs in %s", operator_csv.path)
-            # Replace pullspecs everywhere, not just in locations in which they
-            # are expected to be found - OCP 4.4 workaround
-            operator_csv.replace_pullspecs_everywhere(replacement_pullspecs)
-            operator_csv.dump()
+            if not operator_csv.has_related_images():
+                self.log.info("Replacing pullspecs in %s", operator_csv.path)
+                # Replace pullspecs everywhere, not just in locations in which they
+                # are expected to be found - OCP 4.4 workaround
+                operator_csv.replace_pullspecs_everywhere(replacement_pullspecs)
+
+                self.log.info("Creating relatedImages section in %s", operator_csv.path)
+                operator_csv.set_related_images()
+
+                operator_csv.dump()
+            else:
+                self.log.warning("%s has a relatedImages section, skipping", operator_csv.path)
 
     def _get_operator_manifest(self):
         if self.user_config is None:
@@ -151,9 +163,20 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
         return operator_manifest
 
     def _get_pullspecs(self, operator_manifest):
+        self.log.info("Looking for pullspecs in operator CSV files")
         pullspec_set = set()
+
         for operator_csv in operator_manifest.files:
-            pullspec_set.update(operator_csv.get_pullspecs())
+            if not operator_csv.has_related_images():
+                self.log.info("Getting pullspecs from %s", operator_csv.path)
+                pullspec_set.update(operator_csv.get_pullspecs())
+            elif operator_csv.has_related_image_envs():
+                msg = ("Both relatedImages and RELATED_IMAGE_* env vars present in {}. "
+                       "Please remove the relatedImages section, it will be reconstructed "
+                       "automatically.".format(operator_csv.path))
+                raise RuntimeError(msg)
+            else:
+                self.log.warning("%s has a relatedImages section, skipping", operator_csv.path)
 
         # Make sure pullspecs are handled in a deterministic order
         # ImageName does not implement ordering, use str() as key for sorting
--- a/atomic_reactor/utils/operator.py
+++ b/atomic_reactor/utils/operator.py
@@ -10,6 +10,7 @@ from __future__ import absolute_import, unicode_literals
 
 import os
 import logging
+from collections import OrderedDict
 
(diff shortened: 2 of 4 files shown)


chmeliik:
It is not yet clear what should be done if the `relatedImages` section already exists (current approach - check for conflicts, append), see CLOUDBLD-601

atomic_reactor/utils/operator.py
```diff
@@ -17,6 +17,13 @@
 from atomic_reactor.util import ImageName, chain_get
 
 
+__all__ = [
```
MartinBasti, line 20: Do we need this? Somebody may not expect this and have hard times with imports

chmeliik: True, I mostly just wanted to express that `NamedPullspec` and its subclasses are not really a public part of the module, but it's not that important. I'll remove this.

chmeliik:
Possibly not final, but if you have time then PTAL because @MartinBasti wants to get this into the next release :)

tests/plugins/test_pin_operator_digests.py
```diff
@@ -116,23 +118,48 @@ def mock_env(docker_tasker, tmpdir, orchestrator,
...
         'spec': {
+            'relatedImages': [],
             # It does not really matter where in the CSV these pullspecs go
-            # as long as operator_util is known to work properly
-            'relatedImages': [
-                {'name': 'foo-{}'.format(i + 1), 'image': image}
-                for i, image in enumerate(pullspecs)
-            ]
+            # as long as utils.operator is known to work properly, just do not
+            # put them in relatedImages because worker should update those
```
lcarva, line 135: I don't understand what is meant by `as long as utils.operator is known to work properly`. Do we expect it to not work properly? 

Also, what is "worker" ?

atomic_reactor/utils/operator.py
```diff
@@ -87,153 +179,131 @@ def replace_pullspecs(self, replacement_pullspecs):
...
+        """
+        Find pullspecs in predefined locations and add all of them to the
+        .spec.relatedImages section (if not already present)
+        """
+        named_pullspecs = self._named_pullspecs()
+
+        # Check for conflicts (same name, different image)
+        by_name = {}
+        for new in named_pullspecs:
+            old = by_name.get(new.name)
```
lcarva, line 212: You could simplify this:
```python
old = by_name.setdefault(new.name, new)
if new.image != old.image:
  ...
```

atomic_reactor/utils/operator.py
```diff
@@ -87,153 +179,131 @@ def replace_pullspecs(self, replacement_pullspecs):
...
+            if old is None:
+                by_name[new.name] = new
+            elif new.image != old.image:
+                msg = ("{path} - Conflict for pullspec {old.name}: "
+                       "{old.image} ({old.description}) X {new.image} ({new.description})"
+                       .format(path=self.path, old=old, new=new))
+                raise RuntimeError(msg)
```
rcerven, line 219: collect first all conflicts, and raise error after for loop with all of them

atomic_reactor/utils/operator.py
```diff
@@ -87,153 +179,131 @@ def replace_pullspecs(self, replacement_pullspecs):
...
-    def _replace_init_container_pullspecs(self, replacement_pullspecs):
-        init_containers = self._get_init_containers()
-        log_template = (
-            "%(path)s - Replaced pullspec for initContainer %(name)s: %(old)s -> %(new)s"
-        )
-        for c in init_containers:
-            self._replace_pullspec(c, "image", replacement_pullspecs, log_template)
+        if new is not None and new != old:
+            log.debug("%s - Replaced pullspec: %s -> %s", self.path, old, new)
+            obj[key] = str(new)  # `new` is an ImageName
```
rcerven, line 293: if you would use new.to_str() it would be more clearer that it is actually ImageName

tests/plugins/test_pin_operator_digests.py
rcerven: worker cluster?

chmeliik: > I don't understand what is meant by as long as utils.operator is known to work properly. Do we expect it to not work properly?

That's just my justification for why I only put the pullspecs in `.spec.install.{...}.containers` instead of having also initContainers, RELATED_IMAGE vars etc. Basically, I'm saying that since `utils.operator` is tested relatively thoroughly, the content of the CSV file is not that important for testing the plugin.

And yeah, "worker" is worker cluster :D

atomic_reactor/utils/operator.py
```diff
@@ -30,6 +30,95 @@ class NotOperatorCSV(Exception):
...
+    _image_key = "image"
+
+    def __init__(self, data):
+        """
+        Initialize a NamedPullspec
+
+        :param data: Dict-like object in JSON/YAML data
+                     in which the name and image can be found
+        """
+        self.data = data
```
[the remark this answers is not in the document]
chmeliik: I originally did that, but then I found it useful to access the data in [_related_image_env_pullspecs()](https://github.com/chmeliik/atomic-reactor/blob/c4aab84dccea5ced47ad433e7e9ef26a9e0866ff/atomic_reactor/utils/operator.py#L273). I could make `data` a property, but I didn't think it was too important since the class is meant to be internal to the module anyway

chmeliik:
Realized `update_related_images()` could append duplicates, fixed that

chmeliik:
Docs PR: https://github.com/containerbuildsystem/osbs-docs/pull/143

chmeliik:
CLOUDBLD-601 is resolved now.

Tried to keep changes to a minimum, which is why:
OperatorCSV will still happily `get | replace | set` pullspecs from relatedImages section, it is up to the plugin to exclude files which already have one.

Last commit is the change which reflects CLOUDBLD-601 resolution.
Still very much a WIP, need to improve logging, make some things nicer etc.

Unittests will fail horribly for now.

chmeliik:
Improved logging, fixed unit tests, added unit tests. Should be final-ish form

MartinBasti:
`AttributeError: 'list' object has no attribute 'clear'` seems you are having py3 party :)

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -151,9 +163,20 @@ def _get_operator_manifest(self):
...
+        self.log.info("Looking for pullspecs in operator CSV files")
         pullspec_set = set()
+
         for operator_csv in operator_manifest.files:
-            pullspec_set.update(operator_csv.get_pullspecs())
+            if not operator_csv.has_related_images():
+                self.log.info("Getting pullspecs from %s", operator_csv.path)
+                pullspec_set.update(operator_csv.get_pullspecs())
+            elif operator_csv.has_related_image_envs():
+                msg = ("Both relatedImages and RELATED_IMAGE_* env vars present in {}. "
```
MartinBasti, line 174: Nope, we want opposite here. Keeping RELATED_IMAGE_ and letting OSBS to construct relatedImages section

atomic_reactor/utils/operator.py
```diff
@@ -87,153 +192,138 @@ def replace_pullspecs(self, replacement_pullspecs):
...
+                       .format(old=old, new=new))
+                conflicts.append(msg)
+
+        if conflicts:
+            raise RuntimeError("{} - Found conflicts when setting relatedImages:\n{}"
+                               .format(self.path, "\n".join(conflicts)))
+
+        related_images = (self.data.setdefault("spec", CommentedMap())
+                                   .setdefault("relatedImages", CommentedSeq()))
+        related_images.clear()
```
MartinBasti, line 242: :panda_face:

atomic_reactor/utils/operator.py
```diff
@@ -87,153 +192,138 @@ def replace_pullspecs(self, replacement_pullspecs):
...
+        .spec.relatedImages section (if it already exists, clear it first)
+        """
+        named_pullspecs = self._named_pullspecs()
+
+        by_name = OrderedDict()
+        conflicts = []
+
+        for new in named_pullspecs:
+            # Keep track only of the first instance with a given name.
+            # Ideally, existing relatedImages should come first in the list,
```
MartinBasti, line 227: There should be no existing relatedImages, you can assert that if you want :)

atomic_reactor/utils/operator.py
chmeliik: Sad panda

atomic_reactor/plugins/pre_pin_operator_digest.py
chmeliik: @lcarva commented in ticket:
> if relatedImages is set and there are RELATED_IMAGE_* environment variables, fail the build. This is to avoid confusion and ambiguous behavior.

Is that not what we want?

atomic_reactor/utils/operator.py
chmeliik: I don't think `OperatorCSV` should worry about this, it makes sense to me to re-include existing relatedImages if there are any. IMO whether we use it like that or not should be up to the plugin

atomic_reactor/plugins/pre_pin_operator_digest.py
chmeliik: Oh you meant the error message right?

atomic_reactor/utils/operator.py
MartinBasti: ok

atomic_reactor/plugins/pre_pin_operator_digest.py
MartinBasti: yes, message should be different

chmeliik: Ok, fixed in last force-push

MartinBasti:
release notes:
- [behavior change] operator manifests: if `relatedImages` section is defined, OSBS skips pullspec digest pinning. If `relatedImages` section is not defined OSBS will create it by pinning image pullspecs from well known locations (please see documentation for more details)

PR #1438, opened 2020-04-23
Fetch cachito request configuration files

athos-ribeiro:

### DIFF
--- a/atomic_reactor/plugins/pre_download_remote_source.py
+++ b/atomic_reactor/plugins/pre_download_remote_source.py
@@ -10,12 +10,14 @@ Downloads and unpacks the source code archive from Cachito and sets appropriate 
 
 from __future__ import absolute_import
 
+import base64
 import os
 import tarfile
 
 from atomic_reactor.constants import REMOTE_SOURCE_DIR
-from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.download import download_url
+from atomic_reactor.plugin import PreBuildPlugin
+from atomic_reactor.utils.cachito import CFG_TYPE_B64
 
 
 class DownloadRemoteSourcePlugin(PreBuildPlugin):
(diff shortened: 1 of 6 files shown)


athos-ribeiro:
Cc @mprahl

atomic_reactor/plugins/pre_download_remote_source.py
```diff
@@ -58,6 +64,22 @@ def run(self):
...
             tf.extractall(dest_dir)
 
+        # Inject cachito provided configuration files
+        for config in self.config_files:
+            config_path = os.path.join(dest_dir, config['path'])
+            if config['type'] == CFG_TYPE_B64:
+                data = base64.b64decode(config['content'])
+                with open(config_path, 'wb') as f:
+                    f.write(data)
+            elif config['type'] == CFG_TYPE_URL:
```
mprahl, line 74: So I decided to not add this type to Cachito after all. We may have to eventually, but for now, just the base64 type is supported.

Until that support is added to Cachito, I think OSBS shouldn't expect that configuration type. We can always add it in later.

atomic_reactor/plugins/pre_resolve_remote_source.py
```diff
@@ -116,6 +117,7 @@ def set_worker_params(self, source_request, remote_source_url):
         }
         override_build_kwarg(self.workflow, 'remote_source_url', remote_source_url)
         override_build_kwarg(self.workflow, 'remote_source_build_args', build_args)
+        override_build_kwarg(self.workflow, 'remote_source_configs', remote_source_configs)
```
mprahl, line 120: What does this do?

atomic_reactor/plugins/pre_download_remote_source.py
athos-ribeiro: I removed the 'url' type

atomic_reactor/plugins/pre_resolve_remote_source.py
athos-ribeiro: I sets the parameter to be forwarded to the worker builds. This is then set as a param to the download_remote_source plugin.

This is needed because while this plugin runs in the orchestrator build, the download_remote_sources plugin (which prepares the sources to be copied into the build image) runs in the worker builds.

athos-ribeiro:
No docs PR: this is part of adding JS support to cachito builds. This will be documented when the work on supporting JS is complete.

Release notes:
- Cachito-enabled builds now query cachito for possible extra configuration files for the requested sources

PR #1439, opened 2020-04-25
Add extra build metadata reflecting fetched artifacts

### LINKED ISSUE #1435: Record fetch archive information in build metadata
The `fetch-archive-url.yaml` and `fetch-archive-koji.yaml` configurations should be represented in the build metadata.

Right now there appears to be no indication of the link between the resulting build and any archives. In particular, it would be great if we could easily identify that particular build NVRs (via the `fetch-archive-koji.yaml`) have become components of the container build.

obriencj:
Augments the pre_fetch_maven_artifacts plugin to produce more detailed results, which are then used by the exit_koji_import plugin to add extra build data declaring what artifacts were pulled in by `fetch_artifacts_koji.yaml` and `fetch_artifacts_url.yaml`

Adds keys two new keys -- "koji_artifacts" and "url_artifacts" to build["extra"]["image"], named obviously after the two files that specify the relevant artifacts to fetch.

I'd like feedback if there should be more or less information recorded to extra, or if the keys I came up with are silly and you'd prefer something else.

I tested this via test.sh, under OS=fedora OS_VERSION=29 PYTHON_VERSION=3
I encountered some issues with pip not being functional using the default configuration in test.sh, it seems that there's some odd conflicts in that centos image that result in pip.main not existing.

Fixes #1435 

### DIFF
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -42,6 +42,7 @@ except ImportError:
 from atomic_reactor.constants import (
     PROG,
     PLUGIN_KOJI_IMPORT_PLUGIN_KEY,
+    PLUGIN_FETCH_MAVEN_KEY,
     PLUGIN_FETCH_WORKER_METADATA_KEY, PLUGIN_GROUP_MANIFESTS_KEY, PLUGIN_RESOLVE_COMPOSES_KEY,
     PLUGIN_VERIFY_MEDIA_KEY,
     PLUGIN_PUSH_OPERATOR_MANIFESTS_KEY,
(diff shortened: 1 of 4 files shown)


obriencj:
looks like stickler is more strict than flake8, I'll nix that excessive blank line after I have some feedback on the rest of it and whether it's worth pursuing :)

obriencj:
Looks like my branch happened just before a release, so there's some new tests which aren't in my fork. I'll update the tests

lcarva:
Thanks for being proactive and implementing this!

> Adds keys two new keys -- "koji_artifacts" and "url_artifacts" to build["extra"]["image"], named obviously after the two files that specify the relevant artifacts to fetch.

I'd prefer if `build.extra.image` wasn't used for this information. As I previously mentioned in the email thread, this information is also needed for building source containers with middleware content. The [design doc](https://docs.google.com/document/d/1kX8hqeKQcBTqvf6tJzXk67bhZusvCIloMeLnCKdrR4U/edit#) proposes adding this information to the [content generator metadata](https://docs.pagure.org/koji/content_generator_metadata/#buildroots) of the Koji build.

There's also some enhancements required for `fetch_artifacts_url.yaml` so that sources can also be properly recorded.

I think it's important to add this information to the correct location first to avoid having to coordinate a data change later on.

cc @twaugh

obriencj:
@lcarva will be excited to see a PR matching your designs then. When can I expect that?

obriencj:
Closing this out in preference of a PR that uses the buildroot components instead.

PR #1444, opened 2020-04-30
ResolveComposesPlugin allows to handle FROM-scratch image

tkdchen:
* CLOUDBLD-127

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -149,8 +149,14 @@ class ResolveComposesPlugin(PreBuildPlugin):
             self.all_compose_ids = []
 
     def adjust_for_inherit(self):
+        if self.workflow.builder.base_from_scratch:
+            self.log.debug('This is a base image based on scratch. '
+                           'Skipping adjusting composes for inheritance.')
+            return
+
         if not self.plugin_result:
             return
+
         build_info = self.plugin_result[BASE_IMAGE_KOJI_BUILD]
         parent_compose_ids = []
         parent_repourls = []
@@ -219,6 +225,11 @@ class ResolveComposesPlugin(PreBuildPlugin):
         self.adjust_signing_intent_from_parent()
 
     def adjust_signing_intent_from_parent(self):
+        if self.workflow.builder.base_from_scratch:
+            self.log.debug('This is a base image based on scratch. '
+                           'Signing intent will not be adjusted for it.')
+            return
+
         if not self.plugin_result:
             self.log.debug("%s plugin didn't run, signing intent will not be adjusted",
                            PLUGIN_KOJI_PARENT_KEY)
--- a/tests/plugins/test_resolve_composes.py
+++ b/tests/plugins/test_resolve_composes.py
@@ -93,6 +93,7 @@ class MockInsideBuilder(object):
         self.image_id = 'image_id'
         self.image = 'image'
         self.source = MockSource(tmpdir)
+        self.base_from_scratch = False
 
 
 @pytest.fixture()
@@ -1402,3 +1403,19 @@ class TestResolveComposes(object):
                 .get(WORKSPACE_KEY_OVERRIDE_KWARGS, {})
                 .get(arch, {})
                 .get('yum_repourls'))
+
+    def test_skip_adjust_composes_for_inheritance_if_image_is_based_on_scratch(
+            self, workflow, caplog):
+        plugin = ResolveComposesPlugin(workflow.builder.tasker, workflow)
+        workflow.builder.base_from_scratch = True
+        plugin.adjust_for_inherit()
+        assert ('This is a base image based on scratch. '
+                'Skipping adjusting composes for inheritance.' in caplog.text)
+
+    def test_skip_adjust_signing_intent_from_parent_if_image_is_based_on_scratch(
+            self, workflow, caplog):
+        plugin = ResolveComposesPlugin(workflow.builder.tasker, workflow)
+        workflow.builder.base_from_scratch = True
+        plugin.adjust_signing_intent_from_parent()
+        assert ('This is a base image based on scratch. '
+                'Signing intent will not be adjusted for it.' in caplog.text)


atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -151,7 +151,13 @@ def adjust_for_autorebuild(self):
     def adjust_for_inherit(self):
         if not self.plugin_result:
             return
-        build_info = self.plugin_result[BASE_IMAGE_KOJI_BUILD]
+
+        build_info = self.plugin_result.get(BASE_IMAGE_KOJI_BUILD)
+        if build_info is None:
+            self.log.debug('No base image koji build info is found. '
+                           'Skip to adjust inheritance for composes.')
```
ben-alkov, line 158: @tkdchen; Based on this comment, I'm currently confused as to what's being skipped and why. Do you perhaps mean "Skipping forward to adjust inheritance for composes."?

tkdchen: @ben-alkov Yeah, you are right. Sorry, "skip to" was used incorrectly.

tkdchen:
@athos-ribeiro 

The comment inside `InsideBuilder.__init__` mentions that if Dockerfile needs to be generated from `container.yml`, a plugin will create the dockerfile and set base image that sets `workflow.builder.base_from_scratch`. Which plugin does this job? Is this plugin optional to run for some cases before resolve_composes plugin?

athos-ribeiro:
> @athos-ribeiro
> 
> The comment inside `InsideBuilder.__init__` mentions that if Dockerfile needs to be generated from `container.yml`, a plugin will create the dockerfile and set base image that sets `workflow.builder.base_from_scratch`. Which plugin does this job? Is this plugin optional to run for some cases before resolve_composes plugin?

Currently, this is the case for Flatpak builds. There is a pre_create_flatpak_dokerfile plugin that does this. It rund before the resolve_composes plugin in the orchestrator build.

tkdchen:
@athos-ribeiro PR is updated and using `workflow.builder.base_from_scratch` now. PTAL.

athos-ribeiro:
LGTM!

The tests are checking for calls to log.debug, but there are other calls to `debug` in those methods. Could we improve those tests to either check for the logged strings or to check that the plugin will skip in the desired cases?

tkdchen:
@athos-ribeiro 

Thanks for your review. I updated this PR by checking the logged strings.

BTW, I also moved the check of `workflow.builder.base_from_scratch` to the front of the check of `self.plugin_result`. This is not for making the tests pass. My understand is if an image is based on scratch, the code should be actually unnecessary to consider the result from previous plugin, the koji_parent, because the `BASE_IMAGE_KOJI_BUILD` is the only thing `adjust_for_inherit` and `adjust_signing_intent_from_parent` care about.

athos-ribeiro:
Waiving osbs-docs PR and feature toggling through configuration requirements. This is a fix for a bug in the OSBS build workflow.

tests/plugins/test_resolve_composes.py
```diff
@@ -1365,3 +1366,21 @@ def get_override_yum_repourls(self, workflow, arch=ODCS_COMPOSE_DEFAULT_ARCH):
                 .get(WORKSPACE_KEY_OVERRIDE_KWARGS, {})
                 .get(arch, {})
                 .get('yum_repourls'))
+
+    def test_skip_adjust_inheritance_for_composes_if_image_is_based_on_scratch(self, workflow):
+        plugin = ResolveComposesPlugin(workflow.builder.tasker, workflow)
+        flexmock(plugin.log).should_call('debug').once().with_args(
```
rcerven, line 1372: can we use just caplog? and check text in caplog.text

tests/plugins/test_resolve_composes.py
```diff
@@ -1365,3 +1366,21 @@ def get_override_yum_repourls(self, workflow, arch=ODCS_COMPOSE_DEFAULT_ARCH):
...
+            'This is a base image based on scratch. '
+            'Skipping forward to adjust inheritance for composes.'
+        )
+        flexmock(plugin.workflow.builder, base_from_scratch=True)
```
rcerven, line 1376: we can just set it plugin.workflow.builder.base_from_scratch=True
we don't need flexmock for that

tests/plugins/test_resolve_composes.py
```diff
@@ -1365,3 +1366,21 @@ def get_override_yum_repourls(self, workflow, arch=ODCS_COMPOSE_DEFAULT_ARCH):
...
+        plugin.adjust_for_inherit()
+
+    def test_skip_adjust_signing_intent_from_parent_if_image_is_based_on_scratch(self, workflow):
+        plugin = ResolveComposesPlugin(workflow.builder.tasker, workflow)
+        flexmock(plugin.log).should_call('debug').once().with_args(
```
rcerven, line 1381: use caplog

tests/plugins/test_resolve_composes.py
```diff
@@ -1365,3 +1366,21 @@ def get_override_yum_repourls(self, workflow, arch=ODCS_COMPOSE_DEFAULT_ARCH):
...
+            'This is a base image based on scratch. '
+            'Signing intent will not be adjusted for it.'
+        )
+        flexmock(plugin.workflow.builder, base_from_scratch=True)
```
rcerven, line 1385: no need for flexmock

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -149,8 +149,14 @@ def adjust_for_autorebuild(self):
             self.all_compose_ids = []
 
     def adjust_for_inherit(self):
+        if self.workflow.builder.base_from_scratch:
+            self.log.debug('This is a base image based on scratch. '
+                           'Skipping forward to adjust inheritance for composes.')
```
rcerven, line 154: 'Skipping forward to adjust inheritance for composes.' =>
'Skipping adjusting composes for inheritance'

tkdchen:
@rcerven Thanks for your good points. All your comments are addressed. PTAL.

chmeliik:
Release note: resolve_composes plugin will correctly handle FROM-scratch images

PR #1445, opened 2020-04-30
Use DockerBuildWorkflow fixture in unittests

chmeliik:
* CLOUDBLD-73

Replace most occurrences of DockerBuildWorkflow in unittests with the
existing `workflow` fixture.


chmeliik:
This is just the result of a fairly mechanical process:
- remove `DockerBuildWorkflow` occurrences in tests
- add `workflow` params to test methods
- if test uses a method to mock the workflow (most of them do), also pass `workflow` to this method
- reindent for 100 character line length limit

Next step could be to try and identify a common subset of workflow setup that is done in most tests and enhance the fixture to do that. Unfortunately, I'm afraid finding that subset would be quite difficult and would require significant refactoring in some tests. Also, setting up more than the bare minimum in the fixture would make it hard to guarantee that some tests do not pass by accident. Not sure if that is worth it.

Speaking of "not worth it", given how much mocking most tests need to do with the workflow, not sure if this is worth it at all.

chmeliik:
@MartinBasti PTAL if this is what you had in mind with CLOUDBLD-73

MartinBasti:
Yes I wanted this. I hoped that this may reduce more duplications.

I imagine that this workflow should provide functional workflow object with dockerfile path, stub inside builder, if MOCK: mock_builder() section. Using tmpdir fixture for dockerfile path shouldn't force us to change many tests.

This fixture can be base fixture for other more specific workflow fixtures.

But if there is not a huge overlap then it may not be worth fixing indeed. I'll take another look to try find clusters. Keep in mind that some tests have just copy pasted mock functions. I wanted to avoid this copy pasting in future by introducing fixtures with sane defaults.

MartinBasti:
I think that this can be used as very basic fixture. It should be easy for tests to update attributes for their needs.
```
workflow = DockerBuildWorkflow(....)
workflow.source = StubSource()
workflow.builder = StubInsideBuilder().for_workflow(workflow)
return workflow
```

lcarva:
It might be more efficient if we only target a single test file to begin with.

chmeliik:
Closing in favor of a nicer solution

PR #1447, opened 2020-04-30
Bump check for build existence to higher scope...

ben-alkov:
...to catch missing flatpak modules.

Using the same dist-git from the original bug report, log now looks like

```text
"2020-04-30 19:01:24,478 platform:- - atomic_reactor.plugin - ERROR - plugin 'flatpak_update_dockerfile' raised an exception: RuntimeError: No build found for python27:2.7:8030020200422185346:851f4228"
```

* CLOUDBLD-876

### DIFF
--- a/atomic_reactor/utils/koji.py
+++ b/atomic_reactor/utils/koji.py
@@ -246,8 +246,8 @@ def get_koji_module_build(session, module_spec):
                     else:
                         build = b
 
-        if build is None:
-            raise RuntimeError("No build found for {}".format(module_spec.to_str()))
+    if build is None:
+        raise RuntimeError("No build found for {}".format(module_spec.to_str()))
 
     archives = session.listArchives(buildID=build['build_id'])
     # The RPM list for the 'modulemd.txt' archive has all the RPMs, recent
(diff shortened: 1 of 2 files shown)


rcerven:
looks good, just add unit test

atomic_reactor/utils/koji.py
```diff
@@ -246,8 +246,8 @@ def get_koji_module_build(session, module_spec):
                     else:
                         build = b
 
-        if build is None:
-            raise RuntimeError("No build found for {}".format(module_spec.to_str()))
+    if build is None:
+        raise RuntimeError("No build found for {}".format(module_spec.to_str()))
```
athos-ribeiro, line 250: Aha!

ben-alkov:
@athos
> Could we have an unit test to cover this case?

It's already covered, according to Coveralls ¯\\\_(ツ)_/¯

athos-ribeiro:
> @athos
> 
> > Could we have an unit test to cover this case?
> 
> It's already covered, according to Coveralls ¯\_(ツ)_/¯

It only covers the case where the module_spec.context is None (that was the issue).

It should also cover the case where we go through

`if module_spec.context is not None:`

And the call to `session.getBuild(koji_nvr)` in it returns None

ben-alkov:
@athos-ribeiro, @rcerven; Now with 0.00009% more tests :stuck_out_tongue_closed_eyes:

PR #1448, opened 2020-05-04
Remove koji fallback config

SalatskySal:

### DIFF
--- a/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
+++ b/atomic_reactor/plugins/exit_store_metadata_in_osv3.py
@@ -176,7 +176,7 @@ class StoreMetadataInOSv3Plugin(ExitPlugin):
         koji's configuration to be included in the build_annotations.json file,
         which will be attached in the koji task output.
         """
-        koji_config = get_koji(self.workflow, {})
+        koji_config = get_koji(self.workflow)
         whitelist = koji_config.get('task_annotations_whitelist')
         if whitelist:
             annotations['koji_task_annotations_whitelist'] = json.dumps(whitelist)
--- a/atomic_reactor/plugins/pre_reactor_config.py
+++ b/atomic_reactor/plugins/pre_reactor_config.py
@@ -57,8 +57,8 @@ def get_value(workflow, name, fallback):
         raise
 
 
-def get_koji(workflow, fallback=NO_FALLBACK):
-    koji_map = get_value(workflow, 'koji', fallback)
+def get_koji(workflow):
+    koji_map = get_value(workflow, 'koji', NO_FALLBACK)
 
     if 'auth' in koji_map:
         krb_principal = koji_map['auth'].get('krb_principal')
@@ -69,8 +69,8 @@ def get_koji(workflow, fallback=NO_FALLBACK):
     return koji_map
 
 
-def get_koji_session(workflow, fallback):
-    config = get_koji(workflow, fallback)
+def get_koji_session(workflow):
+    config = get_koji(workflow)
 
     from atomic_reactor.utils.koji import create_koji_session
 
@@ -86,8 +86,8 @@ def get_koji_session(workflow, fallback):
     return create_koji_session(config['hub_url'], auth_info, use_fast_upload)
 
 
-def get_koji_path_info(workflow, fallback):
-    config = get_koji(workflow, fallback)
+def get_koji_path_info(workflow):
+    config = get_koji(workflow)
     from koji import PathInfo
 
     # Make sure koji root_url doesn't end with a slash since the url
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -48,8 +48,6 @@ class ResolveComposesPlugin(PreBuildPlugin):
 
     def __init__(self, tasker, workflow,
                  koji_target=None,
-                 koji_hub=None,
-                 koji_ssl_certs_dir=None,
                  signing_intent=None,
                  compose_ids=tuple(),
                  repourls=None,
@@ -59,9 +57,6 @@ class ResolveComposesPlugin(PreBuildPlugin):
         :param tasker: ContainerTasker instance
         :param workflow: DockerBuildWorkflow instance
         :param koji_target: str, contains build tag to be used when requesting compose from "tag"
-        :param koji_hub: str, koji hub (xmlrpc), required if koji_target is used
-        :param koji_ssl_certs_dir: str, path to "cert", and "serverca"
-                                   used when Koji's identity certificate is not trusted
         :param signing_intent: override the signing intent from git repo configuration
         :param compose_ids: use the given compose_ids instead of requesting a new one
         :param repourls: list of str, URLs to the repo files
@@ -77,14 +72,8 @@ class ResolveComposesPlugin(PreBuildPlugin):
         self.compose_ids = compose_ids
 
         self.koji_target = koji_target
-        self.koji_fallback = {
-            'hub_url': koji_hub,
-            'auth': {
-                'ssl_certs_dir': koji_ssl_certs_dir
-            }
-        }
         if koji_target:
-            if not get_koji(self.workflow, self.koji_fallback)['hub_url']:
+            if not get_koji(self.workflow)['hub_url']:
                 raise ValueError('koji_hub is required when koji_target is used')
 
         self.minimum_time_to_expire = minimum_time_to_expire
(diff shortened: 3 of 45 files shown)


MartinBasti:
```diff
diff --git a/tests/plugins/test_add_filesystem.py b/tests/plugins/test_add_filesystem.py
index 16d2c60..42c6d5b 100644
--- a/tests/plugins/test_add_filesystem.py
+++ b/tests/plugins/test_add_filesystem.py
@@ -180,9 +180,23 @@ def mock_image_build_file(tmpdir, contents=None):
     return file_path
 
 
+def mock_koji_config(workflow):
+    workflow.plugin_workspace.setdefault(ReactorConfigPlugin.key, {})
+    reactor_config = workflow.plugin_workspace[ReactorConfigPlugin.key].get(WORKSPACE_CONF_KEY)
+    if not reactor_config:
+        reactor_config = ReactorConfig({})
+    reactor_config.conf['koji'] = {
+        'auth': {},
+        'hub_url': 'https://koji.example.com/koji_hub',
+        'root_url': 'https://koji.example.com/koji_root/',
+    }
+    workflow.plugin_workspace[ReactorConfigPlugin.key][WORKSPACE_CONF_KEY] = reactor_config
+
+
 def mock_workflow(tmpdir, dockerfile=DEFAULT_DOCKERFILE,
                   scratch=False, for_orchestrator=False):
     workflow = DockerBuildWorkflow('test-image', source=MOCK_SOURCE)
+    mock_koji_config(workflow)
     workflow.user_params['scratch'] = scratch
     mock_source = MockSource(tmpdir)
     setattr(workflow, 'builder', X())
```

I think that this is missing piece. Workflow for plugins that needs koji has to be updated. (Of course without copy paste, this function should be reusable)

SalatskySal:
> /packti copr-build

@MartinBasti Please, fix your command

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -57,8 +57,52 @@ def get_value(workflow, name, fallback):
         raise
 
 
-def get_koji(workflow, fallback=NO_FALLBACK):
-    koji_map = get_value(workflow, 'koji', fallback)
+def add_koji_map_in_workflow(workflow, hub_url, root_url=None, reserve_build=None,
```
rcerven, line 60: why and what is this? 

this is used just in tests to set koji values to reactor config,
and in tests we usually already passing dict to ReactorConfig, so passing it to function which will set it isn't useful

MartinBasti: :eyes: yup mocking function should be in tests.

@rcerven this mocking has been introduced, because koji attribute is mandatory now, and tests we failing with missing `koji` config key

SalatskySal: I added this function because some plugins' tests do not configure koji config as usual and I need to manually add all configs to each internal test. So I use this function instead of creating configuration dicts every time.

SalatskySal: This function has been moved to tests/util.py

MartinBasti:
> Do we need an osbs-docs update to mention that koji configuration is now required in configmap?

That's good idea.  :eyes: @SalatskySal

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
```diff
@@ -176,7 +176,7 @@ def set_koji_task_annotations_whitelist(self, annotations):
         koji's configuration to be included in the build_annotations.json file,
         which will be attached in the koji task output.
         """
-        koji_config = get_koji(self.workflow, {})
+        koji_config = get_koji(self.workflow)
```
lcarva, line 179: This looks like a change in behavior. Should this be wrapped in `try/except` block?

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -475,7 +476,14 @@ class ReactorConfig(object):
     Class to parse the atomic-reactor configuration file
     """
 
-    DEFAULT_CONFIG = {ReactorConfigKeys.VERSION_KEY: 1}
+    DEFAULT_CONFIG = {
+        ReactorConfigKeys.VERSION_KEY: 1,
+        ReactorConfigKeys.KOJI_KEY: {
```
lcarva, line 481: Why is this needed?

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -77,14 +72,8 @@ def __init__(self, tasker, workflow,
...
         self.koji_target = koji_target
-        self.koji_fallback = {
-            'hub_url': koji_hub,
-            'auth': {
-                'ssl_certs_dir': koji_ssl_certs_dir
-            }
-        }
         if koji_target:
-            if not get_koji(self.workflow, self.koji_fallback)['hub_url']:
+            if not get_koji(self.workflow).get('hub_url'):
```
lcarva, line 76: `hub_url` is a required attribute of the `koji` Json Object. Shouldn't this be `['hub_url']` instead?

atomic_reactor/schemas/config.json
```diff
@@ -791,5 +791,5 @@
         "required": ["url"]
     }
   },
-  "required": ["version"]
+  "required": ["version", "koji"]
```
lcarva, line 794: Why is this needed as part of this change?

atomic_reactor/plugins/pre_reactor_config.py
SalatskySal: Some tests are using 'DEFAULT_CONFIG' to configure reactor_config_map. 
Inasmuch as 'koji' is mandatory, I've decided to add 'koji' into 'DEFAULT_CONFIG'.

atomic_reactor/plugins/pre_resolve_composes.py
SalatskySal: My mistake, I'll fix that

atomic_reactor/plugins/exit_store_metadata_in_osv3.py
SalatskySal: Even if we delete empty dict, it won't change the behavior of this function, because of 'koji' will always be in reactor_config_map('koji' is required in the schema).

atomic_reactor/plugins/pre_resolve_composes.py
SalatskySal: Fixed

atomic_reactor/plugins/pre_reactor_config.py
MartinBasti: Fix tests, don't modify code with random default values because tests are failing

SalatskySal: Fixed

tests/plugins/test_download_remote_source.py
```diff
@@ -36,12 +36,16 @@
...
+            hub_url: /
+            root_url: ''
+            auth: {}
         cachito:
            api_url: 'example.com'
            insecure: {}
            auth:
                ssl_certs_dir: /some/dir
-        """.format(insecure))
+        """.format("{}", insecure))
```
chmeliik, line 48: When you need a literal `{}` in a formatted string, you can use this:
```python
"a: {{}}, b: {}".format(1)  # -> "a: {}, b: 1"
```
IMO it is more readable.

Note that this appears in other places as well.

SalatskySal: Wow, thanks :+1: :+1: :+1:

SalatskySal: Fixed

MartinBasti:
> @SalatskySal @MartinBasti I think that also all affected plugins, should now require reactor_config_map, and not have it optional, like most of them do

can you elaborate more what you mean?

rcerven:
> > @SalatskySal @MartinBasti I think that also all affected plugins, should now require reactor_config_map, and not have it optional, like most of them do
> 
> can you elaborate more what you mean?

all affected plugins should do this cleanup :
https://github.com/containerbuildsystem/atomic-reactor/pull/1448/files#diff-32d320c81fb4648113f8cd623cf0caf1L383

removing reactor_config param from tests and ALWAYS create one

MartinBasti:
Can we merge this and do it in another PR?

rcerven:
> Can we merge this and do it in another PR?

well he is doing it already here for some plugins, and it is clear from PR which plugins are affected, but up to you

SalatskySal:
> > Can we merge this and do it in another PR?
> 
> well he is doing it already here for some plugins, and it is clear from PR which plugins are affected, but up to you

@rcerven affected plugins have been updated

rcerven:
> > > Can we merge this and do it in another PR?
> > 
> > 
> > well he is doing it already here for some plugins, and it is clear from PR which plugins are affected, but up to you
> 
> @rcerven affected plugins have been updated

looks good, only 1 missing now test_tag_and_push

MartinBasti:
Release notes:
 - Removed deprecated koji fallback methods (`koji` configuration is mandatory now)

PR #1450, opened 2020-05-04
Make parent image check more robust

lkolacek:
CLOUDBLD-558

Customer facing registry might contain additional v2 digests in manifest lists. We should check digests only for platforms that are presented in Koji.

### DIFF
--- a/atomic_reactor/plugins/pre_koji_parent.py
+++ b/atomic_reactor/plugins/pre_koji_parent.py
@@ -14,11 +14,13 @@ from atomic_reactor.constants import (
 )
 from atomic_reactor.plugins.pre_reactor_config import (
     get_deep_manifest_list_inspection, get_koji_session, get_source_registry,
-    get_skip_koji_check_for_base_image, get_fail_on_digest_mismatch
+    get_skip_koji_check_for_base_image, get_fail_on_digest_mismatch,
+    get_platform_to_goarch_mapping
 )
 from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
 from atomic_reactor.util import (
-    base_image_is_custom, get_manifest_list, get_manifest_media_type, is_scratch_build
+    base_image_is_custom, get_manifest_list, get_manifest_media_type, is_scratch_build,
+    get_platforms
 )
 from copy import copy
 from osbs.utils import Labels
@@ -34,6 +36,7 @@ DEFAULT_POLL_INTERVAL = 10  # 10 seconds
 
 class KojiParentBuildMissing(ValueError):
     """Expected to find a build for the parent image in koji, did not find it within timeout."""
+
 
 class KojiParentPlugin(PreBuildPlugin):
     """Wait for Koji build of parent images to be available
@@ -81,6 +84,7 @@ class KojiParentPlugin(PreBuildPlugin):
         self._parent_builds = {}
         self._poll_start = None
         self._source_registry = get_source_registry(workflow, {})
+        self.platforms = get_platforms(self.workflow)
         self._deep_manifest_list_inspection = get_deep_manifest_list_inspection(self.workflow,
                                                                                 fallback=True)
 
@@ -242,12 +246,31 @@ class KojiParentPlugin(PreBuildPlugin):
             v2_digest = archive['extra']['docker']['digests'][v2_type]
             koji_archives_data[arch] = v2_digest
 
-        if koji_archives_data == manifest_list_data:
-            self.log.info('Deeper manifest list check verified v2 manifest references match')
-            return True
-        self.log.warning('Manifest list refs "%s" do not match koji archive refs "%s"',
-                         manifest_list_data, koji_archives_data)
-        return False
+        platform_to_arch_dict = get_platform_to_goarch_mapping(self.workflow)
+        architectures = [platform_to_arch_dict[platform] for platform in self.platforms]
+
+        missing_arches = [a for a in architectures if a not in koji_archives_data]
+        if missing_arches:
+            self.log.warning('Architectures "%s" are missing in Koji archives "%s"',
+                             missing_arches, koji_archives_data)
+            return False
+
+        # manifest lists can be manually pushed to the registry to make sure a specific tag
+        # (e.g., latest) is available for all platforms.
+        # In such cases these manifest lists may include images from different koji builds.
+        # We only want to check the digests for the images built in the current parent koji build
+        err_msg = 'Manifest list digest %s differs from Koji archive digest %s for platform %s'
+        unmatched_digests = False
+        for arch in architectures:
+            if manifest_list_data[arch] != koji_archives_data[arch]:
+                unmatched_digests = True
+                self.log.warning(err_msg, manifest_list_data[arch], koji_archives_data[arch], arch)
+
+        if unmatched_digests:
+            return False
+
+        self.log.info('Deeper manifest list check verified v2 manifest references match')
+        return True
 
     def detect_parent_image_nvr(self, image_name, inspect_data=None):
         """
--- a/tests/plugins/test_koji_parent.py
+++ b/tests/plugins/test_koji_parent.py
@@ -13,7 +13,8 @@ import koji
 
 import atomic_reactor
 from atomic_reactor.constants import (
-    INSPECT_CONFIG, BASE_IMAGE_KOJI_BUILD, PARENT_IMAGES_KOJI_BUILDS
+    INSPECT_CONFIG, BASE_IMAGE_KOJI_BUILD, PARENT_IMAGES_KOJI_BUILDS,
+    PLUGIN_CHECK_AND_SET_PLATFORMS_KEY
 )
 from atomic_reactor.inner import DockerBuildWorkflow
 from atomic_reactor.plugin import PreBuildPluginsRunner, PluginFailedException
@@ -27,6 +28,7 @@ from atomic_reactor.constants import SCRATCH_FROM
 from flexmock import flexmock
 from tests.constants import MOCK, MOCK_SOURCE
 from tests.stubs import StubInsideBuilder
+from copy import deepcopy
 
 import pytest
 
@@ -354,13 +356,17 @@ class TestKojiParent(object):
             [{'digest': 'stubDigest', 'mediaType': 'unexpected', 'platform': {
                 'architecture': 'amd64'}}]},
         {}])
-    def test_deep_digest_inspection(self, workflow, koji_session, reactor_config_map, parent_tag,
+    def test_deep_digest_inspection(self, workflow, koji_session, parent_tag,
                                     caplog, has_registry, manifest_list, feature_flag,
                                     mismatch_failure):  # noqa
+        reactor_config_map = True
         image_str = 'base'
         if has_registry:
             image_str = '/'.join(['example.com', image_str])
         extra = {'image': {'index': {'digests': {V2_LIST: 'stubDigest'}}}}
+
+        workflow.prebuild_results[PLUGIN_CHECK_AND_SET_PLATFORMS_KEY] = ['x86_64']
+
         koji_build = dict(nvr='base-image-1.0-99',
                           id=KOJI_BUILD_ID,
                           state=KOJI_STATE_COMPLETE,
@@ -403,6 +409,8 @@ class TestKojiParent(object):
         workflow.builder.parent_images = parent_images
 
         rebuild_str = 'This parent image MUST be rebuilt'
+        manifest_list_check_passed = ('Deeper manifest list check verified v2 manifest '
+                                      'references match')
 
         defective_v2 = (not has_registry
                         or manifest_list.get('manifests', [{}])[0].get('digest') != 'stubDigest'
@@ -436,8 +444,9 @@ class TestKojiParent(object):
                         assert rebuild_str in caplog.text
                     elif manifest_list['manifests'][0]['digest'] == 'stubDigest':
                         assert rebuild_str not in caplog.text
+                        assert manifest_list_check_passed in caplog.text
                     else:
-                        assert 'do not match koji archive refs' in caplog.text
+                        assert 'differs from Koji archive digest' in caplog.text
                         assert rebuild_str in caplog.text
                 else:
                     fetch_error = 'Could not fetch manifest list for {}:latest'.format(image_str)
@@ -506,7 +515,10 @@ class TestKojiParent(object):
                 ReactorConfig({'version': 1, 'koji': koji_map,
                                'deep_manifest_list_inspection': deep_inspection,
                                'fail_on_digest_mismatch': mismatch_failure,
-                               'skip_koji_check_for_base_image': external_base})
+                               'skip_koji_check_for_base_image': external_base,
+                               'platform_descriptors': [{'architecture': 'amd64',
+                                                         'platform': 'x86_64'}]
+                               })
 
         runner = PreBuildPluginsRunner(
             workflow.builder.tasker,
@@ -523,6 +535,7 @@ class TestKojiParent(object):
                                PARENT_IMAGES_KOJI_BUILDS: {
                                    base_img: KOJI_BUILD}}
             if is_isolated is not None:
+                expected_result = deepcopy(expected_result)
                 expected_result[BASE_IMAGE_KOJI_BUILD]['extra']['image']['isolated'] = is_isolated
                 expected_result[PARENT_IMAGES_KOJI_BUILDS][base_img]['extra']['image']['isolated']\
                     = is_isolated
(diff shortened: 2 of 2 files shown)


atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -229,6 +230,16 @@ def manifest_list_entries_match(self, image, build_id):
         if koji_archives_data == manifest_list_data:
             self.log.info('Deeper manifest list check verified v2 manifest references match')
             return True
+
+        # Customer facing registry might contain additional v2 digests in manifest lists
+        self.log.info('Limit data in manifest list to platforms which are presented in Koji')
```
lkolacek, line 235: In case this information is useless to us .. I can use previous check with modified manifest_list_data_koji_platforms.  

Or I can avoid storing additional platforms (which are not presented in Koji) for manifest list and make it even shorter. [here](https://github.com/containerbuildsystem/atomic-reactor/blob/5ce68db5f863f57ecf847fe326dc871825b3e2de/atomic_reactor/plugins/pre_koji_parent.py#L221)

lkolacek:
PTAL @ben-alkov @athos-ribeiro @chmeliik @rcerven  @MartinBasti

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -229,6 +230,16 @@ def manifest_list_entries_match(self, image, build_id):
         if koji_archives_data == manifest_list_data:
             self.log.info('Deeper manifest list check verified v2 manifest references match')
             return True
+
+        # Customer facing registry might contain additional v2 digests in manifest lists
```
MartinBasti, line 234: This had nothing with registries. Any tag can contain more arches we care about. API just allows that

atomic_reactor/plugins/pre_koji_parent.py
MartinBasti: IMHO we don't need this information and updating previous check would be sufficient

tests/plugins/test_koji_parent.py
```diff
@@ -359,7 +359,32 @@ def test_unexpected_digest_data(self, workflow, koji_session, reactor_config_map
         {'manifests':
             [{'digest': 'stubDigest', 'mediaType': 'unexpected', 'platform': {
                 'architecture': 'amd64'}}]},
-        {}])
+        {},
+        # Customer facing registry might contain additional v2 digests in manifest lists
```
MartinBasti, line 363: :eyes:

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -229,6 +230,16 @@ def manifest_list_entries_match(self, image, build_id):
...
+        self.log.info('Limit data in manifest list to platforms which are presented in Koji')
+        manifest_list_data_koji_platforms = {k: v for (k, v) in manifest_list_data.items()
+                                             if k in koji_archives_data}
+
+        if koji_archives_data == manifest_list_data_koji_platforms:
+            self.log.info('Deeper manifest list check verified v2 manifest references match')
+            return True
```
athos-ribeiro, line 241: Could we remove the duplication of this chunk of code here?

We can always run the filter above. 

It would also be nice to throw a warning in case any platform data is removed by the filter. This way we can further debug future complaints for "missing arches in latest tags" in the future.

athos-ribeiro:
I am wondering if we need further changes/design here:

- imagine we have an image X-1 built for arches A and B
-- X-2 is now built only for arch A
- Someone creates X:latest manifest list in the registry. X:latest has {A: X-2, B: X-1}
- We build Y-1 for A and B on top of X:latest

What happens with Y-1 now? For instance, what is a source container for it? Will the comparing components fail for Y-1 later?

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -226,9 +227,16 @@ def manifest_list_entries_match(self, image, build_id):
             v2_digest = archive['extra']['docker']['digests'][v2_type]
             koji_archives_data[arch] = v2_digest
 
+        # Additional platforms might contain v2 digests of different builds
+        for k in list(manifest_list_data.keys()):
+            if k not in koji_archives_data:
+                self.log.warning('Removing missing %s platform in Koji from manifest list', k)
```
athos-ribeiro, line 233: Could we say "manifest list platform %s is not present in koji archive and will not be veriied"; If it is easily available, we can add the archive id there.

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -226,9 +227,16 @@ def manifest_list_entries_match(self, image, build_id):
             v2_digest = archive['extra']['docker']['digests'][v2_type]
             koji_archives_data[arch] = v2_digest
 
+        # Additional platforms might contain v2 digests of different builds
```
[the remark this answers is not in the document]
lkolacek: I thought @MartinBasti was against mentioning external registries in this upstream repo.  What would you like to have in there? 

```
# Additional platforms might contain v2 digests of different builds
# Identified previously with external non-osbs re-pushed manifest list
```

atomic_reactor/plugins/pre_koji_parent.py
lkolacek: Koji archives are platform based so I can't provide ID of non-existing ones, but how about adding koji-build-id which should clearly inform which platforms are expected in manifest list?

```
 self.log.warning('Manifest list platform %s is not present in Koji build and will not be verified. koji-build-id=%s', k, build_id)
```

atomic_reactor/plugins/pre_koji_parent.py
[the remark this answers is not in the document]
athos-ribeiro: I believe what Martin did not want to see here (and I back him on this) is that we do not want mentions to deployment specific details, such as where/how a specific container registry is accessed.

What I would like to see is a clearer comment on why we are removing certain architectures from the manifest list. It may not be clear for a reader that may be thinking that it was OSBS who created the manifest list at the first place.

This could be a longer comment like: 
```
manifest lists can be manually pushed to the registry to make sure a specific tag
(e.g., latest) is available for all platforms.
In such cases these manifest lists may include images from different koji builds.
We only want to check the digests for the images built in the current parent koji build
```

atomic_reactor/plugins/pre_koji_parent.py
athos-ribeiro: Great!

athos-ribeiro:
> @athos I don't know if I fully understand the case you described but it looks like it will fail in any case while digest will be compared to parent image in koji.

I though the purpose of this patch was to avoid this specific failure.

My question is, if we allow this check to pass, won't we face a further problem during components comparison since the parents are different and so are the RPMs installed in each image?

atomic_reactor/plugins/pre_koji_parent.py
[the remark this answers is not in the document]
rcerven: looks good, only  maybe :
"We only want to check the digests for the images built in the current parent koji build" =>
"We only want to check the digests for platforms built in the current koji build"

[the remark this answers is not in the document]
lkolacek: Added check for requested platforms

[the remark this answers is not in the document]
lkolacek: @athos @rcerven  if this is what we wanted I'll update unit tests

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -216,22 +220,35 @@ def manifest_list_entries_match(self, image, build_id):
                 return False
 
             arch = manifest['platform']['architecture']
+            if arch not in self.architectures:
+                self.log.warning('Requested platform %s is missing in manifest list', arch)
```
rcerven, line 224: we don't need this,
it is already checked in pull_base_image

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -216,22 +220,35 @@ def manifest_list_entries_match(self, image, build_id):
...
             v2_digest = manifest['digest']
             manifest_list_data[arch] = v2_digest
 
         archives = self.koji_session.listArchives(build_id)
         koji_archives_data = {}
         for archive in (a for a in archives if a['btype'] == KOJI_BTYPE_IMAGE):
             arch = archive['extra']['docker']['config']['architecture']
+            if arch not in self.architectures:
+                self.log.warning('Requested platform %s is missing in Koji archive', arch)
+                return False
```
rcerven, line 235: we should first find all missing arches in koji archives and then report all of them

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -216,22 +220,35 @@ def manifest_list_entries_match(self, image, build_id):
...
-            return True
-        self.log.warning('Manifest list refs "%s" do not match koji archive refs "%s"',
-                         manifest_list_data, koji_archives_data)
-        return False
+        # manifest lists can be manually pushed to the registry to make sure a specific tag
+        # (e.g., latest) is available for all platforms.
+        # In such cases these manifest lists may include images from different koji builds.
+        # We only want to check the digests for the images built in the current parent koji build
+        for arch in self.architectures:
+            if manifest_list_data(arch) != koji_archives_data(arch):
```
rcerven, line 244: we could also first check for which arches digests don't match, 
and then return only those arches for which it didn't match

atomic_reactor/plugins/pre_koji_parent.py
lkolacek: So how about this?
```
        if not all(arch in manifest_list_data for arch in self.architectures):
            self.log.warning('Requested platforms "%s" are missing in manifest list "%s"',
                             self.architectures, manifest_list_data)
            return False
````

File: atomic_reactor/plugins/pre_koji_parent.py (hunk 9 above)
Comment (lkolacek) [reply atomic_reactor/plugins/pre_koji_parent.py:249]: This?
```
        msg = 'Manifest list digest {} differs from Koji archive digest {} for platform {},'
        error_msg = ''
        for arch in self.architectures:
            if manifest_list_data(arch) != koji_archives_data(arch):
                error_msg += msg.format(manifest_list_data(arch), koji_archives_data(arch), arch)

        if error_msg:
            self.log.warning(error_msg)
            return False

```

File: atomic_reactor/plugins/pre_koji_parent.py (hunk 7 above)
Comment (lkolacek) [reply atomic_reactor/plugins/pre_koji_parent.py]: Removed, thanks

File: atomic_reactor/plugins/pre_koji_parent.py [hunk 10]
```diff
@@ -226,12 +230,27 @@ def manifest_list_entries_match(self, image, build_id):
...
+
+        # manifest lists can be manually pushed to the registry to make sure a specific tag
+        # (e.g., latest) is available for all platforms.
+        # In such cases these manifest lists may include images from different koji builds.
+        # We only want to check the digests for the images built in the current parent koji build
+        msg = 'Manifest list digest {} differs from Koji archive digest {} for platform {},'
+        error_msg = ''
+        for arch in self.architectures:
+            if manifest_list_data(arch) != koji_archives_data(arch):
+                error_msg += msg.format(manifest_list_data(arch), koji_archives_data(arch), arch)
```
rcerven, line 246: you can log that msg already here

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -63,6 +65,7 @@ def __init__(self, tasker, workflow, koji_hub=None, koji_ssl_certs_dir=None,
                                    used when Koji's identity certificate is not trusted
         :param poll_interval: int, seconds between polling for Koji build
         :param poll_timeout: int, max amount of seconds to wait for Koji build
+        :param architectures: list<str>, list of arches to build on (orchestrator)
```
MartinBasti, line 68: :eyes: where is this param?

rcerven: yeah that can be removed

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -120,7 +124,6 @@ def run(self):
                 else:
                     self.log.error(err_msg)
                     raise RuntimeError(err_msg)
-
```
rcerven, line 123: keep the empty line

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -226,12 +228,27 @@ def manifest_list_entries_match(self, image, build_id):
             v2_digest = archive['extra']['docker']['digests'][v2_type]
             koji_archives_data[arch] = v2_digest
 
-        if koji_archives_data == manifest_list_data:
-            self.log.info('Deeper manifest list check verified v2 manifest references match')
-            return True
-        self.log.warning('Manifest list refs "%s" do not match koji archive refs "%s"',
-                         manifest_list_data, koji_archives_data)
-        return False
+        if not all(arch in koji_archives_data for arch in self.architectures):
```
rcerven, line 231: we will have to convert self.arches to that different format 
self.arches has x86, koji archive/manifest_list amd64

tests/plugins/test_koji_parent.py
```diff
@@ -359,14 +360,40 @@ def test_unexpected_digest_data(self, workflow, koji_session, reactor_config_map
...
+             'platform': {'architecture': 's390x'}}]},
+        ])
     def test_deep_digest_inspection(self, workflow, koji_session, reactor_config_map, parent_tag,
                                     caplog, has_registry, manifest_list, feature_flag,
                                     mismatch_failure):  # noqa
         image_str = 'base'
         if has_registry:
             image_str = '/'.join(['example.com', image_str])
         extra = {'image': {'index': {'digests': {V2_LIST: 'stubDigest'}}}}
+        workflow.prebuild_results[PLUGIN_CHECK_AND_SET_PLATFORMS_KEY] = ['amd64']
```
rcerven, line 396: this should be 'x86_64'

rcerven:
missing coverage

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -226,12 +229,32 @@ def manifest_list_entries_match(self, image, build_id):
...
-        if koji_archives_data == manifest_list_data:
-            self.log.info('Deeper manifest list check verified v2 manifest references match')
-            return True
-        self.log.warning('Manifest list refs "%s" do not match koji archive refs "%s"',
-                         manifest_list_data, koji_archives_data)
-        return False
+        platform_to_arch_dict = get_platform_to_goarch_mapping(self.workflow)
+        architectures = []
+        for arch in self.architectures:
+            architectures.append(platform_to_arch_dict[arch])
```
chmeliik, line 235: I think renaming `self.architectures` to `self.platforms` might help avoid confusion

lkolacek: Nice catch, fixed!

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -226,12 +231,32 @@ def manifest_list_entries_match(self, image, build_id):
...
+            return False
+
+        # manifest lists can be manually pushed to the registry to make sure a specific tag
+        # (e.g., latest) is available for all platforms.
+        # In such cases these manifest lists may include images from different koji builds.
+        # We only want to check the digests for the images built in the current parent koji build
+        err_msg = 'Manifest list digest %s differs from Koji archive digest %s for platform %s'
+        unmatched_digests = False
+        for arch in architectures:
+            if manifest_list_data[arch] != koji_archives_data[arch]:
```
tkdchen, line 250: As the comment mentions, manifest lists can be manually pushed to the registry, is there a mechanism outside OSBS to ensure the arches included in the pushed manifest lists must match or contain arches defined in reactor config `platform_descriptor`? If no, `manifest_list_data[arch]` could raise `KeyError`. Is this an expected special case?

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -226,12 +231,32 @@ def manifest_list_entries_match(self, image, build_id):
...
-        self.log.warning('Manifest list refs "%s" do not match koji archive refs "%s"',
-                         manifest_list_data, koji_archives_data)
-        return False
+        platform_to_arch_dict = get_platform_to_goarch_mapping(self.workflow)
+        architectures = []
+        for platform in self.platforms:
+            architectures.append(platform_to_arch_dict[platform])
+
+        if not all(arch in koji_archives_data for arch in architectures):
+            self.log.warning('Requested platforms "%s" are missing in Koji archives "%s"',
```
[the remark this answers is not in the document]
tkdchen: I think this log could be reworded to be more specific, because it does not tell whether all or which ones are missing in Koji archives data. That could probably means the call of `all` should be replaced with another way to find the difference.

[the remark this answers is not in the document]
lkolacek: If this warning is raised it means that user requested platforms which are not presented in Koji Archive. For us it means that parent check failed and build won't be created. This message suppose to work as debug info why build failed.

I can change the check to following and print only missing arches:
```
        missing_arches = [arch for arch in architectures if arch not in koji_archives_data]
        if missing_arches:
            self.log.warning('Architectures "%s" are missing in Koji archives "%s"',
                             missing_arches, koji_archives_data)
```

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -226,12 +231,32 @@ def manifest_list_entries_match(self, image, build_id):
...
-        if koji_archives_data == manifest_list_data:
-            self.log.info('Deeper manifest list check verified v2 manifest references match')
-            return True
-        self.log.warning('Manifest list refs "%s" do not match koji archive refs "%s"',
-                         manifest_list_data, koji_archives_data)
-        return False
+        platform_to_arch_dict = get_platform_to_goarch_mapping(self.workflow)
+        architectures = []
+        for platform in self.platforms:
+            architectures.append(platform_to_arch_dict[platform])
```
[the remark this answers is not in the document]
lkolacek: sure

atomic_reactor/plugins/pre_koji_parent.py
lkolacek: @rcerven ^

rcerven: if there would be wrong arches in manifest_list, it would fail already in pull_base_image plugin

lkolacek: Thanks robert

lkolacek:
@athos I opened change for documentation update  https://github.com/containerbuildsystem/osbs-docs/pull/149

lkolacek:
@athos change you requested was implemented can you approve this?

athos-ribeiro:
Release notes added: 

- Only perform parent image digest verification for platforms requested in the current build

PR #1451, opened 2020-05-06
operators: Skip pinning when not configured

chmeliik:
* CLOUDBLD-974

In orchestrator, plugin is skipped already, but in workers, plugin
previously relied on the absence of replacement pullspecs to determine
when it should not run.

That is no longer the case, because even when there are no replacement
pullspecs, plugin may still have to create relatedImages sections.

Use same check for orchestrator and for worker.


chmeliik:
Release notes: Fix: pin_operator_digest plugin will now properly be skipped when not configured

PR #1452, opened 2020-05-06
Heuristic for finding pullspecs in operator CSVs

chmeliik:
* CLOUDBLD-907

### DIFF
--- a/atomic_reactor/utils/operator.py
+++ b/atomic_reactor/utils/operator.py
@@ -12,6 +12,7 @@ import os
 import logging
 from collections import OrderedDict
 
+import six
 from ruamel.yaml import YAML
 from ruamel.yaml.comments import CommentedMap, CommentedSeq
 
@@ -23,6 +24,28 @@ log = logging.getLogger(__name__)
 
 
 OPERATOR_CSV_KIND = "ClusterServiceVersion"
+
+
+def is_dict(obj):
+    """
+    Check if object is a dict or the ruamel.yaml equivalent of a dict
+    """
+    return isinstance(obj, (dict, CommentedMap))
+
+
+def is_list(obj):
+    """
+    Check if object is a list or the ruamel.yaml equivalent of a list
+    """
+    return isinstance(obj, (list, CommentedSeq))
+
+
+def is_str(obj):
+    """
+    Check if object is a string or bytes. On python 3, checking for string
+    would be sufficient, but on python 2, it may not be.
+    """
+    return isinstance(obj, (bytes, six.text_type))
 
 
 class NotOperatorCSV(Exception):
@@ -314,13 +337,13 @@ class OperatorCSV(object):
 
     def _replace_pullspecs_everywhere(self, obj, k_or_i, replacement_pullspecs):
         item = obj[k_or_i]
-        if isinstance(item, CommentedMap):
+        if is_dict(item):
             for k in item:
                 self._replace_pullspecs_everywhere(item, k, replacement_pullspecs)
-        elif isinstance(item, CommentedSeq):
+        elif is_list(item):
             for i in range(len(item)):
                 self._replace_pullspecs_everywhere(item, i, replacement_pullspecs)
-        elif isinstance(item, str):
+        elif is_str(item):
             # Doesn't matter if string was not a pullspec, it will simply not match anything
             # in replacement_pullspecs and no replacement will be done
             self._replace_unnamed_pullspec(obj, k_or_i, replacement_pullspecs)
--- a/atomic_reactor/utils/operator.py
+++ b/atomic_reactor/utils/operator.py
@@ -238,19 +238,50 @@ class RelatedImageEnv(NamedPullspec):
 
 
 class Annotation(NamedPullspec):
-    _image_key = NotImplemented
+    """
+    Annotation pullspecs are special, they may be found under any key
+    and there may be more than one pullspec under a single key
+    """
+
+    def __init__(self, data):
+        super(Annotation, self).__init__(data)
+        self._image_key = NotImplemented
+        self._start_i = NotImplemented
+        self._end_i = NotImplemented
+
+    @property
+    def image(self):
+        i, j = self._start_i, self._end_i
+        return self.data[self._image_key][i:j]
+
+    @image.setter
+    def image(self, value):
+        # If there are 2 or more pullspecs in the same annotation text,
+        # they *must* be replaced starting with the last one, otherwise
+        # replacing one pullspec would invalidate the (start, end) indices
+        # of the others. This is up to the code that uses this class.
+        i, j = self._start_i, self._end_i
+        text = self.data[self._image_key]
+        self.data[self._image_key] = text[:i] + value + text[j:]
 
     @property
     def name(self):
-        # Construct name by taking image repo and adding suffix
-        return ImageName.parse(self.image).repo + "-annotation"
+        # Construct name by taking repo and tag from image and adding suffix
+        image = ImageName.parse(self.image)
+        tag = image.tag
+        # If tag is a digest, strip "sha256:" prefix
+        if tag.startswith("sha256:"):
+            tag = tag[len("sha256:"):]
+        return "{}-{}-annotation".format(image.repo, tag)
 
     @property
     def description(self):
         return "{} annotation".format(self._image_key)
 
-    def with_key(self, image_key):
+    def in_key(self, image_key, start=None, end=None):
         self._image_key = image_key
+        self._start_i = start or 0
+        self._end_i = end or len(self.data[image_key])
         return self
 
 
@@ -258,23 +289,32 @@ class OperatorCSV(object):
     """
     A single ClusterServiceVersion file in an operator manifest.
 
-    Can find and replace pullspecs for container images in predefined locations.
+    Can find and replace pullspecs for container images in predefined locations
+    and all annotations (using a heuristic to guess what is a pullspec).
     """
 
-    def __init__(self, path, data):
+    # Annotation keys that are expected to contain pullspecs
+    _known_annotation_keys = ("containerImage",)
+
+    def __init__(self, path, data, pullspec_heuristic=default_pullspec_heuristic):
         """
         Initialize an OperatorCSV
 
         :param path: Path where data was found or where it should be written
         :param data: ClusterServiceVersion yaml data
+        :param pullspec_heuristic: Function that takes a string and returns
+            a list of pullspecs (substrings). If not specified, a default
+            implementation will be used. For more information, see the
+            default_pullspec_heuristic() function in this module.
         """
         if data.get("kind") != OPERATOR_CSV_KIND:
             raise NotOperatorCSV("Not a ClusterServiceVersion")
         self.path = path
         self.data = data
+        self._pullspec_heuristic = pullspec_heuristic
 
     @classmethod
-    def from_file(cls, path):
+    def from_file(cls, path, **kwargs):
         """
         Make an OperatorCSV from a file
 
@@ -283,7 +323,7 @@ class OperatorCSV(object):
         """
         with open(path) as f:
             data = yaml.load(f)
-            return cls(path, data)
+            return cls(path, data, **kwargs)
 
     def dump(self):
         """
@@ -310,14 +350,11 @@ class OperatorCSV(object):
 
         :return: set of ImageName pullspecs
         """
-        named_pullspecs = self._named_pullspecs()
         pullspecs = set()
-
-        for p in named_pullspecs:
+        for p in self._named_pullspecs():
             image = ImageName.parse(p.image)
             log.debug("%s - Found pullspec for %s: %s", self.path, p.description, image)
             pullspecs.add(image)
-
         return pullspecs
 
     def replace_pullspecs(self, replacement_pullspecs):
@@ -326,16 +363,8 @@ class OperatorCSV(object):
 
         :param replacement_pullspecs: mapping of pullspec -> replacement
         """
-        named_pullspecs = self._named_pullspecs()
-
-        for p in named_pullspecs:
-            old = ImageName.parse(p.image)
-            new = replacement_pullspecs.get(old)
-
-            if new is not None and old != new:
-                log.debug("%s - Replaced pullspec for %s: %s -> %s",
-                          self.path, p.description, old, new)
-                p.image = new.to_str()  # `new` is an ImageName
+        for p in self._named_pullspecs():
+            self._replace_named_pullspec(p, replacement_pullspecs)
 
     def replace_pullspecs_everywhere(self, replacement_pullspecs):
         """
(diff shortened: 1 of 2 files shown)


chmeliik:
Fixed python2 failures,  now there should only be 3 failing tests (where the regex-search heuristic finds pullspecs that the default one does not)

chmeliik:
Should be ready now, still need to squash and reorganize commits though. There is also one "xfailed" test for a scenario which is unlikely to ever happen, but I will need some opinions on that.

There is also the question of what to do with false positives. The heuristic is pretty conservative, but a single false positive *will* make the build fail. Not sure if we want to ignore errors though, IDK.

atomic_reactor/utils/operator.py
```diff
@@ -25,6 +27,136 @@
...
+    _name = r"[a-zA-Z0-9\-._]"
+    # Base16 characters
+    _base16 = r"[a-fA-F0-9]"
+
+    # A basic name is anything that contains only alphanumeric and name
+    # characters and starts and ends with an alphanumeric character
+    _basic_name = r"(?:(?:{alnum}{name}*{alnum})|{alnum})".format(alnum=_alnum, name=_name)
+
+    # A basic tag is ':' followed by a basic name
+    _basic_tag = r"(?::{basic_name})".format(basic_name=_basic_name)
```
lcarva, line 69: `_named_tag` ?

atomic_reactor/utils/operator.py
```diff
@@ -25,6 +27,136 @@
...
+    # followed by an optional port number
+    _registry = r"(?:{alnum}{name}*\.{name}*{alnum}(?::\d+)?)".format(alnum=_alnum, name=_name)
+
+    # Namespace is a basic name
+    _namespace = _basic_name
+
+    # Repo is a basic name followed by a tag
+    _repo = _basic_name + _tag
+
+    # NOTE: Tag is REQUIRED, otherwise regex picks up too many false positives,
```
lcarva, line 86: This comment is important but it seems out of place. Can we move it up to where `_repo` is defined?

atomic_reactor/utils/operator.py
```diff
@@ -25,6 +27,136 @@
...
+    # such as URLs, math and possibly many others.
+
+    # Pullspec is registry/repo or registry/namespace/repo
+    _pullspec = r"{registry}/(?:{namespace}/)?{repo}".format(registry=_registry,
```
lcarva, line 90: There could be 0 or many namespaces. Although quay doesn't allow for nested namespaces, the Docker's registry does.

atomic_reactor/utils/operator.py
```diff
@@ -104,7 +236,31 @@ def description(self):
...
+    def image(self):
+        i, j = self._start_i, self._end_i
+        return self.data[self._image_key][i:j]
+
+    @image.setter
+    def image(self, value):
+        # If there are 2 or more pullspecs in the same annotation text,
+        # they *must* be replaced starting with the last one, otherwise
+        # replacing one pullspec would invalidate the (start, end) indices
+        # of the others. This is up to the code that uses this class.
```
lcarva, line 260: To clarify, the method `OperatorCSV.replace_pullspecs_everywhere` will use this class correctly?

atomic_reactor/utils/operator.py
```diff
@@ -25,6 +27,136 @@
...
+    _base16 = r"[a-fA-F0-9]"
+
+    # A basic name is anything that contains only alphanumeric and name
+    # characters and starts and ends with an alphanumeric character
+    _basic_name = r"(?:(?:{alnum}{name}*{alnum})|{alnum})".format(alnum=_alnum, name=_name)
+
+    # A basic tag is ':' followed by a basic name
+    _basic_tag = r"(?::{basic_name})".format(basic_name=_basic_name)
+    # A digest is "@sha256:" followed by base16 characters
+    _digest = r"(?:@sha256:{base16}+)".format(base16=_base16)
```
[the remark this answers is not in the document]
chmeliik: Good idea! Just to be sure, less than 64 characters would always be invalid, right? There is no "can be any unique prefix" sort of thing?

atomic_reactor/utils/operator.py
chmeliik: Oh, I had no idea, thanks!

atomic_reactor/utils/operator.py
chmeliik: Any method that needs to replace annotations, so both `OperatorCSV.replace_pullspecs_everywhere` and the variant without `_everywhere` must use this correctly.

To further clarify, I'm *only* using the heuristic for replacing annotations, nothing else. So the `replace_pullspecs_everywhere` method now comes down to:
* Replace bare pullspecs (not embedded in text) everywhere except metadata.annotations
* Replace annotations using the heuristic

atomic_reactor/utils/operator.py
```diff
@@ -276,11 +430,17 @@ def _container_pullspecs(self):
         ]
 
     def _annotation_pullspecs(self):
+        # Known sources of pullspecs in annotations
+
+        # TODO: Should this be a thing? Have known sources of pullspecs in annotations
+        # and handle them separately from the unknown ones? If so, should only the top
+        # level .metadata.annotations be considered here, or all the nested ones also?
```
chmeliik, line 437: Not sure if it is worth handling the "containerImage" annotation separately from the other (heuristic) annotations, WDYT?

chmeliik:
Squashed and reordered commits, hopefully in a way that makes sense.

Note: when reviewing, you can completely disregard "Rename variables for pullspecs" commit, it is just a search+replace of variable names

atomic_reactor/utils/operator.py
MartinBasti: IMO yes, this is not heuristic, it's clear that this annotation must contain pullspec.

atomic_reactor/utils/operator.py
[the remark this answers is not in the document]
lcarva: I'm not aware of any such restrictions. BTW, this should be *exactly* 64 characters, not less, nor more.

atomic_reactor/utils/operator.py
```diff
@@ -115,32 +269,40 @@ def name(self):
...
         """
         if data.get("kind") != OPERATOR_CSV_KIND:
             raise NotOperatorCSV("Not a ClusterServiceVersion")
         self.path = path
         self.data = data
+        self._pullspec_heuristic = pullspec_heuristic or default_pullspec_heuristic
 
     @classmethod
-    def from_file(cls, path):
+    def from_file(cls, path, **kwds):
```
MartinBasti, line 305: `**kwargs` is common term?

atomic_reactor/utils/operator.py
```diff
@@ -276,11 +430,17 @@ def _container_pullspecs(self):
         ]
 
     def _annotation_pullspecs(self):
+        # Known sources of pullspecs in annotations
+
+        # TODO: Should this be a thing? Have known sources of pullspecs in annotations
```
MartinBasti, line 435: We want replace it and prevent using incorrect registry names in annotation

tests/utils/test_operator.py
```diff
@@ -13,15 +13,178 @@
...
+    ("(a.b/c:1)", ["a.b/c:1"]),
+    ("[a.b/c:1]", ["a.b/c:1"]),
+    ("{a.b/c:1}", ["a.b/c:1"]),
+    # Enclosed in various pullspec characters
+    (".a.b/c:1.", ["a.b/c:1"]),
+    ("-a.b/c:1-", ["a.b/c:1"]),
+    ("_a.b/c:1_", ["a.b/c:1"]),
+    ("/a.b/c:1/", ["a.b/c:1"]),
+    ("@a.b/c:1@", ["a.b/c:1"]),
+    (":a.b/c:1:", ["a.b/c:1"]),
```
MartinBasti, line 70: `8.0-1136.1589201512` can we also test tag like this?

atomic_reactor/utils/operator.py
lcarva: Great!

atomic_reactor/utils/operator.py
```diff
@@ -25,6 +27,134 @@
...
+        i, j = _adjust_for_arbitrary_text(text, i, j)
+        candidate = text[i:j]
+        if PullspecRegex.FULL.match(candidate):
+            pullspecs.append((i, j))
+            log.debug("Pullspec heuristic: %s looks like a pullspec", candidate)
+    return pullspecs
+
+
+def _pullspec_candidates(text):
+    return [match.span() for match in PullspecRegex.CANDIDATE.finditer(text)]
```
MartinBasti, line 145: could be generator?

atomic_reactor/utils/operator.py
```diff
@@ -25,6 +27,134 @@
...
+    # Alphanumeric characters
+    _alnum = r"[a-zA-Z0-9]"
+    # Characters that you might find in a registry, namespace, repo or tag
+    _name = r"[a-zA-Z0-9\-._]"
+    # Base16 characters
+    _base16 = r"[a-fA-F0-9]"
+
+    # A basic name is anything that contains only alphanumeric and name
+    # characters and starts and ends with an alphanumeric character
+    _basic_name = r"(?:(?:{alnum}{name}*{alnum})|{alnum})".format(alnum=_alnum, name=_name)
```
rcerven, line 66: in your regex, the 1st will cover 2+ chars  and 2nd 1 char
why not just ? (this will cover 1+ chars:
r"(?:{alnum}{name}*{alnum})"

MartinBasti:
I cannot find how multivalued annotations are named in this diff. All of them shares the same name in `relatedImages` section?

atomic_reactor/utils/operator.py
rcerven, line 145: why don't we use PullspecRegex.PULLSPEC here?

which isn't used at all.

so instead of using CANDIDATE just use PULLSPEC

rcerven: also when we do this, we won't need _adjust_for_arbitrary_tex

atomic_reactor/utils/operator.py
```diff
@@ -210,7 +361,9 @@ def replace_pullspecs_everywhere(self, replacement_pullspecs):
         :param replacement_pullspecs: mapping of pullspec -> replacment
         """
         for k in self.data:
-            self._replace_pullspecs_everywhere(self.data, k, replacement_pullspecs)
+            self._replace_pullspecs_not_in_annotations(self.data, k, replacement_pullspecs)
+        for annotation in self._annotation_pullspecs() + self._guess_annotation_pullspecs():
```
rcerven, line 365: do we still need "self._annotation_pullspecs()" ???
self._guess_annotation_pullspecs() will find them anyway

atomic_reactor/utils/operator.py
rcerven: no

atomic_reactor/utils/operator.py
lcarva: I had the same question, but there are a few use cases that break if the candidate is not used. I think the main reason is so things like `foo.com/repo:latest/foobar` produce no matches instead of `foo.com/repo:latest`.

rcerven: ahh ok ic

atomic_reactor/utils/operator.py
chmeliik: Same for `**kwds`: https://docs.python.org/3/library/functools.html#functools.wraps
But I don't really have a preference

atomic_reactor/utils/operator.py
chmeliik: Ok, I was thinking the same thing. What about the second point? Is "containerImage" only expected in the top-level metadata.annotations, or potentially also other metadata.annotations in the file?

tests/utils/test_operator.py
chmeliik: Sure,  the "special characters everywhere" cases are kind of similar but I'll add a case

atomic_reactor/utils/operator.py
chmeliik: That's exactly it, I'll add a comment to the PULLSPEC regex (or I could remove that regex, but that would probably leave the same questions)

chmeliik:
> My concern is about tuples (i, j) and keeping it's reverse order. I'm afraid that this might be error prone. Is there a reason why we cannot store original pullpsec and just do substring replacement (exactly once per annotation object)?

It's hard to do string replacement exactly once when there are multiple things to replace, I'm pretty confident that with a naive `str.replace`, 2 or 3 of the cases here would break: https://github.com/chmeliik/atomic-reactor/blob/1217a331fa7817261beadb740e31efe122964cf3/tests/utils/test_operator.py#L961

The first one (1st is a substring of 2nd, and the reverse) could be solved by replacing in order from longest to shortest, but I'm not sure about the others.

You would have to replace them all at the same time, it could probably be done with regex magic, should I try?


> I cannot find how multivalued annotations are named in this diff. All of them shares the same name in relatedImages section?

https://github.com/chmeliik/atomic-reactor/blob/1217a331fa7817261beadb740e31efe122964cf3/tests/utils/test_operator.py#L762

They are named `anX-annotation` (X is between 2 and 7) based on their repo name.

atomic_reactor/utils/operator.py
chmeliik: I see, my question was more about whether we want non-heuristic annotations or just heuristic ones, and you've already answered that, thanks :)

chmeliik:
> They are named anX-annotation (X is between 2 and 7) based on their repo name.

Now that I think about it, if you have two different pullspecs with the same repo in annotations, it will fail, e.g: `registry.io/foo:v1, registry.io/foo:v2` - both will be `foo-annotation`, which is a conflict. We may have to ensure the generated names are unique

MartinBasti:
> > They are named anX-annotation (X is between 2 and 7) based on their repo name.
> 
> Now that I think about it, if you have two different pullspecs with the same repo in annotations, it will fail, e.g: `registry.io/foo:v1, registry.io/foo:v2` - both will be `foo-annotation`, which is a conflict. We may have to ensure the generated names are unique

Yes, that's why I was asking that

MartinBasti:
> > My concern is about tuples (i, j) and keeping it's reverse order. I'm afraid that this might be error prone. Is there a reason why we cannot store original pullpsec and just do substring replacement (exactly once per annotation object)?
> 
> It's hard to do string replacement exactly once when there are multiple things to replace, I'm pretty confident that with a naive `str.replace`, 2 or 3 of the cases here would break: https://github.com/chmeliik/atomic-reactor/blob/1217a331fa7817261beadb740e31efe122964cf3/tests/utils/test_operator.py#L961
> 
> The first one (1st is a substring of 2nd, and the reverse) could be solved by replacing in order from longest to shortest, but I'm not sure about the others.
> 
> You would have to replace them all at the same time, it could probably be done with regex magic, should I try?
> 

Keep it as is for now. 
Thank you!

chmeliik:
Changed how `Annotation` generates names - now, it is `<repo>-<tag>-annotation`
Example:
* registry.io/foo:v1 -> foo-v1-annotation
* registry.io/foo@sha265:abcdef... -> foo-abcdef...-annotation

This can still break (as seen in new unit test) if you use 2 pullspecs with the same repo and tag but different registry or namespace. That probably should not be done anyway.

atomic_reactor/utils/operator.py
chmeliik: I'll collect potential "containerImage" keys from all metadata.annotations objects and ignore "containerImage" in the heuristic, that way we should not find the same annotation pullspec twice

chmeliik:
Settled on handling "known" annotations separately from unknown ones. The "containerImage" annotations will now be collected from all metadata.annotations objects, not just the top-level one. Correspondingly, the heuristic will no longer be used for "containerImage" annotations.

lcarva:
Updated release notes draft:
> - Find embedded pullspecs in operator CSVs `metadata.annotations`

PR #1453, opened 2020-05-07
Include content sets in the image only when compose with pulp_repos are

rcerven:
in container.yaml

* CLOUDBLD-764

### DIFF
--- a/atomic_reactor/plugins/pre_add_content_sets.py
+++ b/atomic_reactor/plugins/pre_add_content_sets.py
@@ -55,7 +55,9 @@ class AddContentSetsPlugin(PreBuildPlugin):
         file_path = os.path.join(workdir, REPO_CONTENT_SETS_CONFIG)
         content_sets = {}
 
-        if os.path.exists(file_path):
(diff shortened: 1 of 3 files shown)


atomic_reactor/plugins/pre_add_content_sets.py
```diff
@@ -55,7 +55,9 @@ def get_output_json(self):
         file_path = os.path.join(workdir, REPO_CONTENT_SETS_CONFIG)
         content_sets = {}
 
-        if os.path.exists(file_path):
+        config_data = self.workflow.source.config.compose or {}
+
+        if os.path.exists(file_path) and config_data.get('pulp_repos'):
```
lcarva, line 60: `content_sets.yml` serve multiple purposes. One of them is to allow ODCS integration via the `compose.pulp_repos` flag in `container.yaml`. The second one is to aid in grading. So although ODCS integration may not be enabled, we still want to add the content sets to the image.

rcerven: ok I'll close this

PR #1458, opened 2020-05-13
util: Fix read_yaml_from_url

chmeliik:
* CLOUDBLD-1063

The method incorrectly assumed that passing `decode_unicode=True` to the
`iter_content()` method of a response would always produce unicode
chunks. That it not the case when the Content-Type of the response is
not appropriate. Instead, make sure chunks are always bytes and decode
them explicitly.


chmeliik:
Release notes: Fixed method of reading yaml files from URLs to support responses with binary Content-Type

PR #1461, opened 2020-05-13
Refactor KojiImportPlugin to reduce complexity...

ben-alkov:
...accrued due to `if` logic surrounding source
container builds.

- Associated fixes for unit tests

i.e. implement
class KojiImportBase
    |- class KojiImportPlugin
    |- class KojiImportSourceContainerPlugin

* CLOUDBLD-121

### DIFF
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -25,8 +25,10 @@ from atomic_reactor.plugins.pre_add_filesystem import AddFilesystemPlugin
 from atomic_reactor.plugins.pre_check_and_set_rebuild import is_rebuild
 from atomic_reactor.util import (OSBSLogs, get_parent_image_koji_data, get_manifest_media_version,
                                  is_manifest_list)
+from atomic_reactor.utils.koji import get_buildroot as koji_get_buildroot
+from atomic_reactor.utils.koji import get_output as koji_get_output
 from atomic_reactor.utils.koji import (
-        get_buildroot, get_output, generate_koji_upload_dir, add_custom_type,
+        generate_koji_upload_dir, add_custom_type,
         get_source_tarball_output, get_remote_source_json_output
 )
 from atomic_reactor.plugins.pre_reactor_config import get_openshift_session
@@ -41,7 +43,7 @@ except ImportError:
 
 from atomic_reactor.constants import (
     PROG,
-    PLUGIN_KOJI_IMPORT_PLUGIN_KEY,
+    PLUGIN_KOJI_IMPORT_PLUGIN_KEY, PLUGIN_KOJI_IMPORT_SOURCE_CONTAINER_PLUGIN_KEY,
     PLUGIN_FETCH_WORKER_METADATA_KEY, PLUGIN_GROUP_MANIFESTS_KEY, PLUGIN_RESOLVE_COMPOSES_KEY,
     PLUGIN_VERIFY_MEDIA_KEY,
     PLUGIN_PUSH_OPERATOR_MANIFESTS_KEY,
@@ -70,7 +72,7 @@ from osbs.utils import Labels
 
 
 @label('koji-build-id')
-class KojiImportPlugin(ExitPlugin):
+class KojiImportBase(ExitPlugin):
     """
     Import this build to Koji
 
@@ -92,7 +94,7 @@ class KojiImportPlugin(ExitPlugin):
     plugins.
     """
 
-    key = PLUGIN_KOJI_IMPORT_PLUGIN_KEY
+    key = None
     is_allowed_to_fail = False
 
     def __init__(self, tasker, workflow, kojihub=None, url=None,
@@ -118,7 +120,7 @@ class KojiImportPlugin(ExitPlugin):
         :param target: str, koji target
         :param poll_interval: int, seconds between Koji task status requests
         """
-        super(KojiImportPlugin, self).__init__(tasker, workflow)
+        super(KojiImportBase, self).__init__(tasker, workflow)
 
         self.koji_fallback = {
             'hub_url': kojihub,
@@ -144,70 +146,14 @@ class KojiImportPlugin(ExitPlugin):
         self.build_id = None
         self.session = None
         self.reserve_build = get_koji(self.workflow, self.koji_fallback).get('reserve_build', False)
-        self.source_build = bool(self.workflow.build_result.oci_image_path)
 
-    def get_output(self, worker_metadatas, buildroot_id):
-        """
-        Build the output entry of the metadata.
+    def get_output(self, *args):
+        # Must be implemented by subclasses
+        raise NotImplementedError
 
-        :return: list, containing dicts of partial metadata
-        """
-        outputs = []
-        output_file = None
-
-        if self.source_build:
-
-            registry = self.workflow.push_conf.docker_registries[0]
-
-            build_name = get_unique_images(self.workflow)[0]
-            pullspec = copy.deepcopy(build_name)
-            pullspec.registry = registry.uri
-
-            outputs, output_file = get_output(workflow=self.workflow, buildroot_id=buildroot_id,
-                                              pullspec=pullspec, platform=os.uname()[4],
-                                              source_build=True, logs=None)
-
-        else:
-            for platform in worker_metadatas:
-                for instance in worker_metadatas[platform]['output']:
-                    instance['buildroot_id'] = '{}-{}'.format(platform, instance['buildroot_id'])
-                    outputs.append(instance)
-
-        return outputs, output_file
-
-    def get_buildroot(self, worker_metadatas):
-        """
-        Build the buildroot entry of the metadata.
-
-        :return: list, containing dicts of partial metadata
-        """
-        buildroots = []
-
-        if self.source_build:
-            buildroot = get_buildroot(build_id=self.build_id, tasker=self.tasker,
-                                      osbs=self.osbs, rpms=False)
-            buildroot['id'] = '{}-{}'.format(buildroot['container']['arch'], buildroot['id'])
-
-            registry = self.workflow.push_conf.docker_registries[0]
-            build_name = get_unique_images(self.workflow)[0].to_str()
-
-            manifest_digest = registry.digests[build_name]
-            digest_version = get_manifest_media_version(manifest_digest)
-            media_type = get_manifest_media_type(digest_version)
-
-            buildroot['extra']['osbs']['koji'] = {
-                'build_name': build_name,
-                'builder_image_id': {media_type: manifest_digest.default}
-            }
-
-            buildroots.append(buildroot)
-        else:
-            for platform in sorted(worker_metadatas.keys()):
-                for instance in worker_metadatas[platform]['buildroots']:
-                    instance['id'] = '{}-{}'.format(platform, instance['id'])
-                    buildroots.append(instance)
-
-        return buildroots
+    def get_buildroot(self, *args):
+        # Must be implemented by subclasses
+        raise NotImplementedError
 
     def set_help(self, extra, worker_metadatas):
         all_annotations = [get_worker_build_info(self.workflow, platform).build.get_annotations()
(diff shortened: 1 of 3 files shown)


ben-alkov:
@reviewers

pylint's complaints are:

- atomic-reactor-fedora-29-py2
  - atomic_reactor/plugins/exit_koji_import.py:380: [E1101(no-member), KojiImportBase.combine_metadata_fragments] Instance of 'KojiImportBase' has no 'get_build' member; maybe '_get_build'?)
  - atomic_reactor/plugins/exit_koji_import.py:470: [E1101(no-member), KojiImportBase.run] Instance of 'KojiImportBase' has no 'get_server_dir' member)
- atomic-reactor-fedora-31-py3
  - atomic_reactor/plugins/exit_koji_import.py:380: [E1101(no-member), KojiImportBase.combine_metadata_fragments] Instance of 'KojiImportBase' has no 'get_build' member; maybe '_get_build'?)
  - atomic_reactor/plugins/exit_koji_import.py:470: [E1101(no-member), KojiImportBase.run] Instance of 'KojiImportBase' has no 'get_server_dir' member)

I'd like to waive them, given that
- I wrote this based on my understanding of Python's classes
- All unit tests pass
- CPython doesn't seem to care nearly as much about this as pylint does :)

chmeliik:
@ben-alkov IMO pylint is right here.

You are calling the `get_build()` and `get_server_dir()` methods in the parent class `KojiImportBase`, but they are only defined in the child classes. If they do need to be called in the base class, they should be defined there, perhaps as abstract if no default definition makes sense - e.g.:
```python
def get_server_dir(self):
    # Must be implemented by subclasses
    raise NotImplementedError
```

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -155,23 +156,10 @@ def get_output(self, worker_metadatas, buildroot_id):
...
-
-        else:
-            for platform in worker_metadatas:
-                for instance in worker_metadatas[platform]['output']:
-                    instance['buildroot_id'] = '{}-{}'.format(platform, instance['buildroot_id'])
-                    outputs.append(instance)
+        for platform in worker_metadatas:
+            for instance in worker_metadatas[platform]['output']:
+                instance['buildroot_id'] = '{}-{}'.format(platform, instance['buildroot_id'])
+                outputs.append(instance)
```
chmeliik, line 162: Is this implementation specific to non-SourceContainer builds? Should it be in the appropriate subclass instead?

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -183,29 +171,10 @@ def get_buildroot(self, worker_metadatas):
...
-            buildroots.append(buildroot)
-        else:
-            for platform in sorted(worker_metadatas.keys()):
-                for instance in worker_metadatas[platform]['buildroots']:
-                    instance['id'] = '{}-{}'.format(platform, instance['id'])
-                    buildroots.append(instance)
+        for platform in sorted(worker_metadatas.keys()):
+            for instance in worker_metadatas[platform]['buildroots']:
+                instance['id'] = '{}-{}'.format(platform, instance['id'])
+                buildroots.append(instance)
```
chmeliik, line 177: See `get_output()` above

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -633,3 +504,215 @@ def run(self):
...
+    def set_media_types(self, extra, worker_metadatas):
+        media_types = []
+        for platform in worker_metadatas:
+            annotations = get_worker_build_info(self.workflow,
+                                                platform).build.get_annotations()
+            if annotations.get('media-types'):
+                media_types = json.loads(annotations['media-types'])
+                break
+        super(KojiImportPlugin, self).set_media_types(extra, worker_metadatas,
+                                                      media_types)
```
chmeliik, line 537: Rather than passing the extra media types to the base method, could we instead call the base method first then adjust the content of `extra['image']['media_types']` here as necessary? Or even adjust extra first, then call the base method to finalize it?

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -633,3 +504,215 @@ def run(self):
...
+
+    def get_build(self, metadata, worker_metadatas):
+        start_time = int(atomic_reactor_start_time)
+        extra = super(KojiImportPlugin, self)._setup_get_build()
```
chmeliik, line 541: This seems to be identical for both source and non-source builds, is there a reason why it could not be done in the base `_get_build()` method? You also would not need to pass the `start_time, extra` params to it in that case.

ben-alkov: AFAICT from local unit testing, `start_time` *has* to be invoked at the head of the method which ends up using it (if that makes sense), otherwise the unit tests of the timestamps fail...

ben-alkov:
> @ben-alkov IMO pylint is right here.
> 
> You are calling the `get_build()` and `get_server_dir()` methods in the parent class `KojiImportBase`, but they are only defined in the child classes. If they do need to be called in the base class, they should be defined there, perhaps as abstract if no default definition makes sense - e.g.:
> 
> ```python
> def get_server_dir(self):
>     # Must be implemented by subclasses
>     raise NotImplementedError
> ```

OHH! *That's* what I was trying to remember... :)

I don't believe that they are ever called from the base class, but I wanted to stub them - I forgot about `NotImplementedError`

atomic_reactor/plugins/exit_koji_import.py
ben-alkov: Absolutely correct.

atomic_reactor/plugins/exit_koji_import.py
ben-alkov: Likewise, as above :)

atomic_reactor/plugins/exit_koji_import.py
ben-alkov: I realized after I went to bed last night that I should refactor `get_build` to be "inside out" from the way it is now :)

ben-alkov: @chmeliik, @athos-ribeiro, @MartinBasti; Can one of you PTAL - I've refactored `set_media_types`, and now AFAICT it's not even being called, and I can't figure out why...

The pytests from Travis should show the same failure I'm seeing.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -633,3 +495,243 @@ def run(self):
...
+            annotations = get_worker_build_info(self.workflow,
+                                                platform).build.get_annotations()
+            if annotations.get('media-types'):
+                media_types = json.loads(annotations['media-types'])
+                break
+
+        # Append new if any, and re-sort media_types
+        existing_mts = extra['image'].get('media_types')
+        if media_types and existing_mts:
+            extra['image']['media_types'] = sorted(set(existing_mts.append(media_types)))
```
chmeliik, line 535: AFAIK append/extend (it should be extend in this case) do not return the object they are called on, i.e. `some_list.extend(some_other_list) -> modifies some_list, returns None`. You could use the `+` operator instead

Not sure if this is related to the failure in unittests though

ben-alkov: D'OH!

ben-alkov: It was apparently a combination of this, plus the fact that (inexplicably) `set([].extend([]))` throws "'NoneType' object is not iterable"

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -633,3 +494,247 @@ def run(self):
...
+                 blocksize=None,
+                 target=None, poll_interval=5):
+        super(KojiImportSourceContainerPlugin, self).__init__(tasker, workflow,
+                                                              kojihub, url,
+                                                              verify_ssl, use_auth,
+                                                              koji_ssl_certs, koji_proxy_user,
+                                                              koji_principal, koji_keytab,
+                                                              blocksize,
+                                                              target, poll_interval)
+        self.source_build = bool(self.workflow.build_result.oci_image_path)
```
rcerven, line 675: not needed anymore

ben-alkov: What?

ben-alkov: Ah. Somehow it wasn't connecting in my mind that this is (obviously) setting the `source_build` flag, which is what *was* used to switch through various the `if`s

ben-alkov:
The 5 new lines not covered by unit tests are all method stubs in the parent class which raise NotImplementedError.

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -633,3 +494,250 @@ def run(self):
...
+
+        build_name = get_unique_images(self.workflow)[0]
+        pullspec = copy.deepcopy(build_name)
+        pullspec.registry = registry.uri
+
+        outputs, output_file = koji_get_output(workflow=self.workflow, buildroot_id=buildroot_id,
+                                               pullspec=pullspec, platform=os.uname()[4],
+                                               source_build=True, logs=None)
+
+        return outputs, output_file
```
tkdchen, line 691: Why not just `return koji_get_output(...)`? :)

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -633,3 +494,250 @@ def run(self):
...
+        digest_version = get_manifest_media_version(manifest_digest)
+        media_type = get_manifest_media_type(digest_version)
+
+        buildroot['extra']['osbs']['koji'] = {
+            'build_name': build_name,
+            'builder_image_id': {media_type: manifest_digest.default}
+        }
+
+        buildroots.append(buildroot)
+        return buildroots
```
tkdchen, line 718: Looks this method only returns a list containing a single builtroot. Why not just `return [buildroot]` directly?

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -633,3 +494,250 @@ def run(self):
...
+                break
+
+        # Copy existing; append with new, if any; de-dupe and re-sort
+        extra_image_mts = copy.deepcopy(extra['image'].get('media_types', []))
+        # FWIW: `set([].extend([]))` (the case where both
+        #     `extra['image']['media_types']` *and* `media_types` are `[]`)
+        #     throws "'NoneType' object is not iterable"
+        extra_image_mts.extend(media_types)
+        if media_types and (type(extra_image_mts) == list):
+            extra['image']['media_types'] = sorted(set(extra_image_mts))
```
chmeliik, line 538: The comment is not exactly accurate IMO. This is how `extend()` works in python:
```python
>>> some_list = [1, 2, 3]
>>> some_other_list = [4, 5, 6]
>>> print(some_list.extend(some_other_list))
None
>>> print(some_list)
[1, 2, 3, 4, 5, 6]
```

`extend()` only modifies `some_list`, it does not return anything - i.e. in Python terms: returns `None`

There should also be no need to check the type of `extra_image_mts`

atomic_reactor/plugins/exit_koji_import.py
ben-alkov: Valid point :)

atomic_reactor/plugins/exit_koji_import.py
ben-alkov: I see. This was brought in from the general case where there might have been more than one...

atomic_reactor/plugins/exit_koji_import.py
```diff
@@ -365,85 +303,19 @@ def set_group_manifest_info(self, extra, worker_metadatas):
                                 digests = get_digests_map_from_annotations(annotations['digests'])
                                 instance['extra']['docker']['digests'] = digests
 
+    def _process_extra(self, *args):
```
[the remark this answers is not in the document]
ben-alkov: A leftover from an earlier refactor. Both subclasses now have the same arg list. BUT it does need a docstring :)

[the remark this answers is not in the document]
ben-alkov: Good points. 

I'll rename `_process_extra` to `_update_extra` to highlight it's side effects.

atomic_reactor/plugins/exit_koji_import.py
ben-alkov: We have `buildroot_id = buildroot[0]['id']` in `combine_metadata_fragments()`, which is not differentiated by build type. Doesn't seem worthwhile to add complexity *there*, in order to (possibly) reduce confusion *here*.

atomic_reactor/plugins/exit_koji_import.py
ben-alkov: (；⌣̀_⌣́) see below...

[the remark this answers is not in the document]
ben-alkov: I *swear* that that's essentially the first thing I tried, and I persistently got a weird `'NoneType' object is not iterable` exception...

Works now ¯\\\_(ツ)_/¯

tkdchen:
LGTM

ben-alkov:
> The refactoring looks great!
> 
> I'd like to to test this before merging, and for that, we will also need a patch to osbs-client updating the plugins configuration for source container builds.

Incoming!

PR #1462, opened 2020-05-15
resolve_composes: strip out profiles from modules

owtaylor:
For Flatpaks, we want to allow specifying modules as name:stream/profile
to allow fine-tuning which packages from the module get installed into
the container - but when we switched Flatpaks to use resolve_composes
instead of having a separate plugin, the logic to strip out profiles
before triggering the build got lost.

Parsing the specs to remove the /profile causes us to reject a bare
'name' with a clear:

   atomic_reactor.plugin.PluginFailedException: plugin 'resolve_composes'
   raised an exception: ValueError: Module specification spam_modules should
   be in NAME:STREAM[:VERSION[:CONTEXT]][/PROFILE] format

exception - previously this would been passed to ODCS, which would fail
the compose. The tests are updated accordingly so all modules are in
name:stream format.

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -12,6 +12,8 @@ from datetime import datetime, timedelta
 import os
 from collections import defaultdict
 
+from osbs.repo_utils import ModuleSpec
+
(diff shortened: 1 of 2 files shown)


athos-ribeiro:
Release notes added:

- Do not pass module profiles to ODCS when resolving Flatpak composes

PR #1469, opened 2020-05-19
Add pull_registries option to reactor config map

chmeliik:
* CLOUDBLD-397

pull_registries is a list of additional registries that may be used for
pulling container images (source_registry is the default one)

### DIFF
--- a/atomic_reactor/schemas/config.json
+++ b/atomic_reactor/schemas/config.json
@@ -513,18 +513,7 @@
                     "description": "Don't check SSL certificate for url",
                     "type": "boolean"
                 },
-                "auth": {
-                    "description": "Authentication information",
-                    "type": "object",
-                    "properties": {
-                        "cfg_path": {
-                            "description": "Path to directory  containing .dockercfg for registry auth",
-                            "type": "string"
-                        }
-                    },
-                    "additionalProperties": false,
-                    "required": ["cfg_path"]
-                },
+                "auth": {"$ref": "#/definitions/registry_auth"},
                 "expected_media_types": {"$ref": "#/definitions/media_types"}
             },
             "additionalProperties": false,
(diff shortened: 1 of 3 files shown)


atomic_reactor/schemas/config.json
```diff
@@ -584,22 +573,33 @@
...
+        "type": "array",
+        "items": {
+            "description": "Any container registry other than the default one",
+            "type": "object",
+            "properties": {
+                "hostname": {
+                    "description": "Registry hostname (not a URL, i.e. no http(s), no port)",
+                    "type": "string",
+                    "pattern": "^(?!http(s)?://).*$"
+                },
```
chmeliik, line 592: @lkolacek @MartinBasti PTAL, do I understand correctly that it must be a plain hostname, not a URL? Or does it not really matter?

MartinBasti: yes hostname, but I'm checking the implementation and it seems that current code expects URLs. so we may need both?

MartinBasti: or having just URL and parsing hostname from that

chmeliik: Ok, will use URL instead

chmeliik:
Now using URL instead of hostname for pull_registries (also merged the definitions for `source_registry` and `pull_registries.items` into one). Hostname will be parsed from URL as the `docker_uri` attribute of `RegistryURI`

atomic_reactor/schemas/config.json
lkolacek: OK I will change it in design and PR in osbs-playbooks

MartinBasti:
Looks good! We need also docs :-/

chmeliik:
Docs PR: https://github.com/containerbuildsystem/osbs-docs/pull/150

chmeliik:
Release notes not needed, internal support for configuration option that is not yet used anywhere

PR #1470, opened 2020-05-20
Create RegistryClient class

chmeliik:
* CLOUDBLD-397

Depends on #1469 

The `create_from_config()` method of RegistrySession will eventually be used in `pin_operator_digest` to support other registries.

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -669,6 +669,239 @@ class RegistrySession(object):
         return self._do(self.session.delete, relative_url, **kwargs)
 
 
+class RegistryClient(object):
+    """
+    Registry client, provides methods for looking up image digests and configs
+    in container registries.
+
+    Methods that accept the `image` parameter expect that the registry of the
+    image matches the one of the client instance, but they do not check whether
+    that is true.
+    """
+
+    def __init__(self, registry_session):
+        self._session = registry_session
+
+    def get_manifest(self, image, version):
+        saved_not_found = None
+        media_type = get_manifest_media_type(version)
+        try:
+            response = query_registry(self._session, image, digest=None, version=version)
+        except (HTTPError, RetryError) as ex:
+            if ex.response is None:
+                raise
+            if ex.response.status_code == requests.codes.not_found:
+                saved_not_found = ex
+            # If the registry has a v2 manifest that can't be converted into a v1
+            # manifest, the registry fails with status=400 (BAD_REQUEST), and an error code of
+            # MANIFEST_INVALID. Note that if the registry has v2 manifest and
+            # you ask for an OCI manifest, the registry will try to convert the
+            # v2 manifest into a v1 manifest as the default type, so the same
+            # thing occurs.
+            if version != 'v2' and ex.response.status_code == requests.codes.bad_request:
+                logger.warning('Unable to fetch digest for %s, got error %s',
+                               media_type, ex.response.status_code)
+                return None, saved_not_found
+            # Returned if the manifest could not be retrieved for the given
+            # media type
+            elif (ex.response.status_code == requests.codes.not_found or
+                  ex.response.status_code == requests.codes.not_acceptable):
+                logger.debug("skipping version %s due to status code %s",
+                             version, ex.response.status_code)
+                return None, saved_not_found
+            else:
+                raise
+
+        if not manifest_is_media_type(response, media_type):
+            logger.warning("content does not match expected media type")
+            return None, saved_not_found
+        logger.debug("content matches expected media type")
+        return response, saved_not_found
+
+    def get_manifest_digests(self,
+                             image,
+                             versions=('v1', 'v2', 'v2_list', 'oci', 'oci_index'),
+                             require_digest=True):
+        """Return manifest digest for image.
+
+        :param image: ImageName, the remote image to inspect
+        :param versions: tuple, which manifest schema versions to fetch digest
+        :param require_digest: bool, when True exception is thrown if no digest is
+                                     set in the headers.
+
+        :return: dict, versions mapped to their digest
+        """
+
+        digests = {}
+        # If all of the media types return a 404 NOT_FOUND status, then we rethrow
+        # an exception, if all of the media types fail for some other reason - like
+        # bad headers - then we return a ManifestDigest object with no digests.
+        # This is interesting for the Pulp "retry until the manifest shows up" case.
+        all_not_found = True
+        saved_not_found = None
+        for version in versions:
+            media_type = get_manifest_media_type(version)
+            response, saved_not_found = get_manifest(image, self._session, version)
+
+            if saved_not_found is None:
+                all_not_found = False
+
+            if not response:
+                continue
+            # set it to truthy value so that koji_import would know pulp supports these digests
+            digests[version] = True
+
+            if not response.headers.get('Docker-Content-Digest'):
+                logger.warning('Unable to fetch digest for %s, no Docker-Content-Digest header',
+                               media_type)
+                continue
+
+            digests[version] = response.headers['Docker-Content-Digest']
+            context = '/'.join([x for x in [image.namespace, image.repo] if x])
+            tag = image.tag
+            logger.debug('Image %s:%s has %s manifest digest: %s',
+                         context, tag, version, digests[version])
+
+        if not digests:
+            if all_not_found and len(versions) > 0:
+                raise saved_not_found   # pylint: disable=raising-bad-type
+            if require_digest:
+                raise RuntimeError('No digests found for {}'.format(image))
+
+        return ManifestDigest(**digests)
+
+    def get_manifest_list(self, image):
+        """Return manifest list for image.
+
+        :param image: ImageName, the remote image to inspect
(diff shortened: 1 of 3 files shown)


chmeliik:
I would like to resolve the cyclic import in `RegistrySession.create_from_config`, this would probably mean creating a new utils module for `Registry{Session,Client}`, not sure if it should be done in this PR

@MartinBasti PTAL, I sort of did the inverse of what you suggested (`RegistryClient` implements the existing methods, existing methods depend on `RegistryClient`). This way, the existing methods do not need an extra `registry_session` param

MartinBasti:
> @MartinBasti PTAL, I sort of did the inverse of what you suggested (`RegistryClient` implements the existing methods, existing methods depend on `RegistryClient`). This way, the existing methods do not need an extra `registry_session` param

:open_mouth:

atomic_reactor/util.py
```diff
@@ -738,6 +738,41 @@ def __init__(self, registry, insecure=False, dockercfg_path=None, access=None):
 
         self.session = get_retrying_requests_session()
 
+    @classmethod
+    def create_from_config(cls, registry, workflow, access=None):
```
MartinBasti, line 742: IMHO `registry=None` should be defined. We may not have always registry part available, default registry should be used in this case

tests/plugins/test_reactor_config.py
```diff
@@ -148,6 +149,115 @@ def test_get_docker_registry(self, config, fallback, valid):
...
+         """,
+         [
+             {"uri": RegistryURI("registry.io"),
+              "insecure": False,
+              "dockercfg_path": None},
+         ]),
+        ("""\
+            version: 1
+            pull_registries:
+            - url: https://registry.io
```
MartinBasti, line 171: IIRC this is with API endpoint `https://registry.io/v2`

atomic_reactor/util.py
```diff
@@ -767,6 +802,254 @@ def delete(self, relative_url, **kwargs):
...
+    that is true.
+
+    Use RegistryClient.create() to get a client for a specific registry.
+    """
+
+    def __init__(self, registry_session):
+        self._session = registry_session
+
+    @classmethod
+    def create(cls, registry, workflow, access=None):
```
MartinBasti, line 821: I'm not sure about this. It doesn't have additional value and it violates dependency injection principle.

IMHO sufficient is having just
```python
client = RegistryClient(RegistrySession.create_from_config(workflow, image.registry))
```

MartinBasti: I'll let others express about this too :)

atomic_reactor/util.py
```diff
@@ -738,6 +738,41 @@ def __init__(self, registry, insecure=False, dockercfg_path=None, access=None):
...
+        use the default source_registry configuration (if present, else an error is thrown).
+
+        :param registry: str, registry to create session for (as hostname[:port], no http(s))
+        :param workflow: DockerBuildWorkflow, contains configuration for registries
+        :param access: List of actions this session is allowed to perform, e.g. ('push', 'pull')
+
+        :return: RegistrySession for the specified registry
+        """
+        # Has to be imported here (cyclic import)
+        from atomic_reactor.plugins.pre_reactor_config import (get_source_registry,
```
MartinBasti, line 756: Keep it for now here, we can move it to custom submodule (or resolve cyclic import in different way) in another PR if needed.

MartinBasti: Or decouple workflow from it, only ask for default and pull_registries configuration in params

tests/plugins/test_reactor_config.py
MartinBasti: Hmm, this like this in `registries` section, I see inconsistency. Ok, keep this consistent with source_registry

atomic_reactor/util.py
chmeliik: Got it, what about the case when the registry is defined but does not match any pull_registries or source_registry? Currently I just use source_registry in that case, is that correct? Or should it use default configuration `{'uri': registry, 'insecure': False, 'dockercfg_path': None}`

chmeliik:
Made `registry` an optional param

atomic_reactor/util.py
chmeliik: I tried the decoupling solution, it might be kind of hard to use, you would have to do this before creating a registry session:
```python
source_registry = get_source_registry(workflow, fallback=None)
pull_registries = get_pull_registries(workflow, fallback=[])
```
Maybe the cyclic import is fine for now, IDK

chmeliik:
```
Message: maximum recursion depth exceeded
Exception: RecursionError('maximum recursion depth exceeded')
```

packit? :eyes:

atomic_reactor/util.py
MartinBasti: I'm thinking how much ugly it would be to add sort of registry factory method into workflow. It will encapsulate everything, hides workflow/configuration implementation, and can cache instances :thinking:

MartinBasti: keep it with cyclic imports for now
functionality first, cleanup later :D

atomic_reactor/util.py
MartinBasti: Keep current behavior, so default source_registry.

Also you can open ticket about failing when registry doesn't match.

atomic_reactor/util.py
```diff
@@ -738,6 +738,46 @@ def __init__(self, registry, insecure=False, dockercfg_path=None, access=None):
...
+        """
+        if registry is None:
+            if source_registry is not None:
+                matched_registry = source_registry
+            else:
+                msg = "No source_registry configured, cannot create default session"
+                raise RuntimeError(msg)
+        else:
+            # Try to match a registry in pull_registries, if no match found, use source_registry
+            matched_registry = next(
```
ben-alkov, line 767: @chmeliik; Why `next()`? This skips the first iteration, right?

chmeliik: `next()` returns the, well, *next* element in an iterator, e.g.:
```python
>>> it = iter([1, 2, 3])
>>> next(it)
1
>>> next(it)
2
>>> next(it)
3
>>> next(it)
StopIteration
```
I'm not sure how idiomatic my code is, but this is what it does:
```python
# 1. Make iterator of all registries that match the registry given as a parameter
# (could be empty)
it = (reg for reg in pull_registries if reg['uri'].docker_uri == registry)

# 2. Take first element from this iterator, or the default if iterator is empty
matched_registry = next(it, source_registry)
```

chmeliik:
> I like the organization of the code. Although, I think ideally the RegistryClient would be used directly. Is the plan to do that at a later time?

Yes, the plan is to gradually move to `RegistryClient`, starting with plugins that actually need the support for `pull_registries`, e.g. `pin_operator_digest`

atomic_reactor/util.py
```diff
@@ -767,6 +807,258 @@ def delete(self, relative_url, **kwargs):
...
+            # set it to truthy value so that koji_import would know pulp supports these digests
+            digests[version] = True
+
+            if not response.headers.get('Docker-Content-Digest'):
+                logger.warning('Unable to fetch digest for %s, no Docker-Content-Digest header',
+                               media_type)
+                continue
+
+            digests[version] = response.headers['Docker-Content-Digest']
+            context = '/'.join([x for x in [image.namespace, image.repo] if x])
```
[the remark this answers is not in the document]
chmeliik: Probably, I just copy-pasted the implementations from their original location, I'll take a look

atomic_reactor/util.py
chmeliik: Any other opinions? I wouldn't mind dropping this

MartinBasti: Drop it please. Thanks

atomic_reactor/util.py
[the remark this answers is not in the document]
MartinBasti: this is just copy pasted code. Don't change it in this PR to avoid any disruptions

chmeliik:
Dropped last commit (the cyclic import thing) removed `RegistryClient.create()`

chmeliik:
Rebased (also removed unused `RegistryClient` import from test_util - it is currently tested using only existing tests for get_manifest_digests etc.)

atomic_reactor/util.py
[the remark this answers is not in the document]
tkdchen: I searched through the codebase, there are 3 occurrences of `[x for x in [image.namespace, image.repo] if x]` in `util.py`. Perhaps we could replace them separately.

lkolacek:
Looks really nice Adam!

chmeliik:
Noticed that some of the methods in `RegistryClient` were using the module-level `get_manifest` instead of the one in RegistryClient, fixed (it was functionally equivalent but weird)

atomic_reactor/util.py
ben-alkov: Got it. Thanks!

rcerven:
Code looks good

chmeliik:
Release notes not needed, more internal support for pull_registries, not yet used

PR #1476, opened 2020-05-26
Create mechanism for mocking workflow in unit tests

chmeliik:
* CLOUDBLD-73

### DIFF
--- a/tests/mock_env.py
+++ b/tests/mock_env.py
@@ -1,0 +1,215 @@
+"""
+Copyright (c) 2020 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import unicode_literals, absolute_import
+
+from flexmock import flexmock
+
+from atomic_reactor.constants import PLUGIN_BUILD_ORCHESTRATE_KEY
+from atomic_reactor.plugin import (PreBuildPluginsRunner,
+                                   BuildStepPluginsRunner,
+                                   PostBuildPluginsRunner,
+                                   PrePublishPluginsRunner,
+                                   ExitPluginsRunner)
+from atomic_reactor.inner import DockerBuildWorkflow
+from atomic_reactor.plugins.pre_reactor_config import (ReactorConfig,
+                                                       ReactorConfigPlugin,
+                                                       WORKSPACE_CONF_KEY)
+
+from tests.constants import TEST_IMAGE, MOCK_SOURCE
+from tests.stubs import StubSource, StubInsideBuilder
+
+
+class MockEnv(object):
+    """
+    Mock environment for unit tests.
+
+    Provides methods for setting up workflow (DockerBuildWorkflow) and runner (PluginsRunner)
+    for a specific test scenario.
+
+    Example usage:
+    >>> runner = (MockEnv()
+    >>>           .for_plugin('prebuild', 'my_plugin')
+    >>>           .set_scratch(True)
+    >>>           .make_orchestrator()
+    >>>           .create_runner(docker_tasker))  # docker_tasker is a fixture
+    >>> runner.run()
+    """
+
+    _plugin_phases = ('prebuild', 'buildstep', 'postbuild', 'prepublish', 'exit')
+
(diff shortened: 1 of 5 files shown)


chmeliik:
@MartinBasti PTAL, instead of a fixture I made a class for mocking workflow

tests/plugins/test_export_operator_manifests.py
```diff
@@ -97,18 +76,24 @@ def mock_env(tmpdir, docker_tasker,
...
-    workflow.user_params['scratch'] = scratch
+
+    env = (MockEnv()
+           .for_plugin('postbuild', ExportOperatorManifestsPlugin.key)
+           .set_scratch(scratch))
+    if orchestrator:
+        env.make_orchestrator()
+
+    env.workflow.builder.set_df_path(str(tmpdir))
+    env.workflow.builder.tasker = flexmock()
```
MartinBasti, line 87: Should be this set as default in MockEnv?

tests/mock_env.py
```diff
@@ -0,0 +1,207 @@
...
+        """
+        Set reactor config map (in the ReactorConfigPlugin's workspace).
+
+        :param config: dict or ReactorConfig, if dict, will be converted to ReactorConfig
+        """
+        if not isinstance(config, ReactorConfig):
+            config = ReactorConfig(config)
+        workspace = self._get_reactor_config_workspace()
+        workspace[WORKSPACE_CONF_KEY] = config
+        return self
```
ben-alkov, line 171: Is this correct? `return self`?

ben-alkov: For method chaining?!

chmeliik: @ben-alkov 

> For method chaining?!

Exactly! This is essentially the Builder pattern, you can do things like:
```python
runner = (MockEnv()
          .set_reactor_config({'version': 1, 'operator_manifests': {}})
          .make_orchestrator()
          .create_runner())
```

All methods except `create_runner()` somehow modify state and return `self`, `create_runner()` is the "finisher" that takes the current state of the instance and creates a plugins runner from it.

chmeliik:
Rebased on master and resolved conflict in test_pin_operator_digest

ben-alkov:
Amazing!

tests/mock_env.py
```diff
@@ -0,0 +1,210 @@
...
+
+        :param docker_tasker: docker_tasker fixture from conftest
+        """
+        if self._phase is None:
+            raise ValueError('No plugin configured (use for_plugin() to configure one)')
+        runner_cls = self._runner_for_phase[self._phase]
+        plugins_conf = getattr(self.workflow, self._plugins_for_phase[self._phase])
+        return runner_cls(docker_tasker, self.workflow, plugins_conf)
+
+    def for_plugin(self, phase, plugin_key, args=None):
```
tkdchen, line 85: Looks it doesn't make sense to call `for_plugin` multiple times, since the `create_runner` only creates a runner for the last phase. If this is the design, can this be documented in the docstring? Or, prevent the second call in the code and raise an error?

tests/mock_env.py
```diff
@@ -0,0 +1,210 @@
...
+        self.workflow.builder.tasker = flexmock()
+
+        self._phase = None
+        self._plugin_key = None
+
+    def create_runner(self, docker_tasker):
+        """
+        Create runner for current plugin (configured using for_plugin())
+
+        :param docker_tasker: docker_tasker fixture from conftest
```
tkdchen, line 77: It would be good the document the return value as well. This method is different from other MockEnv instance methods by not return itself.

tests/mock_env.py
chmeliik: Good idea, added the check and error

tests/mock_env.py
chmeliik: Good point! Added

chmeliik:
Had the last change in the wrong commit, oops (fridays, right?)

chmeliik:
I'm pretty happy with this now, not sure if I should update more tests or if the 4 I have are good enough as proof of concept

chmeliik:
No release notes, internal change for testing only

PR #1477, opened 2020-05-26
Add Image Content Manifest definition

athos-ribeiro:
* CLOUDBLD-1115

### DIFF
--- a/atomic_reactor/schemas/content_manifest.json
+++ b/atomic_reactor/schemas/content_manifest.json
@@ -1,0 +1,82 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Image Content Manifest v1 definition",
+
+  "type": ["object", "null"],
+  "properties": {
+    "metadata": {
+      "description": "Information about the Image Content Manifest (ICM) file",
+      "type": "object",
+      "properties": {
+        "icm_version": {
+          "description": "ICM Version",
+          "type": "integer",
+          "default": 1
+        },
+        "icm_spec": {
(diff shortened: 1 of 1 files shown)


ben-alkov:
@athos-ribeiro; LGTM. Do we have any/require unit tests for something like this?

athos-ribeiro:
> @athos-ribeiro; LGTM. Do we have any/require unit tests for something like this?

We should add some tests when we start using the schema. For now, since we are not validating it, we just need to find a home for the schema

atomic_reactor/schemas/content_manifest.json
```diff
@@ -0,0 +1,82 @@
...
+      "properties": {
+        "icm_version": {
+          "description": "ICM Version",
+          "type": "integer",
+          "default": 1
+        },
+        "icm_spec": {
+          "description": "URL for the ICM specification (this document)",
+          "type": "string",
+          "example": "https://github.com/containerbuildsystem/atomic-reactor/blob/COMMIT_REF/atomic_reactor/schemas/content_manifest.json"
```
chmeliik, line 19: Latest JSON Schema draft defines ["examples"](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.5) as the keyword for examples (and it must be an array), but since we are using draft 4 here (which does not define any example keyword), it is probably fine.

chmeliik:
Waiving test coverage and osbs-docs, assuming both will be added when the schema is actually used somewhere

athos-ribeiro:
No release notes added: we are just shipping the manifest here, which is still not being used and has no effects on atomic-reactor runs.

PR #1483, opened 2020-05-28
allow pulling from external registries

rcerven:
* CLOUDBLD-747

### DIFF
--- a/atomic_reactor/build.py
+++ b/atomic_reactor/build.py
@@ -156,8 +156,9 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
         # arguments for build
         self.source = source
         self.base_image = None
-        self.base_image_insecure = None
-        self.base_image_dockercfg_path = None
+        # configuration of source_registy and pull_registries with insecure and
+        # dockercfg_path, by registry key
+        self.pull_registries = {}
         self.original_base_image = None
         self._base_image_inspect = None
         self.parents_pulled = False
@@ -269,7 +270,7 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
             parent_images[key] = val
         self.parent_images = parent_images
 
-    def set_base_image(self, base_image, parents_pulled=True, insecure=False, dockercfg_path=None):
+    def set_base_image(self, base_image, parents_pulled=True):
         self.base_from_scratch = base_image_is_scratch(base_image)
         if not self.custom_base_image:
             self.custom_base_image = base_image_is_custom(base_image)
@@ -280,8 +281,6 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
         if not self.base_from_scratch:
             self.parent_images[self.original_base_image] = self.base_image
         self.parents_pulled = parents_pulled
-        self.base_image_insecure = insecure
-        self.base_image_dockercfg_path = dockercfg_path
         logger.info("set base image to '%s' with original base '%s'", self.base_image,
                     self.original_base_image)
 
@@ -306,11 +305,13 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
                     # as this property should behave like a dict
                     raise KeyError("Unprocessed base image Dockerfile cannot be inspected")
             else:
+                insecure = self.pull_registries[self.base_image.registry]['insecure']
+                dockercfg_path = self.pull_registries[self.base_image.registry]['dockercfg_path']
                 self._base_image_inspect =\
                     atomic_reactor.util.get_inspect_for_image(self.base_image,
                                                               self.base_image.registry,
-                                                              self.base_image_insecure,
-                                                              self.base_image_dockercfg_path)
+                                                              insecure,
+                                                              dockercfg_path)
 
             base_image_str = str(self.base_image)
             if base_image_str not in self._parent_images_inspect:
@@ -329,11 +330,13 @@ class InsideBuilder(LastLogger, BuilderStateMachine):
             if self.parents_pulled:
                 self._parent_images_inspect[image_name] = self.tasker.inspect_image(image)
             else:
+                insecure = self.pull_registries[image_name.registry]['insecure']
+                dockercfg_path = self.pull_registries[image_name.registry]['dockercfg_path']
                 self._parent_images_inspect[image_name] =\
                     atomic_reactor.util.get_inspect_for_image(image_name,
                                                               image_name.registry,
-                                                              self.base_image_insecure,
-                                                              self.base_image_dockercfg_path)
+                                                              insecure,
+                                                              dockercfg_path)
 
         return self._parent_images_inspect[image_name]
 
--- a/atomic_reactor/plugins/pre_pull_base_image.py
+++ b/atomic_reactor/plugins/pre_pull_base_image.py
@@ -19,33 +19,30 @@ from __future__ import unicode_literals, absolute_import
 import docker
 
 from atomic_reactor.plugin import PreBuildPlugin
-from atomic_reactor.util import (get_build_json, get_manifest_list,
-                                 get_config_from_registry,
-                                 get_platforms, base_image_is_custom,
+from atomic_reactor.util import (get_build_json, get_platforms, base_image_is_custom,
                                  get_checksums, get_manifest_media_type,
-                                 get_all_manifests)
+                                 RegistrySession, RegistryClient)
 from atomic_reactor.core import RetryGeneratorException
 from atomic_reactor.plugins.pre_reactor_config import (get_source_registry,
                                                        get_platform_to_goarch_mapping,
-                                                       get_registries_organization)
+                                                       get_registries_organization,
+                                                       get_pull_registries)
 from io import BytesIO
 from requests.exceptions import HTTPError, RetryError, Timeout
-from osbs.utils import RegistryURI, ImageName
+from osbs.utils import ImageName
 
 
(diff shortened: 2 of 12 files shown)


atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -206,49 +206,36 @@ def _resolve_base_image(self, build_json):
...
             try:
-                self.tasker.pull_image(image, insecure=self.parent_registry_insecure,
-                                       dockercfg_path=self.parent_registry_dockercfg_path)
+                self.tasker.pull_image(image, insecure=reg_client.insecure,
+                                       dockercfg_path=reg_client.dockercfg_path)
                 self.workflow.pulled_base_images.add(image.to_str())
-            except RetryGeneratorException as exc:
-                # getting here means the pull itself failed. we may want to retry if the
-                # image being pulled lacks a namespace, like e.g. "rhel7". we cannot count
-                # on the registry mapping this into the docker standard "library/rhel7" so
```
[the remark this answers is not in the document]
rcerven: that was just fallback to use 'library' I don't think we want that anymore, as we don't want any other magic with FROM

I'll update current commit msg mentioning it

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -19,33 +19,30 @@
...
-                 check_platforms=False, inspect_only=False, parent_images_digests=None):
+    def __init__(self, tasker, workflow, check_platforms=False, inspect_only=False,
+                 parent_images_digests=None):
         """
         constructor
 
         :param tasker: ContainerTasker instance
         :param workflow: DockerBuildWorkflow instance
-        :param parent_registry: registry to enforce pulling from
-        :param parent_registry_insecure: allow connecting to the registry over plain http
```
chmeliik, line 48: I was trying to find where these arguments were passed from and found them in `api.py`, is that OK? Do we want to leave them there?

atomic_reactor/build.py
```diff
@@ -156,8 +156,8 @@ def __init__(self, source, image, **kwargs):
         # arguments for build
         self.source = source
         self.base_image = None
-        self.base_image_insecure = None
-        self.base_image_dockercfg_path = None
+        # configuration of pull_registries with insecure and dockercfg_path, by registry key
+        self.pull_registries = {}
```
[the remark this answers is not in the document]
rcerven: yeah all registries used and allowed (source + pull) are included, I'll add comment

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: I've deleted em from api as well

tests/plugins/test_change_from_in_df.py
```diff
@@ -110,8 +115,8 @@ def test_update_base_image_inspect_broken(tmpdir, caplog, docker_tasker):
     workflow.builder.base_image = image_name
     workflow.builder.set_parent_inspection_data(image_str, dict(no_id="here"))
 
-    with pytest.raises(NoIdInspection):
-        ChangeFromPlugin(docker_tasker, workflow).run()
+    run_plugin(workflow, docker_tasker, allow_failure=True)
+    assert "raised an exception: NoIdInspection" in caplog.text
```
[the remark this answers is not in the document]
rcerven: when we use run_plugin it will use PreBuildPluginsRunner, which runs plugins, catches exceptions and collects result

so here all I get are logs

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -55,23 +52,26 @@ def __init__(self, tasker, workflow, parent_registry=None, parent_registry_insec
...
-            self.parent_registry = source_registry['uri'].docker_uri
-            self.parent_registry_insecure = source_registry['insecure']
-            self.parent_registry_dockercfg_path = source_registry.get('dockercfg_path', None)
-        else:
-            self.parent_registry = None
-            self.parent_registry_insecure = False
-            self.parent_registry_dockercfg_path = None
+
+        self.pull_registries = get_pull_registries(workflow, [])
+        self.source_registry = get_source_registry(self.workflow)['uri'].docker_uri
```
lcarva, line 57: Should this be called `self.source_registry_docker_uri` ?

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -55,23 +52,26 @@ def __init__(self, tasker, workflow, parent_registry=None, parent_registry_insec
...
-        if source_registry.get('uri'):
-            self.parent_registry = source_registry['uri'].docker_uri
-            self.parent_registry_insecure = source_registry['insecure']
-            self.parent_registry_dockercfg_path = source_registry.get('dockercfg_path', None)
-        else:
-            self.parent_registry = None
-            self.parent_registry_insecure = False
-            self.parent_registry_dockercfg_path = None
+
+        self.pull_registries = get_pull_registries(workflow, [])
```
lcarva, line 56: It doesn't look like this should be in `self` since it's only used within this method.

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -55,23 +52,26 @@ def __init__(self, tasker, workflow, parent_registry=None, parent_registry_insec
...
+
         if parent_images_digests:
             metadata = self.workflow.builder.parent_images_digests
             metadata.update(parent_images_digests)
 
         self.manifest_list_cache = {}
+        # RegistryClient instances cached by registry name
+        self.registry_clients = {}
+
+    def _should_enclose(self, image):
```
lcarva, line 70: Consider enhancing `ImageName.enclose` to take an optional registry parameter to perform this logic.

tests/plugins/test_change_from_in_df.py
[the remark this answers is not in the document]
lcarva: Doesn't `PreBuildPluginsRunner` return the result for each plugin though?

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: sure I can rename it

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: good idea will do

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: right I'll delete 'self'

tests/plugins/test_change_from_in_df.py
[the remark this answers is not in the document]
rcerven: yes it returns results, but when plugin fails that result is None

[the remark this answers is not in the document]
chmeliik: `PreBuildPluginsRunner` also re-raises any caught exceptions as `PluginFailedException`, doesn't it? At the end of the day, this doesn't matter that much, but I think it would work.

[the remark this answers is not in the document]
rcerven: change_from_in_df doesn't have is_allowed_to_fail defined, so it default to True, so it won't raise

atomic_reactor/plugins/pre_pull_base_image.py
rcerven: let's leave it be for now, and we will see if we will need it in more plugins

tests/plugins/test_change_from_in_df.py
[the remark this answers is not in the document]
chmeliik: Ah gotcha, makes sense :+1:

rcerven:
> LGTM, but commit message should be updated (see Athos' comment) and an osbs-docs PR might be needed.

commit message already updated, what we would need in osbs-docs?

chmeliik:
> > LGTM, but commit message should be updated (see Athos' comment) and an osbs-docs PR might be needed.
> 
> commit message already updated, what we would need in osbs-docs?

I'm not sure if docs mention anything about *Where you can pull images from*, if not, then maybe an update is not needed. Though it might still be nice to document that you can only pull images from registries that are defined in config map.

rcerven:
> > > LGTM, but commit message should be updated (see Athos' comment) and an osbs-docs PR might be needed.
> > 
> > 
> > commit message already updated, what we would need in osbs-docs?
> 
> I'm not sure if docs mention anything about _Where you can pull images from_, if not, then maybe an update is not needed. Though it might still be nice to document that you can only pull images from registries that are defined in config map.

ic, I'll have to check

rcerven:
> > > > LGTM, but commit message should be updated (see Athos' comment) and an osbs-docs PR might be needed.
> > > 
> > > 
> > > commit message already updated, what we would need in osbs-docs?
> > 
> > 
> > I'm not sure if docs mention anything about _Where you can pull images from_, if not, then maybe an update is not needed. Though it might still be nice to document that you can only pull images from registries that are defined in config map.
> 
> ic, I'll have to check

well I've found both source and pull_registries mentioned in here as sources

https://osbs.readthedocs.io/en/latest/build_parameters.html?highlight=config#reactor-configuration

I am not exactly sure where we could put some new info

rcerven:
rn updated: allow pulling from external registries

PR #1490, opened 2020-06-05
Test internet connection only once

tkdchen:

### DIFF
--- a/tests/util.py
+++ b/tests/util.py
@@ -41,7 +41,7 @@ class InternetConnectionChecker(object):
     def __bool__(self):
         if self._has_conn is None:
             try:
-                requests.get("https://github.com/")
+                requests.get("http://github.com/", allow_redirects=False, timeout=5)
                 self._has_conn = True
             except requests.ConnectionError:
                 self._has_conn = False
--- a/tests/util.py
+++ b/tests/util.py
@@ -32,13 +32,23 @@ def is_string_type(obj):
                for strtype in string_types)
 
 
-def has_connection():
-    try:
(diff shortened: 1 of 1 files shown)


tests/util.py
```diff
@@ -32,13 +32,23 @@ def is_string_type(obj):
...
+                self._has_conn = True
+            except requests.ConnectionError:
+                self._has_conn = False
+        return self._has_conn
+
+has_connection = InternetConnectionChecker()
 
 
 # In case we run tests in an environment without internet connection.
-requires_internet = pytest.mark.skipif(not has_connection(), reason="requires internet connection")
```
chmeliik, line 44: Wait a sec, this was already done only once, no? If I'm not mistaken, the expression is evaluated at import time, and `requires_internet` simply becomes `pytest.mark.skipif(False, reason="requires internet connection")`

Speeding up the check is still valid though.

tests/util.py
```diff
@@ -32,13 +32,23 @@ def is_string_type(obj):
...
+class InternetConnectionChecker(object):
+    """Indicate whether there is a usable Internet connection"""
+
+    def __init__(self):
+        self._has_conn = None
+
+    def __bool__(self):
+        if self._has_conn is None:
+            try:
+                requests.get("http://github.com/", allow_redirects=False, timeout=5)
```
athos-ribeiro, line 44: If we are speeding things up, should we use `head` instead of `get` here?

tkdchen:
Hi, thanks for your review. Since the major change is not valid, I'll close this pr and open another one for the check speed-up.

PR #1491, opened 2020-06-05
Make Internet connection check a little faster

tkdchen:
Testing whether Internet connect is available only cares about if the
remote server is able to response something, but not the concrete
content returned. So, when remote server responses 301, it's enough, no
need to wait for a 200 response with the real content from redirected
URL.

### DIFF
--- a/tests/util.py
+++ b/tests/util.py
@@ -34,7 +34,7 @@ def is_string_type(obj):
 
 def has_connection():
     try:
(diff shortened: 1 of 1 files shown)


tkdchen:
@athos-ribeiro @chmeliik PTAL

athos-ribeiro:
No release notes added. Test suite change.

PR #1495, opened 2020-06-09
Defer Cachito configurations download to workers

athos-ribeiro:
Cachito provided additional remote source configurations may contain large files. Forwarding them as user params from the orchestrator to the workers may result in large chunks of data being passed in the parameters and being appended to the build logs. This may lead to build failures.

Instead, we defer querying the Cachito configurations endpoint to the    worker builds.

* CLOUDBLD-427

This PR also proposes using cachito configurations to downoad the sources in the download remote sources plugin:

If Cachito configuration is set to insecure, i.e., to not verify SSL certificates, that configuration should also be propagated to the download_remote_sources plugin when fetching the source bundles from Cachito.

* CLOUDBLD-203
* CLOUDBLD-427

### DIFF
--- a/atomic_reactor/utils/cachito.py
+++ b/atomic_reactor/utils/cachito.py
@@ -184,21 +184,17 @@ class CachitoAPI(object):
         request_id = self._get_request_id(request)
         return '{}/api/v1/requests/{}/download'.format(self.api_url, request_id)
 
-    def get_request_config(self, request):
-        """Get the configuration files associated with a request
+    def assemble_request_config_url(self, request):
+        """Return the URL to download the configuration files associated with a request
 
         :param request: int or dict, either the Cachito request ID or a dict with 'id' key
 
-        :return: list<dict>, configuration data for the given request.
-                 entries include path, type, and content
+        :return: str, the URL to download the configuration files
         """
         request_id = self._get_request_id(request)
         logger.debug('Retrieving configuration files for request %ds', request_id)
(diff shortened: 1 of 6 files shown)


atomic_reactor/utils/cachito.py
```diff
@@ -184,21 +184,17 @@ def assemble_download_url(self, request):
...
 
         :param request: int or dict, either the Cachito request ID or a dict with 'id' key
 
-        :return: list<dict>, configuration data for the given request.
-                 entries include path, type, and content
+        :return: str, the URL to download the configuration files
         """
         request_id = self._get_request_id(request)
         logger.debug('Retrieving configuration files for request %ds', request_id)
         url = '{}/api/v1/requests/{}/configuration-files'.format(self.api_url, request_id)
```
lcarva, line 196: This URL can be retrieved from the request directly.

athos-ribeiro:
Addressed @lcarva comments in the third commit. Re-requesting reviews.

athos-ribeiro:
No release notes added: Fix for a change included in this release cycle.

athos-ribeiro:
As pointed by Martin, the release cycle has changed and release notes are indeed needed here.

# Bug fixes
 Fetch cachito configurations in worker builds to avoid OCP failures due to long user parameters. 

# Improvements
- allow skipping ssl verifications for cachito downloads
- use cachito provided endpoint for configurations instead of assembling an URL

PR #1498, opened 2020-06-11
Support Cachito environment variables that are not paths

mprahl:
Cachito provides environment variable values that are relative paths or literal values that OSBS sets as build arguments in the user's Dockerfile. For the gomod package manager, all the environment variables were relative paths and OSBS converted them to absolute paths in the container. Cachito recently added support for the npm package manager, which specifies environment variables that are literal values and not relative paths. OSBS needs to detect these values and not convert them to absolute paths.

Eventually, Cachito will add the concept of environment variable types, which will allow OSBS to stop making educated guesses on which values are paths.

Resolves CLOUDBLD-1483

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_remote_source.py
+++ b/atomic_reactor/plugins/pre_resolve_remote_source.py
@@ -8,6 +8,7 @@ of the BSD license. See the LICENSE file for details.
 from __future__ import unicode_literals, absolute_import
 
 import os.path
+import re
 
 from atomic_reactor.constants import PLUGIN_RESOLVE_REMOTE_SOURCE, REMOTE_SOURCE_DIR
 from atomic_reactor.utils.koji import get_koji_task_owner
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_resolve_remote_source.py
```diff
@@ -108,13 +109,23 @@ def run(self):
...
-        build_args = {
-            # Turn the environment variables into absolute paths that
-            # represent where the remote sources are copied to during
-            # the build process.
-            env_var: os.path.join(REMOTE_SOURCE_DIR, value)
-            for env_var, value in source_request.get('environment_variables', {}).items()
-        }
+        build_args = {}
+        # This matches values such as 'deps/gomod' but not 'true'
+        rel_path_regex = re.compile(r'^[^/]+/[^/]+(?:/[^/]+)*$')
```
athos-ribeiro, line 114: Should we keep a list of variables that should be converted to absolute paths or have Cachito to also return additional information (e.g., type) on the env variables? This way we wouldn't need to worry if additional variables are added in the future using a slash for some different reason or if we get a relative path with no nested entries (e.g., converting 'usr' to /usr).

mprahl: @athos-ribeiro the idea is that Cachito will eventually be enhanced to provide the environment variable type. We can discuss this on Monday to determine if I should do that first in Cachito and then patch OSBS the right way or pursue the approach in the PR as a quick fix.

MartinBasti:
I'm not happy to see that atomic-reactor has to guess something

mprahl:
> I'm not happy to see that atomic-reactor has to guess something

I agree, but it's a temporary solution until we teach Cachito to specify the environment variable type. This temporary workaround relieves the pressure to get that done in a tight time frame. I expect it'll be done in the next month though, but can't say for sure due to other higher priority work.

MartinBasti:
ACK then (workarounds everywhere)

mprahl:
@athos-ribeiro I updated the commit message to include the JIRA ticket this resolves.

mprahl:
@MartinBasti and @athos-ribeiro I updated the commit message to be more descriptive as requested. Thanks for the reviews!

athos-ribeiro:
Waiving docs and feature toggle requirements. This is an internal fix for how OSBS handles cachito provided data.

athos-ribeiro:
Release notes added:

- Add support for non-path Cachito environment variables

PR #1499, opened 2020-06-15
Fix Requests Session verify parameter

athos-ribeiro:
Due to bad variable naming, the wrong value was being passed to the
verify session parameter. Let's rename the variables to reflect the
insecure SSL parameter semantics.

CLOUDBLD-203


athos-ribeiro:
No release notes added:

There is already an entry in the bug fixes for the next release:

- Fetch cachito configurations in worker builds to avoid OCP failures due to long user parameters.

PR #1500, opened 2020-06-16
refactor parent images

rcerven:


atomic_reactor/plugins/pre_koji.py
```diff
@@ -66,7 +66,8 @@ def run(self):
             self.log.info('no target provided, skipping plugin')
             return
 
-        if self.workflow.builder.base_from_scratch and not self.workflow.builder.parent_images:
+        if (self.workflow.builder.dockerfile_images.base_from_scratch and
+                not self.workflow.builder.dockerfile_images.keys()):
```
MartinBasti, line 70: `.keys()` are not needed

rcerven: they are needed, because IF we will have just scratch base image
dockerfile_images WILL be defined BUT there won't be any pullable_parents that is why I am checking keys()

MartinBasti: Now I'm curious about that object protocol when `bool()` and `len()` are not tied with `keys()`

MartinBasti:
It's very hard to review this. Too many changes just 1 commit, I'm lost here.

atomic_reactor/plugins/pre_koji.py
rcerven: nvm, you are right, keys() can be removed

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,177 @@ def chain_get(d, path, default=None):
...
+        provides local tag for base_image, or base image with registry and enclosed
+        when we don't know local tag yet
+        """
+        if self.base_from_scratch:
+            return SCRATCH_FROM
+
+        if self._local_parents and self._local_parents[0]:
+            return deepcopy(self._local_parents[0])
+        elif self._pullable_parents and self._pullable_parents[0]:
+            return deepcopy(self._pullable_parents[0])
```
MartinBasti, line 1716: so
```
else:
    return self.base_image_key
```
?

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,177 @@ def chain_get(d, path, default=None):
...
+            if base_image_is_custom(image_name.to_str()):
+                self._custom_parent_image = True
+                if index == 0:
+                    self._custom_base_image = True
+
+            if image_name not in self._pullable_parents:
+                self._pullable_parents.append(image_name)
+                self._local_parents.append(None)
+
+    def set_source_registry_and_org(self, source_registry, organization=None):
```
MartinBasti, line 1821: Methods having `_and_` in name are suspicious. (Single responsibility principle)

You have already properties for source_registry and organization, so this should be split into two setters

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,177 @@ def chain_get(d, path, default=None):
...
+            return SCRATCH_FROM
+
+        if self._pullable_parents and self._pullable_parents[0]:
+            return deepcopy(self._pullable_parents[0])
+        else:
+            raise KeyError('base_image')
+
+    @property
+    def base_from_scratch(self):
+        return self._base_from_scratch
```
MartinBasti, line 1736: Can we have lazy evaluation here?

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,177 @@ def chain_get(d, path, default=None):
...
+
+    @property
+    def custom_base_image(self):
+        return self._custom_base_image
```
MartinBasti, line 1740: Can we have lazy evaluation here?

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,177 @@ def chain_get(d, path, default=None):
...
+
+    @property
+    def custom_parent_image(self):
+        return self._custom_parent_image
```
MartinBasti, line 1744: Can we have lazy evaluation here?

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,177 @@ def chain_get(d, path, default=None):
...
+        provides base image with registry and enclosed
+        (original base image with added registry and enclosed)
+        """
+        if self.base_from_scratch:
+            return SCRATCH_FROM
+
+        if self._pullable_parents and self._pullable_parents[0]:
+            return deepcopy(self._pullable_parents[0])
+        else:
+            raise KeyError('base_image')
```
MartinBasti, line 1732: I'm not sure if KeyError is proper one. Maybe it should return `None` as unset property?

atomic_reactor/util.py
rcerven: value will be set once class is instantiated and it isn't suppose to change

atomic_reactor/util.py
rcerven: value will be set once class is instantiated and it isn't suppose to change

atomic_reactor/util.py
rcerven: value will be set once class is instantiated and it isn't suppose to change

atomic_reactor/util.py
rcerven: reason for having just 1 method is that we will set it just once, so 1 method seemed enough,
but if you insist I can split it

atomic_reactor/util.py
rcerven: well that is basically corner case for Dockerfile without any FROM
so calling 'base_image'/'base_image_key' is basically KeyError
I can change it to different error, but it should raise error for sure

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,177 @@ def chain_get(d, path, default=None):
...
+    def __init__(self, dfp_parents):
+        self._original_parents = deepcopy(dfp_parents) if dfp_parents else []
+        self._custom_parent_image = False
+        self._custom_base_image = False
+        self._base_from_scratch = False
+        self._source_registry = None
+        self._organization = None
+        self._original_base_image = self._original_parents[-1] if self._original_parents else None
+        self._pullable_parents = []
+        self._local_parents = []
```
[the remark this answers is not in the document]
rcerven: sure

atomic_reactor/util.py
rcerven: right will change

atomic_reactor/util.py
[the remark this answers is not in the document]
rcerven: added

atomic_reactor/util.py
ben-alkov: Is there a good name for the combined source_reg+org? If so, then this becomes `set_FOO()`

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,186 @@ def chain_get(d, path, default=None):
...
+    def set_source_registry_and_org(self, source_registry, organization=None):
+        """
+        set source registry and organization
+        has to be done before using with expectation of having
+        already source registry and organization,
+        set in  reactor_config, and in flatpak_create_dockerfile
+        """
+        self._source_registry = source_registry
+        self._organization = organization
+        self._enclose_pullable()
```
chmeliik, line 1839: `set_source_registry_and_org()` is expected to be used at most once, right? If I'm not mistaken, using it a second time would kinda break everything.

Could `source_registry` and `organization` be passed in `__init__()` instead? Or are they not available yet at some point where this class is used?

rcerven: it is expected to be called once as there isn't any other reason to call it with different source & organization
it wouldn't exactly break things, but there isn't reason for it

class is created in inner.py with set_df_path and for flatpak set_df_path is called in flatpak_dockerfile_plugin

source & organization is known in reactor_config plugin

rcerven:
> I'm still not fully on board with what the `DockerfileImages` class does. A dict that dynamically updates its own keys and does magic with `__getitem__` and `__setitem__` sounds dangerous unless its behaviour is very well defined.
> 
> Would it be possible to isolate all the enclosing logic? Quick idea (maybe impossible, I haven't looked at all the use cases): Have a `make_enclosed()` method which takes `registry, organization` and, instead of modifying existing keys, returns a new dict.
> 
> Edit:
> The concerns above are under the assumption that the answer to this is a no:
> 
> > Could `source_registry` and `organization` be passed in `__init__()` instead? Or are they not available yet at some point where this class is used?
> 
> If all the enclosing magic could be done only once in `__init__()`, that would probably be fine.

the whole idea of having this class it to have 1 and only instance which you won't be able to modify apart from
setting source & org
and add local tags 

that way we have 1 instance in whole atomic_reactor

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,186 @@ def chain_get(d, path, default=None):
...
+
+    def _create_pullable(self):
+        for index, image in enumerate(reversed(self._original_parents)):
+            image_name = ImageName.parse(image)
+            if base_image_is_scratch(image_name.get_repo()):
+                if index == 0:
+                    self._base_from_scratch = True
+                continue
+            if base_image_is_custom(image_name.to_str()):
+                self._custom_parent_image = True
```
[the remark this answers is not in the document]
rcerven: I wanted to keep the same names as we had before

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,186 @@ def chain_get(d, path, default=None):
...
+        """
+        provides unmodified base image from Dockerfile
+        """
+        return self._original_base_image
+
+    @property
+    def base_image(self):
+        """
+        provides local tag for base_image, or base image with registry and enclosed
+        when we don't know local tag yet
```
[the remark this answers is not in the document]
rcerven: sure

atomic_reactor/util.py
rcerven: I didn't try to build with Dockerfile without FROM
this is really just corner case, also useful for unit tests

I don't have problem with RuntimeError

chmeliik:
> and will even do enclosing for you

That's the thing though, I have no issue with the class apart from the fact that it also does enclosing. I'm wondering if enclosing couldn't be cleanly separated from the rest of the logic in this class.

But if that is not a viable solution, it might be nice to at least make the `set_source_registry_and_org()` method raise an error if called a second time (that would be an error, wouldn't it?) and document expected usage / behaviour of the class a bit more

athos-ribeiro:
FYI: I tested this in osbs-box with both an external and an osbs-box-built parent image. Basic cases work fine.

I did find an issue when triggering a build from a non existent tag (build fails on OSBS client request with odd error) which seem to be unrelated.

rcerven:
> > and will even do enclosing for you
> 
> That's the thing though, I have no issue with the class apart from the fact that it also does enclosing. I'm wondering if enclosing couldn't be cleanly separated from the rest of the logic in this class.
> 
> But if that is not a viable solution, it might be nice to at least make the `set_source_registry_and_org()` method raise an error if called a second time (that would be an error, wouldn't it?) and document expected usage / behaviour of the class a bit more

having enclosing in there is something we want, so we don't have to worry about it

sure set_source_registry_and_org() can raise error if called 2nd time

rcerven:
> FYI: I tested this in osbs-box with both an external and an osbs-box-built parent image. Basic cases work fine.
> 
> I did find an issue when triggering a build from a non existent tag (build fails on OSBS client request with odd error) which seem to be unrelated.

I did try this with all our dev tests in jenkins

atomic_reactor/util.py
rcerven: the only thing is, that the class init gets list of parents (which are from Dockerfile, but in theory you can just use any list of parents) so that keyerror seems more appropriate

rcerven:
added more docstrings

and also set_source_and_org will raise when called more than once

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,207 @@ def chain_get(d, path, default=None):
...
+        return self._source_registry
+
+    @property
+    def organization(self):
+        return self._organization
+
+    def __len__(self):
+        return len(self._pullable_parents)
+
+    def __setitem__(self, key, item):
```
chmeliik, line 1783: Could we also have docstrings for `__getitem__` and `__setitem__` to explain that both enclosed and un-enclosed keys will work (and under what conditions, e.g. if the key already has a different registry, it won't work)?

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,207 @@ def chain_get(d, path, default=None):
...
+        if not isinstance(key, ImageName):
+            key = ImageName.parse(key)
+
+        if not base_image_is_custom(key.to_str()):
+            if not key.registry:
+                key.registry = self._source_registry
+
+            if self._should_enclose(key):
+                key.enclose(self._organization)
```
chmeliik, line 1792: This is repeated in three places, should it be a separate method (e.g. `_enclose_image()`)?

rcerven: yeah I know, but from this it is more clear that it is conditional

_enclose_image is already saying that it will enclose it soo

atomic_reactor/util.py
rcerven: sure

rcerven: added

atomic_reactor/util.py
rcerven: also it is just 2 lines, which we would substitute for 1, so I'd prefer it to be more readable

atomic_reactor/util.py
rcerven: also looking at code, it will actually fail already in set_df_path, with : Runtimerror when base_image is None

atomic_reactor/util.py
chmeliik: ok, it's fine then

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,225 @@ def chain_get(d, path, default=None):
         except (IndexError, KeyError):
             return default
     return obj
+
+
+class DockerfileImages(object):
+    """
+    _pullable_parents and _local_parents form together dictionary for parent_images
+    they are in reversed order than in Dockerfile (base image first)
```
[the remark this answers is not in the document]
rcerven: sure

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,225 @@ def chain_get(d, path, default=None):
...
+        """
+        return self._original_parents
+
+    @property
+    def original_base_image(self):
+        """
+        provides unmodified base image from Dockerfile
+        :return: str or None
+        """
+        return self._original_base_image
```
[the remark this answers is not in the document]
rcerven: sure

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,225 @@ def chain_get(d, path, default=None):
...
+
+    def set_source_registry_and_org(self, source_registry, organization=None):
+        """
+        set source registry and organization
+        has to be done before using with expectation of having
+        already source registry and organization,
+        set in  reactor_config, and in flatpak_create_dockerfile
+        """
+        if self._source_and_org_set:
+            raise RuntimeError('set_source_registry_and_org is supposed to be called only once')
```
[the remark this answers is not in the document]
rcerven: sure

atomic_reactor/util.py
```diff
@@ -1672,3 +1672,225 @@ def chain_get(d, path, default=None):
...
+        if self.base_from_scratch:
+            return SCRATCH_FROM
+
+        if self._local_parents and self._local_parents[0]:
+            return deepcopy(self._local_parents[0])
+        else:
+            return self.base_image_key
+
+    @property
+    def base_image_key(self):
```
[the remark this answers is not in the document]
rcerven: explained in docstring

atomic_reactor/plugin.py
```diff
@@ -378,9 +378,6 @@ def _translate_special_values(self, obj_to_translate):
             'BUILD_SOURCE_PATH': self.workflow.builder.source.path,
         }
 
-        if self.workflow.builder.base_image:
-            translation_dict['BASE_IMAGE'] = self.workflow.builder.base_image.to_str()
```
chmeliik, line 382: What was this for, and why do we not need it anymore?

docs/koji.md
```diff
@@ -89,9 +89,13 @@ Data which is placed here includes
   integration is enabled
 - `build.extra.image.parent_build_id` (int): Koji build id of the parent image,
   if found
-- `build.extra.image.parent_image_builds` (map of maps): With each parent image
-  given in the Dockerfile as a key, its value being a map representing the Koji
+- `build.extra.image.parent_image_builds` (map of maps): Keys are parent images
+  given in the Dockerfile, which are regular images and
+  not 'scratch' or custom image ('koji/image-build), and includes registry,
+  organization and tag, its value being a map representing the Koji
```
chmeliik, line 95: Is the content of `build.extra.image.parent_image_builds` the same as it was before? If the content changed, is there any risk that some existing automation would be broken by this?

atomic_reactor/plugins/pre_change_from_in_df.py
```diff
@@ -37,6 +33,7 @@ class NoIdInspection(RuntimeError):
 
 class ChangeFromPlugin(PreBuildPlugin):
     key = "change_from_in_dockerfile"
+    is_allowed_to_fail = False
```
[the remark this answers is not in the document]
rcerven: yeah it was mistake, it basically made build fail later

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -649,3 +649,10 @@ def run(self):
         self.workflow.default_image_build_method = default_image_build_method
         self.workflow.builder.tasker.build_method = (source_image_build_method or
                                                      default_image_build_method)
+
+        # set source registry and organization
+        if self.workflow.builder.dockerfile_images is not None:
```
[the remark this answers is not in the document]
rcerven: dockerfile_images are set with set_df_path in InsideBuilder (builder.py) WHEN Dockerfile exists (not-flatpak)
so reactor_config checks for None, and sets source_reg & org in that case

in flatpak_create_dockerfile the Dockerile is created so in that case we call set_df path there which sets dockerfile_images, and then set source_registry & org

atomic_reactor/plugin.py
rcerven: in osbs/inputs in those orchestrator/worker inner files, this MACRO could be used, but we don't used that anymore, as we can access base_image from workflow in every plugin

docs/koji.md
rcerven: content didn't change but I wanted to be more specific because what it said before didn't take in account scratch or custom images

chmeliik: I thought so, just wanted to make sure, thanks

atomic_reactor/plugins/pre_reactor_config.py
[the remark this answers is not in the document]
chmeliik: ic, so basically, only plugins that run before `flatpak_create_dockerfile` have to check for `None` in case build is a flatpak build and Dockerfile does not exist?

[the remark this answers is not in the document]
rcerven: exactly

rcerven:
rn updated: refactor parent images

PR #1501, opened 2020-06-17
Remove expect_v2s2 related checks in tag_and_push plugin

tkdchen:
For now, OSBS will not work if v2s2 is not available. Hence, it makes no
sense to check v2s2 in this plugin.

### DIFF
--- a/atomic_reactor/plugins/post_tag_and_push.py
+++ b/atomic_reactor/plugins/post_tag_and_push.py
@@ -15,8 +15,7 @@ import platform
 import random
 
 from atomic_reactor.constants import (IMAGE_TYPE_DOCKER_ARCHIVE, IMAGE_TYPE_OCI, IMAGE_TYPE_OCI_TAR,
-                                      MEDIA_TYPE_DOCKER_V2_SCHEMA2, DOCKER_PUSH_MAX_RETRIES,
-                                      DOCKER_PUSH_BACKOFF_FACTOR)
+                                      DOCKER_PUSH_MAX_RETRIES, DOCKER_PUSH_BACKOFF_FACTOR)
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.plugins.exit_remove_built_image import defer_removal
 from atomic_reactor.plugins.pre_reactor_config import (get_registries, get_group_manifests,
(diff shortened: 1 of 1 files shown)


tkdchen:
I find the `self.registries` is not modified during the two level for-loops. If anything I missed, please point it out. It's appreciated. Thanks.

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -169,12 +178,6 @@ def run(self):
                 registry_image.registry = registry
                 max_retries = DOCKER_PUSH_MAX_RETRIES
 
-                expect_v2s2 = False
-                for registry in self.registries:
-                    media_types = self.registries[registry].get('expected_media_types', [])
-                    if MEDIA_TYPE_DOCKER_V2_SCHEMA2 in media_types:
-                        expect_v2s2 = True
-
```
chmeliik, line 177: The change seems correct, but why was it done like this in the first place?
If I'm not mistaken, this code essentially says `expect V2_SCHEMA2 if any of the registries have it`. Shouldn't it be `expect V2_SCHEMA2 if the current registry has it`?

tkdchen: This is also the question in my mind.

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -150,9 +150,18 @@ def run(self):
                 self.workflow.tag_conf.add_unique_image(source_unique_image)
             else:
                 self.workflow.tag_conf.add_unique_image(self.workflow.image)
+
         config_manifest_digest = None
         config_manifest_type = None
         config_registry_image = None
+
+        expect_v2s2 = False
+        for _, registry_conf in self.registries.items():
```
ben-alkov, line 159: @tkdchen; Don't you want only `values()` here?

MartinBasti:
Schema 2 v 2 seems to be hard requirement for OSBS nowadays. These are leftovers from old times. Maybe it's worth to open ticket.

MartinBasti:
@tkdchen are you going to continue with this PR or shall we close it?

tkdchen:
@rcerven

> I'd just remove the whole thing
> 
> including :
> if not (self.group or expect_v2s2):
> max_retries = 0

What was removed? These two lines of code still exists in TagAndPushPlugin.

chmeliik:
> @rcerven
> 
> > I'd just remove the whole thing
> > including :
> > if not (self.group or expect_v2s2):
> > max_retries = 0
> 
> What was removed? These two lines of code still exists in TagAndPushPlugin.

@tkdchen I think what Robert is saying is that all the code related to checking expect_v2s2 should just be removed, including the 2 lines that he mentioned

tkdchen:
Thanks for your review. Patch is updated. PTAL.

MartinBasti:
@rcerven PTAL

PR #1503, opened 2020-06-18
Fetch Image Content Manifest from Cachito...

ben-alkov:
...and forward to a new pre-build plugin: `add_image_content_manifest`

- 'pre_resolve_remote_source.py' adds the ICM URL as a param for
  `set_worker_params()`

* CLOUDBLD-1132

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_remote_source.py
+++ b/atomic_reactor/plugins/pre_resolve_remote_source.py
@@ -94,7 +94,9 @@ class ResolveRemoteSourcePlugin(PreBuildPlugin):
         remote_source_json = self.source_request_to_json(source_request)
         remote_source_url = self.cachito_session.assemble_download_url(source_request)
         remote_source_conf_url = remote_source_json.get('configuration_files')
-        self.set_worker_params(source_request, remote_source_url, remote_source_conf_url)
+        remote_source_icm_url = remote_source_json.get('content_manifest')
+        self.set_worker_params(source_request, remote_source_url, remote_source_conf_url,
+                               remote_source_icm_url)
 
         dest_dir = self.workflow.source.workdir
         dest_path = self.cachito_session.download_sources(source_request, dest_dir=dest_dir)
@@ -108,7 +110,8 @@ class ResolveRemoteSourcePlugin(PreBuildPlugin):
             'remote_source_path': dest_path,
         }
 
-    def set_worker_params(self, source_request, remote_source_url, remote_source_conf_url):
+    def set_worker_params(self, source_request, remote_source_url, remote_source_conf_url,
+                          remote_source_icm_url):
         build_args = {}
         # This matches values such as 'deps/gomod' but not 'true'
         rel_path_regex = re.compile(r'^[^/]+/[^/]+(?:/[^/]+)*$')
@@ -129,6 +132,7 @@ class ResolveRemoteSourcePlugin(PreBuildPlugin):
         override_build_kwarg(self.workflow, 'remote_source_url', remote_source_url)
         override_build_kwarg(self.workflow, 'remote_source_build_args', build_args)
         override_build_kwarg(self.workflow, 'remote_source_configs', remote_source_conf_url)
+        override_build_kwarg(self.workflow, 'remote_source_icm_url', remote_source_icm_url)
 
     def source_request_to_json(self, source_request):
         """Create a relevant representation of the source request"""
--- a/tests/plugins/test_resolve_remote_source.py
+++ b/tests/plugins/test_resolve_remote_source.py
@@ -42,7 +42,11 @@ CACHITO_REQUEST_DOWNLOAD_URL = '{}/api/v1/{}/download'.format(CACHITO_URL, CACHI
 CACHITO_REQUEST_CONFIG_URL = '{}/api/v1/requests/{}/configuration-files'.format(
     CACHITO_URL,
     CACHITO_REQUEST_ID
-    )
+)
+CACHITO_ICM_URL = '{}/api/v1/requests/{}/content-manifest'.format(
+    CACHITO_URL,
+    CACHITO_REQUEST_ID
+)
 
 REMOTE_SOURCE_REPO = 'https://git.example.com/team/repo.git'
 REMOTE_SOURCE_REF = 'b55c00f45ec3dfee0c766cea3d395d6e21cc2e5a'
@@ -163,13 +167,14 @@ def mock_cachito_api(workflow, user=KOJI_TASK_OWNER, source_request=None,
             ref=REMOTE_SOURCE_REF,
             user=user,
             dependency_replacements=dependency_replacements,
-        )
+         )
         .and_return({'id': CACHITO_REQUEST_ID}))
 
     (flexmock(CachitoAPI)
         .should_receive('wait_for_request')
         .with_args({'id': CACHITO_REQUEST_ID})
         .and_return(source_request))
+
     (flexmock(CachitoAPI)
         .should_receive('download_sources')
         .with_args(source_request, dest_dir=str(workflow._tmpdir))
@@ -384,5 +389,6 @@ def run_plugin_with_args(workflow, dependency_replacements=None, expect_error=No
             'GOPATH': '/remote-source/deps/gomod',
             'GOCACHE': '/remote-source/deps/gomod',
         }
+        assert worker_params['remote_source_icm_url'] == CACHITO_ICM_URL
 
     return results


atomic_reactor/utils/cachito.py
```diff
@@ -82,7 +83,7 @@ def request_sources(self, repo, ref, flags=None, pkg_managers=None, user=None,
         # Remove None values
         payload = {k: v for k, v in payload.items() if v}
 
-        url = '{}/api/v1/requests'.format(self.api_url)
+        url = self.api_endpoint
         logger.debug('Making request %s with payload:\n%s', url, json.dumps(payload, indent=4))
         response = self.session.post(url, json=payload)
```
athos-ribeiro, line 88: In this case, can we just use the self.api_endpoint in the post call (and in the log msg) and get rid of the attribution to url?

atomic_reactor/utils/cachito.py
```diff
@@ -182,7 +183,28 @@ def assemble_download_url(self, request):
         :return: str, the URL to download the sources
         """
         request_id = self._get_request_id(request)
-        return '{}/api/v1/requests/{}/download'.format(self.api_url, request_id)
+        return '{}/{}/download'.format(self.api_endpoint, request_id)
+
+    def get_icm(self, request):
+        """Get and validate the ICM ("Image Content Manifest") from the Cachito API
+
```
athos-ribeiro, line 190: do we need this blank line here, since the next line is a continuation?

atomic_reactor/utils/cachito.py
```diff
@@ -182,7 +183,28 @@ def assemble_download_url(self, request):
...
+        request_id = self._get_request_id(request)
+        url = '{}/{}/content-manifest'.format(self.api_endpoint, request_id)
+        logger.debug('Making request %s', url)
+        response = self.session.get(url)
+        response.raise_for_status()
+        icm = json.dumps(response.json())
+        logger.debug('Cachito ICM request response:\n%s', icm)
+        # Validate
+        read_yaml(icm, 'schemas/content_manifest.json')
+        return icm
```
athos-ribeiro, line 207: Since we are still going to manipulate the ICM later, should we return the dict here instead? So we won't need to decode it again later.

atomic_reactor/utils/cachito.py
```diff
@@ -182,7 +183,28 @@ def assemble_download_url(self, request):
...
+                        with 'id' key
+
+        :return: str, the ICM as JSON
+        """
+        request_id = self._get_request_id(request)
+        url = '{}/{}/content-manifest'.format(self.api_endpoint, request_id)
+        logger.debug('Making request %s', url)
+        response = self.session.get(url)
+        response.raise_for_status()
+        icm = json.dumps(response.json())
```
chmeliik, line 203: How about `icm = response.text` instead?

atomic_reactor/utils/cachito.py
ben-alkov: Aha. Major question which @rcerven and I discussed. I wasn't certain if the ICM would be further mangled or not :)

ben-alkov:
@athos-ribeiro; Where might I find an example of this?

> feature toggle configuration for this one in the reactor configmap

rcerven:
> It would be nice to have some check in tests/plugins/test_resolve_remote_source.py. For instace, check if the worker build params were overwritten to include the ICM.
> 
> It would also be nice to have the feature toggle configuration for this one in the reactor configmap: Since the API endpoint is still not available in cachito, any integration tests performed with a cachito instance would break if the endpoint is not available.

it doesn't make sense adding feature toggle for this, because IT is dependent on CLOUDBLD-1119
also having feature toggle for 'get_icm' in resolve_remote plugin is weird,
since it isn't even "feature" but just fragment

ben-alkov:
@ athos;

> It would be nice to have some check in tests/plugins/test_resolve_remote_source.py. For instace, check if the worker build params were overwritten to include the ICM.

I've extended the assertions in 'tests/plugins/test_resolve_remote_source.py::run_plugin_with_args()' to include `worker_params['remote_source_content_manifest']` - the tests all pass (woo); would you like to see anything else?

athos-ribeiro:
> e toggle for this, because IT is dependent on CLOUDBLD-1119
> also having feature toggle for 'get_icm' in resolve_remote plugin is weird,

This would simply mean "support cachito >= X" through configuration. It does make sense to me. If it is too much trouble, we can waive that.

@ben-alkov: great job!

atomic_reactor/utils/cachito.py
```diff
@@ -182,7 +183,30 @@ def assemble_download_url(self, request):
...
+    def get_icm(self, request):
+        """
+        Get and validate the ICM from the Cachito API `content-manifest` endpoint.
+
+        N.B. ICM = Image Content Manifest
+
+        :param request: int or dict, either the Cachito request ID or a dict
+                        with 'id' key
+
+        :return: str, the ICM as a Python dict
```
athos-ribeiro, line 197: should it be dict instead of str?

ben-alkov: 😱

ben-alkov:
> If it is too much trouble, we can waive that.

I'd just as soon; this PR is dependent on other work being merged first anyway.

atomic_reactor/plugins/pre_resolve_remote_source.py
```diff
@@ -94,7 +94,9 @@ def run(self):
         remote_source_json = self.source_request_to_json(source_request)
         remote_source_url = self.cachito_session.assemble_download_url(source_request)
         remote_source_conf_url = remote_source_json.get('configuration_files')
-        self.set_worker_params(source_request, remote_source_url, remote_source_conf_url)
+        remote_source_content_manifest = self.cachito_session.get_icm(source_request)
+        self.set_worker_params(source_request, remote_source_url, remote_source_conf_url,
```
chmeliik, line 98: I think this will also require an osbs-client update (add `remote_source_content_manifest` to user_params). It probably doesn't need to be done right away but only when the new plugin is added, is that the plan?

Ideally, `remote_source_content_manifest` should also be added here: [atomic_reactor/schemas/user_params.json](../../tree/master/atomic_reactor/schemas/user_params.json), but since all the other `remote_source_*` params are missing, it would just create an inconsistency.

atomic_reactor/plugins/pre_resolve_remote_source.py
```diff
@@ -129,6 +132,8 @@ def set_worker_params(self, source_request, remote_source_url, remote_source_con
         override_build_kwarg(self.workflow, 'remote_source_url', remote_source_url)
         override_build_kwarg(self.workflow, 'remote_source_build_args', build_args)
         override_build_kwarg(self.workflow, 'remote_source_configs', remote_source_conf_url)
+        override_build_kwarg(self.workflow, 'remote_source_content_manifest',
```
chmeliik, line 135: I think this will also require an osbs-client update (add `remote_source_content_manifest` to user_params). It probably doesn't need to be done right away but only when the new plugin is added, is that the plan?

Ideally, `remote_source_content_manifest` should also be added here: [atomic_reactor/schemas/user_params.json](../../tree/master/atomic_reactor/schemas/user_params.json), but since all the other `remote_source_*` params are missing, it would just create an inconsistency.

tests/utils/test_cachito.py
```diff
@@ -298,3 +298,105 @@ def test_download_sources_bad_request_type(tmpdir):
...
+                },
+                {
+                  "purl": "pkg:golang/golang.org%2Fx%2Ftools@v0.0.0"
+                },
+                {
+                  "purl": "pkg:golang/gopkg.in%2Fcheck.v1@v1.0.0"
+                },
+                {
+                  "purl": "pkg:golang/gopkg.in%2Fyaml.v2@v2.2.2"
+                }
```
tkdchen, line 382: Is it possible to use fewer `purl` in this test data? Fewer lines would make it easier to read.

atomic_reactor/plugins/pre_resolve_remote_source.py
ben-alkov: @chmeliik; Exactly; this won't be merged until other work is completed, so the osbs-client work is contingent upon all of that :)

I'll open a separate PR to add the `remote_source_*` params to the schema

tests/utils/test_cachito.py
ben-alkov: Done

atomic_reactor/plugins/pre_resolve_remote_source.py
```diff
@@ -129,6 +132,7 @@ def set_worker_params(self, source_request, remote_source_url, remote_source_con
         override_build_kwarg(self.workflow, 'remote_source_url', remote_source_url)
         override_build_kwarg(self.workflow, 'remote_source_build_args', build_args)
         override_build_kwarg(self.workflow, 'remote_source_configs', remote_source_conf_url)
+        override_build_kwarg(self.workflow, 'remote_source_request_id', remote_source_request_id)
```
athos-ribeiro, line 135: Could we do it as we do above with the configuration files and pass the content-manifest url? Then the download plugin will not need to deal with the cacito API client again.

The cachito response will also include a `content_manifest` field analogous to the configuration_files one: https://github.com/release-engineering/cachito/pull/187/files#diff-7cbde6013c2753c0bb969ab72bd9d02eR499. You could either start using it now or just build the URL here and we can change it after the cachito PR is merged.

ben-alkov: @athos-ribeiro; I must be missing something. The way I'm reading this, I'm supposed to get the ICM URL from the original request, and then essentially do the work which currently happens in`CachitoAPI::get_icm()` in the 'add_image_content_manifest' plugin instead... otherwise, what am I supposed to do with just the request URL?

*Somewhere* we have to fetch the actual ICM using the 'content_manifest' URL. We can't fetch the *content* and then pass it around (too much data), but it doesn't make sense to me to do the `request` work in the 'add_image_content_manifest' plugin vs. doing it in 'cachito.py' - seems like the actual Cachito API access should happen in the latter...

athos-ribeiro: I am suggesting to do the same thing as the we do for the `remote_source_conf_url`: What we will be passing around is a URL. The `add_image_content_manifest` plugin will not care if it comes from cachito or not. It will just download the contents in that URL and verify it is a valid ICM.

> seems like the actual Cachito API access should happen in the latter...

It is happening there, since the URL will come from the cachito request response.

How does that sound?

PR #1504, opened 2020-06-22
auth: Use retries when querying for token

chmeliik:
* CLOUDBLD-1292

Move get_retrying_requests_session() out of util and into its own module
to avoid a cyclic import between auth and util.

Use a retrying session when querying for token in HTTPBearerAuth.

### DIFF
--- a/atomic_reactor/utils/retries.py
+++ b/atomic_reactor/utils/retries.py
@@ -1,0 +1,60 @@
+"""
+Copyright (c) 2020 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import absolute_import, unicode_literals
+
+import requests
+from requests.adapters import HTTPAdapter
+from urllib3.util import Retry
+
+from atomic_reactor.constants import (HTTP_CLIENT_STATUS_RETRY,
+                                      HTTP_MAX_RETRIES,
+                                      HTTP_BACKOFF_FACTOR,
+                                      HTTP_REQUEST_TIMEOUT)
+
+
+class SessionWithTimeout(requests.Session):
+    """
+    requests Session with added timeout
+    """
+    def __init__(self, *args, **kwargs):
+        super(SessionWithTimeout, self).__init__(*args, **kwargs)
+
+    def request(self, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


tests/test_auth.py
```diff
@@ -204,6 +204,23 @@ def unsupported_callback(request):
...
+        responses.add_callback(responses.GET, url, bearer_unauthorized_callback)
+        responses.add_callback(responses.GET, url, bearer_success_callback)
+
+        auth = HTTPBearerAuth()
+
+        response = requests.get(url, auth=auth)
+        assert response.json() == 'success'
+        assert response.status_code == 200
+        assert len(responses.calls) == 4
+
```
chmeliik, line 223: Test does not work, I suspect that `responses` does not support retries.

Does this change need a test? If so, does anyone have a good way to mock responses for retries?

atomic_reactor/utils/retries.py
```diff
@@ -0,0 +1,60 @@
+"""
```
MartinBasti, line 1: Was this move necessary? It should have also separate test_suite file then.

tests/test_auth.py
MartinBasti: https://github.com/getsentry/responses/issues/135  :(

Probably not

atomic_reactor/utils/retries.py
chmeliik: It was, otherwise there would be a cyclic import between `auth` and `util` (commit message says this too). I will look for relevant tests and move them too.

chmeliik: Plot twist: there were no tests, I added some basic ones (and remove the failing one in `test_auth`)

tests/test_auth.py
chmeliik: Ok, removed test

ben-alkov: Do we care enough about this to use something like 'httpretty' as in e.g. https://github.com/getsentry/responses/issues/135#issuecomment-644384328

chmeliik: IMO the test is not so important that I would add an extra tests dependency, it would essentially be testing 3rd party functionality anyway (retries with `requests` + `urllib3`)

chmeliik:
Release notes: Registry authentication will use retries when querying for bearer token

(probably not needed 'cause it's a minor change)

PR #1505, opened 2020-06-23
Limit maximum image size for regular images before pushing to registry

tkdchen:
* CLOUDBLD-1095

### DIFF
--- a/atomic_reactor/plugins/post_tag_and_push.py
+++ b/atomic_reactor/plugins/post_tag_and_push.py
@@ -21,7 +21,8 @@ from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.plugins.exit_remove_built_image import defer_removal
 from atomic_reactor.plugins.pre_reactor_config import (get_registries, get_group_manifests,
                                                        get_koji_session,
-                                                       get_registries_organization)
+                                                       get_registries_organization,
+                                                       get_image_size_limit)
 from atomic_reactor.plugins.pre_fetch_sources import PLUGIN_FETCH_SOURCES_KEY
 from atomic_reactor.util import (get_manifest_digests, get_config_from_registry, Dockercfg,
                                  get_all_manifests)
@@ -31,6 +32,10 @@ from osbs.constants import RAND_DIGITS
 
 
 __all__ = ('TagAndPushPlugin', )
+
+
+class ExceedsImageSizeError(RuntimeError):
+    """Error of exceeding image size"""
 
 
 class TagAndPushPlugin(PostBuildPlugin):
@@ -149,9 +154,12 @@ class TagAndPushPlugin(PostBuildPlugin):
                 self.workflow.tag_conf.add_unique_image(source_unique_image)
             else:
                 self.workflow.tag_conf.add_unique_image(self.workflow.image)
+
         config_manifest_digest = None
         config_manifest_type = None
         config_registry_image = None
+        image_size_limit = get_image_size_limit(self.workflow)
+
         for registry, registry_conf in self.registries.items():
             insecure = registry_conf.get('insecure', False)
             push_conf_registry = \
@@ -163,6 +171,17 @@ class TagAndPushPlugin(PostBuildPlugin):
             for image in self.workflow.tag_conf.images:
                 if image.registry:
                     raise RuntimeError("Image name must not contain registry: %r" % image.registry)
+
+                if not source_oci_image_path:
+                    image_size = sum(item['size'] for item in self.workflow.layer_sizes)
+                    config_image_size = image_size_limit['binary_image']
+                    # Only handle the case when size is set > 0 in config
+                    if config_image_size and image_size > config_image_size:
+                        raise ExceedsImageSizeError(
+                            'The size {} of image {} exceeds the limitation {} '
+                            'configured in reactor config.'
+                            .format(image_size, image, image_size_limit)
+                        )
 
                 registry_image = image.copy()
                 registry_image.registry = registry
--- a/atomic_reactor/plugins/pre_reactor_config.py
+++ b/atomic_reactor/plugins/pre_reactor_config.py
@@ -439,6 +439,13 @@ def get_operator_manifests(workflow, fallback=NO_FALLBACK):
     return get_value(workflow, 'operator_manifests', fallback)
 
 
+def get_image_size_limit(workflow):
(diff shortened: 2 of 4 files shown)


atomic_reactor/schemas/config.json
```diff
@@ -734,6 +734,18 @@
         "additionalProperties": false,
         "required": ["name", "value"]
       }
+    },
+    "image_size_limitation": {
```
chmeliik, line 738: I'd prefer `image_size_limit` as the name for this.

To me, "limitation" sounds like something that is out of your control (technological limitation etc.).

atomic_reactor/schemas/config.json
```diff
@@ -734,6 +734,18 @@
...
+      "description": "Define maximum image size for regular image checked before push",
```
chmeliik, line 739: It would be nice to document the unit of size (is it bytes?)

atomic_reactor/schemas/config.json
```diff
@@ -734,6 +734,18 @@
...
+      "type": "object",
+      "properties": {
+        "regular_image": {
+          "description": "Maximum size of regular images. Set to 0 or omit to skip the check",
```
chmeliik, line 743: IMO 0 to skip check is kinda confusing, omit to skip should be good enough. If someone sets 0, build should probably just fail when checking size.

MartinBasti: In configs I'm using, 0 usually means unlimited.

atomic_reactor/schemas/config.json
MartinBasti: And mention it's uncompressed image size.

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -33,6 +34,10 @@
 __all__ = ('TagAndPushPlugin', )
 
 
+class ExceedImageSizeError(Exception):
```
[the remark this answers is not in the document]
tkdchen: Or `ExceedsImageSizeError`? Not sure as well :) Any idea from anyone else? Thanks.

atomic_reactor/schemas/config.json
tkdchen: Done

atomic_reactor/schemas/config.json
tkdchen: How does OSBS handle such value in a consistent way? On the other hand, IMO, if 0 makes no sense for this case, it could be easy to set minimum value to 1 in this schema to validate it.

atomic_reactor/plugins/post_tag_and_push.py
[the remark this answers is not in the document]
tkdchen: `RuntimeError` is the base class now.

atomic_reactor/schemas/config.json
chmeliik: Not sure about OSBS, but in python, unlimited is usually indicated with `-1`. I don't really mind though, if 0 for unlimited is common then we can keep it this way.

Setting minimum to 1 probably does not make that much sense, in practice both 1 and 0 would make all builds fail anyway.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -439,6 +439,13 @@ def get_operator_manifests(workflow, fallback=NO_FALLBACK):
     return get_value(workflow, 'operator_manifests', fallback)
 
 
+def get_image_size_limitation(workflow):
+    config = get_value(workflow, 'image_size_limitation', {})
+    return {
+        'regular_image': config.get('regular_image', 0),
+    }
```
chmeliik, line 446: Should be renamed to `limit` as well

tkdchen: @chmeliik Fixed this and other variable names having "limitation". Thanks!

atomic_reactor/plugins/post_tag_and_push.py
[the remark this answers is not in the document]
tkdchen: Done. Use name `ExceedsImageSizeError`.

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -165,6 +173,17 @@ def run(self):
                 if image.registry:
                     raise RuntimeError("Image name must not contain registry: %r" % image.registry)
 
+                if not source_oci_image_path:
```
chmeliik, line 176: What does this condition mean? I'm not familiar with this part of the code

MartinBasti: source container images are not checked (they are .... different)

chmeliik: Ah, so `source_oci_image_path` == `it is a source container build`?

MartinBasti:
This feature should be also documented in https://github.com/containerbuildsystem/osbs-docs/blob/master/docs/admins.rst

atomic_reactor/plugins/post_tag_and_push.py
MartinBasti: yes

atomic_reactor/schemas/config.json
athos-ribeiro: IMO, 0 does make sense for meaning unlimited here.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -439,6 +439,13 @@ def get_operator_manifests(workflow, fallback=NO_FALLBACK):
     return get_value(workflow, 'operator_manifests', fallback)
 
 
+def get_image_size_limit(workflow):
+    config = get_value(workflow, 'image_size_limit', {})
+    return {
+        'regular_image': config.get('regular_image', 0),
```
athos-ribeiro, line 445: Are we use regular to mean binary here?

If so, we should rename it to binary_image, since this is the name we have been using everywhere else to mean non-source-images.

tkdchen: Done.

tkdchen:
> This feature should be also documented in https://github.com/containerbuildsystem/osbs-docs/blob/master/docs/admins.rst

@MartinBasti Which part of this doc is proper to document this feature?

MartinBasti:
Release notes:
- - added ability to limit max size of binary images via reactor config map

PR #1506, opened 2020-06-23
Modular koji tags

rcerven:

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -56,7 +56,8 @@ class ResolveComposesPlugin(PreBuildPlugin):
         """
         :param tasker: ContainerTasker instance
         :param workflow: DockerBuildWorkflow instance
-        :param koji_target: str, contains build tag to be used when requesting compose from "tag"
+        :param koji_target: str, koji target contains build tag to be used
+                            when requesting compose from "tag"
         :param signing_intent: override the signing intent from git repo configuration
         :param compose_ids: use the given compose_ids instead of requesting a new one
         :param repourls: list of str, URLs to the repo files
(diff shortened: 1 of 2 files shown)


rcerven:
related osbs-client PR https://github.com/containerbuildsystem/osbs-client/pull/988

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -416,7 +417,13 @@ def __init__(self, data, pulp_data, odcs_config, koji_tag=None, arches=None):
         self.arches = arches or []
         self.multilib_arches = []
         self.multilib_method = None
-        self.modular_tags = data.get('modular_koji_tags', [])
+        self.modular_tags = data.get('modular_koji_tags')
+        self.koji_tag = koji_tag
+
+        if self.modular_tags is True:
```
tkdchen, line 423: Just `if self.modular_tags:` is enough?

chmeliik: From the related osbs-client PR (https://github.com/containerbuildsystem/osbs-client/pull/988) it seems that `modular_koji_tags` may use different types to mean different things.

My question is: what exactly does `True` mean here? I couldn't find details in the related JIRA issue.

chmeliik: > My question is: what exactly does `True` mean here? I couldn't find details in the related JIRA issue.


Nevermind, found it:

> we can support modular_koji_tags: true in osbs by simply passing the buildroot tag to ODCS

^ is that still valid / correct? IIUC we are passing the build target tag, not the buildroot tag here.

rcerven: buildroot target in this context means koji buildroot for target, don't have it confused with OUR buildroot :)

rcerven: and yes we support either True or list

in case we have list we will use that as modular_koji_tags

if it is True we will use tag from target in list

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -416,7 +417,13 @@ def __init__(self, data, pulp_data, odcs_config, koji_tag=None, arches=None):
...
+            if not self.koji_tag:
+                raise ValueError('koji_tag is required when modular_koji_tags is True')
```
athos-ribeiro, line 425: Could we say that a koji target is required (this error msg will be displayed to the user, and `koji_tag` won't mean much in this sense)

atomic_reactor/plugins/pre_resolve_composes.py
rcerven: updated docstr

atomic_reactor/plugins/pre_resolve_composes.py
rcerven: well the same message is in validation

thing is, there could be koji target, without tag I'd say soooo

rcerven: also this can happen only when using osbs-client directly

rcerven:
tests need that schema change in osbs-client PR

rcerven:
rn updated: forward "modular_koji_tags" to odcs

PR #1510, opened 2020-06-29
 Implement new 'add_image_content_manifest' pre-build plugin

ben-alkov:
- Loosely based on, and deprecates, `add_content_sets` pre-build
  plugin
- Receives 'remote_source_icm_url' from the 'resolve_remote_source'
  pre-build plugin
- Downloads the ICM using above
- Injects current layer number, and existing 'content_sets' metadata
  from dist git, into the ICM
- Validates the updated ICM
- Writes the ICM into the same directory as the Dockerfile
- Adds a COPY line to the Dockerfile, to put the ICM into the
  resulting container.

* CLOUDBLD-1133

### DIFF
--- a/atomic_reactor/plugins/pre_add_image_content_manifest.py
+++ b/atomic_reactor/plugins/pre_add_image_content_manifest.py
@@ -1,0 +1,214 @@
+"""
+Copyright (c) 2020 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import absolute_import, unicode_literals
+
+import json
+import os
+
+from copy import deepcopy
+
+from osbs.utils import Labels
+
+from atomic_reactor.constants import (IMAGE_BUILD_INFO_DIR, INSPECT_ROOTFS,
+                                      INSPECT_ROOTFS_LAYERS,
+                                      PLUGIN_ADD_IMAGE_CONTENT_MANIFEST,
+                                      REPO_CONTENT_SETS_CONFIG)
+from atomic_reactor.plugin import PreBuildPlugin
+from atomic_reactor.plugins.pre_reactor_config import get_cachito
+from atomic_reactor.util import (base_image_is_scratch, df_parser, read_yaml,
+                                 read_yaml_from_file_path, get_retrying_requests_session,
+                                 )
+
+
+class AddImageContentManifestPlugin(PreBuildPlugin):
+    """
+    Add the ICM JSON file to the IMAGE_BUILD_INFO_DIR/content_manifests
+    directory, for the current platform. Filename will be '{IMAGE_NVR}.json'
+
+    ICM examples:
+
+    WITHOUT content_sets specified:
+
+    {
+      "metadata": {
+        "icm_version": 1,
+        "icm_spec": "https://link.to.icm.specification",
+        "image_layer_index": 3
+      },
+      "content_sets" : [],
+      "image_contents": [
+        {
+          "purl": "pkg:golang/github.com%2Frelease-engineering%2Fretrodep%2Fv2@v2.0.2",
+          "dependencies": [{"purl": "pkg:golang/github.com%2Fop%2Fgo-logging@v0.0.0"}],
+          "sources": [{"purl": "pkg:golang/github.com%2FMasterminds%2Fsemver@v1.4.2"}]
+        }
+      ]
+    }
+
+    WITH content_sets specified:
+
+    {
+      "metadata": {
+        "icm_version": 1,
+        "icm_spec": "https://link.to.icm.specification",
+        "image_layer_index": 2
+      },
+      "content_sets": [
+          "rhel-8-for-x86_64-baseos-rpms",
+          "rhel-8-for-x86_64-appstream-rpms"
+      ],
+      "image_contents": [
+        {
+          "purl": "pkg:golang/github.com%2Frelease-engineering%2Fretrodep%2Fv2@v2.0.2",
+          "dependencies": [{"purl": "pkg:golang/github.com%2Fop%2Fgo-logging@v0.0.0"}],
+          "sources": [{"purl": "pkg:golang/github.com%2FMasterminds%2Fsemver@v1.4.2"}]
+        }
+      ]
+    }
+    """
+    key = PLUGIN_ADD_IMAGE_CONTENT_MANIFEST
+    is_allowed_to_fail = False
+    minimal_icm = {
+        'metadata': {
+            'icm_version': 1,
+            'icm_spec': ('https://raw.githubusercontent.com/containerbuildsystem/atomic-reactor/'
+                         'master/atomic_reactor/schemas/content_manifest.json'),
+            'image_layer_index': 1
+        },
+        'content_sets': [],
+        'image_contents': [],
+    }
+
+    def __init__(self, tasker, workflow, remote_source_icm_url=None, destdir=IMAGE_BUILD_INFO_DIR):
+        """
+        :param tasker: ContainerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param icm_url: str, URL of the ICM from the Cachito request.
+        :param destdir: image path to carry content_manifests data dir
+        """
+        super(AddImageContentManifestPlugin, self).__init__(tasker, workflow)
+        self.content_manifests_dir = os.path.join(destdir, 'content_manifests')
+        self.icm_url = remote_source_icm_url
+        self.dfp = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        labels = Labels(self.dfp.labels)
+        _, image_name = labels.get_name_and_value(Labels.LABEL_TYPE_COMPONENT)
+        _, image_version = labels.get_name_and_value(Labels.LABEL_TYPE_VERSION)
+        _, image_release = labels.get_name_and_value(Labels.LABEL_TYPE_RELEASE)
+        self.icm_file_name = '{}-{}-{}.json'.format(image_name, image_version, image_release)
+        self.content_sets = []
+        self._cachito_verify = None
+        self._layer_index = None
+        self._icm = None
+
+    @property
+    def layer_index(self):
+        if self._layer_index is None:
+            # Default layer index is 1, because base and 'FROM scratch' images
+            #     *always* have 2 layers
+            self._layer_index = 1
+        if not base_image_is_scratch(self.dfp.baseimage):
+            inspect = self.workflow.builder.base_image_inspect
+            self._layer_index = len(inspect[INSPECT_ROOTFS][INSPECT_ROOTFS_LAYERS])
+        return self._layer_index
+
+    @property
+    def cachito_verify(self):
+        if self._cachito_verify is None:
+            try:
+                cachito_conf = get_cachito(self.workflow)
+            except KeyError:
+                cachito_conf = {}
+            # Get the value of Cachito's 'insecure' key from the active reactor config map,
+            #    *flip it*, and let the result tell us whether to verify or not
+            self._cachito_verify = not cachito_conf.get('insecure', False)
+        return self._cachito_verify
+
+    @property
+    def icm(self):
+        """
+        Get and validate the ICM from the Cachito API `content-manifest` endpoint.
+
+        :return: dict, the ICM as a Python dict
+        """
+        if self.icm_url is None and self._icm is None:
+            self._icm = deepcopy(self.minimal_icm)
+        if self._icm is None:
+            session = get_retrying_requests_session()
+            session.verify = self.cachito_verify
+            self.log.debug('Making request to "%s"', self.icm_url)
+            response = session.get(self.icm_url)
+            response.raise_for_status()
+            self._icm = response.json()  # Returns dict
+            self.log.debug('Cachito ICM request response:\n%s', self._icm)
+
+            # Validate; `json.dumps()` converts `icm` to str. Confusingly, `read_yaml`
+            #     *will* validate JSON
+            read_yaml(json.dumps(self._icm), 'schemas/content_manifest.json')
+        return self._icm
+
+    def _populate_content_sets(self):
+        """
+        Get the list of the current platform's content_sets from
+        'content_sets.yml' in dist-git, and set `self.content_sets` to same.
+        """
+        current_platform = self.workflow.user_params['platform']
+        workdir = self.workflow.builder.df_dir
+        fpath = os.path.join(workdir, REPO_CONTENT_SETS_CONFIG)
+        if os.path.exists(fpath):
+            content_sets_yml = read_yaml_from_file_path(fpath, 'schemas/content_sets.json') or {}
+            if current_platform in content_sets_yml:
+                self.content_sets = content_sets_yml[current_platform]
+        self.log.debug('Output content_sets: %s', self.content_sets)
+
+    def _update_icm_data(self):
+        # Inject the content_sets data into the ICM JSON object
+        self.icm['content_sets'] = self.content_sets
+
+        # Inject the current image layer index number into the ICM JSON object metadata
+        self.icm['metadata']['image_layer_index'] = self.layer_index
+
+        # Convert dict -> str
+        icm_json = json.dumps(self.icm, indent=4)
+
+        # Validate the updated ICM with the ICM JSON Schema
+        read_yaml(icm_json, 'schemas/content_manifest.json')
+
+        self.log.debug('Output ICM: %s', icm_json)
+
+    def _write_json_file(self):
+        out_file_path = os.path.join(self.workflow.builder.df_dir,
+                                     self.icm_file_name)
+        if os.path.exists(out_file_path):
+            raise RuntimeError('File {} already exists in repo'.format(out_file_path))
+        with open(out_file_path, 'w') as outfile:
+            json.dump(self.icm, outfile, indent=4)
+        self.log.debug('ICM JSON saved to: %s', out_file_path)
+
+    def _add_to_dockerfile(self):
+        """
+        Put an ADD instruction into the Dockerfile (to include the ICM file
+        into the container image to be built)
+        """
+        dest_file_path = os.path.join(self.content_manifests_dir, self.icm_file_name)
+        content = 'ADD {0} {1}'.format(self.icm_file_name, dest_file_path)
+        lines = self.dfp.lines
+
+        # Put it before last instruction
+        lines.insert(-1, content + '\n')
+        self.dfp.lines = lines
+
+    def run(self):
+        """
+        run the plugin
+        """
+        self._populate_content_sets()
+        self._update_icm_data()
+        self._write_json_file()
+        self._add_to_dockerfile()
+        self.log.info('added "%s" to "%s"', self.icm_file_name, self.content_manifests_dir)
--- a/tests/plugins/test_add_image_content_manifest.py
+++ b/tests/plugins/test_add_image_content_manifest.py
@@ -1,0 +1,300 @@
+"""
+Copyright (c) 2020 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from __future__ import absolute_import, print_function, unicode_literals
+
+import json
+import os
+import sys
+
+from copy import deepcopy
+from textwrap import dedent
+
+import pytest
+import yaml
+
+from tests.mock_env import MockEnv
+from tests.utils.test_cachito import CACHITO_URL, CACHITO_REQUEST_ID
+
+from atomic_reactor import util
+from atomic_reactor.constants import INSPECT_ROOTFS, INSPECT_ROOTFS_LAYERS
+from atomic_reactor.plugin import PluginFailedException
+from atomic_reactor.plugins.pre_add_image_content_manifest import AddImageContentManifestPlugin
+
+CONTENT_SETS = {
+    'x86_64': ['pulp-spamx86-rpms', 'pulp-baconx86-rpms'],
+    'ppc64': ['pulp-spamppc64-rpms', 'pulp-baconppc64-rpms'],
+    's390x': ['pulp-spams390x-rpms', 'pulp-bacons390x-rpms'],
+}
+CACHITO_ICM_URL = '{}/api/v1/requests/{}/content-manifest'.format(CACHITO_URL,
+                                                                  CACHITO_REQUEST_ID)
+ICM_MINIMAL_DICT = {
+    'metadata': {
+        'icm_version': 1,
+        'icm_spec': ('https://raw.githubusercontent.com/containerbuildsystem/atomic-reactor/'
+                     'master/atomic_reactor/schemas/content_manifest.json'),
+        'image_layer_index': 1
+    },
+    'content_sets': [],
+    'image_contents': [],
+}
+ICM_DICT = {
+    'metadata': {
+        'icm_version': 1,
+        'icm_spec': 'https://link.to.icm.specification',
+        'image_layer_index': 1,
+    },
+    'content_sets': [],
+    'image_contents': [
+        {
+            'purl':
+            'pkg:golang/github.com%2Frelease-engineering%2Fretrodep%2Fv2@v2.0.2',
+            'dependencies': [
+                {
+                    'purl':
+                    'pkg:golang/github.com%2Fop%2Fgo-logging@v0.0.0'
+                },
+            ],
+            'sources': [
+                {
+                    'purl':
+                    'pkg:golang/github.com%2FMasterminds%2Fsemver@v1.4.2'
+                },
+            ]
+        },
+    ]
+}
+ICM_JSON = dedent(
+    '''\
+    {
+        "metadata": {
+        "icm_version": 1,
+        "icm_spec": "https://link.to.icm.specification",
+        "image_layer_index": 1
+        },
+        "content_sets": [],
+        "image_contents": [
+        {
+            "purl": "pkg:golang/github.com%2Frelease-engineering%2Fretrodep%2Fv2@v2.0.2",
+            "dependencies": [
+            {
+                "purl": "pkg:golang/github.com%2Fop%2Fgo-logging@v0.0.0"
+            }
+            ],
+            "sources": [
+            {
+                "purl": "pkg:golang/github.com%2FMasterminds%2Fsemver@v1.4.2"
+            }
+            ]
+        }
+        ]
+    }
+    '''
+)
+
+
+def setup_function(*args):
+    # IMPORTANT: This needs to be done to ensure mocks at the module
+    # level are reset between test cases.
+    sys.modules.pop('pre_add_image_content_manifest', None)
+
+
+def teardown_function(*args):
+    # IMPORTANT: This needs to be done to ensure mocks at the module
+    # level are reset between test cases.
+    sys.modules.pop('pre_add_image_content_manifest', None)
+
+
+def mock_content_sets_config(tmpdir, empty=False):
+    content_dict = {}
+    if not empty:
+        for arch, repos in CONTENT_SETS.items():
+            content_dict[arch] = repos
+    tmpdir.join('content_sets.yml').write(yaml.safe_dump(content_dict))
+
+
+def mock_get_icm(requests_mock):
+    requests_mock.register_uri('GET', CACHITO_ICM_URL, text=ICM_JSON)
+
+
+def mock_env(tmpdir, docker_tasker, platform='x86_64', base_layers=0,
+             icm_url=CACHITO_ICM_URL, r_c_m_override=None,
+             ):
+    inspection_data = {
+        INSPECT_ROOTFS: {
+            INSPECT_ROOTFS_LAYERS: list(range(base_layers))
+        }
+    }
+    if r_c_m_override is None:
+        r_c_m = {
+            'version': 1,
+            'cachito': {
+                'api_url': CACHITO_URL,
+                'auth': {
+                    'ssl_certs_dir': str(tmpdir),
+                },
+            },
+        }
+    else:
+        r_c_m = r_c_m_override
+    env = (MockEnv()
+           .for_plugin('prebuild', AddImageContentManifestPlugin.key,
+                       {'remote_source_icm_url': icm_url})
+           .set_reactor_config(r_c_m)
+           .make_orchestrator()
+           )
+    tmpdir.join('cert').write('')
+    env.workflow.builder.set_inspection_data(inspection_data)
+    env.workflow.user_params['platform'] = platform
+
+    return env.create_runner(docker_tasker)
+
+
+@pytest.mark.parametrize('manifest_file_exists', [True, False])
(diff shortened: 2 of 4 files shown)


ben-alkov:
Draft (mostly) because it still needs unit tests.

ben-alkov:
New unit tests will not pass until #1503 is merged.

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,159 @@
...
+    def _write_json_file(self):
+        if os.path.exists(self.out_file_path):
+            raise RuntimeError('File {} already exists in repo'.format(self.out_file_path))
+        with open(self.out_file_path, 'w') as outfile:
+            json.dump(self.icm, outfile, indent=4)
+        self.log.debug('ICM JSON saved to: %s', self.out_file_path)
+
+    def _add_to_dockerfile(self):
+        """
+        Put a COPY/ADD instruction into the Dockerfile (to include the ICM file
```
rcerven, line 140: remove : COPY/

tests/plugins/test_add_image_content_manifest.py
```diff
@@ -0,0 +1,321 @@
...
+from atomic_reactor.plugin import PluginFailedException, PreBuildPluginsRunner
+from atomic_reactor.plugins import pre_reactor_config
+from atomic_reactor.plugins.pre_add_image_content_manifest import AddImageContentManifestPlugin
+from atomic_reactor.plugins.pre_reactor_config import (WORKSPACE_CONF_KEY,
+                                                       ReactorConfig,
+                                                       ReactorConfigPlugin)
+from atomic_reactor.source import SourceConfig
+from atomic_reactor.utils.cachito import CachitoAPI
+
+PULP_MAPPING = {
```
rcerven, line 40: CONTENT_SETS

tests/plugins/test_add_image_content_manifest.py
```diff
@@ -0,0 +1,321 @@
...
+    builder = StubInsideBuilder().for_workflow(workflow)
+    builder.set_df_path(str(tmpdir))
+    builder.tasker = flexmock()
+    workflow.builder = flexmock(builder)
+    workflow.buildstep_plugins_conf = [{'name': PLUGIN_BUILD_ORCHESTRATE_KEY}]
+    mock_repo_config(workflow)
+    mock_reactor_config(workflow)
+    mock_build_json()
+    mock_cachito_api(workflow)
+    mock_koji()
```
rcerven, line 200: why are you mocking koji, when plugin isn't using koji?

tests/plugins/test_add_image_content_manifest.py
```diff
@@ -0,0 +1,321 @@
...
+            self.workdir = workdir
+
+    workflow.source = MockSource(str(tmpdir))
+    builder = StubInsideBuilder().for_workflow(workflow)
+    builder.set_df_path(str(tmpdir))
+    builder.tasker = flexmock()
+    workflow.builder = flexmock(builder)
+    workflow.buildstep_plugins_conf = [{'name': PLUGIN_BUILD_ORCHESTRATE_KEY}]
+    mock_repo_config(workflow)
+    mock_reactor_config(workflow)
```
rcerven, line 197: you don't need reactor_config_map

tests/plugins/test_add_image_content_manifest.py
```diff
@@ -0,0 +1,321 @@
...
+    mock_build_json()
```
rcerven, line 198: you don't need build_json

tests/plugins/test_add_image_content_manifest.py
```diff
@@ -0,0 +1,321 @@
...
+    mock_cachito_api(workflow)
```
rcerven, line 199: you don't need cachito

tests/plugins/test_add_image_content_manifest.py
```diff
@@ -0,0 +1,321 @@
...
+            super(MockSource, self).__init__()
+            self.workdir = workdir
+
+    workflow.source = MockSource(str(tmpdir))
+    builder = StubInsideBuilder().for_workflow(workflow)
+    builder.set_df_path(str(tmpdir))
+    builder.tasker = flexmock()
+    workflow.builder = flexmock(builder)
+    workflow.buildstep_plugins_conf = [{'name': PLUGIN_BUILD_ORCHESTRATE_KEY}]
+    mock_repo_config(workflow)
```
rcerven, line 196: you don't need container.yaml

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,159 @@
...
+          "dependencies": [{"purl": "pkg:golang/github.com%2Fop%2Fgo-logging@v0.0.0"}],
+          "sources": [{"purl": "pkg:golang/github.com%2FMasterminds%2Fsemver@v1.4.2"}]
+        }
+      ]
+    }
+    """
+    key = PLUGIN_ADD_IMAGE_CONTENT_MANIFEST
+    is_allowed_to_fail = False
+
+    def __init__(self, tasker, workflow, remote_source_content_manifest,
```
MartinBasti, line 73: `remote_source_content_manifest` could be this too big to be transferred to workers via buildconfig? What is OCP limit?

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,159 @@
...
+                 destdir=IMAGE_BUILD_INFO_DIR):
```
MartinBasti, line 74: Will osbs-client specify location?

ben-alkov: Don't know. @athos-ribeiro?

atomic_reactor/plugins/pre_add_image_content_manifest.py
ben-alkov: @MartinBasti; This is the design I worked from... @athos-ribeiro, @rcerven?

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,159 @@
...
+        self.icm_dir = os.path.join(destdir, 'content_manifests')
+        self.icm = remote_source_content_manifest
+        self.dfp = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        labels = Labels(self.dfp.labels)
+        _, image_name = labels.get_name_and_value(Labels.LABEL_TYPE_COMPONENT)
+        _, image_version = labels.get_name_and_value(Labels.LABEL_TYPE_VERSION)
+        _, image_release = labels.get_name_and_value(Labels.LABEL_TYPE_RELEASE)
+        self.icm_file_name = '{}-{}-{}.json'.format(image_name, image_version, image_release)
+        self.out_file_path = os.path.join(self.workflow.builder.df_dir,
+                                          self.icm_file_name)
```
tkdchen, line 91: This icm filename is only used in method `_write_json_file`. How about just define this variable in that method?

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # Convert dict -> str
+        icm_json = json.dumps(self.icm, indent=4)
+
+        # Validate the ICM with the ICM JSON Schema
+        read_yaml(icm_json, 'schemas/content_manifest.json')
+
+        self.log.debug('Output ICM: %s', icm_json)
+
+    def _write_json_file(self):
```
tkdchen, line 131: `AddContentSetsPlugin` has same method. Reuse the code?

atomic_reactor/plugins/pre_add_image_content_manifest.py
athos-ribeiro: In theory, the file could actualy be quite large. Thanks for spotting this, @MartinBasti.

@ben-alkov, we should use the same approach as the one used in https://github.com/containerbuildsystem/atomic-reactor/pull/1495 and defer downloading the ICM here in the worker; i.e., https://github.com/containerbuildsystem/atomic-reactor/pull/1503 should be changed to forward the ICM URL instead.

atomic_reactor/plugins/pre_add_image_content_manifest.py
athos-ribeiro: No. This signature was designed by @rcerven to allow further customization in the destdir if needed.

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        super(AddImageContentManifestPlugin, self).__init__(tasker, workflow)
+        self.icm_dir = os.path.join(destdir, 'content_manifests')
+        self.icm = remote_source_content_manifest
+        self.dfp = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        labels = Labels(self.dfp.labels)
+        _, image_name = labels.get_name_and_value(Labels.LABEL_TYPE_COMPONENT)
+        _, image_version = labels.get_name_and_value(Labels.LABEL_TYPE_VERSION)
+        _, image_release = labels.get_name_and_value(Labels.LABEL_TYPE_RELEASE)
+        self.icm_file_name = '{}-{}-{}.json'.format(image_name, image_version, image_release)
```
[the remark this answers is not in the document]
athos-ribeiro: I don't follow the proposal. Could you elaborate? I don't think dockerfile_parser should know about koji/OSBS specific labels...

Note that this plugin will deprecate pre_add_content_sets, which will be removed after this plugin is set to run in osbs-client

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,159 @@
...
+    def _populate_content_sets(self):
+        """
+        Get the list of the current platform's content_sets from
+        'content_sets.yml' in dist-git, and set `self.content_sets` to same.
+        """
+        current_platform = self.workflow.user_params['platform']
+        workdir = self.workflow.builder.df_dir
+        fpath = os.path.join(workdir, REPO_CONTENT_SETS_CONFIG)
+        if os.path.exists(fpath):
+            content_sets_yml = read_yaml_from_file_path(fpath, 'schemas/content_sets.json') or {}
```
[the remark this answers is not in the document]
athos-ribeiro: pre_add_content_sets will be removed. This plugin is mostly extending the add_content_sets features (and re-naming the plugin).

The reason we are not also removing the other plugin altogether is that we will first make the change in osbs-client and make sure everything is alright before removing the deprecated plugin.

atomic_reactor/plugins/pre_add_image_content_manifest.py
athos-ribeiro: see deprecation comments above

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,159 @@
...
+        with open(self.out_file_path, 'w') as outfile:
+            json.dump(self.icm, outfile, indent=4)
+        self.log.debug('ICM JSON saved to: %s', self.out_file_path)
+
+    def _add_to_dockerfile(self):
+        """
+        Put an ADD instruction into the Dockerfile (to include the ICM file
+        into the container image to be built)
+        """
+        content = 'ADD {0} {1}'.format(self.icm_file_name, self.icm_dir)
```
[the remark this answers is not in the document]
athos-ribeiro: see deprecation comments above.

However, having common code reused with the add_helper plugin is a good idea

atomic_reactor/plugins/pre_add_image_content_manifest.py
tkdchen: @athos-ribeiro Thanks for reminding.

@ben-alkov Please ignore my comments that are not valid due to deprecate `add_content_sets`.

atomic_reactor/plugins/pre_add_image_content_manifest.py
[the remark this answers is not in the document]
tkdchen: @athos-ribeiro something like:

```python
def generate_xxx_filename(df_parser):
    labels = Labels(df_parser.labels)
    _, image_name = labels.get_name_and_value(Labels.LABEL_TYPE_COMPONENT)
    _, image_version = labels.get_name_and_value(Labels.LABEL_TYPE_VERSION)
    _, image_release = labels.get_name_and_value(Labels.LABEL_TYPE_RELEASE)
    return '{}-{}-{}.json'.format(image_name, image_version, image_release)
```

atomic_reactor/plugins/pre_add_image_content_manifest.py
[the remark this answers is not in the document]
tkdchen: @athos-ribeiro I get it. Thanks for explanation. The separate function could still be shared by the `pre_resolve_composes` and this new plugin. :)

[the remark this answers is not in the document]
ben-alkov: @tkdchen, @athos-ribeiro; cqi has a point:

1. Do we think it's worth the the effort?
1. If `1`, where should this code live, and how should it be accessed in other places?

ben-alkov:
@tkdchen; Will you open a/some stories to track any repeated-code tech debt you find? You points are *absolutely* valid, but I don't think that **this** PR (and associated tracking story) is necessarily the correct place to address them.

Of course, if you find instances where I could easily *call existing code*, please point them out :)

ben-alkov:
@tkdchen; Will you open a/some stories to track any repeated-code tech debt you find? You points are *absolutely* valid, but I don't think that this PR (and associated tracking story) is necessarily the correct place to address them.

Of course, if you find instances where I could easily *call existing code*, please point them out :)

tkdchen:
@ben-alkov I agree with you. I'll file new card for the reuse.

ben-alkov:
[Checks](https://github.com/ben-alkov/atomic-reactor/pull/3/checks) from a PR in my fork with content from both this PR and #1503

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        :param tasker: ContainerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param request_id: (int) ID of the remote source request. Used to get the ICM.
+        :param destdir: image path to carry content_manifests data dir
+        """
+        super(AddImageContentManifestPlugin, self).__init__(tasker, workflow)
+        self.content_manifests_dir = os.path.join(destdir, 'content_manifests')
+        self._cachito_session = None
+        self.icm = self.cachito_session.get_icm(request_id)
```
athos-ribeiro, line 85: As mentioned in https://github.com/containerbuildsystem/atomic-reactor/pull/1503, we can ass the content-manifest URL directly as we do with configuration files to avoid getting another cachito session here.

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,186 @@
...
+    @property
+    def icm(self):
+        """
+        Get and validate the ICM from the Cachito API `content-manifest` endpoint.
+
+        :return: dict, the ICM as a Python dict
+        """
+        if not self._icm:
+            def make_session():
+                # `method_whitelist=False` allows retrying non-idempotent methods like POST
```
athos-ribeiro, line 115: Do we need this, since we are only performing GET requests?

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,186 @@
...
+                session = get_retrying_requests_session(method_whitelist=False)
+                session.verify = True
+                return session
+            session = make_session()
```
athos-ribeiro, line 119: Why do we need this function here instead of getting the session directly? It seems it just got a bit more verbose.

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+        Get and validate the ICM from the Cachito API `content-manifest` endpoint.
+
+        :return: dict, the ICM as a Python dict
+        """
+        if not self._icm:
+            def make_session():
+                # `method_whitelist=False` allows retrying non-idempotent methods like POST
+                session = get_retrying_requests_session(method_whitelist=False)
+                session.verify = True
```
athos-ribeiro, line 117: We need to get this value from the Cachito "insecure" configuration. This will be important for development environments (such as osbs-box).

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,186 @@
...
+        return self._layer_index
+
+    @property
+    def icm(self):
+        """
+        Get and validate the ICM from the Cachito API `content-manifest` endpoint.
+
+        :return: dict, the ICM as a Python dict
+        """
+        if not self._icm:
```
athos-ribeiro, line 113: Should we use `is not None` here in case cachito returns an empty dict for some reason?

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def run(self):
+        """
+        run the plugin
+        """
+        self._populate_content_sets()
+        self._update_icm_data()
+        self._write_json_file()
+        self._add_to_dockerfile()
+        self.log.info("added %s", self.content_manifests_dir)
```
athos-ribeiro, line 186: Should this also include the file name added?

atomic_reactor/plugins/pre_add_image_content_manifest.py
ben-alkov: @athos-ribeiro; I don't know what "get this value from the Cachito "insecure" configuration" means... Set `session.verify = False`? Or is there more to it?

atomic_reactor/plugins/pre_add_image_content_manifest.py
ben-alkov: @athos-ribeiro; No, given that this is a property, we have to check for `None`. Maybe an additional check for an empty dict?

atomic_reactor/plugins/pre_add_image_content_manifest.py
ben-alkov: ack

atomic_reactor/plugins/pre_add_image_content_manifest.py
ben-alkov: ack

atomic_reactor/plugins/pre_add_image_content_manifest.py
chmeliik: @ben-alkov IIUC, what athos means is to use the Cachito configuration from reactor-config-map to determine whether session should be insecure, see `get_cachito_session()` in `plugins.pre_reactor_config`.

Speaking of `get_cachito_session()`, would it make sense to add a method to download the ICM to `CachitoAPI` from `utils.cachito`?

athos-ribeiro: @chmeliik, the ICM URL is present in the initial response request, meaning it could be hosted anywhere (OSBS should not bother with it. I have no preference on whether we should download it in the plugin or in the cachito utils itself.

atomic_reactor/plugins/pre_add_image_content_manifest.py
athos-ribeiro: I was actually proposing to check `if self._icm is None` instead of `if not self_icm`, in case the cachito response for this icm is just an empty dict, so we will not need to process it again. Sorry for the "not" in there.

atomic_reactor/plugins/pre_add_image_content_manifest.py
chmeliik: @athos-ribeiro I see, in that case adding it to cachito utils probably does not make that much sense.

atomic_reactor/plugins/pre_add_image_content_manifest.py
athos-ribeiro: NVM, I see you did change that! Thanks!

athos-ribeiro:
Great work

We just need to fetch the proper value for `verify` from the cachito configuration, if available!

Thank you, Ben!

ben-alkov:
osbs-box tests available from my [osbs-client PR](https://github.com/containerbuildsystem/osbs-client/pull/990#issuecomment-657789788)

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,200 @@
...
+
+        :return: dict, the ICM as a Python dict
+        """
+        if self.icm_url is None:
+            self._icm = self.minimal_icm
+        if self._icm is None:
+            session = get_retrying_requests_session()
+
+            # Gets the value of Cachito's 'insecure' key from the active reactor config map
+            session.verify = get_cachito(self.workflow).get('insecure', False)
```
athos-ribeiro, line 130: If insecure is `True`, should it mean we do **not** want to verify the certificates? In other words, should we toggle whatever the actual insecure value is?

Also, should we default to having a secure connection (i.e., verify if `insecure` is not available)?

Finally, what if, for some reason, cachito configuration is not available? I understand that the resolve_remote_source would skip execution and we would not enter this block. But maybe we should not rely on another plugin to avoid failure here and also set `verify` to `True` in this case, just to be safe.

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,200 @@
...
+
+    @property
+    def icm(self):
+        """
+        Get and validate the ICM from the Cachito API `content-manifest` endpoint.
+
+        :return: dict, the ICM as a Python dict
+        """
+        if self.icm_url is None:
+            self._icm = self.minimal_icm
```
athos-ribeiro, line 125: Good catch

atomic_reactor/plugins/pre_add_image_content_manifest.py
ben-alkov: - Absolutely correct; the logic here is wrong (reversed)
- Yes, agreed to default to secure (`insecure = False`)
- I'll make and use a `cachito_verify` instance attribute which defaults to True

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,208 @@
...
+            #     *always* have 2 layers
+            self._layer_index = 1
+        if not base_image_is_scratch(self.dfp.baseimage):
+            inspect = self.workflow.builder.base_image_inspect
+            self._layer_index = len(inspect[INSPECT_ROOTFS][INSPECT_ROOTFS_LAYERS])
+        return self._layer_index
+
+    @property
+    def cachito_verify(self):
+        if self._cachito_verify is True:
```
chmeliik, line 120: Why is the condition `is True`? If `cachito_verify` is supposed to be a lazily evaluated property, wouldn't it be better to set `self._cachito_verify = None` initially and check `if self._cachito_verify is None`?

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,208 @@
...
+            cachito_conf = get_cachito(self.workflow) or {}
```
chmeliik, line 121: If I'm not mistaken, `get_cachito()` raises `KeyError` when cachito is not configured. You may have to use a try/except, or use `get_cachito(self.workflow, fallback={})` (but get_cachito does not currently accept the `fallback` param, you would have to update it)

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def __init__(self, tasker, workflow, remote_source_icm_url=None, destdir=IMAGE_BUILD_INFO_DIR):
+        """
+        :param tasker: ContainerTasker instance
+        :param workflow: DockerBuildWorkflow instance
+        :param icm_url: str, URL of the ICM from the Cachito request.
+        :param destdir: image path to carry content_manifests data dir
+        """
+        super(AddImageContentManifestPlugin, self).__init__(tasker, workflow)
+        self.content_manifests_dir = os.path.join(destdir, 'content_manifests', '')
```
chmeliik, line 94: What is the purpose of that extra empty string at the end?

atomic_reactor/plugins/pre_add_image_content_manifest.py
athos-ribeiro: Good catch!

atomic_reactor/plugins/pre_add_image_content_manifest.py
athos-ribeiro: +1 to setting it to None and after the first evaluation, in case it cannot be determined (e.g., no cachito conf), set it to True.

MartinBasti: also it should be just `if self._cachito_verify:`. `is True` may be tricky sometimes

atomic_reactor/plugins/pre_add_image_content_manifest.py
chmeliik: Oh I think I got it:
```diff
- ADD eggs-1.0-42.json /root/buildinfo/content_manifests
+ ADD eggs-1.0-42.json /root/buildinfo/content_manifests/
```
Previously, the file would be added as `content_manifests` instead of being put into the `content_manifests/` dir

Another way to achieve that would be this:
```diff
- ADD eggs-1.0-42.json /root/buildinfo/content_manifests
+ ADD eggs-1.0-42.json /root/buildinfo/content_manifests/eggs-1.0.42.json
```
IMO that is a little less confusing and more explicit, what do you think?

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,208 @@
...
+
+        :return: dict, the ICM as a Python dict
+        """
+        if self.icm_url is None:
+            self._icm = self.minimal_icm
+        if self._icm is None:
+            session = get_retrying_requests_session()
+            session.verify = self.cachito_verify
+            self.log.debug('Making request to "%s"', self.icm_url)
+            response = session.get(self.icm_url)
```
MartinBasti, line 140: Don't we have `CachitoAPI` class that should implement all cachito operations?

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,208 @@
...
+        into the container image to be built)
+        """
+        content = 'ADD {0} {1}'.format(self.icm_file_name, self.content_manifests_dir)
+        lines = self.dfp.lines
+
+        # Put it before last instruction
+        lines.insert(-1, content + '\n')
+        self.dfp.lines = lines
+
+    def run(self):
```
MartinBasti, line 200: Where should this plugin run? Orchestrator/worker/both? (I assume worker, so we should make sure that this plugin is skipped on orchestrator)

atomic_reactor/plugins/pre_add_image_content_manifest.py
athos-ribeiro: We do, but the URL comes from the API response itself and then forwarded here so we do not need to add an additional dependency on the cachito util module (see resolve_remote_sources). We do the same for the configuration files in the download remote sources plugin.

atomic_reactor/plugins/pre_add_image_content_manifest.py
ben-alkov: @chmeliik; I'd rather keep it platform-neutral and add a comment, rather than hard-code a '/'

chmeliik: I'm not saying hard-code a '/', you could change [line 193](https://github.com/ben-alkov/atomic-reactor/blob/b23f39e1a9e180a0f0ba1f655377cd38a7a4f715/atomic_reactor/plugins/pre_add_image_content_manifest.py#L193) to this instead:
```python
dest_file_name = os.path.join(self.content_manifests_dir, self.icm_file_name)
content = 'ADD {0} {1}'.format(self.icm_file_name, dest_file_name)
```

Which would give you something like the second diff in my previous comment.
IMO using the full filename is more reliable than depending on a trailing '/'

atomic_reactor/plugins/pre_add_image_content_manifest.py
rcerven: plugin will be only in worker_inner:6.json and not in orchestrator_inner:6.json
so it will run only where it should

atomic_reactor/plugins/pre_add_image_content_manifest.py
chmeliik: The empty string in `os.path.join(destdir, 'content_manifests', '')` should not be necessary anymore (it will work anyway, but IMO is confusing)

Also, that `dest_file_name` should probably be `dest_file_path`, my bad

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,214 @@
...
+            inspect = self.workflow.builder.base_image_inspect
+            self._layer_index = len(inspect[INSPECT_ROOTFS][INSPECT_ROOTFS_LAYERS])
+        return self._layer_index
+
+    @property
+    def cachito_verify(self):
+        if self._cachito_verify is None:
+            try:
+                cachito_conf = get_cachito(self.workflow)
+            except KeyError:
```
athos-ribeiro, line 125: We are missing coverage for this block

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -0,0 +1,214 @@
...
+        self._layer_index = None
+        self._icm = None
+
+    @property
+    def layer_index(self):
+        if self._layer_index is None:
+            # Default layer index is 1, because base and 'FROM scratch' images
+            #     *always* have 2 layers
+            self._layer_index = 1
+        if not base_image_is_scratch(self.dfp.baseimage):
```
tkdchen, line 115: `base_image_is_scratch` compares `self.dfp.baseimage` with string `scratch`. What is the difference between this and the 'FROM scratch'?

athos-ribeiro:
Docs PR: https://github.com/containerbuildsystem/osbs-docs/pull/164

atomic_reactor/plugins/pre_add_image_content_manifest.py
ben-alkov: @tkdchen; As I understand it (I brought this line in from the previous 'add_content_sets' plugin), this is an abstraction over `if self.dfp.baseimage == scratch`

The advantages of the current construction are

- Clarity. It makes it clear that we're checking that the base image a is SCRATCH - regardless of what SCRATCH means in practical terms (implementation hiding)
- We don't care what SCRATCH means *in the code* (could become a boolean in the future...)

athos-ribeiro:
Release notes:

- new 'add_image_content_manifest' pre-build plugin to generate Image Content Manifests

PR #1522, opened 2020-07-16
Exclude closed source container first content from source containers

rcerven:
* CLOUDBLD-1784

### DIFF
--- a/atomic_reactor/plugins/pre_fetch_sources.py
+++ b/atomic_reactor/plugins/pre_fetch_sources.py
@@ -8,10 +8,13 @@ of the BSD license. See the LICENSE file for details.
 from __future__ import absolute_import
 
 import os
+import shutil
 import tempfile
 
 import koji
 from six import string_types
+import tarfile
+import yaml
 
 from atomic_reactor.constants import PLUGIN_FETCH_SOURCES_KEY
 from atomic_reactor.plugin import PreBuildPlugin
@@ -78,7 +81,7 @@ class FetchSourcesPlugin(PreBuildPlugin):
         koji_config = get_koji(self.workflow)
         insecure = koji_config.get('insecure_download', False)
         urls = self.get_srpm_urls(signing_intent['keys'], insecure=insecure)
-        urls_remote = self.get_remote_urls()
+        urls_remote, remote_sources_map = self.get_remote_urls()
 
         if not urls and not urls_remote:
             msg = "No srpms or remote sources found for source container," \
@@ -93,6 +96,7 @@ class FetchSourcesPlugin(PreBuildPlugin):
         if urls_remote:
             remote_sources_dir = self.download_sources(urls_remote, insecure=insecure,
                                                        download_dir=self.REMOTE_SOUCES_DOWNLOAD_DIR)
+            self.exclude_files_from_remote_sources(remote_sources_map, remote_sources_dir)
         return {
                 'sources_for_koji_build_id': self.koji_build_id,
                 'sources_for_nvr': self.koji_build_nvr,
@@ -195,14 +199,21 @@ class FetchSourcesPlugin(PreBuildPlugin):
         self.log.debug('archives: %s', archives)
         remote_sources_path = self.pathinfo.typedir(koji_build, btype='remote-sources')
         remote_sources_urls = []
+        dest_name = None
+        cachito_json_url = None
 
         for archive in archives:
             if archive['type_name'] == 'tar':
                 remote_source = {}
                 remote_source['url'] = os.path.join(remote_sources_path, archive['filename'])
                 remote_source['dest'] = '-'.join([koji_build['nvr'], archive['filename']])
+                dest_name = remote_source['dest']
                 remote_sources_urls.append(remote_source)
-        return remote_sources_urls
+
+            elif archive['type_name'] == 'json' and archive['filename'] == 'remote-source.json':
+                cachito_json_url = os.path.join(remote_sources_path, archive['filename'])
+
+        return remote_sources_urls, {dest_name: cachito_json_url}
 
     def get_remote_urls(self):
         """Fetch remote source urls from all builds
@@ -210,19 +221,24 @@ class FetchSourcesPlugin(PreBuildPlugin):
         :return: list, dicts with URL pointing to remote sources
         """
         remote_sources_urls = []
+        remote_sources_map = {}
 
-        remote_sources_urls.extend(self._get_remote_urls_helper(self.koji_build))
+        remote_source, remote_json = self._get_remote_urls_helper(self.koji_build)
+        remote_sources_urls.extend(remote_source)
+        remote_sources_map.update(remote_json)
 
         koji_build = self.koji_build
 
         while 'parent_build_id' in koji_build['extra']['image']:
             koji_build = self.session.getBuild(koji_build['extra']['image']['parent_build_id'],
                                                strict=True)
-            remote_sources_urls.extend(self._get_remote_urls_helper(koji_build))
+            remote_source, remote_json = self._get_remote_urls_helper(koji_build)
+            remote_sources_urls.extend(remote_source)
+            remote_sources_map.update(remote_json)
 
-        return remote_sources_urls
+        return remote_sources_urls, remote_sources_map
 
-    def get_denylisted_sources(self):
+    def get_denylisted_srpms(self):
         src_config = get_source_container(self.workflow, fallback={})
         denylist_srpms = src_config.get('denylist_srpms')
         if not denylist_srpms:
@@ -275,7 +291,7 @@ class FetchSourcesPlugin(PreBuildPlugin):
         rpms = [rpm for archive in archives
                 for rpm in self.session.listRPMs(imageID=archive['id'])]
 
-        denylist_srpms = self.get_denylisted_sources()
+        denylist_srpms = self.get_denylisted_srpms()
 
         srpm_build_paths = {}
         for rpm in rpms:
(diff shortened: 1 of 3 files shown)


atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -348,3 +365,126 @@ def get_signing_intent(self):
...
+                    break
+        return delete_app
+
+    def _delete_app_directory(self, remote_source_dir, unpack_dir, remote_archive):
+        vendor_dir = os.path.join(remote_source_dir, unpack_dir, 'app', 'vendor')
+
+        if os.path.exists(vendor_dir):
+            shutil.move(vendor_dir, remote_source_dir)
+            self.log.debug('Removing app from "%s"', remote_archive)
+            shutil.rmtree(os.path.join(unpack_dir, 'app'))
```
chmeliik, line 410: Shouldn't the path be `os.path.join(remote_source_dir, unpack_dir, 'app')`? Or am I missing something?

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -348,3 +365,126 @@ def get_signing_intent(self):
...
+            shutil.move(os.path.join(remote_source_dir, 'vendor'), vendor_dir)
```
chmeliik, line 411: Again, I might be missing something, but how can you move the vendor dir to `/somewhere/app/vendor` if `/somewhere/app/` was just deleted?

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -348,3 +365,126 @@ def get_signing_intent(self):
...
+    def _get_excluded_matches(self, unpack_dir, exclude_list):
+        matches = []
+        # py2 glob.glob doesn't support recursive, hence os.walk & fnmatch
+        # py3 can use: glob.glob(os.path.join(unpack_dir, '**', exclude), recursive=True)
+        for root, dirnames, filenames in os.walk(unpack_dir):
+            for entry in dirnames + filenames:
+                full_path = os.path.join(root, entry)
+
+                for exclude in exclude_list:
+                    if fnmatch.fnmatch(full_path, '*' + exclude):
```
chmeliik, line 426: Is `fnmatch` necessary here? Could it be just `full_path.endswith(exclude)`? Or is `exclude` also expected to include wildcards?

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -348,3 +365,126 @@ def get_signing_intent(self):
 
         signing_intent = odcs_config.get_signing_intent_by_name(self.signing_intent)
         return signing_intent
+
+    def _get_exclude_list(self, request_session, exclude_list_url):
+        response = request_session.get(exclude_list_url)
+        response.raise_for_status()
+        exclude_list_yaml = yaml.safe_load(response.text)
+        # prepend os.sep so it starts with exact match, glob.glob doesn't need it
+        return [os.path.join(os.sep, k, item) for k, v in exclude_list_yaml.items() for item in v]
```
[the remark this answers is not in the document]
rcerven: adding better explanation

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -348,3 +365,126 @@ def get_signing_intent(self):
...
+        for package in remote_json['packages']:
+            if delete_app:
+                break
+
+            for exclude_path in exclude_list:
+                if package.get('name').endswith(exclude_path):
+                    self.log.debug('Package excluded: "%s" from "%s"', package.get('name'),
+                                   remote_archive)
+                    delete_app = True
+                    break
```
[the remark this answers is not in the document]
rcerven: yop :+1:

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: remote_source_dir is the dir where those downloaded archives are and unpack_dir is 
archive name with pull path with _unpacked (just for unpacking)
I am moving vendor to main dir, so I can delete whole app dir in unpack_dir 
and then moving it back

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: vendor_dir   :    /somewhere/unpacked_dir/app/vendor
first we move 'vendor' to /somewhere/
then we delete : /somewhere/unpacked_dir/app/vendor
then we move vendor back   to /somewhere/unpacked_dir/app/vendor  from /somewhere/vendor

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: right we can use endswith

atomic_reactor/plugins/pre_fetch_sources.py
chmeliik: If I undestand it correctly, the entire process is similar to this (`tree` command just for visualization):
```shell
$ tree 'remote_sources'
remote_sources
└── foo_unpacked
    └── app
        └── vendor
$ mv 'remote_sources/foo_unpacked/app/vendor' 'remote_sources'
$ tree 'remote_sources'
remote_sources
├── foo_unpacked
│   └── app
└── vendor
$ rm -rf 'remote_sources/foo_unpacked/app'
$ tree 'remote_sources'
remote_sources
├── foo_unpacked
└── vendor
$ mv 'remote_sources/vendor' 'remote_sources/foo_unpacked/app/vendor'
mv: cannot move 'remote_sources/vendor' to 'remote_sources/foo_unpacked/app/vendor': No such file or directory
```
If that is correct, then how come the last `shutil.move` command does not fail? From the little test that I tried, `shutil.move` does not seem to create missing directories along the way, just like `mv` doesn't

atomic_reactor/plugins/pre_fetch_sources.py
chmeliik: Is `unpack_dir` a subdirectory of `remote_source_dir`? I'm confused because
- `vendor_dir` uses `os.path.join(remote_source_dir, unpack_dir, 'app', 'vendor')`
- here it's just `os.path.join(unpack_dir, 'app')`

If it is a subdirectory, it would be nice to keep it consistent

rcerven: remote_source_dir & unpack_dir   are full path dirs 

this is creating confusion :          vendor_dir = os.path.join(remote_source_dir, unpack_dir, 'app', 'vendor')

it works BUT it should be just:

vendor_dir = os.path.join(unpack_dir, 'app', 'vendor')

os.path.join just joins those dirs with full path in mind so it does it works

I'll change this

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: yes that is correct but shutil.move works:
here is example:

>>> shutil.move('/tmp/tmpo44pO_/remote_sources/unpacked/app/vendor', '/tmp/tmpo44pO_/remote_sources/')
>>> shutil.rmtree('/tmp/tmpo44pO_/remote_sources/unpacked/app')
>>> shutil.move('/tmp/tmpo44pO_/remote_sources/vendor', '/tmp/tmpo44pO_/remote_sources/unpacked/app/vendor')

rcerven: shutil.move(src, dst)
Recursively move a file or directory (src) to another location (dst).

chmeliik: Oh ok, so there seems to be a difference between how `shutil.move` handles directories vs. simple files:
```python
>>> shutil.move('/tmp/foo.txt', '/tmp/nonexistent/foo.txt')
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/nonexistent/foo.txt'
>>> shutil.move('/tmp/bar/', '/tmp/nonexistent/bar')
'/tmp/nonexistent/bar'
```
Could you add a comment that `shutil.move` will create missing parent directories in `dst` when `src` is a directory?

chmeliik:
Unit tests in CI are failing on python 3, looks like a `bytes` vs. `str` issue

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: sure

rcerven:
> Unit tests in CI are failing on python 3, looks like a `bytes` vs. `str` issue

fixed

atomic_reactor/schemas/config.json
```diff
@@ -665,6 +665,10 @@
                   }
               },
               "required": ["denylist_url", "denylist_key"]
+          },
+          "excludelist_content": {
```
chmeliik, line 669: Is there a specific reason why this is called "excludelist" rather than "denylist"? This serves practically the same purpose as `denylist_srpms`, maybe a better name would be something like `denylist_container_first` or just `denylist_sources`?

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -348,3 +364,125 @@ def get_signing_intent(self):
...
+
+            # if any package in cachito json matched excluded entry,
+            # remove 'app' from sources, except 'app/vendor' when exists
+            if delete_app and os.path.exists(os.path.join(unpack_dir, 'app')):
+                self._delete_app_directory(remote_sources_dir, unpack_dir, remote_archive)
+
+            # search for excluded matches
+            matches = self._get_excluded_matches(unpack_dir, exclude_list)
+
+            self._remove_excluded_matches(matches)
```
tkdchen, line 477: My understand of the lines from 464 to 477 is they are filtering out files from extracted cachito archive file by finding out excluded items. May I suggest another way to implement, that is to use use `members` argument of `TarFile.extractall`, and `TarFile.getmembers` is helpful to get the full list of names within the archive file.

In this way, we don't have to extract the whole archive file and do all the necessary IO operations through the whole unpack directory tree. IIRC, I was told that there might be much big archive files in cachito. However, I'm not sure if that would affect source container build.

Basically, the workflow could look like:

```
def filter_func(members):
     yield the included member only

tarfile.extractall(unpack_dir, members=filter_func(tarfile))
remove remote_archive
create new remote_archive from unpack_dir
```

atomic_reactor/schemas/config.json
MartinBasti: also for RPMS we had to switch to fetching config via HTTP because that list is maintained by somebody else and needs more often updates which may lead to updated OSBS config very often

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -195,32 +199,44 @@ def _get_remote_urls_helper(self, koji_build):
...
             if archive['type_name'] == 'tar':
                 remote_source = {}
                 remote_source['url'] = os.path.join(remote_sources_path, archive['filename'])
                 remote_source['dest'] = '-'.join([koji_build['nvr'], archive['filename']])
+                dest_name = remote_source['dest']
                 remote_sources_urls.append(remote_source)
-        return remote_sources_urls
+
+            elif archive['type_name'] == 'json' and archive['filename'] == 'remote-source.json':
+                cachito_json_url = os.path.join(remote_sources_path, archive['filename'])
```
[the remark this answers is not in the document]
rcerven: well I want to be sure it is really there, hence in loop

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -195,32 +199,44 @@ def _get_remote_urls_helper(self, koji_build):
...
+
+        return remote_sources_urls, {dest_name: cachito_json_url}
```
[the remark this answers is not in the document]
rcerven: it is getting archives fort btype : remote-sources

and there is just 1 archive with sources

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -348,3 +364,125 @@ def get_signing_intent(self):
...
+
+    def _get_exclude_list(self, request_session, exclude_list_url):
+        response = request_session.get(exclude_list_url)
+        response.raise_for_status()
+        exclude_list_yaml = yaml.safe_load(response.text)
+        # prepend os.sep for 2 reasons:
+        # - so fnmatch will match exact dir/file when using * + exclude
+        #   glob.glob doesn't need it
+        # - so endswith for package will match also full name
+        return [os.path.join(os.sep, k, item) for k, v in exclude_list_yaml.items() for item in v]
```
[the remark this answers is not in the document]
rcerven: well in that exclude list is basically subpath1 and subpath2  .... it isn't necessary subdir/filename so I just using k/v

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -348,3 +364,125 @@ def get_signing_intent(self):
...
+        for remote_source, remote_source_json in remote_sources_map.items():
+            remote_source_archive = os.path.join(remote_sources_dir, remote_source)
+
+            response = request_session.get(remote_source_json)
+            response.raise_for_status()
+            response_json = response.json()
+            full_remote_sources_map[remote_source_archive] = response_json
+        return full_remote_sources_map
+
+    def _check_if_package_excluded(self, remote_json, exclude_list, remote_archive):
```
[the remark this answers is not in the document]
rcerven: sure

atomic_reactor/plugins/pre_fetch_sources.py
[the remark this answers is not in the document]
chmeliik: I see, but there does not seem to be any special behaviour for the case when it is missing. The URL will end up being `None` and it will eventually fail with `request_session.get(None)` when creating the full mapping? Is that OK?

atomic_reactor/schemas/config.json
rcerven: changed to : denylist_sources

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: well I need to first unpack archive,
delete matched dirs/files from it 
and then pack it again 

so I don't see how useful is it 

also keep in mind, that we will be deleting just few things from the archive

atomic_reactor/plugins/pre_fetch_sources.py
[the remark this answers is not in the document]
rcerven: well remote-source.json will always be in source container build as well as just 1 archive

so I don't think we need extra check for that

in the worst case it will fail on that get(None) (IF source container build is missing remote-source.json)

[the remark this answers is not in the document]
chmeliik: ok

atomic_reactor/schemas/config.json
chmeliik: I noticed that the plugin already has a `get_denylisted_sources` helper method, but it gets denylisted SRPMs, which will now become confusing. Could it be renamed to `get_denylisted_srpms`? Could the methods/variables that have "excludelist" in the name be renamed as well?

rcerven: ok renamed what made sense

tests/plugins/test_fetch_sources.py
```diff
@@ -217,19 +224,61 @@ def body_callback(request, context):
             requests_mock.register_uri('GET', url, body=body_callback)
 
     def body_remote_callback(request, context):
-        f = MockBytesIO(b"remote sources content")
+        f = MockBytesIO(targz_bytes)
         return f
+
+    if 'app' not in dirs_in_remote:
+        os.mkdir(os.path.join(str(tmpdir), 'app'))
+    if 'app' not in dirs_in_remote:
```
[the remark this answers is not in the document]
rcerven: because typo 2nd one should be 'deps' :D 

fixed

rcerven:
rn updated:  Exclude closed source container first content from source containers

rcerven:
rn updated:  Exclude closed source container first content from source containers

PR #1523, opened 2020-07-17
Use a single HTTP session in download_remote_source.run

tkdchen:

### DIFF
--- a/atomic_reactor/plugins/pre_download_remote_source.py
+++ b/atomic_reactor/plugins/pre_download_remote_source.py
@@ -44,26 +44,19 @@ class DownloadRemoteSourcePlugin(PreBuildPlugin):
         self.buildargs = remote_source_build_args or {}
         self.remote_source_conf_url = remote_source_configs
 
-    def get_remote_source_config(self, url, insecure=False):
+    def get_remote_source_config(self, session, url, insecure=False):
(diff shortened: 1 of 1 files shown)


atomic_reactor/plugins/pre_download_remote_source.py
```diff
@@ -44,27 +44,6 @@ def __init__(self, tasker, workflow, remote_source_url=None,
...
         self.remote_source_conf_url = remote_source_configs
 
-    def get_remote_source_config(self, url, insecure=False):
-        """Get the configuration files associated with the remote sources
-
-        :param url: str, URL to cachito remote source configurations
-        :param insecure: bool, whether to verify SSL certificates
-
-        :return: list<dict>, configuration data for the given request.
-                 entries include path, type, and content
```
chmeliik, line 54: Could we keep this part of the docstring somewhere? It would be nice to have the format of the response documented.

MartinBasti: Can we keep this function and just use session param to have better abstraction?

tkdchen: Fixed.

MartinBasti:
Internal change only, no release notes

PR #1529, opened 2020-07-22
Timeout when waiting for Cachito requests based on the last update time

mprahl:
This will reduce the amount of times that OSBS will fail the build
when Cachito has a long queue of tasks.

Additionally, since the burst length resets after every state
change, the default `burst_retry` value is set from `1` second
to `3` seconds to reduce the load on Cachito's REST API and database.

Resolves CLOUDBLD-2007.

Signed-off-by: mprahl <mprahl@users.noreply.github.com>

### DIFF
--- a/atomic_reactor/utils/cachito.py
+++ b/atomic_reactor/utils/cachito.py
@@ -101,7 +101,7 @@ class CachitoAPI(object):
         return response_json
 
     def wait_for_request(
-            self, request, burst_retry=1, burst_length=30, slow_retry=10):
+            self, request, burst_retry=3, burst_length=30, slow_retry=10):
         """Wait for a Cachito request to complete
 
         :param request: int or dict, either the Cachito request ID or a dict with 'id' key
@@ -119,7 +119,8 @@ class CachitoAPI(object):
         url = '{}/api/v1/requests/{}'.format(self.api_url, request_id)
         logger.info('Waiting for request %s to complete...', request_id)
 
-        start_time = time.time()
+        last_updated_value = None
+        last_update_time = None
         while True:
             response = self.session.get(url)
             response.raise_for_status()
@@ -147,14 +148,22 @@ class CachitoAPI(object):
 
             # All other states are expected to be transient and are not checked.
 
-            elapsed = time.time() - start_time
+            # If "last_updated_value" does not match the "updated" value of the
+            # request from Cachito, then we know Cachito has performed some work
+            # since the last check, so the timer resets.
+            if last_updated_value is None or last_updated_value != response_json['updated']:
+                last_updated_value = response_json['updated']
+                last_update_time = time.time()
+
+            elapsed = time.time() - last_update_time
             if elapsed > self.timeout:
                 logger.error(dedent("""\
-                    Request %s not completed after %s seconds
+                    Request %s not completed after %s seconds of not being updated
                     Details: %s
                     """), url, self.timeout, json.dumps(response_json, indent=4))
                 raise CachitoAPIRequestTimeout(
-                    'Request %s not completed after %s seconds' % (url, self.timeout))
+                    'Request %s not completed after %s seconds of not being updated'
+                    % (url, self.timeout))
             else:
                 if elapsed > burst_length:
                     time.sleep(slow_retry)
--- a/tests/utils/test_cachito.py
+++ b/tests/utils/test_cachito.py
@@ -19,6 +19,7 @@ import json
 import os.path
 import re
 import time
+from datetime import datetime
 from textwrap import dedent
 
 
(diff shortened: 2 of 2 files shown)


atomic_reactor/utils/cachito.py
```diff
@@ -101,7 +102,7 @@ def request_sources(self, repo, ref, flags=None, pkg_managers=None, user=None,
         return response_json
 
     def wait_for_request(
-            self, request, burst_retry=1, burst_length=30, slow_retry=10):
+            self, request, burst_retry=5, burst_length=30, slow_retry=10):
```
athos-ribeiro, line 105: Why is this also being changed? Should this change be in a different PR or also be explained in the commit message?

atomic_reactor/utils/cachito.py
```diff
@@ -147,14 +147,16 @@ def wait_for_request(
 
             # All other states are expected to be transient and are not checked.
 
-            elapsed = time.time() - start_time
+            updated_time = datetime.fromisoformat(response_json["updated"])
+            elapsed = (datetime.utcnow() - updated_time).total_seconds()
```
athos-ribeiro, line 151: Should we trust the time on an external service here? We could make atomic-reactor start counting from the last time it has perceived  a new update instead.

atomic_reactor/utils/cachito.py
```diff
@@ -147,14 +147,16 @@ def wait_for_request(
 
             # All other states are expected to be transient and are not checked.
 
-            elapsed = time.time() - start_time
+            updated_time = datetime.fromisoformat(response_json["updated"])
```
athos-ribeiro, line 150: While `fromisoformat` was introduced in python 3.7, atomic-reactor still supports python 2.7 (and it would also be nice to keep 3.6 compatibility since we plan on supporting CentOS 8 for buildroots soon).

atomic_reactor/utils/cachito.py
mprahl: I added an explanation in the commit message and PR description. I also changed this to `3` seconds.

atomic_reactor/utils/cachito.py
mprahl: Thanks for catching that. I changed it to use `strptime` instead.

atomic_reactor/utils/cachito.py
mprahl: That is interesting to consider, however, since Cachito uses Kerberos for its workers to authenticate with the REST API, there would be a Cachito outage if there was more than a 5 minute time gap.

I'll add an extra commit which implements your suggested approach. Then you and other reviewers can let me know which approach you prefer.

mprahl: @athos-ribeiro I ended up just amending the original commit since this seemed pretty clean.

atomic_reactor/utils/cachito.py
```diff
@@ -147,14 +149,22 @@ def wait_for_request(
 
             # All other states are expected to be transient and are not checked.
 
-            elapsed = time.time() - start_time
+            # If "last_updated_value" does not match the "updated" value of the
+            # request from Cachito, then we know Cachito has performed some work
+            # since the last check, so the timer resets.
+            if last_updated_value is None or last_updated_value != response_json['updated']:
```
mprahl, line 155: I added `last_updated_value is None` to the if statement in the event Cachito has a bug and returns a null `updated` value. Let me know if you think this is too defensive and I can remove it.

lcarva: You could simply do:
```
last_update_time = response_json['updated'] or last_update_time or time.time()
```

However, in either case, if there's a Cachito bug this could run indefinitely.

atomic_reactor/utils/cachito.py
```diff
@@ -147,14 +148,22 @@ def wait_for_request(
...
+                last_updated_value = response_json['updated']
+                last_update_time = time.time()
+
+            elapsed = time.time() - last_update_time
```
lcarva, line 158: This will cause a failure if Cachito doesn't return an `updated` value initially.

mprahl: It won't because `last_update_time` always gets set to `time.time()` on the first run because of `if last_updated_value`.

atomic_reactor/utils/cachito.py
mprahl: `last_update_time` is float of seconds but `response_json['updated']` is a string, so this won't work.

athos-ribeiro:
Release notes added:

 - Cachito requests will now timeout based on the last update time

I will wait for @lcarva to ack his comments before merging (feel free to merge, Lui, RN already added).

athos-ribeiro:
@lcarva, we decided to merge this so we can test it in time for the next release. If you think this should not have been merged, let me know and I will work on the missing bits :)

PR #1531, opened 2020-07-23
Update CI to Fedora 32 and drop F30

ben-alkov:
Keep F29 for Py2 Pylint for now.

* CLOUDBLD-1521

### DIFF
--- a/.github/workflows/pylint.yaml
+++ b/.github/workflows/pylint.yaml
@@ -23,7 +23,7 @@ jobs:
             engine: docker
 
           - name: fedora
-            version: 31
+            version: 32
             python: 3
             engine: docker
 
--- a/.github/workflows/bandit.yaml
+++ b/.github/workflows/bandit.yaml
@@ -23,7 +23,7 @@ jobs:
             engine: docker
 
           - name: fedora
-            version: 31
+            version: 32
             python: 3
             engine: docker
 
--- a/.github/workflows/test_sh.yaml
+++ b/.github/workflows/test_sh.yaml
@@ -24,12 +24,12 @@ jobs:
             engine: docker
 
           - name: fedora
-            version: 30
+            version: 31
             python: 3
             engine: docker
 
           - name: fedora
-            version: 31
+            version: 32
             python: 3
             engine: docker
 


.github/workflows/pylint.yaml
```diff
@@ -18,12 +18,12 @@ jobs:
       matrix:
         os:
           - name: fedora
-            version: 29
-            python: 2
+            version: 31
+            python: 3
```
athos-ribeiro, line 22: We still need to run pylint in python 2. IMO it does not make much sense to run it in both supported version of fedora for python 3.

It seems to be available in epel7 though, through this package: https://koji.fedoraproject.org/koji/buildinfo?buildID=1403259

We could run it in CentOS7 with EPEL for python 2. Thoughts?

ben-alkov: Cent makes more sense than Fedora, for sure.

ben-alkov: @athos-ribeiro; Alrighty then. I'm assuming that we were stuck on F29 x Py2 because "No package python-pylint available" on Cent7 x Py2.

athos-ribeiro: We need to install it from EPEL, which is not automatically enabled in CentOS (you have to install the epel package). This was the reason we did not go with it a while ago. But now we do not have python 2 in Fedora anymore, it makes sense to go this way.

ben-alkov: @athos-ribeiro; Latest checks show that I've won the battle but lost the war. EPEL's pylint (v1.6.5 - the only one available; indeed it's the same for Py3) is either too stupid to recognize our pylint config, or too stupid to lint *only* the delta in this PR. Either way, it returns about 10 million hits (i.e. instead of 0)  ¯\\\_(ツ)_/¯

ben-alkov: As we're dropping Py2 support as soon as humanly possible, we've decided to stick with F29 for Py2 pylint but note that it's essentially deprecated, and open an issue to drop Py2 pylint when possible.

test.sh
```diff
@@ -30,109 +30,119 @@ function setup_osbs() {
...
-    PIP_PKG="python-pip"
+
+  if [[ $OS == "centos" ]]; then
+    PYTHON="python"
+    PIP_PKG="$PYTHON-pip"
     PIP="pip"
     PKG="yum"
     ENABLE_REPO=
-    PKG_EXTRA="yum-utils epel-release git-core desktop-file-utils flatpak ostree skopeo python2-libmodulemd2"
+    PKG_EXTRA="yum-utils git-core desktop-file-utils flatpak ostree skopeo ${PYTHON}2-libmodulemd2"
```
chmeliik, line 43: IMO `${PYTHON}2` does not make much sense, ideally the `$PYTHON` variable should already include the version number (but that might break something in the centos workflow, in which case leaving it harcoded as `python2-libmodulemd2` would be better IMO)

test.sh
```diff
@@ -30,109 +30,119 @@ function setup_osbs() {
...
+    PKG_EXTRA="dnf-plugins-core desktop-file-utils flatpak ostree skopeo $PYTHON-libmodulemd glibc-langpack-en $PYTHON-pylint"
+    BUILDDEP="dnf builddep"
   fi
 
-  PIP_INST=("$PIP" install --index-url "${PYPI_INDEX:-https://pypi.org/simple}")
+  # Common install dependencies
+  PKG_COMMON_EXTRA="git gcc krb5-devel python-devel popt-devel"
+  PKG_EXTRA="$PKG_EXTRA $PKG_COMMON_EXTRA"
+
+  PIP_INST=("$PIP" install --index-url https://pypi.org/simple)
```
chmeliik, line 59: I think we want to keep `"${PYPI_INDEX:-https://pypi.org/simple}"`, @tkdchen added this with the reasoning that specifying a PyPI mirror might make local tests faster for some people

test.sh
```diff
@@ -30,109 +30,119 @@ function setup_osbs() {
   OSBS_CLIENT_REPO=${OSBS_CLIENT_REPO:-https://github.com/containerbuildsystem/osbs-client}
   OSBS_CLIENT_BRANCH=${OSBS_CLIENT_BRANCH:-master}
 
-  # Pull fedora images from registry.fedoraproject.org
-  if [[ $OS == "fedora" ]]; then
-    IMAGE="registry.fedoraproject.org/$IMAGE"
-  fi
```
chmeliik, line 36: Can you clarify why pulling from the "official" repos is not desirable?

MartinBasti:
Why refactoring was needed? Ticket was about replacing F30 with F32, nothing about centos  We will soon drop py2 support and change that test.sh again and eventually migrate to centos8.

Each refactoring change should have separate commit with explanation why.

test.sh
ben-alkov: It's not *possible*; GH just fails.

Regardless even of that, that code has been skipped for months now - it's in the wrong place (which is why GH not pulling from registries other than docker.io hasn't been a problem up to this point).

test.sh
ben-alkov: ok

test.sh
ben-alkov: ok

ben-alkov:
> Why refactoring was needed? Ticket was about replacing F30 with F32, nothing about centos

@MartinBasti; It was actually necessary to get pylint working under Cent, but we've dropped that idea as it was working, but not *properly*. I'll open a separate PR for test.sh.

MartinBasti:
> > Why refactoring was needed? Ticket was about replacing F30 with F32, nothing about centos
> 
> @MartinBasti; It was actually necessary to get pylint working under Cent, but we've dropped that idea as it was working, but not _properly_. I'll open a separate PR for test.sh.

Why are you moving pylint under centos? pylint is so old there. We will drop it soon anyway py3-only rulez

test.sh
chmeliik: I see, makes sense (weird that it would fail though, huh)

ben-alkov:
> Why are you moving pylint under centos? pylint is so old there. We will drop it soon anyway py3-only rulez

You're preaching to the choir, but you were OOO during these discussions.

The move to Cent was to
- Keep Py 2 pylinting in our CI
- Move off of very old Fedora 29

which didn't work, so we're just keeping Py2 pylint running on F29 *for now*

PR #1532, opened 2020-07-23
test_flatpak_create_oci.py: Use flatpak_metadata=both for most tests

owtaylor:
Importing a Flatpak created with flatpak_metadata=annotations is only supported
for Flatpak < 1.6, and importing a Flatpak created with flatpak_metdata=labels
is only supported for for Flatpak >= 1.6, so run most tests with
flatpak_metadata=both, and only run basic tests with just labels/annotations.

* CLOUDBLD-1519


athos-ribeiro:
No release notes added: test suite change.

PR #1533, opened 2020-07-23
Reserve builds for isolated builds

rcerven:
* CLOUDBLD-1732

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -147,7 +147,7 @@ class BumpReleasePlugin(PreBuildPlugin):
         return next_release
 
     def reserve_build_in_koji(self, component, version, release, release_label,
-                              dockerfile_labels, source_build=False):
+                              dockerfile_labels, source_build=False, user_provided_release=False):
         """
         reserve build in koji, and set reserved build id an token in workflow
         for koji_import
@@ -157,7 +157,7 @@ class BumpReleasePlugin(PreBuildPlugin):
                 'name': component,
                 'version': version,
             }
-            if source_build:
+            if source_build or user_provided_release:
                 nvr_data['release'] = release
             else:
                 nvr_data['release'] = dockerfile_labels[release_label]
@@ -167,9 +167,14 @@ class BumpReleasePlugin(PreBuildPlugin):
                 reserve = self.xmlrpc.CGInitBuild(PROG, nvr_data)
                 break
             except GenericError as exc:
+                if release and user_provided_release:
+                    self.log.error("CGInitBuild failed, not retrying because"
+                                   " release was explicitly provided by user")
+                    raise RuntimeError(exc)
+
                 if release and not source_build:
                     self.log.error("CGInitBuild failed, not retrying because"
-                                   " release was explicitly specified in Dockerfile ")
+                                   " release was explicitly specified in Dockerfile")
                     raise RuntimeError(exc)
 
                 if counter < KOJI_RESERVE_MAX_RETRIES:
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -147,7 +147,7 @@ def get_next_release(self, build_info):
         return next_release
 
     def reserve_build_in_koji(self, component, version, release, release_label,
-                              dockerfile_labels, source_build=False):
```
MartinBasti, line 150: can we just use `release` param, a.k.a if define always use it?

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -291,6 +292,21 @@ def run(self):
         # label is present
         release_label = labels.LABEL_NAMES[Labels.LABEL_TYPE_RELEASE][0]
 
+        # Reserve build for isolated builds as well (or any build with supplied release)
+        user_provided_release = self.workflow.user_params.get('release')
```
MartinBasti, line 296: is this block needed? IMHO this should be incorporated into block at the end (or bellow), it already checks release existence, do magic when scratch build, etc..

Does this also miss update of dockerfile (it's in the last code block)?
```
dockerfile_labels[release_label] = release
```

chmeliik:
> Cyclomatic complexity of this plugin is overwhelming me. It's very error prone, I'd like to avoid adding more ifs. IMHO this plugin may be good candidate for refactoring.

Agreed, IMO one of the main issues is that half of the `run()` method is just getting NVR from labels, that should be a separate method for sure. If we don't want to do it in this PR, it would be worth looking at in the future.

atomic_reactor/plugins/pre_bump_release.py
rcerven: in next_release_general is dockerfile updted with new release, so I need it

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -169,7 +169,12 @@ def reserve_build_in_koji(self, component, version, release, release_label,
...
                 if release and not source_build:
                     self.log.error("CGInitBuild failed, not retrying because"
-                                   " release was explicitly specified in Dockerfile ")
+                                   " release was explicitly specified in Dockerfile")
+                    raise RuntimeError(exc)
+
+                if release and user_provided_release:
+                    self.log.error("CGInitBuild failed, not retrying because"
+                                   " release was explicitly provided by user")
                     raise RuntimeError(exc)
```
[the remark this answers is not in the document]
rcerven: right, switching em

atomic_reactor/plugins/pre_bump_release.py
rcerven: I would rather keep it like this, because WHEN release is user_provided all we want to do is just check if exists, and reserve build, and not do anything else  : append, update dockerfile (that is done in add_labels plugin)

atomic_reactor/plugins/pre_bump_release.py
MartinBasti: ok

atomic_reactor/plugins/pre_bump_release.py
MartinBasti: I see duplicated code, it's error prone because changes to isolated builds may be ommited again

rcerven: it is more clear to have user_provided block separated, as we want to just check build and reserve it

instead of making block bellow more complex by adding IF user_provided to share 2 lines

rcerven:
rn updated: Reserve builds for isolated builds

PR #1534, opened 2020-07-24
Decouple DockerBuildWorkflow with OUTPUT_IMAGE for source container b…

tkdchen:
…uild

* CLOUDBLD-123

OSv3InputPlugin reads environment variable OUTPUT_IMAGE set by OpenShift
to get the image tag, and then the value will be set to
DockerBuildWorkflow image argument. However, the plugins used for
building source container image do not use the value of image from
either property workflow.image or workflow.builder.image.

To make the interface of DockerBuildWorkflow clear for both binary and
source container image build, this patch does following things to
decouple it with OUTPUT_IMAGE:

* removes image argument from DockerBuildWorkflow.
* DockerBuildWorkflow object still has a property named image, because
  plugins for binary image build also have the code 'workflow.image'.
* DockerBuildWorkflow uses self.user_params['image_tag'] internally to
  initialize InsideBuilder object by passing the value to its argument
  image.
* InsideBuilder still requires image as it is used in a few places, for
  example in BuildStepPluginRunner.run and
  BuilderStateMachine.ensure_is_built.


tkdchen:
Hi, please review if I'm on the correct direction. Thanks!

MartinBasti:
+1 I think that direction is good

tkdchen:
@MartinBasti @rcerven Thanks for your review. Anything else I missed to change for the refactor?

atomic_reactor/inner.py
```diff
@@ -483,7 +485,7 @@ def build_docker_image(self):
         :return: BuildResult
         """
         exception_being_handled = False
-        self.builder = InsideBuilder(self.source, self.image)
+        self.builder = InsideBuilder(self.source, self.user_params['image_tag'])
```
rcerven, line 488: why not use already that image property? 
meaning you can leave there : self.image

atomic_reactor/plugins/input_osv3.py
```diff
@@ -136,7 +136,6 @@ def run(self):
         build_json = get_build_json()
         git_url = os.environ['SOURCE_URI']
         git_ref = os.environ.get('SOURCE_REF', None)
-        image = os.environ['OUTPUT_IMAGE']
```
rcerven, line 139: there is also the same in : OSv3SourceContainerInputPlugin  line 216

atomic_reactor/plugins/input_osv3.py
```diff
@@ -169,7 +168,6 @@ def run(self):
                     'git_branch': git_branch,
                 },
             },
-            'image': image,
```
rcerven, line 172: there is also the same in : OSv3SourceContainerInputPlugin  line 235

rcerven:
we should also remove 'OUTPUT_IMAGE' from both classmethods 'is_autousable' in input_osv3.py

tkdchen:
@rcerven Updated. PTAL.

tkdchen:
Can this be merged before the markdownlint issue is fixed?

rcerven:
> Can this be merged before the markdownlint issue is fixed?

yeah

MartinBasti:
> Can this be merged before the markdownlint issue is fixed?

Yes. Lets merge it and deal with markdown later.

PR #1536, opened 2020-07-28
Refactor test.sh

ben-alkov:
Best reviewed one commit at a time. See messages therein.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -30,11 +30,6 @@ function setup_osbs() {
   OSBS_CLIENT_REPO=${OSBS_CLIENT_REPO:-https://github.com/containerbuildsystem/osbs-client}
   OSBS_CLIENT_BRANCH=${OSBS_CLIENT_BRANCH:-master}
 
-  # Pull fedora images from registry.fedoraproject.org
-  if [[ $OS == "fedora" ]]; then
-    IMAGE="registry.fedoraproject.org/$IMAGE"
-  fi
-
   # PIP_PREFIX: osbs-client provides input templates that must be copied into /usr/share/...
   ENVS='-e PIP_PREFIX=/usr'
   RUN="$ENGINE exec -i ${ENVS} $CONTAINER_NAME"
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -30,64 +30,65 @@ function setup_osbs() {
...
+    PKG_EXTRA="yum-utils git-core desktop-file-utils flatpak ostree skopeo python2-libmodulemd2"
     BUILDDEP="yum-builddep"
-    PYTHON="python"
+  else
+    PYTHON="python$PYTHON_VERSION"
+    PIP_PKG="$PYTHON-pip"
+    PIP="pip$PYTHON_VERSION"
+    PKG="dnf"
+    ENABLE_REPO="--enablerepo=updates-testing"
+    PKG_EXTRA="dnf-plugins-core desktop-file-utils flatpak ostree skopeo python$PYTHON_VERSION-libmodulemd glibc-langpack-en $PYTHON-pylint"
```
tkdchen, line 50: This works, but array is recommended.

ben-alkov: @tkdchen; I don't think I'm following... Are you saying that `PKG_EXTRA` should be an array?

tkdchen: @ben-alkov 

Yeah, this is what I learned before from shellcheck. By using array like `PKG_EXTRA=(dnf-plugins-core desktop-file-utils flatpak)`, it is helpful to avoid issue SC2086, iirc. In addition, it is also easier to format the code into multiple lines when more elements to be added. Anyway, current code is ok for me as well.

ben-alkov: @tkdchen; Nice! Thx.

test.sh
```diff
@@ -30,64 +30,63 @@ function setup_osbs() {
   OSBS_CLIENT_REPO=${OSBS_CLIENT_REPO:-https://github.com/containerbuildsystem/osbs-client}
   OSBS_CLIENT_BRANCH=${OSBS_CLIENT_BRANCH:-master}
 
-  # Pull fedora images from registry.fedoraproject.org
-  if [[ $OS == "fedora" ]]; then
-    IMAGE="registry.fedoraproject.org/$IMAGE"
-  fi
```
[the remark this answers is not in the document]
ben-alkov: Yeah, sorry there was previous discussion about this:

@chmeliik
Can you clarify why pulling from the "official" repos is not desirable?

@ben-alkov
It's not *possible*; GH just fails.

Regardless even of that, that code has been skipped for months now - it's in the wrong place (which is why GH not pulling from registries other than docker.io hasn't been a problem up to this point).

@chmeliik
I see, makes sense (weird that it would fail though, huh)

PR #1538, opened 2020-08-03
Defaults to Python 3 in Fedora test container

tkdchen:

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -5,10 +5,14 @@ set -eux
 ENGINE=${ENGINE:="podman"}
 OS=${OS:="centos"}
 OS_VERSION=${OS_VERSION:="7"}
-PYTHON_VERSION=${PYTHON_VERSION:="2"}
-ACTION=${ACTION:="test"}
 IMAGE="$OS:$OS_VERSION"
+if [[ "$OS" == "fedora" ]]; then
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -5,10 +5,14 @@ set -eux
 ENGINE=${ENGINE:="podman"}
 OS=${OS:="centos"}
 OS_VERSION=${OS_VERSION:="7"}
-PYTHON_VERSION=${PYTHON_VERSION:="2"}
-ACTION=${ACTION:="test"}
 IMAGE="$OS:$OS_VERSION"
+if [[ "$OS" == "fedora" ]]; then
```
MartinBasti, line 9: for py2 pylint we need F29 with py2

tkdchen: So, it makes no sense to make this change. Thanks for the info. :)

ben-alkov: Also, if you're doing anything like this in future, please switch on Cent instead *even if there's no action for Cent*, e.g.

```bash
if [[ "$OS" == "centos" ]]; then
    :
else  # or even `elif [[ "$OS" == "fedora" ]]` to make it explicit
    FOO
fi
```

IMO it's easier to reason about the thing as a whole if we always switch on *one* OS (or version of Python, for that matter).

PR #1542, opened 2020-08-13
make packages required parameter

arewm:
Cachito should always provide it, so making it
required to make sure that later tools built on top
of it are not accidentally broken if it is removed.

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_remote_source.py
+++ b/atomic_reactor/plugins/pre_resolve_remote_source.py
@@ -136,8 +136,8 @@ class ResolveRemoteSourcePlugin(PreBuildPlugin):
 
     def source_request_to_json(self, source_request):
         """Create a relevant representation of the source request"""
-        required = ('ref', 'repo')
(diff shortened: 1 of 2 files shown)


lcarva:
It would be great to add `packages` to this tuple to ensure it's working as expected: https://github.com/containerbuildsystem/atomic-reactor/blob/master/tests/plugins/test_resolve_remote_source.py#L295

ben-alkov:
@arewm; We need a "[Signed-off-by](https://github.com/apps/dco)" line; when you commit, do `git commit [-s|--signoff] ...`

ben-alkov:
@arewm; Also, please [squash](https://gist.github.com/patik/b8a9dc5cd356f9f6f980#the-harder-and-less-flexible-way)[1] your commits. We're fine with having unit tests go into the same commit as the feature they support.

[1] This gist looks good to me, but maybe ask @lcarva, or @MartinBasti for specifics if it doesn't seem clear enough.

arewm:
Thanks, @ben-alkov, I have fixed that.

lcarva:
Updated release notes draft:
> - Verify `packages` is always provided in Cachito's response

PR #1546, opened 2020-08-19
Fix py2/py3 reqs

MartinBasti:


lkolacek (approved):
LGTM :+1:

MartinBasti:
@lkolacek updated

tests/plugins/test_resolve_composes.py
```diff
@@ -1363,7 +1364,10 @@ def run_plugin_with_args(self, workflow, plugin_args=None,
         if expect_error:
             with pytest.raises(PluginFailedException) as exc_info:
                 runner.run()
-            assert expect_error in str(exc_info.value)
+            if hasattr(expect_error, 'search'):  # py2/3 compat, different re.Compile objects
```
tkdchen, line 1367: How about use `six.PY2` or `six.PY3`? No comment is required and the code is much straightforward.

MartinBasti: Can you provide exaple? I don't think that using six makes code more straightforward in this case

ben-alkov: Actually, I'd like to see the comment explain *which* Python re.Compile object has  `search` (so I don't have to look it up :p)

MartinBasti:
> Actually, I'd like to see the comment explain which Python re.Compile object has search (so I don't have to look it up :p)

```
>>> r = re.compile('test')
# py2
>>> type(r)
<type '_sre.SRE_Pattern'>

# py3
>>> type(r)
<class 're.Pattern'>
```

I'm not sure if I understand what you mean. It differs between python versions. In py2 it's not even object

ben-alkov:
> I'm not sure if I understand what you mean. It differs between python versions. In py2 it's not even object

IOW, maybe the comment could be something like

```python
if hasattr(expect_error, 'search'):  # For Py 3; Py 2's re isn't an object
    assert expect_error.search(str(exc_info.value))
else:
    assert expect_error in str(exc_info.value) # Py 2
```
or whatever you're comfortable with :)

MartinBasti:
Oh, it's misunderstanding. if-else is there because I want to support both string and regexp.

I added comment about py2/3 compatibility, because I cannot use `isinstance(expect_error, re.Pattern)` in py2

MartinBasti:
No release notes, no functional change

PR #1557, opened 2020-10-27
Fix unwanted tracebacks recorded

SalatskySal:

### DIFF
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -397,7 +397,7 @@ class SendMailPlugin(ExitPlugin):
                 try:
                     koji_task_owner_emails = self._get_koji_owners()
                 except Exception:
-                    self.log.exception("Failed to include a package owner")
+                    self.log.info("Failed to include a package owner")
                 else:
                     receivers_list += koji_task_owner_emails
 
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -163,7 +163,7 @@ class SendMailPlugin(ExitPlugin):
             metadata = get_build_json().get("metadata", {})
             self.original_koji_task_id = int(metadata['labels']['original-koji-task-id'])
         except Exception:
-            self.log.exception("Failed to fetch original koji task ID")
+            self.log.info("Failed to fetch original koji task ID")
             self.original_koji_task_id = None
         else:
             self.log.info("original Koji task ID: %s", self.original_koji_task_id)
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -330,7 +330,10 @@ class SendMailPlugin(ExitPlugin):
         else:
             if not self.email_domain:
                 raise RuntimeError("Empty email_domain specified")
-            return '@'.join([obj['name'], self.email_domain])
+            elif not obj.get('name'):
+                raise RuntimeError("Koji task owner name is missing")
+            else:
+                return '@'.join([obj['name'], self.email_domain])
 
     def _get_koji_submitter(self):
         koji_task_owner = get_koji_task_owner(self.session,
@@ -386,7 +389,7 @@ class SendMailPlugin(ExitPlugin):
                 try:
                     koji_task_owner_email = self._get_koji_submitter()
                 except Exception:
-                    self.log.exception("Failed to include a task submitter")
+                    self.log.info("Failed to include a task submitter")
                 else:
                     receivers_list.append(koji_task_owner_email)
 
--- a/tests/plugins/test_sendmail.py
+++ b/tests/plugins/test_sendmail.py
@@ -17,6 +17,7 @@ from atomic_reactor.plugins.exit_koji_import import KojiImportPlugin
 from atomic_reactor.plugins.pre_reactor_config import (ReactorConfigPlugin,
                                                        WORKSPACE_CONF_KEY,
                                                        ReactorConfig)
+from atomic_reactor.utils.koji import get_koji_task_owner
 from tests.util import add_koji_map_in_workflow
 from osbs.api import OSBS
 from osbs.exceptions import OsbsException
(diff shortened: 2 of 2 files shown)


atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -159,14 +159,18 @@ def __init__(self, tasker, workflow,
             else:
                 self.log.info("No koji task")
 
+        self.original_koji_task_id = None
         try:
             metadata = get_build_json().get("metadata", {})
-            self.original_koji_task_id = int(metadata['labels']['original-koji-task-id'])
+            original_koji_task_id = metadata['labels'].get('original-koji-task-id')
```
MartinBasti, line 165:
```suggestion
            original_koji_task_id = metadata.get('labels', {}).get('original-koji-task-id')
```

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -334,7 +334,11 @@ def _get_email_from_koji_obj(self, obj):
         else:
             if not self.email_domain:
                 raise RuntimeError("Empty email_domain specified")
-            return '@'.join([obj['name'], self.email_domain])
+            elif obj.get('name'):
+                return '@'.join([obj['name'], self.email_domain])
+            else:
+                self.log.info('Koji object name is missing')
+                return None
```
tkdchen, line 341: If I understand the code path correctly, if the obj has no key `name`, raising an error should be the right way to ignore the obj. Otherwise, if return None, it will cause None is included in receivers list and will be passed to `sendmail`.

BTW, is there a test case covering the case of this change?

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -159,14 +159,18 @@ def __init__(self, tasker, workflow,
             else:
                 self.log.info("No koji task")
 
+        self.original_koji_task_id = None
         try:
             metadata = get_build_json().get("metadata", {})
-            self.original_koji_task_id = int(metadata['labels']['original-koji-task-id'])
```
tkdchen, line 164: This line is wrapped into:

```python
try:
    ...
except Exception:
```

Looks the error `"KeyError: 'original-koji-task-id'"` mentioned in issue should have no chance to be raised outside. How can it happen?

I'm on master branch at commit 8f2fc140552835fd7e6a216adb71dc5548fadad3.

atomic_reactor/plugins/exit_sendmail.py
SalatskySal: Good catch, thanks!
I checked test_sendmail and looks like there is no coverage for this

atomic_reactor/plugins/exit_sendmail.py
SalatskySal: @tkdchen I'd like to present [this commit](https://github.com/containerbuildsystem/atomic-reactor/commit/fef5b7ed42771f1d199b8fdf2ce2473c0e8689a3#diff-a078811c45cb9386db6ea253e9c97d6d6a3265bf167a5234032fb30413506217), where was the same problem.
Actually, you can reproduce it using the command from the issue description. It looks weird, I know, but it is.

rcerven: exception is logged because of this:

self.log.exception("Failed to fetch original koji task ID")

just change it to self.log.warning 

and no other changes are necessary

all exceptions are catched correctly, be with this we are also logging them

SalatskySal: Nevertheless, this change describes better what happened, so I'd leave this unchanged

ben-alkov:
Needs unit tests

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -159,14 +159,18 @@ def __init__(self, tasker, workflow,
...
+            original_koji_task_id = metadata.get('labels', {}).get('original-koji-task-id')
         except Exception:
-            self.log.exception("Failed to fetch original koji task ID")
-            self.original_koji_task_id = None
+            self.log.info("Failed to fetch original koji task ID")
         else:
-            self.log.info("original Koji task ID: %s", self.original_koji_task_id)
+            if original_koji_task_id:
+                self.original_koji_task_id = int(original_koji_task_id)
```
tkdchen, line 170: @SalatskySal 

Moving `int()` call here changes the original behavior. The change above in try-except is

```python
try:
    metadata = get_build_json().get("metadata", {})
    original_koji_task_id = metadata.get('labels', {}).get('original-koji-task-id')
except Exception:
    self.log.info("Failed to fetch original koji task ID")
```

My understand of the `get_build_json()` call is that the exit_sendmail plugin should run as one step through the container build workflow, there should be no use case to run this plugin itself only. Thus, through the workflow, `get_build_json` should be called already in the previous plugins before exit_sendmail runs. If no `$BUILD`, the build will fail earlier.

So, that two lines inside try-except does not raise any error in this context. The original code also catches potential error raised from `int()`, but this new change doesn't.

IMO, Robert's suggestion is a safe way. It keeps the change in a small scope and the original behavior, and does not introduce any new issue potentially.

Just FYI. :)

SalatskySal: @tkdchen thanks for the great explanation! It makes sense

atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -330,7 +330,10 @@ def _get_email_from_koji_obj(self, obj):
         else:
             if not self.email_domain:
                 raise RuntimeError("Empty email_domain specified")
-            return '@'.join([obj['name'], self.email_domain])
+            elif 'name' not in obj:
+                raise RuntimeError("Koji object name is missing")
+            else:
```
MartinBasti, line 335: How this will prevent tracebacks in logs, when in `_get_receivers_list` this Exception will be caught and traceback printed via `log.exception()`

SalatskySal: Just change log.exception into log.info, as @rcerven proposed above

rcerven:
change also self.log.exception("Failed to include a package owner")  to self.log.info

MartinBasti (approved):

lkolacek (approved):

MartinBasti:
no release notes

PR #1562, opened 2020-11-23
exclude package from sources only based on key/value from denylist wi…

rcerven:
…thout leading '/'

as package names from cachito can have differnt form of names (depedning on package types):
github.com/some/package
some/package
@some/package

* CLOUDBLD-3411

### DIFF
--- a/atomic_reactor/plugins/pre_fetch_sources.py
+++ b/atomic_reactor/plugins/pre_fetch_sources.py
@@ -408,8 +408,12 @@ class FetchSourcesPlugin(PreBuildPlugin):
 
     def _check_if_package_excluded(self, packages, denylist_sources, remote_archive):
         # check if any package in cachito json matches excluded entry
+        # strip leading os.sep as package names can include git path with '/' before package name
+        # or just package name, or package name with leading '@' depending on package type
+        denylist_packages = {k.lstrip(os.sep) for k in denylist_sources}
+
         for package in packages:
-            for exclude_path in denylist_sources:
+            for exclude_path in denylist_packages:
                 if package.get('name').endswith(exclude_path):
                     self.log.debug('Package excluded: "%s" from "%s"', package.get('name'),
                                    remote_archive)
--- a/tests/plugins/test_fetch_sources.py
+++ b/tests/plugins/test_fetch_sources.py
@@ -193,7 +193,8 @@ def get_remote_url(koji_build, file_name=REMOTE_SOURCES_FILE):
 
 
 def mock_koji_manifest_download(tmpdir, requests_mock, retries=0, dirs_in_remote=('app', 'deps'),
-                                files_in_remote=(), cachito_package_names=None):
+                                files_in_remote=(), cachito_package_names=None,
+                                change_package_names=True):
     class MockBytesIO(io.BytesIO):
         reads = 0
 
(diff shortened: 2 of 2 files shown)


tkdchen:
LGTM. There is a typo "differnt" in the commit message.

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -408,8 +408,12 @@ def _create_full_remote_sources_map(self, request_session, remote_sources_map,
 
     def _check_if_package_excluded(self, packages, denylist_sources, remote_archive):
         # check if any package in cachito json matches excluded entry
+        # strip leading os.sep as package names can include git path with '/' before package name
+        # or just package name, or package name with leading '@' depending on package type
+        denylist_packages = [k[1:] for k in denylist_sources]
```
mkosiarc, line 413: How about changing this in the __get_denylist_sources_  method where we adding the os.separator just to remove it later?

rcerven: it is used in 2 places, for checking package name and for excluding path (this uses leading separator)

chmeliik:
Sorry for troublemaking, but is this the agreed-upon solution for CLOUDBLD-3411? There was also the question of excluding based on the repo name in addition to the package names, is that no longer relevant? CC @arewm

rcerven:
> Sorry for troublemaking, but is this the agreed-upon solution for CLOUDBLD-3411? There was also the question of excluding based on the repo name in addition to the package names, is that no longer relevant? CC @arewm

this is how discoball is doing it afaik, @lcarva 
and we need to do it the same way

mkosiarc (approved):

rcerven:
> LGTM. There is a typo "differnt" in the commit message.

fixed

atomic_reactor/plugins/pre_fetch_sources.py
lcarva: Should we use `lstrip` to better convey the meaning?
```python
denylist_packages = [k.lstrip('/') for k in denylist_sources]
```

lcarva:
> > Sorry for troublemaking, but is this the agreed-upon solution for CLOUDBLD-3411? There was also the question of excluding based on the repo name in addition to the package names, is that no longer relevant? CC @arewm
> 
> this is how discoball is doing it afaik, @lcarva
> and we need to do it the same way

I don't think we should match on the repo name. We should revert the change in Discoball.

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: sure I can do that

lcarva (approved):

tkdchen (approved):

chmeliik (approved):

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -408,8 +408,12 @@ def _create_full_remote_sources_map(self, request_session, remote_sources_map,
 
     def _check_if_package_excluded(self, packages, denylist_sources, remote_archive):
         # check if any package in cachito json matches excluded entry
+        # strip leading os.sep as package names can include git path with '/' before package name
+        # or just package name, or package name with leading '@' depending on package type
+        denylist_packages = [k.lstrip(os.sep) for k in denylist_sources]
```
MartinBasti, line 413: can be this `set()` instead?

rcerven: sure updated

lkolacek (approved):

MartinBasti (approved):

rcerven:
rn updated: exclude package from sources only based on key/value from denylist without leading '/'

PR #1578, opened 2021-01-22
List RPM content without in-image rpm util for FROM scratch

rcerven:
* CLOUDBLD-4095


atomic_reactor/plugins/post_rpmqa.py
```diff
@@ -70,3 +87,49 @@ def gather_output(self):
...
+
+        try:
+            self.log.info('getting rpms from rpmdb: %s', rpm_cmd)
+            rpm_output = subprocess.check_output(rpm_cmd)
+        except Exception as e:
+            self.log.error("Failed to get rpms from rpmdb: %s", e)
+            raise e
+
+        if isinstance(rpm_output, bytes):
+            rpm_output = rpm_output.decode("utf-8")  # py2 & 3 compat
```
tkdchen, line 132: We don't need to do this by passing `universal_newlines`  to `check_output`.

atomic_reactor/utils/rpm.py
```diff
@@ -35,7 +35,7 @@ def get_rpm_list(tags=None, separator=';'):
     return rpms
 
 
-def rpm_qf_args(tags=None, separator=';'):
+def rpm_qf_args(tags=None, separator=';', output_list=False):
```
tkdchen, line 38: Looks adding this new parameter `output_list` adds a new technical debt, which makes the function return value in another different type. Can we just refactor this function to return value in only one type, that is a list object, and update the existing lines calling `rpm_qf_args`?

rcerven: so this has to return either

"-qa --qf '%{NAME};%{VERSION};%{RELEASE};%{ARCH};%{EPOCH};%{SIZE};%{SIGMD5};%{BUILDTIME};%{SIGPGP:pgpsig};%{SIGGPG:pgpsig}\\n'"

or

['-qa', '--qf', '%{NAME};%{VERSION};%{RELEASE};%{ARCH};%{EPOCH};%{SIZE};%{SIGMD5};%{BUILDTIME};%{SIGPGP:pgpsig};%{SIGGPG:pgpsig}\n']

if I return list and join it it will be:

'-qa --qf %{NAME};%{VERSION};%{RELEASE};%{ARCH};%{EPOCH};%{SIZE};%{SIGMD5};%{BUILDTIME};%{SIGPGP:pgpsig};%{SIGGPG:pgpsig}\n'

which is wrong as that format has to be quoted

if I return string and split it:
['-qa', '--qf', "'%{NAME};%{VERSION};%{RELEASE};%{ARCH};%{EPOCH};%{SIZE};%{SIGMD5};%{BUILDTIME};%{SIGPGP:pgpsig};%{SIGGPG:pgpsig}\\n'"]

which is wrong again because format will have additional ' at the start 

so I'll either leave it as it is or make another method

atomic_reactor/plugins/post_rpmqa.py
rcerven: ok

atomic_reactor/utils/rpm.py
MartinBasti: I don't understand, how did you get extra `'` ?

ben-alkov: Bottom line is that it's very bad practice to return two **completely** different types from the same function.

rcerven: because split quotes it

rcerven: done

ben-alkov (approved):

atomic_reactor/plugins/post_rpmqa.py
```diff
@@ -5,9 +5,19 @@
...
+import subprocess
+import tempfile
+import tarfile
 
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.utils.rpm import rpm_qf_args, parse_rpm_output
+from atomic_reactor.plugins.pre_reactor_config import get_list_rpms_from_scratch
+from docker.errors import APIError
+
+RPMDB_PATH = '/var/lib/rpm'
```
tkdchen, line 18: Can we get this path by calling `rpm.expandMacro` instead of hardcoding it?

rcerven: I'd rather keep explicit path here, so we don't get any surprises

tests/plugins/test_rpmqa.py
```diff
@@ -38,6 +46,43 @@
...
+        rpm_dir.join('Packages').write('')
+
+    archive_path = tmpdir.join('temp.tar')
+    with tarfile.open(str(archive_path), 'w') as archive_tar:
+        archive_tar.add(rpm_dir, arcname=RPMDB_DIR_NAME)
+
+    f = open(str(archive_path), 'rb')
+    for block in iter(partial(f.read, 8), b''):
+        yield block
+    f.close()
```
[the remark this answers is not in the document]
rcerven: ok

atomic_reactor/plugins/post_rpmqa.py
```diff
@@ -70,3 +87,46 @@ def gather_output(self):
                 return output
 
         raise RuntimeError('Unable to gather list of installed packages in container')
+
+    def gather_output_scratch(self):
+        rpmdb_dir = tempfile.mkdtemp()
```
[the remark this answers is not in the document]
rcerven: yeah I can do that but then I have to indent then almost whole code in there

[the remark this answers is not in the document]
MartinBasti: just from lines 112 to 128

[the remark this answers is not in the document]
rcerven: ok done

atomic_reactor/plugins/post_rpmqa.py
```diff
@@ -70,3 +87,45 @@ def gather_output(self):
...
+    def gather_output_scratch(self):
+        container_dict = self.tasker.create_container(self.image_id, command=['/bin/bash'])
+        container_id = container_dict['Id']
+        self._container_ids.append(container_id)
+
+        try:
+            bits, _ = self.tasker.get_archive(container_id, RPMDB_PATH)
+        except APIError as ex:
+            self.log.info('Could not extract rpmdb in %s : %s', RPMDB_PATH, ex)
+            return None
```
mkosiarc, line 100: when does this happen, should we not fail the build as well?

rcerven: when path isn't in the image => rpmdb doesn't exist

in that case we just want to finish without failing

mkosiarc: I see now, since list_rpms_from_scratch is true by default we don't want to affect from scratch images without rpms.

mkosiarc (approved):

MartinBasti (approved):

rcerven:
rn updated: List RPM content without in-image rpm util for FROM scratch

PR #1579, opened 2021-01-26
Use f33 to run linters

tkdchen:
Also fix detected issues:

* W0222(signature-differs)
* W0707(raise-missing-from)

### DIFF
--- a/atomic_reactor/plugin.py
+++ b/atomic_reactor/plugin.py
@@ -212,14 +212,14 @@ class PluginsRunner(object):
             plugin_name = plugin_request['name']
             try:
                 plugin_class = self.plugin_classes[plugin_name]
-            except KeyError:
+            except KeyError as e:
                 if plugin_request.get('required', True):
                     msg = ("no such plugin: '%s', did you set "
                            "the correct plugin type?") % plugin_name
                     exc = PluginFailedException(msg)
                     self.on_plugin_failed(plugin_name, exc)
                     logger.error(msg)
-                    raise exc
+                    raise exc from e
                 else:
                     # This plugin is marked as not being required
                     logger.warning("plugin '%s' requested but not available",
@@ -303,7 +303,7 @@ class PluginsRunner(object):
                         failed_msgs.append(msg)
                 else:
                     logger.error(msg)
-                    raise PluginFailedException(msg)
+                    raise PluginFailedException(msg) from ex
 
                 plugin_response = ex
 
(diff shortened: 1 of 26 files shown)


atomic_reactor/plugin.py
```diff
@@ -467,9 +467,9 @@ def run(self, *args, **kwargs):
         else:
             logger.debug("No Dockerfile path has been specified")
 
-        kwargs['buildstep_phase'] = True
-
-        plugins_results = super(BuildStepPluginsRunner, self).run(*args, **kwargs)
+        plugins_results = super(BuildStepPluginsRunner, self).run(
+            keep_going=keep_going, buildstep_phase=True
```
mkosiarc, line 471: shouldn't this be buildstep_phase=buildstep_phase ?

atomic_reactor/plugin.py
```diff
@@ -467,9 +467,9 @@ def run(self, *args, **kwargs):
         else:
             logger.debug("No Dockerfile path has been specified")
 
-        kwargs['buildstep_phase'] = True
```
tkdchen, line 470: > shouldn't this be buildstep_phase=buildstep_phase ?

@mkosiarc Here.

mkosiarc: I see, what is the reason we are not having this as default in the function?

MartinBasti (approved):

atomic_reactor/plugin.py
```diff
@@ -455,7 +455,7 @@ def __init__(self, dt, workflow, plugin_conf, *args, **kwargs):
         super(BuildStepPluginsRunner, self).__init__(
             dt, workflow, 'BuildStepPlugin', plugin_conf, *args, **kwargs)
 
-    def run(self, *args, **kwargs):
+    def run(self, keep_going=False, buildstep_phase=False):
```
MartinBasti, line 458: Can we have `buildstep_phase=True` here for consistency? It's BuildStepPlugin

mkosiarc (approved):

atomic_reactor/plugin.py
tkdchen: @MartinBasti @mkosiarc Updated the parameter's default value. PTAL.

mkosiarc: ack LGTM

ben-alkov:
Can we please pick one of 'e'/'ex'/'exc' and be consistent?

Personally I vote for 'exc'; we don't have to pay for extra characters in identifiers...

ben-alkov:
@MartinBasti; Would we prefer to see 2-3 separate commits in here?

tkdchen:
> Can we please pick one of 'e'/'ex'/'exc' and be consistent?
> 
> Personally I vote for 'exc'; we don't have to pay for extra characters in identifiers...

Historically, I used to write `e` to reference an error or exception (different programming languages call this kind of thing differently). However, while reviewing the Python document about `raise ... from`, I found out `exc` is much expressive in this context than others.

After we decide whichever name to be used, I suggest to update the name through the code to keep consistency.

PR #1585, opened 2021-02-15
Increase total retries

mkosiarc:
CLOUDBLD-4333

### DIFF
--- a/atomic_reactor/utils/retries.py
+++ b/atomic_reactor/utils/retries.py
@@ -14,8 +14,7 @@ from urllib3.util import Retry
 from atomic_reactor.constants import (HTTP_CLIENT_STATUS_RETRY,
                                       HTTP_MAX_RETRIES,
                                       HTTP_BACKOFF_FACTOR,
-                                      HTTP_REQUEST_TIMEOUT,
-                                      HTTP_CONNECTION_ERROR_RETRIES)
+                                      HTTP_REQUEST_TIMEOUT)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


atomic_reactor/utils/retries.py
```diff
@@ -49,15 +48,13 @@ def hook_log_error_response_content(response, *args, **kwargs):
...
-                                  delay=HTTP_BACKOFF_FACTOR, method_whitelist=None,
-                                  raise_on_status=True):
+                                  times=HTTP_MAX_RETRIES, delay=HTTP_BACKOFF_FACTOR,
+                                  method_whitelist=None, raise_on_status=True):
     if _http_retries_disabled():
         times = 0
 
     retry = Retry(
         total=int(times),
-        connect=connect,
```
lcarva, line 60: > connect (int) –
>
> How many connection-related errors to retry on.
>
> These are errors raised before the request is sent to the remote server, which we assume has not triggered the server to process the request.
> 
> Set to 0 to fail on the first retry of this type.

https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html

Why are we claiming this is not needed? The default value appears to be `None`: https://github.com/urllib3/urllib3/blob/main/src/urllib3/util/retry.py#L177

rcerven: 'Total number of retries to allow. Takes precedence over other counts.'

rcerven (approved):

atomic_reactor/utils/retries.py
mkosiarc: the total has higher priority and retries on ALL errors (the documentation is a bit vague) is says:

"Total number of retries to allow. Takes precedence over other counts."

But you don't actually need to specify separate retries attempts for different errors if you don't want to.
at the moment these 10 retries will retry on all errors.

the code is more convincing than the documentation
https://github.com/urllib3/urllib3/blob/main/src/urllib3/util/retry.py

I could provide you with the testing I have done, and from the brew tasks linked in those related jira isssues you can see that osbs did in fact retry on "connection related" error and our initial assumption as to why was wrong. Please read my comment in CLOUDBLD-4333

MartinBasti (approved):

lkolacek (approved):

atomic_reactor/utils/retries.py
lcarva: Thank you for the explanation. I've chatted with Robert offline about this and he cleared up my confusion. Resolving this.

lcarva (approved):

PR #1588, opened 2021-02-17
Update operator pinning plugin - other metadata replacements

ben-alkov:
Take non-pullspec metadata from _fetch_operator_csv_modifications and use it to add/append/update/replace data in the OperatorManifest

- CLOUDBLD-4179

### DIFF
--- a/atomic_reactor/plugins/pre_pin_operator_digest.py
+++ b/atomic_reactor/plugins/pre_pin_operator_digest.py
@@ -7,7 +7,6 @@ of the BSD license. See the LICENSE file for details.
 """
 
 import logging
-from typing import Iterator, Sequence
 
 from osbs.utils import Labels, ImageName
 
@@ -20,7 +19,8 @@ from atomic_reactor.constants import (
 from atomic_reactor.util import (RegistrySession,
                                  RegistryClient,
                                  has_operator_bundle_manifest,
-                                 read_yaml_from_url, df_parser)
+                                 read_yaml_from_url, df_parser,
+                                 terminal_key_paths,)
 from osbs.utils.yaml import (
     load_schema,
     validate_with_schema,
@@ -29,24 +29,6 @@ from atomic_reactor.plugins.pre_reactor_config import get_operator_manifests
 from atomic_reactor.plugins.build_orchestrate_build import override_build_kwarg
 from atomic_reactor.utils.operator import OperatorManifest
 from atomic_reactor.utils.retries import get_retrying_requests_session
-
-
-def terminal_key_paths(obj: dict) -> Iterator[Sequence]:
-    """Generates path to all terminal keys of nested dicts by yielding
-    tuples of nested dict keys represented as path
-
-    From `{'a': {'b': {'c': 1, 'd': 2}}}` yields `('a', 'b', 'c')`, `('a', 'b', 'd')`
-    """
-    stack = [
-        (obj, tuple())
-    ]
-    while stack:
-        data, path = stack.pop()
-        if isinstance(data, dict):
-            for k, v in data.items():
-                stack.append((v, path + (k, )))
-        else:
-            yield path
 
 
 class PinOperatorDigestsPlugin(PreBuildPlugin):
--- a/atomic_reactor/plugins/pre_pin_operator_digest.py
+++ b/atomic_reactor/plugins/pre_pin_operator_digest.py
@@ -302,6 +302,13 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
             self.log.info("Creating relatedImages section in %s", operator_csv.path)
             operator_csv.set_related_images()
 
+            operator_csv_modifications = self._fetch_operator_csv_modifications()
+            if operator_csv_modifications:
+                operator_csv.modifications_append(
+                    operator_csv_modifications.get('append', {}))
+                operator_csv.modifications_update(
+                    operator_csv_modifications.get('update', {}))
+
             operator_csv.dump()
         else:
             self.log.warning("%s has a relatedImages section, skipping", operator_csv.path)
--- a/atomic_reactor/utils/operator.py
+++ b/atomic_reactor/utils/operator.py
@@ -200,6 +200,37 @@ def check_csv(data, schema):
     validate_with_schema(data, schema)
 
 
+def modify_dict_recursively(target, mods, append=False):
+    """Apply 'mods' dictionary to 'target' dictionary for map entry values,
+
+    creating map entries if missing. If `append` flag is set, update/create
+    leaf lists (also creating any missing mapping entries).
+
+    Properly handles the case where either or both dicts are nested.
+
+    :param target: a dict
+    :param mods: a dict
+    :param append: boolean
+    :return: target, modified
+    """
+    for key, value in mods.items():
+        if isinstance(value, dict):
+            # recursive call
+            target[key] = modify_dict_recursively(target.get(key, {}), value, append=append)
+        else:
+            if append:
+                field = target.setdefault(key, [])
+                if not isinstance(field, list):
+                    # There's a type mismatch in the original CSV - we only *ever* allow
+                    # a list for appending.
+                    raise CSVModifyError('CSV value to append to is not a list'
+                                         f' (found {target[key]})')
+                field.extend(value)
+            else:
+                target[key] = value
+    return target
+
+
 def is_dict(obj):
     """
     Check if object is a dict or the ruamel.yaml equivalent of a dict
@@ -341,6 +372,12 @@ class NotOperatorCSV(Exception):
 class CSVValidationError(Exception):
     """
     CSV validation against schema failed
+    """
+
+
+class CSVModifyError(Exception):
+    """
+    Problem modifying the CSV, i.e. via `modifications_append` or `modifications_update`
     """
 
 
@@ -733,6 +770,28 @@ class OperatorCSV(object):
             # Doesn't matter if string was not a pullspec, it will simply not match anything
             # in replacement_pullspecs and no replacement will be done
             self._replace_unnamed_pullspec(obj, k_or_i, replacements)
+
+    def modifications_append(self, append_mods):
+        """Append to a list entry in `self.data` (or add a list, if one does not exist)
+
+        Also checks the result against schema.
+
+        :param append_mods: a dict
+        :return: None; this modifies 'self.data' in-place
+        """
+        modify_dict_recursively(self.data, append_mods, append=True)
+        check_csv(self.data, _mini_csv_schema)
+
+    def modifications_update(self, update_mods):
+        """Update a dict entry in `self.data` (or add a dict entry, if one does not exist)
+
+        Also checks the result against schema.
+
+        :param update_mods: a dict
+        :return: None; this modifies 'self.data' in-place
+        """
+        modify_dict_recursively(self.data, update_mods)
+        check_csv(self.data, _mini_csv_schema)
 
 
 class OperatorManifest(object):
--- a/tests/plugins/test_pin_operator_digests.py
+++ b/tests/plugins/test_pin_operator_digests.py
@@ -27,6 +27,9 @@ from atomic_reactor.plugins.pre_pin_operator_digest import (
     PinOperatorDigestsPlugin,
     PullspecReplacer,
     terminal_key_paths)
+
+from atomic_reactor.source import PathSource
+
 from tests.util import OPERATOR_MANIFESTS_DIR
 
 from osbs.exceptions import OsbsValidationException
@@ -128,7 +131,6 @@ def mock_env(docker_tasker, repo_dir, orchestrator,
 
     mock_dockerfile(repo_dir, df_base, df_operator_label)
 
-    from atomic_reactor.source import PathSource
     env.workflow.source = PathSource('path', str(repo_dir))
 
     # NOTE: is this path correct?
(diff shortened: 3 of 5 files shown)


atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -449,6 +453,45 @@ def _get_replacement_pullspecs_OSBS_resolution(self, pullspecs):
...
+                    for item in mods[action][action_field]:
+                        if action == 'append':
+                            if isinstance(mods[action][action_field][item], list):
+                                # Append to existing list
+                                if operator_manifest.csv.data[action_field].get(item):
+                                    for value in mods[action][action_field][item]:
+                                        print(value)
+                                        operator_manifest.csv.data[action_field][item].append(value)
+                                else:  # Add list - it doesn't exist in target CSV
+                                    operator_manifest.csv.data[action_field][item] = mods[action][action_field][item]
```
MartinBasti, line 480: All modifications to CSV should be done via OperatorManifest object, don't bypass it, create methods there.

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -449,6 +453,45 @@ def _get_replacement_pullspecs_OSBS_resolution(self, pullspecs):
...
+            if (mods.get(action)):
+                # Loop over mod data and apply action
+                for action_field in mods[action]:
+                    for item in mods[action][action_field]:
+                        if action == 'append':
+                            if isinstance(mods[action][action_field][item], list):
+                                # Append to existing list
+                                if operator_manifest.csv.data[action_field].get(item):
+                                    for value in mods[action][action_field][item]:
+                                        print(value)
```
MartinBasti, line 477: :eyes:

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -449,6 +453,45 @@ def _get_replacement_pullspecs_OSBS_resolution(self, pullspecs):
...
+        passed in
+
+        :param operator_manifest: an OperatorManifest, as created by `from_directory()`
+        :type operator_manifest: instance of OperatorManifest
+        :return: None
+        """
+
+        def do_action(operator_manifest, action):
+            mods = self._fetch_operator_csv_modifications()
+            if (mods.get(action)):
```
MartinBasti, line 468: https://en.wikipedia.org/wiki/Cyclomatic_complexity

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -449,6 +453,45 @@ def _get_replacement_pullspecs_OSBS_resolution(self, pullspecs):
...
+                # Loop over mod data and apply action
+                for action_field in mods[action]:
+                    for item in mods[action][action_field]:
+                        if action == 'append':
+                            if isinstance(mods[action][action_field][item], list):
+                                # Append to existing list
+                                if operator_manifest.csv.data[action_field].get(item):
```
MartinBasti, line 475: Could be simplified
```python
field = operator_manifest.csv.data[action_field].setdefault(item, [])
field.extend(mods[action][action_field][item])
```

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -449,6 +453,45 @@ def _get_replacement_pullspecs_OSBS_resolution(self, pullspecs):
...
+                                #  Raise something - we're expecting a list
+                                pass
+                        elif action == 'update':
+                            # Update existing dict
+                            if operator_manifest.csv.data.get(action_field):
+                                operator_manifest.csv.data[action_field][item] = mods[action][action_field][item]
+                            else:  # Add dict ("object") - it doesn't exist in target CSV
+                                operator_manifest.csv.data[action_field]= mods[action][action_field]
+
+        if self._fetch_operator_csv_modifications():
```
MartinBasti, line 491: This is fetched 3 times, should be just once IMHO

atomic_reactor/plugins/pre_pin_operator_digest.py
MartinBasti: https://en.wikipedia.org/wiki/Separation_of_concerns

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -449,6 +453,45 @@ def _get_replacement_pullspecs_OSBS_resolution(self, pullspecs):
...
+            do_action(operator_manifest, 'append')
```
MartinBasti, line 492: https://en.wikipedia.org/wiki/Single-responsibility_principle  can we obey SRP and have one method for update and one for appending? Just pass only what they need to them, not the whole modification file

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -150,6 +150,9 @@ def run_in_orchestrator(self):
               replaced: <bool>  # plugin modified pullspec (repo/registry/tag changed)
             - original: ........
           created_by_osbs: <bool>
+        SIDE-EFFECT:
```
MartinBasti, line 153: Orchestrator cannot replace data in CSV, that's happening on worker

MartinBasti (changes requested):

atomic_reactor/plugins/pre_pin_operator_digest.py
ben-alkov: @MartinBasti; Agreed. Should I make it a field/property of `PinOperatorDigestsPlugin`?

atomic_reactor/plugins/pre_pin_operator_digest.py
ben-alkov: Nice!

atomic_reactor/plugins/pre_pin_operator_digest.py
ben-alkov: ACK

atomic_reactor/plugins/pre_pin_operator_digest.py
ben-alkov: @MartinBasti; So I should be calling `_replace_other_metadata_from_csv_modifications()` from `run_in_worker()` instead?

rcerven: run_in_worker() inside 'if not operator_csv.has_related_images()'

when has_releated_images is there freshmaker shouldn't be rebuilding images

atomic_reactor/plugins/pre_pin_operator_digest.py
rcerven: just call it once at the begin of _replace_other_metadata_from_csv_modifications and then use it

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -449,6 +453,45 @@ def _get_replacement_pullspecs_OSBS_resolution(self, pullspecs):
 
         return replacements
 
+    def _replace_other_metadata_from_csv_modifications(self, operator_manifest):
```
rcerven, line 456: pass only operator_manifest.csv as you need only that anyway

atomic_reactor/plugins/pre_pin_operator_digest.py
ben-alkov: @rcerven; That leaves us with a total of two calls

atomic_reactor/plugins/pre_pin_operator_digest.py
ben-alkov: @rcerven; ACK exactly what I needed.

atomic_reactor/plugins/pre_pin_operator_digest.py
rcerven, line 475: @mprahl @MartinBasti 
is it possible that freshmaker will pass:
append : spec->skips->['some']

but in CSV even 'spec' doesn't exist?

if yes we should also first check if even : operator_manifest.csv.data[action_field] exists (like in update)

ben-alkov: @rcerven; I could be wrong, but I believe that the top-level 'spec' is required by our schema check. Regardless, I've since refactored and it's significantly more robust.

atomic_reactor/plugins/pre_pin_operator_digest.py
rcerven: huh? why 2? 1 call is on orchestrator another on worker

atomic_reactor/plugins/pre_pin_operator_digest.py
rcerven: ok I've 'spec' & 'metadata' are required in schema so we are good

atomic_reactor/utils/operator.py
```diff
@@ -734,6 +734,28 @@ def _replace_pullspecs_not_in_annotations(self, obj, k_or_i, replacements):
...
+                else:
+                    #  Raise something - we're expecting a list
+                    pass
+
+    def _mods_update(self, update_mods):
+        if not update_mods:
+            return
+        for mod in update_mods:
+            for key in update_mods[mod]:
+                if not self.data.get(mod):
```
rcerven, line 754: 'spec' & 'metadata' are required in schema

atomic_reactor/utils/operator.py
```diff
@@ -734,6 +734,28 @@ def _replace_pullspecs_not_in_annotations(self, obj, k_or_i, replacements):
             # in replacement_pullspecs and no replacement will be done
             self._replace_unnamed_pullspec(obj, k_or_i, replacements)
 
+    def _mods_append(self, append_mods):
+        if not append_mods:
+            return
+        for mod in append_mods:
+            for key in append_mods[mod]:
+                item = append_mods[mod][key]
```
rcerven, line 742: you have to first check if key exists ('skip' doesn't even have to be there)

if key exists then check if it is list => append, or None/empty  => assign list

atomic_reactor/utils/operator.py
ben-alkov: @rcerven; I don't understand your comment in this context.

If the existing CSV is incorrect, it should have already been caught elsewhere; and the modifications file is *not* a complete CSV... so...?

rcerven: you don't need if/else here, since 'mod' will be always in CSV

you can do just: 
self.data[mod][key] = update_mods[mod][key]

ben-alkov: @rcerven; So, you're saying that a modifications file will *always* have an 'update' section, which will *always* contain 'metadata' and 'spec' keys? The example provided in the spec doc is prefaced with the comment that the "...JSON file provided to OSBS via a URL might look something like this" - "**MIGHT**"

atomic_reactor/utils/operator.py
ben-alkov: @rcerven; I took this: "In the example above, there could already be a “skips” array or it could be null" as referring the CSV file being modified, not the *modification* file.

atomic_reactor/utils/operator.py
rcerven: you are already passing JUST update section operator_csv._mods_update(operator_csv_modifications.get('update'))

'mod' is now spec/metadata

ben-alkov: derp. *If* an 'update' section is provided, any mods will *have to* include the full path, starting with 'metadata'/'spec' etc. Not sure what I was thinking...

atomic_reactor/utils/operator.py
```diff
@@ -734,6 +734,28 @@ def _replace_pullspecs_not_in_annotations(self, obj, k_or_i, replacements):
             # in replacement_pullspecs and no replacement will be done
             self._replace_unnamed_pullspec(obj, k_or_i, replacements)
 
+    def _mods_append(self, append_mods):
```
MartinBasti, line 737: This must be public method, you are calling "library"

mprahl:
@ben-alkov if I'm reading the code correctly, I see that it assumes the "append" and "update" values are dictionaries where each key has a value of a dictionary with primitive values. I would rather see the code traverse the dictionaries until a primitive or list of primitive values are reached. Then only those values would be updated.

For example, if we had the following modifications file (made up example). I wouldn't want OSBS to overwrite the `annotations` dictionary with `{"description": "Some new description"}`. I would just want it to modify the `description` key.
```json
{
   "update": {
       "metadata": {
           "name": "app.v1.0.1-01610399900-patched",
           "annotations": {
                 "description": "Some new description"
           }
       },
   }
}
```

The downside of not doing this is Freshmaker would have to pass in the whole `annotations` object, and thus have to read the CSV file on its end which is undesirable.

tests/plugins/test_pin_operator_digests.py
```diff
@@ -1206,3 +1208,83 @@ def test_pullspecs_replacements(self, tmpdir, docker_tasker):
...
+
+    @responses.activate
+    def test_other_metadata_replacements(self, tmpdir, docker_tasker):
+        pull_specs = ['myimage:v1.2.2']
+        test_url = "https://example.com/modifications.json"
+        modification_data = {
+            "pullspec_replacements": PULLSPEC_REPLACEMENTS,
+            "append": {
+                "spec": {
+                    "relatedImages": [
```
MartinBasti, line 1220: This is wrong example, OSBS is doing it, remove relatedImages from test

MartinBasti:
> @ben-alkov if I'm reading the code correctly, I see that it assumes the "append" and "update" values are dictionaries where each key has a value of a dictionary with primitive values. I would rather see the code traverse the dictionaries until a non-dictionary is reached. Then only those values would be updated.
> 
> For example, if we had the following modifications file (made up example). I wouldn't want OSBS to overwrite the `annotations` dictionary with `{"description": "Some new description"}`. I would just want it to modify the `description` key.
> 
> ```json
> {
>    "update": {
>        "metadata": {
>            "name": "app.v1.0.1-01610399900-patched",
>            "annotations": {
>                  "description": "Some new description"
>            }
>        },
>    }
> }
> ```
> 
> The downside of not doing this is Freshmaker would have to pass in the whole `annotations` dictionary, and thus have to read the CSV file on its end which is undesirable.

Do we want to allow modifications to everything here? There won't be any restrictions what can be modified?

mprahl:
> > @ben-alkov if I'm reading the code correctly, I see that it assumes the "append" and "update" values are dictionaries where each key has a value of a dictionary with primitive values. I would rather see the code traverse the dictionaries until a non-dictionary is reached. Then only those values would be updated.
> > For example, if we had the following modifications file (made up example). I wouldn't want OSBS to overwrite the `annotations` dictionary with `{"description": "Some new description"}`. I would just want it to modify the `description` key.
> > ```json
> > {
> >    "update": {
> >        "metadata": {
> >            "name": "app.v1.0.1-01610399900-patched",
> >            "annotations": {
> >                  "description": "Some new description"
> >            }
> >        },
> >    }
> > }
> > ```
> > 
> > 
> > The downside of not doing this is Freshmaker would have to pass in the whole `annotations` dictionary, and thus have to read the CSV file on its end which is undesirable.
> 
> Do we want to allow modifications to everything here? There won't be any restrictions what can be modified?

In my opinion, it'd be ideal for the CSV modification feature to be limited to Freshmaker, but I don't know if that adds too much complexity to OSBS.

I also don't see the harm in not restricting this as long as OSBS flags the CSV as having been modified in the Brew build metadata.

@lcarva, what are your thoughts?

lcarva:
> > > @ben-alkov if I'm reading the code correctly, I see that it assumes the "append" and "update" values are dictionaries where each key has a value of a dictionary with primitive values. I would rather see the code traverse the dictionaries until a non-dictionary is reached. Then only those values would be updated.
> > > For example, if we had the following modifications file (made up example). I wouldn't want OSBS to overwrite the `annotations` dictionary with `{"description": "Some new description"}`. I would just want it to modify the `description` key.
> > > ```json
> > > {
> > >    "update": {
> > >        "metadata": {
> > >            "name": "app.v1.0.1-01610399900-patched",
> > >            "annotations": {
> > >                  "description": "Some new description"
> > >            }
> > >        },
> > >    }
> > > }
> > > ```
> > > 
> > > 
> > > The downside of not doing this is Freshmaker would have to pass in the whole `annotations` dictionary, and thus have to read the CSV file on its end which is undesirable.
> > 
> > 
> > Do we want to allow modifications to everything here? There won't be any restrictions what can be modified?
> 
> In my opinion, it'd be ideal for the CSV modification feature to be limited to Freshmaker, but I don't know if that adds too much complexity to OSBS.
> 
> I also don't see the harm in not restricting this as long as OSBS flags the CSV as having been modified in the Brew build metadata.
> 
> @lcarva, what are your thoughts?

I believe we talked about restricting this to just isolated builds which is not necessarily limited to just Freshmaker.

atomic_reactor/utils/operator.py
tkdchen, line 737: I guess `mods` means modifications, right? Can we please use the full spelling to make it clear?

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -225,6 +225,11 @@ def run_in_worker(self):
             self.log.info("Creating relatedImages section in %s", operator_csv.path)
             operator_csv.set_related_images()
 
+            operator_csv_modifications = self._fetch_operator_csv_modifications()
```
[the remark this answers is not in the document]
MartinBasti: we cannot pass huge data directly to worker, there are limitations in OCP

atomic_reactor/utils/operator.py
ben-alkov: Sure can

tests/plugins/test_pin_operator_digests.py
ben-alkov: ACK

atomic_reactor/utils/operator.py
ben-alkov: ACK

ben-alkov:
@MartinBasti; Needs more tests, but PTAL?

atomic_reactor/utils/operator.py
```diff
@@ -734,6 +734,30 @@ def _replace_pullspecs_not_in_annotations(self, obj, k_or_i, replacements):
             # in replacement_pullspecs and no replacement will be done
             self._replace_unnamed_pullspec(obj, k_or_i, replacements)
 
+    def modifications_append(self, append_mods):
```
MartinBasti, line 737: Docstring

atomic_reactor/utils/operator.py
```diff
@@ -734,6 +734,30 @@ def _replace_pullspecs_not_in_annotations(self, obj, k_or_i, replacements):
...
+        if not append_mods:
+            return
+        for mod in append_mods:
+            for key in append_mods[mod]:
+                item = append_mods[mod][key]
+                if isinstance(item, list):
+                    self.data[mod].setdefault(key, []).extend(item)
+                else:
+                    #  Raise something - we're expecting a list
```
MartinBasti, line 746: :eyes:

atomic_reactor/utils/operator.py
```diff
@@ -734,6 +734,30 @@ def _replace_pullspecs_not_in_annotations(self, obj, k_or_i, replacements):
...
+                    pass
+        check_csv(self.data, _mini_csv_schema)
+
+    def modifications_update(self, update_mods):
```
MartinBasti, line 750: docstring

atomic_reactor/utils/operator.py
```diff
@@ -734,6 +734,30 @@ def _replace_pullspecs_not_in_annotations(self, obj, k_or_i, replacements):
...
+        if not update_mods:
+            return
+        for mod in update_mods:
+            for key in update_mods[mod]:
+                if isinstance(self.data[mod].get(key), dict):
```
MartinBasti, line 755: If I understand correctly, this code works in a following way
```
>>> data = { 'a': {'b': {'c': {'something': 1}}}}                                                                                                                                                                                          
>>> update_spec = { 'a': {'b': {'c': {'else': 9}}}}                                                                                                                                                                                        
>>> data['a']['b'].update(update_spec['a']['b'])                                                                                                                                                                                           
>>> print(data)                                                                                                                                                                                                                                   
{'a': {'b': {'c': {'else': 9}}}}
```

However I'd expect: 
```
{'a': {'b': {'c': {'else': 9, 'something': 1}}}}
```

atomic_reactor/utils/operator.py
ben-alkov: @MartinBasti; Yes... What should I raise here?

atomic_reactor/utils/operator.py
ben-alkov: Ah. It needs to be recursive.

atomic_reactor/utils/operator.py
ben-alkov: `RuntimeError`

ben-alkov:
@MartinBasti; PTAL @ tests - am I headed in the right direction? Anything I should do differently?

atomic_reactor/utils/operator.py
```diff
@@ -734,6 +734,47 @@ def _replace_pullspecs_not_in_annotations(self, obj, k_or_i, replacements):
...
+
+        def deep_update(target, updates):
+            for key, value in updates.items():
+                if isinstance(value, dict):
+                    # tldr; recursive
+                    target[key] = deep_update(target.get(key, {}), value)
+                else:
+                    target[key] = value
+            return target
+        deep_update(self.data, update_mods)
```
MartinBasti, line 775: can we have a newline here?

ben-alkov: @MartinBasti; Before the call to `deep_update()`, or after...?

ben-alkov:
@MartinBasti; While writing more unit tests, I discovered that the 'append' code won't work if the list in question isn't *directly* under a top-level key (e.g. 'spec').

e.g. trying to append to `'spec': {'install': {'spec': {'deployments': []` fails when it hits the first non-list...

atomic_reactor/utils/operator.py
MartinBasti: before

MartinBasti:
> @MartinBasti; While writing more unit tests, I discovered that the 'append' code won't work if the list in question isn't _directly_ under a top-level key (e.g. 'spec').
> 
> e.g. trying to append to `'spec': {'install': {'spec': {'deployments': []` fails when it hits the first non-list...

Should work in the same way as update (recursive lookup for attributes)

ben-alkov:
@MartinBasti; PTAL. Unit tests are broken due to schema validation (LOL), but I tested the 'append' code by hand and it *should* work as intended.

atomic_reactor/utils/operator.py
```diff
@@ -734,6 +736,53 @@ def _replace_pullspecs_not_in_annotations(self, obj, k_or_i, replacements):
...
+
+        def get_value_by_key_path(data: dict, key_path: Tuple) -> Any:
+            return reduce(getitem, key_path, data)
+
+        def set_value_by_key_path(data: dict, key_path: Tuple, append_value: Any):
+            key_ref = get_value_by_key_path(data, key_path[:-1])  # Get key_path "head"
+            key_ref.setdefault(key_path[-1], []).extend(append_value)  # Use key_path "tail"
+
+        for key_path in terminal_key_paths(append_mods):
+            append_item = (get_value_by_key_path(append_mods, key_path))
```
MartinBasti, line 758: Why parentheses are here?

MartinBasti:
Why implementation differs between update and append, when there is only difference how the final element is handled?

atomic_reactor/utils/operator.py
ben-alkov: Mistake

ben-alkov:
> Why implementation differs between update and append, when there is only difference how the final element is handled?
> 
> Update will create missing objects, but append won't. I see this as an inconsistency.

Append *does* create missing objects

atomic_reactor/utils/operator.py
```diff
@@ -200,6 +199,19 @@ def check_csv(data, schema):
...
 
 
+def modify_dict_recursively(target, mods, append=False):
+    for key, value in mods.items():
+        if isinstance(value, dict):
+            # recursive call
+            target[key] = modify_dict_recursively(target.get(key, {}), value, append=append)
+        else:
+            if append:
+                target.setdefault(key, []).extend(value)
```
rcerven, line 209: this will fail IF in CSV is for example skips: "string" instead of skips: ["string"]

tests/plugins/test_pin_operator_digests.py
```diff
@@ -1352,6 +1354,342 @@ def test_not_allowed_attributes_for_modifications(self, tmpdir, docker_tasker, m
...
+                ],
+                (
+                    "[{'foo': {'bar': ['1.2.3']}}] is not of type 'object'"
+                ),
+                id='update_not_dict'
+            ),
+        ]
+    )
+    @responses.activate
+    def test_other_metadata_replacements_fail(self, tmpdir, docker_tasker,
```
rcerven, line 1660: I wouldn't test again failing on validation, just that case when you are trying to extend string

ben-alkov: Agreed

atomic_reactor/utils/operator.py
ben-alkov: Added 'skips' to CSV schema

atomic_reactor/utils/operator.py
```diff
@@ -200,6 +199,19 @@ def check_csv(data, schema):
...
+            else:  # Got `None`, create key & value
+                target[key] = value
```
ssalatsk, line 217: It's just a minor suggestion, but why not use elif statement here?

MartinBasti: `elif` where?

atomic_reactor/utils/operator.py
```diff
@@ -200,6 +199,19 @@ def check_csv(data, schema):
...
 
+def modify_dict_recursively(target, mods, append=False):
+    for key, value in mods.items():
+        if isinstance(value, dict):
+            # recursive call
+            target[key] = modify_dict_recursively(target.get(key, {}), value, append=append)
+        else:
+            if append:
+                target.setdefault(key, []).extend(value)
+            else:  # Got `None`, create key & value
```
MartinBasti, line 216: Why `None` ?

atomic_reactor/utils/operator.py
```diff
@@ -200,6 +199,19 @@ def check_csv(data, schema):
     validate_with_schema(data, schema)
 
 
+def modify_dict_recursively(target, mods, append=False):
```
MartinBasti, line 208: Docstring

atomic_reactor/utils/operator.py
ssalatsk:
```suggestion
        elif append:
            target.setdefault(key, []).extend(value)
        else:  # Got `None`, create key & value
            target[key] = value
```

atomic_reactor/utils/operator.py
ben-alkov: Ah, that's a leftover from something else

atomic_reactor/utils/operator.py
ben-alkov: Inaccurate comment

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -302,6 +302,11 @@ def run_in_worker(self):
...
+            if operator_csv_modifications:
+                operator_csv.modifications_append(operator_csv_modifications.get('append'))
```
MartinBasti, line 307: if you'd use `get(x, {})` you don't have to add `if not ...` to every `modifications_*` method, because it's would be an empty iteration.

MartinBasti:
https://github.com/containerbuildsystem/osbs-docs/blob/master/docs/admins.rst#server-side-configuration-for-atomic-reactor

Please add also documentation about new config option ^

atomic_reactor/plugins/pre_pin_operator_digest.py
ben-alkov: ACK

atomic_reactor/utils/operator.py
```diff
@@ -200,6 +200,40 @@ def check_csv(data, schema):
...
+    for key, value in mods.items():
+        if isinstance(value, dict):
+            # recursive call
+            target[key] = modify_dict_recursively(target.get(key, {}), value, append=append)
+        else:
+            if append:
+                try:
+                    target.setdefault(key, []).extend(value)
+                except AttributeError as exc:
+                    if "object has no attribute 'extend'" in str(exc):
```
MartinBasti, line 225: :eyes:

atomic_reactor/utils/operator.py
```diff
@@ -200,6 +200,40 @@ def check_csv(data, schema):
...
+                        # There's a type mismatch in the original CSV - we only *ever* allow
+                        # a list for appending.
+                        raise CSVModifyError('CSV value to append to is not a list'
+                                             f' (found {target[key]})') from exc
+                    else:
+                        raise exc
```
MartinBasti, line 231:
```suggestion
                field = target.setdefault(key, []).extend(value)
                if not isinstance(field, list):
                       raise CSVModifyError(
                            'CSV value to append to is not a list '
                            f'(found {target[key]})'
                       ) from exc
                field.extend(value)
```

MartinBasti (approved):

tests/plugins/test_pin_operator_digests.py
```diff
@@ -1352,6 +1354,351 @@ def test_not_allowed_attributes_for_modifications(self, tmpdir, docker_tasker, m
...
+                        },
+                        'relatedImages': [{'name': 'foo-1', 'image': 'myimage:v1.2.2'}]
+                    }
+                },
+                id='append_only_deep'
+            )
+        ]
+    )
+    @responses.activate
+    def test_other_metadata_replacements_pass(self, tmpdir, docker_tasker,
```
rcerven, line 1587: missing test which is actually appending to list

for example skips is already in CSV [1,2,3]

and then in replacements in append [4]  resulting in [1,2,3,4]

rcerven (changes requested):

atomic_reactor/utils/operator.py
```diff
@@ -734,6 +771,28 @@ def _replace_pullspecs_not_in_annotations(self, obj, k_or_i, replacements):
             # in replacement_pullspecs and no replacement will be done
             self._replace_unnamed_pullspec(obj, k_or_i, replacements)
 
+    def modifications_append(self, append_mods):
+        """Append to a list entry in `self.data` (or add a list, if one does not exist)
+
+        Also checks the result against schema.
+
+        :param append_mods: a dict
```
tkdchen, line 779: Describing such parameters more will be helpful to understand these methods.

MartinBasti:
no RN part of unreleased feature

MartinBasti:
Comments addressed here https://github.com/containerbuildsystem/atomic-reactor/pull/1599

PR #1594, opened 2021-03-01
Consolidate RPM identification process

mkosiarc:


ben-alkov (approved):

atomic_reactor/plugins/post_rpmqa.py
```diff
@@ -112,10 +94,9 @@ def gather_output_scratch(self):
             tar_archive.extractall(rpmdb_dir)
 
             rpmdb_path = os.path.join(rpmdb_dir, RPMDB_DIR_NAME)
-            rpmdb_packages = os.path.join(rpmdb_path, RPMDB_PACKAGES_NAME)
 
-            if not os.path.exists(rpmdb_packages):
-                self.log.info('%s does not exist in rpmdb', RPMDB_PACKAGES_NAME)
+            if not os.path.exists(rpmdb_path) or not os.listdir(rpmdb_path):
```
rcerven, line 98: why are you changing this? this shouldn't be changed

mkosiarc: from jira - Fedora: recent fedora have changed names of files, make sure that folder is not empty don't check for exact names.

tests/plugins/test_rpmqa.py
```diff
@@ -163,10 +173,10 @@ def test_rpmqa_plugin(caplog, docker_tasker, base_from_scratch, list_rpms_from_s
         assert workflow.image_components == parse_rpm_output(ignore_autogenerated["package_list"])
 
 
-@pytest.mark.parametrize('packages_exists', [True, False])
```
rcerven, line 166: unit test should test file existence for os.listdir

atomic_reactor/plugins/post_rpmqa.py
rcerven: ack

tests/plugins/test_rpmqa.py
mkosiarc: updated the test, I extracted this part to separate testcase to be more straightforward and easier to follow.
You made a very good point as this caught a error when returning None from function gather_output but not from plugin itself.

tests/plugins/test_rpmqa.py
```diff
@@ -163,10 +178,9 @@ def test_rpmqa_plugin(caplog, docker_tasker, base_from_scratch, list_rpms_from_s
...
 @pytest.mark.parametrize('rpm_failed', [True, False])
 @pytest.mark.parametrize('get_archive_raises', [None, APIError, Exception])
-def test_rpmqa_plugin_base_from_scratch(caplog, tmpdir, docker_tasker, packages_exists, rpm_failed,
+def test_rpmqa_plugin_base_from_scratch(caplog, tmpdir, docker_tasker, rpm_failed,
```
rcerven, line 183: rename this method, it isn't anymore just for FROM scratch

tests/plugins/test_rpmqa.py
```diff
@@ -221,33 +235,57 @@ def test_rpmqa_plugin_base_from_scratch(caplog, tmpdir, docker_tasker, packages_
...
+            results = runner.run()
+            assert log_msg_getting in caplog.text
+            assert results[PostBuildRPMqaPlugin.key] == PACKAGE_LIST
+            assert workflow.image_components == parse_rpm_output(PACKAGE_LIST)
 
     log_msg = "from scratch, list_rpms_from_scratch is True, trying get rpmdb"
     assert log_msg in caplog.text
 
 
+@pytest.mark.parametrize('rpmdb_dir_exists, rpmdb_dir_empty', [(True, False), (False, False)])
```
rcerven, line 257: merge this test with previous test, which was already testing similar thing before

tests/plugins/test_rpmqa.py
```diff
@@ -46,20 +44,25 @@
 pytestmark = pytest.mark.usefixtures('user_params')
 
 
-def generate_archive(tmpdir, empty=False):
+def generate_archive(tmpdir, rpm_dir_exists=True, rpm_dir_empty=False):
     """Generate a fake tar archive
     :param bool empty: whether the archive is empty.
```
tkdchen, line 49: docstring is not updated accordingly.

tkdchen (changes requested):

tests/plugins/test_rpmqa.py
mkosiarc: updated

rcerven:
code looks good

tests look fine, but I'd still prefer to have more complex test instead of multiple simple ones, but up to others

mkosiarc:
It really don't think it's a good idea to have complex testcases especially with conditional branching, they are easier to make mistakes in, harder to read/understand and harder to mantain. That's is also one of the reasons we have that story for refactoring atomic-reactor unit-tests (sure there are other issues).
I think unit test should definitely be short, without branching and test only 1 thing if possible, splitting them had additional advantage that I could now name additional testcases and it's clearer what they test, and if they should fail you can debug much quicker where the issue is.

lkolacek:
I can't agree more with Max. Unit test should be small atomic and easy to maintain. The code looks great. Thanks, Max!

tkdchen (approved):

rcerven:
> It really don't think it's a good idea to have complex testcases especially with conditional branching, they are easier to make mistakes in, harder to read/understand and harder to mantain. That's is also one of the reasons we have that story for refactoring atomic-reactor unit-tests (sure there are other issues).
> I think unit test should definitely be short, without branching and test only 1 thing if possible, splitting them had additional advantage that I could now name additional testcases and it's clearer what they test, and if they should fail we can debug much quicker where the issue is.

yeah problem is that if we will be constantly adding just simple tests, we will end up with tests taking a lot of time

so we should balance between too complex tests and too simple tests

mkosiarc:
> > It really don't think it's a good idea to have complex testcases especially with conditional branching, they are easier to make mistakes in, harder to read/understand and harder to mantain. That's is also one of the reasons we have that story for refactoring atomic-reactor unit-tests (sure there are other issues).
> > I think unit test should definitely be short, without branching and test only 1 thing if possible, splitting them had additional advantage that I could now name additional testcases and it's clearer what they test, and if they should fail we can debug much quicker where the issue is.
> 
> yeah problem is that if we will be constantly adding just simple tests, we will end up with tests taking a lot of time
> 
> so we should balance between too complex tests and too simple tests

Not necessarily, for example when have bigger testcase with many pytest parameters it also has to run for each pytest parameters separately, so when you split it, it will run basically the same code in the same amount of time.

But even if the time necessary for running tests increased slightly, it would still be more beneficial to have more simple testcases as you would save time on understanding/creating/maintaining those tests

rcerven (approved):

MartinBasti:
Comments addressed here https://github.com/containerbuildsystem/atomic-reactor/pull/1599

tkdchen (approved):

MartinBasti (changes requested):
Blocking merging - incompatibilities between RPM DB have been identified. This issue must be resolved first.

MartinBasti:
Please add ticket to commit message and provide better description in commit message

mkosiarc:
@MartinBasti I updated the commit message and changed the destination branch, PTAL

MartinBasti (approved):

mkosiarc:
rn added

PR #1597, opened 2021-03-02
Allow accessing odcs repos via https

tkdchen:

### DIFF
--- a/atomic_reactor/plugins/pre_inject_yum_repo.py
+++ b/atomic_reactor/plugins/pre_inject_yum_repo.py
@@ -11,64 +11,64 @@ Pre build plugin which injects custom yum repository in dockerfile.
 import os
 import shutil
 from io import StringIO
+
 from atomic_reactor.constants import YUM_REPOS_DIR, RELATIVE_REPOS_PATH, INSPECT_CONFIG
 from atomic_reactor.plugin import PreBuildPlugin
+from atomic_reactor.plugins.pre_reactor_config import get_builder_ca_bundle
 from atomic_reactor.util import df_parser
 from atomic_reactor.utils.yum import YumRepo
-
-BUILDER_CA_BUNDLE = '/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem'
-CA_BUNDLE_PEM = os.path.basename(BUILDER_CA_BUNDLE)
-
-
-def add_yum_repos_to_dockerfile(repo_dir, yumrepos, df, inherited_user, base_from_scratch):
-    if df.baseimage is None:
-        raise RuntimeError("No FROM line in Dockerfile")
-
-    # Determine the USER the final image should end with
-    final_user_line = "USER " + inherited_user if inherited_user else None
-    # Look for the last USER after the last FROM... by looking in reverse
-    for insndesc in reversed(df.structure):
-        if insndesc['instruction'] == 'USER':
-            final_user_line = insndesc['content']  # we will reuse the line verbatim
-            break
-        if insndesc['instruction'] == 'FROM':
-            break  # no USER specified in final stage
-
-    shutil.copyfile(BUILDER_CA_BUNDLE, os.path.join(repo_dir, CA_BUNDLE_PEM))
-
-    # Insert the ADD line at the beginning of each stage
-    df.add_lines(
-        f'ADD {CA_BUNDLE_PEM} /tmp/{CA_BUNDLE_PEM}',
-        "ADD %s* %s" % (RELATIVE_REPOS_PATH, YUM_REPOS_DIR),
-        all_stages=True, at_start=True, skip_scratch=True
-    )
-
-    # Insert line(s) to remove the repos
-    cleanup_lines = [
-        "RUN rm -f " + " ".join(["'%s'" % repo for repo in yumrepos]),
-        f'RUN rm -f /tmp/{CA_BUNDLE_PEM}',
-    ]
-    # If needed, change to root in order to RUN rm, then change back.
-    if final_user_line:
-        cleanup_lines.insert(0, "USER root")
-        cleanup_lines.append(final_user_line)
-
-    if not base_from_scratch:
-        df.add_lines(*cleanup_lines)
 
 
 class InjectYumRepoPlugin(PreBuildPlugin):
     key = "inject_yum_repo"
     is_allowed_to_fail = False
 
-    def run(self):
-        """
-        run the plugin
-        """
-        yum_repos = {k: v for k, v in self.workflow.files.items() if k.startswith(YUM_REPOS_DIR)}
-        if not yum_repos:
-            return
-        # absolute path in containers -> relative path within context
+    def _final_user_line(self):
+        user = self._find_final_user()
+        if user:
+            return f'USER {user}'
+
+        builder = self.workflow.builder
+        if not builder.dockerfile_images.base_from_scratch:
+            inspect = builder.base_image_inspect
+            user = inspect.get(INSPECT_CONFIG).get('User')
+            if user:
+                return f'USER {user}'
+
+        return ''
+
+    def _find_final_user(self):
+        """Find the user in USER instruction in the last build stage"""
+        for insndesc in reversed(self._dockerfile.structure):
+            if insndesc['instruction'] == 'USER':
+                return insndesc['content']  # we will reuse the line verbatim
+            if insndesc['instruction'] == 'FROM':
+                break  # no USER specified in final stage
+
+    def _cleanup_lines(self):
+        lines = [
+            "RUN rm -f " + " ".join(
+                (f"'{repo_file}'" for repo_file in self.workflow.files)
+            )
+        ]
+        if self._builder_ca_bundle:
+            lines.append(f'RUN rm -f /tmp/{self._ca_bundle_pem}')
+
+        final_user_line = self._final_user_line()
+        if final_user_line:
+            lines.insert(0, "USER root")
+            lines.append(final_user_line)
+
+        return lines
+
+    def __init__(self, tasker, workflow, *args, **kwargs):
+        super().__init__(tasker, workflow, *args, **kwargs)
+        self._builder_ca_bundle = None
+        self._ca_bundle_pem = None
+        self._dockerfile = None
+
+    def _inject_into_repo_files(self):
+        """Inject repo files into a relative directory inside the build context"""
         host_repos_path = os.path.join(self.workflow.builder.df_dir, RELATIVE_REPOS_PATH)
         self.log.info("creating directory for yum repos: %s", host_repos_path)
         os.mkdir(host_repos_path)
@@ -80,8 +80,8 @@ class InjectYumRepoPlugin(PreBuildPlugin):
                 for line in input_buf:
                     updated_buf.write(line)
                     # Apply sslcacert to every repo in a repofile
-                    if line.lstrip().startswith('['):
-                        updated_buf.write(f'sslcacert=/tmp/{CA_BUNDLE_PEM}\n')
+                    if line.lstrip().startswith('[') and self._builder_ca_bundle:
+                        updated_buf.write(f'sslcacert=/tmp/{self._ca_bundle_pem}\n')
 
                 yum_repo = YumRepo(repourl=repo_filename,
                                    content=updated_buf.getvalue(),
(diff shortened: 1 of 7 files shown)


tkdchen:
Choosing `/opt/osbs-build` instead of `/tmp/` is because bandit reported the hardcoded `/tmp/` issue originally.

Does atomic-reactor already have a dedicated directory to hold such files used during build time?

ben-alkov:
> Choosing `/opt/osbs-build` instead of `/tmp/` is because bandit reported the hardcoded `/tmp/` issue originally.
> 
> Does atomic-reactor already have a dedicated directory to hold such files used during build time?

"I Am Not An Expert", but AFAICT, atomic-reactor uses [`tempfile`][] from the Python 3 libraries for temp files/dirs.

I imagine that bandit's problem is indeed with **hard-coding** the *string* 'tmp' - here's a canonical example of `tempfile` use, modified from the above link:

```python
with tempfile.TemporaryDirectory() as tmpdirname:
    BUILD_TIME_CA_BUNDLE = f'{tmpdirname}/tls-ca-bundle.pem')
    # etc...
```

`TemporaryDirectory` has all of the advantages of `mkdtemp()` [^1], but is usable by a context manager, so that removal of the temp dir and all of its contents is automatic.


[^1]: "Creates a temporary directory in the most secure manner possible. There are no race conditions in the directory’s creation. The directory is readable, writable, and searchable only by the creating user ID."

[`tempfile`]: https://docs.python.org/3.6/library/tempfile.html

tkdchen:
@ben-alkov Thanks for your explanation. The `/tmp/...` is used as the destination of instruction `ADD`during build, it's a location in the image not a place in the builder. So, I think the `with tempfile.TemporaryDirectory() as tmpdirname:` should not be helpful for this case. I asked that question is because I thought it would be good and explicit for maintainers to have one place than different plugins use different directories.

I chose the `/opt/osbs-build` instead of workaround the bandit detection. If there is other similar requirement in the future like the ca-bundle.pem, which are used temporarily during the image build, probably we can simply reuse this directory.

mkosiarc (approved):

lkolacek (approved):

lkolacek:
@tkdchen  looks good. I think it might be worth documenting this change also in osbs-docs. How about an integration test? Are we going to merge it before testing it?

MartinBasti:
with other plugins we are using `/tmp` I'd prefer to stay consistent here.

Will be /opt/.. propagated to the final image?

rcerven:
> > Choosing `/opt/osbs-build` instead of `/tmp/` is because bandit reported the hardcoded `/tmp/` issue originally.
> > Does atomic-reactor already have a dedicated directory to hold such files used during build time?
> 
> "I Am Not An Expert", but AFAICT, atomic-reactor uses [`tempfile`](https://docs.python.org/3.6/library/tempfile.html) from the Python 3 libraries for temp files/dirs.
> 
> I imagine that bandit's problem is indeed with **hard-coding** the _string_ 'tmp' - here's a canonical example of `tempfile` use, modified from the above link:
> 
> ```python
> with tempfile.TemporaryDirectory() as tmpdirname:
>     BUILD_TIME_CA_BUNDLE = f'{tmpdirname}/tls-ca-bundle.pem')
>     # etc...
> ```
> 
> `TemporaryDirectory` has all of the advantages of `mkdtemp()` [^1], but is usable by a context manager, so that removal of the temp dir and all of its contents is automatic.
> 
> EDIT: On further inspection, it seems like all you really want is something like `tmpdirname = tempfile.TemporaryDirectory()`. This will give you a nicely randomized directory name. Note that _the directory is also created_. Because of the way it's being passed around, I wouldn't consider that directory name to be _truly_ secure, but at least it'll silence bandit.
> 
> [^1]: "Creates a temporary directory in the most secure manner possible. There are no race conditions in the directory’s creation. The directory is readable, writable, and searchable only by the creating user ID."

this is about directory during build, tempfile can be created locally in buildroot, but we need to copy it into build itself

rcerven:
> with other plugins we are using `/tmp` I'd prefer to stay consistent here.
> 
> Will be /opt/.. propagated to the final image?

yeah we should just use /tmp in build, the same we are doing when hiding ubi.repo file

atomic_reactor/plugins/pre_inject_yum_repo.py
```diff
@@ -9,11 +9,38 @@
...
+from io import StringIO
+from urllib.parse import urlparse
 from atomic_reactor.constants import YUM_REPOS_DIR, RELATIVE_REPOS_PATH, INSPECT_CONFIG
 from atomic_reactor.plugin import PreBuildPlugin
+from atomic_reactor.plugins.pre_reactor_config import get_odcs
 from atomic_reactor.util import df_parser
 from atomic_reactor.utils.yum import YumRepo
 
+# The location of ca-bundle used during image build
+OSBS_BUILD_DIR = '/opt/osbs-build'
```
rcerven, line 21: use just /tmp
as we are using for hiding files => ubi.repo

atomic_reactor/plugins/pre_inject_yum_repo.py
```diff
@@ -9,11 +9,38 @@
...
+def inject_ca_bundle(dockerfile, base_from_scratch):
+    """Inject CA bundle into Dockerfile for accessing ODCS repos via HTTPS
+
+    :param dockerfile: the Dockerfile to inject the CA bundle.
+    :type dockerfile: DockerfileParser
+    :param bool base_from_scratch: whether the last stage is based on scratch
+        "FROM scrach", if it is a multistage build.
+    """
+    dockerfile.add_lines(
+        f'RUN mkdir -p {OSBS_BUILD_DIR}',
```
rcerven, line 35: when using /tmp, we don't have to care about existence of directory, neither in removing directory

atomic_reactor/plugins/pre_inject_yum_repo.py
```diff
@@ -77,7 +120,9 @@ def run(self):
             inherited_user = inspect.get(INSPECT_CONFIG).get('User', '')
         df = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
         yum_repos = list(self.workflow.files)
-        add_yum_repos_to_dockerfile(yum_repos, df, inherited_user,
-                                    self.workflow.builder.dockerfile_images.base_from_scratch)
+        base_from_scratch = self.workflow.builder.dockerfile_images.base_from_scratch
+        add_yum_repos_to_dockerfile(yum_repos, df, inherited_user, base_from_scratch)
+        if has_odcs_repo:
```
rcerven, line 125: just do it for all repos we are including, not just odcs

atomic_reactor/plugins/pre_inject_yum_repo.py
```diff
@@ -65,10 +92,26 @@ def run(self):
         self.log.info("creating directory for yum repos: %s", host_repos_path)
         os.mkdir(host_repos_path)
 
-        for repo, repo_content in self.workflow.files.items():
-            yum_repo = YumRepo(repourl=repo, content=repo_content, dst_repos_dir=host_repos_path,
-                               add_hash=False)
-            yum_repo.write_content()
+        odcs_config = get_odcs(self.workflow)
```
rcerven, line 95: don't check for odcs

atomic_reactor/plugins/pre_inject_yum_repo.py
```diff
@@ -65,10 +92,26 @@ def run(self):
...
+        has_odcs_repo = False
+
+        for repo_filename, repo_content in self.workflow.files.items():
+            # Update every repo accordingly in a repofile
+            # input_buf ---- updated ----> updated_buf
+            with StringIO(repo_content) as input_buf, StringIO() as updated_buf:
+                for line in input_buf:
+                    updated_buf.write(line)
+                    # Every repo should have only one baseurl or is not set
+                    if line.startswith('baseurl') and configured_odcs_host in line:
```
rcerven, line 106: line.lstrip().startswith('baseurl')

rcerven (changes requested):

tkdchen:
@MartinBasti @rcerven

In the latest update, the ca bundle is copied to `/tmp/`, and odcs is not checked. I'll continue updating the unit tests. Please review the update in the plugin. Thanks.

atomic_reactor/plugins/pre_inject_yum_repo.py
```diff
@@ -9,11 +9,39 @@
...
+    """Inject CA bundle into Dockerfile for accessing ODCS repos via HTTPS
+
+    :param str repo_dir: the container repository directory from where to build
+        the image.
+    :param dockerfile: the Dockerfile to inject the CA bundle.
+    :type dockerfile: DockerfileParser
+    :param bool base_from_scratch: whether the last stage is based on scratch
+        "FROM scrach", if it is a multistage build.
+    """
+    ca_bundle_filename = os.path.basename(BUILDER_CA_BUNDLE)
```
rcerven, line 33: use BUILD_TIME_CA_BUNDLE instead

atomic_reactor/plugins/pre_inject_yum_repo.py
```diff
@@ -9,11 +9,39 @@
...
+    shutil.copyfile(BUILDER_CA_BUNDLE,
+                    os.path.join(repo_dir, ca_bundle_filename))
+    dockerfile.add_lines(
+        f'ADD {ca_bundle_filename} /tmp/{BUILD_TIME_CA_BUNDLE}',
+        all_stages=True, at_start=True, skip_scratch=True
+    )
+
+    # Ensure the cert file is removed in the last non-scratch stage
+    if not base_from_scratch:
+        dockerfile.add_lines(f'RUN rm /tmp/{BUILD_TIME_CA_BUNDLE}')
```
rcerven, line 43: rm -f

the same what we are doing for repo file

atomic_reactor/plugins/pre_inject_yum_repo.py
```diff
@@ -65,10 +93,21 @@ def run(self):
...
-                               add_hash=False)
-            yum_repo.write_content()
+        for repo_filename, repo_content in self.workflow.files.items():
+            # Update every repo accordingly in a repofile
+            # input_buf ---- updated ----> updated_buf
+            with StringIO(repo_content) as input_buf, StringIO() as updated_buf:
+                for line in input_buf:
+                    updated_buf.write(line)
+                    # Apply sslcacert to every repo in a repofile
+                    if line.startswith('['):
```
rcerven, line 103: I still think that there can be spaces before that, so to be sure, I'd just lstrip it

ben-alkov (approved):

tkdchen:
The original tests are migrated to the new test `test_inject_repos`.

atomic_reactor/plugins/pre_inject_yum_repo.py
```diff
@@ -9,13 +9,18 @@
...
 """
 import os
+import shutil
+from io import StringIO
 from atomic_reactor.constants import YUM_REPOS_DIR, RELATIVE_REPOS_PATH, INSPECT_CONFIG
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.util import df_parser
 from atomic_reactor.utils.yum import YumRepo
 
+BUILDER_CA_BUNDLE = '/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem'
```
MartinBasti, line 19: Shouldn't be this configurable?

atomic_reactor/utils/yum.py
```diff
@@ -50,7 +49,7 @@ def filename(self):
         if not basename.endswith(REPO_SUFFIX):
             basename += REPO_SUFFIX
         if self.add_hash:
-            suffix = '-' + md5(self.repourl.encode('utf-8')).hexdigest()[:5]  # nosec
+            suffix = '-' + sha256sum(self.repourl, abbrev_len=5)
```
rcerven, line 52: change also docstring reflecting the change

tkdchen: Updated. PTAL.

ssalatsk (approved):

lkolacek (approved):
Code looks good to me. Should be tested with autotest from CLOUDBLD-4622 though

atomic_reactor/schemas/config.json
```diff
@@ -836,6 +836,11 @@
         },
         "additionalProperties": false,
         "required": ["url"]
+    },
+    "builder_ca_bundle": {
+        "description": "The path to the ca-bundle certificate inside the buildroot.",
+        "type": "string",
+        "default": "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
```
MartinBasti, line 843: Where is this default value used? I cannot see it in the code

atomic_reactor/plugins/pre_inject_yum_repo.py
```diff
@@ -9,75 +9,123 @@
...
-        self.log.info("creating directory for yum repos: %s", host_repos_path)
-        os.mkdir(host_repos_path)
 
-        for repo, repo_content in self.workflow.files.items():
-            yum_repo = YumRepo(repourl=repo, content=repo_content, dst_repos_dir=host_repos_path,
-                               add_hash=False)
-            yum_repo.write_content()
+        self._dockerfile = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        if self._dockerfile.baseimage is None:
+            raise RuntimeError("No FROM line in Dockerfile")
```
MartinBasti, line 121: Is this even valid Dockerfile?

tkdchen: Empty Dockerfile should not be reasonable for an image build. The plugin raises this error originally.

atomic_reactor/schemas/config.json
tkdchen: The call to `get_builder_ca_bundle`

https://github.com/containerbuildsystem/atomic-reactor/pull/1597/files#diff-65b2ed390ee23c54fcfe0ca79c38e0068d64e05f5b554e5a113f5c4a80546fc7R123

atomic_reactor/plugins/pre_inject_yum_repo.py
MartinBasti: Ok

atomic_reactor/schemas/config.json
MartinBasti: I see no fallback there, so it will fail (or return None), but it will NOT use `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` as a default path when it's not configured.

MartinBasti:
@lkolacek @ssalatsk 
I dunno why this is marked as `N/A`
`[n/a] New feature can be disabled from a configuration file`

I'd like to have a flag to disable injecting CA bundles to dockerfile

MartinBasti:
https://github.com/containerbuildsystem/osbs-docs/blob/master/docs/admins.rst#server-side-configuration-for-atomic-reactor

Please add also documentation about new config option ^

MartinBasti:
Please add issue ID to commit msg

tkdchen:
@MartinBasti 

> Please add issue ID to commit msg

Updated.

> Please add also documentation about new config option ^

Adding the `builder_ca_bundle` to reactor config makes me think more. If I understand the tls-ca-bundle certificate file correctly, it could includes multiple certificates to address the SSL authentication by multiple external services. For the issue of this PR, it is the ODCS case of accessing repositories via HTTPS. For other potential cases for OSBS in the future, the added `/tmp/tls-ca-bundle` could also be used as well.

Therefore, after rethinking, from my point of view, it should make much sense to add a flag (should be the same thing you mentioned yesterday) for ODCS integration specifically. That is, whether to use the ca-bundle for accessing repos via HTTPS is determined by two configure options, one is under the `odcs` section, e.g. `enable_https_for_repo_access`, and another one is the already added `builder_ca_bundle` at the top level of the reactor config. The basic logic is:

```
if odcs.enable_https_for_repo_access:
    if builder_ca_bundle is set:
        inject the ca-bundle
    else:
        raise configuration error
```

What do you think?

tkdchen:
@MartinBasti

Setting a feature flag for odcs is not applicable for now, because current solution is to add the `sslcacert` to every added repos not only the YUM repos generated by ODCS.

So far, the osbs-docs is updated as well and all comments have been addressed already I think.

MartinBasti:
I'd go just with one bundle and didn't complicate things ODCS vs yum repo url passed by arg

rcerven:
> I'd go just with one bundle and didn't complicate things ODCS vs yum repo url passed by arg

yes please just keep it simple and do it for all repos

PR #1603, opened 2021-03-17
Get cachito request env vars from cachito API

tkdchen:
* CLOUDBLD-3961

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_remote_source.py
+++ b/atomic_reactor/plugins/pre_resolve_remote_source.py
@@ -115,14 +115,18 @@ class ResolveRemoteSourcePlugin(PreBuildPlugin):
 
         for env_var, value_info in env_vars.items():
             build_arg_value = value_info['value']
-            if value_info['kind'] == 'path':
+            kind = value_info['kind']
+            if kind == 'path':
                 build_arg_value = os.path.join(REMOTE_SOURCE_DIR, value_info['value'])
                 self.log.debug(
                     'Setting the Cachito environment variable "%s" to the absolute path "%s"',
                     env_var,
                     build_arg_value,
                 )
-            build_args[env_var] = build_arg_value
+            elif kind == 'literal':
+                build_args[env_var] = build_arg_value
+            else:
+                raise RuntimeError(f'Unknown kind {kind} got from Cachito.')
 
         # Alias for absolute path to cachito.env script added into buildargs
         build_args[CACHITO_ENV_ARG_ALIAS] = os.path.join(REMOTE_SOURCE_DIR, CACHITO_ENV_FILENAME)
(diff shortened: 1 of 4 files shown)


MartinBasti (approved):

rcerven:
we should have test that build_args was set correctly to path or just value

atomic_reactor/plugins/pre_resolve_remote_source.py
```diff
@@ -112,21 +111,18 @@ def run(self):
...
                 self.log.debug(
                     'Setting the Cachito environment variable "%s" to the absolute path "%s"',
                     env_var,
-                    abs_path,
+                    build_arg_value,
                 )
-                build_args[env_var] = abs_path
-            else:
-                build_args[env_var] = value
+            build_args[env_var] = build_arg_value
```
rcerven, line 125: add here also:

elif value_info['kind'] == 'literal':

and fail for unknown kinds

rcerven (changes requested):

atomic_reactor/plugins/pre_resolve_remote_source.py
tkdchen: @rcerven This changes the original behavior. At this moment, I can't imagine what container builds could be affected potentially by this change. I would suggest to track this change separately and do it once we are sure it is safe to make this change.

mkosiarc (approved):

atomic_reactor/plugins/pre_resolve_remote_source.py
chmeliik: I can guarantee that `path` and `literal` are the only kinds that Cachito currently uses, but I'll leave it up to you guys.
https://github.com/release-engineering/cachito/blob/bda4abd187d9a624073f884d694bc12eaed4bab6/cachito/web/models.py#L1099

rcerven: what do you mean it changes original behavior ?

since we aren't anymore parsing environment_variables from source_request but ONLY from environment_variables endpoint, there will always be 'kind', so if there will be any new kind in the future this will break

we should explicitly check kind in order to interpret it correctly

tkdchen: @rcerven 

The original code detects the path-like env vars by regex, and adds other non-path-like env vars directly as build_args. It does not fail the build.

As @chmeliik mentioned, currently there are only two kinds `path` and `literal`, so both the original and the new change should be safe to work for users.

Validation against kinds is really worth to do as we should always validate any input data to ensure they are what OSBS requires. That is why I suggested above to handle the validation separately instead of extending the scope this PR handles.

WDYT?

MartinBasti: Separate validation sounds good to me

ben-alkov: Have we opened a story to track validation?

rcerven: @tkdchen @MartinBasti not sure what kind of validation, all I want in this PR is check explicitly kind : path,literal, and fail if something else was provided, we don't want separate story for that

rcerven (approved):

PR #1606, opened 2021-03-24
fetch kojifile sources from middleware builds and add them layer by layer to source image

hjmodi:
Fetch middleware sources using kojifile component metadata in image output and build source image using those sources

Epic MMENG-829
Story MMENG-1276

### DIFF
--- a/atomic_reactor/constants.py
+++ b/atomic_reactor/constants.py
@@ -160,6 +160,9 @@ MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST = "application/vnd.docker.distribution.manife
 MEDIA_TYPE_OCI_V1 = "application/vnd.oci.image.manifest.v1+json"
 MEDIA_TYPE_OCI_V1_INDEX = "application/vnd.oci.image.index.v1+json"
 
+# PNC related constants
+PNC_SYSTEM_USER = "newcastle"
+
 REPO_CONTAINER_CONFIG = 'container.yaml'
 REPO_CONTENT_SETS_CONFIG = 'content_sets.yml'
 REPO_FETCH_ARTIFACTS_URL = 'fetch-artifacts-url.yaml'
--- a/atomic_reactor/plugins/build_source_container.py
+++ b/atomic_reactor/plugins/build_source_container.py
@@ -51,13 +51,16 @@ class SourceContainerPlugin(BuildStepPlugin):
         fetch_sources_result = self.workflow.prebuild_results.get(PLUGIN_FETCH_SOURCES_KEY, {})
         source_data_dir = fetch_sources_result.get('image_sources_dir')
         remote_source_data_dir = fetch_sources_result.get('remote_sources_dir')
+        maven_source_data_dir = fetch_sources_result.get('maven_sources_dir')
 
         source_exists = source_data_dir and os.path.isdir(source_data_dir)
         remote_source_exists = remote_source_data_dir and os.path.isdir(remote_source_data_dir)
+        maven_source_exists = maven_source_data_dir and os.path.isdir(maven_source_data_dir)
 
-        if not source_exists and not remote_source_exists:
+        if not any([source_exists, remote_source_exists, maven_source_exists]):
             err_msg = "No SRPMs directory '{}' available".format(source_data_dir)
             err_msg += "\nNo Remote source directory '{}' available".format(remote_source_data_dir)
+            err_msg += "\nNo Maven source directory '{}' available".format(maven_source_data_dir)
             self.log.error(err_msg)
             return BuildResult(logs=err_msg, fail_reason=err_msg)
 
@@ -65,20 +68,28 @@ class SourceContainerPlugin(BuildStepPlugin):
             self.log.warning("SRPMs directory '%s' is empty", source_data_dir)
         if remote_source_exists and not os.listdir(remote_source_data_dir):
             self.log.warning("Remote source directory '%s' is empty", remote_source_data_dir)
+        if maven_source_exists and not os.listdir(maven_source_data_dir):
+            self.log.warning("Maven source directory '%s' is empty", maven_source_data_dir)
 
         image_output_dir = tempfile.mkdtemp()
         cmd = ['bsi', '-d']
-        drivers = []
+        drivers = set()
 
         if source_exists:
-            drivers.append('sourcedriver_rpm_dir')
+            drivers.add('sourcedriver_rpm_dir')
             cmd.append('-s')
             cmd.append('{}'.format(source_data_dir))
 
         if remote_source_exists:
-            drivers.append('sourcedriver_extra_src_dir')
+            drivers.add('sourcedriver_extra_src_dir')
             cmd.append('-e')
             cmd.append('{}'.format(remote_source_data_dir))
+
+        if maven_source_exists:
+            drivers.add('sourcedriver_extra_src_dir')
+            for maven_source_subdir in os.listdir(maven_source_data_dir):
+                cmd.append('-e')
+                cmd.append('{}'.format(os.path.join(maven_source_data_dir, maven_source_subdir)))
 
         driver_str = ','.join(drivers)
         cmd.insert(2, driver_str)
--- a/atomic_reactor/plugins/pre_fetch_sources.py
+++ b/atomic_reactor/plugins/pre_fetch_sources.py
@@ -13,19 +13,21 @@ import koji
 import tarfile
 import yaml
 
-from atomic_reactor.constants import PLUGIN_FETCH_SOURCES_KEY
+from atomic_reactor.constants import PLUGIN_FETCH_SOURCES_KEY, PNC_SYSTEM_USER
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.plugins.pre_reactor_config import (
     get_koji,
     get_koji_path_info,
     get_koji_session,
     get_config,
-    get_source_container
+    get_pnc,
+    get_source_container,
 )
 from atomic_reactor.source import GitSource
 from atomic_reactor.util import get_retrying_requests_session
 from atomic_reactor.download import download_url
 from atomic_reactor.metadata import label_map
+from atomic_reactor.utils import pnc
 
 
 @label_map('sources_for_koji_build_id')
@@ -34,7 +36,8 @@ class FetchSourcesPlugin(PreBuildPlugin):
     key = PLUGIN_FETCH_SOURCES_KEY
     is_allowed_to_fail = False
     SRPMS_DOWNLOAD_DIR = 'image_sources'
-    REMOTE_SOUCES_DOWNLOAD_DIR = 'remote_sources'
+    REMOTE_SOURCES_DOWNLOAD_DIR = 'remote_sources'
+    MAVEN_SOURCES_DOWNLOAD_DIR = 'maven_sources'
 
     def __init__(
         self, tasker, workflow, koji_build_id=None, koji_build_nvr=None, signing_intent=None,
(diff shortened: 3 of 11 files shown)


hjmodi:
@rcerven @MartinBasti

atomic_reactor/constants.py
```diff
@@ -160,6 +160,9 @@
 MEDIA_TYPE_OCI_V1 = "application/vnd.oci.image.manifest.v1+json"
 MEDIA_TYPE_OCI_V1_INDEX = "application/vnd.oci.image.index.v1+json"
 
+# PNC related constants
+PNC_SYSTEM_USER = "newcastle"
```
MartinBasti, line 164: Can you elaborate more about this constant? Is it really a constant?

hjmodi: yes, this is a constant we need to identify if the brew build was created by PNC. Based on that we follow different logic to look for its sources.

atomic_reactor/plugins/build_source_container.py
```diff
@@ -51,35 +51,46 @@ def run(self):
...
-            drivers.append('sourcedriver_extra_src_dir')
+            drivers.add('sourcedriver_extra_src_dir')
             cmd.append('-e')
             cmd.append('{}'.format(remote_source_data_dir))
 
+        if maven_source_exists:
+            drivers.add('sourcedriver_extra_src_dir')
+            for maven_source_subdir in os.listdir(maven_source_data_dir):
+                cmd.append('-e')
+                cmd.append('{}'.format(os.path.join(maven_source_data_dir, maven_source_subdir)))
```
MartinBasti, line 92: I hope we won't reach limits here :D

On my system:
```
xargs --show-limits
Your environment variables take up 4465 bytes
POSIX upper limit on argument length (this system): 2090639
POSIX smallest allowable upper limit on argument length (all systems): 4096
Maximum length of command we could actually use: 2086174
Size of command buffer we are actually using: 131072
Maximum parallelism (--max-procs must be no greater): 2147483647
```

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -79,26 +82,33 @@ def run(self):
         insecure = koji_config.get('insecure_download', False)
         urls = self.get_srpm_urls(signing_intent['keys'], insecure=insecure)
         urls_remote, remote_sources_map = self.get_remote_urls()
+        urls_maven = self.get_kojifile_source_urls()
 
-        if not urls and not urls_remote:
-            msg = "No srpms or remote sources found for source container," \
-                  " would produce empty source container image"
+        if not any([urls, urls_remote, urls_maven]):
```
MartinBasti, line 87: <3

atomic_reactor/utils/pnc.py
```diff
@@ -0,0 +1,41 @@
...
+    response.raise_for_status()
+
+    url = response.headers.get('Location')
+    dest_filename = os.path.basename(url)
+
+    # Often the SCM URL will be gerrit URL which do no have the filename in URL
+    # So we'll have to get the filename from the header if it's not valid.
+    if not re.fullmatch(r'^[\w\-.]+$', dest_filename):
+        dest_filename = session.head(url).headers.get(
+            "Content-disposition").split("filename=")[1].replace('"', '')
```
MartinBasti, line 39: I think I've asked this somewhere already, but I cannot find where. What happen when `Content-disposition` is not there?

hjmodi: should any of these operations fail the build will fail. Since this code will be executed with the filename is not valid, I think it should fail, even if we catch the error, koji will fail in import because of the invalid filename.

atomic_reactor/plugins/build_source_container.py
hjmodi: It is definitely a possibility. A solution could be to file an RFE in the BSI script to iterate through the subdirectories on its own to avoid having to pass the full path of each and every subdirectory.

atomic_reactor/constants.py
MartinBasti: and this is hardcoded to every PNC and cannot be changed?

atomic_reactor/utils/pnc.py
MartinBasti: Can we fail with something better than AttributeErrror?

atomic_reactor/constants.py
hjmodi: yes, this is the service account and all production builds are done using this account.

atomic_reactor/utils/pnc.py
hjmodi: In that case, we'll have to catch both AttributeError and IndexError(if the filename is not preset in 'content-disposition'). We can throw a RuntimeError but it'd probably cause confusion while debugging the source of the error.

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -120,8 +130,13 @@ def download_sources(self, sources, insecure=False, download_dir=SRPMS_DOWNLOAD_
 
         req_session = get_retrying_requests_session()
         for source in sources:
-            download_url(source['url'], dest_dir, insecure=insecure,
-                         session=req_session, dest_filename=source.get('dest'))
+            subdir = os.path.join(dest_dir, source.get('subdir', ''))
+            checksums = source.get('checksums', '')
```
rcerven, line 134: should should be either source.get('checksums')  to return None
or source.get('checksums', {}) 

because download_url is setting it to empty dict only if is None

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -252,6 +323,26 @@ def get_remote_urls(self):
...
+
+        kojifile_source = self._get_kojifile_source_urls_helper(self.koji_build)
+        kojifile_sources.extend(kojifile_source)
+
+        koji_build = self.koji_build
+
+        while 'parent_build_id' in koji_build['extra']['image']:
+            koji_build = self.session.getBuild(koji_build['extra']['image']['parent_build_id'],
+                                               strict=True)
+            kojifile_source = self._get_kojifile_source_urls_helper(self.koji_build)
```
rcerven, line 341: this should use koji_build not self.koji_build, as that is infinite loop because that doesn't ever change

hjmodi: fixed it.

atomic_reactor/plugins/pre_fetch_sources.py
hjmodi: fixed it.

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -35,6 +37,7 @@ class FetchSourcesPlugin(PreBuildPlugin):
     is_allowed_to_fail = False
     SRPMS_DOWNLOAD_DIR = 'image_sources'
     REMOTE_SOUCES_DOWNLOAD_DIR = 'remote_sources'
+    MAVEN_SOUCES_DOWNLOAD_DIR = 'maven_sources'
```
rcerven, line 40: typo  SOUCES vs SOURCES

we have typo also in REMOTE_SOUCES_DOWNLOAD_DIR

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -229,6 +244,62 @@ def _get_remote_urls_helper(self, koji_build):
...
+            source_build = self.session.getBuild(build_id, strict=True)
+            if source_build['owner_name'] == PNC_SYSTEM_USER:
+                pnc_build_id = source_build['extra']['external_build_id']
+                url, dest_filename = pnc.get_scm_archive_from_build_id(
+                                                                    api_request_url=api_request_url,
+                                                                    build_id=pnc_build_id,
+                                                                    session=req_session)
+                source = {'url': url,
+                          'subdir': source_build['nvr'],
+                          'dest': '__'.join([source_build['nvr'], dest_filename])}
```
rcerven, line 279: why are you prepending nvr to filenames? when we have already nvr directory?

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -229,6 +244,62 @@ def _get_remote_urls_helper(self, koji_build):
...
+                        source_archive = archive
+                        break
+                if not source_archive:
+                    raise RuntimeError(f"No sources found for {source_build['nvr']}")
+
+                maven_file_path = self.pathinfo.mavenfile(source_archive)
+                url = maven_build_path + '/' + maven_file_path
+                source = {'url': url,
+                          'subdir': source_build['nvr'],
+                          'dest': '__'.join([source_build['nvr'], source_archive['filename']]),
```
rcerven, line 295: same question as above

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -229,6 +244,62 @@ def _get_remote_urls_helper(self, koji_build):
...
+                                                                    session=req_session)
+                source = {'url': url,
+                          'subdir': source_build['nvr'],
+                          'dest': '__'.join([source_build['nvr'], dest_filename])}
+            else:
+                source_archive = None
+                maven_build_path = self.pathinfo.mavenbuild(source_build)
+                for archive in self.session.listArchives(buildID=source_build['build_id'],
+                                                         type='maven'):
+                    if archive['filename'].endswith('-project-sources.tar.gz'):
```
rcerven, line 285: maybe you can also look for *.src.rpm ?

rcerven: not sure which other filenames could be used for sources, I've even found one maven build which has just jar & pom, but that will fail so that is fine

atomic_reactor/plugins/pre_fetch_sources.py
hjmodi: fixed the typos.

atomic_reactor/plugins/pre_fetch_sources.py
hjmodi: When we pass the directories to the BSI, the directory name is not stored. It becomes a tar named exta-src#.tar when # is the layer index. To distinguish sources for given nvrs we have to append it to the filename itself.

atomic_reactor/plugins/pre_fetch_sources.py
hjmodi: answered above.

atomic_reactor/plugins/pre_fetch_sources.py
hjmodi: hmm, I'm not sure what sources that would result in and whether it would be a valid source build or not. We only have kojifile components for artifacts fetched using fetch-artifacts-*.

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: ah right ok

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -229,6 +245,62 @@ def _get_remote_urls_helper(self, koji_build):
...
+
+        sources = []
+        pnc_map = get_pnc(self.workflow)
+        # using urljoin here causes the API path in the base_api_url to be removed
+        api_request_url = pnc_map['base_api_url'] + '/' + pnc_map['get_scm_archive_path']
+        req_session = get_retrying_requests_session()
+
+        kojifile_build_ids = {kojifile['build_id'] for image in images
+                              for kojifile in self.session.listArchives(imageID=image['id'],
+                                                                        type='maven')}
```
rcerven, line 268: shouldn't type be 'kojifile' ???

hjmodi: The way it's implemented in brew, when we add kojifile components in brew it identifies the build type of artifacts resulting in type 'maven' for fetched artifacts.

rcerven: ahh ok

MartinBasti (approved):

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: ok

rcerven (approved):

atomic_reactor/plugins/build_source_container.py
hjmodi: Created https://github.com/containers/BuildSourceImage/issues/79

hjmodi:
release notes updated.

PR #1612, opened 2021-04-01
Update plugins to consume remote_sources parameter

mkosiarc:
CLOUDBLD-4449

Instead of receiving individual parameters related to
remote_sources such as build_args, configs, icm_url, url
plugins will now receive only one parameter that is a list of dicts
with these values.

For now only first element of the list will be used -
remote_sources[0] as multiple remote_sources are not yet supported.

This is in preparation for completing cachito-multiple-upstreams epic
CLOUDBLD-2838

### DIFF
--- a/atomic_reactor/plugins/pre_add_image_content_manifest.py
+++ b/atomic_reactor/plugins/pre_add_image_content_manifest.py
@@ -17,9 +17,8 @@ from atomic_reactor.constants import (IMAGE_BUILD_INFO_DIR, INSPECT_ROOTFS,
                                       INSPECT_ROOTFS_LAYERS,
                                       PLUGIN_ADD_IMAGE_CONTENT_MANIFEST)
 from atomic_reactor.plugin import PreBuildPlugin
-from atomic_reactor.plugins.pre_reactor_config import get_cachito
+from atomic_reactor.plugins.pre_reactor_config import get_cachito, get_cachito_session
 from atomic_reactor.util import (base_image_is_scratch, df_parser, read_yaml,
-                                 get_retrying_requests_session,
                                  read_content_sets
                                  )
 
@@ -83,16 +82,24 @@ class AddImageContentManifestPlugin(PreBuildPlugin):
         'image_contents': [],
     }
 
-    def __init__(self, tasker, workflow, remote_source_icm_url=None, destdir=IMAGE_BUILD_INFO_DIR):
+    def __init__(self, tasker, workflow, remote_sources=None, destdir=IMAGE_BUILD_INFO_DIR):
         """
         :param tasker: ContainerTasker instance
         :param workflow: DockerBuildWorkflow instance
-        :param icm_url: str, URL of the ICM from the Cachito request.
+        :param remote_sources: list of dicts, each dict contains info about particular
+        remote source with the following keys:
+            build_args: dict, extra args for `builder.build_args`, if any
+            configs: list of str, configuration files to be injected into
+            the exploded remote sources dir
+            request_id: int, cachito request id; used to request the
+            Image Content Manifest
+            url: str, URL from which to download a source archive
+            name: str, name of remote source
         :param destdir: image path to carry content_manifests data dir
         """
         super(AddImageContentManifestPlugin, self).__init__(tasker, workflow)
         self.content_manifests_dir = os.path.join(destdir, 'content_manifests')
-        self.icm_url = remote_source_icm_url
+        self.remote_sources = remote_sources
         self.dfp = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
         labels = Labels(self.dfp.labels)
         _, image_name = labels.get_name_and_value(Labels.LABEL_TYPE_COMPONENT)
@@ -103,6 +110,7 @@ class AddImageContentManifestPlugin(PreBuildPlugin):
         self._cachito_verify = None
         self._layer_index = None
         self._icm = None
+        self._cachito_session = None
 
     @property
     def layer_index(self):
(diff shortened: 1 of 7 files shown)


atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -83,16 +83,24 @@ class AddImageContentManifestPlugin(PreBuildPlugin):
...
-        :param icm_url: str, URL of the ICM from the Cachito request.
+        :param remote_sources: list of dicts, each dict contains info about particular
+        remote source with the following keys:
+            build_args: dict, extra args for `builder.build_args`, if any
+            configs: list of str, configuration files to be injected into
+            the exploded remote sources dir
+            icm_url: str, the Cachito ICM URL; used to request the
+            Image Content Manifest
+            url: str, URL from which to download a source archive
+            name: str, name of remote source
```
tkdchen, line 98: Suggest to create a "remote source" object.

tkdchen (commented):
typo in commit message `receieve`.

atomic_reactor/plugins/pre_add_image_content_manifest.py
mkosiarc: That is not a bad idea, but this was the decision taken in the design

chmeliik: Improvements over the design are always welcome :wink: 

I think the param has to be a dict though. Correct me if I'm wrong, but I think all plugin params must be JSON-compatible objects because that's how osbs-client passes them to atomic-reactor. You could convert the param into some kind of "remote source" object inside the plugin though, I'll leave that up to you.

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -134,13 +142,15 @@ def icm(self):
 
         :return: dict, the ICM as a Python dict
         """
-        if self.icm_url is None and self._icm is None:
+        if not self.remote_sources or (
+                self.remote_sources[0]["icm_url"] is None and self._icm is None
```
chmeliik, line 146: I don't think you need to check for `remote_sources[0]["icm_url"] is None`. I think checking that `remote_sources` is not None should be good enough. You could even prevent icm_url from being None in the schema (atomic_reactor/schemas/user_params.json)

The same pattern appears in other plugins too, checking that `remote_sources` is not None should usually be good enough.

atomic_reactor/plugins/pre_add_image_content_manifest.py
MartinBasti: Yup, it has to be JSONizable (if that is a word, JSON serializable)

mkosiarc: I am going to keep it as dict for now, unless someone is strongly against.

atomic_reactor/plugins/pre_add_image_content_manifest.py
mkosiarc: So if remote_sources was specified, we would expect that all values will be present and not None for each remote source?

From https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/plugins/pre_download_remote_source.py#L107

It looks like configs parameter might be empty?

chmeliik: Yeah, remote_sources will be a list of objects with
```python
build_args: dict[str, str]
configs: str  # this is actually a url
request_id: int
url: str
name: Optional[str]
```
i.e. `name` is the only one which can ever be `None`

atomic_reactor/utils/cachito.py
```diff
@@ -224,6 +224,27 @@ def get_request_env_vars(self, rid):
                    f'but the response contains: {resp.content}.')
             raise ValueError(msg) from exc
 
+    def get_image_content_manifest(self, rid):
+        """
+        Get the image content manifest from endpoint /content-manifest
+        :param int rid: cachito request id
```
rcerven, line 230: this should be list of request_ids

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -115,35 +123,19 @@ def layer_index(self):
...
-            session = get_retrying_requests_session()
-            session.verify = self.cachito_verify
-            self.log.debug('Making request to "%s"', self.icm_url)
-            response = session.get(self.icm_url)
-            response.raise_for_status()
-            self._icm = response.json()  # Returns dict
-
+        elif self._icm is None:
+            self._icm = self.cachito_session.get_image_content_manifest(
+                self.remote_sources[0]["request_id"]
```
rcerven, line 137: this should already pass all request_ids 

since we are allowing only 1 request from pre_resolve_remote_source it won't be issue

atomic_reactor/plugins/pre_resolve_remote_source.py
```diff
@@ -94,26 +94,30 @@ def run(self):
...
+            'annotations': {'remote_source_url': remote_sources[0]['url']},
             # JSON representation of the remote source request
             'remote_source_json': remote_source_json,
             # Local path to the remote source archive
             'remote_source_path': dest_path,
         }
 
-    def set_worker_params(self, source_request, remote_source_url, remote_source_conf_url,
-                          remote_source_icm_url):
+    def set_worker_params(self, source_request, remote_sources):
```
rcerven, line 120: we don't need source_request anymore to get 'id'

as that id is also already in remote_sources

mkosiarc: updated

atomic_reactor/plugins/pre_add_image_content_manifest.py
mkosiarc: updated

atomic_reactor/utils/cachito.py
mkosiarc: updated

rcerven (approved):

lkolacek (approved):

test.sh
```diff
@@ -27,8 +27,8 @@ fi
 function setup_osbs() {
   # Optionally specify repo and branch for osbs-client to test changes
   # which depend on osbs-client patches not yet available in upstream master
-  OSBS_CLIENT_REPO=${OSBS_CLIENT_REPO:-https://github.com/containerbuildsystem/osbs-client}
-  OSBS_CLIENT_BRANCH=${OSBS_CLIENT_BRANCH:-master}
+  OSBS_CLIENT_REPO=${OSBS_CLIENT_REPO:-https://github.com/mkosiarc/osbs-client}
```
MartinBasti, line 30: :eyes:

mkosiarc: otherwise tests would fail, this will be removed before merging.

MartinBasti: just making note :)

mkosiarc:
rn updated : Update plugins to consume remote_sources parameter

PR #1613, opened 2021-04-12
pre_fetch_sources: update to download remote-source-file archives

hjmodi:
Download remote-source-file archives to build middleware sources
images for middleware binary builds

MMENG-1277

### DIFF
--- a/atomic_reactor/plugins/pre_fetch_sources.py
+++ b/atomic_reactor/plugins/pre_fetch_sources.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2019 Red Hat, Inc
+Copyright (c) 2019, 2021 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
@@ -82,7 +82,7 @@ class FetchSourcesPlugin(PreBuildPlugin):
         insecure = koji_config.get('insecure_download', False)
         urls = self.get_srpm_urls(signing_intent['keys'], insecure=insecure)
         urls_remote, remote_sources_map = self.get_remote_urls()
-        urls_maven = self.get_kojifile_source_urls()
+        urls_maven = self.get_kojifile_source_urls() + self.get_remote_file_urls()
 
         if not any([urls, urls_remote, urls_maven]):
             msg = "No srpms or remote sources or maven sources found for source" \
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -301,6 +301,35 @@ def _get_kojifile_source_urls_helper(self, koji_build):
...
+        remote_source_files_path = self.pathinfo.typedir(koji_build, btype='remote-source-file')
+        sources = []
+
+        for archive in archives:
+            if archive['type_name'] == 'tar':
+                # download each remote-source-file archive into it's own subdirectory
+                #  with the same name as the archive
+                source = {'url': os.path.join(remote_source_files_path,
+                                              archive['filename']),
+                          'subdir': archive['filename'].split('.')[0],
```
rcerven, line 325: I am not sure which extensions that file has but  I guess there can be dots in the filename as well
so maybe we could do rsplit('.', 1) ?

hjmodi: hmm, I think we'll have to use rsplit('.', 2) in that case since we're getting a lot of .tar.gz files.

hjmodi: updated.

rcerven (approved):

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -344,6 +373,26 @@ def get_kojifile_source_urls(self):
...
+        :return: list, dicts with URL pointing to remote source files
+        """
+        sources = []
+
+        source = self._get_remote_file_urls_helper(self.koji_build)
+        sources.extend(source)
+
+        koji_build = self.koji_build
+
+        while 'parent_build_id' in koji_build['extra']['image']:
```
lkolacek, line 388: Will this result in endless loop, because condition will be always True?

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -344,6 +373,26 @@ def get_kojifile_source_urls(self):
...
+            koji_build = self.session.getBuild(koji_build['extra']['image']['parent_build_id'],
```
lkolacek, line 389: Renaming `koji_build` to `parent_koji_build` to not to overwrite different object?

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: base image doesn't have 'parent_build_id' because it is generated from koji image builder

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: we aren't overwriting different object but we still need the first build for while loop which is from : 
koji_build = self.koji_build

atomic_reactor/plugins/pre_fetch_sources.py
lkolacek: got it now

atomic_reactor/plugins/pre_fetch_sources.py
lkolacek: got it now

lkolacek (approved):

MartinBasti (approved):

hjmodi:
release notes updated

PR #1615, opened 2021-04-14
convert utils/pnc.py to a class and initialize only when needed

hjmodi:
MMENG-1276

### DIFF
--- a/atomic_reactor/plugins/pre_fetch_sources.py
+++ b/atomic_reactor/plugins/pre_fetch_sources.py
@@ -27,7 +27,7 @@ from atomic_reactor.source import GitSource
 from atomic_reactor.util import get_retrying_requests_session
 from atomic_reactor.download import download_url
 from atomic_reactor.metadata import label_map
-from atomic_reactor.utils import pnc
+from atomic_reactor.utils.pnc import PNCUtil
 
 
 @label_map('sources_for_koji_build_id')
@@ -69,6 +69,17 @@ class FetchSourcesPlugin(PreBuildPlugin):
         self.signing_intent = signing_intent
         self.session = get_koji_session(self.workflow)
         self.pathinfo = get_koji_path_info(self.workflow)
+        self._pnc_util = None
+
+    @property
+    def pnc_util(self):
+        if not self._pnc_util:
+            try:
+                pnc_map = get_pnc(self.workflow)
+            except KeyError:
+                raise RuntimeError('No PNC configuration found in reactor config map') from KeyError
+            self._pnc_util = PNCUtil(pnc_map)
+        return self._pnc_util
 
     def run(self):
         """
@@ -258,10 +269,6 @@ class FetchSourcesPlugin(PreBuildPlugin):
         self.log.debug('images: %s', images)
 
         sources = []
-        pnc_map = get_pnc(self.workflow)
-        # using urljoin here causes the API path in the base_api_url to be removed
-        api_request_url = pnc_map['base_api_url'] + '/' + pnc_map['get_scm_archive_path']
-        req_session = get_retrying_requests_session()
 
         kojifile_build_ids = {kojifile['build_id'] for image in images
                               for kojifile in self.session.listArchives(imageID=image['id'],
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -271,10 +267,10 @@ def _get_kojifile_source_urls_helper(self, koji_build):
             source_build = self.session.getBuild(build_id, strict=True)
             if source_build['owner_name'] == PNC_SYSTEM_USER:
                 pnc_build_id = source_build['extra']['external_build_id']
-                url, dest_filename = pnc.get_scm_archive_from_build_id(
-                                                                    api_request_url=api_request_url,
-                                                                    build_id=pnc_build_id,
-                                                                    session=req_session)
+                if not self.pnc_util:
+                    self.pnc_util = PNCUtil(self.workflow)
```
MartinBasti, line 271: Would be safer to make `pnc_util` lazy property?

MartinBasti:
When there is a PNC build but PNC config is empty, how will it be reported to user? KeyError traceback?

atomic_reactor/plugins/pre_fetch_sources.py
hjmodi: Updated to make it a lazy_property

rcerven:
> When there is a PNC build but PNC config is empty, how will it be reported to user? KeyError traceback?

we should allow either fallback for get_pnc method, and then when it is used and not configured it should log.error and raise RuntimeError or some else

or just expect KeyError in get_pnc call and act upon it

hjmodi:
> When there is a PNC build but PNC config is empty, how will it be reported to user? KeyError traceback?

yes, it's pretty much guaranteed though to have the build id.

hjmodi:
> > When there is a PNC build but PNC config is empty, how will it be reported to user? KeyError traceback?
> 
> we should allow either fallback for get_pnc method, and then when it is used and not configured it should log.error and raise RuntimeError or some else
> 
> or just expect KeyError in get_pnc call and act upon it

it shouldn't be the case after the release. If it is though, I think it'll through a KeyError.

MartinBasti:
> > > When there is a PNC build but PNC config is empty, how will it be reported to user? KeyError traceback?
> > 
> > 
> > we should allow either fallback for get_pnc method, and then when it is used and not configured it should log.error and raise RuntimeError or some else
> > or just expect KeyError in get_pnc call and act upon it
> 
> it shouldn't be the case after the release. If it is though, I think it'll throw a KeyError.

It can be issue anytime, it can be removed from config accidentally and I'd like to see helfpul error message instead of KeyError traceback

hjmodi:
> > > > When there is a PNC build but PNC config is empty, how will it be reported to user? KeyError traceback?
> > > 
> > > 
> > > we should allow either fallback for get_pnc method, and then when it is used and not configured it should log.error and raise RuntimeError or some else
> > > or just expect KeyError in get_pnc call and act upon it
> > 
> > 
> > it shouldn't be the case after the release. If it is though, I think it'll throw a KeyError.
> 
> It can be issue anytime, it can be removed from config accidentally and I'd like to see helfpul error message instead of KeyError traceback

Added changes to cover this case.

atomic_reactor/plugins/pre_reactor_config.py
```diff
@@ -66,7 +66,11 @@ def get_koji(workflow):
 
 
 def get_pnc(workflow):
-    return get_value(workflow, 'pnc', NO_FALLBACK)
+    try:
+        value = get_value(workflow, 'pnc', NO_FALLBACK)
+        return value
+    except KeyError:
```
MartinBasti, line 72: I think this should be in `PNCUtil`. Or in that lazy property (in such case, PNCUtil can receive it as param instead of workflow)

hjmodi: updated.

MartinBasti (approved):

rcerven:
all looks good,
just missing coverage for that raising error

rcerven (approved):

hjmodi:
no rn updates needed.

PR #1620, opened 2021-05-03
Download archives for remote_sources

rcerven:
* CLOUDBLD-4450

### DIFF
--- a/atomic_reactor/plugins/pre_download_remote_source.py
+++ b/atomic_reactor/plugins/pre_download_remote_source.py
@@ -13,12 +13,13 @@ import os
 import tarfile
 from shlex import quote
 
-from atomic_reactor.constants import REMOTE_SOURCE_DIR, CACHITO_ENV_FILENAME
+from atomic_reactor.constants import REMOTE_SOURCE_DIR, CACHITO_ENV_FILENAME, CACHITO_ENV_ARG_ALIAS
 from atomic_reactor.download import download_url
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.plugins.pre_reactor_config import get_cachito
 from atomic_reactor.util import get_retrying_requests_session
 from atomic_reactor.utils.cachito import CFG_TYPE_B64
+from urllib.parse import urlparse
 
 
 class DownloadRemoteSourcePlugin(PreBuildPlugin):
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/pre_download_remote_source.py
```diff
@@ -84,58 +125,61 @@ def run(self):
 
         session = get_retrying_requests_session()
 
-        # Download the source code archive
+        self.new_remote_sources = self.remote_sources[0]['name'] is not None
```
mkosiarc, line 128: one alternative idea might be to check _self.workflow.source.config.remote_sources_ instead of using self.new_remote_sources.

We could possibly also issue warning/raise error based on _self.workflow.source.config.remote_sources_ if user is using remote_sources but used  COPY $REMOTE_SOURCE $REMOTE_SOURCE_DIR in the dockerfile.

mkosiarc:
Maybe the commit message could have short description as well, even though more details can be found in the jira.

atomic_reactor/plugins/pre_download_remote_source.py
```diff
@@ -42,6 +43,7 @@ def __init__(self, tasker, workflow, remote_sources=None):
         """
         super(DownloadRemoteSourcePlugin, self).__init__(tasker, workflow)
         self.remote_sources = remote_sources
+        self.new_remote_sources = None
```
MartinBasti, line 46: I don't like `new_` prefix, will we have also `new_new_new_remote_sources` ? Can we rename it to something more feature related names?

atomic_reactor/plugins/pre_download_remote_source.py
rcerven: sure I'll just use self.workflow.source.config.remote_sources

rcerven: we don't have to check dockerfile for that, since those variables won't be defined anyway so it won't work anyway

atomic_reactor/plugins/pre_download_remote_source.py
rcerven: fixed

mkosiarc (approved):

MartinBasti (approved):

rcerven:
rn updated: Download archives for remote_sources

PR #1634, opened 2021-06-16
If all clusters for required platform are disabled, fail the build

rcerven:
When build requires platform which is defined in config map, but
all clusters for that platform are disabled fail the build.
This changes current behavior, which was just silently skipping
such platform.

* CLOUDBLD-6076

### DIFF
--- a/atomic_reactor/plugins/pre_check_and_set_platforms.py
+++ b/atomic_reactor/plugins/pre_check_and_set_platforms.py
@@ -68,19 +68,28 @@ class CheckAndSetPlatformsPlugin(PreBuildPlugin):
                     return set(override_platforms)
         else:
             platforms = get_orchestrator_platforms(self.workflow)
-            self.log.info("No koji platforms. User specified platforms are %s", sorted(platforms))
+            user_platforms = sorted(platforms) if platforms else None
+            self.log.info("No koji platforms. User specified platforms are %s", user_platforms)
 
         if not platforms:
             raise RuntimeError("Cannot determine platforms; no koji target or platform list")
 
         # Filter platforms based on clusters
         enabled_platforms = []
+        defined_but_disabled = []
         for p in platforms:
             if self.reactor_config.get_enabled_clusters_for_platform(p):
                 enabled_platforms.append(p)
+            elif self.reactor_config.cluster_defined_for_platform(p):
+                defined_but_disabled.append(p)
             else:
-                self.log.warning(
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/pre_check_and_set_platforms.py
```diff
@@ -68,19 +68,29 @@ def run(self):
...
         # Filter platforms based on clusters
         enabled_platforms = []
+        defined_but_disabled = []
         for p in platforms:
             if self.reactor_config.get_enabled_clusters_for_platform(p):
                 enabled_platforms.append(p)
             else:
-                self.log.warning(
-                    "No cluster found for platform '%s' in reactor config map, skipping", p)
+                if self.reactor_config.cluster_defined_for_platform(p):
```
mkosiarc, line 84: this could be changed to elif statement, and therefore the above else could be removed.

tests/plugins/test_check_and_set_platforms.py
```diff
@@ -341,3 +338,63 @@ def test_platforms_from_cluster_config(tmpdir, user_params,
...
+    }])
+
+    if fails:
+        with pytest.raises(Exception) as e:
+            runner.run()
+
+        if fails == 'no_platforms':
+            msg = 'No platforms to build for'
+        elif fails == 'disabled':
+            msg = 'Platforms specified in config map, but have all clusters disabled'
```
mkosiarc, line 381: Minor thing, the error message could be incorporated directly into the parameters, and then it would not have to be assigned in the code based on fails parameter. and we would have only fails_with_error_msg or something like that.

rcerven: yeah I know, but that message is quite long, so I didn't want to add it to parameters

atomic_reactor/plugins/pre_check_and_set_platforms.py
rcerven: ok

rcerven: fixed

tests/plugins/test_check_and_set_platforms.py
mkosiarc: I feel like long message is better than additional conditionals but it's not important.

ssalatsk (approved):

MartinBasti (approved):

rcerven:
rn updated: If all clusters for required platform are disabled, fail the build

PR #1637, opened 2021-06-21
Add check for Dockerfile label version in 'check_user_settings' plugin

ssalatsk:
…lugin

*CLOUDBLD-3374

A given Dockerfile can have label 'version', which in turn
may have unexpected characters, e.g. '/' or '!'.
Make sure that the given value can only contain
alphanumeric characters, underscores, and dots between keywords.

Signed-off-by: Serhii Salatskyi <ssalatsk@redhat.com>

### DIFF
--- a/atomic_reactor/plugins/pre_check_user_settings.py
+++ b/atomic_reactor/plugins/pre_check_user_settings.py
@@ -10,6 +10,7 @@ from atomic_reactor.constants import (PLUGIN_CHECK_USER_SETTINGS, CONTAINER_DOCK
                                       CONTAINER_IMAGEBUILDER_BUILD_METHOD)
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.util import (
+    df_parser,
     has_operator_appregistry_manifest,
     has_operator_bundle_manifest,
     is_isolated_build,
@@ -18,6 +19,8 @@ from atomic_reactor.util import (
     read_fetch_artifacts_pnc,
     read_fetch_artifacts_url
 )
+
+from osbs.utils import Labels
 
 
 class CheckUserSettingsPlugin(PreBuildPlugin):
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_check_user_settings.py
```diff
@@ -52,10 +57,35 @@ def dockerfile_checks(self):
...
+            "Dockerfile label version can only contain "
+            "alphanumeric characters, underscores, and dots between keywords"
+        )
+
+        parser = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
+        dockerfile_labels = parser.labels
+        labels = Labels(parser.labels)
+        component_label = labels.get_name(Labels.LABEL_TYPE_VERSION)
+
+        if component_label not in dockerfile_labels:
```
rcerven, line 81: version is required label, and build without it would fail already in osbs-client

atomic_reactor/plugins/pre_check_user_settings.py
```diff
@@ -52,10 +57,35 @@ def dockerfile_checks(self):
...
+           can only contain alphanumeric characters, underscores, and dots between keywords.
+
+           Skip the check if there is no defined version
+        """
+        msg = (
+            "Dockerfile label version can only contain "
+            "alphanumeric characters, underscores, and dots between keywords"
+        )
+
+        parser = df_parser(self.workflow.builder.df_path, workflow=self.workflow)
```
rcerven, line 76: you should also take in account flatpak, which doesn't have Dockerfile

atomic_reactor/plugins/pre_check_user_settings.py
```diff
@@ -52,10 +57,35 @@ def dockerfile_checks(self):
...
+        dockerfile_labels = parser.labels
+        labels = Labels(parser.labels)
+        component_label = labels.get_name(Labels.LABEL_TYPE_VERSION)
+
+        if component_label not in dockerfile_labels:
+            return
+
+        label_version = dockerfile_labels[component_label]
+        for part in label_version.split('.'):
```
rcerven, line 85: we should just check for character '/' in whole version label

atomic_reactor/plugins/pre_check_user_settings.py
ssalatsk: It's already prevented [here](https://github.com/containerbuildsystem/atomic-reactor/pull/1637/files#diff-f03b0cdbe34622ae6254586f69ab8d5aa2784be9e59b48e44bfb769dbd41152fL49)

atomic_reactor/plugins/pre_check_user_settings.py
ssalatsk: Fixed

atomic_reactor/plugins/pre_check_user_settings.py
rcerven: ???

if in df_path isn't Dockerfile (flatpak), parser.labels will fail with : FileNotFoundError: [Errno 2] No such file or directory: './Dockerfile'

rcerven (changes requested):

atomic_reactor/plugins/pre_check_user_settings.py
mkosiarc: @rcerven  self.label_version_check is called in function dockerfile_checks(), which returns at the start if self.flatpak is true

mkosiarc (approved):

atomic_reactor/plugins/pre_check_user_settings.py
rcerven: oh ic

rcerven (approved):

MartinBasti (approved):

rcerven (approved):

rcerven:
rn updated: Add check for Dockerfile label version in 'check_user_settings' plugin

PR #1638, opened 2021-06-30
Reword user_warning for compose inheritance

MartinBasti:
Make clear about which inheritance OSBS warns about

https://github.com/containerbuildsystem/osbs-client/issues/1027

Signed-off-by: Martin Basti <mbasti@redhat.com>

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -128,7 +128,7 @@ class ResolveComposesPlugin(PreBuildPlugin):
         """Returns boolean if composes can be inherited"""
         if not self.workflow.source.config.inherit:
             return False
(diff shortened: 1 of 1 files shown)


atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -131,7 +131,8 @@ def allow_inheritance(self):
         self.log.info("Inheritance requested in config file")
```
ktdreyer, line 131: It seems like we could update this log statement as well to mention the "config file" we're talking about

ktdreyer (approved):

atomic_reactor/plugins/pre_resolve_composes.py
MartinBasti: added

ben-alkov (approved):

PR #1649, opened 2021-08-26
Support header-only signed RPMs

chmeliik:
CLOUDBLD-7002

There are 4 RPM tags that can contain signatures [1]. Two of them are
calculated from the header+payload - SIGPGP and SIGGPG. The other two
are calculated only from the header - DSAHEADER and RSAHEADER.

Header signatures were added in rpm v4 and the header+payload ones have
slowly been getting phased out since then [2]. The default rpm behavior
is now to include only the header signature [3].

What this means for OSBS is simply that we need to start looking at the
header signatures as well, where previously we would only consider the
header+payload ones.

[1]: https://rpm-software-management.github.io/rpm/manual/tags.html
[2]: https://github.com/rpm-software-management/rpm/issues/863
[3]: https://github.com/rpm-software-management/rpm/commit/b1aeafef49721d565fcd02ce5e37b653bd96559e


chmeliik:
[RN](https://github.com/containerbuildsystem/atomic-reactor/releases/tag/untagged-b7b5240f6edacd34bb95) updated:
* Support signed RPMs that only have a header signature, no header+payload signature

PR #1654, opened 2021-09-03
convert reactor-config-map plugin into core config module

rcerven:
* CLOUDBLD-6304


atomic_reactor/config.py
```diff
@@ -0,0 +1,567 @@
...
+        cluster_configs = self._get_cluster_configuration()
+        if platform not in cluster_configs:
+            return []
+        return [conf for conf in cluster_configs[platform] if conf.enabled]
+
+    def cluster_defined_for_platform(self, platform):
+        cluster_configs = self._get_cluster_configuration()
+        return platform in cluster_configs
+
+    def get_odcs_config(self):
```
MartinBasti, line 210: Property?

atomic_reactor/config.py
```diff
@@ -0,0 +1,567 @@
...
+
+    def _get_cluster_configuration(self):
+        all_cluster_configs = {}
+        for platform, clusters in self.clusters.items():
+            cluster_configs = [ClusterConfig(priority=priority, **cluster)
+                               for priority, cluster in enumerate(clusters)]
+            all_cluster_configs[platform] = cluster_configs
+        return all_cluster_configs
+
+    def get_enabled_clusters_for_platform(self, platform):
```
MartinBasti, line 200: Property?

atomic_reactor/config.py
```diff
@@ -0,0 +1,567 @@
...
+            "krb_keytab": self.koji['auth'].get('krb_keytab_path')
+        }
+
+        use_fast_upload = self.koji.get('use_fast_upload', True)
+
+        return create_koji_session(self.koji['hub_url'], auth_info, use_fast_upload)
+
+    @property
+    def koji_path_info(self):
+        from koji import PathInfo
```
MartinBasti, line 267: Why import is here?

atomic_reactor/build.py
```diff
@@ -9,16 +9,11 @@
 Classes which implement tasks which builder has to be capable of doing.
 Logic above these classes has to set the workflow itself.
 """
-import re
```
MartinBasti, line 12: Why changes in build.py were done?

atomic_reactor/config.py
rcerven: it was like that in reactor_config_map plugin, but I can move all those imports to the top if you want

atomic_reactor/build.py
rcerven: in config I need dockerfileimages, which were in InsideBuilder (which will be removed), so I've already moved useful attributes to workflow

        self.df_dir = build_file_dir
        self._df_path = None
        self.original_df = None
        self.buildargs = {}  # --buildargs for container build
        self.dockerfile_images = DockerfileImages([])

atomic_reactor/config.py
rcerven: it needs platform arg

atomic_reactor/config.py
rcerven: sure

rcerven:
input & build plugins (except for orchestrate & source) removed

also moved from InsideBuilder to workflow

    self.df_dir = build_file_dir
    self._df_path = None
    self.original_df = None
    self.buildargs = {}  # --buildargs for container build
    self.dockerfile_images = DockerfileImages([])

as InsideBuilder will be removed

schema now requires also : "openshift", "registries", "source_registry"
as those are mandatory for osbs anyway

removed fallbacks from plugins as well, as all configuration should be read from reactor_config_map

atomic_reactor/build.py
MartinBasti: Why dockerfile images must be in config?

rcerven: they are only used there to set source registry & organization 
https://github.com/containerbuildsystem/atomic-reactor/pull/1654/files#diff-3079e0bf8bc543b512b09e347622e9a2457237692c644efe6319d113c0ef4f1eR185

atomic_reactor/config.py
```diff
@@ -0,0 +1,568 @@
...
+
+
+class Configuration(object):
+    """
+    Class to parse the atomic-reactor configuration file
+    """
+    DEFAULT_CONFIG = {ReactorConfigKeys.VERSION_KEY: 1}
+
+    def __init__(self, workflow, config_path=None, basename=REACTOR_CONFIG_BASE_NAME,
+                 env_name=REACTOR_CONFIG_ENV_NAME, raw_config=None):
```
chmeliik, line 147: Do we need all these ways to init a Config? From a quick look, it seems that only `config_path` and `raw_config` are ever used, and `raw_config` only for unit tests.

How about something like this
```python
class Configuration:
    def __init__(self, workflow, raw_config: dict):
        ...

    @classmethod
    def from_file(cls, workflow, path: str) -> Configuration:
        ...
```
(also, I'm not convinced that `workflow` should be there, see next comment)

atomic_reactor/config.py
```diff
@@ -0,0 +1,568 @@
...
+    def set_workflow_based_on_config(self):
+        """
+        Sets attributes in workflow based on config
+        """
+
+        # set source registry and organization
+        if self.workflow.dockerfile_images:
+            source_registry_docker_uri = self.source_registry['uri'].docker_uri
+            organization = self.registries_organization
+            self.workflow.dockerfile_images.set_source_registry(source_registry_docker_uri,
```
chmeliik, line 189: Does `workflow` need to be an attribute of Configuration? That seems backwards to me (configuration doesn't have a workflow, the workflow has some configuration). How about instead giving this method a `workflow` param? The `openshift_session` property would also need to be changed, probably turned into a method with a `namespace` param.

atomic_reactor/plugins/post_push_operator_manifest.py
```diff
@@ -1,172 +0,0 @@
...
-from atomic_reactor.util import (
-    is_isolated_build,
-    is_scratch_build,
-    has_operator_appregistry_manifest,
-    has_operator_bundle_manifest,
-    get_retrying_requests_session,
-)
-
-
-class PushOperatorManifestsPlugin(PostBuildPlugin):
```
chmeliik, line 32: Could the OMPS removal have its own commit (or maybe its own PR)? And do we want to remove `utils.omps` as well?

chmeliik: And docs will need to be updated too.

atomic_reactor/config.py
```diff
@@ -0,0 +1,568 @@
...
+        root_url = self.koji['root_url'].rstrip('/')
+        return PathInfo(topdir=root_url)
+
+    @property
+    def pnc(self):
+        return self._get_value(ReactorConfigKeys.PNC_KEY)
+
+    @property
+    def odcs(self):
+        return self._get_value(ReactorConfigKeys.ODCS_KEY, fallback={})
```
chmeliik, line 281: It will be confusing if some properties have fallbacks and others don't.
```python
>>> conf = Configuration(workflow, raw_config={})
>>> conf.odcs
{}
>>> conf.pnc
KeyError: 'pnc'  # why does one fail and the other does not, if neither was in the config?
```
Could this be done more consistently?

atomic_reactor/config.py
rcerven: I did leave reading from env variable be for now

raw_config is just for testing purposes

I prefer to have it as params and call all inits the same way

atomic_reactor/config.py
rcerven: workflow has configuration, which uses/changes workflow

also it is easy to use it like that

atomic_reactor/plugin.py
```diff
@@ -445,12 +445,13 @@ def __init__(self, dt, workflow, plugin_conf, *args, **kwargs):
...
-            source_method = workflow.builder.source.config.image_build_method
-            system_method = workflow.default_image_build_method
-            plugin_conf = [{'name': source_method or system_method, 'is_allowed_to_fail': False}]
+# no build plugins
+#        else:
+#            # if no buildstep_plugins configured, which is typical for worker builds,
+#            # use what the source says or the system default.
+#            source_method = workflow.source.config.image_build_method
+#            system_method = workflow.default_image_build_method
+#            plugin_conf = [{'name': source_method or system_method, 'is_allowed_to_fail': False}]
```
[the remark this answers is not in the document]
rcerven: I'll remove it

atomic_reactor/config.py
rcerven: that is based on schema default values

before we had fallback randomly used in plugins, now we have it only based on schema default values, and without forcing it

chmeliik: In the schema, these top-level properties have a `default` value:
```
group_manifests
prefer_schema1_digest
skip_koji_check_for_base_image
deep_manifest_list_inspection
fail_on_digest_mismatch
list_rpms_from_scratch
allow_multiple_remote_sources
```
A lot of properties have fallbacks and no default in the schema

atomic_reactor/plugins/post_push_operator_manifest.py
rcerven: yeah I'll remove utils.omps as well

I'll move it to separate commit

atomic_reactor/config.py
chmeliik: Easy to use but IMO harder to understand and keep clean. But I'd like some more opinions on this

atomic_reactor/config.py
rcerven: ok I'll add fallback to everything except of required keys in the schema

atomic_reactor/config.py
MartinBasti: I agree with Adam here.

atomic_reactor/config.py
rcerven: changed

atomic_reactor/plugins/post_push_operator_manifest.py
rcerven: moved

atomic_reactor/config.py
```diff
@@ -0,0 +1,568 @@
...
+    def smtp_session(self):
+        import smtplib
+        return smtplib.SMTP(self.smtp['host'])
+
+    @property
+    def cachito(self):
+        return self._get_value(ReactorConfigKeys.CACHITO_KEY, fallback={})
+
+    @property
+    def cachito_session(self):
```
MartinBasti, line 322: `*_session` methods shouldn't be part of class, you can keep them in module but it's very weird having session property as part of configuration

atomic_reactor/config.py
rcerven: workflow removed

atomic_reactor/config.py
rcerven: *_session moved outside of class as standalone functions

atomic_reactor/schemas/config.json
```diff
@@ -280,38 +230,6 @@
         "additionalProperties": false,
         "required": ["api_url", "auth", "signing_intents", "default_signing_intent"]
     },
-    "omps": {
```
chmeliik, line 283: Should this also go in the Remove OMPS commit?

atomic_reactor/config.py
```diff
@@ -0,0 +1,565 @@
...
+    """
+    DEFAULT_CONFIG = {ReactorConfigKeys.VERSION_KEY: 1}
+
+    def __init__(self, workflow, config_path=None, basename=REACTOR_CONFIG_BASE_NAME,
+                 env_name=REACTOR_CONFIG_ENV_NAME, raw_config=None):
+
+        self.reactor_config_env = env_name
+        self.reactor_config_path = None
+        if config_path and basename:
+            self.reactor_config_path = os.path.join(config_path, basename)
```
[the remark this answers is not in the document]
rcerven: they don't have to, I thought it might be useful to have it accessible

atomic_reactor/config.py
```diff
@@ -0,0 +1,565 @@
...
+        else:
+            logger.info("using default config: %s", self.DEFAULT_CONFIG)
+
+        version = self.conf[ReactorConfigKeys.VERSION_KEY]
+        if version != 1:
+            raise ValueError("version %r unknown" % version)
+
+        logger.info("reading config content %s", self.conf)
+
+        self.set_workflow_based_on_config(workflow)
```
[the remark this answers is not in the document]
rcerven: well, it should be called right after init, so why not have it in the init already

atomic_reactor/schemas/config.json
rcerven: moved there

atomic_reactor/config.py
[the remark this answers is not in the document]
chmeliik: Because then you need a worklfow object to create a Configuration object. It should be possible create a Configuration independently of workflow.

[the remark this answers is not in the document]
rcerven: ok I'll remove it from there

atomic_reactor/config.py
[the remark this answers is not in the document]
rcerven: removed from self

tests/test_config.py
```diff
@@ -0,0 +1,1470 @@
...
+    def test_set_user_params(self, tmpdir):
+        filename = os.path.join(str(tmpdir), 'config.yaml')
+        with open(filename, 'w') as fp:
+            fp.write(dedent(REQUIRED_CONFIG))
+
+        workflow = self.prepare()
+        Configuration(config_path=str(tmpdir))
+
+        for k, v in USER_PARAMS.items():
+            assert workflow.user_params[k] == v
```
chmeliik, line 1387: This doesn't test anything related to configuration, does it?

tests/test_config.py
```diff
@@ -0,0 +1,1470 @@
...
+    def test_set_source_registry(self, tmpdir, dockerfile_images, organization):
+        config = REQUIRED_CONFIG
+
+        if organization:
+            config += "\nregistries_organization: " + organization
+
+        filename = os.path.join(str(tmpdir), 'config.yaml')
+        with open(filename, 'w') as fp:
+            fp.write(dedent(config))
+        workflow = self.prepare()
```
chmeliik, line 1435: To me this looks like the only test that actually needs a `workflow` object, could it be removed from the other tests?

chmeliik (commented):
config.py LGTM, minor comments for unit tests

I only skimmed through the rest of the changes, but they did look OK :smile:

tests/test_config.py
rcerven: right moving to inner

tests/test_config.py
rcerven: removed

chmeliik:
Note on the failing coverage check: I believe most (if not all) the code that lost coverage will be removed anyway

MartinBasti (approved):

rcerven:
osbs-docs https://github.com/containerbuildsystem/osbs-docs/pull/194

rcerven:
rn updated

PR #1661, opened 2021-09-20
Implement the supporting structure for switching to the tasks model

chmeliik:
Implements the structure, but no business logic yet, of the CLI and the individual tasks.


chmeliik:
Removed some plugins as per https://github.com/containerbuildsystem/osbs-client/pull/1033

atomic_reactor/cli/parser.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def _add_common_task_args(task_parser: argparse.ArgumentParser) -> None:
+    """Add arguments common to all tasks to the task subparser."""
+    task_parser.add_argument(
+        "--build-dir", metavar="DIR", help="directory for the build input files"
+    )
+    task_parser.add_argument(
+        "--context-dir", metavar="DIR", help="shared working directory for tasks"
+    )
```
chmeliik, line 97: @MartinBasti @rcerven @lkolacek @mkosiarc @ben-alkov 

Should there be a default build_dir and context_dir? Should the arguments be required?

MartinBasti: let's make them required

chmeliik: changed

chmeliik:
PTAL, I realize some unit tests are failing but I can't be bothered with anything older than F33 :smile: 
(I *might* look into why F32 is failing at some point)

chmeliik:
> PTAL, I realize some unit tests are failing but I can't be bothered with anything older than F33 smile
> (I _might_ look into why F32 is failing at some point)

Fixed, argparse on python3.8 cannot handle a required subparser without a metavar

chmeliik:
atomic-reactor now requires python3.8 or greater

tests/tasks/test_common.py
```diff
@@ -0,0 +1,100 @@
...
+        (
+            flexmock(util)
+            .should_receive("read_yaml")
+            .with_args(userparams_str, "schemas/user_params.json")
+            .and_return(userparams)
+        )
+        params = common.TaskParams.from_cli_args({**TASK_ARGS, "user_params": userparams_str})
+        assert params.user_params == userparams
+
+    def test_user_prams_from_file(self):
```
lkolacek, line 76: typo params

lkolacek:
Nice PR! Separate commits with explanation <3

lkolacek (approved):

tests/tasks/test_common.py
chmeliik: thanks, fixed :)

atomic_reactor/tasks/common.py
```diff
@@ -0,0 +1,92 @@
...
+        args = cls._drop_known_unset_args(args)
+        params_str = args.pop("user_params", None)
+        params_file = args.pop("user_params_file", None)
+
+        if params_str:
+            user_params = util.read_yaml(params_str, "schemas/user_params.json")
+        elif params_file:
+            user_params = util.read_yaml_from_file_path(params_file, "schemas/user_params.json")
+        else:
+            user_params = {}
```
chmeliik, line 65: Now that I think about it, should be this be allowed? The schema marks a bunch of properties as [required](https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/schemas/user_params.json#L89). IMO none of those make sense anymore, but if we're going by the current schema then these `user_params` are invalid.

MartinBasti: let's disallow it, to find issues with incorrectly defined user_params caused by refactoring.

Will it make unittests harder?

chmeliik: Hopefully not, if you need a `TaskParams` instance for unit tests you can create one directly, you don't need to go through `from_cli_args`. And the unit tests for the `common` module won't be harder either, I'm mocking the read_yaml methods anyway. Only unit tests for the parser (if I make one of the user params arguments required) will be slightly uglier.

chmeliik: Or I can just raise an error here and leave the params as optional in the CLI, we may one day reach a point where --user-params can be empty

chmeliik: Did that, keeping them optional in the CLI, lmk what you think

MartinBasti: +1

chmeliik:
Removed unused logger, fixed flake8 problem and removed redundant assert

atomic_reactor/cli/parser.py
```diff
@@ -0,0 +1,132 @@
...
+        help="shared working directory for tasks",
+    )
+    task_parser.add_argument(
+        "--config-file",
+        metavar="FILE",
+        default=REACTOR_CONFIG_FULL_PATH,
+        help=f"{PROG} configuration file",
+    )
+    # Two different ways to pass user params
+    userparams_ex = task_parser.add_mutually_exclusive_group()
```
rcerven, line 111: shouldn't user params be required?

atomic_reactor/cli/parser.py
```diff
@@ -0,0 +1,132 @@
...
+
+def _add_source_args(subtask_parser: argparse.ArgumentParser) -> None:
+    """Add --source-uri and --source-ref args to a specific task."""
+    subtask_parser.add_argument(
+        "--source-uri",
+        metavar="URI",
+        required=True,
+        help="git uri of the repository containing the build inputs",
+    )
+    subtask_parser.add_argument(
```
rcerven, line 128: @MartinBasti what about making mutually exclusive ref & branch which will be required?

also later in the code IF the branch is used we will internally convert it to the commit so it isn't floating

chmeliik:
I found out just now (from #1663) that we have a separate user_params schema for source containers :O

Fixed

atomic_reactor/schemas/plugins.json
```diff
@@ -19,9 +19,11 @@
       "allOf": [
         {"$ref": "#/definitions/base_plugins_phase"},
         {
-          "properties": {
-            "is_allowed_to_fail": {"type": "boolean"},
-            "required": {"type": "boolean"}
+          "items": {
```
rcerven, line 22: @MartinBasti what about completely removing required/is_allowed_to_fail ?

rcerven (changes requested):

atomic_reactor/tasks/common.py
chmeliik: @rcerven on the topic of user params being required. Do you think we should have them as required in the CLI for now? I'm thinking we might eventually refactor them away, so I only put a check in the TaskParams class and not in the CLI parser as that was easier to unit test.

atomic_reactor/tasks/sources.py
```diff
@@ -0,0 +1,44 @@
...
+class SourceBuildTaskParams(common.TaskParams):
+    """Source build task parameters (this task only uses common parameters)."""
+
+
+class SourceBuildTask(common.Task):
+    """Source container build task."""
+
+    plugins_def = common.PluginsDef(
+        prebuild=[
+            {"name": "reactor_config"},
```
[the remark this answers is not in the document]
chmeliik: Oops, good catch! Removed

atomic_reactor/cli/parser.py
```diff
@@ -0,0 +1,132 @@
...
+    userparams_ex = task_parser.add_mutually_exclusive_group()
+    userparams_ex.add_argument(
+        "--user-params", metavar="JSON", help="JSON string with user configuration"
+    )
+    userparams_ex.add_argument(
+        "--user-params-file", metavar="FILE", help="JSON file with user configuration"
+    )
+
+
+def _add_source_args(subtask_parser: argparse.ArgumentParser) -> None:
```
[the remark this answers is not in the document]
chmeliik: I called it _add_git_source_args, does that work?

mkosiarc (approved):

atomic_reactor/cli/parser.py
MartinBasti: I don't follow what are you asking me.

In any case don't change current behavior

atomic_reactor/cli/parser.py
```diff
@@ -0,0 +1,132 @@
...
+
+def _add_git_source_args(subtask_parser: argparse.ArgumentParser) -> None:
+    """Add --source-uri and --source-ref args to a specific task."""
+    subtask_parser.add_argument(
+        "--source-uri",
+        metavar="URI",
+        required=True,
+        help="git uri of the repository containing the build inputs",
+    )
+    subtask_parser.add_argument(
```
rcerven, line 128: please make source-ref & branch required, as it is now

atomic_reactor/cli/parser.py
```diff
@@ -0,0 +1,132 @@
...
+        "--source-ref",
+        metavar="REF",
+        help="git ref of the repository containing the build inputs",
+    )
```
rcerven, line 132: also please add required args for source container

which is either sources-for-koji-build-nvr OR sources-for-koji-build-id  (ideally mutually exclusive, even though you can pass both, there is check in plugin to compare them anyway)

rcerven (changes requested):

atomic_reactor/cli/parser.py
chmeliik: Where is it currently required?

rcerven: just remove those whole _add_git_source_args

those should be args in osbs-client, in atomic-reactor we will just read it from user_params

and only those general args context/build_dir/rcm will be from args (and later maybe even rcm will be just from user_params

chmeliik: Removed, also rebased on master

rcerven (approved):

PR #1665, opened 2021-09-24
Read plugin args from userparams

chmeliik:
CLOUDBLD-6300

The argument rendering logic that was previously handled by osbs-client
is now handled by atomic-reactor plugins themselves.

### DIFF
--- a/atomic_reactor/plugins/post_koji_upload.py
+++ b/atomic_reactor/plugins/post_koji_upload.py
@@ -13,7 +13,7 @@ from tempfile import NamedTemporaryFile
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.constants import PLUGIN_KOJI_UPLOAD_PLUGIN_KEY
 from atomic_reactor.config import get_koji_session, get_openshift_session
-from atomic_reactor.util import (get_build_json, is_scratch_build)
+from atomic_reactor.util import get_build_json, is_scratch_build, map_to_user_params
 from atomic_reactor.utils.koji import get_buildroot, get_output, get_output_metadata
 from osbs.exceptions import OsbsException
 from osbs.utils import ImageName
@@ -68,8 +68,9 @@ class KojiUploadPlugin(PostBuildPlugin):
     key = PLUGIN_KOJI_UPLOAD_PLUGIN_KEY
     is_allowed_to_fail = False
 
-    def __init__(self, workflow, koji_upload_dir, blocksize=None,
-                 platform='x86_64', report_multiple_digests=False):
+    args_from_user_params = map_to_user_params("koji_upload_dir", "platform")
+
+    def __init__(self, workflow, koji_upload_dir, blocksize=None, platform='x86_64'):
         """
         constructor
 
@@ -77,14 +78,11 @@ class KojiUploadPlugin(PostBuildPlugin):
         :param koji_upload_dir: str, path to use when uploading to hub
         :param blocksize: int, blocksize to use for uploading files
         :param platform: str, platform name for this build
-        :param report_multiple_digests: bool, whether to report both schema 1
-            and schema 2 digests
         """
         super(KojiUploadPlugin, self).__init__(workflow)
 
         self.blocksize = blocksize
         self.koji_upload_dir = koji_upload_dir
-        self.report_multiple_digests = report_multiple_digests
 
         self.osbs = get_openshift_session(self.workflow.conf,
                                           self.workflow.user_params.get('namespace'))
--- a/atomic_reactor/plugins/post_tag_from_config.py
+++ b/atomic_reactor/plugins/post_tag_from_config.py
@@ -6,48 +6,51 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
 
-import os
-import re
+import functools
 
 from atomic_reactor.plugin import PostBuildPlugin
-from atomic_reactor.constants import INSPECT_CONFIG, TAG_NAME_REGEX
+from atomic_reactor.constants import INSPECT_CONFIG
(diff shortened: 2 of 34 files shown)


chmeliik:
This moves *almost* all the logic from the `render_*` methods in https://github.com/containerbuildsystem/osbs-client/blob/master/osbs/build/plugins_configuration.py to atomic-reactor.

The one plugin that I have not yet updated is tag_from_config, that one will be an even bigger change than the two that got their own commits (because the code path that uses the 'additional-tags' file is now unreachable, and has been for a while AFAICT). I'll either add it to this PR later or update it in a different PR.

chmeliik:
TODO
- [ ] Upgrade flake8 and fix newfound problems: https://github.com/containerbuildsystem/actions/pull/2 (and linked PRs)
- [ ] Deal with unittest failures caused by https://github.com/containerbuildsystem/osbs-client/pull/1036 (@rcerven will you be removing build methods from A-R too?)
- [ ] Fix epel builds failing because they use python3.6 (how?)
- [ ] Update the tag_from_config plugin

chmeliik:
So the, last WIP commit fixes the EPEL builds AND test.sh (apart from the unit tests broken by the osbs-client schema change), but it is not the right solution. If atomic-reactor is built for python3.8 and osbs-client for python3.6, then the RPMs will either fail to install or the installation will be broken (atomic-reactor will go to `/usr/lib/python3.8`, osbs-client to `/usr/lib/python3.6` -> atomic-reactor cannot import osbs-client).

The possible solutions:
- Keep building everything for python3.6. Doesn't work if we want to keep `dataclasses` (if atomic-reactor gets installed in `/usr/lib/python3.6`, it has to be executed with `python3.6`, `dataclasses` module doesn't exist)
- Build everything (atomic-reactor, osbs-client, dockerfile-parse, koji-containerbuild) for python3.8. Assumes that all the build and runtime dependencies are already packaged for python3.8
- Wait for RHEL 9, ignore failing EPEL 8 builds

chmeliik:
> So the, last WIP commit fixes the EPEL builds AND test.sh (apart from the unit tests broken by the osbs-client schema change), but it is not the right solution. If atomic-reactor is built for python3.8 and osbs-client for python3.6, then the RPMs will either fail to install or the installation will be broken (atomic-reactor will go to `/usr/lib/python3.8`, osbs-client to `/usr/lib/python3.6` -> atomic-reactor cannot import osbs-client).

Also, while the RPM build itself is fine, I'm pretty sure the final product is broken as hell. All the dependencies would have to be python38-\*, not python3-\*

MartinBasti:
So to unblock development, for now allow fail for centos and epel in CI

chmeliik:
PTAL, epel builds will fail as we know, also 2 unittests fail because of https://github.com/containerbuildsystem/osbs-client/pull/1036

MartinBasti:
Can you remove image_build_method from the test?

chmeliik:
> Can you remove image_build_method from the test?

I can, but I though @rcerven would remove build methods from atomic-reactor since he did so for osbs-client

rcerven:
> > Can you remove image_build_method from the test?
> 
> I can, but I though @rcerven would remove build methods from atomic-reactor since he did so for osbs-client

fixed in https://github.com/containerbuildsystem/atomic-reactor/pull/1668

rcerven:
why are we still passing those user_params as args? I was hoping that we could actually get rid of plugins args and just use user_params

chmeliik:
> why are we still passing those user_params as args? I was hoping that we could actually get rid of plugins args and just use user_params @MartinBasti @chmeliik

Mainly because there are 100s of unit tests that create plugin runners in ad-hoc ways, and fixing all of those would take way too long. This approach requires almost no changes in unit tests.

I would also argue that keeping plugin arguments has its advantages, e.g. you can immediately see what the plugin needs from user_params just by looking at the init signature. And you get to keep docstrings, which IMO is a pretty big deal too.

chmeliik:
Rebased on master

atomic_reactor/plugins/post_koji_upload.py
```diff
@@ -68,6 +68,8 @@ class KojiUploadPlugin(PostBuildPlugin):
     key = PLUGIN_KOJI_UPLOAD_PLUGIN_KEY
     is_allowed_to_fail = False
 
+    args_from_user_params = map_to_user_params("koji_upload_dir", "platform")
+
     def __init__(self, tasker, workflow, koji_upload_dir, blocksize=None,
                  platform='x86_64', report_multiple_digests=False):
```
rcerven, line 74: you can also remove report_multiple_digests arg since that isn't used anymore

atomic_reactor/plugins/post_tag_from_config.py
```diff
@@ -6,48 +6,51 @@
...
-    Configuration file must be named "additional-tags" and it must
-    reside in repository as a sibling of Dockerfile. Each line in file
-    is considered as a different tag to be applied. Empty lines and
-    tag names containing hyphens are ignored. Tags will be prefixed by
-    the value of Name label.
+    """Computes tags to be applied to the built image.
 
-    For example, using the following configuration file:
+    The tags are saved in the tag configuration object in the build workflow. They are later
+    applied by the tag_and_push plugin.
```
rcerven, line 21: and push_floating_tags plugin

atomic_reactor/plugins/post_koji_upload.py
chmeliik: Good catch, deleted

atomic_reactor/plugins/post_tag_from_config.py
chmeliik: Added

rcerven (approved):

MartinBasti (approved):

PR #1669, opened 2021-10-01
Minor fixes for tasks

chmeliik:
Getting all the minor fixes out of the way ahead of the final PR that will make tasks actually do stuff.

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -356,9 +356,9 @@ class DockerBuildWorkflow(object):
         """
         tmp_dir = tempfile.mkdtemp()
         if source is None:
-            self.source = DummySource(None, None, tmpdir=tmp_dir)
+            self.source = DummySource(None, None, workdir=tmp_dir)
         else:
-            self.source = get_source_instance_for(source, tmpdir=tmp_dir)
+            self.source = get_source_instance_for(source, workdir=tmp_dir)
 
         self.prebuild_plugins_conf = prebuild_plugins
         self.buildstep_plugins_conf = buildstep_plugins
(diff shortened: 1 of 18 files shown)


atomic_reactor/inner.py
```diff
@@ -664,7 +663,7 @@ def build_docker_image(self):
                 if not exception_being_handled:
                     raise ex
             finally:
-                self.source.remove_tmpdir()
+                self.source.remove_workdir()  # OSBS2 TBD: Don't remove here, remove in exit task?
```
rcerven, line 666: well this is running already after exit plugins

exit task will run only exit plugins, so I am not sure if this comment has much value
since we will have to split runners to tasks anyway

rcerven (approved):

mkosiarc (approved):

lkolacek (approved):

atomic_reactor/inner.py
chmeliik: That's true, but it runs even if there are no exit plugins.

Example:
```python
class PluginBasedTask:
    def execute(self):
        workflow = DockerBuildWorkflow(...)
        return workflow.build_docker_image()


class PreBuildTask(PluginBasedTask):
    plugins_def = PluginsDef(prebuild=[...])


class PostBuildTask(PluginBasedTask):
    plugins_def = PluginsDef(postbuild=[...])


class ExitTask(PluginBasedTask):
    plugins_def = PluginsDef(exit=[...])
```

If `PluginBasedTask.execute()` is implemented as shown above (as I'm planning to), then all three tasks would remove the source dir at the end. I assume the tasks, or at least workflow execution, will look very different in the future, but for now I think the comment is relevant.

PR #1674, opened 2021-10-07
pre_resolve_compose: add a new compose: module_resolve_tags key

rcerven:
The module_resolve_tags key specifies that modules that are incompletely
specified (name:stream, not name:stream:version:context) are looked up
in the given set of Koji tags, rather than using the latest built
version in MBS.

This is mapped to koji_modular_tags in the ODCS request, since ODCS
uses koji_modular_tags for both the existing koji_modular_tags meaning
in container.yaml, and also for this meaning. See
https://pagure.io/odcs/pull-request/496

* CLOUDBLD-4225


rcerven:
rn updated

PR #1677, opened 2021-10-08
Add tekton definitions for source containers

rcerven:
* CLOUDBLD-6302


tekton/tasks/build-source-container.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+  steps:
+    - name: build-source-container-image
+      image: $(params.OSBS_IMAGE)
+      script: >
+        atomic-reactor task source-build
+        --user-params="$(params.USER_PARAMS)"
+        --build-dir=$(workspaces.ws-build-dir.path)
+        --context-dir=$(workspaces.ws-context-dir.path)
+        --config-file=$(workspaces.ws-reactor-config-map.path)/config.yaml
```
chmeliik, line 31: In the current implementation, only the `task` subparser accepts the common arguments
```suggestion
        atomic-reactor task
        --user-params="$(params.USER_PARAMS)"
        --build-dir=$(workspaces.ws-build-dir.path)
        --context-dir=$(workspaces.ws-context-dir.path)
        --config-file=$(workspaces.ws-reactor-config-map.path)/config.yaml
         source-build
```

tekton/tasks/build-source-container.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: build-source-container
```
MartinBasti, line 4: So we don't do versioning of tasks?

rcerven: added

tekton/pipelines/source-container.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: source-container-0-1  # dot is not allowed in the name
+spec:
+  params:
+    - name: OSBS_IMAGE
```
ben-alkov, line 7: I think that we need to decide on standards as early as possible, because things we *don't* decide on will stick just as hard as things we *do* decide on.

Given that; I'd like to see UPPER_UNDER_STYLE reserved for env vars (and for use in e.g. embedded shell scripts), with everything else being lower-kebab-style. @MartinBasti, @rcerven; WDYT?

MartinBasti: I don't have strong preference, I saw pipelines with both styles for params

PR #1683, opened 2021-10-21
stop exporting buildroot metadata

hjmodi:
Since we're moving to OpenShift Pipelines
OSBS will not be a buildroot anymore, making
the buildroot metadata related to OSBS
unnecessary

* CLOUDBLD-7747


atomic_reactor/utils/koji.py
```diff
@@ -326,30 +320,10 @@ def get_buildroot(build_id, osbs, rpms):
             'version': atomic_reactor_version,
         },
         'container': {
-            'type': 'docker',
+            'type': 'none',
             'arch': os.uname()[4],
         },
-        'tools': [
```
rcerven, line 332: I think we should still keep osbs tools versions

cc @MartinBasti

PR #1689, opened 2021-10-29
create all tmpdirs inside build_dir

hjmodi:
* CLOUDBLD-7743

### DIFF
--- a/atomic_reactor/plugins/build_source_container.py
+++ b/atomic_reactor/plugins/build_source_container.py
@@ -8,7 +8,6 @@ of the BSD license. See the LICENSE file for details.
 import os
 import shutil
 import subprocess
-import tempfile
 
 from atomic_reactor.inner import BuildResult
 from atomic_reactor.constants import (PLUGIN_SOURCE_CONTAINER_KEY, EXPORTED_SQUASHED_IMAGE_NAME,
@@ -27,7 +26,7 @@ class SourceContainerPlugin(BuildStepPlugin):
     key = PLUGIN_SOURCE_CONTAINER_KEY
 
     def export_image(self, image_output_dir):
-        output_path = os.path.join(tempfile.mkdtemp(), EXPORTED_SQUASHED_IMAGE_NAME)
+        output_path = os.path.join(self.workflow.source.workdir, EXPORTED_SQUASHED_IMAGE_NAME)
 
         cmd = ['skopeo', 'copy']
         source_img = 'oci:{}'.format(image_output_dir)
@@ -88,7 +87,8 @@ class SourceContainerPlugin(BuildStepPlugin):
         if maven_source_exists and not os.listdir(maven_source_data_dir):
             self.log.warning("Maven source directory '%s' is empty", maven_source_data_dir)
 
-        image_output_dir = tempfile.mkdtemp()
+        image_output_dir = os.path.join(self.workflow.source.workdir, 'output')
+        os.makedirs(image_output_dir, exist_ok=True)
         cmd = ['bsi', '-d']
         drivers = set()
 
--- a/atomic_reactor/plugins/post_export_operator_manifests.py
+++ b/atomic_reactor/plugins/post_export_operator_manifests.py
@@ -81,6 +81,7 @@ class ExportOperatorManifestsPlugin(PostBuildPlugin):
         if not self.should_run():
             return
 
+        # OSBS2 TBD this should go in the context dir
         manifests_archive_dir = tempfile.mkdtemp()
 #        image = self.workflow.image
         # OSBS2 TBD
--- a/atomic_reactor/plugins/pre_fetch_sources.py
+++ b/atomic_reactor/plugins/pre_fetch_sources.py
@@ -7,7 +7,6 @@ of the BSD license. See the LICENSE file for details.
 """
 import os
 import shutil
-import tempfile
 
 import koji
 import tarfile
@@ -135,7 +134,8 @@ class FetchSourcesPlugin(PreBuildPlugin):
         :param download_dir: str, directory where to download content
         :return: str, paths to directory with downloaded sources
         """
-        workdir = tempfile.mkdtemp()
+        workdir = os.path.join(self.workflow.source.workdir, 'sources')
+        os.makedirs(workdir, exist_ok=True)
         dest_dir = os.path.join(workdir, download_dir)
         if not os.path.exists(dest_dir):
             os.makedirs(dest_dir)
--- a/tests/plugins/test_build_source_container.py
+++ b/tests/plugins/test_build_source_container.py
@@ -30,6 +30,7 @@ class MockSource(object):
         self.dockerfile_path = os.path.join(tmpdir, 'Dockerfile')
         self.path = tmpdir
         self.config = flexmock(image_build_method=None)
+        self.workdir = tmpdir
 
     def get_build_file_path(self):
         return self.dockerfile_path, self.path
@@ -100,11 +101,11 @@ def test_running_build(tmpdir, caplog, user_params,
         }]
     )
 
-    temp_image_output_dir = os.path.join(str(tmpdir), 'image_output_dir')
-    temp_image_export_dir = os.path.join(str(tmpdir), 'image_export_dir')
+    temp_image_output_dir = os.path.join(str(tmpdir), 'output')
+    temp_image_export_dir = str(tmpdir)
     tempfile_chain = flexmock(tempfile).should_receive("mkdtemp").and_return(temp_image_output_dir)
     tempfile_chain.and_return(temp_image_export_dir)
-    os.mkdir(temp_image_export_dir)
+    os.makedirs(temp_image_export_dir, exist_ok=True)
     os.makedirs(os.path.join(temp_image_output_dir, 'blobs', 'sha256'))
     # temp dir created by bsi
     flexmock(os).should_receive('getcwd').and_return(str(tmpdir))


rcerven (dismissed):

chmeliik:
Noticed one more: https://github.com/containerbuildsystem/atomic-reactor/blob/eb2f4efe480d458ccd663282d8f369636de871c6/atomic_reactor/plugins/post_rpmqa.py#L93

What about `NamedTemporaryFile`s? I guess those can just stay in /tmp?

rcerven:
> Noticed one more:
> 
> https://github.com/containerbuildsystem/atomic-reactor/blob/eb2f4efe480d458ccd663282d8f369636de871c6/atomic_reactor/plugins/post_rpmqa.py#L93
> 
> What about `NamedTemporaryFile`s? I guess those can just stay in /tmp?

mmm actually there is also plugins/post_rpmqa.py:        with tempfile.TemporaryDirectory() as rpmdb_dir:

I'd say all should be in that workdir

chmeliik:
Just throwing this idea out there - would it be better to put all the temporary directories and files in a "temp" subdirectory of the build_dir rather than directly in the build_dir root?

The build_dir already contains the cloned repo and I'm also planning to keep all the per-architecture build directories in it. While naming conflicts between tempfiles and per-arch build dirs are very unlikely, I would still feel calmer if I knew everything temporary is hidden away in a subdir

rcerven:
> Just throwing this idea out there - would it be better to put all the temporary directories and files in a "temp" subdirectory of the build_dir rather than directly in the build_dir root?
> 
> The build_dir already contains the cloned repo and I'm also planning to keep all the per-architecture build directories in it. While naming conflicts between tempfiles and per-arch build dirs are very unlikely, I would still feel calmer if I knew everything temporary is hidden away in a subdir

sounds good to me

ben-alkov:
@hjmodi; I see `os.path.join(workflow.source.workdir, 'temp')` a **lot** - couldn't it be stashed in the workflow somehow (e.g. `workflow.tmp_dir` or something)?

atomic_reactor/plugins/build_source_container.py
```diff
@@ -27,7 +27,10 @@ class SourceContainerPlugin(BuildStepPlugin):
     key = PLUGIN_SOURCE_CONTAINER_KEY
 
     def export_image(self, image_output_dir):
-        output_path = os.path.join(tempfile.mkdtemp(), EXPORTED_SQUASHED_IMAGE_NAME)
+        os.makedirs(os.path.join(self.workflow.source.workdir, 'temp'), exist_ok=True)
+        output_path = os.path.join(tempfile.mkdtemp(dir=os.path.join(self.workflow.source.workdir,
+                                                                     'temp')),
+                                   EXPORTED_SQUASHED_IMAGE_NAME)
```
chmeliik, line 33: I wouldn't make this a tmpdir, let's just use something like `{workdir}/{EXPORTED_SQUASHED_IMAGE_NAME}`. The output is not temporary, it is the most important output in the source container pipeline (the built image)

atomic_reactor/plugins/build_source_container.py
```diff
@@ -88,7 +91,8 @@ def run(self):
         if maven_source_exists and not os.listdir(maven_source_data_dir):
             self.log.warning("Maven source directory '%s' is empty", maven_source_data_dir)
 
-        image_output_dir = tempfile.mkdtemp()
+        os.makedirs(os.path.join(self.workflow.source.workdir, 'temp'), exist_ok=True)
+        image_output_dir = tempfile.mkdtemp(dir=os.path.join(self.workflow.source.workdir, 'temp'))
```
chmeliik, line 95: Same here, I would make this `{workdir}/output/`

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -81,7 +81,9 @@ def run(self):
         if not self.should_run():
             return
 
-        manifests_archive_dir = tempfile.mkdtemp()
+        os.makedirs(os.path.join(self.workflow.source.workdir, 'temp'), exist_ok=True)
+        manifests_archive_dir = tempfile.mkdtemp(dir=os.path.join(self.workflow.source.workdir,
+                                                                  'temp'))
```
chmeliik, line 86: Not needed for now, you could leave an OSBS2 TBD comment that this should go in the context dir

atomic_reactor/plugins/post_rpmqa.py
```diff
@@ -90,7 +90,9 @@ def gather_output(self):
             rpmdb_archive.flush()
             tar_archive = tarfile.TarFile(rpmdb_archive.name)
 
-        with tempfile.TemporaryDirectory() as rpmdb_dir:
+        os.makedirs(os.path.join(self.workflow.source.workdir, 'temp'), exist_ok=True)
+        with tempfile.TemporaryDirectory(dir=os.path.join(self.workflow.source.workdir, 'temp')) \
+                as rpmdb_dir:
```
chmeliik, line 95: Not needed, should stay in /tmp

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -135,7 +135,8 @@ def download_sources(self, sources, insecure=False, download_dir=SRPMS_DOWNLOAD_
         :param download_dir: str, directory where to download content
         :return: str, paths to directory with downloaded sources
         """
-        workdir = tempfile.mkdtemp()
+        os.makedirs(os.path.join(self.workflow.source.workdir, 'temp'), exist_ok=True)
+        workdir = tempfile.mkdtemp(dir=os.path.join(self.workflow.source.workdir, 'temp'))
```
chmeliik, line 139: how about `{workdir}/inputs/` or `{workdir}/sources/` instead?

atomic_reactor/plugins/pre_resolve_remote_source.py
```diff
@@ -222,9 +222,10 @@ def process_request(self, source_request, name, remote_sources_workers_params,
             tarball_filename = REMOTE_SOURCE_TARBALL_FILENAME
             json_filename = REMOTE_SOURCE_JSON_FILENAME
 
+        os.makedirs(os.path.join(self.workflow.source.workdir, 'temp'), exist_ok=True)
         tarball_dest_path = self.cachito_session.download_sources(
             source_request,
-            dest_dir=self.workflow.source.workdir,
+            dest_dir=os.path.join(self.workflow.source.workdir, 'temp'),
```
chmeliik, line 228: no change needed

atomic_reactor/utils/koji.py
```diff
@@ -363,7 +363,8 @@ def get_source_tarballs_output(workflow):
 def get_remote_sources_json_output(workflow):
     plugin_results = workflow.prebuild_results.get(PLUGIN_RESOLVE_REMOTE_SOURCE) or {}
     output_json_files = []
-    tmpdir = tempfile.mkdtemp()
+    os.makedirs(os.path.join(workflow.source.workdir, 'temp'), exist_ok=True)
+    tmpdir = tempfile.mkdtemp(dir=os.path.join(workflow.source.workdir, 'temp'))
```
chmeliik, line 367: Not needed, should stay in /tmp

chmeliik (commented):
After understanding the reasoning behind creating tmpdirs inside build_dir a bit better, we have come to the following conclusion
* inputs for the build should just go directly into the build_dir
* temp files that are only used locally by a single plugin can stay in /tmp
* files that are not build inputs and need to be shared between tasks should go somewhere in the context dir
  * not yet, the context sharing needs more design

Left comments accordingly

atomic_reactor/plugins/build_source_container.py
hjmodi: updated.

atomic_reactor/plugins/post_export_operator_manifests.py
hjmodi: reverted and added comment

atomic_reactor/plugins/build_source_container.py
hjmodi: updated.

atomic_reactor/plugins/post_rpmqa.py
hjmodi: reverted.

atomic_reactor/plugins/pre_fetch_sources.py
hjmodi: updated to `{workdir}/sources/`

atomic_reactor/plugins/pre_resolve_remote_source.py
hjmodi: reverted.

atomic_reactor/utils/koji.py
hjmodi: reverted.

hjmodi:
> @hjmodi; I see `os.path.join(workflow.source.workdir, 'temp')` a **lot** - couldn't it be stashed in the workflow somehow (e.g. `workflow.tmp_dir` or something)?

we're using different names now :D

rcerven (approved):

atomic_reactor/plugins/build_source_container.py
```diff
@@ -88,7 +87,8 @@ def run(self):
         if maven_source_exists and not os.listdir(maven_source_data_dir):
             self.log.warning("Maven source directory '%s' is empty", maven_source_data_dir)
 
-        image_output_dir = tempfile.mkdtemp()
+        os.makedirs(os.path.join(self.workflow.source.workdir, 'output'), exist_ok=True)
+        image_output_dir = os.path.join(self.workflow.source.workdir, 'output')
```
mkosiarc, line 91: Could the order of these 2 be replaced? so we would have:
```
image_output_dir = os.path.join(self.workflow.source.workdir, 'output')
os.makedirs(image_output_dir, exist_ok=True)
```

hjmodi: updated.

mkosiarc (approved):

chmeliik (approved):

PR #1699, opened 2021-11-05
Add a few type annotations

tkdchen:

### DIFF
--- a/atomic_reactor/plugins/pre_pull_base_image.py
+++ b/atomic_reactor/plugins/pre_pull_base_image.py
@@ -13,15 +13,17 @@ trigger instead of what is in the Dockerfile.
 Tag each image to a unique name (the build name plus a nonce) to be used during
 this build so that it isn't removed by other builds doing clean-up.
 """
+from io import BytesIO
+from typing import Optional, Union
 
 import docker
+import requests
 
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.util import (get_platforms, base_image_is_custom,
                                  get_checksums, get_manifest_media_type,
                                  RegistrySession, RegistryClient, map_to_user_params)
 from atomic_reactor.utils import imageutil
-from io import BytesIO
 from requests.exceptions import HTTPError, RetryError, Timeout
 from osbs.utils import ImageName
 
@@ -106,7 +108,7 @@ class PullBaseImagePlugin(PreBuildPlugin):
             raise RuntimeError('Error when extracting parent images manifest digests: {}'
                                .format(digest_fetching_exceptions))
 
-    def _get_image_with_digest(self, image):
+    def _get_image_with_digest(self, image: ImageName) -> Optional[ImageName]:
         image_str = image.to_str()
         try:
             image_metadata = self.workflow.parent_images_digests[image_str]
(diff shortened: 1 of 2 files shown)


atomic_reactor/util.py
```diff
@@ -5,7 +5,7 @@
 This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
-
```
MartinBasti, line 8: Newline should stay, but I need little bit of explanation why `_hashlib` ?

tkdchen: I wanted to describe the type of return value from hashlib methods like `hashlib.md5()` or `hashlib.sha256()`. `type(returned object)` returns `_hashlib.HASH`.

lkolacek (approved):

atomic_reactor/util.py
tkdchen: The newline is back.

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -176,7 +178,7 @@ def _ensure_image_registry(self, image):
             raise RuntimeError("Shouldn't happen, images should have already "
                                "registry set in dockerfile_images")
 
-    def _pull_and_tag_image(self, image, nonce):
+    def _pull_and_tag_image(self, image: ImageName, nonce: str) -> Optional[ImageName]:
```
chmeliik, line 181: I don't think the return value is Optional, the only `return` returns an ImageName and if the return doesn't happen then you get a RuntimeError

atomic_reactor/util.py
```diff
@@ -793,7 +798,7 @@ def get_manifest_digests(self,
 
         return ManifestDigest(**digests)
 
-    def get_manifest_list(self, image):
+    def get_manifest_list(self, image: ImageName) -> requests.Response:
```
chmeliik, line 801: The return value is Optional I think

atomic_reactor/util.py
```diff
@@ -1671,41 +1680,41 @@ def base_image_key(self):
...
-    def organization(self):
+    def organization(self) -> Optional[str]:
         return self._organization
 
-    def __len__(self):
+    def __len__(self) -> int:
         return len(self._pullable_parents)
 
-    def __setitem__(self, key, item):
+    def __setitem__(self, key: Union[str, ImageName], item: str):
```
chmeliik, line 1717: Technically `item` can be an ImageName too, not sure if it's used that way though (`ImageName.parse()` works on ImageName objects too)

atomic_reactor/plugins/pre_pull_base_image.py
```diff
@@ -124,7 +126,7 @@ def _get_image_with_digest(self, image):
         new_image = '{}@sha256:{}'.format(image_name, digest)
         return ImageName.parse(new_image)
 
-    def _store_manifest_digest(self, image, use_original_tag):
+    def _store_manifest_digest(self, image: ImageName, use_original_tag: bool) -> None:
```
ben-alkov, line 129: Redundant? Functions return `None` by default...

tkdchen: This is really a good question. I think there should be many aspects of the consideration to answer this question. In short, explicit would be better than implicit to describe the behavior of a function. Annotating None here means this method does not return anything.

chmeliik (approved):
LGTM :+1:

ben-alkov (approved):

PR #1701, opened 2021-11-05
create stud tasks and pipeline defs

hjmodi:
* CLOUDBLD-7957

### DIFF
--- a/atomic_reactor/cli/parser.py
+++ b/atomic_reactor/cli/parser.py
@@ -57,6 +57,34 @@ def parse_args(args: Optional[Sequence[str]] = None) -> dict:
     )
     source_build.set_defaults(func=task.source_build)
 
+    binary_container_prebuild = tasks.add_parser(
+        "binary-container-prebuild",
+        help="binary container pre-build step",
+        description="Execute binary container pre-build steps.",
+    )
+    binary_container_prebuild.set_defaults(func=task.binary_container_prebuild)
+
+    binary_container_build = tasks.add_parser(
+        "binary-container-build",
+        help="build a binary container",
+        description="Build a binary container.",
+    )
+    binary_container_build.set_defaults(func=task.binary_container_build)
+
+    binary_container_postbuild = tasks.add_parser(
+        "binary-container-postbuild",
+        help="binary container post-build step",
+        description="Execute binary container post-build steps.",
+    )
+    binary_container_postbuild.set_defaults(func=task.binary_container_postbuild)
+
+    binary_container_exit = tasks.add_parser(
+        "binary-container-exit",
+        help="exit a binary container build",
+        description="Execute binary container exit steps.",
+    )
+    binary_container_exit.set_defaults(func=task.binary_container_exit)
+
     return vars(parser.parse_args(args))
 
 
--- a/atomic_reactor/tasks/binary.py
+++ b/atomic_reactor/tasks/binary.py
@@ -1,0 +1,83 @@
+"""
+Copyright (c) 2021 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+from atomic_reactor.tasks import plugin_based
+
+
+class BinaryPreBuildTask(plugin_based.PluginBasedTask):
+    """Binary container pre-build task."""
+
+    plugins_def = plugin_based.PluginsDef(
+        prebuild=[
+            {"name": "check_user_settings"},
+            {"name": "check_and_set_platforms"},
+            {"name": "flatpak_create_dockerfile"},
+            {"name": "inject_parent_image"},
+            {"name": "pull_base_image"},
+            {"name": "koji_parent"},
+            {"name": "resolve_composes"},
+            {"name": "add_filesystem"},
+            {"name": "flatpak_update_dockerfile"},
+            {"name": "bump_release"},
+            {"name": "add_labels_in_dockerfile"},
+            {"name": "resolve_remote_source"},
+            {"name": "pin_operator_digest"},
+            {"name": "change_from_in_dockerfile"},
+            {"name": "add_help"},
+            {"name": "fetch_maven_artifacts"},
+            {"name": "add_image_content_manifest"},
+            {"name": "add_dockerfile"},
+            {"name": "distgit_fetch_artefacts"},
+            {"name": "inject_yum_repos"},
+            {"name": "hide_files"},
+            {"name": "distribution_scope"},
+            {"name": "add_buildargs_in_dockerfile"},
+        ],
+    )
+
+
+class BinaryBuildTask(plugin_based.PluginBasedTask):
+    """Binary container build task."""
+
+    plugins_def = plugin_based.PluginsDef(
+        buildstep=[
+            {"name": "binary_container"},
+        ],
+    )
+
+
+class BinaryPostBuildTask(plugin_based.PluginBasedTask):
+    """Binary container post-build task."""
+
+    plugins_def = plugin_based.PluginsDef(
+        postbuild=[
+            {"name": "flatpak_create_oci"},
+            {"name": "all_rpm_packages"},
+            {"name": "tag_from_config"},
+            {"name": "export_operator_manifests"},
+            {"name": "compress"},
+            {"name": "fetch_worker_metadata"},
+            {"name": "compare_components"},
+            {"name": "group_manifests"},
+            {"name": "generate_maven_metadata"},
+        ],
+    )
+
+
+class BinaryExitTask(plugin_based.PluginBasedTask):
+    """Binary container exit-build task."""
+
+    plugins_def = plugin_based.PluginsDef(
+        exit=[
+            {"name": "verify_media", "required": False},
+            {"name": "koji_import"},
+            {"name": "koji_tag_build"},
+            {"name": "store_metadata"},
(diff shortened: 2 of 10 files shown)


atomic_reactor/cli/parser.py
```diff
@@ -57,6 +57,13 @@ def parse_args(args: Optional[Sequence[str]] = None) -> dict:
     )
     source_build.set_defaults(func=task.source_build)
 
+    binary_build = tasks.add_parser(
+        "binary-build",
+        help="build a binary container",
+        description="Build a binary container.",
+    )
+    binary_build.set_defaults(func=task.binary_build)
```
chmeliik, line 65: I think we should already split the binary build into the individual tasks. We will need 4 tasks, something like `pre-binary-build`, `binary-build`, `post-binary-build`, `exit-binary-build`. All of them should have the same arguments, just execute a different list of plugins (prebuild task only the prebuild plugins, postbuild task the postbuild plugins etc.)

hjmodi: I thought so too at first but then I assumed it's done internally in https://github.com/containerbuildsystem/atomic-reactor/pull/1701/files#diff-2a8511529ad7a2ff1fa4bdcb888a4c1a933de7e715e21b2569f5f8f4c35097e7R30

chmeliik: Those internal plugin phases are a leftover from OSBS 1, you can assume that each task just executes a flat list of plugins with no way to distinguish between `prebuild`, `buildstep`, etc.

hjmodi: split into tasks.

atomic_reactor/tasks/binary.py
```diff
@@ -0,0 +1,103 @@
...
+class BinaryBuildTaskParams(common.TaskParams):
+    """Binary build task parameters (this task only uses common parameters)."""
+
+    # Validate with the source containers schema instead
+    user_params_schema: ClassVar[str] = "schemas/user_params.json"
+
+    @property
+    def source(self) -> source.Source:
+        """Return a dummy source for a binary container build."""
+        return source.DummySource(None, None, workdir=self.build_dir)
```
chmeliik, line 27: No need to override, this can return a real Source just like the common TaskParams

atomic_reactor/tasks/binary.py
```diff
@@ -0,0 +1,103 @@
...
+from atomic_reactor.tasks import common
+from atomic_reactor.tasks import plugin_based
+
+
+@dataclass(frozen=True)
+class BinaryBuildTaskParams(common.TaskParams):
+    """Binary build task parameters (this task only uses common parameters)."""
+
+    # Validate with the source containers schema instead
+    user_params_schema: ClassVar[str] = "schemas/user_params.json"
```
chmeliik, line 22: "binary containers schema", also this is the default in the common TaskParams

atomic_reactor/tasks/binary.py
```diff
@@ -0,0 +1,103 @@
...
+            {"name": "distribution_scope"},
+            {"name": "add_buildargs_in_dockerfile"},
+            {"name": "pull_base_image"},
+            {"name": "change_from_in_dockerfile"},
+            {"name": "add_help"},
+            {"name": "fetch_maven_artifacts"},
+            {"name": "add_dockerfile"},
+            {"name": "distgit_fetch_artefacts"},
+            {"name": "inject_yum_repos"},
+        ],
```
chmeliik, line 60: @tkdchen could you double-check that the plugins were split into tasks in the right order? I think there is at least one problem, `fetch_maven_artifacts` and `resolve_remote_source` both need to run before `add_image_content_manifest`

tekton/pipelines/binary-container.yaml
```diff
@@ -0,0 +1,115 @@
...
+          workspace: ws-context-dir
+        - name: ws-registries-secret
+          workspace: ws-registries-secret
+        - name: ws-koji-secret
+          workspace: ws-koji-secret
+        - name: ws-worker-tokens-secrets
+          workspace: ws-worker-tokens-secrets
+        - name: ws-reactor-config-map
+          workspace: ws-reactor-config-map
+        - name: ws-reactor-config-map-scratch
```
rcerven, line 63: remove the scratch reactor config map

we are using just 1 reactor config, and setting it to correct one in osbs-client

hjmodi: removed it.

atomic_reactor/tasks/binary.py
hjmodi: updated. Switched to common.TaskParams.

atomic_reactor/tasks/binary.py
hjmodi: switched to common.TaskParams.

tekton/pipelines/binary-container.yaml
```diff
@@ -0,0 +1,106 @@
...
+    - name: USER_PARAMS
+      type: string
+      description: User parameters in JSON format
+
+  workspaces:
+    - name: ws-build-dir
+    - name: ws-context-dir
+    - name: ws-registries-secret
+    - name: ws-koji-secret
+    - name: ws-worker-tokens-secrets
```
chmeliik, line 19: What is `worker-tokens-secrets`?

rcerven: we should remove it, it was for ocp tokens for worker

chmeliik: +1, we might have need some secret for the remote build hosts but I wouldn't worry about those yet

hjmodi: piece of outdated design. Removed it.

tests/tasks/test_binary.py
```diff
@@ -0,0 +1,50 @@
...
+"""
+import pytest
+
+from atomic_reactor import source
+from atomic_reactor.tasks import common
+
+GIT_URI = "git://example.org/namespace/repo"
+
+
+class TestBinaryBuildTaskParams:
```
chmeliik, line 16: We probably don't need any additional unit tests for the `binary` module, it doesn't implement anything, just declares plugin lists

chmeliik (commented):
LGTM apart from the plugin order. But I think we need to get that right *now*, before we start changing any plugins. It would be difficult to work on plugin changes while knowing that the order in which they will be executed is not final

MartinBasti (approved):
:rocket:

tests/tasks/test_binary.py
hjmodi: removed it

atomic_reactor/tasks/binary.py
tkdchen: @hjmodi I updated the plugin list of this Pre task. There is a slight change to the order.

tekton/pipelines/binary-container.yaml
```diff
@@ -0,0 +1,97 @@
...
+        - name: ws-koji-secret
+          workspace: ws-koji-secret
+        - name: ws-reactor-config-map
+          workspace: ws-reactor-config-map
+      params:
+        - name: OSBS_IMAGE
+          value: "$(params.OSBS_IMAGE)"
+        - name: USER_PARAMS
+          value: "$(params.USER_PARAMS)"
+    - name: exit-binary-container-build
```
tkdchen, line 79: I think we need `runAfter` to ensure the order of the tasks:

1. prebuild-binary-container-0-1
2. build-binary-container-0-1
3. postbuild-binary-container-0-1
4. exit-binary-container-build-0-1

tkdchen (commented):
Regarding the name of the tekton tasks, current ones are:

- prebuild-binary-container-0-1
- build-binary-container-0-1
- postbuild-binary-container-0-1
- exit-binary-container-build-0-1

How about:

- binary-container-prebuild-0-1
- binary-container-build-0-1
- binary-container-postbuild-0-1
- binary-container-exit-0-1

This follows a format: `{build type}-{task}-{version}`.

tekton/pipelines/binary-container.yaml
hjmodi: added `runAfter`

atomic_reactor/tasks/binary.py
hjmodi: updated the order here as well

tekton/pipelines/binary-container.yaml
```diff
@@ -0,0 +1,103 @@
...
+        - name: ws-reactor-config-map
+          workspace: ws-reactor-config-map
+      params:
+        - name: OSBS_IMAGE
+          value: "$(params.OSBS_IMAGE)"
+        - name: USER_PARAMS
+          value: "$(params.USER_PARAMS)"
+    - name: binary-container-exit
+      runAfter:
+        - binary-container-postbuild
```
chmeliik, line 85: Now that I think about it, the exit task should actually run always, so it should be in the [finally](https://tekton.dev/docs/pipelines/pipelines/#adding-finally-to-the-pipeline) section

chmeliik (commented):
Nice! Just one detail about running the exit task in `finally`

tekton/pipelines/binary-container.yaml
hjmodi: moved it to finally.

tkdchen (approved):
Nice. :rocket:

chmeliik (approved):
:tada:

PR #1709, opened 2021-11-12
Implement the build_dir abstraction

tkdchen:
* CLOUDBLD-7977

### DIFF
--- a/atomic_reactor/dirs.py
+++ b/atomic_reactor/dirs.py
@@ -1,0 +1,231 @@
+"""
+Copyright (c) 2021 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+import logging
+
+from pathlib import Path
+from shutil import copyfile, copytree
+from typing import Dict, Any, List, Callable, Iterable, Optional
+
+from dockerfile_parse import DockerfileParser
+
+from atomic_reactor.constants import DOCKERFILE_FILENAME
+from atomic_reactor.source import Source
+from atomic_reactor.types import ImageInspectionData
+
+logger = logging.getLogger(__name__)
+
+
+class DockerfileNotExist(Exception):
+    """Dockerfile does not exist."""
+
+
+class BuildDir(object):
+    """Representing a directory which is specific to a platform."""
+
+    def __init__(self, path: Path, platform: str) -> None:
+        if not path.exists():
+            raise FileNotFoundError(f"Build directory {path} does not exist.")
+        self.path = path
+        self.platform = platform
+
+    @property
+    def dockerfile_path(self) -> Path:
+        """An absolute path to the dockerfile within this build directory."""
+        f = self.path / DOCKERFILE_FILENAME
+        real_path = f.resolve()
+        if real_path.parent != self.path:
+            raise DockerfileNotExist(
+                f"Dockerfile is linked from {real_path}, which is not supported."
+            )
+        return f
+
+    @property
+    def dockerfile(self) -> DockerfileParser:
+        """Return the parsed Dockerfile.
+
+        :return: the parsed Dockerfile.
+        :rtype: DockerfileParser
+        """
+        return DockerfileParser(str(self.dockerfile_path))
+
+    @staticmethod
+    def _get_env_from_inspection(data: ImageInspectionData) -> Optional[Dict[str, str]]:
+        """Get environment variables defined by ENV.
+
+        :param data: the data inspected from an image.
+        :type data: dict[str, any]
+        :return: a mapping of environment variables got from the image
+            inspection data. If no Env is found from the inspection data, None
+            will be returned.
+        :rtype: None or dict[str, str]
+        """
+        envs = data.get("Env")
+        if envs is None:
+            return None
+        if isinstance(envs, dict):
+            return envs
+        if isinstance(envs, list):
+            return dict(item.split("=", 1) for item in envs)
+
+    def dockerfile_with_parent_env(self, parent_inspect: ImageInspectionData) -> DockerfileParser:
+        """Get the parsed Dockerfile with parent information injected.
+
+        :param parent_inspect: a mapping containing the image inspection data,
+            which is used as parent image's inspection data to be injected into
+            the parsed Dockerfile.
+        :type parent_inspect: dict[str, any]
+        :return: the parsed Dockerfile
+        :rtype: DockerfileParser
+        """
+        envs = self._get_env_from_inspection(parent_inspect)
+        if envs is None:
+            logger.debug("Parent Environment not found, not applied to Dockerfile")
+        return DockerfileParser(str(self.dockerfile_path), parent_env=envs)
+
+
+FileCreationFunc = Callable[[BuildDir], Iterable[Path]]
+
+
+class RootBuildDir(object):
+    """A directory containing all artifacts for building images.
+
+    The following is a typical directory structure:
+
+    /path/to/root_build_dir
+    |-- aarch64/
+    |-- ppc64le/
+    |-- s390x/
+    +-- x86_64/
+    """
+
+    def __init__(self, path: Path, platforms: List[str]) -> None:
+        if not path.exists():
+            raise FileNotFoundError(f"Path {path} does not exist.")
+        self.path = path
+        if not platforms:
+            raise ValueError("At least one platform must be specified.")
+        self.platforms = sorted(platforms)
+
+    def copy_sources(self, source: Source) -> None:
+        """Create platform-specific build directories from source.
+
+        :param source: the original source from where to copy the content into
+            every platform-specific directory.
+        :type source: Source
+        """
+        src_path = source.path
+        for platform in self.platforms:
+            copytree(src_path, self.path / platform)
+
+    @property
+    def has_sources(self) -> bool:
+        """Check if all platform-specific directories exist.
+
+        :return: True if all platform-specific directories exist. Otherwise,
+            False is returned.
+        :rtype: bool
+        """
+        for platform in self.platforms:
+            if not self.path.joinpath(platform).exists():
+                return False
+        return True
+
+    @property
+    def any_build_dir(self) -> BuildDir:
+        """Get a platform-specific build directory.
+
+        This is typically used by code that does not care about platform. The
+        returned directory is guaranteed to be the same one between the
+        multiple calls.
+
+        :return: a source directory.
+        :rtype: BuildDir
+        """
+        platform: str = self.platforms[0]
+        return BuildDir(self.path / platform, platform)
+
+    def for_each(self, action: Callable[[BuildDir], Any]) -> Dict[str, Any]:
+        """Apply an action on every platform-specific directory.
+
+        The action callable will be applied to the platform-specific
+        directories in the order of the platforms specified when initializing
+        this RootBuildDir object. The caller should not depend on or assume the
+        order how action is applied.
+
+        The action callable might raise errors, that will cause the
+        ``for_each`` terminates immediately and the raised error is propagated
+        to the caller. As a result, the action will not be applied to the rest
+        of the platforms.
+
+        :param action: a callable object that will be applied on every
+            platform-specific directory. This callable must accept one single
+            argument in BuildDir type, and it can return data in any type.
+        :type action: Callable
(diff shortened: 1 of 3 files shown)


atomic_reactor/dirs.py
```diff
@@ -0,0 +1,227 @@
...
+        """Get a platform-specific build directory.
+
+        This is typically used by code that does not care about platform. The
+        returned directory is guaranteed to be the same one between the
+        multiple calls.
+
+        :return: a source directory.
+        :rtype: BuildDir
+        """
+        platform: Platform = sorted(self.platforms, key=lambda item: item.value)[0]
```
chmeliik, line 153: Maybe sort the platforms aleady in `__init__`? IMO that would also make the `for_each()` and `for_all_copy()` methods slightly more predictable (though the order still shouldn't be relied on)

atomic_reactor/dirs.py
```diff
@@ -0,0 +1,227 @@
...
+    def for_all_copy(self, action: Callable[[BuildDir], Iterable[Path]]) -> List[Path]:
+        """Ensure created files are present in all platform-specific directories.
+
+        :param action: a callable that creates files in a given build directory
+            and returns the files it created. The action accepts one single
+            argument in type BuildDir, and returns an iterable object that
+            yields paths of the created files.
+        :type action: callable
+        :return: the list of absolute paths of the created files in all build
+            directories.
```
chmeliik, line 192: I'd say this should return the files for just one build directory - they are guaranteed to be identical anyway

atomic_reactor/dirs.py
```diff
@@ -0,0 +1,227 @@
...
+        build_dir = self.path / first_platform.value
+        created_files = action(BuildDir(build_dir, first_platform))
+
+        # Ensure the created files' path is the absolute path, and make some
+        # checks.
+        the_new_files: List[Path] = []
+        filepath: Path
+        for filepath in created_files:
+            if not filepath.is_absolute():
+                filepath = filepath.resolve()
```
chmeliik, line 205: Maybe the path should be treated as relative to the BuildDir in this case. I think the callable should be allowed to return relative paths

atomic_reactor/dirs.py
```diff
@@ -0,0 +1,227 @@
...
+            if filepath.is_dir():
+                raise ValueError(f"{filepath} is a directory. A normal file is expected.")
```
chmeliik, line 207: I think we will have valid use cases for directories. For example the Cachito plugin, it unpacks a tarball into the build dir and I think it's OK to just return the path to the unpacked dir, the plugin shouldn't have to list all regular files

atomic_reactor/dirs.py
```diff
@@ -0,0 +1,227 @@
...
+            if not str(filepath).startswith(str(build_dir)):
+                raise ValueError(
+                    f"File must be created inside the build directory. "
+                    f"File {filepath} is not allowed."
+                )
```
chmeliik, line 212: Big +1 for this, but I think the check could be a bit more robust:
```python
build_dir in filepath.parents
# or
try:
    filepath.relative_to(build_dir)
except ValueError:
    ...
```
The current code may miss some edge cases like
```python
"/foobar/some-file".startswith("/foo")
```

atomic_reactor/dirs.py
```diff
@@ -0,0 +1,227 @@
...
+
+        copied_files: List[Path] = []
+        for platform in self.platforms[1:]:
+            for src_file in the_new_files:
+                dest_build_dir = Path(
+                    str(src_file.parent)
+                    .replace(first_platform.value, platform.value)
+                )
+                dest_build_dir.mkdir(parents=True, exist_ok=True)
+                dest_file = dest_build_dir / src_file.name
```
chmeliik, line 223:
```suggestion
                dest = self.path / platform.value / src_file.relative_to(build_dir)
                dest.parent.mkdir(parents=True, exist_ok=True)
```
Would this be simpler? I mainly wanted to avoid the string replacement, it could cause problems if the src_file path contains platform names in unexpected places

atomic_reactor/dirs.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+class DockerfileNotExist(Exception):
+    """Dockerfile does not exist."""
+
+
+class BuildDir(object):
+    """Representing a directory which is specific to a platform."""
+
+    def __init__(self, path: Path, platform: Platform) -> None:
```
chmeliik, line 30: I can't decide how I feel about using the Platform enum vs. a plain string. On the one hand, I do think it's good to limit the valid platforms, but on the other, a plain string would usualy be nicer to work with. I'm also not sure if we should hardcode the valid platforms, since technically they are configurable in the reactor config map.

atomic_reactor/dirs.py
chmeliik: > I think we lack information whether the order of passed platforms matters for the caller

In my opinion, we should never have code that relies on processing the platforms in a specific order

> For what cases of this predictability?

We will have at least two different ways to create the RootBuildDir - in the prebuild task, it will be done by the `check_and_set_platforms` plugin, in all other tasks after that, it will be done differently. If the order is up to the caller, then we would have to make sure every initialization passes the platforms in the same order, otherwise different tasks could process platforms in different order (which wouldn't matter that much, since IMO the order shouldn't matter, but it would still be nice to ensure consistency).

atomic_reactor/dirs.py
chmeliik: I meant that if we use a string for the platform name, then we have infinitely many possible values, the `Platform` enum has just 4 (and that's a good thing). But I don't know if we can really enumerate all the possible platforms in the upstream code, our instance of OSBS only handles those 4 but I'm not sure if those are the only options, even our docs mention another one: https://osbs.readthedocs.io/en/osbs_ocp3/users.html?highlight=armhfp#image-configuration

atomic_reactor/dirs.py
chmeliik: Ooh, I didn't know about [is_relative_to](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_relative_to), nice!

But it's new in 3.9, so we can't use it yet :cry:

tkdchen: Oh, I didn't notice that. I'll change it to `relative_to`. :sweat_smile:

atomic_reactor/dirs.py
```diff
@@ -0,0 +1,229 @@
...
+        created_files = action(BuildDir(build_dir, first_platform))
+
+        the_new_files: List[Path] = []
+        file_path: Path
+        for file_path in created_files:
+            if not file_path.is_absolute():
+                file_path = build_dir / file_path
+            file_path = file_path.resolve()
+            if file_path == build_dir:
+                continue
```
chmeliik, line 208: Probably shouldn't be allowed

tests/test_dirs.py
```diff
@@ -0,0 +1,276 @@
...
+from dockerfile_parse import DockerfileParser
+
+
+def test_builddir_failure_on_nonexisting_path():
+    with pytest.raises(FileNotFoundError, match="does not exist"):
+        BuildDir(Path("some_dir"), Platform.x86_64)
+
+
+def test_builddir_dockerfile_path(tmpdir):
+    dir_path = Path(tmpdir)
```
chmeliik, line 27: You can use [tmp_path](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture) instead of `tmpdir` (completely optional, I'm fine with keeping it this way)

tests/test_dirs.py
```diff
@@ -0,0 +1,276 @@
...
+])
+def test_builddir_get_parsed_dockerfile_with_parent_env(
+    inspection_data: Dict[str, Any], expected_envs, tmpdir
+):
+    dir_path = Path(tmpdir)
+    dir_path.joinpath(DOCKERFILE_FILENAME).write_text(
+        "FROM base-image\nENV HOME=$HOME", "utf-8"
+    )
+    build_dir = BuildDir(dir_path, Platform.x86_64)
+    inspection_data = ImageInspectionData(inspection_data)
```
chmeliik, line 80: Oh, I didn't realize that `NewType` requires you to wrap the value in the new type. Would a type alias be enough? https://docs.python.org/3/library/typing.html#type-aliases

chmeliik (commented):
A few more minor things, overall looks very nice!

@tkdchen if I remember corretly, you mentioned you would update a few plugins to use this, are you planning to do that in a follow-up PR?

tests/test_dirs.py
```diff
@@ -0,0 +1,276 @@
+"""
+Copyright (c) 2021 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+import os
+import tempfile
+from pathlib import Path
```
ben-alkov, line 10: 🎉

tests/test_dirs.py
tkdchen: Cool. Type alias is better in this case. :+1:

atomic_reactor/dirs.py
tkdchen: Done. Updated to use `relative_to`.

atomic_reactor/dirs.py
tkdchen: Platform enum is removed and the platform is passed as string now.

atomic_reactor/dirs.py
```diff
@@ -0,0 +1,229 @@
...
+        if envs is None:
+            logger.debug("Parent Environment not found, not applied to Dockerfile")
+        return DockerfileParser(str(self.dockerfile_path), parent_env=envs)
+
+
+class RootBuildDir(object):
+    """A directory containing all artifacts for building images.
+
+    This directory contains two kinds of directories, one is the container
+    repository cloned from the remote, and the other one is the
```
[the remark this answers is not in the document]
tkdchen: These lines are removed from the docstring.

atomic_reactor/dirs.py
tkdchen: Now, treat it as an error.

tkdchen:
@chmeliik 

> if I remember corretly, you mentioned you would update a few plugins to use this, are you planning to do that in a follow-up PR?

Yes. That is for making an example by using this new API. I chose plugin flatpak_create_dockerfile to update. More time is needed. So I think once this PR is good enough, it can be merged firstly and I'll open another PR for the plugin update.

chmeliik (commented):
LGTM :+1:

The only question left for me is if we can ignore the `.git/` dir or not when copying

atomic_reactor/dirs.py
```diff
@@ -0,0 +1,227 @@
...
+        """Create platform-specific build directories from source.
+
+        :param source: the original source from where to copy the content into
+            every platform-specific directory.
+        :type source: Source
+        """
+        src_path = source.path
+        ignore = ignore_patterns(".git")
+        for platform in self.platforms:
+            copytree(src_path, self.path / platform.value, ignore=ignore)
```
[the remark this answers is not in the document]
MartinBasti: Git should stay, people may use git ref during their builds in mutlistage builds, or weherever.

MartinBasti:
> The only question left for me is if we can ignore the `.git/` dir or not when copying

No, keep it

MartinBasti:
LGTM (I just read it quickly, so I'm not sure enough to give full ACK)

tkdchen:
Hi, this PR needs approvals for merge. :)

chmeliik (approved):
LGTM :+1:

mkosiarc (approved):

atomic_reactor/dirs.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def __init__(self, path: Path, platform: str) -> None:
+        if not path.exists():
+            raise FileNotFoundError(f"Build directory {path} does not exist.")
+        self.path = path
+        self.platform = platform
+
+    @property
+    def dockerfile_path(self) -> Path:
+        """An absolute path path to the dockerfile within this build directory."""
```
lkolacek, line 38: double path

lkolacek (approved):

atomic_reactor/dirs.py
tkdchen: @lkolacek Thank you. :)

PR #1711, opened 2021-11-18
Fetch logs from pipeline run for source container builds

mkosiarc:

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -1442,7 +1442,7 @@ class OSBSLogs(object):
 
         return metadata
 
-    def get_log_files(self, osbs, build_id):
+    def get_log_files(self, osbs, pipeline_run_name):
         """
         Build list of log files
 
@@ -1454,30 +1454,24 @@ class OSBSLogs(object):
 
         # Collect logs from server
         try:
-            logs = osbs.get_build_logs(build_id)
+            logs = osbs.get_build_logs(pipeline_run_name)
         except OsbsException as ex:
             self.log.error("unable to get build logs: %s", ex)
             return output
-        except TypeError:
-            # Older osbs-client has no get_orchestrator_build_logs
-            self.log.error("OSBS client does not support get_orchestrator_build_logs")
-            return output
 
-        platform_logs = {}
-        for entry in logs:
-            platform = entry.platform
-            if platform not in platform_logs:
-                filename = 'orchestrator' if platform is None else platform
-                platform_logs[platform] = NamedTemporaryFile(prefix="%s-%s" %
-                                                             (build_id, filename),
-                                                             suffix=".log", mode='r+b')
-            platform_logs[platform].write((entry.line + '\n').encode('utf-8'))
+        # OSBS2 TBD logs for specific arches for binary builds
+        filename = 'osbs_build'
+        logfile = NamedTemporaryFile(prefix=f"{pipeline_run_name}-{filename}",
+                                     suffix=".log", mode='wb')
 
-        for platform, logfile in platform_logs.items():
-            logfile.flush()
-            filename = 'orchestrator' if platform is None else platform
-            metadata = self.get_log_metadata(logfile.name, "%s.log" % filename)
-            output.append(Output(file=logfile, metadata=metadata))
+        # Correct log order depends on dict iteration order
+        # and on osbs.get_build_logs returning correctly ordered dict
+        for containers in logs.values():
+            for log_message in containers.values():
+                logfile.write((log_message + '\n').encode('utf-8'))
+        logfile.flush()
+        metadata = self.get_log_metadata(logfile.name, f"{filename}.log")
+        output.append(Output(file=logfile, metadata=metadata))
 
         return output
 
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -122,6 +122,7 @@ class KojiImportBase(ExitPlugin):
         self.koji_task_id = None
         self.session = None
         self.reserve_build = self.workflow.conf.koji.get('reserve_build', False)
+        self.pipeline_run_name = None
 
     def get_output(self, *args):
         # Must be implemented by subclasses
@@ -412,7 +413,7 @@ class KojiImportBase(ExitPlugin):
             return Output(file=logfile, metadata=metadata)
 
         try:
-            self.build_id = self.workflow.user_params['pipeline_run_name']
+            self.pipeline_run_name = self.workflow.user_params['pipeline_run_name']
         except KeyError:
             self.log.error("No pipeline_run_name found")
             raise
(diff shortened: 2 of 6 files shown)


atomic_reactor/util.py
```diff
@@ -1454,30 +1454,21 @@ def get_log_files(self, osbs, build_id):
...
-            filename = 'orchestrator' if platform is None else platform
-            metadata = self.get_log_metadata(logfile.name, "%s.log" % filename)
-            output.append(Output(file=logfile, metadata=metadata))
+        # OSBS2 TBD logs for specific arches for binary builds
+        filename = 'orchestrator'
+        logfile = NamedTemporaryFile(prefix=f"{pipeline_run_name}-{filename}",
+                                     suffix=".log", mode='r+b')
+        for containers in logs.values():
+            for log_message in containers.values():
+                logfile.write((log_message + '\n').encode('utf-8'))
```
chmeliik, line 1468: This depends on dict iteration order (which *is* deterministic in python >= 3.7), and also on the `get_build_logs()` method inserting taskruns and containers in the right order.

That's probably fine, but it might be good to mention that in a comment (or to change the return value of `get_build_logs()` to something more appropriate, but that would be a bigger change).

atomic_reactor/util.py
```diff
@@ -1454,30 +1454,21 @@ def get_log_files(self, osbs, build_id):
 
         # Collect logs from server
         try:
-            logs = osbs.get_build_logs(build_id)
+            logs = osbs.get_build_logs(pipeline_run_name)
```
chmeliik, line 1457: So the pipeline is getting its own logs while it's still running? Is it expected that the logs will only contain messages written up to the point when this executes?

I guess that's probably the original behavior, but god that's weird.

atomic_reactor/util.py
```diff
@@ -1454,30 +1454,21 @@ def get_log_files(self, osbs, build_id):
...
-            output.append(Output(file=logfile, metadata=metadata))
+        # OSBS2 TBD logs for specific arches for binary builds
+        filename = 'orchestrator'
+        logfile = NamedTemporaryFile(prefix=f"{pipeline_run_name}-{filename}",
+                                     suffix=".log", mode='r+b')
+        for containers in logs.values():
+            for log_message in containers.values():
+                logfile.write((log_message + '\n').encode('utf-8'))
+        logfile.flush()
+        metadata = self.get_log_metadata(logfile.name, f"{filename}.log")
```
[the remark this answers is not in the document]
mkosiarc: I kept it for now, until we decide on a name

atomic_reactor/util.py
mkosiarc: I don't mind doing the change, I leave this for more discussion from other team members.

atomic_reactor/util.py
mkosiarc: Now that you mentioned it, that does seem weird. The code currently and before fetched the logs up to the current point of execution. It does not even seem to be the last plugin to run, which could mean that some information could be lost.

MartinBasti: Yes, this is how things are working. You cannot upload logs to koji build after, so it will just contain truncated logs

atomic_reactor/util.py
```diff
@@ -1454,30 +1454,21 @@ def get_log_files(self, osbs, build_id):
...
-                                                             suffix=".log", mode='r+b')
-            platform_logs[platform].write((entry.line + '\n').encode('utf-8'))
-
-        for platform, logfile in platform_logs.items():
-            logfile.flush()
-            filename = 'orchestrator' if platform is None else platform
-            metadata = self.get_log_metadata(logfile.name, "%s.log" % filename)
-            output.append(Output(file=logfile, metadata=metadata))
+        # OSBS2 TBD logs for specific arches for binary builds
+        filename = 'orchestrator'
```
rcerven, line 1463: rename this log to 'osbs_build.log' like in https://github.com/containerbuildsystem/koji-containerbuild/blob/master/koji_containerbuild/plugins/builder_containerbuild.py#L309

atomic_reactor/util.py
rcerven: you should get_info for pipeline run and read         task_runs = pipeline_run['status']['taskRuns']

and use that order

mkosiarc: I don't think that would change anything here, as it should be already in that order https://github.com/containerbuildsystem/osbs-client/blob/master/osbs/tekton.py#L631

rcerven: it is looped in order, but then it puts it in dict, and entries from dict aren't read in the same order

mkosiarc: what do you mean that entries from dict aren't read in the same order?
It is first looped in order and put in a dict in that order, and from python 3.6+ it is guaranteed that it will be read in that order or am I missing something?

atomic_reactor/util.py
mkosiarc: done

atomic_reactor/util.py
rcerven: ah ok so the order is guaranteed by insertion order

mkosiarc: @chmeliik so as agreed that no changes will be done to get_build_logs() itself, I added a comment here.

chmeliik (approved):
LGTM :+1:

atomic_reactor/util.py
```diff
@@ -1454,30 +1454,24 @@ def get_log_files(self, osbs, build_id):
...
-
-        for platform, logfile in platform_logs.items():
-            logfile.flush()
-            filename = 'orchestrator' if platform is None else platform
-            metadata = self.get_log_metadata(logfile.name, "%s.log" % filename)
-            output.append(Output(file=logfile, metadata=metadata))
+        # OSBS2 TBD logs for specific arches for binary builds
+        filename = 'osbs_build'
+        logfile = NamedTemporaryFile(prefix=f"{pipeline_run_name}-{filename}",
+                                     suffix=".log", mode='r+b')
```
MartinBasti, line 1465: Why `r+b`  instead `wb` ?

mkosiarc: I have no idea, that part is not new :D

MartinBasti: Can we have it fixed please then?

mkosiarc: changed it to wb

mkosiarc: wait though, do you think it is not required that the file be readable?

mkosiarc: We'll see on testing I guess

MartinBasti: The only usage I was able to track is this, and you don't need read filemode because it will created it's own read descriptor
```python
        session.uploadWrapper(output.file.name, serverdir, name=name,
                              callback=upload_logger.callback, **kwargs)
```

where `output.file is logfile`

MartinBasti (approved):

MartinBasti (approved):

PR #1715, opened 2021-11-22
Create flatpak Dockerfile in platform-specific build dirs

tkdchen:


atomic_reactor/inner.py
```diff
@@ -434,7 +437,11 @@ def __init__(
         :param plugin_files: load plugins also from these files
         :param client_version: osbs-client version used to render build json
         """
+        self.build_dir = build_dir
```
chmeliik, line 440: Could this be `self._build_dir` to avoid accidents with "build_dir" vs "build_dirs"?

chmeliik (commented):
Nice!

So many unit test changes, workflow is the bane of our existence again :cry:

atomic_reactor/inner.py
tkdchen: Fixed.

atomic_reactor/inner.py
```diff
@@ -434,7 +437,11 @@ def __init__(
         :param plugin_files: load plugins also from these files
         :param client_version: osbs-client version used to render build json
         """
+        self._build_dir = build_dir
+
         self.source = source or DummySource(None, None)
+        # Will be set later to once the platforms are determined for the build.
+        self.build_dirs: Optional[RootBuildDir] = None
```
chmeliik, line 444: I'm thinking about how to do this for source containers. There is no check_and_set_platforms plugin in the source container workflow, maybe it would be better to pass an already initialized `RootBuildDir` object (or a subclass) to workflow (instead of passing `build_dir: Path`).

The problem of course, is that the prebuild task cannot do that, because it doesn't know what platforms there are yet. Maybe the `RootBuildDir` initialization has to be a 3-step process:
```python
# in the execute method (for source containers, a subclass might be better)
d = RootBuildDir(path)
# binary containers: in the check_and_set_platforms
# source containers: probably never, there are no platforms
d.set_platforms(platforms)
# binary containers: in the check_and_set_platforms plugin
# source containers: in the execute method?
d.copy_sources(source)
```
This doesn't have to  be resolved in this PR though, just throwing out ideas for the next step when we use this with source containers

atomic_reactor/plugins/pre_flatpak_create_dockerfile.py
```diff
@@ -135,19 +137,23 @@ def run(self):
...
+                                                 relative_repos_path=RELATIVE_REPOS_PATH,
+                                                 rpm_qf_args=rpm_qf_args(),
+                                                 yum_repos_dir=YUM_REPOS_DIR)
+            build_dir.dockerfile_path.write_text(content, "utf-8")
+            return [build_dir.dockerfile_path]
+
+        created_files = self.workflow.build_dirs.for_all_copy(_create_dockerfile)
+
+        dockerfile_path = created_files[0]
+        self.workflow.set_df_path(str(dockerfile_path))
```
chmeliik, line 156: I'm putting the `df_path` property on my "list of things to remove when we're done". For now it has to stay, removing it would break too many things.

Note to self: the method will still be needed, should be renamed to `set_dockerfile_images`

chmeliik (approved):
LGTM :tada:

mkosiarc (approved):

atomic_reactor/plugins/pre_flatpak_create_dockerfile.py
tkdchen: In addition, it would be also good to separate the Dockerfile validation from the set operation.

tkdchen:
Rebased on the master. I'll merge this PR once the CI passes.

tkdchen:
:tada:

PR #1718, opened 2021-11-25
Initialize build_dir for source build

tkdchen:
* CLOUDBLD-7978

### DIFF
--- a/atomic_reactor/dirs.py
+++ b/atomic_reactor/dirs.py
@@ -168,7 +168,7 @@ class RootBuildDir(object):
         platform: str = self.platforms[0]
         return BuildDir(self.path / platform, platform)
 
-    def for_each(self, action: Callable[[BuildDir], Any]) -> Dict[str, Any]:
+    def for_each_platform(self, action: Callable[[BuildDir], Any]) -> Dict[str, Any]:
         """Apply an action on every platform-specific directory.
 
         The action callable will be applied to the platform-specific
@@ -197,7 +197,7 @@ class RootBuildDir(object):
             results[platform] = action(build_dir)
         return results
 
-    def for_all_copy(self, action: FileCreationFunc) -> List[Path]:
+    def for_all_platforms_copy(self, action: FileCreationFunc) -> List[Path]:
         """Ensure created files are present in all platform-specific directories.
 
         ``for_all_copy`` accepts either absolute or relative path returned from
--- a/atomic_reactor/dirs.py
+++ b/atomic_reactor/dirs.py
@@ -24,6 +24,13 @@ class DockerfileNotExist(Exception):
     """Dockerfile does not exist."""
 
 
+class BuildDirIsNotInitialized(Exception):
+    """Build directories are not initialized."""
+
+    def __init__(self, msg: Optional[str] = None):
+        super().__init__(msg or "Build directory is not initialized yet.")
+
+
 class BuildDir(object):
     """Representing a directory which is specific to a platform."""
 
@@ -107,7 +114,7 @@ class RootBuildDir(object):
         if not path.exists():
             raise FileNotFoundError(f"Path {path} does not exist.")
         self.path = path
-        self.platforms: List[str] = []
+        self.platforms: Optional[List[str]] = None
 
     def _copy_sources(self, source: Source) -> None:
         """Create platform-specific build directories from source.
@@ -128,6 +135,8 @@ class RootBuildDir(object):
             False is returned.
         :rtype: bool
         """
+        if not self.platforms:
+            return False
         for platform in self.platforms:
             if not self.path.joinpath(platform).exists():
                 return False
@@ -138,8 +147,6 @@ class RootBuildDir(object):
 
         :param list[str] platforms: a list of platforms to build for.
         """
-        if not platforms:
-            raise ValueError("At least one platform must be specified.")
         self.platforms = sorted(platforms)
         if self.has_sources:
             return
@@ -156,6 +163,8 @@ class RootBuildDir(object):
         :return: a source directory.
         :rtype: BuildDir
         """
+        if not self.has_sources:
+            raise BuildDirIsNotInitialized()
         platform: str = self.platforms[0]
         return BuildDir(self.path / platform, platform)
 
@@ -180,6 +189,8 @@ class RootBuildDir(object):
             function which is called for that platform.
         :rtype: dict[str, any]
         """
+        if not self.has_sources:
+            raise BuildDirIsNotInitialized()
         results: Dict[str, Any] = {}
         for platform in self.platforms:
             build_dir = BuildDir(self.path / platform, platform)
@@ -202,6 +213,9 @@ class RootBuildDir(object):
         :return: the list of absolute paths of the created files.
         :rtype: list[pathlib.Path]
         """
+        if not self.has_sources:
+            raise BuildDirIsNotInitialized()
+
         first_platform: str = self.platforms[0]
         build_dir = self.path / first_platform
         created_files = action(BuildDir(build_dir, first_platform))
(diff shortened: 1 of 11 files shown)


atomic_reactor/dirs.py
```diff
@@ -135,6 +133,18 @@ def has_sources(self) -> bool:
...
+        """Initialize the root build directory with specific determined platforms.
+
+        :param list[str] platforms: a list of platforms to build for.
+        """
+        if not platforms:
+            raise ValueError("At least one platform must be specified.")
+        self.platforms = sorted(platforms)
+        if self.has_sources:
+            return
+        self.copy_sources(source)
```
chmeliik, line 146: Should copy_sources be private now? It could be easy to misuse, if you call copy_sources() before setting platforms it will do nothing

atomic_reactor/inner.py
```diff
@@ -430,18 +429,18 @@ def __init__(
...
+        :type build_dir: RootBuildDir
         :param source: where/how to get source code to put in image
         :param plugins: the plugins to be executed in this workflow
         :param user_params: user (and other) params that control various aspects of the build
         :param reactor_config_path: path to atomic-reactor configuration file
         :param plugin_files: load plugins also from these files
         :param client_version: osbs-client version used to render build json
         """
-        self._build_dir = build_dir
+        self.build_dir = build_dir
```
chmeliik, line 441: I would still call it `build_dirs`, I think that more closely matches how we want to use this object (to handle multiple build dirs)

atomic_reactor/tasks/binary.py
```diff
@@ -6,10 +6,21 @@
 of the BSD license. See the LICENSE file for details.
 """
 
+from atomic_reactor.dirs import RootBuildDir
 from atomic_reactor.tasks import plugin_based
 
 
-class BinaryPreBuildTask(plugin_based.PluginBasedTask):
+class BinaryContainerBuildTask(plugin_based.PluginBasedTask):
```
chmeliik, line 13: Maybe just `BinaryTask` or `BinaryContainerTask`? I think the current name is too similar to `BinaryBuildTask`

atomic_reactor/tasks/common.py
```diff
@@ -85,3 +86,6 @@ def __init__(self, params: TaskParams):
     @abc.abstractmethod
     def execute(self):
         """Execute this task."""
+
+    def _get_build_dir(self) -> RootBuildDir:
+        """Create a build directory object for this container build task."""
```
chmeliik, line 91: Should this be an abstract method?

chmeliik (commented):
Looks good! Just a few nitpicks

atomic_reactor/tasks/common.py
tkdchen: The clone task does not require a build dir.

chmeliik: Hmm, then maybe the method doesn't need to be in the Task class at all?

atomic_reactor/dirs.py
```diff
@@ -103,13 +103,11 @@ class RootBuildDir(object):
     +-- x86_64/
     """
 
-    def __init__(self, path: Path, platforms: List[str]) -> None:
+    def __init__(self, path: Path) -> None:
         if not path.exists():
             raise FileNotFoundError(f"Path {path} does not exist.")
         self.path = path
-        if not platforms:
```
rcerven, line 110: so if you now won't raise error here, you can just call copy_sources after init, and it won't do anything because you didn't call init_build_dirs, that doesn't seems nice

atomic_reactor/dirs.py
rcerven: I mean everything here now can be misused without platforms

rcerven (changes requested):

atomic_reactor/tasks/binary.py
rcerven: yeah this is very confusing

since this is class used for other classes, maybe even something like BinaryGeneralTask .....

atomic_reactor/dirs.py
chmeliik: That's a good point :disappointed:

atomic_reactor/dirs.py
tkdchen: @rcerven 

In this patch, a RootBuildDir object is initialized and passed to a workflow object before initializing the build directories inside with determined platforms, because for the binary container build, the initialization has to be delayed into the check_and_set_platforms plugin.

atomic_reactor/dirs.py
tkdchen: @rcerven Sorry, I can't get your point. How can everything be misused?

atomic_reactor/tasks/common.py
tkdchen: Oh, good catch! :sweat_smile:

atomic_reactor/inner.py
chmeliik: Yeah, that's a good point. I'm just afraid `workflow.build_dir.for_each()` is going to look weird now. Maybe the methods could be renamed to `for_each_platform` and `for_all_platforms_copy`? But maybe the current naming is fine, not sure

atomic_reactor/dirs.py
chmeliik: I think the potential problem is that in the prebuild task, if you call `workflow.build_dir.for_each()` or `workflow.build_dir.for_all_copy()` before the "check_and_set_platforms" plugin did the initialization, the methods will silently do nothing. But maybe if we just run the check platforms plugin at the start of the prebuild task, that wouldn't be a big problem.

Or if we kill the plugin and instead determine the platforms in `PreBuildTask.execute` :-)

atomic_reactor/inner.py
chmeliik: Alternatively, maybe plugins shouldn't access `build_dirs.path`. Maybe there should be predefined directories for everything that doesn't go in the platform-specific build dirs - like `build_dirs.source_container_inputs`, `build_dirs.output`

atomic_reactor/dirs.py
tkdchen: I see. But, based on current plugins design, if a plugin calls either the for_each or for_all_copy before check_and_set_platforms plugin run, this case should be an error and not allowed. :)

tkdchen: > Or if we kill the plugin and instead determine the platforms in PreBuildTask.execute :-)

I actually do think as well the platforms can be determined before the workflow starts the image build, and the determined platforms is an input for the image build, probably can be set into the user_params.

atomic_reactor/inner.py
tkdchen: > Yeah, that's a good point. I'm just afraid workflow.build_dir.for_each() is going to look weird now. Maybe the methods could be renamed to `for_each_platform` and `for_all_platforms_copy`? But maybe the current naming is fine, not sure

I like these two new names. 😃 I think the new name reflects the intention much straightforward than the old ones.

atomic_reactor/dirs.py
tkdchen: > , probably can be set into the user_params.

Or, just as a workflow object attribute, e.g. `workflow.platforms`.

tkdchen:
@chmeliik 

Added another new commit, PTAL.

atomic_reactor/dirs.py
```diff
@@ -24,6 +24,15 @@ class DockerfileNotExist(Exception):
     """Dockerfile does not exist."""
 
 
+class BuildDirsNotInitialized(Exception):
+    """Build directories are not initialized."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            "Build directories are not initialized inside the root build directory."
+        )
```
chmeliik, line 33: It's a bit weird that the args and kwargs are just ignored. Maybe the init shouldn't have any args? Or, instead of overriding init, avoid duplicating the code some other way (constant, helper function...)?

chmeliik (approved):
LGTM :+1:

rcerven (approved):

PR #1727, opened 2021-12-07
Registry utils

chmeliik:


atomic_reactor/util.py
```diff
@@ -912,6 +914,26 @@ def _config_by_digest(self, image: ImageName, config_digest: str) -> dict:
         blob_config = config_response.json()
         return blob_config
 
+    def _config_and_id_from_manifest_list(
+        self, image: ImageName, manifest_list: dict, arch: Optional[str]
+    ) -> Tuple[dict, str]:
```
tkdchen, line 919: How about just pass `manifest_list["manifests"]` directly?

atomic_reactor/util.py
```diff
@@ -912,6 +914,26 @@ def _config_by_digest(self, image: ImageName, config_digest: str) -> dict:
...
+        # if an architecture was specified, filter the correct architecture
+        filtered_manifests = (
+            m for m in manifest_list["manifests"]
+            if not arch or m.get("platform", {}).get("architecture") == arch
+        )
+        manifest = next(filtered_manifests, None)
```
tkdchen, line 925: Logically, if arch is None or `''`, the `filtered_manifests` might have more than one manifests. If this happens, why only handle the first one?

tkdchen (commented):
Nice.

atomic_reactor/util.py
```diff
@@ -915,15 +921,11 @@ def get_config_and_id_from_registry(self, image, digest, version='v2'):
...
+        response = query_registry(self._session, image, digest=digest, version=version)
         manifest_config = response.json()
-        config_digest = manifest_config['config']['digest']
-
-        config_response = query_registry(
-            self._session, image, digest=config_digest, version=version, is_blob=True)
 
-        blob_config = config_response.json()
+        config_digest = manifest_config['config']['digest']
+        blob_config = self._config_by_digest(image, config_digest)
```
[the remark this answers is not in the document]
chmeliik: Oh true, I didn't notice 'v1' was the default. I don't think it matters when querying the blob config, because both v1 and v2 are wrong:

v1 -> `application/vnd.docker.distribution.manifest.v1+json`
v2 -> `application/vnd.docker.distribution.manifest.v2+json`

but the content type of blob config is `application/vnd.docker.container.image.v1+json`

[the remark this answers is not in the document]
chmeliik: But yeah, we probably should eventually drop v1, because OSBS will not work correctly with it anyway

atomic_reactor/util.py
chmeliik: I would rather keep it as it is. `manifest_list["manifests"]` is a detail of the [manifest list](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list) format, currently only the `_config_and_id_from_manifest_list` method has to understand those details

atomic_reactor/util.py
chmeliik: That would mean the caller doesn't care what architecture is being inspected, so we just inspect the first image in the list. Often it is not important to inspect a specific architecture, because we expect some properties to be the same between arches.

Btw, there may be multiple manifests even if arch is not falsy, e.g. `skopeo inspect --raw docker://docker.io/library/busybox:1.34.1 | jq '.manifests[] | select(.platform.architecture == "arm")'`
```json
{
  "digest": "sha256:0252da5f2df7425dcf48afb4bc337966dfeb2d87079ea3f7fe25051d5b9e9c26",
  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
  "platform": {
    "architecture": "arm",
    "os": "linux",
    "variant": "v5"
  },
  "size": 527
}
{
  "digest": "sha256:b27cc98025245c0e746b201d5c773faff99869ae58585090182e18d7c5e8a5e2",
  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
  "platform": {
    "architecture": "arm",
    "os": "linux",
    "variant": "v6"
  },
  "size": 527
}
{
  "digest": "sha256:7c416c3c45da3176e6160c81f2ed7b94265d98ebfeac11b12eb3ce3cc289644c",
  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
  "platform": {
    "architecture": "arm",
    "os": "linux",
    "variant": "v7"
  },
  "size": 527
}
```
I don't actually know if that's OK, but I hope it is, because OSBS has no concept of architecture variants

atomic_reactor/util.py
```diff
@@ -906,6 +907,11 @@ def get_inspect_for_image(self, image):
 
         return image_inspect
 
+    def _config_by_digest(self, image: ImageName, config_digest: str) -> dict:
```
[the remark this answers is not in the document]
chmeliik: updated

atomic_reactor/util.py
```diff
@@ -844,37 +844,34 @@ def get_all_manifests(
...
         all_man_digests = self.get_all_manifests(image)
-        blob_config = None
-        config_digest = None
-        image_inspect = {}
 
-        # we have manifest list (get digest for 1st platform)
+        # we have manifest list
         if 'v2_list' in all_man_digests:
-            man_list_json = all_man_digests['v2_list'].json()
-            if man_list_json['manifests'][0]['mediaType'] != MEDIA_TYPE_DOCKER_V2_SCHEMA2:
```
rcerven, line 862: why was this removed?

atomic_reactor/util.py
```diff
@@ -891,21 +888,52 @@ def get_inspect_for_image(self, image):
...
+    def _blob_config_by_digest(self, image: ImageName, config_digest: str) -> dict:
+        config_response = query_registry(self._session, image, digest=config_digest, is_blob=True)
+        blob_config = config_response.json()
+        return blob_config
+
+    def _config_and_id_from_manifest_list(
+        self, image: ImageName, manifest_list: dict, arch: Optional[str]
+    ) -> Tuple[dict, str]:
+        # if an architecture was specified, filter the correct architecture
+        filtered_manifests = (
```
rcerven, line 921: can we have just list comprehension ? so you have access nicely to whole list if needed

atomic_reactor/util.py
```diff
@@ -891,21 +888,52 @@ def get_inspect_for_image(self, image):
...
+            m for m in manifest_list["manifests"]
+            if not arch or m.get("platform", {}).get("architecture") == arch
+        )
+        manifest = next(filtered_manifests, None)
+
+        if not manifest:
+            raise RuntimeError(
+                f"Image {image}: No manifest for {arch} architecture in manifest list"
```
rcerven, line 929: well you don't know if architecture was even provided

lkolacek: +1, None for architecture might be misleading in this error message

chmeliik: This can only happen if the manifest list is empty, but technically the [OCI spec](https://github.com/opencontainers/image-spec/blob/main/image-index.md#image-index-property-descriptions) allows that (and the docker spec doesn't say anything on the matter) so I'll add a check for this

atomic_reactor/util.py
chmeliik: Yup, fixed as part of other changes

atomic_reactor/util.py
chmeliik: fixed

chmeliik:
Added checks to ensure that
a) the manifest list is not empty
b) when an architecture is specified, it must match exactly one image in the list

chmeliik:
Rebased on master (otherwise no changes), PTAL

lkolacek (approved):

hjmodi (approved):

chmeliik:
No release notes, this change by itself is inconsequential refactoring

PR #1731, opened 2021-12-13
Fetch sources into platform-specific build dirs

tkdchen:
* CLOUDBLD-8122

### DIFF
--- a/atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
+++ b/atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
@@ -10,9 +10,12 @@ To have everything for a build in dist-git you need to fetch artefacts using 'fe
 
 This plugin should do it.
 """
-import os
+import shutil
 import subprocess
+from pathlib import Path
+from typing import List
 
+from atomic_reactor.dirs import BuildDir
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.constants import PLUGIN_DISTGIT_FETCH_KEY
 
@@ -26,12 +29,38 @@ class DistgitFetchArtefactsPlugin(PreBuildPlugin):
         constructor
 
         :param workflow: DockerBuildWorkflow instance
-        :param command: str, command to use to get artefacts (e.g. 'make sources')
-                             it is executed in cloned git repo
+        :type workflow: atomic_reactor.inner.DockerBuildWorkflow
         """
         # call parent constructor
         super(DistgitFetchArtefactsPlugin, self).__init__(workflow)
         self.command = self.workflow.conf.sources_command
+
+    def _fetch_sources(self, build_dir: BuildDir) -> List[Path]:
+        """Fetch sources files.
+
+        :param build_dir: download the sources files into this directory.
+        :type build_dir: BuildDir
+        :return: a list of downloaded file names.
+        :rtype: list[pathlib.Path]
+        """
+        # Create a dedicated directory to hold the fetched sources files, from
+        # where to generated the downloaded file list.
+        build_dir_path = build_dir.path
+        sources_outdir = build_dir_path / 'outdir'
+        sources_outdir.mkdir()
+
+        sources_cmd = self.command.split()
+        sources_cmd.append('--outdir')
+        sources_cmd.append(str(sources_outdir))
+        self.log.debug('Fetching sources: %r', sources_cmd)
+        subprocess.check_call(sources_cmd, cwd=build_dir_path)
+
+        fetched_sources = []
+        for file_name in sources_outdir.iterdir():
+            shutil.move(str(file_name), str(build_dir_path))
+            fetched_sources.append(file_name.relative_to(sources_outdir))
+        sources_outdir.rmdir()
+        return fetched_sources
 
     def run(self):
         """
@@ -41,10 +70,4 @@ class DistgitFetchArtefactsPlugin(PreBuildPlugin):
             self.log.info('no sources command configuration, skipping plugin')
             return
 
-        source_path = self.workflow.source.path
-        cur_dir = os.getcwd()
-        os.chdir(source_path)
-        try:
-            subprocess.check_call(self.command.split())
-        finally:
-            os.chdir(cur_dir)
+        self.workflow.build_dir.for_all_platforms_copy(self._fetch_sources)
--- a/atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
+++ b/atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
@@ -10,14 +10,18 @@ To have everything for a build in dist-git you need to fetch artefacts using 'fe
 
 This plugin should do it.
 """
+import os
+import re
 import shutil
 import subprocess
 from pathlib import Path
-from typing import List
+from typing import Final, List
 
 from atomic_reactor.dirs import BuildDir
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.constants import PLUGIN_DISTGIT_FETCH_KEY
+
+EXPLODED_SOURCES_FILE: Final[str] = 'source-repos'
 
 
 class DistgitFetchArtefactsPlugin(PreBuildPlugin):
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
```diff
@@ -10,13 +10,41 @@
...
+
+def read_filenames_from_sources(sources_file: Path) -> Iterable[str]:
+    """Read and return the file names listed in the sources file.
+
+    :param sources_file: the file name of the sources file.
+    :type sources_file: pathlib.Path
+    :return: a iterator of file names.
+    :rtype: iterable[str]
+    """
+    known_patterns = (re.compile(BSD_CHECKSUM_RE), re.compile(MD5_CHECKSUM_RE))
```
chmeliik, line 35: Is the sources file format documented somewhere?

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
```diff
@@ -10,13 +10,41 @@
...
+    with sources_file.open("r", encoding="utf-8") as sources:
+        for line in sources:
+            matched = False
+            for regex in known_patterns:
+                if match := regex.match(line):
+                    yield match.group(1)
+                    matched = True
+                    break
+            if not matched:
+                raise ValueError(f"Unrecognizable sources line: '{line}'.")
```
chmeliik, line 45: Btw this can be slightly simplified with for-else (completely optional)
```suggestion
            for regex in known_patterns:
                if match := regex.match(line):
                    yield match.group(1)
                    break
            else:
                raise ValueError(f"Unrecognizable sources line: '{line}'.")
```

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
```diff
@@ -10,13 +10,41 @@
...
 import subprocess
+from pathlib import Path
+from typing import Iterable, List
 
+from atomic_reactor.dirs import BuildDir
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.constants import PLUGIN_DISTGIT_FETCH_KEY
 
 
+MD5_CHECKSUM_RE = r'^[a-f0-9]+ [* ](.+)\n?$'
```
rcerven, line 23: instead of ' [* ]'

why not just ' +' ?

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
rcerven: yeah that would be useful to know

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
```diff
@@ -10,13 +10,41 @@
...
+BSD_CHECKSUM_RE = r'^[^ ]+ \((.+)\) = [a-f0-9]+\n?$'
```
rcerven, line 24: sha512 works just fine with space at the beginning

we should really make correct regex here ..

or in worst case we could just use output of rhpkg sources

OR even better would actually be to run rhpkg sources in NEWLY created empty directory, and just use file list from there !

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
rcerven: I still don't think we should be validating sources file which is up for rhpkg sources

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
ben-alkov: Do we use real, actual, [BSD checksum](https://en.wikipedia.org/wiki/BSD_checksum)[1], or is this just an odd name?

[1] If so, why?

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
tkdchen: AFAIK, no such document. I asked the rpkg maintainer, let's see if there is any news.

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
tkdchen: @rcerven 

This regex actually follows the one used in rpkg. The lines of sources are not generated by sha512sum command, instead they are written by the rpkg code which does not put a space at the beginning.

tkdchen:
Thank you guys for your review. I'll rewrite this PR according to Robert's comment.

tkdchen:
This implementation does not work with the exploded sources. Command sources does not store the generated tarball to the `--outdir` directory. It has to be handled specifically.

chmeliik:
> This implementation does not work with the exploded sources. Command sources does not store the generated tarball to the `--outdir` directory. It has to be handled specifically.

Another option might be to run the command in the build dir, not in a new directory, and compare the list of files before/after. But I don't know how rpkg handles existing files, it could be difficult to determine if any of those changed or not

tkdchen:
> Another option might be to run the command in the build dir, not in a new directory, and compare the list of files before/after. But I don't know how rpkg handles existing files, it could be difficult to determine if any of those changed or not

@chmeliik

Based on my previous experience on maintaining rpkg, it does not have a concept of handling existing files, and it is able to ignore files by modifying `.gitignore`. However, `.gitignore` is not controlled by rpkg only, it is free for maintainer to modify on-demand.

At minimum, a dist-git repo can have `sources` and a SPEC file. IIRC, `sources` is also optional.

When we start to care about the concrete downloaded file list, I originally thought to import rpkg and get the sources file list from the `SourcesFile`, that would cause many irrelevant dependent packages to be installed, which are required by the rpkg CLI commands, but we just need the metadata (the fetched sources file list) of a container dist-git repo.

The source-repos is a rhpkg-specific thing. Unfortunately, it seems there isn't a good way to get the final generated tarball files information from rhpkg. The current situation is 1) rhpkg has no API to expose the exploded sources tarballs, 2) rhpkg does not output final tarball filenames to the terminal, even with `-d -v`, the output is difficult for parse. Therefore, the rest is we have to parse the file `source-repos` by ourselves and find out the generated tarballs.

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
```diff
@@ -26,13 +29,59 @@ def __init__(self, workflow):
...
+        subprocess.check_call(sources_cmd, cwd=build_dir_path)
+
+        fetched_sources = []
+        for file_name in sources_outdir.iterdir():
+            shutil.move(str(file_name), str(build_dir_path))
+            fetched_sources.append(file_name.relative_to(sources_outdir))
+        sources_outdir.rmdir()
+
+        # Collect the exploded sources if there is
+        fetched_sources += self._collect_exploded_sources_files(build_dir_path)
```
chmeliik, line 81: It feels odd that some files will be fetched into the build_dir itself while others will be fetched into the new empty dir and then moved. I would also prefer not to parse any rpkg-specific files if possible.

Could the following approach work?
```
files_before = set(all files from build_dir)
run rpkg in build_dir, fetch everything into build_dir
files_after = set(all files from build_dir)

fetched_files = files_after - files_before
```
(this has the potential problem with existing files - e.g. if ./some/file exists, is it possible that rpkg will overwrite ./some/file with a new one?)

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
```diff
@@ -26,13 +29,59 @@ def __init__(self, workflow):
...
+        :type build_dir_path: pathlib.Path
+        """
+        source_repos = build_dir_path / 'source-repos'
+        filenames = []
+        if not source_repos.exists():
+            return filenames
+        with source_repos.open('r', encoding='utf-8') as f:
+            for line in f:
+                git_ref = line.rstrip().split()[-1]
+                filenames.append(next(build_dir_path.glob(f'*-{git_ref}.tar.gz')))
```
chmeliik, line 51: Is it safe to glob with just the git ref? We don't know what the dist-git repo may contain. Could we instead return the filename exactly as it would be constructed by rhpkg?

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
chmeliik: Discussed offline, this is probably not worth doing

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
tkdchen: @chmeliik almost rewrite this method. Add some checks on the file names.

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
```diff
@@ -11,11 +11,18 @@
...
+import shutil
 import subprocess
+from pathlib import Path
+from typing import List, final
 
+from atomic_reactor.dirs import BuildDir
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.constants import PLUGIN_DISTGIT_FETCH_KEY
 
+EXPLODED_SOURCES_FILE: final = 'source-repos'
```
chmeliik, line 24:
```suggestion
EXPLODED_SOURCES_FILE: Final = 'source-repos'
```
According to https://www.python.org/dev/peps/pep-0591/#motivation, `final` is a decorator and `Final` is the type annotation.

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
```diff
@@ -26,13 +33,80 @@ def __init__(self, workflow):
...
+            neither of these issues should be captured by this method, since
+            it runs after rhpkg generates the tarballs from the repos listed
+            within source-repos and any potential issue should be handled by
+            rhpkg.
+        """
+        source_repos = build_dir_path / EXPLODED_SOURCES_FILE
+        filenames = []
+        if not source_repos.exists():
+            return filenames
+        regex = re.compile(r'^([\S]+)\s+([0-9a-f]+)$')
```
chmeliik, line 62: If I'm looking at the right rhpkg version, the regex they use is `r'^([\S]+)\s+([\S]+)$'` (which could be simplified to `r'^(\S+)\s+(\S+)$'`). It seems that our regex puts additional restrictions on the git ref, is that OK?

And another detail is that rhpkg matches this regex against `line.strip()`, so it allows extra whitespace on both ends. Now that I think about it, the rhpkg approach is just an unnecessarily complicated way to do `line.split()` and check that it splits to 2 elements :smile:

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
chmeliik: +1, I like this approach more, just a minor detail about potential inconsistency with the way rhpkg does things

atomic_reactor/plugins/pre_pyrpkg_fetch_artefacts.py
chmeliik: (nice find on `Final` though, I didn't know it existed :+1:)

tkdchen: Good catch! Thanks :smiley:

chmeliik (approved):
LGTM :+1:

lkolacek (approved):

PR #1732, opened 2021-12-13
Implement part of ImageUtil

chmeliik:
CLOUDBLD-8036

Implement the methods that will be needed by prebuild plugins.

Add the workflow.imageutil property that provides an ImageUtil object
for plugins.

Remove stubs of newly implemeted imageutil methods and fix all the
plugins that used those stubs.

### DIFF
--- a/atomic_reactor/plugins/exit_store_metadata.py
+++ b/atomic_reactor/plugins/exit_store_metadata.py
@@ -17,7 +17,6 @@ from atomic_reactor.constants import (PLUGIN_KOJI_UPLOAD_PLUGIN_KEY,
                                       SCRATCH_FROM)
 from atomic_reactor.config import get_openshift_session
 from atomic_reactor.plugin import ExitPlugin
-from atomic_reactor.utils import imageutil
 
 
 class StoreMetadataPlugin(ExitPlugin):
@@ -220,8 +219,8 @@ class StoreMetadataPlugin(ExitPlugin):
                     not self.workflow.dockerfile_images.base_from_scratch):
                 base_image_name = base_image
                 try:
-                    # OSBS2 TBD
-                    base_image_id = imageutil.base_image_inspect().get('Id', "")
+                    # OSBS2 TBD: we probably don't need this and many other annotations anymore
+                    base_image_id = self.workflow.imageutil.base_image_inspect().get('Id', "")
                 except KeyError:
                     base_image_id = ""
             else:
--- a/atomic_reactor/plugins/pre_add_image_content_manifest.py
+++ b/atomic_reactor/plugins/pre_add_image_content_manifest.py
@@ -19,9 +19,7 @@ from atomic_reactor.constants import (IMAGE_BUILD_INFO_DIR, INSPECT_ROOTFS,
                                       PLUGIN_FETCH_MAVEN_KEY)
 from atomic_reactor.config import get_cachito_session
 from atomic_reactor.plugin import PreBuildPlugin
-from atomic_reactor.util import (base_image_is_scratch, df_parser, read_yaml,
-                                 read_content_sets, map_to_user_params
-                                 )
+from atomic_reactor.util import df_parser, read_yaml, read_content_sets, map_to_user_params
 from atomic_reactor.utils.pnc import PNCUtil
 
 
@@ -120,13 +118,19 @@ class AddImageContentManifestPlugin(PreBuildPlugin):
 
     @property
     def layer_index(self):
-        if self._layer_index is None:
-            # Default layer index is 1, because base and 'FROM scratch' images
-            #     *always* have 2 layers
-            self._layer_index = 1
-        if not base_image_is_scratch(self.dfp.baseimage):
-            # OSBS2 TBD: decide if we need to inspect a specific arch
-            inspect = self.workflow.imageutil.base_image_inspect()
+        if self._layer_index is not None:
+            return self._layer_index
+
+        # OSBS2 TBD: decide if we need to inspect a specific arch
+        inspect = self.workflow.imageutil.base_image_inspect()
+        if not inspect:
+            # Base images ('FROM koji/image-build') and 'FROM scratch' images do not have any
+            #   base image. When building with `podman build --squash`, such images get squashed
+            #   to only 1 layer => the layer index in this case is 0 (the first and only layer).
+
+            # OSBS2 TBD: this is only true for build tasks that behave like `podman build --squash`
+            self._layer_index = 0
+        else:
             self._layer_index = len(inspect[INSPECT_ROOTFS][INSPECT_ROOTFS_LAYERS])
         return self._layer_index
 
--- a/atomic_reactor/plugins/pre_add_image_content_manifest.py
+++ b/atomic_reactor/plugins/pre_add_image_content_manifest.py
@@ -22,7 +22,6 @@ from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.util import (base_image_is_scratch, df_parser, read_yaml,
                                  read_content_sets, map_to_user_params
                                  )
-from atomic_reactor.utils import imageutil
 from atomic_reactor.utils.pnc import PNCUtil
 
 
@@ -126,8 +125,8 @@ class AddImageContentManifestPlugin(PreBuildPlugin):
             #     *always* have 2 layers
             self._layer_index = 1
         if not base_image_is_scratch(self.dfp.baseimage):
-            # OSBS2 TBD
-            inspect = imageutil.base_image_inspect()
+            # OSBS2 TBD: decide if we need to inspect a specific arch
+            inspect = self.workflow.imageutil.base_image_inspect()
             self._layer_index = len(inspect[INSPECT_ROOTFS][INSPECT_ROOTFS_LAYERS])
         return self._layer_index
 
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -60,7 +60,6 @@ from atomic_reactor.constants import INSPECT_CONFIG
 from atomic_reactor.util import (df_parser,
                                  label_to_string,
                                  LabelFormatter)
-from atomic_reactor.utils import imageutil
 from osbs.utils import Labels
 import json
 import datetime
@@ -278,8 +277,8 @@ class AddLabelsPlugin(PreBuildPlugin):
             base_image_labels = {}
         else:
             try:
-                # OSBS2 TBD
-                config = imageutil.base_image_inspect()[INSPECT_CONFIG]
+                # OSBS2 TBD: inspect the correct architecture
+                config = self.workflow.imageutil.base_image_inspect()[INSPECT_CONFIG]
             except KeyError as exc:
                 message = "base image was not inspected"
                 self.log.error(message)
--- a/atomic_reactor/plugins/pre_change_from_in_df.py
+++ b/atomic_reactor/plugins/pre_change_from_in_df.py
@@ -11,7 +11,6 @@ to the more specific names given by the builder.
 """
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.util import df_parser, base_image_is_scratch
-from atomic_reactor.utils import imageutil
 
 
 class BaseImageMismatch(RuntimeError):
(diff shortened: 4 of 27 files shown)


atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -124,10 +123,12 @@ def layer_index(self):
         if self._layer_index is None:
             # Default layer index is 1, because base and 'FROM scratch' images
             #     *always* have 2 layers
+            # TODO: was this true for 'FROM scratch' images? Don't those have just 1 layer?
             self._layer_index = 1
```
chmeliik, line 127: Does anyone know if this is true? I think scratch images get squashed to 1 layer, and since we will be converting custom base images to `FROM scratch` as well, those will get 1-squashed too.

atomic_reactor/plugins/exit_store_metadata.py
```diff
@@ -220,8 +219,8 @@ def run(self):
                     not self.workflow.dockerfile_images.base_from_scratch):
                 base_image_name = base_image
                 try:
-                    # OSBS2 TBD
-                    base_image_id = imageutil.base_image_inspect().get('Id', "")
+                    # OSBS2 TBD: inspect the correct architecture?
```
rcerven, line 222: well it is adding annotation annotations['base-image-id']

and as of now that ran also on the orchestrator so it was x86

but I am not sure if that information is used anywhere

atomic_reactor/plugins/pre_add_image_content_manifest.py
rcerven: it is truth now, I am not sure what do we plan to do with podman and how that squashes

atomic_reactor/utils/imageutil.py
```diff
@@ -5,6 +5,91 @@
...
+        self._dockerfile_images = dockerfile_images
+        self._conf = conf
+
+    def set_dockerfile_images(self, dockerfile_images: util.DockerfileImages) -> None:
+        """Set a new dockerfile_images instance."""
+        self._dockerfile_images = dockerfile_images
+
+    def get_inspect_for_image(
+        self, image: Union[str, ImageName], platform: Optional[str] = None
+    ) -> dict:
```
tkdchen, line 45: How about reuse the type alias `ImageInspectionData`?

atomic_reactor/plugins/pre_add_image_content_manifest.py
chmeliik: Oh interesting, if you have a FROM scratch image then `podman build --squash` squashes it to 1 layer but imagebuilder squashes to 2 layers (the first one is an empty tarball). So we don't really know what the index will be, it depends on the build task implementation :thinking:

atomic_reactor/utils/imageutil.py
chmeliik: Good idea, I'll update the corresponding RegistryClient method as well

atomic_reactor/plugins/pre_add_image_content_manifest.py
rcerven: yep

atomic_reactor/utils/imageutil.py
chmeliik: updated

atomic_reactor/plugins/exit_store_metadata.py
chmeliik: It looks like it's not used. Changed the TBD comment, we might want to remove it eventually

atomic_reactor/plugins/pre_add_image_content_manifest.py
chmeliik: Changed the index to 0 for now with a note about the potential issue with different build tasks

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -121,13 +118,19 @@ def __init__(self, workflow, remote_sources=None, destdir=IMAGE_BUILD_INFO_DIR):
...
-            # Default layer index is 1, because base and 'FROM scratch' images
-            #     *always* have 2 layers
-            self._layer_index = 1
-        if not base_image_is_scratch(self.dfp.baseimage):
-            # OSBS2 TBD
-            inspect = imageutil.base_image_inspect()
+        if self._layer_index is not None:
+            return self._layer_index
+
+        # OSBS2 TBD: decide if we need to inspect a specific arch
```
rcerven, line 124: we don't because image for all arches should have same amount of layers

chmeliik: That's what I'm thinking as well, but technically, nothing in the docker/OCI specification says that all images in a manifest list must be "the same" (and it's not clear to me what exactly that would mean anyway).

It's probably OK to assume that all images in a manifest list have the same number of layers, but I'd rather leave this here to be resolved when we update the content manifest plugin

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -278,8 +277,8 @@ def run(self):
             base_image_labels = {}
         else:
             try:
-                # OSBS2 TBD
-                config = imageutil.base_image_inspect()[INSPECT_CONFIG]
+                # OSBS2 TBD: decide if we need to inspect a specific arch
```
rcerven, line 280: atm we are generating auto labels which are arch specific (but they don't depend on base image ones)

and adding aliases for old labels (which shouldn't be arch specific)

but any of this could change in the future, we we probably should get arch specific just to be sure

atomic_reactor/plugins/pre_change_from_in_df.py
```diff
@@ -92,8 +91,8 @@ def run(self):
                 new_parents.append(df_img)
                 continue
             local_image = self.workflow.dockerfile_images[df_img]
-            # OSBS2 TBD
-            inspection = imageutil.get_inspect_for_image(local_image)
+            # OSBS2 TBD: we don't want to map images to Ids anymore
```
rcerven, line 94: but we still want there arch specific image sha

atomic_reactor/plugins/pre_distribution_scope.py
```diff
@@ -89,11 +88,11 @@ def run(self):
             scope = self.get_scope('current', labels)
 
             # Find out the parent's intended scope
-            # OSBS2 TBD
-            inspect = imageutil.base_image_inspect()
+            # OSBS2 TBD: decide if we need to inspect a specific arch
```
rcerven, line 91: distribution scope is set by osbs, so it will be same for all arches of the parent so we don't nee arch specific

atomic_reactor/plugins/pre_hide_files.py
```diff
@@ -91,9 +90,9 @@ def _update_dockerfile(self, from_structure, end_structure, stage_user):
         if parent_image_id == SCRATCH_FROM:
             return
 
-        # OSBS2 TBD
-        inspect = imageutil.get_inspect_for_image(parent_image_id)
-        inherited_user = inspect.get(INSPECT_CONFIG).get('User', '')
+        # OSBS2 TBD: decide if we need to inspect a specific arch
```
rcerven, line 93: USER in base image shouldn't be arch dependent

atomic_reactor/plugins/pre_inject_yum_repo.py
```diff
@@ -29,9 +28,9 @@ def _final_user_line(self):
             return user
 
         if not self.workflow.dockerfile_images.base_from_scratch:
-            # OSBS2 TBD
-            inspect = imageutil.base_image_inspect()
-            user = inspect.get(INSPECT_CONFIG).get('User')
+            # OSBS2 TBD: decide if we need to inspect a specific arch
```
rcerven, line 31: USER in base image shouldn't be arch dependent

atomic_reactor/plugins/pre_koji_parent.py
```diff
@@ -83,8 +82,8 @@ def run(self):
                 self.workflow.dockerfile_images.custom_base_image):
             self._base_image_nvr = self.detect_parent_image_nvr(
                 self.workflow.dockerfile_images.base_image,
-                # OSBS2 TBD
-                inspect_data=imageutil.base_image_inspect(),
+                # OSBS2 TBD: decide if we need to inspect a specific arch
```
rcerven, line 85: koji parent is checking for labels to get NVR of brew build

release/version/component will be same for all arches

atomic_reactor/utils/imageutil.py
```diff
@@ -5,18 +5,95 @@
...
+
+        The result is cached, this method will not query the registry more than once
+        (if successful) for the same image reference (within one Python process).
+
+        :param image: The image to inspect
+        :param platform: Optionally, inspect the base image for a specific platform.
+            This can be either the platform name (e.g. x86_64) or the GOARCH name (amd64).
+        """
+        if not inspectable(image):
+            raise ValueError(f"{image!r} is not inspectable")
```
rcerven, line 59: mmm I wonder if we shouldn't just return {}

I mean code shouldn't try to inspect scratch/custom base images anyway, but still

atomic_reactor/utils/imageutil.py
```diff
@@ -5,18 +5,95 @@
...
+        if not inspectable(base_image):
+            return {}
+
+        return self.get_inspect_for_image(base_image, platform)
+
+    @functools.lru_cache(maxsize=None)
+    def _cached_inspect_image(
+        self, image: str, goarch: Optional[str] = None
+    ) -> ImageInspectionData:
+        # Important: this method must take the image name as a string, not an ImageName.
```
rcerven, line 82: and why don't you just convert ImageName (if provided) to string and then parse it again ?

rcerven: but well if it will be used just internally here it doesn't matter

atomic_reactor/plugins/pre_change_from_in_df.py
chmeliik: Manifest list digest, right? updated

atomic_reactor/plugins/pre_add_labels_in_df.py
chmeliik: Yup, changed comment to suggest inspecting correct arch

chmeliik:
@rcerven updated the TBDs where we *do* want to inspect the right arch. Left the others in place, the decision *not* to inspect the correct arch is potentially dangerous and I don't want to make them in this PR

atomic_reactor/utils/imageutil.py
chmeliik: > and why don't you just convert ImageName (if provided) to string and then parse it again ?

Yeah that's what I'm doing. `get_inspect_for_image` accepts both string and ImageName, converts to string, and `_cached_inspect_image`  parses back to ImageName.

atomic_reactor/utils/imageutil.py
chmeliik: Hmm, yeah that would be more consistent with how `base_image_inspect` behaves. On the other hand, unlike the `base_image_inspect` method, when calling this one the caller knows what image they are passing as an argument so maybe it should be up to them to make sure the image is valid.

I'm fine with either, I'd like to know what others think

atomic_reactor/utils/imageutil.py
```diff
@@ -5,6 +5,91 @@
...
+import functools
+from typing import Optional, Union
+
+from osbs.utils import ImageName
+
+from atomic_reactor import config
+from atomic_reactor import util
+
+
+def inspectable(image: Union[str, ImageName]) -> bool:
```
mkosiarc, line 19: Since this function is not a part of a class, the name itself - inspectable does not tell me much without that docs, so maybe something like  is_image_inspectable would improve readability

mkosiarc (approved):

atomic_reactor/utils/imageutil.py
chmeliik: +1, renamed

lkolacek (approved):

atomic_reactor/plugins/pre_change_from_in_df.py
rcerven: right or manifest list digest

rcerven (approved):

chmeliik:
No release notes, internal util module

PR #1736, opened 2021-12-15
pin_operator_digest: update plugin to work in tekton

hjmodi:
Remove worker and orchestrator differences completely,
and unify the login in the run() method of the plugin.
In OSBS 2 there will not be any worker orchestrator
abstraction.

* CLOUDBLD-8112

### DIFF
--- a/atomic_reactor/plugins/pre_pin_operator_digest.py
+++ b/atomic_reactor/plugins/pre_pin_operator_digest.py
@@ -7,6 +7,7 @@ of the BSD license. See the LICENSE file for details.
 """
 
 import logging
+import os.path
 
 from osbs.utils import Labels, ImageName
 
@@ -258,7 +259,7 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
                 operator_csv.modifications_update(
                     operator_csv_modifications.get('update', {}))
 
-            operator_csv.dump()
+            self.workflow.build_dir.for_each_platform(operator_csv.dump)
         else:
             self.log.warning("%s has a relatedImages section, skipping", operator_csv.path)
 
@@ -302,9 +303,13 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
                                "bundles".format(component))
 
     def _get_operator_manifest(self):
-        manifests_dir = self.workflow.source.manifests_dir
-        self.log.info("Looking for operator CSV files in %s", manifests_dir)
-        operator_manifest = OperatorManifest.from_directory(manifests_dir)
+        if self.workflow.source.config.operator_manifests is None:
+            raise RuntimeError("operator_manifests configuration missing in container.yaml")
+        self.log.info("Looking for operator CSV files in %s", self.workflow.build_dir.path)
+        manifests_dir = self.workflow.source.config.operator_manifests["manifests_dir"]
+        kwargs = {'repo_dir': self.workflow.build_dir.any_build_dir.path}
+        operator_manifest = OperatorManifest.from_directory(
+            os.path.join(self.workflow.build_dir.any_build_dir.path, manifests_dir), **kwargs)
         self.log.info("Found operator CSV file: %s", operator_manifest.csv.path)
 
         return operator_manifest
--- a/atomic_reactor/plugins/pre_pin_operator_digest.py
+++ b/atomic_reactor/plugins/pre_pin_operator_digest.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2020 Red Hat, Inc
+Copyright (c) 2021 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
@@ -26,7 +26,6 @@ from osbs.utils.yaml import (
     load_schema,
     validate_with_schema,
 )
-from atomic_reactor.plugins.build_orchestrate_build import override_build_kwarg
 from atomic_reactor.utils.operator import OperatorManifest
 from atomic_reactor.utils.retries import get_retrying_requests_session
 
@@ -35,17 +34,13 @@ class PinOperatorDigestsPlugin(PreBuildPlugin):
     """
     Plugin runs for operator manifest bundle builds.
 
-    When running in orchestrator:
     - finds container pullspecs in operator ClusterServiceVersion files
     - computes replacement pullspecs:
         - replaces tags with manifest list digests
         - replaces repos (and namespaces) based on operator_manifests.repo_replacements
           configuration in container.yaml and r-c-m*
         - replaces registries based on operator_manifests.registry_post_replace in r-c-m*
-
-    When running in a worker:
-    - receives replacement pullspec mapping computed by orchestrator
-    - replaces pullspecs in ClusterServiceVersion files based on said mapping
+    - replaces pullspecs in ClusterServiceVersion files based on replacement pullspec mapping
     - creates relatedImages sections in ClusterServiceVersion files
 
     Files that already have a relatedImages section are excluded.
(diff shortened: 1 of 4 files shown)


atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -67,8 +62,7 @@ def __init__(self, workflow, replacement_pullspecs=None,
         Initialize pin_operator_digests plugin
 
         :param workflow: DockerBuildWorkflow instance
-        :param replacement_pullspecs: Dict[str, str], computed in orchestrator,
-                                      provided to workers by osbs-client
+        :param replacement_pullspecs: Dict[str, str]
```
chmeliik, line 65: This param can be removed, and ideally the `self.replacement_pullspecs` attribute as well (now, `replacement_pullspecs` will be just be computed in the run method)

chmeliik (commented):
Nice, now the only necessary changes should be to `_get_operator_manifest()` from a build_dir rather than the source dir and make the changes (pretty much just `operator_csv.dump()`?) in all build dirs

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -267,35 +227,17 @@ def run_in_orchestrator(self):
         else:
             if pullspecs:
                 replacement_pullspecs = self._get_replacement_pullspecs(pullspecs)
```
rcerven, line 229: this variable should be set to {} or None before IF

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -267,35 +227,17 @@ def run_in_orchestrator(self):
...
-
-        Exclude CSVs which already have a relatedImages section.
-        """
-        operator_manifest = self._get_operator_manifest()
-
-        if self._skip_all():
-            self.log.warning("skip_all defined, not running on worker")
-            return
-
         replacement_pullspecs = {
```
rcerven, line 235: you should use replacement_pullspecs which you've got previously

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -318,6 +260,23 @@ def run_in_worker(self):
...
+
+        :return: bool, should plugin run?
+        """
+        if not has_operator_bundle_manifest(self.workflow):
+            self.log.info("Not an operator manifest bundle build, skipping plugin")
+            return False
+        if not self.workflow.conf.operator_manifests:
+            msg = "operator_manifests configuration missing in reactor config map, aborting"
+            self.log.warning(msg)
+            return False
```
rcerven, line 277: I wonder wonder shouldn't this actually fail? I mean you want to use feature so you expect it to work, but this way we are just silently doing nothing if it isn't configured

@chmeliik @MartinBasti

chmeliik: Hmm, hard to say. The current behavior is not ideal, but immediately failing if the configuration is missing may not be ideal either. Users have ways to do the pinning themselves and tell OSBS to skip processing, so maybe that should be evaluated before checking for server config and raising errors.

TL;DR, I don't think it should be fixed in this PR

atomic_reactor/plugins/pre_pin_operator_digest.py
hjmodi: updated.

atomic_reactor/plugins/pre_pin_operator_digest.py
hjmodi: updated.

atomic_reactor/plugins/pre_pin_operator_digest.py
hjmodi: that one is a list and the latest parts require a dictionary so this step converts it.

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -451,18 +410,11 @@ def _get_replacement_pullspecs_from_csv_modifications(self, pullspecs):
         pullspecs_set = set(pullspecs)
         mod_pullspecs_set = set((ImageName.parse(p['original']) for p in mod_pullspec_repl))
 
-        missing = pullspecs_set - mod_pullspecs_set
-        if missing:
+        diff = pullspecs_set - mod_pullspecs_set
```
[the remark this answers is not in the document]
hjmodi: updated the testcase to do the same.

hjmodi:
> Nice, now the only necessary changes should be to `_get_operator_manifest()` from a build_dir rather than the source dir and make the changes (pretty much just `operator_csv.dump()`?) in all build dirs

hmm, let me look into it. My assumption was that we're getting config from the source repo which should ideally be in the build_dir only so we don't have to worry about it here.

I'll try to ensure that's the case.

chmeliik:
> hmm, let me look into it. My assumption was that we're getting config from the source repo which should ideally be in the build_dir only so we don't have to worry about it here.
> 
> I'll try to ensure that's the case.

Config can be read from the source repo, but the manifest files themselves should probably be read from one of the platform-specific build dirs. Though it might be fine to *read* manifests from the source repo as long as all *write* operations are done only in build dirs, so if it's not an easy change to make it's probably fine

tests/plugins/test_pin_operator_digests.py
```diff
@@ -713,26 +713,6 @@ def test_replacement_opt_out(self, pin_digest, replace_repo, replace_registry,
...
-        original_content = csv.read_text("utf-8")
-
-        user_config = get_user_config(OPERATOR_MANIFESTS_DIR)
-        runner = mock_env(workflow, source_dir, site_config=get_site_config(),
-                          user_config=user_config)
-        runner.run()
-
-        assert "Replacing pullspecs" not in caplog.text
-        assert "Creating relatedImages section" not in caplog.text
-        assert csv.read_text("utf-8") == original_content
```
chmeliik, line 724: This test checked that OSBS will skip CSV files if they already have a relatedImages section. Is that still tested somewhere?

chmeliik (commented):
@hjmodi the multi-arch handling still needs to be added, right? The plugin will have to use `workflow.build_dir.for_each_platform()` or `for_all_platforms_copy()` to make changes in build dirs

tests/plugins/test_pin_operator_digests.py
hjmodi: It's sort of a weird situation with this testcase. It was ensuring that the `run_in_worker` didn't fail if there was a conflict between `relateImages` and env variables of the same since it should have been caught in `run_in_orchestrator`(it throws run time error if skip_all is False and if it's True the worker doesn't run.). This didn't make sense to me since `run_in_orchestrator` was always running before `run_in_worker` and this testcase was simulating something that wasn't actually supposed to happen.

Furthermore, in OSBS2 `run_in_orchestrator` and `run_in_worker` are merged in the `run` method so `run_in_orchestrator` steps would always execute before `run_in_worker` steps which invalidate this testcase.

chmeliik: I think the comment at the top is the confusing part here. The test doesn't check the conflict handling, it checks that the worker part of the plugin will not touch CSV files if they already have `relatedImages`, regardless of whether there is a conflict or not. I think the fix for this test would be to remove the `has_envs` parametrization and set it to False (so that there is no conflict). The conflict handling should already have its own test I think.

hjmodi: Oh I see. It makes sense now. Updated the testcase instead of removing it.

hjmodi:
> @hjmodi the multi-arch handling still needs to be added, right? The plugin will have to use `workflow.build_dir.for_each_platform()` or `for_all_platforms_copy()` to make changes in build dirs

updated the code to handle multi-arch :)

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -341,9 +303,13 @@ def _skip_all(self):
...
-        manifests_dir = self.workflow.source.manifests_dir
-        self.log.info("Looking for operator CSV files in %s", manifests_dir)
-        operator_manifest = OperatorManifest.from_directory(manifests_dir)
+        if self.workflow.source.config.operator_manifests is None:
+            raise RuntimeError("operator_manifests configuration missing in container.yaml")
+        self.log.info("Looking for operator CSV files in %s", self.workflow.build_dir.path)
+        manifests_dir = self.workflow.source.config.operator_manifests["manifests_dir"]
+        kwargs = {'repo_dir': self.workflow.build_dir.any_build_dir.path}
+        operator_manifest = OperatorManifest.from_directory(
+            os.path.join(self.workflow.build_dir.any_build_dir.path, manifests_dir), **kwargs)
```
rcerven, line 312: why are you passing kwargs again with the same path?
originally there was just OperatorManifest.from_directory(manifests_dir)

rcerven (approved):

lkolacek (approved):

atomic_reactor/plugins/pre_pin_operator_digest.py
```diff
@@ -302,9 +303,13 @@ def _skip_all(self):
                                "bundles".format(component))
 
     def _get_operator_manifest(self):
-        manifests_dir = self.workflow.source.manifests_dir
-        self.log.info("Looking for operator CSV files in %s", manifests_dir)
-        operator_manifest = OperatorManifest.from_directory(manifests_dir)
+        if self.workflow.source.config.operator_manifests is None:
+            raise RuntimeError("operator_manifests configuration missing in container.yaml")
+        self.log.info("Looking for operator CSV files in %s", self.workflow.build_dir.path)
+        manifests_dir = self.workflow.source.config.operator_manifests["manifests_dir"]
```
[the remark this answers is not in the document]
chmeliik: Filed internal tracker

PR #1744, opened 2022-01-05
Add unittests for json schemas

mkosiarc:
CLOUDBLD-7880

### DIFF
--- a/tests/test_schemas.py
+++ b/tests/test_schemas.py
@@ -1,0 +1,311 @@
+"""
+Copyright (c) 2022 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+
+import pytest
+
+from atomic_reactor.util import validate_with_schema
+from osbs.exceptions import OsbsValidationException
+
+
+SOURCE_CONTAINERS_USER_PARAMS_ALL_PROPERTIES = {
+    "build_json_dir": "/usr/share/osbs/",
+    "component": "osbs-test-base-container-source",
+    "image_tag": "user/osbs-test-base-container-source:osbs-test-1.0-rhel-7",
+    "kind": "source_containers_user_params",
+    "koji_target": "osbs-test-1.0-rhel-7-containers-candidate",
+    "namespace": "exd-sp-guild-container-build--osbs-qa-source-containers",
+    "pipeline_run_name": "source-container-0-1-49063-20211129161231",
+    "reactor_config_map": "reactor-config-map-user",
+    "sources_for_koji_build_nvr": "osbs-test-base-container-1.0.distscope.base-2396",
+    "user": "user",
+}
+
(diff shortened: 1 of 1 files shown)


mkosiarc:
Other schemas are covered in other places, sometimes as parts of testing the whole plugins - uniting those all in one file would take more effort. And I am also interested in your opinion on those cases I commented with "TODO should this fail?"

tests/test_schemas.py
```diff
@@ -0,0 +1,303 @@
+import pytest
```
tkdchen, line 1: For a new file, should we add the license declaration at the beginning like others?

tests/test_schemas.py
```diff
@@ -0,0 +1,303 @@
...
+        (OPERATOR_CSV_MODIFICATIONS_MISSING_ORIGINAL_PULLSPEC,
+         "schemas/operator_csv_modifications.json",
+         r"validating 'required' has failed \('original' is a required property\)"),
+        (OPERATOR_CSV_MODIFICATIONS_MISSING_NEW_PULLSPEC, "schemas/operator_csv_modifications.json",
+         r"validating 'required' has failed \('new' is a required property\)"),
+        (OPERATOR_CSV_MODIFICATIONS_MISSING_PINNED_PULLSPEC,
+         "schemas/operator_csv_modifications.json",
+         r"validating 'required' has failed \('pinned' is a required property\)"),
+        ({"pullspec_replacements": []},
+         "schemas/operator_csv_modifications.json", None),  # TODO should this fail?
```
chmeliik, line 271: Empty pullspec replacements would be valid if the original CSV had no pullspecs. Not sure how much sense that makes, but it looks like the code is ready to handle that, so this seems fine.

tests/test_schemas.py
```diff
@@ -0,0 +1,303 @@
...
+         "schemas/user_params.json",
+         r"validating 'required' has failed \('user' is a required property\)"),
+        (USER_PARAMS_MISSING_GIT_URI,
+         "schemas/user_params.json",
+         r"validating 'required' has failed \('git_uri' is a required property\)"),
+        (USER_PARAMS_MISSING_GIT_REF,
+         "schemas/user_params.json",
+         r"validating 'required' has failed \('git_ref' is a required property\)"),
+        (USER_PARAMS_ADDITIONAL_NONDEFINED_PROPERTY,
+         "schemas/user_params.json", None),  # TODO should this fail?
```
chmeliik, line 256: Probably should, but if we change it now, we potentially break existing builds :disappointed:

tests/test_schemas.py
mkosiarc: updated

tests/test_schemas.py
mkosiarc: This is maybe not an big issue and can probably be left as is, I'll keep this conversation unresolved in case anybody else wants to pitch in.

chmeliik (approved):
LGTM :+1:

lkolacek (approved):

PR #1747, opened 2022-01-10
Update Pipeline definition for parallel build tasks

mkosiarc:
and add --platform CLI option for binary-container-build
CLOUDBLD-8135

### DIFF
--- a/atomic_reactor/cli/task.py
+++ b/atomic_reactor/cli/task.py
@@ -6,7 +6,7 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
 from atomic_reactor.tasks.binary import BinaryBuildTask, BinaryExitTask, \
-    BinaryPostBuildTask, BinaryPreBuildTask
+    BinaryPostBuildTask, BinaryPreBuildTask, BinaryBuildTaskParams
 from atomic_reactor.tasks.clone import CloneTask
 from atomic_reactor.tasks.common import TaskParams
 from atomic_reactor.tasks.orchestrator import OrchestratorTask, OrchestratorTaskParams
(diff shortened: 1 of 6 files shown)


rcerven:
code looks good, but shouldn't there be already some check based on workflow.json that it will stop/cancel build for not requested platform? 

cc @chmeliik

chmeliik:
> code looks good, but shouldn't there be already some check based on workflow.json that it will stop/cancel build for not requested platform?
> 
> cc @chmeliik

workflow.json is not implemented yet, so no check needed for now. It is tracked in CLOUDBLD-8136

atomic_reactor/cli/task.py
```diff
@@ -69,7 +69,7 @@ def binary_container_build(task_args: dict):
 
     :param task_args: CLI arguments for a binary-container-build task
     """
-    params = TaskParams.from_cli_args(task_args)
+    params = BinaryContainerBuildTaskParams.from_cli_args(task_args)
     task = BinaryBuildTask(params)
```
chmeliik, line 73: I'd keep the naming consistent as `<TaskName>Params`

chmeliik (commented):
LGTM, just a minor nitpick about the class name

rcerven (approved):

atomic_reactor/cli/task.py
hjmodi: +1. Should we remove `container` from `REQUIRED_PLATFORM_FOR_BINARY_CONTAINER_BUILD`, `binary-container-prebuild`, `binary-container-postbuild` etc. as well? I think the container part is a bit redundant already.

mkosiarc: are we completely sure? It would have to be changed in code, in pipeline/task definitions and possibly some other places to keep consistency - for example the pvc names.

chmeliik: Hmm that's true, some of the names might even be scattered across different projects (like downstream playbooks). I don't mind leaving it as is

hjmodi: I'll vote yes

chmeliik (approved):
LGTM :+1:

atomic_reactor/cli/task.py
mkosiarc: @hjmodi would you mind if we leave it as it is for now in this mr and change it later on?

PR #1749, opened 2022-01-11
resolve_composes: don't set worker args

hjmodi:
Instead return results in plugin results. In the new
result format everything will be per arch and noarch repo
will be added to all arches that we're currently building for.

* CLOUDBLD-8107

### DIFF
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -77,6 +77,7 @@ class ResolveComposesPlugin(PreBuildPlugin):
         self.has_complete_repos = len(self.repourls) > 0
         self.plugin_result = self.workflow.prebuild_results.get(PLUGIN_KOJI_PARENT_KEY)
         self.all_compose_ids = list(self.compose_ids)
+        self.new_compose_ids = []
         self.parent_compose_ids = []
         self.include_koji_repo = False
         self.yum_repourls = defaultdict(list)
@@ -101,13 +102,8 @@ class ResolveComposesPlugin(PreBuildPlugin):
             self.wait_for_composes()
         except WaitComposeToFinishTimeout as e:
             self.log.info(str(e))
-            preserve_composes = set(self.compose_ids).union(self.parent_compose_ids)
-            cancel_composes = set(self.all_compose_ids) - preserve_composes
-            if cancel_composes:
-                self.log.info('Canceling unfinished composes which were created by the build: %s',
-                              cancel_composes)
 
-            for compose_id in cancel_composes:
+            for compose_id in self.new_compose_ids:
                 if self.odcs_client.get_compose_status(compose_id) in ['wait', 'generating']:
                     self.log.info('Canceling the compose %s', compose_id)
                     self.odcs_client.cancel_compose(compose_id)
@@ -254,7 +250,8 @@ class ResolveComposesPlugin(PreBuildPlugin):
 
         for compose_request in self.compose_config.render_requests():
             compose_info = self.odcs_client.start_compose(**compose_request)
-            self.all_compose_ids.append(compose_info['id'])
+            self.new_compose_ids.append(compose_info['id'])
+        self.all_compose_ids.extend(self.new_compose_ids)
 
     def wait_for_composes(self):
         self.log.debug('Waiting for ODCS composes to be available: %s', self.all_compose_ids)
@@ -277,6 +274,7 @@ class ResolveComposesPlugin(PreBuildPlugin):
 
                 compose_info = self.odcs_client.renew_compose(compose_id, sigkeys)
                 compose_id = compose_info['id']
+                self.new_compose_ids.append(compose_id)
                 compose_info = self.odcs_client.wait_for_compose(compose_id)
 
             self.composes_info.append(compose_info)
--- a/atomic_reactor/plugins/pre_resolve_composes.py
+++ b/atomic_reactor/plugins/pre_resolve_composes.py
@@ -1,25 +1,24 @@
 """
-Copyright (c) 2017 Red Hat, Inc
+Copyright (c) 2017-2022 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
+from collections import defaultdict
 from copy import deepcopy
 from datetime import datetime, timedelta
-from collections import defaultdict
 
-from atomic_reactor import util
-from atomic_reactor.utils.odcs import WaitComposeToFinishTimeout
 from osbs.repo_utils import ModuleSpec
 
+from atomic_reactor import util
+from atomic_reactor.config import get_koji_session, get_odcs_session
 from atomic_reactor.constants import (PLUGIN_KOJI_PARENT_KEY,
                                       PLUGIN_RESOLVE_COMPOSES_KEY,
                                       BASE_IMAGE_KOJI_BUILD)
-from atomic_reactor.config import get_koji_session, get_odcs_session
 from atomic_reactor.plugin import PreBuildPlugin
-from atomic_reactor.plugins.build_orchestrate_build import override_build_kwarg
 from atomic_reactor.util import get_platforms, is_isolated_build, is_scratch_build
+from atomic_reactor.utils.odcs import WaitComposeToFinishTimeout
 
 ODCS_DATETIME_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
 MINIMUM_TIME_TO_EXPIRE = timedelta(hours=2).total_seconds()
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -105,13 +105,8 @@ def run(self):
             self.wait_for_composes()
         except WaitComposeToFinishTimeout as e:
             self.log.info(str(e))
-            preserve_composes = set(self.compose_ids).union(self.parent_compose_ids)
-            cancel_composes = set(self.all_compose_ids) - preserve_composes
-            if cancel_composes:
-                self.log.info('Canceling unfinished composes which were created by the build: %s',
-                              cancel_composes)
 
-            for compose_id in cancel_composes:
```
chmeliik, line 108: I feel like this may have been intentional. Not 100% sure though, the plugin updates `all_compose_ids` so many times that I've lost track of what it's supposed to represent at this point in the code

chmeliik: I *think* that:
- compose_ids is the list of IDs that the user specified when starting the build
- parent_compose_ids is the list of compose IDs from the parent build
- all_compose_ids is compose_ids + parent_compose_ids + IDs for composes that OSBS requested for this build

So the cancel_composes list probably *was* what we wanted - the list of IDs for composes that were requested in this build

But I would appreciate if someone could double-check my theory, I'm still confused by this plugin

rcerven: yeah your logic is correct that is exactly how it works

so by removing that part of code, you are also checking status of composes finished already in the past, so I'd just leave that code here and check only really those newly requested ones

rcerven: 1 note/bug is that in case that compose_ids were provided AND they required renewal, we aren't canceling those

but that is because in 'wait_for_composes' it is looping through them and requesting renewal if required, and once loop finishes it updates all_compose_ids

so maybe it would be best just create something like self.new_compose_ids
and add there ids in 'request_compose_if_needed' and 'wait_for_composes' (if self._needs_renewal)

that way we will collect all newly requested ones and we can use just that

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -329,58 +330,49 @@ def resolve_signing_intent(self):
...
-        for arch, repofiles in repos_by_arch.items():
-            # OSBS2 TBD: `override_build_kwarg` is imported from build_orchestrate_build
-            override_build_kwarg(self.workflow, 'yum_repourls', repofiles, arch)
-
-        # Only set the None override if there are no other repos
-        if not repos_by_arch:
-            # OSBS2 TBD: `override_build_kwarg` is imported from build_orchestrate_build
-            override_build_kwarg(self.workflow, 'yum_repourls', noarch_repos, None)
+            for repourl in self.repourls:
+                self.yum_repourls['noarch'].add(repourl)
```
rcerven, line 348: just use union?

self.yum_repourls['noarch'].union(self.repourls)

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -329,58 +330,49 @@ def resolve_signing_intent(self):
...
+        if 'noarch' in self.yum_repourls:
+            noarch_repos = self.yum_repourls.pop('noarch')
+            for arch in self.yum_repourls:
+                for repo in noarch_repos:
+                    self.yum_repourls[arch].add(repo)
```
rcerven, line 353: again use union

tests/plugins/test_resolve_composes.py
```diff
@@ -360,13 +356,9 @@ def test_inherit_parents(self, mocked_env, parent_compose, parent_repourls,
...
         current_repourl = ["http://example.com/current.repo"]
-        expected_yum_repourls = []
-        if repo_provided:
-            expected_yum_repourls = list(current_repourl)
+        expected_yum_repourls = defaultdict(set)
         if not ids and compose_defined:
-            expected_yum_repourls.append(ODCS_COMPOSE['result_repofile'])
-        if allow_inherit and parent_repourls:
-            expected_yum_repourls.append(parent_repo)
+            expected_yum_repourls[ODCS_COMPOSE_DEFAULT_ARCH].add(ODCS_COMPOSE['result_repofile'])
```
rcerven, line 361: since you are adding always only default arch x86

change for consistency also : arches = ['ppc64le', 'x86_64']

because that ppc is basically ignored, because you request composes for both arches and return compose just for x86

rcerven: or just improve that compose_defined scenario and return results for both arches

atomic_reactor/plugins/pre_resolve_composes.py
hjmodi: got it. Thanks for the suggestion @rcerven . I have followed your suggestion and updated the logic accordingly.

atomic_reactor/plugins/pre_resolve_composes.py
hjmodi: that's what I did at first and after much pain, I found that for some reason `union` is not working on the `defaultdict`.

tests/plugins/test_resolve_composes.py
hjmodi: updated the test to include both arches.

atomic_reactor/plugins/pre_resolve_composes.py
hjmodi: that worked!! thanks :D

atomic_reactor/plugins/pre_resolve_composes.py
hjmodi: updated to use `.update()` since union wasn't working

rcerven (approved):

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -329,58 +334,47 @@ def resolve_signing_intent(self):
...
+        signing_intent = None
+        signing_intent_overridden = False
+        if self.compose_config:
+            signing_intent = self.compose_config.signing_intent['name']
+            signing_intent_overridden = self.compose_config.has_signing_intent_changed()
         result = {
             'composes': self.composes_info,
-            'signing_intent': self.compose_config.signing_intent['name'],
-            'signing_intent_overridden': self.compose_config.has_signing_intent_changed(),
+            'yum_repourls': self.yum_repourls,
```
chmeliik, line 370: It might be good to convert these to a `dict[str, list]` rather than `dict[str, set]` in this method - plugin results will soon need to be JSON-serializable

(@tkdchen can correct me if I'm wrong)

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -78,7 +77,13 @@ def __init__(self, workflow, koji_target=None, signing_intent=None, compose_ids=
         self.has_complete_repos = len(self.repourls) > 0
         self.plugin_result = self.workflow.prebuild_results.get(PLUGIN_KOJI_PARENT_KEY)
         self.all_compose_ids = list(self.compose_ids)
+        self.new_compose_ids = []
         self.parent_compose_ids = []
+        self.include_koji_repo = False
+        self.yum_repourls = defaultdict(set)
+        self.architectures = get_platforms(self.workflow)
+        if not self.architectures:
+            self.architectures = []
```
chmeliik, line 86:
```suggestion
        self.architectures = get_platforms(self.workflow) or []
```

atomic_reactor/plugins/pre_resolve_composes.py
```diff
@@ -88,24 +93,20 @@ def run(self):
         try:
             self.read_configs()
         except SkipResolveComposesPlugin as abort_exc:
-            # OSBS2 TBD: `override_build_kwarg` is imported from build_orchestrate_build
-            override_build_kwarg(self.workflow, 'yum_repourls', self.repourls, None)
             self.log.info('Aborting plugin execution: %s', abort_exc)
-            return
+            if self.repourls:
+                for arch in self.architectures:
+                    self.yum_repourls[arch].update(self.repourls)
```
chmeliik, line 99:
```suggestion
            for arch in self.architectures:
                self.yum_repourls[arch].update(self.repourls)
```
No need for the conditional, if self.repourls is an empty list it will behave as intended

chmeliik (commented):
Looks good :+1:

Just one suggestion about the return value and minor style suggestions

atomic_reactor/plugins/pre_resolve_composes.py
hjmodi: incorporated suggestion

atomic_reactor/plugins/pre_resolve_composes.py
hjmodi: incorporated suggestion

atomic_reactor/plugins/pre_resolve_composes.py
hjmodi: incorporated suggestion

chmeliik (approved):
LGTM, just one pylint issue

PR #1751, opened 2022-01-12
Fix wrong inject_yum_repo plugin name in prebuild task definition

tkdchen:
* CLOUDBLD-8243


chmeliik:
Didn't we agree that the plugin would be renamed to "inject_yum_repos" after the plugins get merged?

hjmodi:
I am working on merging the plugins to inject repos. Those plugins, once merged, are renamed to `inject_yum_repos` so this change will need to be changed back. I'd say rather than changing it back we just leave it as is.

tkdchen:
Hi, sorry for confusion. I'll close this PR and please go ahead with what you are doing now. I made this change to unblock my test.

PR #1753, opened 2022-01-13
Fetch sources earlier

tkdchen:

- [X] Commit messages are descriptive enough
- [X] Code coverage from testing does not decrease and new code is covered


chmeliik:
I'm really not a fan of the plugin system and how everything works implicitly, based on plugin order and global side effects -_-

But apart from my personal issues with the entire core of OSBS, this should work. The downside is that this means we will have one extra copy of everything fetched by rhpkg (1 in the source dir, 1 in each per-platform dir), but we can probably live with that.

lkolacek (approved):

rcerven (approved):

PR #1754, opened 2022-01-13
Update the content manifest plugin for Tekton

chmeliik:
CLOUDBLD-8121

### DIFF
--- a/atomic_reactor/plugins/pre_add_image_content_manifest.py
+++ b/atomic_reactor/plugins/pre_add_image_content_manifest.py
@@ -16,7 +16,8 @@ from osbs.utils import Labels
 from atomic_reactor.constants import (IMAGE_BUILD_INFO_DIR, INSPECT_ROOTFS,
                                       INSPECT_ROOTFS_LAYERS,
                                       PLUGIN_ADD_IMAGE_CONTENT_MANIFEST,
-                                      PLUGIN_FETCH_MAVEN_KEY)
+                                      PLUGIN_FETCH_MAVEN_KEY,
+                                      PLUGIN_RESOLVE_REMOTE_SOURCE)
 from atomic_reactor.config import get_cachito_session
 from atomic_reactor.dirs import BuildDir
 from atomic_reactor.plugin import PreBuildPlugin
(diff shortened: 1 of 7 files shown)


chmeliik:
PTAL, this is the final PR for the Cachito-related plugin changes.

The first 2 commits are from #1756

atomic_reactor/plugins/pre_add_image_content_manifest.py
```diff
@@ -100,22 +102,25 @@ def __init__(self, workflow, remote_sources=None, destdir=IMAGE_BUILD_INFO_DIR):
...
         fetch_maven_results = workflow.prebuild_results.get(PLUGIN_FETCH_MAVEN_KEY) or {}
         self.pnc_artifact_ids = fetch_maven_results.get('pnc_artifact_ids') or []
 
-    @property
+    @functools.cached_property
     def icm_file_name(self):
         """Determine the name for the ICM file (name-version-release.json)."""
-        labels = Labels(self.dfp.labels)
+        # parse Dockerfile for any platform, the N-V-R labels should be equal for all platforms
+        dockerfile = self.workflow.build_dir.any_platform.dockerfile_with_parent_env(
```
rcerven, line 107: can we use just 'dockerfile' ? all we need is NVR we don't need parent injection at all, I know it doesn't hurt, but might give impression that something from parent is inherited

tests/plugins/test_add_image_content_manifest.py
```diff
@@ -112,11 +116,11 @@
 )
 
 REMOTE_SOURCES = [{
-    'build_args': None,
-    'configs': None,
-    'request_id': CACHITO_REQUEST_ID,
-    'url': None,
-    'name': None,
+    'id': CACHITO_REQUEST_ID,
+    # 'url': 'some url',
```
rcerven, line 120: these can be removed I'd say, since they aren't used by that plugin, or you can just leave them here for consistency

tkdchen (approved):

mkosiarc (approved):

atomic_reactor/plugins/pre_add_image_content_manifest.py
chmeliik: Do we know that the NVR labels cannot reference parent env vars? The plugin previously used `df_parser(..., workflow=self.workflow)`, which does inject parent environment

tests/plugins/test_add_image_content_manifest.py
chmeliik: yeah, the idea was to use only `id` but keep the others there as comments to make it look consistent with the plugin result

atomic_reactor/plugins/pre_add_image_content_manifest.py
rcerven: mmm they probably could, ok then

rcerven (approved):

PR #1758, opened 2022-01-17
flatpak_update_dockerfile: get composes from resolve_compose result

hjmodi:
We don't need to re-fetch composes from compose IDs again since
in OSBS 2 all plugins will be in the same pipeline.

* CLOUDBLD-8107

### DIFF
--- a/atomic_reactor/plugins/pre_flatpak_update_dockerfile.py
+++ b/atomic_reactor/plugins/pre_flatpak_update_dockerfile.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2017 Red Hat, Inc
+Copyright (c) 2017-2022 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_flatpak_update_dockerfile.py
```diff
@@ -244,18 +204,43 @@ def run(self):
...
 
         # Create the cleanup script
-
-        cleanupscript = os.path.join(self.workflow.df_dir,
+        cleanupscript = os.path.join(build_dir.path,
                                      FLATPAK_CLEANUPSCRIPT_FILENAME)
         with open(cleanupscript, 'w') as f:
             f.write(builder.get_cleanup_script())
         os.chmod(cleanupscript, 0o0500)
+        return [Path(includepkgs_path), Path(cleanupscript)]
```
chmeliik, line 221: You could take full advantage of Path objects like this:
```python
cleanupscript = build_dir.path / FLATPAK_CLEANUPSCRIPT_FILENAME
cleanupscript.write_text(builder.get_cleanup_script())
cleanupscript.chmod(0o0500)
return [includepkgs_path, cleanupscript]
```
(and similarly for the includepkg_path)

Completely optional, I just happen to be a fan of [pathlib](https://docs.python.org/3/library/pathlib.html) :smile:

rcerven (approved):

chmeliik (approved):
LGTM :+1:

PR #1759, opened 2022-01-17
Manage workflow data

tkdchen:
* CLOUDBLD-8271

This PR contains changes to manage workflow data so that it can be shared by different tasks through the build pipeline. The data is written into the context_dir, which is an input from the atomic-reactor command line. This PR covers most of the original design except:

* build_result is kept as usual and written into the workflow.json together with others. The build result should be platform-specific, the build task should handle the details of build result for every platform. The load/dump for build result objects can be implemented based on the changes in PR.

* some data are not removed since corresponding plugins are not updated yet and still reference those data, e.g. `plugins_workspace` and `files`.

* `*_results` attributes are not combined into one `plugin_results`. After looking into more details, I found it probably introduce new problem by merging `*_results`, because every runner object has its own attribute `plugins_results` which references to a `_results` individually. From my point of view at this moment, it is probably not worth to merge and maybe we should reconsider a clearer way to collect results from plugins.


atomic_reactor/dirs.py
```diff
@@ -256,29 +255,75 @@ def for_all_platforms_copy(self, action: FileCreationFunc) -> List[Path]:
...
+class ContextDir(BaseDataDir):
+    """Representing the context directory for the whole container build pipeline."""
+
+    @property
+    def workflow_data_file(self):
+        return self._root_path / "workflow.json"
+
+
+class RootBuildDir(BaseDataDir):
+    """A directory containing all artifacts for building images."""
```
chmeliik, line 293: I still don't see that much value in making `ContextDir` behave the same way as `RootBuildDir`. Will the workflow data management involve the platform-specific behavior? I feel like the workflow data management that we need *currently* will only really touch workflow.json - the per-platform behavior will not be utilized at all until we get to the build-result.json - and even then, each build task only handles one platform, not multiple.

atomic_reactor/dirs.py
```diff
@@ -45,6 +46,7 @@ def __init__(self, path: Path, platform: str) -> None:
         self.path = path
         self.platform = platform
         self.exported_squashed_image: Path = self.path / EXPORTED_SQUASHED_IMAGE_NAME
+        self.build_result_file = self.path / "build_result.json"
```
chmeliik, line 49: This feels wrong - the directory will never have both the `build_result_file` and an `exported_squashed_image`, `dockerfile`, etc. It will only have the `build_result_file` if it is used from a ContextDir, and it will only have the others if it's used from a RootBuildDir

atomic_reactor/inner.py
```diff
@@ -282,8 +284,18 @@ def add_floating_images(self, images):
         for image in images:
             self.add_floating_image(image)
 
+    def load(self, data: Dict[str, Any]) -> None:
+        pass
```
chmeliik, line 288: Maybe load should be a classmethod?
```suggestion
    @classmethod
    def load(cls, data: Dict[str, Any]) -> 'TagConf':
        pass
```
Not sure what your plan for loading is yet, but it seems like it could be difficult to use `load()` as an instance method (how would you create the instance before knowing the data to initialize it with?)

atomic_reactor/inner.py
```diff
@@ -404,6 +445,110 @@ def _update(data):
         return new_data
 
 
+@dataclass
+class ImageBuildWorkflowData(ISerializer):
+    """Manage workflow data.
+
+    Workflow data is those data which are generated values by plugins through
+    the whole build workflow (pipeline) and must be shared in every build tasks.
```
chmeliik, line 453: Nice! :tada: 

I love how much data you were able to remove from DockerBuildWorkflow. It almost looks like a sane class now :smile:

atomic_reactor/inner.py
```diff
@@ -225,7 +308,7 @@ def floating_images(self):
         """
         return self._floating_images
 
-    def add_primary_image(self, image):
+    def add_primary_image(self, image: str) -> None:
         """
         add new primary image
```
rcerven, line 314: can we remove 'used by ....' comments from all methods?

atomic_reactor/inner.py
```diff
@@ -282,8 +365,33 @@ def add_floating_images(self, images):
         for image in images:
             self.add_floating_image(image)
 
+    def add_unique_images(self, images: List[str]) -> None:
```
rcerven, line 368: do we really need both methods (image & images) ? we had now also some *_images methods, which weren't used at all

now you will use it in load but you can just as well loop there, or rename those methods to _...

atomic_reactor/inner.py
```diff
@@ -282,8 +365,33 @@ def add_floating_images(self, images):
...
+    def dump(self) -> Dict[str, Any]:
+        return {
+            "primary_images": [image.to_str() for image in self._primary_images],
+            "unique_images": [image.to_str() for image in self._unique_images],
+            "floating_images": [image.to_str() for image in self.floating_images],
+        }
 
-class Registry(object):
+
+class Registry(ISerializer):
```
rcerven, line 394: will be removed with support of multiple registries

atomic_reactor/inner.py
```diff
@@ -294,6 +402,18 @@ def __init__(self, uri, insecure=False):
...
+            raise ValueError(
+                f"Missing uri from data {data!r} to create a {cls.__name__} object."
+            )
+        return Registry(uri, insecure=data.get("insecure", False))
+
+    def dump(self) -> Dict[str, Any]:
+        return {"uri": self.uri, "insecure": self.insecure}
+
 
 class DockerRegistry(Registry):
```
rcerven, line 418: will be removed with support of multiple registries

atomic_reactor/inner.py
```diff
@@ -303,21 +423,70 @@ def __init__(self, uri, insecure=False):
...
+        result["digests"] = {
+            tag: manifest_digest.dump()
+            for tag, manifest_digest in self.digests.items()
+        }
+        result["config"] = self.config
+        return result
+
 
-class PushConf(object):
+class PushConf(ISerializer):
```
rcerven, line 467: will be removed with support of multiple registries

atomic_reactor/inner.py
```diff
@@ -91,13 +94,18 @@ def decode(self, obj):
         return results
 
 
-class BuildResult(object):
+class BuildResult(ISerializer):
```
rcerven, line 97: so this will be now per arch result or ?

atomic_reactor/inner.py
```diff
@@ -404,6 +597,170 @@ def _update(data):
...
+    Workflow data is those data which are generated values by plugins through
+    the whole build workflow (pipeline) and must be shared in every build tasks.
+
+    These data can be dumped into dictionary object in order to be saved into a
+    file as JSON data, and then be loaded while atomic-reactor launches again to
+    execute another set of plugins for a different build task.
+    """
+
+    dockerfile_images: DockerfileImages = field(default_factory=DockerfileImages)
+    tag_conf: TagConf = field(default_factory=TagConf)
```
rcerven, line 613: since this TagConf will be for the final image (manifest list)

do we need also TagConf in results for arch specific unique images ?

atomic_reactor/inner.py
```diff
@@ -404,6 +597,170 @@ def _update(data):
...
+    push_conf: PushConf = field(default_factory=PushConf)
+
+    prebuild_results: Dict[str, Any] = field(default_factory=dict)
+    buildstep_result: Dict[str, Any] = field(default_factory=dict)
+    postbuild_results: Dict[str, Any] = field(default_factory=dict)
+    prepub_results: Dict[str, Any] = field(default_factory=dict)
+    exit_results: Dict[str, Any] = field(default_factory=dict)
+    plugin_workspace: Dict[str, Any] = field(default_factory=dict)
```
rcerven, line 621: in design is written 'remove plugin_workspace' and here it is again

atomic_reactor/inner.py
```diff
@@ -404,6 +596,168 @@ def _update(data):
...
+    #  Each plugin that transforms the image should save it as a new file and append it to
+    #  the end of exported_image_sequence. Other plugins should then operate with last
+    #  member of this structure. Example:
+    #  [{'path': '/tmp/foo.tar', 'size': 12345678, 'md5sum': '<md5>', 'sha256sum': '<sha256>'}]
+    #  You can use util.get_exported_image_metadata to create a dict to append to this list.
+    exported_image_sequence: List[Dict[str, Union[str, int]]] = field(default_factory=list)
+
+    # mapping of downloaded files; DON'T PUT ANYTHING BIG HERE!
+    # "path/to/file" -> "content"
+    files: Dict[str, str] = field(default_factory=dict)
```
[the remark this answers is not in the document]
rcerven: we don't

atomic_reactor/schemas/workflow_data.json
```diff
@@ -0,0 +1,194 @@
...
+    "koji_source_source_url": {"type": ["string", "null"]},
+    "koji_source_manifest": {"type": ["object", "null"]},
+
+    "buildargs": {"type": "object"},
+    "built_image_inspect": {"type": ["object", "null"]},
+    "layer_sizes": {
+      "type": "array",
+      "items": {"$ref": "#/definitions/layer_size"}
+    },
+    "pulled_base_images": {"type": "array", "uniqueItems": true},
```
MartinBasti, line 42: why `true` is red here?

rcerven: well we don't even need pulled_base_images :)

atomic_reactor/inner.py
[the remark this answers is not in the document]
tkdchen: `files` is still used in `add_yum_repo_by_url` currently. The basic idea of making this PR for manage workflow data is that, part of the work is just to refactor the data attributes from DockerBuildWorkflow to a specific workflow data class. Since part of the attributes, that are planned to be removed, are still being used in plugins, they are not removed from the workflow data in this PR to avoid making the scope bigger.

So, similarly with the plugins_workspace to the flatpak plugins, when the implementation of writing repos into the context_dir as files is done, `files` can be removed from the workflow data.

This consideration also applies to the registry classes mentioned above that will be replaced with arch-specific registries.

atomic_reactor/inner.py
chmeliik: You would have to ask @mkosiarc what exactly his plans are, but IIUC arch-specific tags will be a new addition to the TagConf class and/or the tag_from_config plugin result. Either way, this PR shouldn't have to care about that

atomic_reactor/dirs.py
```diff
@@ -282,3 +282,40 @@ def for_all_platforms_copy(self, action: FileCreationFunc) -> List[Path]:
...
         return the_new_files
+
+
+class ContextDir:
+    """Represents the context directory holding data through the whole image build pipeline."""
+
+    def __init__(self, path: Path):
+        if not path.exists():
+            logger.debug("Context directory %s does not exist yet. Create it.", path)
+            path.mkdir()
```
chmeliik, line 293: shouldn't this have parents=True?

atomic_reactor/inner.py
```diff
@@ -393,7 +393,7 @@ def __eq__(self, other: "DockerRegistry") -> bool:
         )
 
     def __hash__(self) -> int:
-        return hash((self.uri, self.insecure, self.digests, self.config))
+        return hash((self.uri, self.insecure, str(self.digests), str(self.config)))
```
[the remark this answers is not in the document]
ben-alkov: Mark it with a `# TODO` so it's not forgotten?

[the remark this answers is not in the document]
tkdchen: __hash__ is not really necessary for osbs currently, no need to hash any objects. @chmeliik You understand is correct. I modified the commit to make pylint ignore w1641 for now.

atomic_reactor/dirs.py
tkdchen: Updated.

atomic_reactor/types.py
```diff
@@ -5,6 +5,45 @@
...
+    @abstractmethod
+    def load(cls, data: Dict[str, Any]):
+        """Create a concrete object from the given serialized data.
+
+        :param data: load attribute values from this mapping to create a new
+            object in specific type. The keys corresponds to the object attributes.
+            There is no rule to define how keys are mapped to the corresponding
+            object attributes back and forth, which is left to the subclass
+            which implementing this interface.
+            Unknown key will be just ignored and no error is raised for it.
```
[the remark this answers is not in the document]
tkdchen: The validation happens in `load_from_dir` and this is for describing the behavior of `load/dump` interface.

atomic_reactor/schemas/docker_registry.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "atomic-reactor docker registry schema for class DockerRegistry",
+  "type": [
+    "object",
+    "null"
```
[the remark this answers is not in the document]
tkdchen: docker_registry schema is moved to the workflow_data.json.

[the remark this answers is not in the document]
chmeliik: ManifestDigest still has this problem

atomic_reactor/inner.py
```diff
@@ -858,11 +858,19 @@ def reset_dockerfile_images(self, path: str) -> None:
...
+        # e.g. prebuild, now subsequent task starts to run and the saved data
+        # is loaded into the dockerfile_images object. In this case, no need
+        # to update the restored dockerfile_images data.
+        if df_images.is_empty:
+            df_images = self._parse_dockerfile_images(path)
+            self.data.dockerfile_images = df_images
+            self.conf.update_dockerfile_images_from_config(df_images)
+
+        # But, still need to do this
+        self.imageutil.set_dockerfile_images(df_images)
```
chmeliik, line 796: :+1: nice catch

atomic_reactor/inner.py
```diff
@@ -603,9 +603,6 @@ class ImageBuildWorkflowData(ISerializer):
     built_image_inspect: Dict[str, Any] = None
     layer_sizes: List[Dict[str, Union[str, int]]] = field(default_factory=list)
 
-    # list of images pulled during the build, to be deleted after the build
-    pulled_base_images: Set = field(default_factory=set)
```
chmeliik: IIRC there was special handling for `set` objects somewhere in this class, can we also remove that now?

.pylintrc
```diff
@@ -64,7 +64,8 @@ disable=I,
         useless-import-alias,  # nice to have
         useless-super-delegation,  # nice to have
         wrong-import-order,
-        wrong-import-position
+        wrong-import-position,
+        eq-without-hash  # no object with __eq__ defined requires a __hash__ to be hashable
```
chmeliik, line 68: Could you clarify? I'm not sure I understand what you mean by the comment

chmeliik (commented):
Mostly LGTM apart from a few details.

Full disclosure: I didn't check almost any unit tests and am not planning to do so :smile:

chmeliik (approved):
LGTM, I think :smile: 

Will you squash the commits to a more reasonable number before merging?

Nice work btw! :tada:

chmeliik (approved):

rcerven (approved):

rcerven:
rn updated

chmeliik:
Finalized in #1789

PR #1760, opened 2022-01-17
pre_add_filesystem: download/don't stream the filesystem

hjmodi:
Instead of importing the filesystem with docker API we'll now
download filesystem for all arches into arch-specific directories
and replace `FROM koji/image-build` with `FROM scratch\nADD
<filesystem> /`

* CLOUDBLD-8107

### DIFF
--- a/atomic_reactor/plugins/pre_add_filesystem.py
+++ b/atomic_reactor/plugins/pre_add_filesystem.py
@@ -1,10 +1,11 @@
 """
-Copyright (c) 2016 Red Hat, Inc
+Copyright (c) 2016-2022 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
+import functools
 from configparser import ConfigParser
 from io import StringIO
 
@@ -13,17 +14,18 @@ from textwrap import dedent
 import re
 import os
 
+from atomic_reactor.dirs import BuildDir
+
 from atomic_reactor.constants import (DEFAULT_DOWNLOAD_BLOCK_SIZE, PLUGIN_ADD_FILESYSTEM_KEY,
                                       PLUGIN_RESOLVE_COMPOSES_KEY)
 from atomic_reactor.config import get_koji_session
 from atomic_reactor.plugin import PreBuildPlugin, BuildCanceledException
-from atomic_reactor.plugins.exit_remove_built_image import defer_removal
 from atomic_reactor.utils.koji import TaskWatcher, stream_task_output
 from atomic_reactor.utils.yum import YumRepo
-from atomic_reactor.util import get_platforms, df_parser, base_image_is_custom, map_to_user_params
+from atomic_reactor.util import get_platforms, base_image_is_custom, map_to_user_params
 from atomic_reactor.metadata import label_map
 from atomic_reactor import util
-from osbs.utils import Labels, ImageName
+from osbs.utils import Labels
 
 
 @label_map('filesystem-koji-task-id')
@@ -36,9 +38,10 @@ class AddFilesystemPlugin(PreBuildPlugin):
     creating the base image:
     https://docs.pagure.org/koji/image_build/
 
-    Once image build task is complete the tarball is downloaded and
-    it's imported into docker. This creates a new image. The existing
-    FROM instruction value is replaced with the ID of this new image.
+    Once image build task is complete the tarball is downloaded.
+    The existing FROM instruction value is replaced with a
+    FROM scratch and ADD <filesystem> <to_image> for Dockerfile
+    of each platform.
 
     The "FROM" instruction should be in the following format:
         FROM koji/image-build[:image-build-conf]
@@ -72,37 +75,30 @@ class AddFilesystemPlugin(PreBuildPlugin):
         ''')
 
     args_from_user_params = map_to_user_params(
-        "from_task_id:filesystem_koji_task_id",
         "repos:yum_repourls",
-        "architecture:platform",
         "koji_target",
     )
 
-    def __init__(self, workflow, from_task_id=None, poll_interval=5,
-                 blocksize=DEFAULT_DOWNLOAD_BLOCK_SIZE,
-                 repos=None, architecture=None, koji_target=None):
+    def __init__(self, workflow, poll_interval=5, blocksize=DEFAULT_DOWNLOAD_BLOCK_SIZE,
+                 repos=None, koji_target=None):
         """
         :param workflow: DockerBuildWorkflow instance
-        :param from_task_id: int, use existing Koji image task ID
         :param poll_interval: int, seconds between polling Koji while waiting
                               for task completion
-        :param blocksize: int, chunk size for streaming files from koji
+        :param blocksize: int, chunk size for downloading files from koji
         :param repos: list<str>: list of yum repo URLs to be used during
                       base filesystem creation. First value will also
                       be used as install_tree. Only baseurl value is used
                       from each repo file.
-        :param architecture: str, arch to build on (worker)
         :param koji_target: str, koji target name
         """
         # call parent constructor
         super(AddFilesystemPlugin, self).__init__(workflow)
 
-        self.from_task_id = from_task_id
         self.poll_interval = poll_interval
         self.blocksize = blocksize
         self.repos = repos or []
         self.architectures = get_platforms(self.workflow)
-        self.architecture = architecture
         self.scratch = util.is_scratch_build(self.workflow)
         self.koji_target = koji_target
         self.session = None
@@ -165,7 +161,7 @@ class AddFilesystemPlugin(PreBuildPlugin):
 
         :param config: ConfigParser object
         """
-        labels = Labels(df_parser(self.workflow.df_path).labels)
+        labels = Labels(self.workflow.build_dir.any_platform.dockerfile.labels)
         for config_key, label in (
             ('name', Labels.LABEL_TYPE_COMPONENT),
             ('version', Labels.LABEL_TYPE_VERSION),
@@ -192,8 +188,6 @@ class AddFilesystemPlugin(PreBuildPlugin):
 
         if self.architectures:
             config.set('image-build', 'arches', ','.join(self.architectures))
-        elif self.architecture:
-            config.set('image-build', 'arches', self.architecture)
         # else just use what was provided by the user in image-build.conf
 
         config_str = StringIO()
@@ -238,21 +232,9 @@ class AddFilesystemPlugin(PreBuildPlugin):
 
         return image_name, args, {'opts': opts}
 
-    def get_filesystem_regex(self, image_name):
-        prefix = image_name
-        if self.architecture:
-            prefix = '{}.*{}'.format(prefix, self.architecture)
-
-        pattern = (r'{}.*(\.tar|\.tar\.gz|\.tar\.bz2|\.tar\.xz)$'
-                   .format(prefix))
-        filesystem_regex = re.compile(pattern, re.IGNORECASE)
-
-        return filesystem_regex
-
     def build_filesystem(self, image_build_conf):
         # Image build conf file should be in the same folder as Dockerfile
-        build_file_dir = self.workflow.source.get_build_file_path()[1]
-        image_build_conf = os.path.join(build_file_dir, image_build_conf)
+        image_build_conf = self.workflow.build_dir.any_platform.path / image_build_conf
         if not os.path.exists(image_build_conf):
             raise RuntimeError('Image build configuration file not found: {}'
                                .format(image_build_conf))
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -374,6 +330,29 @@ def update_repos_from_composes(self):
...
 
+    def _add_filesystem_to_dockerfile(self, file_name, file_path, build_dir: BuildDir):
+        """
+        Put an ADD instruction into the Dockerfile (to include the filesystem
+        into the container image to be built)
+        """
+        content = 'ADD {0} {1}'.format(file_path, file_name)
+        lines = build_dir.dockerfile.lines
+
+        lines[0] = 'FROM scratch\n'
```
rcerven, line 341: this is wrong, you are assuming that 1st stage is custom image

but you have to take in account multistage builds, WHERE multiple stages in any place can be custom image

so you have to basically find all custom image stages and substitute those

rcerven (changes requested):

atomic_reactor/plugins/pre_add_filesystem.py
chmeliik: Oh, there can be multiple custom base images? :open_mouth:

rcerven: also keep in mind that you are suppose to change just the image for scratch, so this should still work:

FROM koji/image-build AS importedimage

hjmodi: made the requested changes.

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -165,7 +161,7 @@ def update_config_from_dockerfile(self, config):
 
         :param config: ConfigParser object
         """
-        labels = Labels(df_parser(self.workflow.df_path).labels)
+        labels = Labels(self.workflow.build_dir.any_platform.dockerfile.labels)
```
chmeliik, line 164: Hmm, sometimes when parsing labels we inject the parent environment and sometimes we don't.

The content_manifest plugin, for example, uses `dockerfile_with_parent_env`, this just uses  `dockerfile`. And that *is* consistent with how it was done before - this used df_parser without the workflow param, so no parent env :thinking: 

It doesn't seem to matter that much, I'll just file something as potential future improvement

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -238,20 +232,9 @@ def parse_image_build_config(self, config_file_name):
...
-        pattern = (r'{}.*(\.tar|\.tar\.gz|\.tar\.bz2|\.tar\.xz)$'
-                   .format(prefix))
-        filesystem_regex = re.compile(pattern, re.IGNORECASE)
-
-        return filesystem_regex
-
     def build_filesystem(self, image_build_conf):
         # Image build conf file should be in the same folder as Dockerfile
-        build_file_dir = self.workflow.source.get_build_file_path()[1]
+        build_file_dir = self.workflow.build_dir.path
```
chmeliik, line 237:
```suggestion
        build_file_dir = self.workflow.build_dir.any_platform.path
```
There's nothing in the root `build_dir` except the per-platform directories. The Dockerfile and image build config are in the per-platform dirs (kinda confusing, maybe the `RootBuildDir.path` attribute should be private)

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -291,29 +270,21 @@ def download_filesystem(self, task_id, filesystem_regex):
                                .format(filesystem_regex.pattern))
         task_id, file_name = found
 
-        self.log.info('Streaming filesystem: %s from task ID: %s',
+        self.log.info('Downloading filesystem: %s from task ID: %s',
                       file_name, task_id)
 
-        contents = stream_task_output(self.session, task_id, file_name,
-                                      self.blocksize)
+        file_path = self.workflow.build_dir.path / file_name
```
chmeliik, line 276: Needs to be downloaded to the per-platform build dirs instead. This is also related to the fact that the filepath for ADD has to be relative

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -374,6 +330,34 @@ def update_repos_from_composes(self):
             self.log.info('adding repo file from compose: %s', compose_info['result_repofile'])
             self.repos.append(compose_info['result_repofile'])
 
+    def _add_filesystem_to_dockerfile(self, file_path, build_dir: BuildDir):
+        """
+        Put an ADD instruction into the Dockerfile (to include the filesystem
+        into the container image to be built)
+        """
+        content = 'ADD {0} /'.format(file_path)
```
chmeliik, line 338: The path needs to be relative:
```dockerfile
ADD filesystem.tar.gz /
```
The current implementation would use this path, which wouldn't work
```dockerfile
ADD /workspace/build-dir/filesystem.tar.gz /
```

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -374,6 +330,34 @@ def update_repos_from_composes(self):
...
+        lines = build_dir.dockerfile.lines
+
+        custom_image_from = 'FROM koji/image-build'
+        scratch_from = 'FROM scratch'
+
+        for index, line in enumerate(lines):
+            if f'{custom_image_from}\n' not in line and f'{custom_image_from} ' not in line:
+                continue
+            lines[index] = lines[index].replace(custom_image_from, scratch_from)
```
chmeliik, line 347: In the change_from_in_df plugin, we use the `parent_images` property to set parent images. Maybe this plugin could use a similar approach?

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -374,6 +330,34 @@ def update_repos_from_composes(self):
...
+        """
+        content = 'ADD {0} /'.format(file_path)
+        lines = build_dir.dockerfile.lines
+
+        custom_image_from = 'FROM koji/image-build'
+        scratch_from = 'FROM scratch'
+
+        for index, line in enumerate(lines):
+            if f'{custom_image_from}\n' not in line and f'{custom_image_from} ' not in line:
+                continue
```
chmeliik, line 346: I would prefer if we use the `base_image_is_custom()` util method to do this check. IIRC, the image doesn't have to be exactly `koji/image-build`, but can be `koji/image-build:some-tag`.

The util method doesn't work on entire lines of course, but if you work with `parent_images` instead, it should work nicely

atomic_reactor/plugins/pre_add_filesystem.py
hjmodi: ok

atomic_reactor/plugins/pre_add_filesystem.py
hjmodi: updated to use platform-specific directories.

atomic_reactor/plugins/pre_add_filesystem.py
hjmodi: updated to use platform-specific dirs and do relative ADD

atomic_reactor/plugins/pre_add_filesystem.py
hjmodi: updated.

atomic_reactor/plugins/pre_add_filesystem.py
hjmodi: updated.

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -374,6 +329,39 @@ def update_repos_from_composes(self):
...
+        """
+        Put an ADD instruction into the Dockerfile (to include the filesystem
+        into the container image to be built)
+        """
+        content = 'ADD {0} /\n'.format(file_name)
+        lines = build_dir.dockerfile.lines
+
+        for index, line in enumerate(lines):
+            if 'koji/image-build' in line:
+                lines.insert(index+1, content)
```
chmeliik, line 342: I'm still not entirely convinced by this. I could imagine someone putting a comment about "koji/image-build" in their Dockerfile, and then we would end up putting the ADD instruction after that comment.

```Dockerfile
# OSBS will replace koji/image-build with a filesystem tarball downloaded from Brew
FROM koji/image-build
```
Does [DockerfileParser](https://github.com/containerbuildsystem/dockerfile-parse/blob/master/dockerfile_parse/parser.py) provide anything that could be used to do this more reliably? Maybe the `structure` property?

Side note: inserting to a list while enumerating that list actually works as intended? :open_mouth: (it does, but only if you insert *after* the current index)

ben-alkov: Note the side-effect that the _inserted_ line will still be checked by the `if`... probably not dangerous here, unless the filesystem archive name contains 'koji/image-build' :grin:

hjmodi: updated the logic to read from `structure` and insert into `lines` and use `parent_images` to replace `koji/image-build` with `scratch`

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -374,6 +329,42 @@ def update_repos_from_composes(self):
...
+        """
+        content = 'ADD {0} /\n'.format(file_name)
+        lines = build_dir.dockerfile.lines
+        # as we insert elements we have to keep track of the increment for inserting
+        increment = 1
+
+        for item in build_dir.dockerfile.structure:
+            if item['instruction'] == 'FROM' and base_image_is_custom(item['value'].split(' ')[0]):
+                lines.insert(item['endline']+increment, content)
+                increment += item['endline'] - item['startline'] + 1
```
chmeliik, line 345: Why is the increment based on the `endline` and `startline` of the FROM instruction?

If my understanding is correct, the increment is for keeping track of the "offset" between the original structure of the Dockerfile and the updated structure that we are modifying by inserting new content. We insert one line every time, so the offset should increase by 1, shouldn't it? Regardless of the size of the FROM instruction?

atomic_reactor/plugins/pre_add_filesystem.py
```diff
@@ -374,6 +329,42 @@ def update_repos_from_composes(self):
...
+        Put an ADD instruction into the Dockerfile (to include the filesystem
+        into the container image to be built)
+        """
+        content = 'ADD {0} /\n'.format(file_name)
+        lines = build_dir.dockerfile.lines
+        # as we insert elements we have to keep track of the increment for inserting
+        increment = 1
+
+        for item in build_dir.dockerfile.structure:
+            if item['instruction'] == 'FROM' and base_image_is_custom(item['value'].split(' ')[0]):
```
chmeliik, line 343:
```suggestion
            if item['instruction'] == 'FROM' and base_image_is_custom(item['value'].split()[0]):
```
Would suggest to split on any whitespace, this will probably never be relevant but just to be sure

hjmodi: used suggestion

atomic_reactor/plugins/pre_add_filesystem.py
hjmodi: oh yes you're right, I misunderstood my own logic

tests/plugins/test_add_filesystem.py
```diff
@@ -210,45 +209,109 @@ def make_and_store_reactor_config_map(workflow, additional_koji=None):
...
     mock_koji_session(scratch=scratch)
-    mock_image_build_file(source_dir)
+    mock_image_build_file(workflow)
+    image_name = 'fedora-23-1.0.x86_64.tar.gz'
+    expected_dockerfile = expected_dockerfile.replace('ADD {} {}\n',
+                                                      f'ADD {image_name} '
+                                                      '/\n')
 
     make_and_store_reactor_config_map(workflow, {'root_url': '', 'auth': {}})
+    workflow.prebuild_results[PLUGIN_RESOLVE_COMPOSES_KEY] = {'composes': []}
```
chmeliik, line 279: Is this necessary? The plugin would still work if there was no resolve_composes result, right? I don't mind keeping this here, just wanted to be sure that this is not a regression

chmeliik (approved):
LGTM :+1:

Just one note about the unit tests, but assuming the code  *does* work when there is no resolve_composes result, I don't care too much

tests/plugins/test_add_filesystem.py
hjmodi: yes it does work without it, removed it so that it's clear.

hjmodi:
> LGTM +1
> 
> Just one note about the unit tests, but assuming the code _does_ work when there is no resolve_composes result, I don't care too much

updated.

Please also take a look at the maintainer checklist.

chmeliik:
> Please also take a look at the maintainer checklist.

Maybe I'd change the wording from "stream" to "import", since the operation we used to do was `docker import`. Otherwise LGTM

hjmodi:
> > Please also take a look at the maintainer checklist.
> 
> Maybe I'd change the wording from "stream" to "import", since the operation we used to do was `docker import`. Otherwise LGTM

updated.

mkosiarc (approved):

PR #1764, opened 2022-01-18
merge koji, add_yum_repo_by_url and inject_yum_repo in one

hjmodi:
These plugins coordinated yum repo injection for different
platfrom by coordinating between between orchestrator and
worker. In OSBS 2, there is no orchestrator and worker so
everything can be done in one plugin.

* CLOUDBLD-8107


atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,264 @@
...
+            raise ValueError('Errors found while checking yum repo urls: \n{}'
+                             .format('\n'.join(errors)))
+
+    def _final_user_line(self):
+        user = self._find_final_user()
+        if user:
+            return user
+
+        if not self.workflow.dockerfile_images.base_from_scratch:
+            # OSBS2 TBD: decide if we need to inspect a specific platform
```
hjmodi, line 56: not sure what advantage inspecting for specific architecture will offer here since we're already looking up `USER` in platform-specific `Dockerfile` in `self._find_final_user()`

hjmodi: I've reached to the following conclusion after observing another PR. We can inspect any platform since the `User` should be same for all platforms.

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,264 @@
...
+            self.log.info("Setting yum proxy to %s", self.proxy)
+            repo['proxy'] = self.proxy
+
+        yum_repo = YumRepo(os.path.join(YUM_REPOS_DIR, self.target))
+        path = yum_repo.dst_filename
+        self.log.info("yum repo of koji target: '%s'", path)
+        yum_repo.content = render_yum_repo(repo, escape_dollars=False)
+        self.yum_repos.append(yum_repo)
+
+    def __init__(self, workflow, target=None, repourls=None, inject_proxy=None):
```
chmeliik, line 144: I didn't expect a constructor this far down :smile: 

It's just cosmetics, but could this be moved above the other methods?

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,264 @@
...
+            return
+
+        resolve_comp_result = self.workflow.prebuild_results.get(PLUGIN_RESOLVE_COMPOSES_KEY)
+        self.include_koji_repo = resolve_comp_result['include_koji_repo']
+
+        if self.include_koji_repo:
+            self.add_koji_repo()
+        else:
+            self.log.info("'include_koji_repo parameter is set to '%s', not including koji repo",
+                          self.include_koji_repo)
```
chmeliik, line 229: If I'm basing basing this only on how the pre_koji plugin was implemented, the condition would be
```python
if self.include_koji_repo or not self.repos:
```
I don't know if that's necessary though, maybe the behavior is already equivalent? It would make more sense to me if the resolve_composes plugin set the include_koji_repo param accordingly, so that this plugin wouldn't need the extra check.

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,264 @@
...
+        :param repourls: list of str, URLs to the repo files
+        :param inject_proxy: set proxy server for this repo
+        """
+        super().__init__(workflow)
+        self.target = target
+
+        self.xmlrpc = None
+        self.pathinfo = None
+        self.proxy = None
+        self.repourls = repourls or []
```
chmeliik, line 159: repourls currently come from user_params? I think they should come from the resolve_composes plugin result instead (and they are different for each arch, so that may need more changes)

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,264 @@
...
+        if final_user_line:
+            lines.insert(0, "USER root")
+            lines.append(final_user_line)
+
+        return lines
+
+    def add_koji_repo(self):
+        self.xmlrpc = get_koji_session(self.workflow.conf)
+        self.pathinfo = self.workflow.conf.koji_path_info
+        self.proxy = self.workflow.conf.yum_proxy
```
chmeliik, line 91: It feels strange to initialize these attributes here. If they're not needed outside this method, maybe make them local variables? Otherwise, could they be initialized in `__init__`?

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,264 @@
...
+
+        if not self.target:
+            self.log.info('no target provided, skipping plugin')
```
chmeliik, line 94:
```suggestion
            self.log.info('no target provided, not adding koji repo')
```
(the rest of the plugin is not skipped)

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,264 @@
...
+        if not self.workflow.dockerfile_images.base_from_scratch:
+            build_dir.dockerfile.add_lines(*self._cleanup_lines())
+
+    def run(self):
+        """
+        run the plugin
+        """
+        if (self.workflow.dockerfile_images.base_from_scratch and
+                not self.workflow.dockerfile_images):
+            self.log.info("Skipping plugin, from scratch single stage can't add repos")
```
rcerven, line 219: this isn't just scratch single stage

there can be even multiple stages, but all scratch

atomic_reactor/plugins/pre_inject_yum_repos.py
rcerven: resolve_composes plugin already took in account explicit repourls, and reflect that in include_koji_repo

atomic_reactor/plugins/pre_inject_yum_repos.py
rcerven: yeah you should use yum_repourls from resolve_compose plugin

and add all noarch repos to all arches, and arch specific ones just to those arches

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,264 @@
...
+        self.yum_repos = []
+        self.allowed_domains = self.workflow.conf.yum_repo_allowed_domains
+        self.include_koji_repo = False
+        self._builder_ca_bundle = None
+        self._ca_bundle_pem = None
+        self._dockerfile = None
+
+    def _inject_into_repo_files(self, build_dir: BuildDir):
+        """Inject repo files into a relative directory inside the build context"""
+        if build_dir.dockerfile.baseimage is None:
```
rcerven, line 170: you can use just self.workflow.dockerfile_images

or even better in run just check for not self.workflow.dockerfile_images  (which indicates no pullable images => all scratch or not any FROM)

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,264 @@
...
+
+        for repo in self.yum_repos:
+            # Update every repo accordingly in a repofile
+            # input_buf ---- updated ----> updated_buf
+            with StringIO(repo.content.decode()) as input_buf, StringIO() as updated_buf:
+                for line in input_buf:
+                    updated_buf.write(line)
+                    # Apply sslcacert to every repo in a repofile
+                    if line.lstrip().startswith('[') and self._builder_ca_bundle:
+                        updated_buf.write(f'sslcacert=/tmp/{self._ca_bundle_pem}\n')
```
rcerven, line 185: this is weird now, because you will have to copy that file to each arch builder ..... 

cc @chmeliik

chmeliik: I think `/tmp/{self._ca_bundle_pem}` is the path inside the container. We copy the cert from the build-dir to the container with `ADD {self._ca_bundle_pem} /tmp/{self._ca_bundle_pem}`, so that should still work

atomic_reactor/plugins/pre_inject_yum_repos.py
hjmodi: moved up

atomic_reactor/plugins/pre_inject_yum_repos.py
hjmodi: updated to use `yum_repourls` from `resolve_composes` result

atomic_reactor/plugins/pre_inject_yum_repos.py
hjmodi: made them local

atomic_reactor/plugins/pre_inject_yum_repos.py
hjmodi: added this change

atomic_reactor/plugins/pre_inject_yum_repos.py
hjmodi: switched to check for `not self.workflow.dockerfile_images`

atomic_reactor/plugins/pre_inject_yum_repos.py
hjmodi: removed the `RuntimeError` and switched to check for `self.workflow.dockerfile_images`

atomic_reactor/plugins/pre_inject_yum_repos.py
hjmodi: I double-checked it and yes, it is as @chmeliik said. we copy ca_bundle to /tmp inside container. We are also copying it from its original path to platform-specific build dirs and then providing the relative path inside the build dir so it's definitely there when adding

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,265 @@
...
+        if not self.allowed_domains:
+            return
+        errors = []
+
+        for platform in self.platforms:
+            for repourl in self.repourls.get(platform, []):
+                repo_domain = urlparse(repourl).netloc
+                if repo_domain not in self.allowed_domains:
+                    errors.append('Yum repo URL {} is not in list of allowed domains: {}'
+                                  .format(repourl, self.allowed_domains))
```
chmeliik, line 67: We expect some of the urls to be duplicated between platforms, don't we? Maybe it would make sense to de-duplicate them before checking them?

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,265 @@
...
+
+        if self.repourls and not is_scratch_build(self.workflow):
+            self.validate_yum_repo_files_url()
+
+        for platform in self.platforms:
+            for repourl in self.repourls.get(platform, []):
+                yum_repo = YumRepo(repourl)
+                self.log.info("fetching yum repo from '%s'", yum_repo.repourl)
+                try:
+                    yum_repo.fetch()
```
chmeliik, line 237: Same here with the duplication, would be good to fetch each repo just once

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,265 @@
...
+                        updated_buf.write(f'sslcacert=/tmp/{self._ca_bundle_pem}\n')
+
+                yum_repo = YumRepo(repourl=repo.dst_filename,
+                                   content=updated_buf.getvalue(),
+                                   dst_repos_dir=host_repos_path,
+                                   add_hash=False)
+                yum_repo.write_content()
+
+    def _inject_into_dockerfile(self, build_dir: BuildDir):
+        self._dockerfile = build_dir.dockerfile
```
chmeliik, line 192: Let's not initialize it here. And it should be `dockerfile_with_parent_env()`, the USER line may reference a parent variable

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,265 @@
...
+from atomic_reactor.utils.yum import YumRepo
+
+
+class InjectYumReposPlugin(PreBuildPlugin):
+    key = "inject_yum_repos"
+    is_allowed_to_fail = False
+
+    args_from_user_params = map_to_user_params(
+        "target:koji_target",
+        "repourls:yum_repourls",
```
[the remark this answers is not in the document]
hjmodi: removed

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+        run the plugin
+        """
+        if not self.workflow.dockerfile_images:
+            self.log.info("Skipping plugin, from scratch stage(s) can't add repos")
+            return
+
+        resolve_comp_result = self.workflow.prebuild_results.get(PLUGIN_RESOLVE_COMPOSES_KEY)
+        self.include_koji_repo = resolve_comp_result['include_koji_repo']
+        self.repourls = resolve_comp_result['yum_repourls']
```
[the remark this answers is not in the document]
hjmodi: moved to init

atomic_reactor/plugins/pre_inject_yum_repos.py
hjmodi: de-deduplicated

atomic_reactor/plugins/pre_inject_yum_repos.py
hjmodi: de-duplicated

atomic_reactor/plugins/pre_inject_yum_repos.py
hjmodi: changed to use local var and use `dockerfile_with_parent_env()`

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,273 @@
...
+        else:
+            self.log.info("'include_koji_repo parameter is set to '%s', not including koji repo",
+                          self.include_koji_repo)
+
+        if self.repourls and not is_scratch_build(self.workflow):
+            self.validate_yum_repo_files_url()
+
+        for platform in self.platforms:
+            fetched_repourls = set()
+            for repourl in self.repourls.get(platform, []):
```
rcerven, line 238: and what about 'noarch' arch which is for all arches??? :)

hjmodi: it is merged in resolve_composes with all arches.

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,273 @@
...
+            self.log.warning("No repo info was retrieved")
+            return
+
+        # to use urljoin, we would have to append '/', so let's append everything
+        baseurl = pathinfo.repo(repo_info['id'], tag_info['name']) + "/$basearch"
+
+        self.log.info("baseurl = '%s'", baseurl)
+
+        repo = {
+            'name': 'atomic-reactor-koji-plugin-%s' % self.target,
```
rcerven, line 151: since we don't have anymore koji plugin, rename to 'atomic-reactor-koji-target-%s'

rcerven (approved):

atomic_reactor/plugins/pre_inject_yum_repos.py
hjmodi: updated

atomic_reactor/plugins/pre_inject_yum_repos.py
```diff
@@ -0,0 +1,273 @@
...
+            self.add_koji_repo()
+        else:
+            self.log.info("'include_koji_repo parameter is set to '%s', not including koji repo",
+                          self.include_koji_repo)
+
+        if self.repourls and not is_scratch_build(self.workflow):
+            self.validate_yum_repo_files_url()
+
+        for platform in self.platforms:
+            fetched_repourls = set()
```
chmeliik, line 237: This will reset between platforms, which sort of defeats the point of deduplicating :smile:

chmeliik (commented):
LGTM, only one detail

atomic_reactor/plugins/pre_inject_yum_repos.py
hjmodi: good catch! fixed the logic

chmeliik (approved):
LGTM :+1:

PR #1765, opened 2022-01-19
Update the check_user_settings plugin for OSBS 2

chmeliik:
- Parse Dockerfile from build dirs, not source dir
- Update util methods the plugin uses to parse the Dockerfile to do the same as ^
- Move the plugin after check_and_set_platforms to make the changes work
- Update other plugins (mostly their unit tests) affected by the util method changes

### DIFF
--- a/atomic_reactor/plugins/exit_koji_import.py
+++ b/atomic_reactor/plugins/exit_koji_import.py
@@ -61,7 +61,7 @@ from atomic_reactor.constants import (
     KOJI_SOURCE_ENGINE,
 )
 from atomic_reactor.util import (Output,
-                                 df_parser, get_primary_images,
+                                 get_primary_images,
                                  get_floating_images, get_unique_images,
                                  get_manifest_media_type,
                                  get_digests_map_from_annotations, is_scratch_build,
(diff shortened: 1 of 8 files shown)


MartinBasti:
> PTAL.
> 
> Note that none of the changes are strictly speaking necessary, the changed plugins _could_ keep parsing the original Dockerfile from the source dir instead of the modified one in the build dir. But I don't like that idea, if some parts of the code use the original Dockerfile and others use the modified one, that sounds like a recipe for confusion and hidden bugs.
> 
> But the fact that check_user_settings now has to be _third_ plugin is also a little horrifying (cc @MartinBasti). In my mind, it's still better than mixing which Dockerfile we work with, but I'm open to suggestions.

But `check_user_settings` is supposed to check user input, it should really run before OSBS do any changes there (except cloning). IMO it should just check source at that stage.

rcerven:
> > PTAL.
> > Note that none of the changes are strictly speaking necessary, the changed plugins _could_ keep parsing the original Dockerfile from the source dir instead of the modified one in the build dir. But I don't like that idea, if some parts of the code use the original Dockerfile and others use the modified one, that sounds like a recipe for confusion and hidden bugs.
> > But the fact that check_user_settings now has to be _third_ plugin is also a little horrifying (cc @MartinBasti). In my mind, it's still better than mixing which Dockerfile we work with, but I'm open to suggestions.
> 
> But `check_user_settings` is supposed to check user input, it should really run before OSBS do any changes there (except cloning). IMO it should just check source at that stage.

you could also use just Dockerfile from rootbuilddir instead of platform one

chmeliik:
> But `check_user_settings` is supposed to check user input, it should really run before OSBS do any changes there (except cloning). IMO it should just check source at that stage.

It does run before OSBS does any changes. The rhpkg sources plugin should be considered part of the cloning process (well, at least that's the only way I can justify running it before check_and_set_platforms). And I've expressed many times what I think about the check_and_set_platforms plugin. It's not a plugin, it's an integral step of initializing a pre|post|exit task.

I'll think about letting the check_user_settings check the source Dockerfile instead, but it will require some refactoring and possibly changes to 3 more plugins to pay for the sins of the `has_operator_*_manifest` methods

MartinBasti:
Okay, stay with less painful change for now

rcerven:
> Okay, stay with less painful change for now

alright

rcerven (approved):

mkosiarc (approved):

mkosiarc:
LGTM

chmeliik:
RN updated:
- Checking user settings now happens a bit later in the process. Generally, this should make no difference

PR #1767, opened 2022-01-20
Move tag_from_config plugin to prebuild

mkosiarc:
Also update it so it will compute the whole image pullspec
and store it in the TagConf class

CLOUDBLD-8244

### DIFF
--- a/atomic_reactor/plugins/pre_tag_from_config.py
+++ b/atomic_reactor/plugins/pre_tag_from_config.py
@@ -8,13 +8,12 @@ of the BSD license. See the LICENSE file for details.
 
 import functools
 
-from atomic_reactor.plugin import PostBuildPlugin
-from atomic_reactor.constants import INSPECT_CONFIG
-from atomic_reactor.util import df_parser, LabelFormatter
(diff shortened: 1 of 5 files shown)


atomic_reactor/plugins/pre_tag_from_config.py
```diff
@@ -108,7 +112,8 @@ def get_component_name(self):
 
     def run(self):
         self.lookup_labels()
-        return self.parse_and_add_tags()
+        self.parse_and_add_tags()
+        self.add_registry_to_images()
 
     def lookup_labels(self):
         if self.workflow.build_result.is_image_available():
```
chmeliik, line 119: I didn't notice this before - this should probably be removed, now that the plugin runs in prebuild it shouldn't be checking the build result.

And please replace the df_parser call below with the BuildDir equivalent (dockerfile_with_parent_env)

mkosiarc: updated

chmeliik (approved):
LGTM :+1:

hjmodi (approved):

PR #1770, opened 2022-01-24
Refactor bump release plugin

chmeliik:
Various simplifications in preparation for the Tekton changes (which will not be part of this PR).

This should preserve exactly the same functionality as before.

- [n/a] Code coverage from testing does not decrease and new code is covered

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -7,6 +7,8 @@ of the BSD license. See the LICENSE file for details.
 """
 
 import time
+from typing import Optional, Tuple
+
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.util import df_parser
 from osbs.utils import Labels
@@ -248,44 +250,12 @@ class BumpReleasePlugin(PreBuildPlugin):
 
         parser = df_parser(self.workflow.df_path, workflow=self.workflow)
         dockerfile_labels = parser.labels
-        labels = Labels(dockerfile_labels)
-        missing_labels = {}
-        missing_value = 'missing'
-        empty_value = 'empty'
 
-        component_label = labels.get_name(Labels.LABEL_TYPE_COMPONENT)
-        try:
-            component = dockerfile_labels[component_label]
-        except KeyError:
-            self.log.error("%s label: %s", missing_value, component_label)
-            missing_labels[component_label] = missing_value
-
-        version_label = labels.get_name(Labels.LABEL_TYPE_VERSION)
-        try:
-            version = dockerfile_labels[version_label]
-            if not version:
-                self.log.error('%s label: %s', empty_value, version_label)
-                missing_labels[version_label] = empty_value
-        except KeyError:
-            self.log.error('%s label: %s', missing_value, version_label)
-            missing_labels[version_label] = missing_value
-
-        try:
-            release_label, release = labels.get_name_and_value(Labels.LABEL_TYPE_RELEASE)
-        except KeyError:
-            release = None
-        else:
-            if not release:
-                self.log.error('%s label: %s', empty_value, release_label)
-                missing_labels[release_label] = empty_value
-
-        if missing_labels:
-            raise RuntimeError('Required labels are missing or empty or using'
-                               ' undefined variables: {}'.format(missing_labels))
+        component, version, release = self._get_nvr(dockerfile_labels)
 
         # Always set preferred release label - other will be set if old-style
         # label is present
-        release_label = labels.LABEL_NAMES[Labels.LABEL_TYPE_RELEASE][0]
+        release_label = Labels.LABEL_NAMES[Labels.LABEL_TYPE_RELEASE][0]
 
         # Reserve build for isolated builds as well (or any build with supplied release)
         user_provided_release = self.workflow.user_params.get('release')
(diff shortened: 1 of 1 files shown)


atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -274,28 +274,33 @@ def run(self):
...
             )
 
-        # TODO: add the log message about setting release label
+        # Always set preferred release label - other will be set if old-style label is present
+        preferred_release_label = Labels.LABEL_NAMES[Labels.LABEL_TYPE_RELEASE][0]
+
+        if dockerfile_labels.get(preferred_release_label) != release:
+            self.log.info("setting %s=%s", preferred_release_label, release)
+            # Write the label back to the file (this is a property setter)
+            dockerfile_labels[preferred_release_label] = release
```
chmeliik, line 303: Unclear: by removing the "redundant" early return in a previous commit, this is now a change in behavior for scratch builds that have a release label in the Dockerfile. Previously, we wouldn't set the "preferred" release label for those, now we will.

Is there any reason why we wouldn't want to set the preferred release label for scratch builds?

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -319,3 +289,47 @@ def run(self):
...
+
+        release_label = labels.get_name(Labels.LABEL_TYPE_RELEASE)
+        release: Optional[str] = dockerfile_labels.get(release_label)
+
+        missing_labels = {}
+
+        # component: missing X, empty OK
+        if component is None:
+            self.log.error("missing label: %s", component_label)
+            missing_labels[component_label] = "missing"
```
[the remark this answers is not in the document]
rcerven: because it would fail already in osbs-client here : https://github.com/containerbuildsystem/osbs-client/blob/master/osbs/api.py#L111

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -300,25 +274,74 @@ def run(self):
...
+            missing_labels[release_label] = "empty"
+
+        if missing_labels:
+            raise RuntimeError(
+                "Required labels are missing or empty or using undefined variables: {}"
+                .format(missing_labels)
+            )
+
+        # For type-checkers: narrow the type of component and version to str
+        assert component is not None and version is not None
```
rcerven, line 345: do we really now need to add asserts in our code??

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -300,25 +274,74 @@ def run(self):
...
-                    return
-
-        if not release or self.append:
-            self.next_release_general(component, version, release, release_label,
-                                      dockerfile_labels)
+            if not is_scratch_build(self.workflow):
+                self.check_build_existence_for_explicit_release(component, version, release)
+        else:
+            # release not set or release should be appended
+            release = self.next_release_general(component, version, original_release)
```
rcerven, line 290: you created commit about ' bump_release: don't pass null release value ', but you are still passing None

atomic_reactor/plugins/pre_bump_release.py
rcerven: no, it just wasn't necessary since for scratch builds bumping release is just formality not really requirement

atomic_reactor/plugins/pre_bump_release.py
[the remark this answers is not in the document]
chmeliik: Hmm, then my follow-up question would be, why is empty version not OK? :smile: 

I'll just update the code to handle them all the same way (except release, which doesn't have to be set)

atomic_reactor/plugins/pre_bump_release.py
chmeliik: Ack, I'll keep it like this then

atomic_reactor/plugins/pre_bump_release.py
chmeliik: If the code looked like this, we wouldn't:
```python
if not component:
    raise RuntimeError
if not version:
    raise RuntimeError
```
But because the code only saves the errors and raises them later all at once, static analysis wouldn't be able to prove that the types are correct otherwise.

atomic_reactor/plugins/pre_bump_release.py
chmeliik: If `self.append` is True, then original_release may not be None here.

chmeliik:
Updated label check to handle component and version consistently

mkosiarc (approved):

atomic_reactor/plugins/pre_bump_release.py
rcerven: yeah I know it is just one of the cases

atomic_reactor/plugins/pre_bump_release.py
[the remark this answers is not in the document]
rcerven: the o-c checks only existence of 'version' not the emptiness (as it may be using variable), here we are already checking with inheritance so version can't be empty

atomic_reactor/plugins/pre_bump_release.py
chmeliik: Right, in that commit I meant something more like "don't pass a variable which is guaranteed to be None". I couldn't fit that in the commit title, I hope the rest of the commit message explains it

rcerven (approved):

ben-alkov (approved):

tkdchen (approved):

PR #1772, opened 2022-01-25
Fix mypy/annotations in atomic_reactor/util.py

MartinBasti:


ben-alkov (approved):

atomic_reactor/util.py
```diff
@@ -1942,7 +1942,7 @@ def terminal_key_paths(obj: dict) -> Iterator[Sequence]:
 
     From `{'a': {'b': {'c': 1, 'd': 2}}}` yields `('a', 'b', 'c')`, `('a', 'b', 'd')`
     """
-    stack = [
+    stack: List[Tuple[Any, Tuple[str, ...]]] = [
```
chmeliik, line 1948: It would be good for the path type to match the return type of this function (`Tuple[str, ...]` vs. `Sequence`)

atomic_reactor/util.py
```diff
@@ -861,6 +861,8 @@ def get_inspect_for_image(
         """
         all_man_digests = self.get_all_manifests(image)
 
+        config_digest: Optional[str] = None
```
chmeliik, line 864: No need to set a value here, it can just be an annotation

atomic_reactor/util.py
MartinBasti: I can update function definition to `-> Tuple[str, ...]` not opposite. I tried and it doesn't like + operator. But given function indeed returns Tuples, func annotation should be updated

atomic_reactor/util.py
MartinBasti: done

chmeliik (approved):

MartinBasti:
change to annotations in codebase, no release notes needed

PR #1773, opened 2022-01-25
Add/enable mypy

ben-alkov:

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -1,0 +1,19 @@
+[mypy]
+python_version = 3.8
+exclude = .*plugins\/.*
+
+; The following is REQUIRED - `exclude` DOES NOT do what you think it does. It
+; ONLY ignores files encountered while recursively discovering files to check.
+; Without this option, mypy will ALWAYS include 'atomic_reactor/plugins/post_rpmqa.py'
+; Yes it did take me ~1.5 hrs to figure this out.
+[mypy-atomic_reactor.plugins.*]
+ignore_errors = True
+
+[mypy-*.dirs]
+ignore_errors = True
+
+[mypy-*.plugin]
+ignore_errors = True
+
+[mypy-*.inner]
+ignore_errors = True
--- a/.github/workflows/linters.yaml
+++ b/.github/workflows/linters.yaml
@@ -55,6 +55,19 @@ jobs:
       - name: Run markdownlint
         uses: containerbuildsystem/actions/markdownlint@master
 
+  mypy:
+    name: mypy
+    runs-on: ubuntu-latest
+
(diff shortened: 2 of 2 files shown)


ben-alkov:
Tested in [my fork][] (fails due to expected mypy errors)

[my fork]: https://github.com/ben-alkov/atomic-reactor/runs/4943810228?check_suite_focus=true

.mypy.ini
```diff
@@ -0,0 +1,10 @@
+[mypy]
+python_version = 3.8
+exclude = .*plugins\/.*
+
+; The following is REQUIRED - `exclude` DOES NOT do what you think it does. It
+; ONLY ignores files encountered while recursively discovering files to check.
+; Without this option, mypy will ALWAYS include 'atomic_reactor/plugins/post_rpmqa.py'
+; Yes it did take me ~1.5 hrs to figure this out.
```
chmeliik, line 8: Hmm, so what makes atomic_reactor/plugins/post_rpmqa.py special? Does the exclusion work for all plugins except that one?

MartinBasti:
Can you also silence other failing modules too? We cannot merge failing CI. Errors should be fixed incrementally and checks enabled for modules. This is helpful for new code and to prevent regressions.

.mypy.ini
MartinBasti: no idea, but I hit the same

```
mypy atomic_reactor/ --install-types  --ignore-missing-imports --exclude "plugins/" 
...
atomic_reactor/plugins/post_rpmqa.py:24: error: Incompatible types in assignment (expression has type "str", base class "Plugin" defined the type as "None")
```

ben-alkov: > Does the exclusion work for all plugins except that one?

Yes. Based on how I'm *actually* able to ignore it, I'd say it's imported in another module somewhere.

ben-alkov:
> Can you also silence other failing modules too? We cannot merge failing CI. Errors should be fixed incrementally and checks enabled for modules. This is helpful for new code and to prevent regressions.

Sure.

I think, given that we only have hits in 2 files, I'll ignore each file separately; that way, as they are fixed one-by-one, they can be remove from  the config.

.mypy.ini
chmeliik: Ack, that seems plausible :)

MartinBasti:
atomic_reactor/inner.py is failing in CI

ben-alkov:
@MartinBasti, @chmeliik; CI passing, please re-review. smh it's ugly, but it works...

MartinBasti (approved):

lkolacek (approved):

PR #1776, opened 2022-01-27
remove support for multiple registries

hjmodi:
For backword compatibility we'll let the config
be `registries` but `maxItems` is set to `1` and
we're not using `PushConf()` anymore that let us
manipulate registry metadata.

* CLOUDBLD-8654

### DIFF
--- a/atomic_reactor/config.py
+++ b/atomic_reactor/config.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2021 Red Hat, Inc
+Copyright (c) 2021-2022 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
@@ -407,22 +407,19 @@ class Configuration(object):
         return self._get_value(ReactorConfigKeys.REGISTRIES_ORGANIZATION_KEY, fallback=None)
 
     @property
-    def registries(self):
+    def registry(self):
         all_registries = self._get_value(ReactorConfigKeys.REGISTRIES_KEY)
 
-        registries_cm = {}
-        for registry in all_registries:
-            reguri = RegistryURI(registry.get('url'))
-            regdict = {}
-            regdict['version'] = reguri.version
-            if registry.get('auth'):
-                regdict['secret'] = registry['auth']['cfg_path']
-            regdict['insecure'] = registry.get('insecure', False)
-            regdict['expected_media_types'] = registry.get('expected_media_types', [])
+        registry = all_registries[0]
 
-            registries_cm[reguri.docker_uri] = regdict
+        reguri = RegistryURI(registry.get('url'))
+        regdict = {'uri': reguri.docker_uri, 'version': reguri.version}
+        if registry.get('auth'):
+            regdict['secret'] = registry['auth']['cfg_path']
+        regdict['insecure'] = registry.get('insecure', False)
+        regdict['expected_media_types'] = registry.get('expected_media_types', [])
 
-        return registries_cm
+        return regdict
 
     @property
     def docker_registry(self):
--- a/atomic_reactor/plugins/exit_push_floating_tags.py
+++ b/atomic_reactor/plugins/exit_push_floating_tags.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2019 Red Hat, Inc
+Copyright (c) 2019-2022 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
@@ -27,7 +27,7 @@ class PushFloatingTagsPlugin(ExitPlugin):
         :param workflow: DockerBuildWorkflow instance
         """
         super(PushFloatingTagsPlugin, self).__init__(workflow)
-        self.manifest_util = ManifestUtil(workflow, None, self.log)
+        self.manifest_util = ManifestUtil(workflow, self.log)
 
     def add_floating_tags(self, session, manifest_list_data, floating_images):
         list_type = manifest_list_data.get("media_type")
@@ -42,12 +42,7 @@ class PushFloatingTagsPlugin(ExitPlugin):
             self.log.debug("storing %s as %s", target_repo, image.tag)
             self.manifest_util.store_manifest_in_repository(session, manifest, list_type,
                                                             target_repo, target_repo, ref=image.tag)
-        # And store the manifest list in the push_conf
-        push_conf_registry = self.workflow.data.push_conf.add_docker_registry(
-            session.registry, insecure=session.insecure
-        )
-        for image in floating_images:
-            push_conf_registry.digests[image.tag] = manifest_digest
+
         registry_image = get_unique_images(self.workflow)[0]
 
         return registry_image.get_repo(explicit_namespace=False), manifest_digest
@@ -78,8 +73,8 @@ class PushFloatingTagsPlugin(ExitPlugin):
 
         digests = dict()
 
-        for registry in self.manifest_util.registries:
-            session = self.manifest_util.get_registry_session(registry)
-            repo, digest = self.add_floating_tags(session, manifest_data, floating_tags)
-            digests[repo] = digest
+        session = self.manifest_util.get_registry_session()
+        repo, digest = self.add_floating_tags(session, manifest_data, floating_tags)
+        digests[repo] = digest
+
         return digests
--- a/atomic_reactor/plugins/exit_store_metadata.py
+++ b/atomic_reactor/plugins/exit_store_metadata.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2015, 2019 Red Hat, Inc
+Copyright (c) 2015-2022 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
@@ -17,6 +17,7 @@ from atomic_reactor.constants import (PLUGIN_KOJI_UPLOAD_PLUGIN_KEY,
                                       SCRATCH_FROM)
 from atomic_reactor.config import get_openshift_session
 from atomic_reactor.plugin import ExitPlugin
+from atomic_reactor.util import get_manifest_digests
 
 
 class StoreMetadataPlugin(ExitPlugin):
@@ -61,50 +62,37 @@ class StoreMetadataPlugin(ExitPlugin):
         """
 
         digests = {}  # repository -> digest
-        wf_data = self.workflow.data
-        for registry in wf_data.push_conf.docker_registries:
-            for image in wf_data.tag_conf.images:
-                image_str = image.to_str()
-                if image_str in registry.digests:
-                    digest = registry.digests[image_str]
-                    digests[image.to_str(registry=False)] = digest
+        registry = self.workflow.conf.registry
+
+        for image in self.workflow.data.tag_conf.images:
+            image_digests = get_manifest_digests(image, registry['uri'], registry['insecure'],
+                                                 registry.get('secret', None))
+            if image_digests:
+                digests[image.to_str()] = image_digests
 
         return digests
 
-    def _get_registries(self):
-        """
-        Return a list of registries that this build updated
-        """
-        return self.workflow.data.push_conf.all_registries
-
     def get_repositories(self):
-        tag_conf = self.workflow.data.tag_conf
-
         # usually repositories formed from NVR labels
         # these should be used for pulling and layering
         primary_repositories = []
-        for registry in self._get_registries():
-            for image in tag_conf.primary_images:
-                registry_image = image.copy()
-                registry_image.registry = registry.uri
-                primary_repositories.append(registry_image.to_str())
+
+        for image in self.workflow.data.tag_conf.primary_images:
+            primary_repositories.append(image.to_str())
 
         # unique unpredictable repositories
         unique_repositories = []
-        for registry in self._get_registries():
-            for image in tag_conf.unique_images:
-                registry_image = image.copy()
-                registry_image.registry = registry.uri
-                unique_repositories.append(registry_image.to_str())
+
+        for image in self.workflow.data.tag_conf.unique_images:
+            unique_repositories.append(image.to_str())
 
         # floating repositories
         # these should be used for pulling and layering
         floating_repositories = []
-        for registry in self._get_registries():
-            for image in tag_conf.floating_images:
-                registry_image = image.copy()
-                registry_image.registry = registry.uri
-                floating_repositories.append(registry_image.to_str())
+
+        for image in self.workflow.data.tag_conf.floating_images:
+            floating_repositories.append(image.to_str())
+
         return {
             "primary": primary_repositories,
             "unique": unique_repositories,
@@ -114,22 +102,21 @@ class StoreMetadataPlugin(ExitPlugin):
     def get_pullspecs(self, digests):
         # v2 registry digests
         pullspecs = []
-        images = self.workflow.data.tag_conf.images
-        for registry in self._get_registries():
-            for image in images:
-                image_str = image.to_str()
-                if image_str in digests:
-                    digest = digests[image_str]
-                    for digest_version in digest.content_type:
-                        if digest_version not in digest:
-                            continue
-                        pullspecs.append({
-                            "registry": registry.uri,
-                            "repository": image.to_str(registry=False, tag=False),
-                            "tag": image.tag,
-                            "digest": digest[digest_version],
-                            "version": digest_version
-                        })
+
+        for image in self.workflow.data.tag_conf.images:
+            image_str = image.to_str()
+            if image_str in digests:
+                digest = digests[image_str]
+                for digest_version in digest.content_type:
+                    if digest_version not in digest:
+                        continue
+                    pullspecs.append({
+                        "registry": image.registry,
+                        "repository": image.to_str(registry=False, tag=False),
+                        "tag": image.tag,
+                        "digest": digest[digest_version],
+                        "version": digest_version
+                    })
 
         return pullspecs
 
--- a/atomic_reactor/plugins/exit_verify_media_types.py
+++ b/atomic_reactor/plugins/exit_verify_media_types.py
@@ -1,4 +1,4 @@
-"""Copyright (c) 2018, 2019 Red Hat, Inc
+"""Copyright (c) 2018-2022 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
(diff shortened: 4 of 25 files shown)


atomic_reactor/plugins/exit_store_metadata.py
```diff
@@ -61,46 +61,40 @@ def get_digests(self):
         """
 
         digests = {}  # repository -> digest
-        for registry in self.workflow.push_conf.docker_registries:
+        for registry, registry_conf in self.workflow.conf.registries.items():
```
ben-alkov, line 64: Is it the case that 'registry' is unused here?

I.e. `for _, registry_conf in self.workflow.conf.registries.items():` would also work?

atomic_reactor/utils/koji.py
```diff
@@ -581,11 +582,11 @@ def get_repositories_and_digests(workflow, pullspec_image):
     """
     digests = {}  # image -> digests
     typed_digests = {}  # media_type -> digests
-    for registry in workflow.push_conf.docker_registries:
+    for registry, registry_conf in workflow.conf.registries.items():
```
ben-alkov, line 585: Same here as above: would `for _, registry_conf in workflow.conf.registries.items():` also work?

atomic_reactor/plugins/exit_store_metadata.py
hjmodi: updated

atomic_reactor/utils/koji.py
hjmodi: updated

atomic_reactor/plugins/exit_store_metadata.py
```diff
@@ -61,46 +61,40 @@ def get_digests(self):
         """
 
         digests = {}  # repository -> digest
-        for registry in self.workflow.push_conf.docker_registries:
+        for _, registry_conf in self.workflow.conf.registries.items():
```
mkosiarc, line 64: this might still give the impression that we should iterate through registries, but this will run only once, right?

mkosiarc: that brings me to another thought. We decided to keep the "registries" part in the reactor config map to allow backward compatibility, but in code we don't need to reference "registries" and we can change it, or no? If we could change it to single "registry" we would not even need to keep using dict with just 1 key.

hjmodi: yes because we're only allowing config for one registry

hjmodi: hmm, I agree as well. I have updated it to be a config dict only instead of uri to dict mapping.

atomic_reactor/config.py
```diff
@@ -407,22 +408,28 @@ def registries_organization(self):
...
+        registry = all_registries[0]
+
+        reguri = RegistryURI(registry.get('url'))
+        regdict = {'uri': reguri.docker_uri, 'version': reguri.version}
+        if registry.get('auth'):
+            regdict['secret'] = registry['auth']['cfg_path']
+        regdict['insecure'] = registry.get('insecure', False)
+        regdict['expected_media_types'] = registry.get('expected_media_types', [])
+        regdict['digests'] = {}
+        regdict['config'] = None
```
chmeliik, line 428: Why do we need digests and config? The original property didn't have them

atomic_reactor/schemas/config.json
```diff
@@ -453,6 +453,7 @@
                 "expected_media_types": {"$ref": "#/definitions/media_types"}
             },
             "additionalProperties": false,
+            "maxItems": 1,
```
chmeliik, line 456: Would it be reasonable to add `minItems: 1`? I don't think atomic-reactor can work without registries, at least not in OSBS 2

atomic_reactor/config.py
```diff
@@ -407,22 +408,28 @@ def registries_organization(self):
         return self._get_value(ReactorConfigKeys.REGISTRIES_ORGANIZATION_KEY, fallback=None)
 
     @property
-    def registries(self):
-        all_registries = self._get_value(ReactorConfigKeys.REGISTRIES_KEY)
+    def registry(self):
+        if self._registry:
+            return self._registry
```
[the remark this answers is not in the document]
hjmodi: it's so that we can save changes to `digests` and `config`

atomic_reactor/config.py
hjmodi: it was part `DockerRegistry` class. which was used as a member of `PushConf` which would store the registry, their digests and config.

atomic_reactor/schemas/config.json
hjmodi: hmm, I agree.

atomic_reactor/config.py
chmeliik: In that case we should not remove `DockerRegistry` and `PushConf`. I think the idea behind this change was that nothing in those classes would be needed anymore, including digests and config.

@rcerven is that not the case?

atomic_reactor/plugins/exit_store_metadata.py
```diff
@@ -61,46 +61,40 @@ def get_digests(self):
         """
 
         digests = {}  # repository -> digest
-        for registry in self.workflow.push_conf.docker_registries:
+        if self.workflow.conf.registry:
```
mkosiarc, line 64: Do we need these types of conditions?

atomic_reactor/plugins/pre_tag_from_config.py
```diff
@@ -91,7 +91,7 @@ def parse_and_add_tags(self):
 
     def add_registry_to_images(self):
         for image in self.workflow.tag_conf.images:
-            image.registry = next(iter(self.workflow.conf.registries))
+            image.registry = next(iter(self.workflow.conf.registry))
```
mkosiarc, line 94: I believe this should now be` self.workflow.conf.registry['uri']`

atomic_reactor/config.py
rcerven: you shouldn't store anymore digests & config, just query it directly via registry api when needed, it is used just in 2 koji plugins

rcerven: this registry should be really just registry itself and not for storing data for images ....

atomic_reactor/plugins/exit_push_floating_tags.py
```diff
@@ -42,11 +42,9 @@ def add_floating_tags(self, session, manifest_list_data, floating_images):
             self.log.debug("storing %s as %s", target_repo, image.tag)
             self.manifest_util.store_manifest_in_repository(session, manifest, list_type,
                                                             target_repo, target_repo, ref=image.tag)
-        # And store the manifest list in the push_conf
-        push_conf_registry = self.workflow.push_conf.add_docker_registry(session.registry,
-                                                                         insecure=session.insecure)
+
         for image in floating_images:
-            push_conf_registry.digests[image.tag] = manifest_digest
+            self.workflow.conf.registry['digests'][image.tag] = manifest_digest
```
rcerven, line 47: don't store it in registry

atomic_reactor/plugins/exit_push_floating_tags.py
```diff
@@ -42,11 +42,9 @@ def add_floating_tags(self, session, manifest_list_data, floating_images):
...
         registry_image = get_unique_images(self.workflow)[0]
 
         return registry_image.get_repo(explicit_namespace=False), manifest_digest
```
rcerven, line 50: it would be probably more useful if this returned registry_image.to_str and whole plugin would return dict with that as key and digests as values

atomic_reactor/plugins/pre_tag_from_config.py
hjmodi: updated

atomic_reactor/plugins/exit_push_floating_tags.py
hjmodi: I have currently just changed it to query the registry for digest

atomic_reactor/plugins/exit_store_metadata.py
hjmodi: not really, it was only there for keeping a few edge cases working where registry wasn't defined but now with 'minItems' set we don't need it. removed

atomic_reactor/config.py
hjmodi: changed it to query the registry directly instead of storing it.

atomic_reactor/config.py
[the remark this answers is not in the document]
hjmodi: removed it since we're not storing anymore.

atomic_reactor/plugins/exit_store_metadata.py
```diff
@@ -61,50 +62,45 @@ def get_digests(self):
...
         primary_repositories = []
-        for registry in self._get_registries():
-            for image in tag_conf.primary_images:
-                registry_image = image.copy()
-                registry_image.registry = registry.uri
-                primary_repositories.append(registry_image.to_str())
+
+        for image in self.workflow.data.tag_conf.primary_images:
+            registry_image = image.copy()
+            registry_image.registry = self.workflow.conf.registry['uri']
```
mkosiarc, line 84: I think that these two lines are not necessary, because at this point (exit plugin) all images should have the information for registry filled (It happens in pre_tag_from_config plugin)

mkosiarc: The same applies for other occurrences in this PR

hjmodi: removed it from here and everywhere else.

hjmodi:
The uncovered parts were previously uncovered as well but due to change in indentation it shows as uncovered.

atomic_reactor/plugins/exit_verify_media_types.py
```diff
@@ -36,56 +36,55 @@ def run(self):
...
+        expected_media_types = set(registry.get('expected_media_types', []))
+        media_types = set()
+
+        if expect_list_only:
+            expected_media_types = {MEDIA_TYPE_DOCKER_V2_MANIFEST_LIST}
+
+        media_in_registry[registry['uri']] = {'expected': expected_media_types}
+
+        pullspec = image.copy()
+        pullspec.registry = registry['uri']
```
mkosiarc, line 52: should not even be needed here

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -148,86 +148,82 @@ def run(self):
...
+                # Only handle the case when size is set > 0 in config
+                if config_image_size and image_size > config_image_size:
+                    raise ExceedsImageSizeError(
+                        'The size {} of image {} exceeds the limitation {} '
+                        'configured in reactor config.'
+                        .format(image_size, image, image_size_limit)
+                    )
+
+            registry_image = image.copy()
+            registry_image.registry = self.registry['uri']
```
mkosiarc, line 172: here as well

atomic_reactor/plugins/exit_verify_media_types.py
hjmodi: removed

atomic_reactor/plugins/post_tag_and_push.py
hjmodi: this one is a bit confusing. the registry is being set to `None` at https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/plugins/post_tag_and_push.py#L128 so I'm not sure if removing it here is the best idea

atomic_reactor/config.py
```diff
@@ -407,22 +407,19 @@ def registries_organization(self):
         return self._get_value(ReactorConfigKeys.REGISTRIES_ORGANIZATION_KEY, fallback=None)
 
     @property
-    def registries(self):
-        all_registries = self._get_value(ReactorConfigKeys.REGISTRIES_KEY)
+    def registry(self):
+        all_registries = list(self._get_value(ReactorConfigKeys.REGISTRIES_KEY))
```
rcerven, line 411: why list? it is already array in schema

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -148,86 +148,82 @@ def run(self):
...
-                        config_manifest_digest = digests.oci
-                        config_manifest_type = 'oci'
-                    config_registry_image = registry_image
+        insecure = self.registry.get('insecure', False)
+
+        docker_push_secret = self.registry.get('secret', None)
+        self.log.info("Registry %s secret %s", self.registry['uri'], docker_push_secret)
+
+        for image in wf_data.tag_conf.images:
+            if image.registry:
```
rcerven, line 157: pre_tag_from_config is already adding registry

atomic_reactor/plugins/post_tag_and_push.py
rcerven: 'here as well' what?

rcerven: you should just add registry to those additional tags 

        if not tag_conf.unique_images:
            if source_docker_archive:
                tag_conf.add_unique_image(source_unique_image)
            else:
                tag_conf.add_unique_image(self.workflow.image)

as the rest will have registry already from pre_tag_from_config

rcerven: also additional idea is to have new method in TagConf 'add_registry'
and before you use tag_conf for reading you will just force 'add' registry there in bulk to be sure ...

rcerven: or even better you could just call that add_registry once, to set registry for whole TagConf

and when you call images/floating_images etc, it will already return those with added registry

that way you won't have to add it in plugins

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -148,86 +148,82 @@ def run(self):
...
+                    config_manifest_type = 'v2'
+                else:
+                    config_manifest_digest = digests.oci
+                    config_manifest_type = 'oci'
+                config_registry_image = registry_image
 
             if config_manifest_digest:
-                push_conf_registry.config = get_config_from_registry(
-                    config_registry_image, registry, config_manifest_digest, insecure,
+                self.registry['config'] = get_config_from_registry(
```
rcerven, line 225: we don't need this anymore, as you will query for config when needed

also self.registry is just plugin variable adding config there won't change anything as self.workflow.conf.registry
is returning always new dict

atomic_reactor/plugins/post_tag_and_push.py
hjmodi: registry is already there but it is being removed for some reason and then added back to the copy of the image. That's what is confusing me, if it's being done for a particular reason or not

atomic_reactor/config.py
hjmodi: removed type cast

atomic_reactor/plugins/post_tag_and_push.py
hjmodi: removed the check and raise

atomic_reactor/plugins/post_tag_and_push.py
hjmodi: it was a remnant of previous revision. removed it.

atomic_reactor/plugins/post_tag_and_push.py
rcerven: it is being removed just for source image, which name is generated from that binary build, and in koji there can be some old registry, so that is why it is removed

atomic_reactor/plugins/post_tag_and_push.py
```diff
@@ -143,94 +141,56 @@ def run(self):
...
-                                                   insecure, docker_push_secret)
-
-                    if (not (digests.v2 or digests.oci) and (retry < max_retries)):
-                        sleep_time = DOCKER_PUSH_BACKOFF_FACTOR * (2 ** retry)
-                        self.log.info("Retrying push because V2 schema 2 or "
-                                      "OCI manifest not found in %is", sleep_time)
-
-                        time.sleep(sleep_time)
-                    else:
-                        break
```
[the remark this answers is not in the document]
hjmodi: hmm, I missed the retry logic. Added it back

atomic_reactor/plugins/exit_store_metadata.py
```diff
@@ -114,22 +104,21 @@ def get_repositories(self):
...
+
+        for image in self.workflow.data.tag_conf.images:
+            image_str = image.to_str()
+            if image_str in digests:
+                digest = digests[image_str]
+                for digest_version in digest.content_type:
+                    if digest_version not in digest:
+                        continue
+                    pullspecs.append({
+                        "registry": self.workflow.conf.registry['uri'],
```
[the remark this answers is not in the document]
hjmodi: updated.

atomic_reactor/plugins/exit_store_metadata.py
```diff
@@ -61,50 +62,39 @@ def get_digests(self):
...
+        registry = self.workflow.conf.registry
 
-        return digests
+        for image in self.workflow.data.tag_conf.images:
+            image_str = image.to_str()
+            image_digests = get_manifest_digests(image, registry['uri'], registry['insecure'],
+                                                 registry.get('secret', None))
+            if image_str in image_digests:
+                digest = image_digests[image_str]
+                digests[image.to_str()] = digest
```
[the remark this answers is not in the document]
hjmodi: fixed it.

atomic_reactor/plugins/exit_push_floating_tags.py
```diff
@@ -78,8 +73,8 @@ def run(self):
 
         digests = dict()
 
-        for registry in self.manifest_util.registries:
-            session = self.manifest_util.get_registry_session(registry)
+        if self.manifest_util.registry:
+            session = self.manifest_util.get_registry_session(self.manifest_util.registry['uri'])
```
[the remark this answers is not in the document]
hjmodi: removed `registry` parameter

atomic_reactor/plugins/post_tag_and_push.py
hjmodi: updated it to set the registry only for source image, for binary images it is left as as. let me know if it's good

tkdchen (approved):

atomic_reactor/utils/koji.py
```diff
@@ -576,39 +589,38 @@ def get_repositories_and_digests(workflow, pullspec_image):
...
+        if image_digests:
+            digest_list = [select_digest(image_digests)]
+            digests[image.to_str(registry=False)] = digest_list
+            for digest_version in image_digests.content_type:
+                if digest_version == 'v1':
+                    continue
+                if digest_version not in image_digests:
+                    continue
+                digest_type = get_manifest_media_type(digest_version)
+                typed_digests[digest_type] = image_digests[digest_version]
```
chmeliik, line 612: Do we actually need to get the manifest digests for all the images in tag_conf? Couldn't we just get them for the pullspec_image?

This whole method makes very little sense to me, but I *think* the only reason for processing every possible registry+image combination was that the same image in two different registries may have different media types. I don't think we need to do it anymore.

atomic_reactor/utils/koji.py
```diff
@@ -576,39 +589,38 @@ def get_repositories_and_digests(workflow, pullspec_image):
...
+    repositories.append(pullspec)
 
-        digest_list = digests.get(image.to_str(registry=False), ())
-        for digest in digest_list:
-            digest_pullspec = image.to_str(tag=False) + "@" + digest
-            repositories.append(digest_pullspec)
+    digest_list = digests.get(image.to_str(registry=False), ())
+    for digest in digest_list:
+        digest_pullspec = image.to_str(tag=False) + "@" + digest
+        repositories.append(digest_pullspec)
```
chmeliik, line 624: There's no need to `copy()` the pullspec_image anymore (or even have the `image` variable at all)

tests/plugins/test_koji_import.py
```diff
@@ -333,20 +329,66 @@ def mock_environment(workflow, source_dir: Path, session=None, name=None,
...
+    for image in tag_conf.images:
+        tag = image.to_str(registry=False)
+        if oci:
+            digest = ManifestDigest(v1='sha256:not-used',
+                                    oci=fake_digest(image))
+        else:
+            digest = ManifestDigest(v1='sha256:not-used',
+                                    v2=fake_digest(image))
+        if source_build:
+            digest[tag] = 'tag'
```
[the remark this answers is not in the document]
hjmodi: it was left over from when I was trying to mock the digests. Removed it.

tests/test_config.py
```diff
@@ -576,22 +576,23 @@ def test_get_methods(self, parse_from, method, tmpdir, caplog):
...
+            expected = yaml.safe_load(REACTOR_CONFIG_MAP)['registries']
+        else:
+            expected = yaml.safe_load(REACTOR_CONFIG_MAP)[method]
 
-        if method == 'registries':
-            registries_cm = {}
+        if method == 'registry':
+            regdict = {}
             for registry in expected:
                 reguri = RegistryURI(registry.get('url'))
```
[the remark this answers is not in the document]
hjmodi: removed the loop

tests/test_inner.py
```diff
@@ -1589,20 +1408,18 @@ def test_load_from_directory(self, tmpdir):
...
 
-        docker_reg = wf_data.push_conf.all_registries[0]
-        expected_manifest_digest = {
-            "v1": None, "v2": "v2-digest", "v2_list": None, "oci": "oci-digest", "oci_index": None
-        }
-        assert expected_manifest_digest == docker_reg.digests["tag1"]
+        # expected_manifest_digest = {
+        #     "v1": None, "v2": "v2-digest", "v2_list": None, "oci": "oci-digest", "oci_index": None
+        # }
+        # assert expected_manifest_digest == docker_reg.digests["tag1"]
```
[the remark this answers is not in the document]
hjmodi: removed it

atomic_reactor/utils/koji.py
hjmodi: changed it to only pull for pullspec_image

atomic_reactor/utils/koji.py
hjmodi: removed it

atomic_reactor/utils/koji.py
```diff
@@ -576,39 +589,36 @@ def get_repositories_and_digests(workflow, pullspec_image):
...
+    repositories.append(pullspec)
 
-        digest_list = digests.get(image.to_str(registry=False), ())
-        for digest in digest_list:
-            digest_pullspec = image.to_str(tag=False) + "@" + digest
-            repositories.append(digest_pullspec)
+    digest_list = digests.get(pullspec_image.to_str(registry=False), ())
+    for digest in digest_list:
+        digest_pullspec = pullspec_image.to_str(tag=False) + "@" + digest
+        repositories.append(digest_pullspec)
```
chmeliik, line 622: Still can't help but feel that this method is an overly complicated description of something that is now fairly simple

Wouldn't this be equivalent now? (replaces the entire body of this method)
```python
digests = get_manifest_digests(...)
digest_pullspec = f"{pullspec_image.to_str(tag=False)}@{select_digest(digests)}"
repositories = [pullspec_image.to_str(), digest_pullspec]

typed_digests = {
    get_manifest_media_type(version): digest
    for version, digest in digests.items()
    if version != "v1"
}
return repositories, typed_digests
```
This assumes that the digests returned by `get_manifest_digests` cannot be empty, but I think that's true

atomic_reactor/utils/koji.py
```diff
@@ -576,39 +589,36 @@ def get_repositories_and_digests(workflow, pullspec_image):
     """
     Returns a map of images to their repositories and a map of media types to each digest
 
-    it creates a map of images to digests, which is need to create the image->repository
+    it creates a map of images to digests, which is needed to create the image->repository
     map and uses the same loop structure as media_types->digest, but the image->digest
     map isn't needed after we have the image->repository map and can be discarded.
```
chmeliik, line 594: Docstring seems outdated now

atomic_reactor/utils/koji.py
hjmodi: thanks for the suggestion. I also realized, this function is only called from one another function and it also calls `get_manifest_digests` again. So I removed the function and made the code inline.

chmeliik: Nice!

chmeliik (approved):
LGTM :+1:

A small update to the docs may be good, to mention that the configuration must have exactly one registry https://osbs.readthedocs.io/en/latest/build_parameters.html?highlight=registries#reactor-configuration

mkosiarc (approved):

PR #1778, opened 2022-01-28
mypy: fix dirs implementation

MartinBasti:
* `_get_env_from_inspection` must always return something or fail
* None type is not indexable, platforms must be always defined as list

Signed-off-by: Martin Basti <mbasti@redhat.com>

### DIFF
--- a/atomic_reactor/dirs.py
+++ b/atomic_reactor/dirs.py
@@ -93,6 +93,7 @@ class BuildDir(object):
             return envs
         if isinstance(envs, list):
             return dict(item.split("=", 1) for item in envs)
+        raise TypeError(f"Unexpected envs type: {type(envs)}; {envs!r}")
 
     def dockerfile_with_parent_env(self, parent_inspect: ImageInspectionData) -> DockerfileParser:
         """Get the parsed Dockerfile with parent information injected.
@@ -129,7 +130,7 @@ class RootBuildDir(object):
         if not path.exists():
             raise FileNotFoundError(f"Path {path} does not exist.")
         self.path = path
-        self.platforms: Optional[List[str]] = None
+        self.platforms: List[str] = []
 
     @property
     def source_container_sources_dir(self) -> Path:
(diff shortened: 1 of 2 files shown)


ben-alkov (approved):

atomic_reactor/dirs.py
```diff
@@ -93,6 +93,7 @@ def _get_env_from_inspection(data: ImageInspectionData) -> Optional[Dict[str, st
             return envs
         if isinstance(envs, list):
             return dict(item.split("=", 1) for item in envs)
+        assert False, f"Unexpected envs: {envs!r}"
```
chmeliik, line 96: Maybe this should be a proper error? I would generally expect asserts to be unreachable code

ben-alkov: True (and they're removed altogether with the `-O`and `-OO` options, and if $PYTHONOPTIMIZE is set... not that we use any of that? But it's a kind of booby trap which could bite us in the future)

My understanding of best practices for Python assertions is "NEVER use them in production code - use Exceptions instead"

chmeliik: I think they're OK for asserting something that cannot actually happen at runtime, especially when needed for `mypy` analysis

E.g.
```python
if some_value is None:
    errors.append("some error")
...
if errors:
    raise RuntimeError(...)

assert some_value is not None
```
But here, that is not the case, `ImageInspectionData` is just an alias for `Dict[str, Any]`, so envs can technically be anything

MartinBasti: I still see this more like development error than runtime error.

IMO envs can be only list or dicts of strings

MartinBasti: I replaced it with exception to be 100% sure, PTAL

chmeliik (approved):
LGTM :+1:

MartinBasti:
no release notes needed, just mypy cosmetic updates

PR #1783, opened 2022-01-31
WIP - Resource manager tests

tveverka-RH:

### DIFF
--- a/tests/utils/test_resource_manager.py
+++ b/tests/utils/test_resource_manager.py
@@ -10,16 +10,34 @@ from unittest import mock
 from atomic_reactor.utils.resource_manager import Host, ResourceManager
 
 
-def test_host_init():
-    my_host = Host('user', 'hostname')
-    assert my_host.username == 'user'
-    assert my_host.hostname == 'hostname'
-    assert my_host.slots == []
-    assert my_host.free_slots == []
+class TestHost:
+    def test_host_init():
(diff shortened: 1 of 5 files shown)


tests/utils/test_resource_manager.py
```diff
@@ -0,0 +1,25 @@
+"""
+Copyright (c) 2022 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+import pytest
+from unittest import mock
```
MartinBasti, line 9: flexmock?

tests/utils/test_resource_manager.py
```diff
@@ -0,0 +1,25 @@
...
+from atomic_reactor.utils.resource_manager import Host, ResourceManager
+
+
+def test_host_init():
```
MartinBasti, line 13: I'd suggest to add into classes here

```
class TestHost:

  def test_host_init(self):
 ...
  def test_host_read_info_data(self)
...
```

tveverka-RH: Fixed

tests/utils/test_resource_manager.py
tveverka-RH: is it issue? I'm not so familiar with it - I'm trying to change it to fm but it's not so intuitive I would say

PR #1786, opened 2022-02-01
Update add_dockerfile plugin for Tekton

chmeliik:
CLOUDBLD-8123

For each platform, add the platform-specific Dockerfile to the build.

Improve tests to also check the content of the copied Dockerfile when
use_final_dockerfile is set to False.

Note that there is a slight behavior change when not using the final
Dockerfile. Previously, the plugin would copy the current snapshot
during initialization. Now, it will do so in the run() method instead.
In practice, this makes no difference because of how PluginRunners work.

### DIFF
--- a/atomic_reactor/plugins/pre_add_dockerfile.py
+++ b/atomic_reactor/plugins/pre_add_dockerfile.py
@@ -38,10 +38,12 @@ or
 
 import os
 import shutil
-from atomic_reactor.constants import DOCKERFILE_FILENAME, IMAGE_BUILD_INFO_DIR
-from atomic_reactor.plugin import PreBuildPlugin
-from atomic_reactor.util import df_parser
+
 from osbs.utils import Labels
+
+from atomic_reactor.constants import DOCKERFILE_FILENAME, IMAGE_BUILD_INFO_DIR
+from atomic_reactor.dirs import BuildDir
+from atomic_reactor.plugin import PreBuildPlugin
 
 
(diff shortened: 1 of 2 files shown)


mkosiarc (approved):

rcerven (approved):

atomic_reactor/plugins/pre_add_dockerfile.py
```diff
@@ -67,45 +69,47 @@ def __init__(self, workflow, nvr=None, destdir=IMAGE_BUILD_INFO_DIR,
...
-
-    def run(self):
-        """
-        run the plugin
-        """
-        dockerfile = df_parser(self.workflow.df_path, workflow=self.workflow)
+            nvr = self._nvr_from_dockerfile()
+
+        self.df_name = f"{DOCKERFILE_FILENAME}-{nvr}"
+        self.df_path = os.path.join(destdir, self.df_name)
```
ben-alkov, line 75: Pathlib? Or is it not worth the trouble here?

chmeliik: I'd say no, this one represents the destination path inside the container. I considered changing this to `f"{destdir}/{self.df_name}"`, since it is not an os-dependent path at all, but I decided to keep it as is since we handle in-container paths as os paths in others places as well

chmeliik:
Rebased for mypy check

PR #1788, opened 2022-02-01
Update remote host resources manager

lkolacek:

CLOUDBLD-8553

### DIFF
--- a/atomic_reactor/utils/resource_manager.py
+++ b/atomic_reactor/utils/resource_manager.py
@@ -6,193 +6,309 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
 
-import os
-import logging
-import json
-import paramiko
 import backoff
+import json
 import jsonschema
-from typing import List, Any, Dict, Tuple
+import logging
+import paramiko
+import random
+import os
+from contextlib import contextmanager
 from datetime import datetime
 from shlex import quote
-
-PARAMIKO_MAX_RETRIES = 3
-PARAMIKO_BACKOFF_FACTOR = 1
-
-ROOT_PATH = 'resource_manager/'
+from typing import List
 
 RESOURCE_LOCK_SCHEMA = 'schemas/resource-lock-content.json'
+SSH_COMMAND_TIMEOUT = 45
+TIMEOUT_FAIL_BUILD = 3000
+RELATIVE_PATH = 'resource_manager'
+RESOURCE_INFO_JSON = 'info.json'
+RETRY_ON_SSH_EXCEPTIONS = (paramiko.ssh_exception.NoValidConnectionsError,
+                           paramiko.ssh_exception.SSHException)
+BACKOFF_FACTOR = 5
+MAX_RETRIES = 3
 
 logger = logging.getLogger(__name__)
 
 
-class Host:
-    '''
-    Remote host abstract class
-    '''
-    def __init__(self, username: str, hostname: str):
-        self.username = username
-        self.hostname = hostname
-        self.slots: List[Dict[str, Any]] = []
-        self.free_slots: List[int] = []
+class ResourceManagerError(RuntimeError):
+    """ Base module exception """
 
-    def _read_info_data(self):
-        cmd = f"cat {ROOT_PATH}info.json"
-        data = self._ssh_run_remote_command(cmd)
-        return data
 
-    def _read_slot_data(self, slot: int):
-        cmd = f"cat {ROOT_PATH}slot_{slot}.json"
-        data = self._ssh_run_remote_command(cmd)
-        return data
+class LockError(ResourceManagerError):
+    """ Attempt for resource locking returned non-zero exitcode """
+
+
+class LockRetry(ResourceManagerError):
+    """ Retry locking, no resources were available """
+
+
+class RemoteHost:
+
+    def __init__(self, *, hostname: str, username: str, ssh_keyfile: str):
+        """
+        :param hostname: str, remote hostname for ssh connection
+        :param username: str, username for ssh connection
+        :param ssh_keyfile: str, filepath to ssh private key
+        """
+        self._hostname = hostname
+        self._username = username
+        self._ssh_keyfile = ssh_keyfile
+
+    @property
+    def hostname(self):
+        return self._hostname
+
+    @property
+    def username(self):
+        return self._username
+
+    @property
+    def ssh_keyfile(self):
+        return self._ssh_keyfile
 
     def _validate_resource_lock(self, payload: dict):
+        """ Validate resource lock json file against schema """
         with open(RESOURCE_LOCK_SCHEMA, 'r') as f:
             schema_data = f.read()
         schema = json.loads(schema_data)
         try:
             jsonschema.validate(payload, schema)
-            return True
         except jsonschema.ValidationError as err:
-            logger.error(err)
-        return False
+            logger.warning("Invalid json file: %s", err)
 
-    def check_avail(self):
+    @contextmanager
+    def _ssh_session(self):
+        """ Create an SSH connection."""
         try:
-            data = self._read_info_data()
-            if data == '':
-                logger.warning("%s hasn't been deployed yet", self.hostname)
-                return False
-            else:
-                # parse slot count, arch
-                meta = json.loads(data)
-                logger.debug("%s is up", meta['hostname'])
-                # check slots status
-                self.free_slots = []
-                for i in range(meta.get('max_slot_count', 1)):
-                    slot = self._read_slot_data(i)
-                    if slot == b'0\n':
-                        state = 'free'
-                        prid = ''
-                        self.free_slots.append(i)
-                    elif slot == "":
-                        logger.warning("host: %s slot: %s has invalid state", self.hostname, i)
-                        state = 'unknown'
-                        prid = 'unknown'
-                    else:
-                        state = 'running'
-                        slot_data = json.loads(slot)
-                        if self._validate_resource_lock(slot_data):
-                            prid = slot_data.get("prid", "unknown")
-                        else:
-                            logger.warning("host: %s slot: %s has invalid state", self.hostname, i)
-                            state = 'invalid'
-                            prid = 'invalid'
-                    logger.debug("%s slot %s is %s", meta['hostname'], i, state)
-                    self.slots.append({'sid': i, 'state': state, 'prid': prid})
-        except Exception as err:
-            logger.error(err)
-            logger.error("Availiability check failed fo host: %s", self.hostname)
-        return self.free_slots
-
-    def _write_free_slot(self, slot: int):
-        cmd = f'echo \'0\' > {ROOT_PATH}slot_{slot}.json'
-        data = self._ssh_run_remote_command(cmd)
-        return data
-
-    def unlock(self, slot: int, prid: str):
-        logger.debug('Unlocking')
-        retries = 0
-        while retries < 3:
-            ret = self._write_free_slot(slot)
-            if ret == b'':
-                logger.info('Unlocked')
-                return True
-            else:
-                logger.debug('Unlocking failed, retrying')
-                retries += 1
-        return False
-
-    def lock(self, slot: int, prid: str):
-        logger.debug("Aquiring lock")
-        payload = json.dumps({"prid": prid, "locked": str(datetime.now())})
-        retries = 0
-        while retries < 3:
-            filename = os.path.join(ROOT_PATH, f'slot_{slot}.json')
-            _cmd = 'if grep -qw 0 {0} ; then echo {1} > {0} ;else exit 1; fi'.format(
-                    filename,
-                    quote(payload)
-                    )
-            cmd = 'flock {} -c {}; echo $?'.format(filename, quote(_cmd))
-            logger.debug(cmd)
-            ret = self._ssh_run_remote_command(cmd)
-            if ret == b'0\n':
-                logger.info('Locked')
-                return True
-            else:
-                logger.debug('Locking failed, retrying')
-                retries += 1
-        logger.info('Locking failed')
-        raise Exception('Locking failed')
+            client = self._open_ssh_session()
+            yield client
+        finally:
+            client.close()
 
     @backoff.on_exception(
         backoff.expo,
-        Exception,
-        factor=PARAMIKO_BACKOFF_FACTOR,
-        max_tries=PARAMIKO_MAX_RETRIES + 1,  # total tries is N retries + 1 initial attempt
+        RETRY_ON_SSH_EXCEPTIONS,
+        factor=BACKOFF_FACTOR,
+        max_tries=MAX_RETRIES,
         jitter=None,  # use deterministic backoff, do not apply random jitter
+        logger=logger,
     )
-    def _ssh_run_remote_command(self, cmd: str):
-        ssh_client = paramiko.SSHClient()
-        ssh_client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-        ssh_client.connect(hostname=self.hostname,
-                           username=self.username)
-        cmd_out = ssh_client.exec_command(cmd)
-        out = cmd_out[1].read()
-        error = cmd_out[2].read()
-        if error:
-            raise Exception('There was an error pulling the runtime: {}'.format(error.decode()))
-        ssh_client.close()
-        return out
+    def _open_ssh_session(self):
+        """
+        Create a new SSH connection and return connection object.
+        """
+        client = paramiko.SSHClient()
+        client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+        logger.debug("Open SSH connection in remote host %s", self.hostname)
+        client.connect(
+            self.hostname, username=self.username, key_filename=self.ssh_keyfile)
+        return client
+
+    @backoff.on_exception(
+        backoff.expo,
+        RETRY_ON_SSH_EXCEPTIONS,
+        factor=BACKOFF_FACTOR,
+        max_tries=MAX_RETRIES,
+        jitter=None,  # use deterministic backoff, do not apply random jitter
+        logger=logger,
+    )
+    def _ssh_run_remote_cmd(self, client, cmd: List):
+        """
+        Run a Shell command on a remote host
+
+        :return: stdout, stderr and exitcode of shell command
+        """
+        logger.debug("Try to run command %s on remote host %s", cmd, self.hostname)
+        _, stdout, stderr = client.exec_command(cmd, timeout=SSH_COMMAND_TIMEOUT)
+        return stdout, stderr, stdout.channel.recv_exit_status()
+
+    def _ssh_run_remote_cmd_to_get_json(self, client, cmd: List):
+        _, stdout, _ = self._ssh_run_remote_cmd(client, cmd)
+        try:
+            json_stdout = json.loads(stdout)
+        except ValueError as ex:  # Not valid JSON
+            logger.warning("Json load failed. Not a valid JSON. %s", ex)
+            return None
+        except TypeError as ex:  # Not an object
+            logger.warning("Json load failed. Not an object. %s", ex)
+            return None
+        return json_stdout
+
+    def available_slots(self) -> List[int]:
+        """
+        Returns list of available slots
+        """
+        logger.debug("Retrieve list of available slots on host %s", self.hostname)
+
+        with self._ssh_session() as client:
+            cmd = self.get_cmd_to_read_from_json(RESOURCE_INFO_JSON)
+            resource_info = self._ssh_run_remote_cmd_to_get_json(client, cmd)
+
+            available_slots = []
+            for slot_id in range(resource_info.get('max_slot_count', 1)):
+                cmd = self.get_cmd_to_read_from_json(f'slot_{slot_id}.json')
+                slot = self._ssh_run_remote_cmd_to_get_json(client, cmd)
+
+                if slot is None:  # Json load failed - file on remote host might be corrupted
+                    logger.warning("%s - slot %s - json load failed",
+                                   resource_info['hostname'], slot_id)
+                elif slot:  # Non-empty json file was loaded
+                    prid = slot.get('prid')
+                    logger.debug("%s - slot %s is occupied by %s",
+                                 resource_info['hostname'], slot_id, prid)
+                    self._validate_resource_lock(slot)
+                elif not slot:  # Empty json file was loaded
+                    logger.debug("%s - slot %s is available",
+                                 resource_info['hostname'], slot_id)
+                    available_slots.append(slot_id)
+                else:  # Undefined state - should not happen
+                    logger.warning("%s - slot %s invalid state!",
+                                   resource_info['hostname'], slot_id)
+
+        return available_slots
+
+    def get_cmd_to_read_from_json(self, file_name: str):
+        """ Create a command to retrieve content of json file from remote host """
+        return ['cat', os.path.join(RELATIVE_PATH, file_name)]
+
+    def get_cmd_to_write_empty_slot(self, slot_id: int):
+        """ Create a command to unlock slot on remote host """
+        return ['echo', '{}', '>', os.path.join(RELATIVE_PATH, f'slot_{slot_id}.json')]
+
+    def get_cmd_to_write_slot_in_progress(self, file_name: str, payload: str):
+        """ Create a command to lock slot for pipelinerun on remote host """
+        filepath = os.path.join(RELATIVE_PATH, file_name)
+
+        # TBD: Test with remote host that this is actually working with Paramiko
+        _cmd = (f'if grep -qw {{}} {filepath} ; then echo {quote(payload)} > '
+                f'{filepath} ;else exit 1; fi')
+        return f'flock {filepath} -c {quote(_cmd)}'
+
+    @backoff.on_exception(
+        backoff.expo,
+        LockError,
+        factor=BACKOFF_FACTOR,
+        max_tries=MAX_RETRIES,
+        jitter=None,  # use deterministic backoff, do not apply random jitter
+        logger=logger,
+    )
+    def lock(self, slot_id: int, prid: str):
+        logger.debug("Locking slot %s on host %s with pipelinerun ID %s",
+                     slot_id, self.hostname, prid)
+
+        payload = json.dumps({'prid': prid, 'locked': str(datetime.utcnow().isoformat())})
+
+        with self._ssh_session() as client:
+            cmd = self.get_cmd_to_write_slot_in_progress(f'slot_{slot_id}.json', payload)
+            _, _, exitcode = self._ssh_run_remote_cmd(client, cmd)
+
+            if exitcode != 0:
+                raise LockError(f'Attempt to lock slot {slot_id} with pipelinerun ID {prid} '
+                                'returned non-zero exitcode')
+
+        logger.debug("%s - slot %s locked with pipelinerun ID %s", self.hostname, slot_id, prid)
+
+    def unlock(self, slot_id: int, prid: int):
+        logger.debug("Unlocking slot %s on host %s with pipelinerun ID %s",
+                     slot_id, self.hostname, prid)
+
+        with self._ssh_session() as client:
+            cmd = self.get_cmd_to_read_from_json(f'slot_{slot_id}.json')
+            slot = self._ssh_run_remote_cmd_to_get_json(client, cmd)
+
+            if slot:
+                self._validate_resource_lock(slot)
+
+                if slot['prid'] != prid:
+                    logger.warning("%s - slot %s, Found pipelinerun ID: %s, expected %s. Unlocking"
+                                   " skipped!", self.hostname, slot_id, slot['prid'], prid)
+                    return
+
+                cmd = self.get_cmd_to_write_empty_slot(slot_id)
+                _, _, exitcode = self._ssh_run_remote_cmd(client, cmd)
+
+                if exitcode != 0:
+                    logger.warning("Unlocking slot %s on host %s failed!", slot_id, self.hostname)
+            else:
+                logger.warning("%s - slot %s is available. Unlocking skipped!",
+                               self.hostname, slot_id)
 
 
-class ResourceManager:
-    '''
-    Manages build resources
-    '''
+class LockedResource:
 
-    def __init__(self, hosts: List[Host]):
+    def __init__(self, host: RemoteHost, slot: int, prid: str):
+        """
+        :param slot: int, Remote host slot ID
+        :param prid: str, Pipeline run ID
+        """
+        self.slot = slot
+        self.prid = prid
+        self.host = host
+
+    def unlock(self):
+        self.host.unlock(self.slot, self.prid)
+
+
+class RemoteHostsPool:
+
+    def __init__(self, hosts: List[RemoteHost]):
+        """
+        :param hosts: List[RemoteHost], List of Remote hosts
+        """
         self.hosts = hosts
 
-    def find(self) -> List[Host]:
-        try:
-            resources = []
-            for host in self.hosts:
-                host.check_avail()
-                if len(host.free_slots) > 0:
-                    resources.append(host)
-            sorted_result = sorted(resources, key=lambda h: len(h.free_slots))
-            logger.debug('sorted result: %s', sorted_result)
-            return sorted_result
-        except Exception as err:
-            logger.debug(err)
-            logger.error("Unable to read hosts file")
-            return []
+    @classmethod
+    def from_config(cls, config: dict):
+        """ Instantiate remote hosts loaded from configmap
 
-    def obtain_free_resource(self, prid: str) -> Tuple[Host, int]:
-        availiable_hosts = self.find()
-        for host in availiable_hosts:
-            for sid in host.free_slots:
-                if host.lock(sid, prid):
-                    return host, sid
-        raise Exception('No availiable host found')
+        :param config: dict, Arch specific remote hosts dictionary from configmap
 
+        Example:
 
-if __name__ == '__main__':
-    my_hosts = [Host('username', 'hostname')]
-    mngr = ResourceManager(my_hosts)
-    # locking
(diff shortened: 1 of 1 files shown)


atomic_reactor/utils/remote_hosts_resources.py
```diff
@@ -0,0 +1,295 @@
...
+
+    def _validate_resource_lock(self, payload: dict):
+        """ Validate resource lock json file against schema """
+        with open(RESOURCE_LOCK_SCHEMA, 'r') as f:
+            schema_data = f.read()
+        schema = json.loads(schema_data)
+        try:
+            jsonschema.validate(payload, schema)
+        except jsonschema.ValidationError as err:
+            logger.warning("Invalid json file: %s", err) #  TODO - What here? Fail build? Leave it be? Set file to empty json?
```
MartinBasti, line 82: Don't fail build, just skip it

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
 from shlex import quote
+from typing import Any, Dict, List, Tuple
+
+SSH_COMMAND_TIMEOUT = 120
+RELATIVE_PATH = 'resource_manager'
+RESOURCE_INFO_JSON = 'info.json'
+RETRY_ON_SSH_EXCEPTIONS = (paramiko.ssh_exception.AuthenticationException,
+                           paramiko.ssh_exception.ChannelException,
+                           paramiko.ssh_exception.NoValidConnectionsError,
+                           paramiko.ssh_exception.SSHException)
```
MartinBasti, line 26:
```
Subclasses:     AuthenticationException, ChannelException, BadHostKeyException, IncompatiblePeer, ProxyCommandFailure, CouldNotCanonicalize, ConfigParseError
```

I don't think we have to explicitly mention `AuthenticationException` and `ChannelException`

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
 import jsonschema
-from typing import List, Any, Dict, Tuple
+import paramiko
+import random
+import os
 from datetime import datetime
 from shlex import quote
+from typing import Any, Dict, List, Tuple
+
+SSH_COMMAND_TIMEOUT = 120
```
MartinBasti, line 20: quite long?

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
-                logger.info('Locked')
-                return True
-            else:
-                logger.debug('Locking failed, retrying')
-                retries += 1
-        logger.info('Locking failed')
-        raise Exception('Locking failed')
+            client.connect(
+                self.hostname, username=self.username, key_filename=self.ssh_keyfile)
+        except paramiko.ssh_exception.AuthenticationException as ex:
```
MartinBasti, line 107: Do we need this for each subclass?

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
+            logger.warning("SSH Connection error. %s", ex)
+        except paramiko.ssh_exception.SSHException as ex:
+            logger.warning("SSH Connection error. %s", ex)
+        return stdout, stderr, client.recv_exit_status()
+
+    def _ssh_run_remote_cmd_to_get_json(self, client, cmd: List):
+        stdin, stdout, stderr = self._ssh_run_remote_cmd(client, cmd)
+        try:
+            json_stdout = json.loads(stdout)
+        except ValueError as ex:  # Not valid JSON  #  TODO: Question - Do we want to fail build on loading invalid json from remote host????
```
MartinBasti, line 148: Skip, don't fail build. We don't want to explode all builds because one slot/host is misconfigured

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
+                    logger.warning("%s - slot %s - json load failed", resource_info['hostname'], slot_id)
+                elif slot:  # Non-empty json file was loaded
+                    prid = slot.get('prid')
+                    logger.debug("%s - slot %s is occupied by %s", resource_info['hostname'], slot_id, prid)
+                    self._validate_resource_lock(slot)
+                elif not slot:  # Empty json file was loaded
+                    logger.debug("%s - slot %s is available", resource_info['hostname'], slot_id)
+                    available_slots.append(slot_id)
+                else:  # Undefined state - should not happen
+                    raise RuntimeError(f"{resource_info['hostname']} - slot {slot_id} invalid state!")
```
MartinBasti, line 187: :eyes:

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
+
+        return random.shuffle(available_slots)
+
+    def get_cmd_to_read_from_json(self, file_name: str):
+        """ Create a command to retrieve content of json file from remote host """
+        return ['cat', os.path.join(RELATIVE_PATH, file_name)]
```
MartinBasti, line 193: if you define RELATIVE_PATH as pathobject in module, you can just use
```python
str(RELATIVE_PATH / filename)
```

not sure if that's better

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
+        filepath = os.path.join(RELATIVE_PATH, file_name)
+        _cmd = (f'if grep -qw {{}} {filepath} ; then echo {quote(payload)} > '
+                f'{filepath} ;else exit 1; fi') #  TODO: Test with remote host that this is actually working with Paramiko
+        return f'flock {filepath} -c {quote(_cmd)}'
+
+    def lock(self, slot_id: int, prid: str):
+        logger.debug("Locking slot %s on host %s with pipelinerun ID %s",
+                     slot, self._hostname, prid)
+
+        payload = json.dumps({'prid': prid, 'locked': str(datetime.now())})
```
MartinBasti, line 210: I'd prefer this here `datetime.utcnow().isoformat()`

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
+                                'returned non-zero exitcode')
+
+        logger.debug("%s - slot %s locked with pipelinerun ID %s", self._hostname, slot_id, prid)
+
+    def unlock(self, slot_id: int, prid: int):
+        logger.debug("Unlocking slot %s on host %s with pipelinerun ID %s",
+                     slot_id, self._hostname, prid)
+
+        with self._ssh_session() as client:
+            #  TODO: Is critical section needed here? -- Start
```
MartinBasti, line 227: yes, but not part of this PR

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
+    @classmethod
+    def from_config(cls, config: dict):
+        """ Instantiate remote hosts loaded from configmap
+
+        :param config: dict, Arch specific remote hosts dictionary from configmap
+
+        Example:
+
+        hostname-remote-host1:
+          enabled: true
```
MartinBasti, line 267: should this filter out not enabled hosts?

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
+                 for hostname, attr in config.items()]
+
+        return cls(hosts)
+
+    def lock_resource(self, prid: str) -> LockedResource:
+        resources = [
+            (host, host.available_slots())
+            for host in self.hosts
+        ]
+        resources.sorted(key=lambda x: len(x[1]), reverse=True)  # Sort based on number of available slots
```
MartinBasti, line 285: bonus points for sorting based on `available/all` slots

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
+        Returns list of available slots
+
+        Slots should be randomized here, for smoother locking, because there will be
+        hunger games when every pipeline will try to lock slot with lowest number first
+        """
+        logger.debug("Retrieve list of available slots on host %s", self._hostname)
+
+        with self._ssh_session() as client:
+            cmd = self.get_cmd_to_read_from_json(RESOURCE_INFO_JSON)
+            resource_info = self._ssh_run_remote_cmd_to_get_json(client, cmd)
```
MartinBasti, line 170: I'm thinking, shouldn't be this just single method?

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
+        except paramiko.ssh_exception.SSHException as ex:
+            logger.warning("SSH Connection error. %s", ex)
+        return client
 
     @backoff.on_exception(
         backoff.expo,
-        Exception,
-        factor=PARAMIKO_BACKOFF_FACTOR,
-        max_tries=PARAMIKO_MAX_RETRIES + 1,  # total tries is N retries + 1 initial attempt
+        RETRY_ON_SSH_EXCEPTIONS,
```
MartinBasti, line 119: Don't we have quite nested retries here?

MartinBasti (changes requested):

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
-logger = logging.getLogger(__name__)
+class LockedResource:
 
+    def __init__(host: RemoteHost, slot: int, prid: str):
+        """
+        :param slot: int, Remote host slot ID
+        :param prid: str, Pipeline run ID
+        """
+        self.slot = slot
+        self.prid = prid
```
MartinBasti, line 44: and `self.host = host` ?

atomic_reactor/utils/resource_manager.py
lkolacek: depends if we want to do selection of enabled hosts here or on OSBS side. I am OK with selecting them here.

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
-        data = self._ssh_run_remote_command(cmd)
-        return data
 
-    def _read_slot_data(self, slot: int):
-        cmd = f"cat {ROOT_PATH}slot_{slot}.json"
-        data = self._ssh_run_remote_command(cmd)
-        return data
+class RemoteHost:
+
+    def __init__(self, hostname: str, username: str, ssh_keyfile: str):
```
MartinBasti, line 52: There will be more attrs, so it's better to enforce params by name
```suggestion
    def __init__(self, *, hostname: str, username: str, ssh_keyfile: str):
```

atomic_reactor/utils/resource_manager.py
lkolacek: Ack

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
+                    prid = slot.get('prid')
+                    logger.debug("%s - slot %s is occupied by %s", resource_info['hostname'], slot_id, prid)
+                    self._validate_resource_lock(slot)
+                elif not slot:  # Empty json file was loaded
+                    logger.debug("%s - slot %s is available", resource_info['hostname'], slot_id)
+                    available_slots.append(slot_id)
+                else:  # Undefined state - should not happen
+                    raise RuntimeError(f"{resource_info['hostname']} - slot {slot_id} invalid state!")
+
+        return random.shuffle(available_slots)
```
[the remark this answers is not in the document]
MartinBasti: Explanation why, may be handy (I know why, but will I know why in 3 months? :D)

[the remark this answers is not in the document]
MartinBasti: Taking back, it's in docstring :|

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
+        with self._ssh_session() as client:
+            cmd = self.get_cmd_to_read_from_json(RESOURCE_INFO_JSON)
+            resource_info = self._ssh_run_remote_cmd_to_get_json(client, cmd)
+
+            available_slots = []
+            for slot_id in range(resource_info.get('max_slot_count', 1)):
+                cmd = self.get_cmd_to_read_from_json(f'slot_{slot_id}.json')
+                slot = self._ssh_run_remote_cmd_to_get_json(client, cmd)
+
+                if slot is None:  # Json load failed - file on remote host might be corrupted
```
MartinBasti, line 177: How are slots created initially?

atomic_reactor/utils/resource_manager.py
lkolacek: Can you elaborate more here? do you mean dropping `_ssh_run_remote_cmd_to_get_json` and using only `_ssh_run_remote_cmd` with json load separately??

atomic_reactor/utils/resource_manager.py
lkolacek: 30? 60?

atomic_reactor/utils/resource_manager.py
lkolacek: I believe via our configuration playbook for remote hosts, but I'll check it in design.

atomic_reactor/utils/resource_manager.py
lkolacek: If there won't be any strong opinion on this, It will stay as it is now. :)

atomic_reactor/utils/resource_manager.py
chmeliik: Dosn't backoff log something by itself anyway?

Also note that if you suppress the exceptions, backoff will not work

atomic_reactor/utils/resource_manager.py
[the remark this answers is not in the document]
chmeliik: Could we shuffle in lock_resource() instead? Trying the slots in a random order is relevant to the locking mechanism but not to this method.

And note that random.shuffle does `Shuffle list x in place, and return None.`

atomic_reactor/utils/resource_manager.py
rcerven: sorted is function not list method

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
+    def get_cmd_to_write_empty_slot(self, slot_id: int):
+        """ Create a command to unlock slot on remote host """
+        return ['echo', '{}', '>', os.path.join(RELATIVE_PATH, f'slot_{slot_id}.json')]
+
+    def get_cmd_to_write_slot_in_progress(self, file_name: str, payload: str):
+        """ Create a command to lock slot for pipelinerun on remote host """
+        filepath = os.path.join(RELATIVE_PATH, file_name)
+        _cmd = (f'if grep -qw {{}} {filepath} ; then echo {quote(payload)} > '
+                f'{filepath} ;else exit 1; fi') #  TODO: Test with remote host that this is actually working with Paramiko
+        return f'flock {filepath} -c {quote(_cmd)}'
```
rcerven, line 204: this locks the file ONLY for duration of the command ......

you have to use file descriptor

1st shell:
$ exec 99>"lock.file"
$ flock -n 99
$ echo $?
0

2nd shell
$ exec 99>"lock.file"
$ flock -n 99
$ echo $?
1

1st shell:
$ flock -u -n 99

2nd shell 
$ flock -n 99
$ echo $?
0

but all that works ONLY if that shell is still active

rcerven: also keep in mind that file actually isn't locked itself, any process can write to it

rcerven: so basically what it does now is that it is using flock feature just for writing to the file .... (was that really the intent ?)

atomic_reactor/utils/resource_manager.py
lkolacek: Adam is right here. Logging can be added directly to backoff decorator and exception will break backoff completely. I made few tests here. https://gitlab.cee.redhat.com/-/snippets/4833 I'll make a fix.

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
         with open(RESOURCE_LOCK_SCHEMA, 'r') as f:
             schema_data = f.read()
         schema = json.loads(schema_data)
         try:
             jsonschema.validate(payload, schema)
-            return True
         except jsonschema.ValidationError as err:
-            logger.error(err)
-        return False
+            logger.warning("Invalid json file: %s", err) #  TODO - What here? Fail build? Leave it be? Set file to empty json?
```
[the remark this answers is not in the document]
lkolacek: I would leave fixing the slot to cleanup job rather then doing it here. So I'll leave it as it is for now. Thanks for feedback.

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,290 @@
...
-    free_host, slot = mngr.obtain_free_resource('unique-name-for-pipeline-run')
-
-    # Unlocking
-    free_host.unlock(slot, 'unique-name-for-pipeline-run')
+            stdin, stdout, stderr = client.exec_command(
+                cmd, timeout=SSH_COMMAND_TIMEOUT)
+        except paramiko.ssh_exception.AuthenticationException as ex:
+            logger.warning("SSH authentication failed. %s", ex) #  TODO: Question - Probably not needed for exec command
+        except paramiko.ssh_exception.ChannelException as ex:
+            logger.warning("Attempt to open a new Channel failed. %s", ex) #  TODO: Question - Probably not needed for exec command
```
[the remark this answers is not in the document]
lkolacek: I agree, Tested here https://gitlab.cee.redhat.com/-/snippets/4833 I'll make a fix

atomic_reactor/utils/resource_manager.py
lkolacek: This part was not introduced by myself, I would like to re-test it to be sure it's working properly. But Anyway it's not complete, beacuse`critical section` will be required here for locking and unlocking as I mentioned in another comment, but it will be added in a separate PR.

atomic_reactor/utils/remote_hosts_resources.py
lkolacek: ack

atomic_reactor/utils/resource_manager.py
lkolacek: ok I'll remove those from list of exceptions

atomic_reactor/utils/resource_manager.py
lkolacek: I'll set this to 45 for now

atomic_reactor/utils/resource_manager.py
lkolacek: ack

atomic_reactor/utils/resource_manager.py
lkolacek: I agree

atomic_reactor/utils/resource_manager.py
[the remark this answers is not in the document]
lkolacek: Ou I'll fix that. Thanks

atomic_reactor/utils/resource_manager.py
lkolacek: Ok enabled hosts will be selected here in resource_manager.

atomic_reactor/utils/resource_manager.py
lkolacek: Ack. Fixed

atomic_reactor/utils/resource_manager.py
chmeliik: > so basically what it does now is that it is using flock feature just for writing to the file .... (was that really the intent ?)

Yes, I believe it was. `flock` is there only to prevent race conditions while locking and unlocking. The file does not have to stay flock-ed for the entire duration of the podman build

rcerven: well that seems weird anyway such usage of flock

as I've said earlier, anything can still write to that file anyway

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,304 @@
...
+
+        # Slots should be randomized here to avoid always locking the lowest numbers first
+        resources = []
+        for host in self.hosts:
+            available_slots = host.available_slots()
+            random.shuffle(available_slots)
+            resources.append((host, available_slots))
+
+        # Sort based on number of available slots
+        sorted(resources, key=lambda x: len(x[1]), reverse=True)
```
MartinBasti, line 296: This?
```suggestion
        recources.sort(key=lambda x: len(x[1]), reverse=True)
```

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,304 @@
...
+        # Add enabled clusters to list hosts
+        hosts = [RemoteHost(hostname=hostname, username=attr['username'], ssh_keyfile=attr['auth'])
+                 for hostname, attr in config.items() if attr['enabled']]
+
+        return cls(hosts)
+
+    @backoff.on_exception(
+        backoff.constant,
+        LockRetry,
+        max_time=TIMEOUT_FAIL_BUILD,
```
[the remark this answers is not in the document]
MartinBasti: (optional, I'm not sure here)

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,304 @@
...
-ROOT_PATH = 'resource_manager/'
 
-RESOURCE_LOCK_SCHEMA = 'schemas/resource-lock-content.json'
+class LockError(Exception):
+    """ Attempt for resource locking returned non-zero exitcode """
+
+
+class LockRetry(Exception):
+    """ Retry locking, no resources were available """
+
```
MartinBasti, line 41: IMO we should inherit from RuntimeError (this is currently used in OSBS everywhere)
```suggestion
class ResourceManagerError(RuntimeError):
   """Base module exception"""

class LockError(ResourceManagerError):
    """ Attempt for resource locking returned non-zero exitcode """


class LockRetry(ResourceManagerError):
    """ Retry locking, no resources were available """

```

atomic_reactor/utils/resource_manager.py
lkolacek: I hope we should not now.

atomic_reactor/utils/resource_manager.py
lkolacek: Leaving as it is. one method is general for shell commands on remote host and second one is used for commands, which returns json format to be automatically loaded and passed as dict.

atomic_reactor/utils/resource_manager.py
lkolacek: Added note to ticket for follow up PR.

atomic_reactor/utils/resource_manager.py
lkolacek: Added

atomic_reactor/utils/resource_manager.py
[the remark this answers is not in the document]
lkolacek: @rcerven @chmeliik thoughts?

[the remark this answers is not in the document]
rcerven: if locking fails with retries on 1 host we should move on to another candidate host

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,308 @@
...
+RESOURCE_INFO_JSON = 'info.json'
+RETRY_ON_SSH_EXCEPTIONS = (paramiko.ssh_exception.NoValidConnectionsError,
+                           paramiko.ssh_exception.SSHException)
+BACKOFF_FACTOR = 5
+MAX_RETRIES = 3
 
-PARAMIKO_MAX_RETRIES = 3
-PARAMIKO_BACKOFF_FACTOR = 1
+logger = logging.getLogger('resource-manager')
+logger.setLevel(logging.ERROR)
```
chmeliik, line 32:
```suggestion
logger = logging.getLogger(__name__)
```

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,308 @@
...
+        return self._username
+
+    @property
+    def ssh_keyfile(self):
+        return self._ssh_keyfile
 
     def _validate_resource_lock(self, payload: dict):
+        """ Validate resource lock json file against schema """
         with open(RESOURCE_LOCK_SCHEMA, 'r') as f:
             schema_data = f.read()
```
chmeliik, line 74: Will this work after installation? `RESOURCE_LOCK_SCHEMA` is a relative path. We should use `util.validate_with_schema` for this.

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,308 @@
...
-            logger.error(err)
-        return False
+            logger.warning("Invalid json file: %s", err)
+
+    @contextmanager
+    def _ssh_session(self):
+        """ Create an SSH connection."""
+        client = self._open_ssh_session()
+        yield client
+        client.close()
```
chmeliik, line 86: client.close() should be in `finally` block, otherwise connection won't be closed on exceptions

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,308 @@
...
+    )
+    def _ssh_run_remote_cmd(self, client, cmd: List):
+        """
+        Run a Shell command on a remote host
 
-    def check_avail(self):
+        :return: stdout, stderr and exitcode of shell command
+        """
+        logger.debug("Try to run command %s on remote host %s", cmd, self.hostname)
+        _, stdout, stderr = client.exec_command(cmd, timeout=SSH_COMMAND_TIMEOUT)
```
chmeliik, line 122: exec_command does not support passing the command as a list

we could convert it with `cmd_str =  " ".join(map(shlex.quote, cmd))`

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,308 @@
...
+        return stdout, stderr, client.recv_exit_status()
```
chmeliik, line 123:
```suggestion
        return stdout, stderr, stdout.channel.recv_exit_status()
```
The Client class does not have this method.

Also, according to [Paramiko docs](https://docs.paramiko.org/en/stable/api/channel.html#paramiko.channel.Channel.recv_exit_status), `recv_exit_status` may hang indefinitely unless you read from the output streams first. Maybe this method should fully consume stdout and stderr, and return strings rather than the open Channel Files.

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,308 @@
...
-                retries += 1
-        logger.info('Locking failed')
-        raise Exception('Locking failed')
+            json_stdout = json.loads(stdout)
+        except ValueError as ex:  # Not valid JSON
+            logger.warning("Json load failed. Not a valid JSON. %s", ex)
+            return None
+        except TypeError as ex:  # Not an object
+            logger.warning("Json load failed. Not an object. %s", ex)
+            return None
```
chmeliik, line 134: How would this even happen?

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,308 @@
...
+                    logger.debug("%s - slot %s is occupied by %s",
+                                 resource_info['hostname'], slot_id, prid)
+                    self._validate_resource_lock(slot)
+                elif not slot:  # Empty json file was loaded
+                    logger.debug("%s - slot %s is available",
+                                 resource_info['hostname'], slot_id)
+                    available_slots.append(slot_id)
+                else:  # Undefined state - should not happen
+                    logger.warning("%s - slot %s invalid state!",
+                                   resource_info['hostname'], slot_id)
```
chmeliik, line 166: Not just undefined state, straight up impossible unless the `__bool__` method of slots returns non-deterministic results :smile:

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,308 @@
...
+
+        return available_slots
+
+    def get_cmd_to_read_from_json(self, file_name: str):
+        """ Create a command to retrieve content of json file from remote host """
+        return ['cat', os.path.join(RELATIVE_PATH, file_name)]
```
chmeliik, line 172: IMO we should use an absolute path

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,308 @@
...
+
+    def get_cmd_to_write_empty_slot(self, slot_id: int):
+        """ Create a command to unlock slot on remote host """
+        return ['echo', '{}', '>', os.path.join(RELATIVE_PATH, f'slot_{slot_id}.json')]
```
chmeliik, line 176: Redirects are only interpreted by shells, `>` will not work in list form. But then again, `exec_command` does not support lists anyway.

atomic_reactor/utils/resource_manager.py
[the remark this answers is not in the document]
chmeliik: That already happens inside this method. If *all* slots on *all* hosts fail to lock, that's when the method raises LockRetry and does the entire process again.

I don't mind having the extra timeout here.

chmeliik (commented):
LGTM, the comments don't need to be addressed right away (but they do need to be addressed eventually)

chmeliik (approved):

atomic_reactor/utils/resource_manager.py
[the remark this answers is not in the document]
rcerven: ic, well there are now 2 use cases:
1) something is wrong with hosts, that we are unable to lock any resource
2) there is more pipeline runs running than remote hosts capacity

for 2nd case we should wait until there is free slot

for 1st case mmm, there might be some temporarily network issue, or there might be some serious issue with hosts, so if we will wait indefinitely (pipeline run timeout will kill run, but it will take a lot time to discover issue)

atomic_reactor/utils/resource_manager.py
lkolacek: Yeah this should not ever happen :D .. I was not sure If I should add else or leave it be.

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,309 @@
...
-PARAMIKO_MAX_RETRIES = 3
-PARAMIKO_BACKOFF_FACTOR = 1
-
-ROOT_PATH = 'resource_manager/'
+from typing import List
 
 RESOURCE_LOCK_SCHEMA = 'schemas/resource-lock-content.json'
+SSH_COMMAND_TIMEOUT = 45
+TIMEOUT_FAIL_BUILD = 3000
+RELATIVE_PATH = 'resource_manager'
```
tkdchen, line 24: @lkolacek What is the relative path of this RELATIVE_PATH? Is the working directory changed at somewhere or is the working directory determined by the configured ssh login?

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,309 @@
...
+        Run a Shell command on a remote host
+
+        :return: stdout, stderr and exitcode of shell command
+        """
+        logger.debug("Try to run command %s on remote host %s", cmd, self.hostname)
+        _, stdout, stderr = client.exec_command(cmd, timeout=SSH_COMMAND_TIMEOUT)
+        return stdout, stderr, stdout.channel.recv_exit_status()
 
-    def find(self) -> List[Host]:
+    def _ssh_run_remote_cmd_to_get_json(self, client, cmd: List):
```
tkdchen, line 126: What is the json?

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,309 @@
...
-
-    # Unlocking
-    free_host.unlock(slot, 'unique-name-for-pipeline-run')
+            json_stdout = json.loads(stdout)
+        except ValueError as ex:  # Not valid JSON
+            logger.warning("Json load failed. Not a valid JSON. %s", ex)
+            return None
+        except TypeError as ex:  # Not an object
+            logger.warning("Json load failed. Not an object. %s", ex)
+            return None
```
tkdchen, line 135: @lkolacek What does the None mean to the caller?

- In `available_slots`,
  - None is not handled for the value of `resource_info`
  - None means "json load failed" for getting the value of `slot`.
- In `unlock`, None means a slot is available on a host and unlocking skipped. Looks like None is a correct return value as expected.
In

atomic_reactor/utils/resource_manager.py
```diff
@@ -6,193 +6,309 @@
...
-    free_host, slot = mngr.obtain_free_resource('unique-name-for-pipeline-run')
-
-    # Unlocking
-    free_host.unlock(slot, 'unique-name-for-pipeline-run')
+            json_stdout = json.loads(stdout)
+        except ValueError as ex:  # Not valid JSON
+            logger.warning("Json load failed. Not a valid JSON. %s", ex)
+            return None
+        except TypeError as ex:  # Not an object
+            logger.warning("Json load failed. Not an object. %s", ex)
```
tkdchen, line 134: Should we log more info about the invalid stdout? Perhaps we can log even the whole stdout value. That might be convenience for us to debug.

atomic_reactor/utils/resource_manager.py
tkdchen: I think the validation can be done earlier and continue the for-loop immediately when slot is None:

```
if slot is None:
    logger.warning()
    continue

self._validate_resource_lock(slot)

if slot:
    logger.debug("slot %s is occupied by %s")
else:
    logger.debug("slot is available.")
    available_slots.append(slot_id)
```

The data types are validated by the schema. Therefore, no "Undefined state" has to be cared about.

But, in either case, should the validation error be handled specifically? The `lock_resource` will fail if the validation fails. Then, will the pipeline run fail eventually? Or would it be good to log enough information for the invalid slot or the specific host and let the `lock_resource` locks other available resource as much as possible in order to keep the pipeline run?

chmeliik: We discussed this a bit in one of the other threads - we want to log a warning and let the pipeline run try a different slot. Invalid slots could be fixed by the cleanup cronjob

PR #1792, opened 2022-02-02
mypy: fix some plugins

MartinBasti:

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -179,7 +179,7 @@ class AddLabelsPlugin(PreBuildPlugin):
 
         for old, new in self.aliases.items():
             if old not in all_labels:
-                applied_alias = not_applied.append(old)
+                not_applied.append(old)
                 continue
 
             # new label doesn't exists but old label does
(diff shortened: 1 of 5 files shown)


ben-alkov (approved):
LGTM

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -259,8 +259,10 @@ def add_release_env_var(self, df_parser):
                 self.log.warning("environment release variable %s could not be set because no "
                                  "release label found", release_env_var)
 
-    def add_labels_to_df(self, build_dir: BuildDir) -> str:
+    def add_labels_to_df(self, build_dir: BuildDir):
```
chmeliik, line 262: `-> None`

No annotation implicitly means `-> Any`

atomic_reactor/plugins/pre_add_labels_in_df.py
```diff
@@ -179,7 +179,7 @@ def add_as_an_alias(set_to, set_from):
 
         for old, new in self.aliases.items():
             if old not in all_labels:
-                applied_alias = not_applied.append(old)
+                not_applied.append(old)
```
chmeliik, line 182: This also changes the logic, but I'm pretty sure the original logic was a bug and it's more correct now (and it's only for a log message anyway)

atomic_reactor/plugins/pre_add_labels_in_df.py
MartinBasti: TIL, thanks

atomic_reactor/plugins/pre_add_labels_in_df.py
MartinBasti: Honestly I'm confused with that logic there, why is there even function that always returns `True` :-|

chmeliik: Right? :smile: 

I was going to refactor that - I'm pretty sure `applied_alias` is useless now and the condition could just be `if alias_labels and not_applied`. But then I got lost and gave up, better to leave this skeleton in the closet :smile:

chmeliik (approved):
LGTM :+1:

MartinBasti:
no release notes, internal changes to mypy annotations

PR #1798, opened 2022-02-04
Update change_from_in_df plugin for Tekton

chmeliik:
Change the FROM lines in each per-platform Dockerfile.

Slight change in behavior: the plugin now checks for unresolved parents
before checking for mismatched base image. Both of those would be code
bugs, it shouldn't matter which we check first.

### DIFF
--- a/atomic_reactor/plugins/pre_change_from_in_df.py
+++ b/atomic_reactor/plugins/pre_change_from_in_df.py
@@ -9,8 +9,9 @@ of the BSD license. See the LICENSE file for details.
 Pre-build plugin that changes the parent images used in FROM instructions
 to the more specific names given by the builder.
 """
+from atomic_reactor.dirs import BuildDir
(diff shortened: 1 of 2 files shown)


chmeliik:
PTAL, this and #1799 are the last remaining things that need to be updated in prebuild plugins

(that I'm aware of, there will be unexpected problems I'm sure)

mkosiarc (approved):

atomic_reactor/plugins/pre_change_from_in_df.py
```diff
@@ -99,5 +95,18 @@ def run(self):
 
         self.log.debug(
```
rcerven, line 96: can we also for consistency report that if base_from_scratch:

"base image 'scratch' unchanged" or something like that

chmeliik: Ack, added

hjmodi (approved):

PR #1802, opened 2022-02-09
Resource manager enhancements

qixiang:
Update resource_manager module

main changes:

1. Only expose lock operation APIs
2. Logging most exceptions instead of raising, there is already retry
   mechanism, when failed to lock a host slot, we'd like to just try 
   another one, and not be forced to handle the exceptions/errors.
3. Change slot file content format from json to simple line of: 
   `<prid>@<datetime-isoformat`
4. Try to avoid race conditions by running a single flock command
   for most operations

### DIFF
--- a/atomic_reactor/utils/resource_manager.py
+++ b/atomic_reactor/utils/resource_manager.py
@@ -7,22 +7,19 @@ of the BSD license. See the LICENSE file for details.
 """
 
 import backoff
-import json
-import jsonschema
 import logging
-import paramiko
-import random
 import os
+import paramiko
 from contextlib import contextmanager
 from datetime import datetime
 from shlex import quote
-from typing import List
+from typing import List, Optional
 
-RESOURCE_LOCK_SCHEMA = 'schemas/resource-lock-content.json'
 SSH_COMMAND_TIMEOUT = 45
-TIMEOUT_FAIL_BUILD = 3000
-RELATIVE_PATH = 'resource_manager'
-RESOURCE_INFO_JSON = 'info.json'
+TIMEOUT_FAIL_BUILD = 30
+FLOCK_TIMEOUT = 10
+FLOCK_CMD = f"flock -w {FLOCK_TIMEOUT}"
+SLOTS_RELATIVE_PATH = "osbs_slots"
 RETRY_ON_SSH_EXCEPTIONS = (paramiko.ssh_exception.NoValidConnectionsError,
                            paramiko.ssh_exception.SSHException)
 BACKOFF_FACTOR = 5
@@ -31,29 +28,20 @@ MAX_RETRIES = 3
 logger = logging.getLogger(__name__)
 
 
-class ResourceManagerError(RuntimeError):
-    """ Base module exception """
-
-
-class LockError(ResourceManagerError):
-    """ Attempt for resource locking returned non-zero exitcode """
-
-
-class LockRetry(ResourceManagerError):
-    """ Retry locking, no resources were available """
-
-
 class RemoteHost:
 
-    def __init__(self, *, hostname: str, username: str, ssh_keyfile: str):
+    def __init__(self, *, hostname: str, username: str, ssh_keyfile: str, slots: int):
         """
         :param hostname: str, remote hostname for ssh connection
         :param username: str, username for ssh connection
         :param ssh_keyfile: str, filepath to ssh private key
+        :param slots: int, number of max allowed slots on remote host
         """
         self._hostname = hostname
         self._username = username
         self._ssh_keyfile = ssh_keyfile
+        self.slots = slots
+        self._slots_dir = None
 
     @property
     def hostname(self):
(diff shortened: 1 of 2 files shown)


atomic_reactor/utils/resource_manager.py
```diff
@@ -97,10 +97,12 @@ def _validate_resource_lock(self, payload: dict):
     def _ssh_session(self):
         """Create an SSH connection."""
         try:
+            client = None
             client = self._open_ssh_session()
             yield client
         finally:
-            client.close()
+            if client is not None:
+                client.close()
```
chmeliik, line 105: Would this work?
```python
client = self._open_ssh_session()
try:
    yield client
finally:
    client.close()
```

atomic_reactor/utils/resource_manager.py
```diff
@@ -45,6 +45,16 @@ class LockRetry(ResourceManagerError):
     """Retry locking, no resources were available"""
 
 
+class RemoteHostLoggerAdapter(logging.LoggerAdapter):
+    def process(self, msg, kwargs):
+        """Manipulate the message to always prepend hostname"""
```
chmeliik, line 50: Nice! :tada:

atomic_reactor/utils/resource_manager.py
```diff
@@ -18,43 +18,59 @@
...
         :param hostname: str, remote hostname for ssh connection
         :param username: str, username for ssh connection
         :param ssh_keyfile: str, filepath to ssh private key
+        :param logger: logging.Logger, logger to use. Optional, defaults to LOGGER
         """
         self._hostname = hostname
         self._username = username
         self._ssh_keyfile = ssh_keyfile
 
+        self.logger = RemoteHostLoggerAdapter(logger, {"hostname": hostname})
```
MartinBasti, line 72: Can we have per class specific logger. Passing logger via argument is not consistent with other modules (and I think we don't need it)
```suggestion
        self.logger = RemoteHostLoggerAdapter(
            logging.getLogger(f"{self.__class__.__module__}.{self.__class__.__name__}"),
             {"hostname": hostname})
```

MartinBasti: or just keep module logger there without param as is done in other modules in OSBS

ben-alkov:
@qixiang; Do not use 'black'

atomic_reactor/utils/resource_manager.py
qixiang: Done

qixiang:
PR is updated to address the comments and some of the remaining required changes required, but keep in mind, there is still some flaws on handling race conditions, it'd would be better to resolve that first, I'll update this PR after having some discussions regarding the race condition issues.

atomic_reactor/utils/resource_manager.py
```diff
@@ -121,10 +121,13 @@ def _ssh_run_remote_cmd(self, client, cmd: str):
         """
         logger.debug("Try to run command %s on remote host %s", cmd, self.hostname)
         _, stdout, stderr = client.exec_command(cmd, timeout=SSH_COMMAND_TIMEOUT)
-        return stdout, stderr, stdout.channel.recv_exit_status()
+        exitcode = stdout.channel.recv_exit_status()
+        stdout = stdout.read().decode().strip()
+        stderr = stderr.read().decode().strip()
```
chmeliik, line 171: Could we first read and then call `recv_exit_status()`? The Paramiko docs have this [warning](https://docs.paramiko.org/en/stable/api/channel.html#paramiko.channel.Channel.recv_exit_status) that recv_exit_status may hang forever unless you receive something from the channel first

atomic_reactor/utils/resource_manager.py
```diff
@@ -273,14 +275,20 @@ def from_config(cls, config: dict):
...
-        hosts = [RemoteHost(hostname=hostname, username=attr['username'], ssh_keyfile=attr['auth'])
-                 for hostname, attr in config.items() if attr['enabled']]
+        hosts = []
+        for hostname, attr in config.items():
+            if not attr.get("enabled", False):
+                continue
+            host = RemoteHost(
+                hostname=hostname, username=attr["username"], ssh_keyfile=attr["auth"]
+            )
+            host.slots = attr.get("slots", 1)
```
chmeliik, line 323: Could `slots` be set only in `__init__`, like all the other attributes?

atomic_reactor/utils/resource_manager.py
```diff
@@ -76,6 +77,27 @@ def slots(self):
     def slots(self, num: int):
         self._slots = num
 
+    @property
+    def slots_dir(self):
+        if self._slots_dir is None:
+            with self._ssh_session() as client:
+                home, _, _ = self._ssh_run_remote_cmd(client, "pwd")
+            self._slots_dir = os.path.join(home, SLOTS_RELATIVE_PATH)
```
chmeliik, line 87: Huh, so the `resource_manager/` directory is in the current user's $HOME? That seems like it might cause issues. Would it be possible to use some system location?

How does the directory get created? Does the remote-hosts playbook do that?

atomic_reactor/utils/resource_manager.py
qixiang: Good point, I was not aware of that.

atomic_reactor/utils/resource_manager.py
qixiang: > Huh, so the `resource_manager/` directory is in the current user's $HOME? That seems like it might cause issues. Would it be possible to use some system location?

Yes, there could be some issues especially while using network-based accounts, e.g. ldap users but home is not created or mounted.

> How does the directory get created? Does the remote-hosts playbook do that?

We're current creating local users with default home directory created automatically in ansible role, so I'd leave it as it is now and find a more reliable solution when we need to support non-local users.

atomic_reactor/utils/resource_manager.py
qixiang: Easy to change but I'm thinking on having a new `Slot` class and only expose necessary APIs after we make a deal with solution to the locking issues, will probably rework this part of code.

chmeliik: Ack :+1:

atomic_reactor/utils/resource_manager.py
chmeliik: Seems reasonable. What about the `resource_manager/` directory itself? It looks like the code expects this directory to already exist, does something create it?

qixiang: oops, you're right, good catch. what do you think about creating these slot files as `.osbs_slot_{num}.lock` in home directory and don't need to worry about creating the placeholder dir?

chmeliik: Hmm, I liked having a dedicated directory for them, but it's not super important to me. Maybe they don't need to be hidden files? It sort of "feels right" to keep them visible, can't quite explain why though...

atomic_reactor/utils/resource_manager.py
```diff
@@ -67,21 +71,62 @@ def username(self):
...
-        with open(RESOURCE_LOCK_SCHEMA, 'r') as f:
-            schema_data = f.read()
-        schema = json.loads(schema_data)
         try:
-            jsonschema.validate(payload, schema)
-        except jsonschema.ValidationError as err:
-            logger.warning("Invalid json file: %s", err)
+            validate_with_schema(payload, RESOURCE_LOCK_SCHEMA)
+        except OsbsValidationException as e:
+            logger.warning("Invalid json file: %s", str(e))
```
MartinBasti, line 123: just nitpick, but isn't `%s` in format string doing `str()` call? So `str(e)` is duplicated

ben-alkov:
> Not sure why `black` is not recommended, I don't find any coding style doc for osbs, it's a little hard to keep consistent coding style without some tools like `black`, even with careful code reviews. Anyway, I have reverted the `black` format change, we can revisit this topic later, it's not a big deal at this moment.

Because we (Container Build) have not made the collective decision to use 'black', which is *incredibly* invasive. It's true that we don't have an explicit style guide, but black violates many of our existing conventions. We *might* revisit this in the future, but right now we have *many* more pressing matters than reviewing formatting changes.

qixiang:
I'm still working this, but feel free to do some early reviews, I was trying to refactor the code to have a new class `HostSlot` but it turned out to be hard to manage the ssh session in that way and is inefficient, so I gave up on that change. I didn't some basic testing locally and adding more unit tests.

And updated PR summary to list the main changes.

atomic_reactor/utils/resource_manager.py
```diff
@@ -7,22 +7,19 @@
...
+from typing import List, Optional
 
-RESOURCE_LOCK_SCHEMA = 'schemas/resource-lock-content.json'
 SSH_COMMAND_TIMEOUT = 45
-TIMEOUT_FAIL_BUILD = 3000
-RELATIVE_PATH = 'resource_manager'
-RESOURCE_INFO_JSON = 'info.json'
+TIMEOUT_FAIL_BUILD = 30
+FLOCK_TIMEOUT = 10
+FLOCK_CMD = f"flock -w {FLOCK_TIMEOUT}"
```
qixiang, line 21: Unused, will remove it.

atomic_reactor/utils/resource_manager.py
```diff
@@ -67,21 +55,131 @@ def username(self):
...
+        if self._slots_dir is None:
+            home, _, _ = self._run("pwd")
+            self._slots_dir = os.path.join(home, SLOTS_RELATIVE_PATH)
+        return self._slots_dir
+
+    def _get_slot_path(self, slot_id: int) -> str:
+        return os.path.join(self.slots_dir, f"slot_{slot_id}")
+
+    def _is_valid_slot_id(self, slot_id: int) -> bool:
+        valid_slots = range(self.slots)
```
qixiang, line 69:
```suggestion
        valid_slots = list(range(self.slots))
```

qixiang:
Explain on starting a new ssh session for every remote command: it doesn't make sense to force the caller to start a session and reusing this session between operations, the APIs will be inconvenient to use and the caller need to handle possible exceptions.

There is another possible approach, keeping an ssh session for a `RemoteHost` instance in its lifecycle, and only close it when the instance goes away (in `__del__`), but that's probably not a good idea either, it can cause issues when cyclic references are introduced in future code, developer need to be careful with that.

atomic_reactor/utils/resource_manager.py
chmeliik: Wouldn't leaving at as a `range` object be better? The main purpose of valid_slots is to check if slot_id is in bounds, and range objects are perfect for that (they even have an efficient `__contains__` method)
```python
In [11]: 2**29 in range(2**30)  # don't convert to list unless you want your REPL to get oom-killed
Out[11]: True
```

atomic_reactor/utils/resource_manager.py
```diff
@@ -270,45 +289,63 @@ def from_config(cls, config: dict):
...
+
+        # Sort list based on ratio of available_slots/all_slots
+        resources.sort(key=lambda x: len(x[1])/x[0].slots, reverse=True)
 
-        # Try to lock resources
+        # Try to lock a remote host slot for pipelinerun
         for host, slots in resources:
             for slot in slots:
                 try:
                     host.lock(slot, prid)
```
qixiang, line 342: TODO: check the result.

atomic_reactor/utils/resource_manager.py
chmeliik: Ah, makes sense. Well, it's not a big deal to convert it to list right away, since slot counts will be small

atomic_reactor/utils/resource_manager.py
```diff
@@ -67,21 +55,131 @@ def username(self):
...
+    @property
+    def is_operational(self) -> bool:
+        """ Check whether this host is operational """
+        try:
+            _, _, code = self._run(f"mkdir -p {self.slots_dir}")
+        except Exception as e:
+            logger.exception("%s: host is not operational: %s", self.hostname, e)
+            return False
+        if code != 0:
+            logger.error("%s: cann't prepare slots directory", self.hostname)
```
MartinBasti, line 85:
```suggestion
            logger.error("%s: cannot prepare slots directory", self.hostname)
```

MartinBasti (commented):
LGTM so far :)  :rocket:

atomic_reactor/utils/resource_manager.py
```diff
@@ -270,45 +289,63 @@ def from_config(cls, config: dict):
...
                     # Specific exceptions should be handled in nested methods
-                    logger.warning("Locking failed on slot %s, host %s - %s", slot, host, ex)
+                    logger.warning("%s: unable to lock slot %s for pipelinerun %s: %s",
+                                   host.hostname, slot, prid, e)
                 else:
                     return LockedResource(host, slot, prid)
 
-        raise LockRetry(f'No remote host resources were available for pipelinerun ID {prid}')
+        logger.info("Cannot find remote host resource for pipelinerun %s", prid)
+        return None
```
chmeliik, line 351: Do we not want to retry the `lock_resource()` operation anymore? Or do you think it should be up to the code that uses this module?

chmeliik (commented):
Left a few comments that should apply even if this is reworked with the flock context manager solution

atomic_reactor/utils/resource_manager.py
```diff
@@ -31,29 +28,20 @@
...
         """
         :param hostname: str, remote hostname for ssh connection
         :param username: str, username for ssh connection
         :param ssh_keyfile: str, filepath to ssh private key
+        :param slots: int, number of max allowed slots on remote host
         """
         self._hostname = hostname
         self._username = username
         self._ssh_keyfile = ssh_keyfile
+        self.slots = slots
```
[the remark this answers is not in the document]
qixiang: Addressed in #1808

atomic_reactor/utils/resource_manager.py
```diff
@@ -67,21 +55,131 @@ def username(self):
...
-        """ Validate resource lock json file against schema """
-        with open(RESOURCE_LOCK_SCHEMA, 'r') as f:
-            schema_data = f.read()
-        schema = json.loads(schema_data)
+    @property
+    def slots_dir(self):
+        if self._slots_dir is None:
+            home, _, _ = self._run("pwd")
+            self._slots_dir = os.path.join(home, SLOTS_RELATIVE_PATH)
+        return self._slots_dir
```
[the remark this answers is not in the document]
qixiang: Addressed in #1808

atomic_reactor/utils/resource_manager.py
qixiang: Addressed in #1808

atomic_reactor/utils/resource_manager.py
qixiang: Not sure, I'll revisit this and do it in future patch after #1808 is merged.

qixiang: Not sure, I'll revisit this and do it in future patch after #1808 is merged.

qixiang:
There is a new PR #1808 opened with the resource_manager module refactored, please check that PR instead. I'll leave this open for a while before that new proposal is accepted.

MartinBasti:
obsoleted by #1808

PR #1803, opened 2022-02-11
Fix workflow data serialization

tkdchen:

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -9,6 +9,7 @@ of the BSD license. See the LICENSE file for details.
 Script for building docker image. This is expected to run inside container.
 """
 
+import base64
 import functools
 import json
 import logging
@@ -19,7 +20,7 @@ import time
 import re
 from dataclasses import dataclass, field, fields
 from textwrap import dedent
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, Final, List, Optional, Union
 
 from atomic_reactor.dirs import ContextDir, RootBuildDir
 from atomic_reactor.plugin import (
@@ -200,7 +201,7 @@ class BuildResult(ISerializer):
             source_docker_archive=data.get("source_docker_archive"),
         )
 
-    def dump(self) -> Dict[str, Any]:
+    def as_dict(self) -> Dict[str, Any]:
         return {
             "annotations": self.annotations,
             "fail_reason": self.fail_reason,
@@ -279,7 +280,7 @@ class TagConf(ISerializer):
         """
         return self._floating_images
 
-    def add_primary_image(self, image: str) -> None:
+    def add_primary_image(self, image: Union[str, "ImageName"]) -> None:
         """add new primary image
 
         :param image: str, name of image (e.g. "namespace/httpd:2.4")
@@ -287,7 +288,7 @@ class TagConf(ISerializer):
         """
         self._primary_images.append(ImageName.parse(image))
 
-    def add_unique_image(self, image: str) -> None:
+    def add_unique_image(self, image: Union[str, "ImageName"]) -> None:
         """add image with unpredictable name
 
         :param image: str, name of image (e.g. "namespace/httpd:2.4")
@@ -295,7 +296,7 @@ class TagConf(ISerializer):
         """
         self._unique_images.append(ImageName.parse(image))
 
-    def add_floating_image(self, image: str) -> None:
+    def add_floating_image(self, image: Union[str, "ImageName"]) -> None:
         """add image with floating name
 
         :param image: str, name of image (e.g. "namespace/httpd:2.4")
@@ -315,7 +316,7 @@ class TagConf(ISerializer):
     @classmethod
     def load(cls, data: Dict[str, Any]):
         tag_conf = TagConf()
-        image: str
+        image: ImageName
         for image in data.get("primary_images", []):
             tag_conf.add_primary_image(image)
         for image in data.get("unique_images", []):
@@ -324,11 +325,11 @@ class TagConf(ISerializer):
             tag_conf.add_floating_image(image)
         return tag_conf
 
-    def dump(self) -> Dict[str, Any]:
+    def as_dict(self) -> Dict[str, Any]:
         return {
-            "primary_images": [image.to_str() for image in self._primary_images],
-            "unique_images": [image.to_str() for image in self._unique_images],
-            "floating_images": [image.to_str() for image in self.floating_images],
+            "primary_images": self.primary_images,
+            "unique_images": self.unique_images,
+            "floating_images": self.floating_images,
         }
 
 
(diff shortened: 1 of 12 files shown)


mkosiarc (approved):

chmeliik (commented):
There are still some plugins that return incompatible types:
* check_and_set_platforms returns a `set`
* something in the group_manifests result is `bytes`

Tested by running unit tests with this modification:
```diff
diff --git a/atomic_reactor/plugin.py b/atomic_reactor/plugin.py
index bf5a26fb..37425a05 100644
--- a/atomic_reactor/plugin.py
+++ b/atomic_reactor/plugin.py
@@ -278,6 +278,12 @@ class PluginsRunner(object):
                                                                    plugin.conf)
                 self.save_plugin_timestamp(plugin.plugin_class.key, start_time)
                 plugin_response = plugin_instance.run()
+
+                import json
+                from atomic_reactor.inner import WorkflowDataEncoder
+
+                json.dumps(plugin_response, cls=WorkflowDataEncoder)
+
                 plugin_successful = True
                 if buildstep_phase:
                     assert isinstance(plugin_response, atomic_reactor.inner.BuildResult)
```
Not all plugins use PluginsRunner for their unit tests, but I checked the rest and they looked OK

atomic_reactor/inner.py
```diff
@@ -324,11 +324,11 @@ def load(cls, data: Dict[str, Any]):
...
 
-    def dump(self) -> Dict[str, Any]:
+    def as_dict(self) -> Dict[str, Any]:
         return {
-            "primary_images": [image.to_str() for image in self._primary_images],
-            "unique_images": [image.to_str() for image in self._unique_images],
-            "floating_images": [image.to_str() for image in self.floating_images],
+            "primary_images": self.primary_images,
+            "unique_images": self.unique_images,
+            "floating_images": self.floating_images,
```
chmeliik, line 331: Should the `load()` method be changed accordingly? (or at least the type annotations for it: images are now `ImageName`, not `str`)

atomic_reactor/inner.py
```diff
@@ -538,7 +536,53 @@ def save(self, context_dir: ContextDir) -> None:
 
         logger.info("Writing workflow data into %s", context_dir.workflow_json)
         with open(context_dir.workflow_json, "w+") as f:
-            json.dump(self.dump(), f)
+            json.dump(asdict(self), f, cls=WorkflowDataEncoder)
```
chmeliik, line 539: As you mentioned, `dataclasses.asdict()` converts nested data classes recursively. That could cause issues if some of the WorkflowData attributes were dataclasses

tests/tasks/test_plugin_based.py
```diff
@@ -129,115 +133,117 @@ def test_execute_returns_failure(self, task_with_mocked_deps, caplog):
...
-        if build_result == "normal_return":
-            (flexmock(plugin_based.inner.DockerBuildWorkflow)
-             .should_receive("build_docker_image")
-             .and_return(BuildResult(logs=["Build successfully."])))
 
+@pytest.mark.parametrize(
+    "build_result",
+    ["normal_return", "error_raised", "failed", "terminated"]
+)
+def test_ensure_workflow_data_is_saved_in_various_conditions(
```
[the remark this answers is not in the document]
tkdchen: During the work of this PR, I encountered the problem I asked in the chat. As part of the fix, I moved these two tests as standalone test methods at the module level.

atomic_reactor/inner.py
tkdchen: Good catch. The magic is `ImageName.parse` actually accepts either `str` or `ImageName` object. Changing the type annotation should be enough I think.

tkdchen:
@chmeliik Is there any reason the return value from `check_and_set_platforms` must be a set? I'm thinking whether we can make this plugin return a list instead, since value of platforms is just used as a list in the codebase (if I didn't miss something :).

Meanwhile, by considering supporting the serialization for set object, it may make things complicated. The JSONEncoder is the one osbs relies on to convert Python objects to a proper type in order to be `json.dumps`ed. We can convert a set object into a dict like `{"__type__": "set", "value": list(set object)}`, but if that set object itself contains other nested non-JSON serializable objects, JSONEncoder do not give any further chance to handle them. As we talked in the chat, plugin's return value should be in JSON-compatible data type as much as possible. For the case of plugin `check_and_set_platforms`, list should be good enough to contain the platforms. WDYT?

atomic_reactor/inner.py
```diff
@@ -521,10 +509,20 @@ def load_from_dir(cls, context_dir: ContextDir) -> "ImageBuildWorkflowData":
...
         with open(context_dir.workflow_json, "r") as f:
-            workflow_data = json.load(f)
+            data = f.read()
+
+        workflow_data = json.loads(data)
         validate_with_schema(workflow_data, "schemas/workflow_data.json")
-        return cls.load(workflow_data)
+
+        # NOTE: json.loads twice since the data is validated at the first time.
+        workflow_data = json.loads(data, object_hook=workflow_data_object_hook)
```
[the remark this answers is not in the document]
tkdchen: Fixed.

atomic_reactor/inner.py
```diff
@@ -538,7 +536,53 @@ def save(self, context_dir: ContextDir) -> None:
...
+            return data
+        elif isinstance(o, ImageName):
+            return {
+                "__type__": o.__class__.__name__,
+                "str": o.to_str(),
+            }
+        return super().default(o)
+
+
+def create_image_name(data: Dict[str, Optional[str]]) -> ImageName:
```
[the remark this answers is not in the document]
tkdchen: Fixed.

atomic_reactor/inner.py
tkdchen: Restored the ISerializer interface for workflow data object.

chmeliik:
@tkdchen +1, let's change the check_and_set_platforms return value to a list

tests/tasks/test_plugin_based.py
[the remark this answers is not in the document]
chmeliik: Ack

atomic_reactor/inner.py
```diff
@@ -538,7 +540,63 @@ def save(self, context_dir: ContextDir) -> None:
...
+        elif isinstance(o, ImageName):
+            return {
+                "__type__": o.__class__.__name__,
+                "str": o.to_str(),
+            }
+        elif isinstance(o, bytes):
+            return {
+                "__type__": "bytes",
+                "value": base64.b64encode(o).decode(),
+            }
```
chmeliik, line 564: Instead of supporting bytes, could we change the the group_manifest plugin to return a regular string instead? I think the the `bytes` part of the result is only the 'manifest' key here https://github.com/containerbuildsystem/atomic-reactor/blob/e495e668e658b4f05bacc0f2c9c7ee57613a2e48/atomic_reactor/plugins/post_group_manifests.py#L137

and it should be OK to convert to a string

chmeliik (approved):
LGTM :+1:

The group_manifests change can be done separately

rcerven (approved):

PR #1807, opened 2022-02-15
Make group_manifests return manifest in str type

tkdchen:
It's unnecessary to keep the manifest in bytes in the plugin's return
value. With this change, the workflow data encoder and decoder do not
need to serialize bytes object.

By the way, generally, a plugin's return value should be JSON-compatible
as much as possible, unless it must be and the aforementioned encoder
and decoder have to be updated accordingly.


chmeliik (approved):
LGTM :+1:

hjmodi (approved):

rcerven (approved):

PR #1808, opened 2022-02-16
Refactor resource manager module

qixiang:
1. Slot file content is a single line in format of "prid@timestamp".
   Empty slot file means slot is free.

2. While performing any operations against a slot which requires
   locking the slot first, there are two ssh sessions opened to
   remote host, one is for keeping the lock on that slot, the other
   one is for running arbitrary commands, we only use it to run
   commands to read and write the slot in public APIs.

   There are two files exist on host for each slot:

   [1] /path/to/slots/dir/slot_{id}: the slot file, contains data
       of lock info
   [2] /path/to/slots/dir/slot_{id}.lock: lock file of slot, before
       writing to a slot, the lock should be acquired first

The `RemostHost` and `RemoteHostsPool` classes are the main classes
that provide the public APIs.

Example usages:

    host = RemoteHost(
        hostname="remote-host-1", username="user",
        ssh_keyfile="/path/to/key", slots=3
    )

    # check host is operational
    operational = host.is_operational

    # check a slot on host is in free state
    host.is_free(2)

    # lock a slot on host for a pipelinerun
    host.lock(2, "pipelinerun-123")

    # unlock a slot on host owned by a pipelinerun
    host.unlock(2, "pipelinerun-123")

    pool = RemoteHostsPool(hosts=[host1, ...])

    # lock a slot from hosts in the pool, an available
    # free slot will be picked and locked
    pool.lock_resource("pipelinerun-123")

* CLOUDBLD-8553

Signed-off-by: Qixiang Wan <qwan@redhat.com>

### DIFF
--- a/atomic_reactor/utils/resource_manager.py
+++ b/atomic_reactor/utils/resource_manager.py
@@ -7,53 +7,154 @@ of the BSD license. See the LICENSE file for details.
 """
 
 import backoff
-import json
-import jsonschema
 import logging
+import os
 import paramiko
 import random
-import os
+import time
 from contextlib import contextmanager
 from datetime import datetime
+from functools import cached_property
 from shlex import quote
-from typing import List
+from typing import List, Optional, Tuple
+from paramiko.channel import ChannelFile  # just for type annotation
 
-RESOURCE_LOCK_SCHEMA = 'schemas/resource-lock-content.json'
-SSH_COMMAND_TIMEOUT = 45
-TIMEOUT_FAIL_BUILD = 3000
-RELATIVE_PATH = 'resource_manager'
-RESOURCE_INFO_JSON = 'info.json'
+SSH_COMMAND_TIMEOUT = 30
+SLOTS_RELATIVE_PATH = "osbs_slots"
 RETRY_ON_SSH_EXCEPTIONS = (paramiko.ssh_exception.NoValidConnectionsError,
                            paramiko.ssh_exception.SSHException)
-BACKOFF_FACTOR = 5
+BACKOFF_FACTOR = 3
 MAX_RETRIES = 3
 
 logger = logging.getLogger(__name__)
 
-
-class ResourceManagerError(RuntimeError):
-    """ Base module exception """
-
-
-class LockError(ResourceManagerError):
-    """ Attempt for resource locking returned non-zero exitcode """
+__all__ = [
+    "RemoteHost",
+    "RemoteHostsPool",
+    "LockedResource",
+]
 
 
-class LockRetry(ResourceManagerError):
-    """ Retry locking, no resources were available """
+class RemoteHostError(RuntimeError):
+    pass
+
+
+class SlotLockError(RemoteHostError):
+    pass
+
+
+class SlotReadError(RemoteHostError):
+    pass
+
+
+class SlotWriteError(RemoteHostError):
+    pass
+
+
+class SSHRetrySession(paramiko.SSHClient):
+    """ paramiko SSHClient with retry mechanism """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    @backoff.on_exception(
+        backoff.expo,
+        RETRY_ON_SSH_EXCEPTIONS,
+        factor=BACKOFF_FACTOR,
+        max_tries=MAX_RETRIES,
+        jitter=None,  # use deterministic backoff, do not apply random jitter
+        logger=logger,
+    )
+    def exec_command(self, *args, **kwargs):
+        return super().exec_command(*args, **kwargs)
+
+    @backoff.on_exception(
+        backoff.expo,
+        RETRY_ON_SSH_EXCEPTIONS,
+        factor=BACKOFF_FACTOR,
+        max_tries=MAX_RETRIES,
+        jitter=None,  # use deterministic backoff, do not apply random jitter
+        logger=logger,
+    )
+    def connect(self, *args, **kwargs):
+        super().connect(*args, **kwargs)
+
+    def run(self, cmd: str) -> Tuple[str, str, int]:
+        _, stdout, stderr = self.exec_command(cmd, timeout=SSH_COMMAND_TIMEOUT)
+        out = stdout.read().decode().strip()
+        err = stderr.read().decode().strip()
+        code = stdout.channel.recv_exit_status()
+        return out, err, code
+
+
+class SlotData:
+
+    def __init__(self, prid: Optional[str] = None, timestamp: Optional[str] = None):
+        """ Instantiate slot data with values of prid and timestamp """
+        # A valid slot contains empty content or content in format:
+        # "prid@timestamp"
+        # prid: pipelinerun id
+        # timestamp: datetime string in iso format
+        self.prid = prid
+        self.timestamp = timestamp
+
+    @classmethod
+    def from_string(cls, string: Optional[str]):
+        """ Instantiate from a string """
+
+        # We don't validate the string here, call is_valid to check slot data
+        if not string:
+            return cls()
+
+        values = string.split("@")
+        prid = values[0]
+        timestamp = "".join(values[1:])
+        return cls(prid=prid, timestamp=timestamp)
+
+    @property
+    def is_empty(self):
+        return not any((self.prid, self.timestamp))
+
+    @property
+    def is_valid(self):
+        # Empty slot data is valid
+        if self.is_empty:
+            return True
+
+        # String of prid cannot contain "@"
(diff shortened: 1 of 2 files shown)


atomic_reactor/utils/resource_manager.py
```diff
@@ -67,186 +166,378 @@ def username(self):
...
-        :param prid: str, Pipeline run ID
-        """
+        self.host = host
         self.slot = slot
         self.prid = prid
-        self.host = host
 
     def unlock(self):
+        """ Unlock the resource for pipelinerun """
         self.host.unlock(self.slot, self.prid)
```
chmeliik, line 541: This ignores the return value of `host.unlock()`.

Which brings me to the next point - since `LockedResource.unlock()` and `RemoteHost.lock()|unlock()` are part of the public API of this module, maybe returning booleans to indicate failure is not the best way to go. It could be easy to forget to check the return value and assume that calling `unlock()` (and not getting an exception) means the slot is unlocked, kind of like it happened here.

atomic_reactor/utils/resource_manager.py
```diff
@@ -270,45 +561,64 @@ def from_config(cls, config: dict):
...
+            logger.error("There is no remote host slot available for pipelinerun %s", prid)
+            return None
+
+        # Sort list based on ratio of available_slots/all_slots
+        resources.sort(key=lambda x: len(x[1])/x[0].slots, reverse=True)
 
-        # Try to lock resources
+        # Try to lock a remote host slot for pipelinerun
         for host, slots in resources:
             for slot in slots:
```
chmeliik, line 612: Do we still want try the slots in a random order?

atomic_reactor/utils/resource_manager.py
```diff
@@ -270,45 +561,64 @@ def from_config(cls, config: dict):
...
-        """ Lock resources for build """
+    def lock_resource(self, prid: str) -> Optional[LockedResource]:
+        """
+        Lock resource for a pipelinerun
+
+        :param prid: str, pipelinerun ID
+        """
+
+        if not self.hosts:
+            logger.error("This is no available remote host in pool")
```
chmeliik, line 591:
```suggestion
            logger.error("There is no available remote host in pool")
```

atomic_reactor/utils/resource_manager.py
```diff
@@ -7,53 +7,152 @@
...
-BACKOFF_FACTOR = 5
+BACKOFF_FACTOR = 3
 MAX_RETRIES = 3
 
 logger = logging.getLogger(__name__)
 
+__all__ = [
+    "RemoteHost",
+    "RemoteHostsPool",
+]
```
chmeliik, line 33: Should `LockedResource` also be public API? Since that is the return value of pool.lock_resource()?

chmeliik (commented):
Nice, great commit message! :tada: 

I do have a couple minor suggestions, overall I like this :+1:

atomic_reactor/utils/resource_manager.py
```diff
@@ -67,186 +166,378 @@ def username(self):
...
+
+        _errmsg = f"{self.hostname}: failed to acquire lock on slot {slot_id}"
+        try:
+            logger.info("%s: acquiring lock on slot %s", self.hostname, slot_id)
+            stdin, stdout, stderr = session.exec_command(cmd)
+        except Exception as ex:
+            raise SlotLockError(_errmsg) from ex
+
+        # A short time sleep to wait for the socket to be ready
+        time.sleep(0.1)
```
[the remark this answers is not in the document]
qixiang: yes, it could happen when the socket is not ready for write, in my testing `time.sleep(0.01)` is not good enough, so I set it to `0.1` second and seems working without issue. And yes, you're right it also happens when slot is locked by others.

atomic_reactor/utils/resource_manager.py
```diff
@@ -67,186 +166,378 @@ def username(self):
...
+            if not stdout.readline():
+                if stdout.channel.recv_exit_status() == 42:
+                    _errmsg = f"{_errmsg}: slot is locked by others"
+                else:
+                    _errmsg = f"{_errmsg}: no output from cat command"
+                logger.debug(_errmsg)
+                raise SlotLockError(_errmsg)
+
+        # So far so good, the session is blocked there with keeping the slot lock
+        return session, stdin, stdout, stderr
```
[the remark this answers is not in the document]
qixiang: right, I'll test and change this.

atomic_reactor/utils/resource_manager.py
qixiang: It's unnecessary, there is no difference among slots on the same host except the number, so the try order doesn't matter.

atomic_reactor/utils/resource_manager.py
qixiang: It should, but I don't see a case of using it at this moment, so I didn't export it, I'll add it to list.

atomic_reactor/utils/resource_manager.py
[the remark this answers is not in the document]
chmeliik: Ack, seems reasonable

atomic_reactor/utils/resource_manager.py
qixiang: This concern makes sense, the problem is we can't do anything when `unlock` failed, so I just ignore it silently, there could be the following cases that `unlock` fails:
1. slot is not owned by this pr, in theory, it should never happen.
2. slot is not locked, in theory it should never happen.
3. slot operations failed to due connection errors.

when I say "it should never happen", it doesn't count the situations of system reboots, slot be changed by people manually.
For case 1, we should just ignore that, and let clean up jobs to fix bad slots.
For case 2, we can also ignore the return value, the free slot will be used by other pipelinerun.
For case 3, maybe we should add retry policy to `host.lock` and `host.unlock` to handle it, and can still ignore the returned value, wdyt?

atomic_reactor/utils/resource_manager.py
```diff
@@ -67,186 +166,378 @@ def username(self):
...
+        self.id = slot_id
+        self.path = os.path.join(self.host.slots_dir, f"slot_{slot_id}")
 
-            if exitcode != 0:
-                raise LockError(f'Attempt to lock slot {slot_id} with pipelinerun ID {prid} '
-                                'returned non-zero exitcode')
+    @property
+    def _data(self) -> SlotData:
+        content = self._read()
+        return SlotData.from_string(content)
```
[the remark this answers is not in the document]
qixiang: Yes, it impacts the performance, currently I think it's acceptable, this will make it easy implement other APIs without worrying about inconsistent data, I'll do some testing and thinking on how can do this better, but probably not in this PR.

atomic_reactor/utils/resource_manager.py
chmeliik: For case 3, that should already be handled by the `SSHRetrySession`, right? I think the behavior is fine, my only concern is that we are using booleans to indicate the success/failure of an operation which is normally expected to succeed - that is what exceptions are for.

But it's not a dealbreaker for me. Thinking about the integration with the build code, we will only really use `RemoteHostPool.lock_resource()` and `LockedResource.unlock()` anyway, and those pretty much behave as expected. Like you said, there's not much we can do with `unlock()` failures except ignore them, anyway.

atomic_reactor/utils/resource_manager.py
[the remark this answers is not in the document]
chmeliik: Ok, should be fine for now

atomic_reactor/utils/resource_manager.py
qixiang: ah, I see, I didn't realized this is what randomizing trying to resolve, changed in updated version.

atomic_reactor/utils/resource_manager.py
qixiang: > For case 3, that should already be handled by the SSHRetrySession, right?

yes, I should say command errors or so on instead, which won't be captured by SSHRetrySession, they raise `SlotLockError`, `SlotReadError` or `SlotWriteError`.  I've updated patch to retry on these exceptions.

qixiang:
PR is updated, I hope all comments have been addressed except the one performance concern about using `_data` to read slot content, I'll try to find a better solution if possible but won't be part of this PR.

chmeliik (approved):
LGTM :+1:

waiving the rest of the checklist items, config schema changes etc. will be done later

lkolacek (approved):
Looks great! :+1:

MartinBasti (approved):
:rocket:

PR #1809, opened 2022-02-16
Rework metadata *_map decorators

chmeliik:
Previously, the map decorators would extract any number of keys from the
plugin result and save each as a separate annotation or label.

Now, they will save only a single annotation or label, but they allow
the caller to specify a custom transformation to apply to the plugin
result.

Currently, all the plugins that use these decorators only use them with
a single key. This use case is unchanged, the default transformation is
simply result[key].


chmeliik:
PTAL, this should be useful for moving annotations and labels to earlier plugins (I did most of this while investigating the Tekton results design, figured I might as well finish it).

This is how you would use this for e.g. the _remote_sources_ annotation:
```diff
diff --git a/atomic_reactor/plugins/pre_resolve_remote_source.py b/atomic_reactor/plugins/pre_resolve_remote_source.py
index 6ec191e5..37b9bab5 100644
--- a/atomic_reactor/plugins/pre_resolve_remote_source.py
+++ b/atomic_reactor/plugins/pre_resolve_remote_source.py
@@ -25,6 +25,7 @@ from atomic_reactor.constants import (
     REMOTE_SOURCE_TARBALL_FILENAME,
 )
 from atomic_reactor.dirs import BuildDir
+from atomic_reactor.metadata import annotation_map
 from atomic_reactor.plugin import PreBuildPlugin
 from atomic_reactor.util import is_scratch_build, map_to_user_params
 from atomic_reactor.utils.cachito import CFG_TYPE_B64
@@ -63,6 +64,10 @@ class RemoteSource:
             return REMOTE_SOURCE_JSON_FILENAME
 
 
+@annotation_map(
+    "remote_sources",
+    transform=lambda result: [{"name": item["name"], "url": item["url"]} for item in result]
+)
 class ResolveRemoteSourcePlugin(PreBuildPlugin):
     """Initiate a new Cachito request for sources
```

ben-alkov (approved):
:taco:

rcerven (approved):

PR #1810, opened 2022-02-16
Add methods to work with built images

mkosiarc:
CLOUDBLD-8247

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -33,8 +33,6 @@ from atomic_reactor.plugin import (
 )
 from atomic_reactor.constants import (
     DOCKER_STORAGE_TRANSPORT_NAME,
-    INSPECT_ROOTFS,
-    INSPECT_ROOTFS_LAYERS,
     PLUGIN_BUILD_ORCHESTRATE_KEY,
     REACTOR_CONFIG_FULL_PATH,
     DOCKERFILE_FILENAME,
@@ -427,8 +425,6 @@ class ImageBuildWorkflowData(ISerializer):
     koji_source_manifest: Dict[str, Any] = field(default_factory=dict)
 
     buildargs: Dict[str, str] = field(default_factory=dict)  # --buildargs for container build
-    built_image_inspect: Dict[str, Any] = field(default_factory=dict)
-    layer_sizes: List[Dict[str, Union[str, int]]] = field(default_factory=list)
 
     # When an image is exported into tarball, it can then be processed by various plugins.
     #  Each plugin that transforms the image should save it as a new file and append it to
@@ -856,21 +852,6 @@ class DockerBuildWorkflow(object):
             except PluginFailedException as ex:
                 logger.error("one or more prepublish plugins failed: %s", ex)
                 raise
-
-            if self.data.build_result.is_image_available():
-                # OSBS2 TBD
-                self.data.built_image_inspect = imageutil.inspect_built_image()
-                # OSBS2 TBD
-                history = imageutil.get_image_history(self.data.image_id)
-                diff_ids = self.data.built_image_inspect[INSPECT_ROOTFS][INSPECT_ROOTFS_LAYERS]
-
-                # diff_ids is ordered oldest first
-                # history is ordered newest first
-                # We want layer_sizes to be ordered oldest first
-                self.data.layer_sizes = [
-                    {"diff_id": diff_id, "size": layer['Size']}
-                    for (diff_id, layer) in zip(diff_ids, reversed(history))
-                ]
 
             try:
                 postbuild_runner.run()
--- a/atomic_reactor/schemas/workflow_data.json
+++ b/atomic_reactor/schemas/workflow_data.json
@@ -33,11 +33,6 @@
     "koji_source_manifest": {"type": ["object", "null"]},
 
     "buildargs": {"type": "object"},
-    "built_image_inspect": {"type": ["object", "null"]},
-    "layer_sizes": {
-      "type": "array",
-      "items": {"$ref": "#/definitions/layer_size"}
-    },
 
     "exported_image_sequence": {
       "type": "array",
@@ -68,8 +63,7 @@
     "plugin_workspace",
     "plugins_timestamps", "plugins_durations", "plugins_errors", "build_canceled", "plugin_failed",
     "reserved_build_id", "reserved_token", "koji_source_nvr", "koji_source_source_url", "koji_source_manifest",
-    "buildargs", "built_image_inspect", "layer_sizes",
-    "exported_image_sequence", "files", "image_components", "all_yum_repourls",
+    "buildargs", "exported_image_sequence", "files", "image_components", "all_yum_repourls",
     "annotations", "labels", "image_id", "parent_images_digests",
     "build_result"
   ],
@@ -184,15 +178,6 @@
         "source_docker_archive",
         "__type__"
       ],
-      "additionalProperties": false
-    },
-    "layer_size": {
-      "type": "object",
-      "properties": {
-        "diff_id": {"type": "string"},
-        "size": {"type": "integer", "minimum": 0}
-      },
-      "required": ["diff_id", "size"],
       "additionalProperties": false
     },
     "manifest_digest": {
--- a/atomic_reactor/utils/imageutil.py
+++ b/atomic_reactor/utils/imageutil.py
@@ -7,13 +7,22 @@ of the BSD license. See the LICENSE file for details.
 """
 
 import functools
-from typing import Optional, Union
+import subprocess
+import logging
+import tarfile
+import json
+
+from typing import Optional, Union, Dict, List, Any
+from pathlib import Path
 
 from osbs.utils import ImageName
 
 from atomic_reactor import config
 from atomic_reactor import util
 from atomic_reactor.types import ImageInspectionData
+from atomic_reactor.utils import retries
+
+logger = logging.getLogger(__name__)
 
 
 def image_is_inspectable(image: Union[str, ImageName]) -> bool:
@@ -44,7 +53,7 @@ class ImageUtil:
     def get_inspect_for_image(
         self, image: Union[str, ImageName], platform: Optional[str] = None
     ) -> ImageInspectionData:
-        """Inspect an image. Should mainly be used to inspect parent images.
+        """Inspect an image.
 
         The image must be inspectable, passing a non-inspectable image is an error.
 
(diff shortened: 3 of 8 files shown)


atomic_reactor/utils/imageutil.py
```diff
@@ -92,33 +99,98 @@ def _get_registry_client(self, registry: str) -> util.RegistryClient:
...
+
+    def get_uncompressed_image_layer_sizes(self, path: str) -> Dict[str, int]:
+        """Returns data about the uncompressed image layer sizes
+
+        :param path: str path to a image archive tarball
+        :returns Dict[str, int], where key is layer id and int is the size
+                                 in bytes
+        """
+        unpack_cmd = f"tar xf {path}"
+        layers_cmd = "jq < manifest.json '.[0].Layers[]' -r | xargs du -b"
```
chmeliik, line 139: Coud we do this with python's `tarfile` and `json` modules instead of subprocess calls?

The cool thing is that we probably won't even have to extract the entire archive that way:
```python
with tarfile.open(docker_tarball) as tar:
    manifest = json.load(tar.extractfile("manifest.json"))
    layer_size = tar.getmember(...).size
```
And similarly for the other method

mkosiarc: This does make more sense, I don't know why I reasoned myself out of this approach. updated now

atomic_reactor/inner.py
```diff
@@ -858,19 +858,10 @@ def build_docker_image(self) -> BuildResult:
...
-
-                # diff_ids is ordered oldest first
-                # history is ordered newest first
-                # We want layer_sizes to be ordered oldest first
-                self.data.layer_sizes = [
-                    {"diff_id": diff_id, "size": layer['Size']}
-                    for (diff_id, layer) in zip(diff_ids, reversed(history))
-                ]
+                self.data.built_image_inspect = imageutil.get_inspect_for_image(self.data.image_id)
+                image_archive_tarball_path = "" # TODO
```
mkosiarc, line 862: @chmeliik what would be good way to handle this ? Just create directory in the specific platform build dir ?

chmeliik: I'm thinking the best thing to do might be to remove `built_image_inspect` and `layer_sizes` from WorkflowData entirely. The few plugins that need this info can fetch it using imageutil methods instead of taking it from context data.

If it would be too much trouble to remove at this point, maybe just don't set them here and add `# OSBS2 TBD`s about removing them to the affected plugins

mkosiarc:
Do you have idea what to do with this? 
`atomic_reactor/utils/imageutil.py:140: error: Argument 1 to "load" has incompatible type "Optional[IO[bytes]]"; expected "SupportsRead[Union[str, bytes]]"`

It didn't help if I forced it to be str

atomic_reactor/utils/imageutil.py
```diff
@@ -92,33 +100,71 @@ def _get_registry_client(self, registry: str) -> util.RegistryClient:
...
+        with tarfile.open(src_path) as tar:
+            manifest = json.load(tar.extractfile("manifest.json"))
+            layers = manifest[0]['Layers']
+            if len(layers) > 1:
+                raise ValueError(f'Tarball at {src_path} has more than 1 layer')
 
+            layer_file = tar.getmember(layers[0])
+            tar.extract(layer_file, dst_path)
 
-def get_image(image):
```
mkosiarc, line 121: @acmiel post_compress.py calls this method, is `download_image_archive_tarball` doing the same? or should this method do something different?

atomic_reactor/utils/imageutil.py
```diff
@@ -92,33 +100,71 @@ def _get_registry_client(self, registry: str) -> util.RegistryClient:
...
+
+    def extract_filesystem_layer(self, src_path: str, dst_path: str) -> str:
+        """Extract filesystem layer from image archive tarball and
+        saves it at dst_path. This is meant for flatpacks and will work
+        only when the archive has 1 layer.
+
+        :param src_path: str, path to image archive tarball
+        :param dst_path: str, path where the layer will be copied
+
+        :returns str, the filesystem layer name
```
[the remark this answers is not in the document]
mkosiarc: It does, see ```return layers[0]```, is that just github diff shows it below the removed get_image method

atomic_reactor/utils/imageutil.py
```diff
@@ -92,33 +100,71 @@ def _get_registry_client(self, registry: str) -> util.RegistryClient:
...
+        :param dst_path: str, path where to export file/dir
+        """
+        cmd = ['oc', 'image', 'extract', f'{image}', '--path', f'{src_path}:{dst_path}']
 
-def remove_image(image, force=False):
-    # self.tasker.remove_image(image, force=force)
-    # most likely won't be needed at all
-    return {}
+        try:
+            return subprocess.check_output(cmd, stderr=subprocess.STDOUT)
```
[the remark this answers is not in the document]
mkosiarc: updated, to use retries. That's the right solution here.

atomic_reactor/utils/imageutil.py
```diff
@@ -92,33 +100,71 @@ def _get_registry_client(self, registry: str) -> util.RegistryClient:
...
+        except subprocess.CalledProcessError as e:
+            logger.error("Image archive download failed:\n%s", e.output)
+            raise
+
+    def get_uncompressed_image_layer_sizes(self, path: str) -> List[Dict[str, Union[str, int]]]:
+        """Returns data about the uncompressed image layer sizes
+
+        :param path: str path to a image archive tarball
+        :returns Dict[str, int], where key is layer id and int is the size
+                                 in bytes
```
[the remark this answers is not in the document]
mkosiarc: updated

chmeliik:
> Do you have idea what to do with this? `atomic_reactor/utils/imageutil.py:140: error: Argument 1 to "load" has incompatible type "Optional[IO[bytes]]"; expected "SupportsRead[Union[str, bytes]]"`
> 
> It didn't help if I forced it to be str

I'm guessing you need to get rid of the `Optional` in `Optional[IO[bytes]]`. For example like this:
```python
def does_typecheck() -> dict:
    with tarfile.open("./archive.tar") as tf:
        manifest_file = tf.extractfile("manifest.json")
        if not manifest_file:
            raise ValueError("manifest.json is not a file")
        manifest = json.load(manifest_file)

    return manifest
```
[tarfile.extractfile](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractfile) returns None if the member is not a regular file and raises KeyError if the member is not present at all

atomic_reactor/utils/imageutil.py
chmeliik: I think the only difference is that `download_image_archive_tarball` creates a file on disk while the current usage assumes that `get_image` returns an open binary file. We don't need to keep `get_image` for that, just open the downloaded tarball instead

mkosiarc: Thanks, I updated the todo comment in post_compress.py and removed the call to get_image

chmeliik:
I've been playing around with `oc image extract` a bit locally, and some of the behaviors are potentially surprising (tested with oc version 4.9.0).

#### 1. the dst_path must be an existing empty dir, nothing else works
<details>
<summary>Example</summary>

```shell
$ oc image extract docker.io/library/python:3.10.0-alpine --path /usr/local/bin/pip:/tmp
error: directory /tmp must be empty, pass --confirm to overwrite contents of directory

$ oc image extract docker.io/library/python:3.10.0-alpine --path /usr/local/bin/pip:/tmp/output
error: destination path does not exist: /tmp/output

$ mkdir /tmp/output
$ oc image extract docker.io/library/python:3.10.0-alpine --path /usr/local/bin/pip:/tmp/output
$ ls /tmp/output
pip
```
</details>

#### 2. file permissions are not preserved
<details>
<summary>Example</summary>

```shell
$ podman run --rm -ti docker.io/library/python:3.10.0-alpine sh
/ # stat -c '%A %U:%G' /usr/local/bin/pip
-rwxr-xr-x root:root
/ # exit

$ stat -c '%A %U:%G' /tmp/output/pip 
-rw-r----- acmiel:acmiel
```
</details>

#### 3. nonexistent files are silently skipped
<details>
<summary>Example</summary>

```shell
$ oc image extract docker.io/library/python:3.10.0-alpine --path /usr/bin/pip:/tmp/output
$ echo $?              
0

$ ls /tmp/output -a
.  ..
```
</details>

#### 4. different behavior for `--path /src/dir` vs `--path /src/dir/`
<details>
<summary>Example</summary>

```shell
$ oc image extract docker.io/library/python:3.10.0-alpine --path /usr/local/bin:/tmp/output
$ ls /tmp/output
bin
$ ls /tmp/output/bin
2to3       idle3     pip3     pydoc3     python3            python3-config
2to3-3.10  idle3.10  pip3.10  pydoc3.10  python3.10         python-config
idle       pip       pydoc    python     python3.10-config  wheel
```
```shell
$ oc image extract docker.io/library/python:3.10.0-alpine --path /usr/local/bin/:/tmp/output
$ ls /tmp/output
2to3       idle3     pip3     pydoc3     python3            python3-config
2to3-3.10  idle3.10  pip3.10  pydoc3.10  python3.10         python-config
idle       pip       pydoc    python     python3.10-config  wheel
```
</details>

They don't seem too bad, the limitation of not preserving permissions was already known and we can live with it, the rest is just minor traps for unsuspecting devs. It would be good to document them though.

atomic_reactor/utils/imageutil.py
```diff
@@ -92,33 +100,76 @@ def _get_registry_client(self, registry: str) -> util.RegistryClient:
...
+        :param path: str path to a image archive tarball
+        :return: List[Dict[str, object]], where key is layer id and int is the size
+                                 in bytes
+        """
+        with tarfile.open(path) as tar:
+            manifest_file = tar.extractfile('manifest.json')
+            if not manifest_file:
+                raise ValueError(f'manifest.json from {path} is not a regular file')
+            manifest = json.load(manifest_file)
+            layers = manifest[0]['Layers']
```
chmeliik, line 144: Should we also check that there is exactly one item in the manifest? The docker [spec](https://github.com/moby/moby/blob/master/image/spec/v1.2.md#combined-image-json--filesystem-changeset-format) says there might optionally be entries for the parent images, unfortunately it does not say if the current image will be first or last in the list

atomic_reactor/utils/imageutil.py
```diff
@@ -92,33 +100,76 @@ def _get_registry_client(self, registry: str) -> util.RegistryClient:
...
+            layer_sizes = []
+            for layer in layers:
+                layer_id = f'sha256:{layer.replace(".tar", "")}'
```
chmeliik, line 147: Based on the above-mentioned spec, it seems that the file names of layers should not be relied on. The fact that they match the diff_id seems to be a skopeo implementation detail.

Instead, we could get the diff IDs from `rootfs.diff_ids` in the config file (also present in the docker archive and referenced in manifest.json) and rely on the fact that the `rootfs.diff_ids` array and the `Layers` array are in the same order.

atomic_reactor/utils/imageutil.py
```diff
@@ -92,33 +100,76 @@ def _get_registry_client(self, registry: str) -> util.RegistryClient:
...
+                layer_sizes.append({
+                    "diff_id": layer_id,
+                    "size": tar.getmember(layer).size
+                })
+            return layer_sizes
+
+    def extract_filesystem_layer(self, src_path: str, dst_path: str) -> str:
+        """Extract filesystem layer from image archive tarball and
+        saves it at dst_path. This is meant for flatpacks and will work
```
chmeliik, line 156: typo: flatpaks, not flatpacks

(more like flatpak is the typo, but an intentional one :smile:)

atomic_reactor/utils/imageutil.py
```diff
@@ -92,33 +100,76 @@ def _get_registry_client(self, registry: str) -> util.RegistryClient:
...
+        only when the archive has 1 layer.
+
+        :param src_path: str, path to image archive tarball
+        :param dst_path: str, path where the layer will be copied
+        :return: str, the filesystem layer name
```
chmeliik, line 161: Technically doesn't have to be just a name, it can be a relative path (example from the spec: `a65da33792c5187473faa80fa3e1b975acba06712852d1dea860692ccddf3198/layer.tar`)

Shouldn't be a problem though, the usage is still the same (treat the return value of this method as a relative path within dst_path)

tests/utils/test_imageutil.py
```diff
@@ -37,6 +43,18 @@ def test_inspectable(image, is_inspectable):
     assert imageutil.image_is_inspectable(image) == is_inspectable
 
 
+def mock_tarball(tarball_path, layer_files, manifest_file_content) -> str:
```
chmeliik, line 46: Does not return anything

atomic_reactor/schemas/workflow_data.json
```diff
@@ -33,11 +33,6 @@
     "koji_source_manifest": {"type": ["object", "null"]},
 
     "buildargs": {"type": "object"},
-    "built_image_inspect": {"type": ["object", "null"]},
-    "layer_sizes": {
-      "type": "array",
-      "items": {"$ref": "#/definitions/layer_size"}
```
chmeliik, line 39: Can we also remove the layer_size definition now, or is it still used?

atomic_reactor/utils/imageutil.py
```diff
@@ -92,33 +100,76 @@ def _get_registry_client(self, registry: str) -> util.RegistryClient:
...
+        except subprocess.CalledProcessError as e:
+            logger.error("Image archive download failed:\n%s", e.output)
+            raise
+
+    def get_uncompressed_image_layer_sizes(self, path: str) -> List[Dict[str, object]]:
+        """Returns data about the uncompressed image layer sizes
+
+        :param path: str path to a image archive tarball
+        :return: List[Dict[str, object]], where key is layer id and int is the size
+                                 in bytes
```
chmeliik, line 137: If it's not possible to further specify the value type, it should be `Any` rather than `object`. Not sure if there are any practical differences, but `Any` is the idiomatic way to leave the type unspecified.

And the description of the return value appears outdated

atomic_reactor/schemas/workflow_data.json
mkosiarc: removed

atomic_reactor/utils/imageutil.py
mkosiarc: fixed

tests/utils/test_imageutil.py
mkosiarc: fixed

atomic_reactor/utils/imageutil.py
mkosiarc: updated the docstring slightly

atomic_reactor/utils/imageutil.py
mkosiarc: fixed

atomic_reactor/utils/imageutil.py
mkosiarc: updated

atomic_reactor/utils/imageutil.py
```diff
@@ -92,33 +100,76 @@ def _get_registry_client(self, registry: str) -> util.RegistryClient:
         session = util.RegistrySession.create_from_config(self._conf, registry)
         return util.RegistryClient(session)
 
+    def extract_file_from_image(self, image: str, src_path: str, dst_path: str) -> None:
```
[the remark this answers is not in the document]
mkosiarc: fixed

atomic_reactor/utils/imageutil.py
mkosiarc: I am not sure about this one, from quick look I can't really find anything about how to identify the item we want when there is more items. I'll try to look into it some more

chmeliik (commented):
Mostly LGTM, just that one unresolved comment + it would be nice to mention the points from https://github.com/containerbuildsystem/atomic-reactor/pull/1810#issuecomment-1046759412 in the docstring

atomic_reactor/utils/imageutil.py
tkdchen: It would be also good to separate these lines into a function:

```python
        with tarfile.open(src_path) as tar:
            manifest_file = tar.extractfile('manifest.json')
            if not manifest_file:
                raise ValueError(f'manifest.json from {src_path} is not a regular file')
            manifest = json.load(manifest_file)
            layers = manifest[0]['Layers']
```

or in some other ways in order to not repeat similar logic in two methods.

mkosiarc: @tkdchen I am not sure separate method is necessary, especially know that one function also extracts config.json, but If you feel strongly about this I'll refactor this.

mkosiarc: I didn't find any additional info about having the entries for parent images, so I am not sure how to proceed. I am assuming that our images will have only one entry (I was testing with couple of them), but to avoid any surprises in the future, maybe I could document this in the code and also raise an error?

mkosiarc:
> Mostly LGTM, just that one unresolved comment + it would be nice to mention the points from [#1810 (comment)](https://github.com/containerbuildsystem/atomic-reactor/pull/1810#issuecomment-1046759412) in the docstring

I updated the docstring, and also added some checks that might catch some of the issues.

atomic_reactor/utils/imageutil.py
chmeliik: document and raise error sounds good :+1:

chmeliik (commented):
LGTM :taco: 

(just the one "document and raise error" point left)

atomic_reactor/utils/imageutil.py
mkosiarc: @chmeliik updated

chmeliik (approved):

lkolacek (approved):

PR #1812, opened 2022-02-17
Output from bsi should be decoded in the build result

tkdchen:

### DIFF
--- a/atomic_reactor/plugins/build_source_container.py
+++ b/atomic_reactor/plugins/build_source_container.py
@@ -121,7 +121,7 @@ class SourceContainerPlugin(BuildStepPlugin):
         cmd.append('{}'.format(image_output_dir))
 
         try:
-            output = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+            output = subprocess.check_output(cmd, stderr=subprocess.STDOUT, text=True)
         except subprocess.CalledProcessError as e:
             self.log.error("BSI failed with output:\n%s", e.output)
             return BuildResult(logs=e.output, fail_reason='BSI utility failed build source image')


atomic_reactor/plugins/build_source_container.py
```diff
@@ -155,7 +155,7 @@ def run(self):
         shutil.rmtree(image_output_dir)
 
         return BuildResult(
-            logs=output,
```
chmeliik, line 158: We also return a BuildResult on line 127, shouldn't the output be decoded there as well?

Maybe we could pass `text=True` to the `check_output()` call instead of decoding later?

chmeliik: And how come mypy is not failing? BuildResult defines logs as `Optional[List[str]]`

tkdchen: > And how come mypy is not failing? BuildResult defines logs as `Optional[List[str]]`

@MartinBasti any idea?

chmeliik: Even if we don't figure it out, do we want to do something about it? Fix the type annotation or fix the code to use a list?

chmeliik: https://mypy.readthedocs.io/en/stable/getting_started.html#function-signatures-and-dynamic-vs-static-typing
> By default, mypy will not type check dynamically typed functions. This means that with a few exceptions, mypy will not report any errors with regular unannotated Python.

Yeah, so about that "mypy is checking the entire codebase now" - nope :cry:

chmeliik:
```shell
$ venv/bin/mypy --ignore-missing-imports --check-untyped-defs --package atomic_reactor
...
Found 100 errors in 28 files (checked 83 source files)
```
That's more like it :tada:

MartinBasti: https://media.giphy.com/media/7p3e2WCM0VEnm/giphy.gif

tkdchen: > Even if we don't figure it out, do we want to do something about it? Fix the type annotation or fix the code to use a list?

Yes. It would be good to be in a separate dedicated pr I think. :)

rcerven (approved):

atomic_reactor/plugins/build_source_container.py
ben-alkov: > Yeah, so about that "mypy is checking the entire codebase now"...

> `mypy --ignore-missing-imports --check-untyped-defs`

Either that or https://github.com/ben-alkov/actions/pull/11

PR #1813, opened 2022-02-17
Separate source container exit task

tkdchen:
* COULDBLD-8272

Split the exit plugin definition into a separate exit task, and always
execute this exit task as the last step in the pipeline.

### DIFF
--- a/atomic_reactor/cli/parser.py
+++ b/atomic_reactor/cli/parser.py
@@ -50,12 +50,19 @@ def parse_args(args: Optional[Sequence[str]] = None) -> dict:
     )
     worker.set_defaults(func=task.worker)
 
-    source_build = tasks.add_parser(
-        "source-build",
+    source_container_build = tasks.add_parser(
+        "source-container-build",
         help="build a source container",
         description="Build a source container.",
     )
-    source_build.set_defaults(func=task.source_build)
+    source_container_build.set_defaults(func=task.source_container_build)
+
+    source_container_exit = tasks.add_parser(
+        "source-container-exit",
+        help="exit a source container build",
+        description="Execute source container exit steps.",
+    )
+    source_container_exit.set_defaults(func=task.source_container_exit)
 
     clone = tasks.add_parser(
         "clone",
--- a/atomic_reactor/cli/task.py
+++ b/atomic_reactor/cli/task.py
@@ -10,7 +10,9 @@ from atomic_reactor.tasks.binary import BinaryBuildTask, BinaryExitTask, \
 from atomic_reactor.tasks.clone import CloneTask
 from atomic_reactor.tasks.common import TaskParams
 from atomic_reactor.tasks.orchestrator import OrchestratorTask, OrchestratorTaskParams
-from atomic_reactor.tasks.sources import SourceBuildTask, SourceBuildTaskParams
+from atomic_reactor.tasks.sources import (
+    SourceExitTask, SourceBuildTask, SourceBuildTaskParams
+)
 from atomic_reactor.tasks.worker import WorkerTask, WorkerTaskParams
 
 
@@ -34,13 +36,23 @@ def worker(task_args: dict):
     return task.execute()
 
 
-def source_build(task_args: dict):
+def source_container_build(task_args: dict):
     """Run a source container build.
 
-    :param task_args: CLI arguments for a source-build task
+    :param task_args: CLI arguments for a source-container-build task
     """
     params = SourceBuildTaskParams.from_cli_args(task_args)
     task = SourceBuildTask(params)
+    return task.execute()
+
+
+def source_container_exit(task_args: dict):
+    """Run source container exit steps.
+
+    :param task_args: CLI arguments for a source-container-exit task
+    """
+    params = SourceBuildTaskParams.from_cli_args(task_args)
+    task = SourceExitTask(params)
     return task.execute()
 
 
--- a/atomic_reactor/tasks/sources.py
+++ b/atomic_reactor/tasks/sources.py
@@ -28,8 +28,22 @@ class SourceBuildTaskParams(common.TaskParams):
         return source.DummySource(None, None, workdir=self.build_dir)
 
 
-class SourceBuildTask(plugin_based.PluginBasedTask):
-    """Source container build task."""
+class SourceBuildBaseTask(plugin_based.PluginBasedTask):
+    """Base task for defining different build phases for source container build."""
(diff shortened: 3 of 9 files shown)


atomic_reactor/cli/parser.py
```diff
@@ -50,12 +50,19 @@ def parse_args(args: Optional[Sequence[str]] = None) -> dict:
...
         help="build a source container",
         description="Build a source container.",
     )
-    source_build.set_defaults(func=task.source_build)
+    source_container_build.set_defaults(func=task.source_container_build)
+
+    source_container_exit = tasks.add_parser(
+        "source-container-exit",
+        help="Source container build exit task.",
+        description="Run plugins while exiting a source container build.",
```
chmeliik, line 63: Could this be consistent with the binary container exit task?
```suggestion
        help="exit a source container build",
        description="Execute source container exit steps.",
```

atomic_reactor/cli/task.py
```diff
@@ -10,7 +10,9 @@
 from atomic_reactor.tasks.clone import CloneTask
 from atomic_reactor.tasks.common import TaskParams
 from atomic_reactor.tasks.orchestrator import OrchestratorTask, OrchestratorTaskParams
-from atomic_reactor.tasks.sources import SourceBuildTask, SourceBuildTaskParams
+from atomic_reactor.tasks.sources import (
+    SourceBuildExitTask, SourceBuildPhasesTask, SourceBuildTaskParams
```
chmeliik, line 14: Again for consistency with binary tasks, maybe the classes should be `SourceExitTask`, `SourceBuildTask`, `SourceTaskParams`?

atomic_reactor/cli/task.py
```diff
@@ -34,13 +36,23 @@ def worker(task_args: dict):
...
+    :param task_args: CLI arguments for a source-container-build task
     """
     params = SourceBuildTaskParams.from_cli_args(task_args)
-    task = SourceBuildTask(params)
+    task = SourceBuildPhasesTask(params)
+    return task.execute()
+
+
+def source_container_exit(task_args: dict):
+    """Run a source container build.
```
chmeliik, line 50:
```suggestion
    """Run source container exit steps.
```

atomic_reactor/tasks/sources.py
```diff
@@ -44,15 +53,17 @@ class SourceBuildTask(plugin_based.PluginBasedTask):
...
+    """Source container build task."""
+
+    plugins_def = plugin_based.PluginsDef(
         exit=[
             {"name": "verify_media", "required": False},
             {"name": "koji_import_source_container"},
             {"name": "koji_tag_build"},
             {"name": "store_metadata"},
         ],
     )
```
chmeliik, line 69: This won't work, because the `workflow.build_docker_image()` method cannot handle when there are 0 buildstep plugins.

@mkosiarc, you encountered this while working on the build task, do you have a solution yet? If yes, could it be a standalone PR?

mkosiarc: I don't have it yet, but will work on it soon. If I finish this and test that it works, I can create a standalone PR.

chmeliik: Ack, just to make sure @tkdchen and you don't do the same work

chmeliik: It won't work even if you move the plugins from `exit` to `prebuild` :disappointed: 

At least the plugins will get executed this way, but after that the finally task will fail with "No appropriate build step"

tekton/tasks/source-container-exit.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+  workspaces:
+    - name: ws-build-dir
+    - name: ws-context-dir
+    - name: ws-registries-secret  # access with $(workspaces.ws-registries-secret.path)/token
+    - name: ws-koji-secret  # access with $(workspaces.ws-koji-secret.path)/token
+    - name: ws-reactor-config-map
+
+  steps:
+    - name: source-container-exit-image
```
chmeliik, line 24:
```suggestion
    - name: source-container-exit
```

tekton/tasks/source-container-exit.yaml
```diff
@@ -0,0 +1,39 @@
...
+      image: $(params.OSBS_IMAGE)
+      resources:
+        requests:
+          memory: 3000Mi
+          cpu: 500m
+        limits:
+          memory: 8600Mi
+          cpu: 1300m
```
chmeliik, line 32: This could probably use the same resources as the binary container exit task. Only the source-container-build task should need the extra resources

tkdchen:
This pr has to be merged after https://github.com/containerbuildsystem/atomic-reactor/pull/1821.

atomic_reactor/tasks/sources.py
```diff
@@ -44,15 +53,17 @@ class SourceBuildTask(plugin_based.PluginBasedTask):
...
         ],
-        exit=[
+    )
+
+
+class SourceExitTask(SourceBuildBaseTask):
+    """Source container exit task."""
+
+    plugins_def = plugin_based.PluginsDef(
+        prebuild=[
```
chmeliik, line 63: Could this be `exit` now?

I think there is some logic in `plugin.py` which makes it so that the `PrebuildPluginsRunner` will refuse to run exit plugins.

tkdchen: Done. It's now `exit`.

chmeliik (approved):
LGTM :+1:

tekton/pipelines/source-container.yaml
```diff
@@ -38,3 +38,23 @@ spec:
           value: "$(params.OSBS_IMAGE)"
         - name: USER_PARAMS
           value: "$(params.USER_PARAMS)"
+  finally:
+    - name: source-container-exit
```
hjmodi, line 42: the name of the build task above(https://github.com/containerbuildsystem/atomic-reactor/pull/1813/files#diff-37fd879ef818700016f8f18b08a45989f6a7eaa79c3090b68180c1d541a1402cR22) should also be consistent with this naming

hjmodi:
LGTM. Just one minor comment

tekton/pipelines/source-container.yaml
tkdchen: Update, PTAL.

hjmodi (approved):

tkdchen:
Rebased.

PR #1821, opened 2022-02-23
Don't raise error when buildstep provides no results

mkosiarc:
The prebuild task currently fails because in
build_docker_image method we run all plugin runners
and we expect results from buildstep plugin.

This is a workaround and temporary fix. This whole process
of running plugins will be reworked later.

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -837,13 +837,13 @@ class DockerBuildWorkflow(object):
             try:
                 self.data.build_result = buildstep_runner.run()
 
-                if self.data.build_result.is_failed():
+                if self.data.build_result and self.data.build_result.is_failed():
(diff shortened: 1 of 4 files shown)


atomic_reactor/inner.py
```diff
@@ -857,21 +855,6 @@ def build_docker_image(self) -> BuildResult:
                 logger.error("one or more prepublish plugins failed: %s", ex)
                 raise
 
-            if self.data.build_result.is_image_available():
```
mkosiarc, line 860: I removed this to avoid any changes here, as this will be removed anyway in https://github.com/containerbuildsystem/atomic-reactor/pull/1810/. (I will rebase this after that is merged)

rcerven (approved):

tkdchen (approved):

PR #1827, opened 2022-02-28
Update compress plugin

hjmodi:

### DIFF
--- a/atomic_reactor/plugins/post_fetch_docker_archive.py
+++ b/atomic_reactor/plugins/post_fetch_docker_archive.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2015 Red Hat, Inc
+Copyright (c) 2015-2022 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
@@ -16,22 +16,22 @@ from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.util import get_exported_image_metadata, human_size, is_scratch_build
 
 
-class CompressPlugin(PostBuildPlugin):
+class FetchDockerArchivePlugin(PostBuildPlugin):
     """Example configuration:
 
     "postbuild_plugins": [{
-            "name": "compress",
+            "name": "fetch_docker_archive",
             "args": {
                     "method": "gzip",
                     "load_exported_image": true
             }
     }]
 
-    Currently supported compression methods are gzip and lzma; gzip is default.
+    Currently, supported compression methods are gzip and lzma; gzip is default.
     By default, the plugin doesn't work on exported image, you have to explicitly
     ask for it by using `load_exported_image: true`.
     """
-    key = 'compress'
+    key = 'fetch_docker_archive'
     is_allowed_to_fail = False
 
     def __init__(self, workflow, load_exported_image=False, method='gzip'):
@@ -40,7 +40,7 @@ class CompressPlugin(PostBuildPlugin):
         :param load_exported_image: bool, when running squash plugin with `dont_load=True`,
                                     you may load the exported tar with this switch
         """
-        super(CompressPlugin, self).__init__(workflow)
+        super(FetchDockerArchivePlugin, self).__init__(workflow)
         self.load_exported_image = load_exported_image
         self.method = method
         self.uncompressed_size = 0
--- a/atomic_reactor/plugins/post_fetch_docker_archive.py
+++ b/atomic_reactor/plugins/post_fetch_docker_archive.py
@@ -5,11 +5,8 @@ All rights reserved.
 This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
-
-import os
-
-from atomic_reactor.constants import (EXPORTED_COMPRESSED_IMAGE_NAME_TEMPLATE,
-                                      IMAGE_TYPE_DOCKER_ARCHIVE)
+from atomic_reactor.constants import IMAGE_TYPE_DOCKER_ARCHIVE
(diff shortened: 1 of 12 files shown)


atomic_reactor/plugins/post_fetch_docker_archive.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __init__(self, workflow):
+        """
+        :param workflow: DockerBuildWorkflow instance
+        """
+        super(FetchDockerArchivePlugin, self).__init__(workflow)
+        self.source_build = bool(self.workflow.data.build_result.source_docker_archive)
+
+    def download_image(self, build_dir: BuildDir):
+        image = self.workflow.data.tag_conf.get_unique_images_with_platform(build_dir.platform)[0]
```
hjmodi, line 28: here I am assuming the list is of length one and contains the images we're want to download. Please correct me if wrong.

chmeliik: That sounds correct :+1: 

Even if there somehow were multiple unique tags, we would expect all of them to point to the same image, so taking any one of them should be fine.

atomic_reactor/plugins/post_fetch_docker_archive.py
```diff
@@ -6,106 +6,45 @@
...
+        # Set image_tarball_path, if image tarball is not yet
+        # stored somewhere, download it
+        # with imageutil.download_image_archive_tarball() for each platform
+        outfile = os.path.join(self.workflow.source.workdir,
+                               EXPORTED_COMPRESSED_IMAGE_NAME_TEMPLATE)
+        os.mknod(outfile)
         metadata = get_exported_image_metadata(outfile, image_type)
 
-        if self.uncompressed_size != 0:
-            metadata['uncompressed_size'] = self.uncompressed_size
```
chmeliik: Not sure about removing `uncompressed_size`, doesn't some metadata plugin later use it?

atomic_reactor/plugins/post_fetch_docker_archive.py
```diff
@@ -25,6 +24,20 @@ def __init__(self, workflow):
...
+        image = self.workflow.data.tag_conf.get_unique_images_with_platform(build_dir.platform)[0]
+        image_path = os.path.join(build_dir.path, EXPORTED_SQUASHED_IMAGE_NAME)
+        image_type = IMAGE_TYPE_DOCKER_ARCHIVE
+
+        self.log.info('fetching image %s', image)
+        self.workflow.imageutil.download_image_archive_tarball(image, image_path)
+
+        metadata = get_exported_image_metadata(image_path, image_type)
+
+        self.workflow.data.exported_image_sequence.append(metadata)
```
chmeliik, line 37: It doesn't seem right to put the 4 arch-specific images into the `exported_image_sequence` - plugins assume that everything in this sequence is the same image (don't ask me why it is a sequence then :smile:).

IMO this plugin should not update `exported_image_sequence` anymore. All the plugins that currently depend on the sequence -- appears to be `tag_and_push`, `store_metadata` and `koji_import` (through `utils.koji.get_output`) -- will have to be reworked for multiple platforms and, as part of that, they will have to stop using this attr.

It might be good to add `OSBS2 TBD` into the places where the sequence is currently used, that it should not be used anymore

atomic_reactor/plugins/post_fetch_docker_archive.py
chmeliik: Nevermind, nothing uses it, not even in the osbs_ocp3 branch :raised_eyebrow:

atomic_reactor/plugins/post_fetch_docker_archive.py
chmeliik: Maybe, instead of appending to the sequence, we could just `return` the metadata. That means `for_each_platform()` would return a dict of platform=>metadata, which we could return from the plugin

hjmodi: added `OSBS2 TBD` and updated the plugin to return platform=>metadata

atomic_reactor/plugins/post_fetch_docker_archive.py
```diff
@@ -25,6 +24,20 @@ def __init__(self, workflow):
         super(FetchDockerArchivePlugin, self).__init__(workflow)
         self.source_build = bool(self.workflow.data.build_result.source_docker_archive)
 
+    def download_image(self, build_dir: BuildDir):
+        image = self.workflow.data.tag_conf.get_unique_images_with_platform(build_dir.platform)[0]
+        image_path = os.path.join(build_dir.path, EXPORTED_SQUASHED_IMAGE_NAME)
```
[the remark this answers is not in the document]
hjmodi: updated

atomic_reactor/plugins/post_fetch_docker_archive.py
```diff
@@ -6,106 +6,45 @@
...
-            # OSBS2 TBD
-            # Set image_tarball_path, if image tarball is not yet
-            # stored somewhere, download it
-            # with imageutil.download_image_archive_tarball()
-            image_tarball_path = None
-            with open(image_tarball_path, 'rb') as image_tarball_file:
-                outfile = self._compress_image_stream(image_tarball_file)
+        if self.source_build:
+            self.log.info('skipping, no exported source image')
+            return
```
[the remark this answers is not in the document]
hjmodi: updated to not run for source container workflow.

hjmodi:
Just made a minor update removed a parameter that wasn't used

lkolacek (approved):
LGTM :+1:

PR #1835, opened 2022-03-03
Remove build_result attribute from workflow data

chmeliik:
Saving the build_result in workflow data causes various issues, like
accidentally overwriting it with None in tasks that don't run any
buildstep plugins. Plus it causes duplication, the build result is
already getting saved in the buildstep_result attribute (as a plugin
result)

Make build_result a property derived from buildstep_result. Add a
setter to avoid having to fix unit tests that set buildstep_result
directly.

Remove build_result from the workflow.json schema, but note that the
custom JSON-encoder|decoder code is still needed (for the BuildResult
object in buildstep_result).

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -464,11 +464,18 @@ class ImageBuildWorkflowData(ISerializer):
     image_id: Optional[str] = None
     parent_images_digests: Dict[str, Dict[str, str]] = field(default_factory=dict)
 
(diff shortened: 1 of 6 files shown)


chmeliik:
PTAL, this should make #1813 work (and we would have hit the same problem with the binary pipeline)

rcerven (approved):

rcerven (approved):

tkdchen (approved):

PR #1839, opened 2022-03-07
remote_host: support configurable slots dir

qixiang:
Default slots dir is $HOME/$SLOTS_RELATIVE_PATH, this can be changed
while instantiating a RemoteHost:

    host = RemoteHost(
        hostname="remote-host-001",
        username="builder",
        ssh_keyfile="/path/to/key",
        slots=3,
        slots_dir="/path/to/slots/dir"
    )

Signed-off-by: Qixiang Wan <qwan@redhat.com>

### DIFF
--- a/atomic_reactor/utils/remote_host.py
+++ b/atomic_reactor/utils/remote_host.py
@@ -143,18 +143,28 @@ class SlotData:
 
 class RemoteHost:
 
-    def __init__(self, *, hostname: str, username: str, ssh_keyfile: str, slots: int):
+    def __init__(
+        self,
+        *,
+        hostname: str,
+        username: str,
+        ssh_keyfile: str,
+        slots: int,
+        slots_dir: Optional[str] = None,
+    ):
         """ Instantiate RemoteHost with hostname, username, ssh key file and slot number
 
         :param hostname: str, remote hostname for ssh connection
         :param username: str, username for ssh connection
         :param ssh_keyfile: str, filepath to ssh private key
         :param slots: int, number of max allowed slots on remote host
+        :param slots_dir: str, directory path for holding slots files
         """
         self._hostname = hostname
         self._username = username
         self._ssh_keyfile = ssh_keyfile
         self._slots = slots
+        self._slots_dir = slots_dir
 
     @property
     def hostname(self):
@@ -174,8 +184,12 @@ class RemoteHost:
 
     @cached_property
     def slots_dir(self):
-        home, _, _ = self._run("pwd")
-        return os.path.join(home, SLOTS_RELATIVE_PATH)
+        # Place the slots files under `~/$SLOTS_RELATIVE_PATH` when slots_dir is not specified
+        if self._slots_dir is None:
+            home, _, _ = self._run("pwd")
+            return os.path.join(home, SLOTS_RELATIVE_PATH)
+        else:
+            return self._slots_dir
 
     def _is_valid_slot_id(self, slot_id: int) -> bool:
         """ Check if a slot id is valid """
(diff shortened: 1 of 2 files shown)


atomic_reactor/utils/remote_host.py
```diff
@@ -143,18 +143,28 @@ def datetime(self):
...
 
-    def __init__(self, *, hostname: str, username: str, ssh_keyfile: str, slots: int):
+    def __init__(
+        self,
+        *,
+        hostname: str,
+        username: str,
+        ssh_keyfile: str,
+        slots: int,
+        slots_dir: Optional[str] = None,
```
chmeliik, line 153: Shouldn't the `RemoteHostsPool.from_config` method pass the `slots_dir` from config now?

qixiang: ah, yes. The config passed to `RemoteHostsPool.from_config` currently is a list of hosts and their attributions (https://github.com/containerbuildsystem/atomic-reactor/blob/d35d81a8fd4df2ed67673de295292d80c9a287dd/atomic_reactor/utils/remote_host.py#L577-L584), any suggestion how we can pass the `slots_dir` which should be configured per cluster or per arch? @MartinBasti

chmeliik: If we do want to keep `slots_dir` above the individual host level, then I would vote to change the from_config method from `from_config(per_platform_hosts)` to `from_config(all_hosts, platform)` (i.e. not passing just `.remote_hosts.x86_64` from the configuration but the entire `.remote_hosts` field)

rcerven: why we need slots_dir configurable per host or arch?

I'd just have 1 slots_dir for all of them, there isn't any reason to have different ones

qixiang: > If we do want to keep `slots_dir` above the individual host level, then I would vote to change the from_config method from `from_config(per_platform_hosts)` to `from_config(all_hosts, platform)` (i.e. not passing just `.remote_hosts.x86_64` from the configuration but the entire `.remote_hosts` field)

`from_config(all_hosts, ...)` will introduce unnecessary work for reserving a host for a task, because one task is generated for each arch, for example, when we want to get a host for a x86_64 task, we will need to filter out unmatched arch hosts after `from_config(all_hosts, ...)`.

> why we need slots_dir configurable per host or arch?
> I'd just have 1 slots_dir for all of them, there isn't any reason to have different ones

I agree, the story behind this is not we want to configure it per host or arch. In the original code, the `$HOME/osbs_slots` dir is used to place the slots files, however we found the service account doesn't have home directory created on remote hosts by default, so we want to make it configurable. The `RemoteHost` is the only place or the only reasonable place we can make the change, but storing "slots_dir" for every hosts or even arch in cluster configmap doesn't make sense.

Another proposal is using a different default dir other than `$HOME` to place the slot files under `osbs_slots`, for example, the systemd user runtime dir `/run/user/{uid}`(`$XDG_RUNTIME_DIR`) and we `enable-linger` for the ssh user in ansible role, so the user runtime dir will always be created after system boot up.

wdyt?

chmeliik: > `from_config(all_hosts, ...)` will introduce unnecessary work for reserving a host for a task, because one task is generated for each arch, for example, when we want to get a host for a x86_64 task, we will need to filter out unmatched arch hosts after `from_config(all_hosts, ...)`.

I meant like this:
```diff
diff --git a/atomic_reactor/utils/remote_host.py b/atomic_reactor/utils/remote_host.py
index 6bd7abe1..1eddb455 100644
--- a/atomic_reactor/utils/remote_host.py
+++ b/atomic_reactor/utils/remote_host.py
@@ -567,10 +567,10 @@ class RemoteHostsPool:
         self.hosts = hosts
 
     @classmethod
-    def from_config(cls, config: dict):
+    def from_config(cls, config: dict, platform: str):
         """ Instantiate remote hosts loaded from a config in dict format
 
-        :param config: dict, Arch specific remote hosts dictionary from configmap
+        :param config: dict, remote hosts dictionary from configmap
 
         Example:
 
@@ -583,8 +583,10 @@ class RemoteHostsPool:
         hostname-remote-host2:
           ...
         """
+        slots_dir = config.get("slots_dir", some_default)
+        platform_config = config[platform]
         hosts = []
-        for hostname, attr in config.items():
+        for hostname, attr in platform_config.items():
             if not attr.get("enabled", False):
                 continue
             host = RemoteHost(
```
> Another proposal is using a different default dir other than $HOME to place the slot files under osbs_slots, for example, the systemd user runtime dir /run/user/{uid}($XDG_RUNTIME_DIR) and we enable-linger for the ssh user in ansible role, so the user runtime dir will always be created after system boot up.

I'd rather have it be a system-wide dir, since the slots are supposed to be per-host, not per-user (it's true that the hosts only have one user who should be running builds, but it still feels cleaner to use a system dir)

MartinBasti: Works for me, I don't like mixing platfroms with atributes, IMO we should have something like

```
remote_hosts:
    slots_dir: ...
    pools:
        x86_64: ...
        ppc64le: ...        
```

qixiang: > I meant like this:

I see, thanks, PR is updated to adopt this approach.

Beside of this, if we're going to use a system-wide dir and not some tmpfs, should we create the slots directory in the remote hosts provision role? because we'll need proper permissions to allow the ssh user to create files under it.

lkolacek: I believe we wanted to make OSBS more robust to the accidental removal of slots_dir and files under it with its creation here. So we do not need to run manually playbooks to enable our workflow again but it will get fixed itself.

qixiang: > I believe we wanted to make OSBS more robust to the accidental removal of slots_dir and files under it with its creation here. So we do not need to run manually playbooks to enable our workflow again but it will get fixed itself.

we have the creation in `is_operational`check:

https://github.com/containerbuildsystem/atomic-reactor/blob/0ae2bfe6c4f6c3cab2dcdb6dd3ba2834dd3bd241/atomic_reactor/utils/remote_host.py#L314-L317

though if we're using some system-wide directory, it may not work in some cases due to lack of permissions. If we use some dirs owned by the ssh user, it will work. My proposal is using the systemd user runtime dir, I'll open a PR in remote hosts ansible role to enable user lingering for the ssh user, so the `$XDG_RUNTIME_DIR` will always be created after system boot up.

chmeliik (approved):
LGTM :+1:

MartinBasti (approved):

rcerven (approved):

PR #1840, opened 2022-03-07
Update generate maven metadata plugin

hjmodi:


tkdchen (commented):
This change avoids download in `generate_maven_metadata`, however the result looks move to another side, that is the `fetch_maven_artifacts` takes part of the responsibility of the metadata generation.

FYI. IIUC, these two plugins share most of the code logic to iterate artifacts and take different actions based on the same data source in pre- and post- build phases. In prebuild, it fetches artifacts, and in the postbuild, it generates metadata. To achieve the goal of CLOUDBLD-8264, probably it's a good time to refactor these two plugins like what was done for the koji_import plugin.

BTW, can we rename `remote_source_files` to another one that can reflect the meaning of maven artifact? e.g. `maven_artifacts_output` or something like that. "output" should make sense because it is a term used in the Koji Content Generator Metadata.

hjmodi:
> This change avoids download in `generate_maven_metadata`, however the result looks move to another side, that is the `fetch_maven_artifacts` takes part of the responsibility of the metadata generation.
> 
> FYI. IIUC, these two plugins share most of the code logic to iterate artifacts and take different actions based on the same data source in pre- and post- build phases. In prebuild, it fetches artifacts, and in the postbuild, it generates metadata. To achieve the goal of CLOUDBLD-8264, probably it's a good time to refactor these two plugins like what was done for the koji_import plugin.
> 
> BTW, can we rename `remote_source_files` to another one that can reflect the meaning of maven artifact? e.g. `maven_artifacts_output` or something like that. "output" should make sense because it is a term used in the Koji Content Generator Metadata.

Hmm, there seems to be some confusion here. I'll add comments in code to make it less confusing and probably rename `generate_maven_metadata` plugin to `generate_remote_source_files`.

This PR accomplishes the following,
`fetch_maven_artifacts` plugin is now processing all `fetch_*` configurations and it's only done once. While iterating it's downloading the maven artifacts and also generating metadata for those artifacts as much as possible(for nvr and pnc artifacts, this was done in `generate_maven_metadata` before this). It is also creating and returning `source_download_queue` that is a queue of sources for url artifacts that is uploaded to binary brew build as `remote_source_files` to be later fetched by source container build.

`generate_maven_metadata` plugin is no longer generating metadata for all `fetch_*` configs. It is only downloading from `source_download_queue` that `fetch_maven_artifacts` generate and creates `remote_source_files` from them. We've decided to do it in `postbuild` so that we won't have source artifacts in `build_dir` during the binary build. That's why it's not done in pre.

hjmodi:
Coverage seems to be decreasing due to decrease in the number of lines.

rcerven (approved):

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
```diff
@@ -185,16 +235,22 @@ def run(self):
...
-        pnc_artifact_ids = self.get_pnc_artifact_ids(pnc_requests)
-
         return {
+            'components': components,
             'download_queue': [dataclasses.asdict(download) for download in download_queue],
+            'no_source': self.no_source_artifacts,
             'pnc_artifact_ids': pnc_artifact_ids,
+            'pnc_build_metadata': pnc_build_metadata,
+            'source_download_queue': source_download_queue,
+            'source_url_to_artifacts': self.source_url_to_artifacts,
```
chmeliik, line 255: Note that plugin results have to be convertible to JSON now (for context sharing). I'm not sure what the types of these values are, but I think at least `source_download_queue` is not JSON compatible

atomic_reactor/constants.py
```diff
@@ -86,7 +86,7 @@
 PLUGIN_BUMP_RELEASE_KEY = 'bump_release'
 PLUGIN_DISTGIT_FETCH_KEY = 'distgit_fetch_artefacts'
 PLUGIN_FETCH_MAVEN_KEY = 'fetch_maven_artifacts'
-PLUGIN_GENERATE_MAVEN_METADATA_KEY = 'generate_maven_metadata'
+PLUGIN_GENERATE_REMOTE_SOURCE_FILES_KEY = 'generate_remote_source_files'
```
chmeliik, line 89: I'm not entirely sold on this naming, it sounds like it would be related to `resolve_remote_source` but it is not. Is there any way this could be made clearer?

By the way, what exactly *are* remote_source_files? :smile:

atomic_reactor/plugins/post_generate_maven_metadata.py
```diff
@@ -37,98 +32,15 @@ def __init__(self, workflow):
...
-                #  in the URL itself so we'll have to get filename from URL response
-                target = os.path.basename(source_url)
-
-                download_queue.append(DownloadRequest(source_url, target, checksums))
-            else:
-                self.source_url_to_artifacts[source_url].append(artifact)
-        return download_queue
-
-    def download_sources(self, download_queue):
+    def get_remote_source_files(self, download_queue):
```
chmeliik: It would be good to add type annotations here, this code currently expects that `download_queue` is a list of `DownloadRequest` (but that may have to change due to the requirement of plugin results being JSON-compatible)

Plus we have adding type annotations in the checklist now :)

atomic_reactor/constants.py
hjmodi: added type annotations

atomic_reactor/plugins/post_generate_maven_metadata.py
hjmodi: added type annotations

atomic_reactor/constants.py
hjmodi: `remote_source_file` is a koji build type we added to attach maven URL source artifacts to brew binary build task that are later fetched by source build and added to source container.

atomic_reactor/plugins/pre_fetch_maven_artifacts.py
hjmodi: added serialization and deserialization

atomic_reactor/plugins/post_generate_maven_metadata.py
```diff
@@ -34,7 +33,9 @@ def __init__(self, workflow):
         super(GenerateMavenMetadataPlugin, self).__init__(workflow)
         self.source_url_to_artifacts = {}
 
-    def get_remote_source_files(self, download_queue):
+    def get_remote_source_files(
+            self, download_queue: Sequence[DownloadRequest]
+    ) -> List[Dict[str, Collection[str]]]:
```
chmeliik: I'd go with `List[dict]` or `List[Dict[str, Any]]`, the fact that all the values are collection of strings is more misleading than helpful (one value is a string, the other is a dict[str, something], both of which happen to be collection of strings)

atomic_reactor/constants.py
```diff
@@ -86,7 +86,7 @@
 PLUGIN_BUMP_RELEASE_KEY = 'bump_release'
 PLUGIN_DISTGIT_FETCH_KEY = 'distgit_fetch_artefacts'
 PLUGIN_FETCH_MAVEN_KEY = 'fetch_maven_artifacts'
-PLUGIN_GENERATE_MAVEN_METADATA_KEY = 'generate_maven_metadata'
+PLUGIN_URL_SOURCES_METADATA_KEY = 'url_sources_metadata'
```
chmeliik, line 89: Could we have "maven" in the name? Sorry to request a second rename, but I probably wouldn't notice the two plugins are related otherwise

chmeliik (commented):
Looks good, just 2 minor things :+1:

atomic_reactor/plugins/post_generate_maven_metadata.py
hjmodi: updated type annotations to `List[Dict[str, Any]]`

atomic_reactor/constants.py
hjmodi: renamed it to `maven_url_sources_metadata`

tkdchen (approved):

chmeliik (approved):
LGTM :+1:

PR #1843, opened 2022-03-09
Update group_manifests for Tekton

chmeliik:
CLOUDBLD-8263

Stop taking the manifest digests of built images from worker build
annotations. Instead, get the pullspecs of built images from tag_conf
and query the registry to find their manifest digests.

And other cleanups, see individual commits.

# Maintainers will complete the following section

- [x] Commit messages are descriptive enough
- [x] Code coverage from testing does not decrease and new code is covered
- [x] Python type annotations added to new code
- [n/a] JSON/YAML configuration changes are updated in the relevant schema
- [n/a] Changes to metadata also update the documentation for the metadata
- [n/a] Pull request has a link to an osbs-docs PR for user documentation updates
- [n/a] New feature can be disabled from a configuration file

PR #1847, opened 2022-03-11
Remove BuildResult - CLOUDBLD-8975

tkdchen:
Detailed description is in the individual commit message.

### DIFF
--- a/atomic_reactor/plugins/build_source_container.py
+++ b/atomic_reactor/plugins/build_source_container.py
@@ -10,13 +10,12 @@ import shutil
 import subprocess
 from pathlib import Path
 
-from atomic_reactor.inner import BuildResult
 from atomic_reactor.constants import (
     IMAGE_TYPE_DOCKER_ARCHIVE,
     PLUGIN_FETCH_SOURCES_KEY,
     PLUGIN_SOURCE_CONTAINER_KEY,
 )
-from atomic_reactor.plugin import BuildStepPlugin
+from atomic_reactor.plugin import BuildStepPlugin, PluginFailedException
 from atomic_reactor.util import get_exported_image_metadata
 from atomic_reactor.utils import retries
 
@@ -27,9 +26,10 @@ class SourceContainerPlugin(BuildStepPlugin):
     https://github.com/containers/BuildSourceImage
     """
 
+    is_allowed_to_fail = False
     key = PLUGIN_SOURCE_CONTAINER_KEY
 
-    def export_image(self, image_output_dir: Path) -> str:
+    def export_image(self, image_output_dir: Path):
         output_path = self.workflow.build_dir.any_platform.exported_squashed_image
 
         cmd = ['skopeo', 'copy']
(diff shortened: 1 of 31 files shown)


tkdchen:
Note: This pr is related to #1843. If it gets merged firstly, rebase this pr. For the group_manifests.

tkdchen:
Rebased.

atomic_reactor/plugins/build_source_container.py
```diff
@@ -150,13 +150,7 @@ def run(self):
                           maven_source_data_dir)
             shutil.rmtree(maven_source_data_dir)
 
-        image_tar_path = self.export_image(image_output_dir)
+        self.export_image(image_output_dir)
 
         self.log.info('Will remove unpacked image directory: %s', image_output_dir)
         shutil.rmtree(image_output_dir)
-
```
rcerven, line 157: well we shouldn't return BuildResult anymore BUT, we should return the same as dict .....

atomic_reactor/plugins/build_source_container.py
```diff
@@ -46,7 +46,6 @@ def export_image(self, image_output_dir: Path) -> str:
         img_metadata = get_exported_image_metadata(str(output_path), IMAGE_TYPE_DOCKER_ARCHIVE)
         # OSBS2 TBD exported_image_sequence will not work for multiple platform
         self.workflow.data.exported_image_sequence.append(img_metadata)
-        return str(output_path)
```
rcerven, line 49: we still want that

rcerven, line 49: we still want that

atomic_reactor/plugins/build_source_container.py
```diff
@@ -150,13 +150,7 @@ def run(self):
                           maven_source_data_dir)
             shutil.rmtree(maven_source_data_dir)
 
-        image_tar_path = self.export_image(image_output_dir)
+        self.export_image(image_output_dir)
 
         self.log.info('Will remove unpacked image directory: %s', image_output_dir)
         shutil.rmtree(image_output_dir)
-
-        return BuildResult(
```
rcerven, line 158: well we shouldn't return BuildResult anymore BUT, we should return the same as dict .....

rcerven (changes requested):

atomic_reactor/plugins/build_source_container.py
tkdchen: @rcerven Could you be specific why we want this? The source docker archive can be accessible from the `BuildDir.exported_squashed_image`.

atomic_reactor/plugins/build_source_container.py
chmeliik: Hmm, and why do we want that?

chmeliik:
Mostly LGTM

Should we also remove build result handling from the ContextDir?

atomic_reactor/plugins/build_source_container.py
rcerven: ahh right so we know that output dir already in advance, right so logs and skip_layer as return looks good

atomic_reactor/plugins/build_source_container.py
rcerven: right that will do

rcerven (approved):

tkdchen:
@chmeliik Updated. PTAL.

chmeliik (approved):
LGTM :+1:

PR #1852, opened 2022-03-17
Update metadata plugins

tkdchen:
* CLOUDBLD-8265

For details, please refer to each commit message.


atomic_reactor/utils/koji.py
```diff
@@ -322,7 +324,7 @@ def get_buildroot():
         },
         'container': {
             'type': 'none',
-            'arch': os.uname()[4],
+            'arch': arch or os.uname().machine,
```
rcerven, line 328: I think we should just use 'host_arch'

os.uname vs platform.processor give the same results for ppc,x86,s390x, and differ only for arm

arm64 vs aarch64

so I'd choose just one of os.uname or platform.processor as default

atomic_reactor/inner.py
```diff
@@ -345,6 +345,14 @@ class ImageBuildWorkflowData(ISerializer):
     image_id: Optional[str] = None
     parent_images_digests: Dict[str, Dict[str, str]] = field(default_factory=dict)
 
+    # List of output files that are uploaded to Brew/Koji
+    # Each element is a two-strings list, local_filename and dest_filename. E.g.
+    # [
+    #     ["/path/to/data.json", "metadata.json"],
+    #     ["/path/to/build.log", "x86_64-build.log"],
+    # ]
+    cg_output_files: List[List[str]] = field(default_factory=list)
```
rcerven, line 354: can we name it differently ?

brew_upload_files    or something more clear

atomic_reactor/plugins/post_koji_import.py
```diff
@@ -440,127 +440,132 @@ def get_build(self):
...
+        remote_source_file_outputs, kojifile_components = get_maven_metadata(self.workflow.data)
 
         # add maven components alongside RPM components
         for worker_output in output:
             if worker_output['type'] == 'docker-image':
                 worker_output['components'] += kojifile_components
 
         # add remote sources tarballs and remote sources json files to output
-        for remote_source_output in [
+        for item in [
```
rcerven, line 482: call it more meaningful

atomic_reactor/inner.py
rcerven: also would be nice to have list of dicts or named tuples

atomic_reactor/plugins/post_koji_import.py
```diff
@@ -192,7 +192,7 @@ def set_media_types(self, extra):
 
         # Append media_types from verify images
         media_results = self.workflow.data.postbuild_results.get(PLUGIN_VERIFY_MEDIA_KEY)
-        if media_results:
+        if media_results and not isinstance(media_results, Exception):
```
rcerven, line 195: verify media plugin will run, and it should be required now: https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/tasks/binary.py#L59

because now media_types depend on it, so if verify media plugin fails, build should fail as well, as we shouldn't have empty media_types in brew

atomic_reactor/utils/koji.py
lkolacek: +1 for choosing one as default, I believe there is also difference between x86_64 vs amd64 representation

atomic_reactor/plugins/post_koji_import.py
```diff
@@ -133,16 +133,20 @@ def __init__(self, workflow, blocksize=None, poll_interval=5, userdata=None):
...
+
+        :return: a mapping from platform to metadata mapping. e.g. {"x86_64": {...}}
+        """
+        metadatas = self.workflow.data.postbuild_results.get(
+            PLUGIN_GATHER_BUILDS_METADATA_KEY, {}
         )
-        if not worker_metadatas:
+        if not metadatas:
             self.log.warning(
                 "No fetched worker metadata is found. Check if %s plugin ran already.",
```
hjmodi, line 147: Should probably update this message as well.

atomic_reactor/plugins/post_koji_import.py
```diff
@@ -157,7 +161,7 @@ def _iter_work_metadata_outputs(
         :type _filter: dict[str, any] or None
         :return: an iterator that yields a tuple in form (platform, output).
         """
-        for worker_platform, metadata in self._worker_metadatas.items():
+        for worker_platform, metadata in self._builds_metadatas.items():
```
hjmodi, line 165: how about `image_platform` or `build_platform` for clarity?

tests/plugins/test_gather_builds_metadata.py
```diff
@@ -0,0 +1,75 @@
...
+                "koji_upload_dir": "path/to/upload",
+            }
+        }],
+    )
+    with pytest.raises(PluginFailedException, match="No enabled platforms"):
+        runner.run()
+
+
+@patch("koji.ClientSession", new=MockedClientSession)
+def test_fetch_worker_plugin(workflow):
```
hjmodi, line 52: probably a good idea to change this to current plugin name as well

atomic_reactor/plugins/post_koji_import.py
tkdchen: :sweat_smile: You are right.

atomic_reactor/utils/koji.py
rcerven: when I was checking brew medata.json I've seen diff only for aarch64

atomic_reactor/plugins/post_gather_builds_metadata.py
```diff
@@ -0,0 +1,136 @@
+"""
+Copyright (c) 2017, 2019 Red Hat, Inc
```
chmeliik, line 2: Should be updated

atomic_reactor/plugins/post_gather_builds_metadata.py
```diff
@@ -0,0 +1,136 @@
...
+
+    Each metadata mapping follows the format of Koji Content Generator Metadata.
+    This plugin ensures it has keys ``metadata_version``, ``buildroots`` and
+    ``output``.
+    """
+
+    key = PLUGIN_GATHER_BUILDS_METADATA_KEY
+    is_allowed_to_fail = False
+
+    args_from_user_params = map_to_user_params("koji_upload_dir")
```
chmeliik: If I recall correctly, this user param used to be generated by the orchestrator plugin. We will not have it in user params anymore, but maybe we can just generate it here. I think another plugin uses the directory as well, we should make sure it's the same directory for both.

atomic_reactor/plugins/post_gather_builds_metadata.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        Build the metadata needed for importing the build
+
+        :return: tuple, the metadata and the list of Output instances
+        """
+        pullspec_image = self._determine_image_pullspec()
+        buildroot = get_buildroot(platform)
+        output_files, _ = get_output(workflow=self.workflow, buildroot_id=buildroot['id'],
+                                     pullspec=pullspec_image, platform=platform,
+                                     source_build=False)
```
chmeliik, line 64: I think we need to use the platform-specific pullspec here

atomic_reactor/inner.py
```diff
@@ -345,6 +345,14 @@ class ImageBuildWorkflowData(ISerializer):
     image_id: Optional[str] = None
     parent_images_digests: Dict[str, Dict[str, str]] = field(default_factory=dict)
 
+    # List of output files that are uploaded to Brew/Koji
+    # Each element is a two-strings list, local_filename and dest_filename. E.g.
+    # [
+    #     {"local_filename": "/path/to/data.json", "dest_filename": "metadata.json"},
+    #     {"local_filename": "/path/to/build.log", "dest_filename": "x86_64-build.log"},
+    # ]
+    koji_upload_files: List[Dict[str, str]] = field(default_factory=list)
```
chmeliik, line 354: This could also be a good way to save logs from `bsi` (source) / `podman` (binary) as a replacement for the removed BuildResult.logs :+1:

atomic_reactor/plugins/post_koji_import.py
```diff
@@ -442,38 +442,54 @@ def get_build(self):
...
+
+        # Collect log files
         osbs_logs = OSBSLogs(self.log, get_platforms(self.workflow.data))
-        output_files = [add_log_type(add_buildroot_id(md, buildroot_id))
-                        for md in osbs_logs.get_log_files(self.osbs, self.pipeline_run_name)]
+        output_log_files = [
+            add_log_type(add_buildroot_id(md, buildroot_id))
+            for md in osbs_logs.get_log_files(self.osbs, self.pipeline_run_name)
+        ]
+        for remote_source_file_output in output_log_files:
```
chmeliik, line 475: Why is it called `remote_source_file_output`? They're just log files, aren't they? :thinking:

atomic_reactor/plugins/post_koji_import.py
```diff
@@ -573,14 +590,17 @@ def run(self):
...
-        finally:
-            for output in output_files:
-                if output.file:
-                    output.file.close()
+        server_dir = self.get_server_dir()
+        koji_metadata = self.combine_metadata_fragments()
+
+        if is_scratch_build(self.workflow):
+            self.upload_scratch_metadata(koji_metadata, server_dir)
+            return
```
chmeliik, line 588: Previously, the plugin did this before checking the task state. I don't know if that is important, but maybe we want to keep it that way?

chmeliik (commented):
Nice! I like the decision to keep some of the metadata gathering in gather_builds_metadata, that makes the changes look simpler than I thought they were going to be :+1:

atomic_reactor/plugins/post_gather_builds_metadata.py
chmeliik: Addressed in later commit, but this is still relevant for `KojiImportPlugin.get_server_dir()` - now it should generate the upload dir instead of getting it from the orchestrator plugin

atomic_reactor/utils/koji.py
```diff
@@ -400,25 +407,24 @@ def get_maven_metadata(workflow_data):
...
-def get_image_components(workflow):
+def get_image_components(workflow, platform: str) -> List[Dict[str, Union[str, int]]]:
     """
     Re-package the output of the rpmqa plugin into the format required
     for the metadata.
     """
-    output = workflow.data.image_components
-    if output is None:
+    components = workflow.data.postbuild_results.get(PostBuildRPMqaPlugin.key)
+    if components is None:
```
hjmodi, line 421: I don't think the result of `rpmqa` will ever be `None`, at most it'll be something likes `{platfrom: None}` since we are using `build_dir.for_each_platfrom` to return the results

tests/plugins/test_gather_builds_metadata.py
```diff
@@ -0,0 +1,75 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
```
hjmodi, line 2:
```suggestion
Copyright (c) 2017-2022 Red Hat, Inc
```

hjmodi (approved):

hjmodi:
LGTM, the code is much more readable now! :tada: 

Just a couple minor comments

atomic_reactor/utils/koji.py
tkdchen: Hi @hjmodi, thanks for your comment. As you can see the following `logger.error`, this check is checking whether `PostBuildRPMqaPlugin` plugin has run. If not, no result in the `postbuild_results`.

atomic_reactor/plugins/post_koji_import.py
```diff
@@ -15,20 +15,21 @@
...
+import koji_cli.lib
+
 from atomic_reactor.config import get_koji_session, get_openshift_session
 from atomic_reactor import start_time as atomic_reactor_start_time
 from atomic_reactor.plugin import PostBuildPlugin
 from atomic_reactor.plugins.pre_add_help import AddHelpPlugin
 from atomic_reactor.source import GitSource
-from atomic_reactor.plugins.build_orchestrate_build import (get_worker_build_info,
-                                                            get_koji_upload_dir)
+from atomic_reactor.plugins.build_orchestrate_build import get_worker_build_info
```
chmeliik, line 25: We should not be importing this anymore. Should be fine to address this in a separate PR.

tests/utils/test_koji.py
```diff
@@ -416,3 +428,146 @@ def test_with_notable(self, totalsize, step, notable, expected_times):
...
+    if has_export_operator_manifests:
+        assert len(output) == 2
+    else:
+        assert len(output) == 1
+
+    image_metadata = output[0].metadata
+
+    # Assert these image metadata firstly, then remove them and assert the
+    # rest. So, no need to mock anything for get_image_output.
+    assert 'docker-image-None.x86_64.tar.gz' == image_metadata.pop('filename')
```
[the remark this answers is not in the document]
tkdchen: It is the `image_id` (`workflow.data.image_id`), which will be fixed in next step.

tests/utils/test_koji.py
```diff
@@ -416,3 +428,146 @@ def test_with_notable(self, totalsize, step, notable, expected_times):
...
+    workflow.conf.conf = {
+        'registries': [
+            {'url': 'https://registry.host/', 'insecure': False},
+        ],
+    }
+    # Mock get manifest digests
+    # What would happen if there is no both v2 and oci digest?
+    image_manifest_digest = ManifestDigest(
+        {'oci': 'oci-1234'} if no_v2_digest else {'v2': '1234'}
+    )
```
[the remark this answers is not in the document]
chmeliik: ah, right :smile:

tests/utils/test_koji.py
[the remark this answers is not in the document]
chmeliik: Ack, seems reasonable

tests/utils/test_koji.py
```diff
@@ -416,3 +428,146 @@ def test_with_notable(self, totalsize, step, notable, expected_times):
...
+        'arch': platform,
+        'type': 'docker-image',
+        'components': [],
+        'extra': {
+            'image': {'arch': platform},
+            'docker': {
+                'id': None,
+                'repositories': expected_repositories,
+                'layer_sizes': layer_sizes,
+                'tags': sorted([primary_image.tag, unique_image.tag]),
```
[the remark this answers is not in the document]
tkdchen: I'll address in next step as well. Will open another PR.

chmeliik (approved):
LGTM :+1:

Were you going to squash the commits or keep all 18? I don't really mind either way

atomic_reactor/utils/koji.py
hjmodi: hmm, I do understand that. It'll only not run for source build in which case this function will not be called. So I don't think we need to check for that. It might lead to confusion. It'll always run for the binary build, if no RPMs are found it'll return a `{plafrom: None}` dictionary

tkdchen:
The build-osbs-components is still unavailable. I'm going to merge this pr for now.

@chmeliik Keep these 18 commits :sweat_smile:

atomic_reactor/utils/koji.py
tkdchen: My understand, the original decision of this check is based on the plugins relationship rather than what kind of data is set to the `image_components`. For now, I'd like keep this original behavior unchanged. Maybe, `log.warning` is more proper than `log.error`.

PR #1855, opened 2022-03-21
Add tox to test container

tkdchen:
With this patch, I can easily run various tests with a simple tox
command instead of repeating `python3 -m pytest ...', `python3 -m
pylint' and other checks. Meanwhile, the tox.ini also includes flake8
and mypy, which I used frequently before creating a pull request.

Hope this is also helpful for you.

- n/a Code coverage from testing does not decrease and new code is covered

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ dist/
 *.egg-info/
 .eggs/
 build/
+*~
 
 # Unit test / coverage reports
 .cache
@@ -11,3 +12,4 @@ build/
 .coverage.*
 htmlcov
 __pytest_reports
+.tox/
--- a/README.md
+++ b/README.md
@@ -74,6 +74,30 @@ missing C libraries in the Dockerfile(s) as well as the test.sh script
 - [Dockerfile](Dockerfile)
 - [test.sh](test.sh)
 
+## Running tests
+
+The prerequisite of running tests is to create a test environment container by
+`test.sh`. For example:
+
+```bash
+OS=fedora OS_VERSION=35 PYTHON_VERSION=3 ACTION=test ./test.sh
+```
+
+When the container is ready and running, you have choice to test your changes
+by executing `pytest` directly:
+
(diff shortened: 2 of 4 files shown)


ben-alkov:
@tkdchen; IIUC, your 'tox.ini' will use whichever python runtime is the current default on the local system. Is this desirable? Would it be better to pin it with e.g. 

```ini
[tox]
envlist = py3.8,test,flake8,pylint,bandit,mypy
...etc
```

tkdchen:
@ben-alkov python3.8 is installed in the centos container, python3 is linked to it. Similarly, in a Fedora 35 container, python3 is the 3.10.

tkdchen:
@ben-alkov BTW, why do we need a CentOS container to run tests? IIUC, we just have to ensure a-r works with Python version >=3.8, and no need to care about which OS to run inside. Any other testing purpose I missed? Thanks.

chmeliik:
This is really nice! Is there a good place in the README where could put the usage example? It might be confusing for some folk that we don't run tox with just `tox` but with `podman exec -ti <test.sh-container> tox`

tkdchen:
If in the README, perhaps we can add a new section before Usage and call it "Running tests". Or, it looks there is another option in osbs.readthedoc.io, in the page "Contributing to OSBS". Which one is the team's preference to add such a description?

chmeliik:
> If in the README, perhaps we can add a new section before Usage and call it "Running tests". Or, it looks there is another option in osbs.readthedoc.io, in the page "Contributing to OSBS". Which one is the team's preference to add such a description?

I'd vote for the README, this is specific to atomic-reactor whereas osbs.readthedocs is about OSBS as a whole

tkdchen:
@chmeliik PTAL

chmeliik (approved):
:tada:

lkolacek (approved):
Nice! Thanks for doc update :+1:

PR #1856, opened 2022-03-22
Annotate type of Plugin workflow argument

tkdchen:
By adding this type annotation, it will make code edit tools, like IDE
or LSP backend, happy and be easy to navigate to the related
symbols.

### DIFF
--- a/atomic_reactor/plugins/pre_bump_release.py
+++ b/atomic_reactor/plugins/pre_bump_release.py
@@ -77,7 +77,7 @@ class BumpReleasePlugin(PreBuildPlugin):
         """Get next release for build."""
         if is_scratch_build(self.workflow):
             # no need to append for scratch build
-            next_release = self.workflow.user_params.get('pipeline_run_name')
+            next_release = self.workflow.user_params['pipeline_run_name']
         elif self.append:
             next_release = self.get_next_release_append(component, version, release)
         else:
--- a/atomic_reactor/plugin.py
+++ b/atomic_reactor/plugin.py
@@ -22,10 +22,13 @@ import time
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
 from dataclasses import dataclass
-from typing import Any, Dict, Generator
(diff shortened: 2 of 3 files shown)


MartinBasti (approved):

atomic_reactor/inner.py
```diff
@@ -668,12 +659,22 @@ def build_process_failed(self):
...
         build docker image
         """
+        from atomic_reactor.plugin import (
+            BuildStepPluginsRunner,
+            ExitPluginsRunner,
+            PluginFailedException,
+            PostBuildPluginsRunner,
+            PreBuildPluginsRunner,
+            PrePublishPluginsRunner,
+        )
```
chmeliik, line 676: Is this the only way to get around the cyclic import? It feels strange to do it like this :(

ben-alkov (approved):

atomic_reactor/inner.py
tkdchen: PTAL

atomic_reactor/plugins/pre_bump_release.py
```diff
@@ -77,7 +77,7 @@ def next_release_general(
         """Get next release for build."""
         if is_scratch_build(self.workflow):
             # no need to append for scratch build
-            next_release = self.workflow.user_params.get('pipeline_run_name')
+            next_release = self.workflow.user_params['pipeline_run_name']
```
lkolacek, line 80: Why this change? to make it fail with `KeyError` when `pipeline_run_name` is not available?

chmeliik: IIUC: the code doesn't work if `next_release` is None, this change ensures that it's a string (or KeyError, yeah)

chmeliik (approved):
LGTM :+1:

Nice find on `typing.TYPE_CHECKING` :rocket:

hjmodi (approved):

lkolacek (approved):

PR #1858, opened 2022-03-22
update `export_operator_manifests` plugin

hjmodi:
* Get the manifests from image using `oc image extract`
* Remove orchestrator or worker paradigm

* CLOUDBLD-9056


rcerven (approved):

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -78,79 +47,52 @@ def run(self):
...
                     filepath = os.path.join(filedir, f)
                     archive.write(os.path.join(root, f), filepath, zipfile.ZIP_DEFLATED)
             manifest_files = archive.namelist()
-            if not manifest_files:
-                self.log.error('Empty operator manifests directory')
-                raise RuntimeError('Empty operator manifests directory')
             self.log.debug("Archiving operator manifests: %s", manifest_files)
 
-        shutil.rmtree(manifests_path)
+        shutil.rmtree(manifests_dir)
```
tkdchen, line 78: Why remove `manifests_dir` instead of the whole `tmp_dir`?

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -78,79 +47,52 @@ def run(self):
...
-            for chunk in bits:
-                extracted_file.write(chunk)
-            extracted_file.flush()
-            tar_archive = tarfile.TarFile(extracted_file.name)
-
-        tar_archive.extractall(manifests_archive_dir)
-        manifests_path = os.path.join(manifests_archive_dir, MANIFESTS_DIR_NAME)
+        platforms = get_platforms(self.workflow.data)
+        image = self.workflow.data.tag_conf.get_unique_images_with_platform(platforms[0])[0]
+        tmp_dir = tempfile.mkdtemp(dir=self.workflow.build_dir.platform_dir(platforms[0]).path)
```
tkdchen, line 59: How about use `BuildDir.any_platform` consistently in this plugin?

tkdchen (commented):
LGTM. Just two minor comments.

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -78,79 +47,52 @@ def run(self):
...
+            for root, _, files in os.walk(manifests_dir):
                 for f in files:
-                    filedir = os.path.relpath(root, manifests_path)
+                    filedir = os.path.relpath(root, manifests_dir)
                     filepath = os.path.join(filedir, f)
                     archive.write(os.path.join(root, f), filepath, zipfile.ZIP_DEFLATED)
             manifest_files = archive.namelist()
-            if not manifest_files:
-                self.log.error('Empty operator manifests directory')
-                raise RuntimeError('Empty operator manifests directory')
```
chmeliik, line 134: Why remove this check?

Edit: is it because extract_file_from_image already sort of does that?

atomic_reactor/plugins/post_export_operator_manifests.py
hjmodi: I was a little hesitant because I wasn't sure if we should rely on `BuildDir.any_platform` to return the directory for platform[0] only. I'll look into it and make sure it's consitant.

chmeliik: IMO it doesn't matter which platform we use, `any_platform` should be OK

hjmodi: updated

atomic_reactor/plugins/post_export_operator_manifests.py
hjmodi: updated to remove tmp_dir itself.

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -78,79 +47,52 @@ def run(self):
...
         """
         try:
-            image_csv = OperatorManifest.from_directory(manifests_path).csv
+            image_csv = OperatorManifest.from_directory(manifests_dir).csv
         except ValueError as e:
             raise ValueError(f'Operator manifests check in built image failed: {e}') from e
 
-        repo_csv = OperatorManifest.from_directory(self.workflow.source.manifests_dir).csv
+        repo_csv = OperatorManifest.from_directory(os.path.join(self.workflow.source.path,
+                                                                MANIFESTS_DIR_NAME)).csv
```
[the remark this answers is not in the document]
hjmodi: updated to use build dir and added annotations

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -78,79 +53,53 @@ def run(self):
...
+
+        self.workflow.imageutil.extract_file_from_image(image, IMG_MANIFESTS_PATH, tmp_dir)
 
         if has_operator_bundle_manifest(self.workflow):
-            self._verify_csv(manifests_path)
+            self._verify_csv(manifests_dir)
 
-        manifests_zipfile_path = os.path.join(manifests_archive_dir, OPERATOR_MANIFESTS_ARCHIVE)
+        manifests_zipfile_path = (self.workflow.build_dir.any_platform.path /
+                                  OPERATOR_MANIFESTS_ARCHIVE)
```
chmeliik, line 75: Previously, we zipped the manifests that we extracted from the image. Now, we're taking the manifests from the build dir instead. Those should usually be identical, but we can't be sure (the only file we verify is the ClusterServiceVersion one).

Could we keep it as it was before and take the extracted ones?

atomic_reactor/plugins/post_export_operator_manifests.py
```diff
@@ -78,79 +53,53 @@ def run(self):
...
-        with tempfile.NamedTemporaryFile() as extracted_file:
-            for chunk in bits:
-                extracted_file.write(chunk)
-            extracted_file.flush()
-            tar_archive = tarfile.TarFile(extracted_file.name)
-
-        tar_archive.extractall(manifests_archive_dir)
-        manifests_path = os.path.join(manifests_archive_dir, MANIFESTS_DIR_NAME)
+        platforms = get_platforms(self.workflow.data)
+        image: ImageName = self.workflow.data.tag_conf.get_unique_images_with_platform(
```
chmeliik, line 64: IMO type annotations for variables are usually unnecessary (though it doesn't hurt to keep them). It is more important to annotate function declarations, even if it's just `def foo() -> None`, because mypy doesn't check functions that don't have any annotations :disappointed:

atomic_reactor/plugins/post_export_operator_manifests.py
hjmodi: There seems to be some confusion here. We're taking it from extracted dir only(`tmp_dir`). It might be because I have named the extracted manifests directory as `manifest_dir`

chmeliik: Oh, yeah, don't mind me, I mixed up which path is being `os.walk`-ed and which is being zipped :smile:

chmeliik (approved):
LGTM :+1:

PR #1859, opened 2022-03-22
Implement the build task

chmeliik:
CLOUDBLD-8136

In the build task, acquire a slot on a remote host. Make sure to unlock
this slot after the build ends.

Run an actual build on the acquired remote host.

...and more fixes, see individual commits

### DIFF
--- a/atomic_reactor/tasks/binary_container_build.py
+++ b/atomic_reactor/tasks/binary_container_build.py
@@ -73,7 +73,9 @@ class BinaryBuildTask(Task[BinaryBuildTaskParams]):
                 logger.info, "Dockerfile used for build:\n%s", build_dir.dockerfile_path.read_text()
             )
 
-            output_lines = self.build_container(
+            podman_remote = PodmanRemote()
+
+            output_lines = podman_remote.build_container(
                 build_dir=build_dir,
                 build_args=data.buildargs,
                 dest_tag=dest_tag,
@@ -82,7 +84,11 @@ class BinaryBuildTask(Task[BinaryBuildTaskParams]):
                 logger.info(line.rstrip())
 
             logger.info("Build finished succesfully! Pushing image to %s", dest_tag)
-            self.push_container(dest_tag, registry_config=config.registry)
+            podman_remote.push_container(dest_tag, registry_config=config.registry)
+
+
+class PodmanRemote:
+    """Wrapper for running podman --remote commands on a remote host."""
 
     def build_container(
         self,
--- a/atomic_reactor/tasks/binary_container_build.py
+++ b/atomic_reactor/tasks/binary_container_build.py
@@ -11,7 +11,7 @@ import logging
 import shutil
 import subprocess
 from dataclasses import dataclass
-from typing import Any, Dict, Iterator, List, Optional
+from typing import Dict, Iterator, List, Optional
 
 from osbs.utils import ImageName
 
@@ -79,7 +79,9 @@ class BinaryBuildTask(Task[BinaryBuildTaskParams]):
             remote_resource = self.acquire_remote_resource(config.remote_hosts)
             defer.callback(remote_resource.unlock)
 
-            podman_remote = PodmanRemote.setup_for(remote_resource)
+            podman_remote = PodmanRemote.setup_for(
+                remote_resource, registries_authfile=config.registry.get("secret")
+            )
 
             output_lines = podman_remote.build_container(
                 build_dir=build_dir,
@@ -90,7 +92,7 @@ class BinaryBuildTask(Task[BinaryBuildTaskParams]):
                 logger.info(line.rstrip())
 
             logger.info("Build finished succesfully! Pushing image to %s", dest_tag)
-            podman_remote.push_container(dest_tag, registry_config=config.registry)
+            podman_remote.push_container(dest_tag, insecure=config.registry.get("insecure", False))
 
     def acquire_remote_resource(self, remote_hosts_config: dict) -> remote_host.LockedResource:
         """Lock a build slot on a remote host."""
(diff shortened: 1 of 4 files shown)


chmeliik:
For the last commit: it is not the right solution by any means, but it does work for our use cases so I'm thinking we could merge it and address it later.

The right solution might be to pull all base images before starting the build and pass the correct --authfile and --tls-verify to each pull command. Or maybe not, because then builds could end up using the base images pulled by other builds (race cnoditions). Probably not a problem because we pin to digest, but still.

Another solution could be to change the way we do configuration, so that all the registries *have to* share a single authfile. And --tls-verify could be all or nothing - nothing is insecure or everything is insecure.

atomic_reactor/tasks/binary_container_build.py
```diff
@@ -98,7 +178,23 @@ def build_container(
...
         and stderr of the build process as they become available.
         """
-        build_cmd = ["/bin/sh", "-c", "for i in 1 2 3; do echo output $i; sleep 0.1; done"]
+        cli_buildargs = [f"--build-arg={key}={value}" for key, value in build_args.items()]
+        build_cmd = [
+            *self._podman_remote_cmd,
+            "build",
+            f"--tag={dest_tag}",
+            str(build_dir.path),
+            "--memory-swap=0",  # TBD: why is this important?
```
chmeliik, line 187: @lkolacek @MartinBasti I found this in design, but I'm not actually share what it does or why we want it

atomic_reactor/tasks/binary_container_build.py
```diff
@@ -86,9 +92,68 @@ def execute(self) -> None:
...
+    def __init__(self, connection_name: str):
+        """Initialize a PodmanRemote instance.
+
+        :param connection_name: the name of an existing podman-remote connection
+        """
+        self._connection_name = connection_name
+
+    @property
+    def _podman_remote_cmd(self) -> List[str]:
+        return [which_podman(), "--remote", f"--connection={self._connection_name}"]
```
rcerven, line 137: --remote is just for podman binary, podman-remote doesn't have it anymore as it is remote by default

atomic_reactor/tasks/binary_container_build.py
```diff
@@ -86,9 +92,68 @@ def execute(self) -> None:
...
+
+    @classmethod
+    def setup_for(cls, host: remote_host.RemoteHost):
+        """Set up a connection for the specified remote host and return a PodmanRemote instance."""
+        connection_name = host.hostname
```
rcerven, line 142: can we rather use pipeline run name for connection name?

atomic_reactor/tasks/binary_container_build.py
```diff
@@ -107,7 +172,19 @@ def build_container(
...
+        cli_buildargs = [f"--build-arg={key}={value}" for key, value in build_args.items()]
+        build_cmd = [
+            *self._podman_remote_cmd,
+            "build",
+            f"--tag={dest_tag}",
+            str(build_dir.path),
+            "--memory-swap=0",  # TBD: why is this important?
+            "--no-cache",  # make sure the build uses a clean environment
+            "--pull-always",  # as above
+            # TBD: --authfile? --tls-verify?
```
rcerven: we don't need auth for pulling, that should be covered by proxies

atomic_reactor/tasks/binary_container_build.py
chmeliik: Podman-remote does have `--remote`. You're right that --remote is default for podman-remote, but IMO it doesn't hurt to pass it anyway for simplicity and consistency

atomic_reactor/tasks/binary_container_build.py
chmeliik: updated

tkdchen (approved):

atomic_reactor/tasks/binary_container_build.py
lkolacek: I haven't found --remote option for podman-remote but those docs I checked might be outdated. I can't install it on my system to properly check.

atomic_reactor/tasks/binary_container_build.py
```diff
@@ -182,10 +193,14 @@ def build_container(
             "--memory-swap=0",  # TBD: why is this important?
             "--no-cache",  # make sure the build uses a clean environment
             "--pull-always",  # as above
-            # TBD: --authfile? --tls-verify?
             "--squash",
             *cli_buildargs,
         ]
+        if self._registries_authfile:
+            # TBD: this only works if the OSBS deployment uses a single registry secret
+            # TBD: we also can't properly handle "insecure" config for pull registries
```
lkolacek, line 201: I don't see an issue with using 1 single .dockerconfigjson with all secrets. I believe for all supported registries in OSBS we should have an entry in there or do I miss some other valid scenario?

yeah, I haven't thought about handling insecure configs. Is it a blocker now?

atomic_reactor/tasks/binary_container_build.py
lkolacek: since we cannot use --memory limits now it's useless and can be dropped.

lkolacek:
PR looks good to me. :+1: Thanks @chmeliik !

atomic_reactor/tasks/binary_container_build.py
chmeliik: Ack, will remove

atomic_reactor/tasks/binary_container_build.py
chmeliik: Definitely not a blocker, just something to address in the future :)

Same for the .dockerconfigjson - in our configmap, we happen to use the same one for all the registries. But nothing stops an OSBS admin from using two or more different dockerconfigs.

Both problems are non-blockers for the initial implementation IMO

atomic_reactor/tasks/binary_container_build.py
chmeliik: Yeah, it's not in --help but it does work (I ran the task locally, built an actual container on a remote host and pushed it, all with podman-remote --remote)

chmeliik:
Rebased on master and removed the useless --memory-swap option

atomic_reactor/tasks/binary_container_build.py
rcerven: yeah I've filed different story for those multiple possible secrets for registries and to get rid of that

rcerven (approved):

lkolacek (approved):

PR #1863, opened 2022-03-25
move pipeline_run_name from `user_params` to CLI

hjmodi:
This attribute can be fetched from tekton
context. So, we don't need to provide
it using osbs-client in `user_params`

* CLOUDBLD-9053


rcerven:
PIPELINE_RUN_NAME should be param in ALL tasks for binary/source pipelines

rcerven (approved):

tests/plugins/test_koji_import.py
```diff
@@ -606,9 +606,9 @@ def workflow(workflow):
     """Add additional data to provide pipeline_run_name specifically."""
     workflow.user_params.update({
         'namespace': NAMESPACE,
-        'pipeline_run_name': PIPELINE_RUN_NAME,
         'koji_task_id': MockedClientSession.TAG_TASK_ID,
     })
+    workflow.pipeline_run_name = PIPELINE_RUN_NAME
```
tkdchen, line 611: workflow fixture already has `pipeline_run_name` set.

tests/plugins/test_store_metadata.py
```diff
@@ -60,7 +60,7 @@ def update_labels_on_build(build_id, labels):
      .with_args(**config_kwargs))
 
     workflow.user_params['namespace'] = 'namespace'
-    workflow.user_params['pipeline_run_name'] = 'store_metadata_test'
+    workflow.pipeline_run_name = 'store_metadata_test'
```
tkdchen, line 63: workflow fixture already has pipeline_run_name set. Looks no need to set it again.

tests/plugins/test_koji_import.py
hjmodi: updated

tests/plugins/test_store_metadata.py
hjmodi: updated

PR #1866, opened 2022-03-28
Push built image as docker v2s2

chmeliik:
CLOUDBLD-8136

By default, the podman build command uses the oci format. The podman
push command defaults to the format of the built image.

Explicitly specify v2s2, because some parts of the RH container pipeline
do not support oci yet.

### DIFF
--- a/atomic_reactor/tasks/binary_container_build.py
+++ b/atomic_reactor/tasks/binary_container_build.py
@@ -186,7 +186,8 @@ class PodmanRemote:
         Pass the specified build arguments as ARG values using --build-arg.
 
         The built image will be available locally on the machine that built it, tagged with
-        the specified dest_tag.
+        the specified dest_tag. This method does not specify the format of the built image
+        (nor does the format really matter), but podman will typically default to 'oci'.
 
         This method returns an iterator which yields individual lines from the stdout
         and stderr of the build process as they become available.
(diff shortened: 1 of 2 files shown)


hjmodi (approved):

hjmodi:
should we document this somewhere that images are built as `oci` and pushed as `v2s2`?

chmeliik:
> should we document this somewhere that images are built as `oci` and pushed as `v2s2`?

Added to docstrings at least, I don't think there is a good place for this info in osbs.readthedocs yet

@rcerven I also updated the commands to follow the "options-first" format. It gets a bit awkward when adding options dynamically (`insert(-1, option)`), LMK what you think

rcerven (approved):

rcerven:
> > should we document this somewhere that images are built as `oci` and pushed as `v2s2`?
> 
> Added to docstrings at least, I don't think there is a good place for this info in osbs.readthedocs yet
> 
> @rcerven I also updated the commands to follow the "options-first" format. It gets a bit awkward when adding options dynamically (`insert(-1, option)`), LMK what you think

well that podman pushes default oci image is podman implementation, for our users nothing changes, we always create v2s2

atomic_reactor/tasks/binary_container_build.py
```diff
@@ -240,14 +241,16 @@ def build_container(
     def push_container(self, dest_tag: ImageName, *, insecure: bool = False) -> None:
         """Push the built container (named dest_tag) to the registry (as dest_tag).
 
+        Push the container as v2s2 (Docker v2 schema 2) regardless of the original format.
+
         :param dest_tag: the name of the built container, and the destination for the push
         :param insecure: disable --tls-verify?
         """
-        push_cmd = [*self._podman_remote_cmd, "push", str(dest_tag)]
+        push_cmd = [*self._podman_remote_cmd, "push", "--format=v2s2", str(dest_tag)]
```
hjmodi, line 249:
```suggestion
        options = []
        if self._registries_authfile:
            options.append(f"--authfile={self._registries_authfile}")
        if insecure:
            options.append("--tls-verify=false")
         push_cmd = [*self._podman_remote_cmd, "push", *options, str(dest_tag)]
```

how about this?

PR #1869, opened 2022-03-30
pass `namespace` via CLI

hjmodi:
previously it was passed from OSBS Client
as `user_params`.

Pass it from tekton context using CLI

* CLOUDBLD-9054


hjmodi:
The more I look at this the more I think we don't need to pass the namespace at all since it's fixed AFAIK and it's available to OSBS-client from config https://github.com/containerbuildsystem/osbs-client/blob/master/osbs/api.py#L94. I think if we just remove the namespace and don't pass it to o-c at all we'll be fine.

rcerven (approved):

mkosiarc (approved):

tests/plugins/test_koji_import.py
```diff
@@ -605,10 +605,10 @@ def custom_get(method, url, headers, **kwargs):
 def workflow(workflow):
     """Add additional data to provide pipeline_run_name specifically."""
     workflow.user_params.update({
-        'namespace': NAMESPACE,
         'pipeline_run_name': PIPELINE_RUN_NAME,
         'koji_task_id': MockedClientSession.TAG_TASK_ID,
     })
+    workflow.namespace = NAMESPACE
```
tkdchen, line 611: workflow fixture already has `namespace`.

tests/plugins/test_store_metadata.py
```diff
@@ -58,7 +58,7 @@ def update_labels_on_build(build_id, labels):
      .should_call("__init__")
      .with_args(**config_kwargs))
 
-    workflow.user_params['namespace'] = 'namespace'
+    workflow.namespace = 'namespace'
```
tkdchen, line 61: Same here, workflow fixture already has `namespace`.

tkdchen (approved):
Looks good. Just two minor comments.

tests/plugins/test_koji_import.py
hjmodi: you're right :). Updated.

tests/plugins/test_store_metadata.py
hjmodi: updated

PR #1870, opened 2022-03-30
squash all layers for flatpak builds

hjmodi:
* CLOUDBLD-8138

### DIFF
--- a/atomic_reactor/tasks/binary_container_build.py
+++ b/atomic_reactor/tasks/binary_container_build.py
@@ -66,6 +66,8 @@ class BinaryBuildTask(Task[BinaryBuildTaskParams]):
 
         data = self.load_workflow_data()
         enabled_platforms = util.get_platforms(data)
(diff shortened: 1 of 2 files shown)


mkosiarc (approved):

atomic_reactor/tasks/binary_container_build.py
```diff
@@ -201,6 +203,7 @@ def build_container(
         build_dir: dirs.BuildDir,
         build_args: Dict[str, str],
         dest_tag: ImageName,
+        flatpak: bool,
```
chmeliik, line 206: Maybe this could be a `squash_all` param instead? I would rather have the logic of how flatpaks are different from regular containers in the Task itself rather than the PodmanRemote class, that one should really just handle podman remote commands

hjmodi: hmm, that seems a better idea. Update to pass squash all.

PR #1874, opened 2022-04-01
initialize build_dir for post build plugins in plugin_based execute

rcerven:
which is first postbuild plugin

* CLOUDBLD-9425

### DIFF
--- a/atomic_reactor/cli/task.py
+++ b/atomic_reactor/cli/task.py
@@ -71,7 +71,7 @@ def binary_container_postbuild(task_args: dict):
     """
     params = TaskParams.from_cli_args(task_args)
     task = BinaryPostBuildTask(params)
-    return task.execute()
+    return task.execute(init_build_dirs=True)
 
 
 def binary_container_exit(task_args: dict):
--- a/atomic_reactor/tasks/plugin_based.py
+++ b/atomic_reactor/tasks/plugin_based.py
@@ -7,10 +7,12 @@ of the BSD license. See the LICENSE file for details.
 """
 
 import logging
+from typing import Optional
 
 from atomic_reactor import inner
 from atomic_reactor.tasks import PluginsDef
 from atomic_reactor.tasks.common import Task, ParamsT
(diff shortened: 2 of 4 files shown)


chmeliik (commented):
What about the exit task? IMO this should not be done in plugins but in `PluginBasedTask.prepare_workflow` or `Task.get_build_dir`

atomic_reactor/cli/task.py
```diff
@@ -71,7 +71,7 @@ def binary_container_postbuild(task_args: dict):
     """
     params = TaskParams.from_cli_args(task_args)
     task = BinaryPostBuildTask(params)
-    return task.execute()
+    return task.execute(init_build_dirs=True)
```
hjmodi, line 74: we should move the initialization for `prebuild` here as well

chmeliik: that unfortunately wouldn't work, because the platforms needed for initialization are not known at the start of prebuild yet, only after the check_and_set_platforms plugin finishes

maybe we should have a separate task between Clone and Prebuild which figures out the enabled platforms :thinking:

chmeliik: but that sounds more like a future improvement than an immediate fix

hjmodi: hmm, it'll be really helpful if tekton had a sort of `init` phase where we could do these steps needed to init tasks.

hjmodi: agree with doing it in future

hjmodi (approved):

chmeliik (approved):
LGTM, I'll file a ticket to make build-dir initialization less awkward by moving check_and_set_platforms to a separate task or step before prebuild

rcerven:
> LGTM, I'll file a ticket to make build-dir initialization less awkward by moving check_and_set_platforms to a separate task or step before prebuild

sure, with this code we can just easily do init for pre builds later

rcerven:
rn updated

PR #1875, opened 2022-04-04
rework `flatpak_create_oci`

hjmodi:
* use `ImageUtil` to extract filesystem
* return per-platform image metadata

* CLOUDBLD-6338


atomic_reactor/plugins/post_flatpak_create_oci.py
```diff
@@ -75,112 +41,52 @@ def __init__(self, workflow):
...
+        filesystem_path = os.path.join(tmp_dir, image_filesystem)
 
-        image_id = self._get_oci_image_id(outfile)
-        self.log.info('New OCI image ID is %s', image_id)
-        # OSBS2 TBD
-        self.workflow.data.image_id = image_id
+        with open(filesystem_path, 'rb') as f:
+            # this part is 'not ideal' but this function seems to be a prerequisite
+            # for building flatpak image since it does the setup for it
+            flatpak_filesystem, _ = self.builder._export_from_stream(f)
```
hjmodi, line 66: This part is really cumbersome since we're calling an internal function that is reprocessing the filesystem but without this call, the `build_container` call fails.

atomic_reactor/plugins/post_flatpak_create_oci.py
```diff
@@ -75,112 +41,52 @@ def __init__(self, workflow):
...
+        df_labels = build_dir.dockerfile_with_parent_env(
+            self.workflow.imageutil.base_image_inspect()
+        ).labels
 
-        tarred_filesystem, manifest = self._export_filesystem()
-        self.log.info('filesystem tarfile written to %s', tarred_filesystem)
-        self.log.info('manifest written to %s', manifest)
+        self.builder.add_labels(df_labels)
 
-        image_components = self.builder.get_components(manifest)
```
hjmodi, line 152: was this useful? will this return a result different from `rpmqa` and if so should we update https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/utils/koji.py#L422?

atomic_reactor/plugins/post_flatpak_create_oci.py
chmeliik: Turns out we already used the private function in OSBS 1, so this is probably the best we can do :disappointed: 

https://github.com/containerbuildsystem/atomic-reactor/blob/d7ec9ef5ba2d230aa70d12905f37952a1d2f1414/atomic_reactor/plugins/prepub_flatpak_create_oci.py#L81

atomic_reactor/plugins/post_flatpak_create_oci.py
```diff
@@ -75,112 +41,52 @@ def __init__(self, workflow):
...
-            self.log.info('not flatpak build, skipping plugin')
-            return
-
-        source = get_flatpak_source_info(self.workflow)
-        if source is None:
-            raise RuntimeError("flatpak_create_dockerfile must be run before flatpak_create_oci")
-
-        self.builder = FlatpakBuilder(source, self.workflow.source.workdir,
+    def build_flatpak_image(self, source, build_dir: BuildDir) -> Dict[str, Union[str, int]]:
+        self.builder = FlatpakBuilder(source, build_dir.path,
```
chmeliik, line 45: Could it be a local `builder` instead of `self.builder`? Having it as an attribute makes it look like we want to reuse the same builder for all platforms, which is not the case

atomic_reactor/plugins/post_flatpak_create_oci.py
chmeliik: Hmm, good question. I'd say we should, since we're relying on the logic of FlatpakBuilder here so we cannot guarantee that the results will *not* be different

chmeliik: And OSBS 1 did use the results from flatpak_create_oci - at first I thought that wasn't the case, because `rpmqa` runs later, but it has a workaround just for this
https://github.com/containerbuildsystem/atomic-reactor/blob/d7ec9ef5ba2d230aa70d12905f37952a1d2f1414/atomic_reactor/plugins/post_rpmqa.py#L43-L45

atomic_reactor/plugins/post_flatpak_create_oci.py
hjmodi: cool

atomic_reactor/plugins/post_flatpak_create_oci.py
hjmodi: updated to use image components from results of this plugin

atomic_reactor/plugins/post_flatpak_create_oci.py
hjmodi: updated

rcerven (approved):

atomic_reactor/plugins/post_flatpak_create_oci.py
```diff
@@ -68,119 +34,63 @@ def __init__(self, workflow):
...
-
-        :param oci_path: str, path to OCI directory
-        :param name: str, name to be given to the image in the internal storage
-        :param tag: str, tag to apply to name in the internal storage
-        """
-        skopeo_copy_dst = '{}:{}:{}'.format(self.workflow.storage_transport, name, tag)
-        cmd = ['skopeo', 'copy', 'oci:{}'.format(oci_path), skopeo_copy_dst]
+    def build_flatpak_image(
+            self, source, build_dir: BuildDir
+    ) -> Dict[str, Union[Dict[str, Union[str, int]], List[Dict[str, Union[str, int]]]]]:
```
chmeliik, line 45: When it gets this complicated, I feel like a `Dict[str, Any]` annotation would be better

The perfect solution would be a [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict), but we haven't used them anywhere else yet

atomic_reactor/plugins/post_flatpak_create_oci.py
```diff
@@ -68,119 +34,63 @@ def __init__(self, workflow):
...
 
         self.log.info('OCI image is available as %s', outfile)
 
-        metadata = get_exported_image_metadata(tarred_outfile, IMAGE_TYPE_OCI_TAR)
-        metadata['ref_name'] = ref_name
+        shutil.rmtree(tmp_dir)
+
+        return {'metadata': metadata, 'components': image_rpm_components}
+
+    def run(self) -> Optional[Dict[str, Dict[str, Union[str, int]]]]:
```
chmeliik, line 85: I think the type should be `Optional[Dict[str, <ret. type of build_flatpak_image>]]`

tests/utils/test_koji.py
```diff
@@ -533,7 +556,7 @@ def test_binary_build_get_output(has_export_operator_manifests: bool,
         'checksum_type': 'md5',
         'arch': platform,
         'type': 'docker-image',
-        'components': [],
+        'components': expected_components,
```
chmeliik, line 559: Thanks for adding the test! :tada:

test.sh
```diff
@@ -50,7 +50,7 @@ function setup_osbs() {
 
   # List common install dependencies
   PKG_COMMON_EXTRA=(git-core gcc krb5-devel "$PY_PKG-devel" popt-devel redhat-rpm-config
-                    cairo-devel gobject-introspection-devel cairo-gobject-devel)
+                    cairo-devel gobject-introspection-devel cairo-gobject-devel ostree)
```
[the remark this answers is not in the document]
hjmodi: actually, we did have it before and it was removed by https://github.com/containerbuildsystem/atomic-reactor/commit/75a4fa9cb766c49663f392c15ab24b0e5486f73d. My guess is it didn't fail till now since we were skipping flatpak tests.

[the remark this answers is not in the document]
chmeliik: Ah, fair enough, let's keep it then :+1:

atomic_reactor/plugins/post_flatpak_create_oci.py
hjmodi: updated

atomic_reactor/plugins/post_flatpak_create_oci.py
hjmodi: ah, that should be very helpful in future as well

hjmodi: updated

tests/plugins/test_flatpak_create_oci.py
```diff
@@ -459,22 +448,52 @@ def write_docker_file(config, source_dir: Path):
                                   stream=base_module['stream'],
                                   version=base_module['version'])))
 
-    return dockerfile
+
+def mock_extract_filesystem(config, src, dest):
+    build_dir = Path('/'.join(src.split('/')[:-1]))
+    tmp_dir = build_dir / dest.split('/')[-1]
```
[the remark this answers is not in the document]
hjmodi: oh yes, right. Updated.

chmeliik (approved):
LGTM :+1:

PR #1876, opened 2022-04-04
Drop build-host label, log the build host instead

chmeliik:
CLOUDBLD-8832

### DIFF
--- a/atomic_reactor/plugins/pre_add_labels_in_df.py
+++ b/atomic_reactor/plugins/pre_add_labels_in_df.py
@@ -81,8 +81,7 @@ class AddLabelsPlugin(PreBuildPlugin):
                  auto_labels=("build-date",
                               "architecture",
                               "vcs-type",
-                              "vcs-ref",
-                              "com.redhat.build-host"),
+                              "vcs-ref"),
                  aliases=None,
                  dont_overwrite_if_in_dockerfile=("distribution-scope",
                                                   "com.redhat.license_terms")):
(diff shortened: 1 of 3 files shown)


chmeliik:
The ticket mentions that we should also log the NVR. I decided not to do that, because properly figuring out the NVR requires inspecting the base image and that seems like too much effort for a single log message. Knowing the pullspec should be good enough, we can get the NVR with e.g. skopeo:
```bash
skopeo inspect --config docker://$pullspec |
    jq '.config.Labels | .["com.redhat.component"] as $name | "\($name)-\(.version)-\(.release)"'
```

rcerven (approved):

hjmodi (approved):

chmeliik:
docs PR: https://github.com/containerbuildsystem/osbs-docs/pull/212

chmeliik:
So, I just found out that not setting the label doesn't mean it will not be in the final image - if the parent image was built on OSBS 1 and therefore has the build-host label, then the child image will inherit it :disappointed: 

Best we can do is probably set the label to an empty value

ben-alkov (approved):

chmeliik:
> So, I just found out that not setting the label doesn't mean it will not be in the final image - if the parent image was built on OSBS 1 and therefore has the build-host label, then the child image will inherit it disappointed
> 
> Best we can do is probably set the label to an empty value

(this would need to be done explicitly, TBD)

chmeliik:
It was decided that since the problem should be only temporary (until all base images are rebuilt on OSBS 2), we should go ahead with this change

Merging

chmeliik:
Added to release notes (Incompatible Changes section):
> - OSBS no longer provides the `com.redhat.build-host` label
>   - However, if the base image has this label, the child will inherit it and it will be incorrect
>   - This is a temporary problem until base images are rebuilt on OSBS 2

PR #1878, opened 2022-04-06
Refactor koji_import get_output

tkdchen:
The major change is to move code handling outputs to proper koji_import
classes properly.

The exported operator manifests is now collected in the
KojiImportPlugin.get_output.


rcerven (approved):

atomic_reactor/plugins/post_koji_import.py
```diff
@@ -613,20 +564,126 @@ def _filesystem_koji_task_id(self) -> Optional[int]:
...
+
+        for platform, output in self._iter_build_metadata_outputs():
+            buildroot_id = output['buildroot_id']
+            output['buildroot_id'] = f'{platform}-{buildroot_id}'
+            if maven_components and output['type'] == 'docker-image':
+                # add maven components alongside RPM components
+                output['components'] += maven_components
+            outputs.append(output)
+
+        buildroot_id = outputs[0]['buildroot_id']
```
chmeliik, line 679: This method first sets the buildroot_id of all per-platform outputs to `{platform}-{buildroot_id}`, then chooses an arbitrary `{platform}-{buildroot_id}` to use for the rest of the outputs. Shouldn't those remaining outputs use a platform-less id?

atomic_reactor/plugins/post_koji_import.py
```diff
@@ -613,20 +564,126 @@ def _filesystem_koji_task_id(self) -> Optional[int]:
...
+        self._collect_exported_operator_manifests_output(buildroot_id, outputs)
+        self._collect_source_tarballs_outputs(buildroot_id, outputs)
+        self._collect_remote_sources_json_outputs(buildroot_id, outputs)
+        self._collect_maven_metadata(buildroot_id, outputs)
```
chmeliik, line 683: The `collect_*` methods all have a lot of things in common:
* they append to the `outputs` param
* they append to `self.workflow.data.koji_upload_files`
* they set the `buildroot_id` on outputs

Is there a reason to do it in those methods instead of doing it here? Something like
```python
platform_independent_outputs = [
    *self._collect_a(), *self._collect_b(), ...
]
for output in platform_indepent_outputs:
    # set buildroot_id, append to koji_upload_files, ...
```

atomic_reactor/plugins/post_koji_import.py
```diff
@@ -613,20 +564,126 @@ def _filesystem_koji_task_id(self) -> Optional[int]:
...
+            outputs_metadata.append(metadata)
+
+            wf_data.koji_upload_files.append({
+                'local_filename': local_filename,
+                'dest_filename': dest_filename,
+            })
+
+    def _collect_remote_sources_json_outputs(
+            self, buildroot_id: str, outputs_metadata: List[Dict[str, Any]]
+    ) -> None:
```
chmeliik, line 591: I'd prefer just one method for collecting remote source metadata instead of having one for tarballs and one for JSON output

atomic_reactor/plugins/post_koji_import.py
```diff
@@ -722,12 +779,22 @@ class KojiImportSourceContainerPlugin(KojiImportBase):
...
+            workflow=self.workflow,
+            buildroot_id=buildroot_id,
+            pullspec=pullspec,
+            platform=os.uname()[4],
+            source_build=True,
+        )
+        self.workflow.data.koji_upload_files.append({
+            "local_filename": output_file.filename,
+            "dest_filename": metadatas[0]['filename'],
+        })
```
chmeliik, line 795: Hmm, so the `output_file` returned from `koji_get_output()` has to correspond to `metadatas[0]`?

atomic_reactor/plugins/post_koji_import.py
tkdchen: > buildroot_id: The id of the buildroot used to create this file. Must match an entry in the buildroots list.[1]

The Koji doc tells this.

[1] https://docs.pagure.org/koji/content_generator_metadata/#output

atomic_reactor/plugins/post_koji_import.py
tkdchen: One major reason is trying to keep the caller side less code as much as possible. Meanwhile, for the individual collect method, it is also natural to express the meaning, a collect method will collect output metadata and corresponding files into proper place for later operation. As a result, there is no such a cycle, calling a collect method, the method returns something, caller operates the return value.

tkdchen: It was actually a little painful for me when I was handling the outputs related methods. I thought it would be nice to make the collection action like a stream, from where to get the each output metadata and associated file. But, didn't figure out a good way yet.

atomic_reactor/plugins/post_koji_import.py
tkdchen: This is what I understand. For source build, the first element of the return value is a list of metadata, only one there, and the second element is the Output object including the file.

atomic_reactor/plugins/post_koji_import.py
chmeliik: I see, and this _is_ the same as what we had in OSBS 1, right?

atomic_reactor/plugins/post_koji_import.py
chmeliik: I feel like this relies on the internal logic of `koji_get_output()` too much. Could the dest_filename be `output_file.metadata['filename']`?

atomic_reactor/plugins/post_koji_import.py
chmeliik: Ok, seems reasonable

atomic_reactor/plugins/post_koji_import.py
tkdchen: I think so. In OSBS 1, the first platform buildroot id is used to collect the artifacts. In this pr, the first output has the first platform buildroot id.

atomic_reactor/plugins/post_koji_import.py
tkdchen: Updated.

atomic_reactor/plugins/post_koji_import.py
```diff
@@ -564,89 +566,46 @@ def _filesystem_koji_task_id(self) -> Optional[int]:
             self.log.error("invalid task ID %r", fs_task_id, exc_info=1)
             return None
 
-    def _collect_source_tarballs_outputs(
-            self, buildroot_id: str, outputs_metadata: List[Dict[str, Any]]
-    ) -> None:
+    def _collect_remote_source_outputs(self) -> Iterable:
```
chmeliik, line 569: Would it be helpful to further specify `Iterable[<inner type>]`?

chmeliik (commented):
+1 for the merged collect methods :+1: 

Just one minor typing suggestion, otherwise LGTM

chmeliik (approved):
LGTM :+1:

PR #1882, opened 2022-04-07
Upload platform-specific build logs

tkdchen:

### DIFF
--- a/tests/plugins/test_gather_builds_metadata.py
+++ b/tests/plugins/test_gather_builds_metadata.py
@@ -6,7 +6,9 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
 
+import re
 from unittest.mock import patch
+from atomic_reactor.inner import DockerBuildWorkflow
 
 from atomic_reactor.utils.koji import get_buildroot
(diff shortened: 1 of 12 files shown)


chmeliik (commented):
Nice! Some minor suggestions

tests/plugins/test_gather_builds_metadata.py
```diff
@@ -72,4 +89,23 @@ def test_gather_builds_metadata(workflow):
...
+            expected_platform="s390x",
+        )
+
+        upload_file_info = {
+            "local_filename": str(build_log_file),
+            "dest_filename": build_log_file.name,
+        }
+        assert [upload_file_info] == workflow.data.koji_upload_files
+    else:
+        assert re.search(r"not found: .+s390x-build.log", caplog.text, flags=re.MULTILINE)
```
chmeliik, line 107: small detail: I think we don't need MULTILINE, it only affects the behavior of `^` and `$`

chmeliik (approved):
LGTM :+1:

rcerven (approved):

tests/plugins/test_gather_builds_metadata.py
tkdchen: @chmeliik You are right :+1: 

Updated. Since only these two lines change, once the CI passes, I'll merge this PR.

PR #1883, opened 2022-04-08
Fix conversion of config blob to inspect format

chmeliik:
CLOUDBLD-9746

The container_config and docker_version keys are docker-specific. Now
that we have moved to building with podman, we should not try to use
them anymore.

Use the OCI spec [1] as reference, handle the keys accordingly.

Note: this spec is for the OCI config, not the Docker config, but it
aims to be compatible with the Docker format and it is a better formal
specification than the non-existent Docker one.

[1]: https://github.com/opencontainers/image-spec/blob/main/config.md#properties

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -885,16 +885,6 @@ class RegistryClient(object):
             raise RuntimeError("Image {image_name} not found: No v2 schema 1 image, "
                                "or v2 schema 2 image or list, found".format(image_name=image))
 
-        # dictionary to convert config keys to inspect keys
-        config_2_inspect = {
-            'created': 'Created',
-            'os': 'Os',
-            'container_config': 'ContainerConfig',
-            'architecture': 'Architecture',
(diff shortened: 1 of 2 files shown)


atomic_reactor/util.py
```diff
@@ -907,14 +897,16 @@ def get_inspect_for_image(
             # set Id, which isn't in config blob
             # Won't be set for v1,as for that image has to be pulled
             'Id': config_digest,
+            'Architecture': blob_config['architecture'],
+            'Os': blob_config['os'],
+            # According to OCI spec, 'created' and 'config' are optional
+            'Created': blob_config.get('created'),
+            'Config': blob_config.get('config') or {},
```
chmeliik, line 904: Note that Created is not used anywhere in atomic-reactor, so even if it is None, it not break anything. Unlike Config, where some code expects it to always be there as a (possibly empty) dict.

rcerven (approved):

tkdchen (approved):

PR #1886, opened 2022-04-11
Fix flatpak build

hjmodi:


atomic_reactor/plugins/post_flatpak_create_oci.py
```diff
@@ -59,18 +59,24 @@ def build_flatpak_image(self, source, build_dir: BuildDir) -> Dict[str, Any]:
         image_filesystem = self.workflow.imageutil.extract_filesystem_layer(
             str(build_dir.exported_squashed_image), str(tmp_dir))
 
+        build_dir.exported_squashed_image.unlink()
```
chmeliik, line 64: Where does it get recreated? We still need it for later plugins don't we?

atomic_reactor/plugins/post_flatpak_create_oci.py
```diff
@@ -81,6 +83,15 @@ def build_flatpak_image(self, source, build_dir: BuildDir) -> Dict[str, Any]:
         metadata = get_exported_image_metadata(outfile, IMAGE_TYPE_OCI)
         metadata['ref_name'] = ref_name
 
+        cmd = ['skopeo', 'copy', 'oci:{path}:{ref_name}'.format(**metadata), '--format=v2s2',
+               'docker-archive:{}'.format(str(build_dir.exported_squashed_image))]
```
chmeliik, line 87: Oh, it gets recreated in the next commit :D

Do we want to remove the `outfile` after this, or is that used for something later?

atomic_reactor/plugins/post_rpmqa.py
```diff
@@ -37,8 +37,12 @@ def __init__(self, workflow, ignore_autogenerated_gpg_keys=True):
         super(PostBuildRPMqaPlugin, self).__init__(workflow)
         self.ignore_autogenerated_gpg_keys = ignore_autogenerated_gpg_keys
 
-    def run(self):
-        return self.workflow.build_dir.for_each_platform(self.gather_output)
+    def run(self) -> None:
+        if self.workflow.data.image_components:
+            self.log.info('skipping since flatpak_create_oci has already set rpm components')
```
chmeliik, line 45: I would suggest a more generic message, something like the comment we had in OSBS 1
`# If another component has already filled in the image component list, skip`

atomic_reactor/schemas/workflow_data.json
```diff
@@ -40,8 +40,11 @@
 
     "files": {"type": "object"},
     "image_components": {
-      "type": ["array", "null"],
-      "items": {"$ref": "#/definitions/rpm_info"}
+      "type": ["object", "null"],
+      "rpm_components": {
+          "type": ["array"],
+          "items": {"$ref": "#/definitions/rpm_info"}
+      }
```
chmeliik, line 47: I don't think this works as intended, the `rpm_components` key has no meaning to jsonschema

It should be something like this
```json
"image_components": {
  "type": ["object", "null"],
  "patternProperties": {
    ".*": {
      "type": ["array"],
      "items": {"$ref": "#/definitions/rpm_info"}
    }
  }
}
```
(and could this be unit-tested?)

atomic_reactor/inner.py
```diff
@@ -326,9 +326,9 @@ class ImageBuildWorkflowData(ISerializer):
     # "path/to/file" -> "content"
     files: Dict[str, str] = field(default_factory=dict)
 
-    # List of RPMs that go into the final result, as per utils.rpm.parse_rpm_output
+    # Per platform List of RPMs that go into the final result
     # Each RPM inside is a mapping containing the name, version, release and other attributes.
-    image_components: Optional[List[RpmComponent]] = None
+    image_components: Dict[str, List[RpmComponent]] = field(default_factory=dict)
```
chmeliik, line 331: Was there ever a difference between `image_components == None` and `image_components == []`? If yes, we should still make this `Optional`

atomic_reactor/plugins/post_rpmqa.py
```diff
@@ -38,24 +38,21 @@ def __init__(self, workflow, ignore_autogenerated_gpg_keys=True):
         self.ignore_autogenerated_gpg_keys = ignore_autogenerated_gpg_keys
 
     def run(self) -> None:
+        if self.workflow.data.dockerfile_images.base_from_scratch:
+            self.log.info('Skipping plugin, no rpm(s) in FROM-scratch images')
+            return None
```
chmeliik, line 43: I'm not sure if this is true. In fact, now that our base image builds look like this, it's most likely not true
```dockerfile
FROM scratch

ADD base-filesystem.tar.gz /
```

Previously, we would extract the RPMs first, and then, if there were none, check if the image is FROM scratch. I think we still need to do it that way

atomic_reactor/plugins/post_flatpak_create_oci.py
chmeliik: Also, what about the returned metadata? That's still for the oci image, not the v2s2 one?

hjmodi: yes, it's still OCI. I didn't change it to `v2s2` metadata since we are only doing `v2s2` to be able to push the image in a format that the registry supports and it's also not possible to get layer sizes from OCI.

atomic_reactor/inner.py
hjmodi: There is a difference but it's going to be set to either the result of `rpmqa` or `flatpak_create_oci` and in both cases, it'll have keys set to platform so it's not going to be `None`

chmeliik: Oh, true, after some plugin sets this it cannot "falsy" dict since it has platform keys. Should be OK then :+1:

atomic_reactor/plugins/post_rpmqa.py
chmeliik: If it's the right exception, yeah :smile: 

Might be good to give the extract method a parameter to allow the use case where nothing may get extracted

atomic_reactor/plugins/post_flatpak_create_oci.py
chmeliik: I think it's possible, but we would have to update the imageutil method to handle OCI. But there might be more changes needed, so saving it as v2s2 is probably better.

What about removing the outfile? Worth it or no?

atomic_reactor/plugins/post_rpmqa.py
rcerven: 1) extract rpmdb path 
2) check if rpmdb was extracted
3) if it was do query
4) if it wasn't , CHECK if build was FROM scratch , if was return NONE (as scratch are allowed not to have rpmdb)
but IF it has rpmdb (ubi-micro) do rpm query

atomic_reactor/plugins/post_flatpak_create_oci.py
rcerven: we can remove outfile & outfile_tarred I'd say 

since we have already metadata and image in v2s2 ....

hjmodi: unfortunately we cannot remove the outfile since it's the directory that contains the OCI image that is actually pushed by `tag_and_push`.

atomic_reactor/plugins/post_rpmqa.py
hjmodi: updated.

atomic_reactor/schemas/workflow_data.json
hjmodi: fixed and added unitests

atomic_reactor/plugins/post_rpmqa.py
hjmodi: updated to follow this logic

atomic_reactor/plugins/post_flatpak_create_oci.py
rcerven: ic, well we could remove outfile, and just push this v2s2 in tag_and_push plugin

but when everything supports oci images and we would like to push really oci image to registry we would have to revert that change again

so just leave outfile there

rcerven (approved):

atomic_reactor/plugins/post_flatpak_create_oci.py
chmeliik: Hmm, so wait a minute, what is oci and what is v2s2 in the flatpak workflow? We push the oci image and use metadata for the oci image. The v2s2 image is used for getting layer sizes and what else?

This commit also sets the image type to v2s2 for some image (not quite sure what image that is)
https://github.com/containerbuildsystem/atomic-reactor/blob/e10bc1de6635764a4d8663239721b467e52b2bf6/atomic_reactor/utils/koji.py#L517-L518

atomic_reactor/plugins/post_rpmqa.py
```diff
@@ -45,17 +45,19 @@ def run(self) -> None:
...
+            try:
+                self.workflow.imageutil.extract_file_from_image(image, RPMDB_PATH, rpmdb_dir)
+            except ValueError as e:
                 if self.workflow.data.dockerfile_images.base_from_scratch:
-                    return None
-                raise RuntimeError(f'rpmdb directory {RPMDB_PATH} is empty')
+                    self.log.info("scratch image doesn't contain or has empty rpmdb %s", RPMDB_PATH)
+                    return []
+                raise RuntimeError(f"rpmdb directory {RPMDB_PATH} in the image, is empty or "
+                                   "doesn't exist") from e
```
chmeliik, line 58: It seems a bit fragile to assume that "extract_file_from_image() raises ValueError" means "no rpmdb in the image". IMO the extract method either needs a specific exception type for "nothing was extracted" or a param that makes it skip the "nothing was extracted" check

tests/test_schemas.py
```diff
@@ -127,6 +131,85 @@
...
+        with open(os.path.join(d, 'workflow_data.json'), 'w') as f:
+            json.dump(wf_data.as_dict(), f, cls=WorkflowDataEncoder)
+        with open(os.path.join(d, 'workflow_data.json')) as f:
+            workflow_json = json.load(f)
+
+    return workflow_json
+
+
+RPM_COMPONENTS_WORKFLOW_DATA = get_workflow_data_json()
+
```
chmeliik, line 212: I might be missing something but... where is the test? :smile: 

1. pytest only treats functions called `test_*` as tests (AFAIK)
2. this function does not use the workflow data schema

Btw there is a test_inner.py::TestWorkflowData test class, which could also be a good place for the rpm components schema unit test (up to you)

atomic_reactor/plugins/post_flatpak_create_oci.py
chmeliik: I see. Looking closer, it probably doesn't matter what we pass here because the "image" from get_image_output is not an image at all, just the dummy tarball with the README, so v2s2 or oci doesn't really matter

tests/test_schemas.py
hjmodi: I forgot to add it to the fixture. Now it's used in the fixture. I added it to the schema test since we're really testing schema in this test and not the `ImageWorkflowData` object itself, which I'm just using to generate the json.

atomic_reactor/plugins/post_rpmqa.py
hjmodi: refined the handling

chmeliik: Nice! :tada:

chmeliik (approved):
LGTM :+1:

hjmodi:
The failures should be fixed by rebasing on https://github.com/containerbuildsystem/atomic-reactor/pull/1890

PR #1890, opened 2022-04-13
Fix dockerfile parsing in sendmail plugin

chmeliik:
CLOUDBLD-8266

Parse the dockerfile from any build dir, not the source dir.

### DIFF
--- a/atomic_reactor/plugins/exit_sendmail.py
+++ b/atomic_reactor/plugins/exit_sendmail.py
@@ -19,7 +19,6 @@ from atomic_reactor.plugin import ExitPlugin, PluginFailedException
 from atomic_reactor.plugins.post_koji_import import KojiImportPlugin
 from atomic_reactor.plugins.exit_store_metadata import StoreMetadataPlugin
 from atomic_reactor.utils.koji import get_koji_task_owner
-from atomic_reactor.util import df_parser
 from atomic_reactor.constants import PLUGIN_SENDMAIL_KEY
 from atomic_reactor.config import get_koji_session, get_smtp_session
 from osbs.utils import Labels, ImageName
@@ -153,7 +152,9 @@ class SendMailPlugin(ExitPlugin):
     def _get_image_name_and_repos(self):
 
         repos = []
-        dockerfile = df_parser(self.workflow.df_path, workflow=self.workflow)
+        dockerfile = self.workflow.build_dir.any_platform.dockerfile_with_parent_env(
+            self.workflow.imageutil.base_image_inspect()
+        )
         labels = Labels(dockerfile.labels)
         _, image_name = labels.get_name_and_value(Labels.LABEL_TYPE_NAME)
 
--- a/atomic_reactor/cli/task.py
+++ b/atomic_reactor/cli/task.py
@@ -81,4 +81,4 @@ def binary_container_exit(task_args: dict):
     """
     params = TaskParams.from_cli_args(task_args)
     task = BinaryExitTask(params)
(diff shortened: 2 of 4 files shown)


atomic_reactor/plugins/exit_sendmail.py
```diff
@@ -153,7 +152,9 @@ def _should_send(self, success, manual_canceled):
     def _get_image_name_and_repos(self):
 
         repos = []
-        dockerfile = df_parser(self.workflow.df_path, workflow=self.workflow)
+        dockerfile = self.workflow.build_dir.any_platform.dockerfile_with_parent_env(
```
hjmodi, line 155: This will not work without initializing the build dirs

hjmodi:
I am thinking we should remove `df_path` entirely. While testing flatpak builds I realized that `df_path` doesn't get set since flatpak sources don't have a Dockerfile in source. Instead, we can set the platform independent data in `workflow.data` and use that. WDYT?

atomic_reactor/plugins/exit_sendmail.py
chmeliik: Oh, good catch! That's why we need consistency /shakes_fist

chmeliik:
> I am thinking we should remove `df_path` entirely. While testing flatpak builds I realized that `df_path` doesn't get set since flatpak sources don't have a Dockerfile in source. Instead, we can set the platform independent data in `workflow.data` and use that. WDYT?

Definitely remove, it's on the list of things to be removed at some point

What do you mean by platform independent data?

hjmodi:
> > I am thinking we should remove `df_path` entirely. While testing flatpak builds I realized that `df_path` doesn't get set since flatpak sources don't have a Dockerfile in source. Instead, we can set the platform independent data in `workflow.data` and use that. WDYT?
> 
> Definitely remove, it's on the list of things to be removed at some point
> 
> What do you mean by platform independent data?

I am thinking that the Dockerfile might be modified during prebuild, with platform-specific data but my guess is that some things like labels may remain the same for all platforms. We can capture that data if needed instead of trying to retrieve it from `dockerfile` in the exit plugins.

chmeliik:
> I am thinking that the Dockerfile might be modified during prebuild, with platform-specific data but my guess is that some things like labels may remain the same for all platforms. We can capture that data if needed instead of trying to retrieve it from `dockerfile` in the exit plugins.

Hmm, might be worth looking into, but it won't be too simple. We would have to define the data that is platform-independent, and probably have a new plugin / whatever to set this data at the end of the prebuild phase, or at some point in postbuild.

Labels are not platform-indepedent, there is at least one (architecture) that's different between platforms. The name-version-release labels should be independent, and NVR is one of the main use cases of parsing the Dockerfile, so that could be an option. But if we were to do that, we should make it _the_ way to get NVR instead of letting all prebuild plugins do it however they like

TL;DR - for now, the easier solution seems to be letting exit plugins use the dockerfile :)

rcerven (approved):

tkdchen (approved):

hjmodi:
@chmeliik do you need to waive coverage?

chmeliik:
> @chmeliik do you need to waive coverage?

@hjmodi I checked the coveralls report and the 2 newly missed lines are in the `util.df_parser()` method which should be deleted when it's no longer needed (which should be soon, this might have been its last use outside of unit tests)

hjmodi (approved):

hjmodi:
LGTM

PR #1891, opened 2022-04-13
Remove the orchestrator plugin and related code

chmeliik:


atomic_reactor/utils/koji.py
```diff
@@ -550,9 +550,8 @@ def add_buildroot_id(output: Output) -> Output:
     # Add the 'docker save' image to the output
     image = add_buildroot_id(output)
 
-    # when doing regular build, worker already uploads image,
-    # so orchestrator needs only metadata,
-    # but source contaiener build didn't upload that image yet,
+    # when doing a binary build, the build task already uploads image,
+    # so we need only metadata, but source container build didn't upload that image yet,
     # so we want metadata, and the image to upload
```
chmeliik, line 555: Pretty sure this is still nonsense, the thing is that we *do* actually seem to upload something extra for source containers. I'll have to look into what that is

tkdchen: For uploading a built image koji metadata, koji_import seems generate a "fake" image output file (returned from get_image_output) to represent the built image. From my point of view, these lines of comment is a little useless, and the `get_image_output` function should be described clearly instead, especially what concrete content to be uploaded to the Koji/Brew as part of the metadata.

tkdchen (commented):
LGTM :+1:

rcerven (approved):

atomic_reactor/utils/koji.py
chmeliik: Ack, deleted the comment

tkdchen (approved):

PR #1893, opened 2022-04-18
Get rid of types of plugins and runners 9382

tkdchen:


chmeliik (commented):
Mostly LGTM, but I'm not gonna lie, I skipped most of the "obvious" changes and only briefly looked at unit tests :smile: 

Thanks for replacing ad-hoc mocking with MockEnv in many places! :tada:

atomic_reactor/plugin.py
```diff
@@ -135,22 +122,39 @@ def args_from_user_params(user_params: dict, /) -> dict:
...
-        """
-        constructor
+    def __init__(self, plugins_conf: List[Dict[str, Any]], *args, **kwargs) -> None:
+        """constructor
 
-        :param plugin_class_name: str, name of plugin class to filter (e.g. 'PreBuildPlugin')
-        :param plugins_conf: list of dicts, configuration for plugins
+        :param plugins_conf: list of dicts, configuration for plugins,
+            e.g. {'name': 'plugin_a', 'required': True}
+        :type plugins_conf: dict[str, any]
```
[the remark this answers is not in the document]
tkdchen: I think it depends on how we use the docstring, and from my point of view, the types in docstring are for different purpose than the type annotation. If we would like to keep docstring to describe arguments, for the completeness, it would be good to include the type as well.

atomic_reactor/inner.py
```diff
@@ -647,73 +642,15 @@ def build_docker_image(self) -> None:
...
-            raise
+            runner = PluginsRunner(self,
+                                   self.plugins_conf,
+                                   self.plugin_files,
+                                   self.keep_plugins_running,
+                                   plugins_results=self.data.plugins_results)
+            runner.run()
         finally:
-            # We need to make sure all exit plugins are executed
-            signal.signal(signal.SIGTERM, lambda *args: None)
```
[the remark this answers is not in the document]
tkdchen: @MartinBasti @rcerven Any idea about this signal handler? Is it ok to remove this line or not?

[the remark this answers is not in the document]
MartinBasti: I'd remove it, now we have finally task for such purpose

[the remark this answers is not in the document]
chmeliik: The finally task replaces the `finally` block, but do we still want to ignore SIGTERM when running the finally task or not?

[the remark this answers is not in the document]
rcerven: well we should still keep it for finally task, so all finally task plugins will actually finish, mainly removing reservation (now)

atomic_reactor/plugin.py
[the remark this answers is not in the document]
chmeliik: Seems reasonable. The type in the docstring does need to be fixed though (`list[dict[str, any]]`)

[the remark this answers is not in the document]
tkdchen: Oh, sorry, I'll update it later.

tkdchen:
PTAL.

atomic_reactor/inner.py
```diff
@@ -643,77 +638,26 @@ def throw_canceled_build_exception(self, *args, **kwargs):
...
-            exception_being_handled = True
-            raise
+            if not self.keep_plugins_running:
+                signal.signal(signal.SIGTERM, self.throw_canceled_build_exception)
+            runner = PluginsRunner(self,
+                                   self.plugins_conf,
+                                   self.plugin_files,
+                                   self.keep_plugins_running,
+                                   plugins_results=self.data.plugins_results)
+            runner.run()
```
chmeliik, line 659: Shouldn't we ignore SIGTERM explicitly, via `signal.signal(signal.SIGTERM, lambda *args: None)`?

chmeliik: Or maybe better, `signal.signal(signal.SIGTERM, signal.SIG_IGN)`

tkdchen: PTAL

chmeliik (approved):
LGTM :tada:

atomic_reactor/inner.py
```diff
@@ -499,22 +487,25 @@ def __init__(
...
         :param namespace: OpenShift namespace of the task
         :param pipeline_run_name: PipelineRun name to reference PipelineRun
-        :param plugins: the plugins to be executed in this workflow
+        :param plugins_conf: the plugins to be executed in this workflow
+        :type plugins_conf: list[dict[str, any]] or None
         :param user_params: user (and other) params that control various aspects of the build
         :param reactor_config_path: path to atomic-reactor configuration file
         :param plugin_files: load plugins also from these files
         :param client_version: osbs-client version used to render build json
+        :param keep_plugins_running: bool
```
hjmodi, line 496: can you add some description for this param as well?

atomic_reactor/plugin.py
```diff
@@ -134,26 +117,55 @@ def args_from_user_params(user_params: dict, /) -> dict:
         return {}
 
 
-class PluginsRunner(object):
+# Built-in plugins
+class PreBuildSleepPlugin(Plugin):
```
hjmodi, line 121: can we remove this or rename this? maybe `TestPlugin` or `DebugPlugin` will be more clear?

atomic_reactor/plugins/rpmqa.py
```diff
@@ -35,7 +35,7 @@ def __init__(self, workflow, ignore_autogenerated_gpg_keys=True):
         :param workflow: DockerBuildWorkflow instance
         """
         # call parent constructor
-        super(PostBuildRPMqaPlugin, self).__init__(workflow)
+        super().__init__(workflow)
```
hjmodi, line 38: since we're doing this I don't think we need the `__all__`  anymore.

atomic_reactor/tasks/sources.py
```diff
@@ -45,29 +45,22 @@ def prepare_workflow(self) -> inner.DockerBuildWorkflow:
...
-            {"name": "tag_and_push"},
-            {"name": "verify_media", "required": False},
-            {"name": "koji_import_source_container"},
-            {"name": "koji_tag_build"},
-        ],
-    )
+    plugins_conf = [
+        {"name": "fetch_sources"},
+        {"name": "bump_release"},
+        {"name": "source_container"},
```
hjmodi, line 51: Now that we don't have the build phase I think this should be `build_source_container` since just `source_container` donesn't make any sense

docs/api.md
```diff
@@ -480,10 +479,7 @@ This class defines a workflow for building images
 - plugins_errors
 - plugins_timestamps
 - postbuild_plugins_conf
-- postbuild_results
 - prebuild_plugins_conf
-- prebuild_results
-- prepub_results
```
hjmodi, line 486: should add `plugin_results`  here as well as `plugins_conf` and remove the rest of `*_plugins_conf` so that we don't have to clean up later

atomic_reactor/plugins/rpmqa.py
tkdchen: Why? I don't think the usage of `__all__` is relative to this change. `__all__` affects what are exported when use `from ... import *`.

docs/api.md
tkdchen: IMO, The section "##### DockerBuildWorkflow instance variables" should be rewritten at all, . IIRC, we are going to update various documentation for OSBS2, and this one should be part of the that.

atomic_reactor/tasks/sources.py
tkdchen: I'm not sure if it's necessary to rename it and even yes, rename are also involves renaming other things like the plugin class name. I'm going to keep it here in this pr. The rename can be done separately.

mkosiarc (approved):

atomic_reactor/tasks/sources.py
hjmodi: I would say this PR is a more appropriate place to do the rename rather than having PRs for non-functional changes but I'll leave that up to you.

hjmodi (approved):

PR #1903, opened 2022-04-25
allow `kojifile` components in `ImageBuildWorkflowData`

hjmodi:
* CLOUDBLD-9867


rcerven (approved):

tkdchen (approved):

atomic_reactor/schemas/workflow_data.json
```diff
@@ -44,7 +44,12 @@
       "patternProperties": {
         ".*": {
           "type": ["array"],
-          "items": {"$ref": "#/definitions/rpm_info"}
+          "items": {
+            "oneOf": [
+              {"$ref": "#/definitions/kojifile"},
```
chmeliik, line 49: How do kojifile components even get into `image_components`? The only 'kojifile' string that I found was in fetch_maven_artifacts, and that one didn't go into image_components

chmeliik: The only other culprit that comes to mind is `FlatpakBuilder.get_components()` in
https://github.com/containerbuildsystem/atomic-reactor/blob/fd88949d911b329bdf7ab2e5f2a9a2c22f578d0d/atomic_reactor/plugins/post_flatpak_create_oci.py#L77

Is that the one? Could that be explained in the commit message?

hjmodi: hmm, in that case, it would only happen for flatpak build but it does happen for normal builds as well so I'm not entirely sure. I highly doubt it's something from `koji_import`

chmeliik: Is it even supposed to be there? The compare_components plugin raises an exception for unsupported types in `image_components`, and the only supported type is "rpm"

https://github.com/containerbuildsystem/atomic-reactor/blob/308b5b9b4b450e69b69ded1a7a200db7e8da19e2/atomic_reactor/plugins/compare_components.py#L17-L18

https://github.com/containerbuildsystem/atomic-reactor/blob/308b5b9b4b450e69b69ded1a7a200db7e8da19e2/atomic_reactor/plugins/compare_components.py#L111-L119

(checked OSBS 1 code, seems to be the same)

hjmodi: something is adding it there afterwards

chmeliik: I think I found it, it's almost certainly a bug.

koji_import does `output['components'] += maven_metadata`
https://github.com/containerbuildsystem/atomic-reactor/blob/46e5db699cf070812b95fd1c48af56dd17fa35be/atomic_reactor/plugins/koji_import.py#L640-L645

`output['components'` is this
https://github.com/containerbuildsystem/atomic-reactor/blob/46e5db699cf070812b95fd1c48af56dd17fa35be/atomic_reactor/utils/koji.py#L480

which is
https://github.com/containerbuildsystem/atomic-reactor/blob/46e5db699cf070812b95fd1c48af56dd17fa35be/atomic_reactor/utils/koji.py#L364-L365

Why is this a problem? Because `output['components']` is a reference to the same list as `image_components[image_platform]`. So modifying `output['components']` also modifies `image_components`.

I love mutable objects

hjmodi: Very interesting find. Seems like that's exactly the reason.

hjmodi: tested and that's the issue. changed the PR to fix that.

chmeliik (approved):
LGTM :+1:

rcerven (approved):

PR #1904, opened 2022-04-26
Run tests in parallel by default in test container

tkdchen:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -29,3 +29,11 @@ commands =
         --non-interactive \
         --ignore-missing-imports \
         --package {posargs:"atomic_reactor"}
+
+[coverage:report]
+skip_covered = true
+sort = Cover
+
+[pytest]
+addopts = -n=2 -ra --color=auto --html=__pytest_reports/atomic-reactor-unit-tests.html --self-contained-html
+render_collapsed = True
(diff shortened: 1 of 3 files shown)


rcerven (approved):

tox.ini
```diff
@@ -29,3 +29,14 @@ commands =
...
+
+[paths]
+source = atomic_reactor
+
+[coverage:report]
+skip_covered = true
+sort = Cover
+
+[pytest]
+addopts = -n=2 -ra --color=auto --html=__pytest_reports/atomic-reactor-unit-tests.html --self-contained-html
```
chmeliik, line 41: Would `-n=auto` be better? (attempts to auto-detect number of cores and use that many processes)

tox.ini
```diff
@@ -29,3 +29,14 @@ commands =
         --non-interactive \
         --ignore-missing-imports \
         --package {posargs:"atomic_reactor"}
+
+[paths]
```
chmeliik, line 33: shouldn't this be `[coverage:paths]`?

tox.ini
tkdchen: I would say it depends. From what I observed locally, `-n=auto` takes longer time to run the tests and the major incremental part is the preparation for processes and waiting the execution starts. Different environment might have different effect resulted by the `auto`. `-n=2` should have a good enough effect for local and CI as a default value.

tox.ini
tkdchen: Removed. It's actually useless for this pr.

mkosiarc (approved):

chmeliik (approved):
Nice! :tada:

PR #1907, opened 2022-04-29
Fix operator manifest issues

hjmodi:
* CLOUDBLD-8489
* CLOUDBLD-9912


hjmodi:
Added fix for two bugs here since it's very related.

rcerven (approved):

atomic_reactor/plugins/pin_operator_digest.py
```diff
@@ -309,10 +309,13 @@ def _get_operator_manifest(self):
         if self.workflow.source.config.operator_manifests is None:
             raise RuntimeError("operator_manifests configuration missing in container.yaml")
         self.log.info("Looking for operator CSV files in %s", self.workflow.build_dir.path)
-        manifests_dir = self.workflow.source.config.operator_manifests["manifests_dir"]
+        manifests_dir = os.path.normpath(self.workflow.build_dir.any_platform.path
+                                         / self.workflow.source.config.operator_manifests[
+                                             "manifests_dir"])
+        if not manifests_dir.startswith(str(self.workflow.build_dir.any_platform.path)):
+            raise RuntimeError("manifests_dir points outside of cloned repository")
```
chmeliik, line 316: The previous check was slightly more robust in that it used `realpath` rather than `normpath` (realpath resolves symlinks, so you can't get around the check by having a `./manifests -> ../foo` symlink in the repo):
https://github.com/containerbuildsystem/atomic-reactor/blob/d7ec9ef5ba2d230aa70d12905f37952a1d2f1414/atomic_reactor/source.py#L104-L110

both the previous and current check have a small issue caused by using `str.startswith`, but that's probably fine (though we could use pathlib to make the check 100% correct)
```python
>>> "/foobar".startswith("/foo")
True
>>> Path("/foobar").relative_to(Path("/foo"))
...
ValueError: '/foobar' is not in the subpath of '/foo' OR one path is relative and the other is absolute.
```

hjmodi: that is a very good idea. Updated accordingly.

chmeliik (approved):
LGTM :+1:

PR #1911, opened 2022-05-10
Ignore annotation repositories properly in sendmail

tkdchen:
* CLOUDBLD-9874

annotations/repositories is included inside the return value from
tag_and_push plugin. In case tag_and_push has no chance to run due to
some previous plugin fails, there will be no repositories key
inside annotations. Once it happens, this commit simply ignores the
non-existing reporitories to render an empty Repositories line in the
mail body.


chmeliik (approved):
LGTM :+1:

rcerven (approved):

PR #1915, opened 2022-05-12
Fix GitSource, delete LazyGit

chmeliik:
CLOUDBLD-9971

### DIFF
--- a/atomic_reactor/source.py
+++ b/atomic_reactor/source.py
@@ -9,13 +9,14 @@ of the BSD license. See the LICENSE file for details.
 Code for getting source code to put inside container.
 """
 
+import functools
 import logging
 import copy
 import os
 import shutil
 import tempfile
 from textwrap import dedent
-from typing import Any, List
+from typing import Any, List, Callable, TypeVar
 import collections
 try:
     from urlparse import urlparse
@@ -87,6 +88,22 @@ class SourceConfig(object):
         return self.platforms['only']
 
 
+SourceT = TypeVar('SourceT', bound='Source')
+T = TypeVar('T')
+
+
+def path_must_exist(method: Callable[[SourceT], T]) -> Callable[[SourceT], T]:
+    """Make a Source method check that self.path exists before doing anything else."""
+
+    @functools.wraps(method)
+    def method_with_path_check(self: SourceT) -> T:
+        if not os.path.exists(self.path):
+            raise RuntimeError(f'Expected source path {self.path} does not exist')
+        return method(self)
+
+    return method_with_path_check
+
+
 class Source(object):
     def __init__(self, provider, uri, dockerfile_path=None, provider_params=None, workdir=None):
         self._config = None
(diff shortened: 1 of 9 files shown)


chmeliik:
PTAL, will remove temp commit after osbs-client PR is merged (and bump the ref instead)

Makefile
```diff
@@ -3,7 +3,7 @@ PYTHON_VERSION_VENV ?= python3.8
 .PHONY: venv
 venv:
 	virtualenv --python=${PYTHON_VERSION_VENV} venv
-	venv/bin/pip install --upgrade pip
+	venv/bin/pip install --upgrade 'pip<22'
```
hjmodi, line 6: why are we pinning pip version?

hjmodi (approved):

Makefile
chmeliik: This was just a temp change because pip-compile failed with pip 22. We probably don't have to pin it, I expect a new release of pip-compile will fix that

atomic_reactor/source.py
```diff
@@ -137,35 +142,22 @@ def get_vcs_info(self):
...
-    def get_vcs_info(self):
-        return VcsInfo(
-            vcs_type='git',
-            vcs_url=self.lg.git_url,
-            vcs_ref=self.lg.commit_id
-        )
-
-    def reset(self, git_reference):
-        self.lg.reset(git_reference)
+        return get_commit_id(self.path)
```
rcerven, line 147: is it intentional that this will raise error when the repo isn't cloned ?

atomic_reactor/source.py
```diff
@@ -137,35 +142,22 @@ def get_vcs_info(self):
...
+            commit = self.provider_params.get('git_commit', None)
+            branch = self.provider_params.get('git_branch', None)
+            depth = self.provider_params.get('git_commit_depth', None)
+            clone_git_repo(
+                self.uri, target_dir=self.path, commit=commit, branch=branch, depth=depth
+            )
+        return self.path
+
+    def get_vcs_info(self) -> VcsInfo:
+        return VcsInfo(vcs_type='git', vcs_url=self.uri, vcs_ref=self.commit_id)
```
rcerven, line 160: is it intentional that this will raise error when the repo isn't cloned ?

atomic_reactor/source.py
rcerven: you could also call self.get() and that would solve it

chmeliik: Looking into it some more, I think there are a few options:

1. make `Source.path` always call get()
2. for methods that expect the path to already exist (commit_id, get_vcs_info(), config, figure_out_build_file_path()), raise an error if the path doesn't exist
3. only document that many methods expect the path to exist

I'll go with 2 for now, LMK what you think

chmeliik:
Added check for path existence, needs a unit test though

chmeliik:
Replaced temp commit with actual osbs-client bump, should be ready now

mkosiarc (approved):

rcerven (approved):

PR #1918, opened 2022-05-13
allowlist for source containers using lookaside cache

rcerven:
* CLOUDBLD-9969

### DIFF
--- a/atomic_reactor/plugins/pre_fetch_sources.py
+++ b/atomic_reactor/plugins/pre_fetch_sources.py
@@ -8,10 +8,12 @@ of the BSD license. See the LICENSE file for details.
 import os
 import shutil
 import tempfile
+import re
 
 import koji
 import tarfile
 import yaml
+from typing import List, Dict, Any
 
 from atomic_reactor.constants import (PLUGIN_FETCH_SOURCES_KEY, PNC_SYSTEM_USER,
                                       REMOTE_SOURCE_JSON_FILENAME, REMOTE_SOURCE_TARBALL_FILENAME,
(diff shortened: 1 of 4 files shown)


hjmodi (approved):

atomic_reactor/schemas/config.json
```diff
@@ -695,6 +695,10 @@
               "description": "Url with denylist yaml file, which will be used to exclude sources from cachito ",
               "type": "string"
           },
+          "allowlist_cache": {
+              "description": "Url with allowlist yaml file, which allows usage of lookaside cache",
```
chmeliik, line 699: I would prefer something like "lookaside_cache_allowlist", the current name sounds like something that would specify if/where/how to cache some allowlist

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -194,12 +211,44 @@ def check_lookaside_cache_usage(self):
...
+            return
+
+        allowlist_cache_yaml = self._get_cache_allowlist()
+        should_raise = True
+
+        if allowlist_cache_yaml:
+            build_component = self.koji_build['package_name']
+            build_task = self.koji_build['extra'].get('container_koji_task_id')
+            task_info = self.session.getTaskInfo(build_task, request=True)
+            build_target = task_info['request'][1]
```
chmeliik, line 231: Is it expected there might not be a 'container_koji_task_id' in `koji_build['extra']`? Is calling `getTaskInfo(None, ...)` what we want to do in that case?

If it's not expected, could we do `koji_build['extra']['container_koji_task_id']` instead?

atomic_reactor/plugins/pre_fetch_sources.py
```diff
@@ -186,6 +187,22 @@ def set_koji_image_build_data(self):
         if not self.koji_build_nvr:
             self.koji_build_nvr = self.koji_build['nvr']
 
+    def _get_cache_allowlist(self):
```
chmeliik, line 190:
```suggestion
    def _get_cache_allowlist(self) -> List[Dict[str, Any]]:
```

atomic_reactor/schemas/config.json
rcerven: ok

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: right it has to be there

atomic_reactor/plugins/pre_fetch_sources.py
rcerven: sure

chmeliik (approved):
LGTM :+1:

Might be good to document the new config option later

rcerven:
> LGTM +1
> 
> Might be good to document the new config option later

it is in the schema, and I don't think cache is used upstream at all

chmeliik:
> > LGTM +1
> > Might be good to document the new config option later
> 
> it is in the schema, and I don't think cache is used upstream at all

Fair point, waiving docs then

rcerven:
rn updated: allowlist for source containers using lookaside cache

PR #1926, opened 2022-05-23
Set include_koji_repo correctly when plugin is aborted

mkosiarc:
This fixes bug that was introduced in
https://github.com/containerbuildsystem/atomic-reactor/pull/1764/

When plugin is aborted and no yum_repourls are present,
include_koji_repo should be set to True

CLOUDBLD-10020


chmeliik (commented):
Alternatively, with a little refactoring, how about getting rid of the `self.include_koji_repo` attribute entirely and returning `'include_koji_repos': not self.has_complete_repos` instead? I'm pretty sure that would be equivalent (and less error-prone)

chmeliik (approved):
LGTM :+1:

chmeliik (approved):

PR #1928, opened 2022-05-26
Fix start time > build time

hjmodi:

### DIFF
--- a/atomic_reactor/plugins/add_help.py
+++ b/atomic_reactor/plugins/add_help.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2017 Red Hat, Inc
+Copyright (c) 2017-2022 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
@@ -18,16 +18,17 @@ Example configuration:
 """
 
 import errno
-from datetime import datetime as dt
 from pathlib import Path
 from subprocess import check_output, CalledProcessError, STDOUT
 from typing import List
 
-from atomic_reactor import start_time as atomic_reactor_start_time
+
 from atomic_reactor.dirs import BuildDir
 from atomic_reactor.plugin import Plugin
 from atomic_reactor.metadata import annotation_map
 from osbs.utils import Labels
+
+from atomic_reactor.util import get_pipeline_run_start_time
 
 DEFAULT_HELP_FILENAME = "help.md"
 
@@ -89,14 +90,16 @@ class AddHelpPlugin(Plugin):
 
         help_path = build_dir.path / self.help_file
 
+        start_time = get_pipeline_run_start_time(self.workflow.osbs,
+                                                 self.workflow.pipeline_run_name)
+
         with open(help_path, 'r+') as help_file:
             lines = help_file.readlines()
 
             if not lines[0].startswith("% "):
                 lines.insert(0, "%% %s (1) Container Image Pages\n" % name)
                 lines.insert(1, "%% %s\n" % maintainer)
-                lines.insert(2, "%% %s\n" % dt.fromtimestamp(atomic_reactor_start_time)
-                             .strftime("%B %-d, %Y"))
+                lines.insert(2, "%% %s\n" % start_time.strftime("%B %-d, %Y"))
 
                 help_file.seek(0)
                 help_file.truncate()
--- a/atomic_reactor/plugins/add_labels_in_df.py
+++ b/atomic_reactor/plugins/add_labels_in_df.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2015 Red Hat, Inc
+Copyright (c) 2015-2022 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
@@ -54,17 +54,15 @@ with different values, the value from the first in the list will be used
 to set value for the missing ones.
 """
 
-import datetime
 import json
 from typing import Dict
 
 from osbs.utils import Labels
 
-from atomic_reactor import start_time as atomic_reactor_start_time
 from atomic_reactor.constants import INSPECT_CONFIG
 from atomic_reactor.dirs import BuildDir
 from atomic_reactor.plugin import Plugin
-from atomic_reactor.util import label_to_string, LabelFormatter
+from atomic_reactor.util import get_pipeline_run_start_time, label_to_string, LabelFormatter
 
 
 class AddLabelsPlugin(Plugin):
@@ -133,10 +131,10 @@ class AddLabelsPlugin(Plugin):
             raise RuntimeError("equal_labels have to be list")
 
     def generate_auto_labels(self, platform: str) -> Dict[str, str]:
-        # build date
-        dt = datetime.datetime.utcfromtimestamp(atomic_reactor_start_time)
+        start_time = get_pipeline_run_start_time(self.workflow.osbs,
+                                                 self.workflow.pipeline_run_name)
 
-        generated = {'build-date': dt.isoformat(), 'architecture': platform}
+        generated = {'build-date': start_time.isoformat(), 'architecture': platform}
 
         # VCS info
         vcs = self.workflow.source.get_vcs_info()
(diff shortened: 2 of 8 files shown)


atomic_reactor/plugins/add_labels_in_df.py
```diff
@@ -133,8 +133,8 @@ def __init__(self, workflow, labels=None, dont_overwrite=None,
             raise RuntimeError("equal_labels have to be list")
 
     def generate_auto_labels(self, platform: str) -> Dict[str, str]:
-        # build date
-        dt = datetime.datetime.utcfromtimestamp(atomic_reactor_start_time)
+        # build date, using the current time here to be > start time of koji build
+        dt = datetime.datetime.utcfromtimestamp(time.time())
```
hjmodi, line 137: We definitely cannot use the start time of prebuild task as the start time here as it has to be > start time of koji build but do let me know if there is an alternate time we can use that captures build-date more accurately.

hjmodi:
Coverage is decreasing due to a decrease in the number of lines.

atomic_reactor/inner.py
```diff
@@ -291,6 +291,8 @@ class ImageBuildWorkflowData(ISerializer):
     plugins_errors: Dict[str, str] = field(default_factory=dict)
     build_canceled: bool = False
     plugin_failed: bool = False
+    # build start time
+    start_time: float = field(default=time.time())
```
chmeliik, line 295: Does this mean the start_time of the build will be the time when `inner.py` was first imported? That seems questionable, maybe we could still import the start_time from `atomic_reactor` and set that as the default?

atomic_reactor/plugins/add_labels_in_df.py
hjmodi: I tried that. Somehow, it doesn't work. This `start_time` doesn't seem to have any effect on the start time of the koji build. It's always set to the time `bump_release` reserved the build.

atomic_reactor/inner.py
hjmodi: Yes, I also looked into setting the start time to pipeline start time but we don't get any pipeline metadata in our env.

atomic_reactor/plugins/add_labels_in_df.py
chmeliik: Oh, I see :thinking: I thought that didn't explain why it worked in OSBS 1, but it actually does. In OBS 1, `bump_release` runs in orchestrator prebuild, and `add_labels` runs in worker. That means the start_time for add_labels is after the time when bump_release ran

If Koji ignores the start_time and we don't use it in add_labels, is it still useful for anything? Maybe in case the bump_release plugin doesn't do the reservation?

chmeliik: Oh I see, it's also used in add_help. But maybe it should be the same time as the label?

atomic_reactor/inner.py
rcerven: why not add it to DockerBuildWorkflow ? which is really used at the start

hjmodi: I don't think it will persist through task runs.

hjmodi: that would require passing it from osbs-client. I don't think that will be any better and adding someway in here to somehow pass `start_time` doesn't seem like a good idea.

rcerven: I didn't mean to add it to workflow, but just use start_time from prun, where ever you need directly

atomic_reactor/plugins/add_labels_in_df.py
hjmodi: hmm, I agree. If that's the case, we can remove the `start_time` and `end_time` from build metadata(since it doesn't take effect anyway) then we don't need it to be in `workflow.data` since we just need it in the prebuild task

rcerven: we want to keep metadata in koji build, those times aren't used for koji build 'started' , 'completed' which are irrelevant anyway

when user downloads metadata json, those times have to be there

and you can use real pipeline start time , as its relation to koji build itself is irrelevant

hjmodi: ok. For the build-date label, should we use start time as well or start time of koji build?

atomic_reactor/inner.py
hjmodi: updated to get start time from pipeline run

atomic_reactor/plugins/add_help.py
```diff
@@ -89,13 +90,19 @@ def render_help_file(self, build_dir: BuildDir) -> List[Path]:
...
+        start_ts = dt.strptime(start_time, '%Y-%m-%dT%H:%M:%SZ').timestamp()
+
         with open(help_path, 'r+') as help_file:
             lines = help_file.readlines()
 
             if not lines[0].startswith("% "):
                 lines.insert(0, "%% %s (1) Container Image Pages\n" % name)
                 lines.insert(1, "%% %s\n" % maintainer)
-                lines.insert(2, "%% %s\n" % dt.fromtimestamp(atomic_reactor_start_time)
+                lines.insert(2, "%% %s\n" % dt.fromtimestamp(start_ts)
```
chmeliik, line 105: this part converts from datetime to timestamp and then back again

atomic_reactor/plugins/add_labels_in_df.py
chmeliik: Same time we use in the other plugins IMO

hjmodi: hmm, in that case we'll have to change the test case

tkdchen, line 137: Why not use the pipeline run start time as the build date?

atomic_reactor/plugins/add_help.py
```diff
@@ -89,13 +90,19 @@ def render_help_file(self, build_dir: BuildDir) -> List[Path]:
 
         help_path = build_dir.path / self.help_file
 
+        osbs = get_openshift_session(self.workflow.conf,
+                                     self.workflow.namespace)
+        build_info = osbs.get_build(self.workflow.pipeline_run_name)
+        start_time = build_info['status']['startTime']
+        start_ts = dt.strptime(start_time, '%Y-%m-%dT%H:%M:%SZ').timestamp()
```
tkdchen, line 97: Let's extract these lines into a method `get_pipeline_run_start_time()`.

atomic_reactor/plugins/add_help.py
hjmodi: removed reconversion

atomic_reactor/plugins/add_labels_in_df.py
hjmodi: changed it to use start time

atomic_reactor/plugins/add_help.py
hjmodi: extracted out

atomic_reactor/util.py
```diff
@@ -370,6 +371,19 @@ def get_image_upload_filename(image_type, image_id, platform):
                            platform=platform, ext=ext)
 
 
+def get_pipeline_run_start_time(conf: 'atomic_reactor.config.Configuration' = None,
```
hjmodi, line 374: This method can be refactored to be better after https://github.com/containerbuildsystem/atomic-reactor/pull/1932 is merged

atomic_reactor/plugins/add_labels_in_df.py
```diff
@@ -133,8 +132,12 @@ def __init__(self, workflow, labels=None, dont_overwrite=None,
             raise RuntimeError("equal_labels have to be list")
 
     def generate_auto_labels(self, platform: str) -> Dict[str, str]:
+        start_time = get_pipeline_run_start_time(self.workflow.conf,
+                                                 self.workflow.namespace,
+                                                 self.workflow.pipeline_run_name)
+        start_ts = start_time.timestamp()
         # build date
-        dt = datetime.datetime.utcfromtimestamp(atomic_reactor_start_time)
+        dt = datetime.utcfromtimestamp(start_ts)
```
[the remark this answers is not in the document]
hjmodi: updated. The pipeline timestamp is UTC only so you're right we don't need to redo the conversation.

chmeliik (approved):
LGTM :+1:

ben-alkov (approved):

PR #1932, opened 2022-05-30
Fix build failure and cancellation handling

chmeliik:
CLOUDBLD-10117

Instead of relying on the plugin_failed property, check the status of
the Tekton pipeline to find out if any task failed.

Get rid of the plugin_failed property.

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -19,7 +19,7 @@ import time
 import re
 from dataclasses import dataclass, field, fields
 from textwrap import dedent
-from typing import Any, Callable, Dict, Final, List, Optional, Union
+from typing import Any, Callable, Dict, Final, List, Optional, Union, Tuple
 
 from dockerfile_parse import DockerfileParser
 
@@ -631,12 +631,18 @@ class DockerBuildWorkflow(object):
     def osbs(self) -> OSBS:
         return get_openshift_session(self.conf, self.namespace)
 
-    @property
-    def build_process_failed(self):
+    def check_build_outcome(self) -> Tuple[bool, bool]:
+        """Did the build process fail? Was the build cancelled?
+
+        :return: Tuple of (failed, cancelled). Note that a cancelled build also counts as failed.
         """
-        Has any aspect of the build process failed?
-        """
-        return self.data.plugin_failed
+        cancelled = self.osbs.build_has_any_cancelled_tasks(self.pipeline_run_name)
+        failed = (
+            cancelled  # cancelled counts as failed
+            or self.osbs.build_has_any_failed_tasks(self.pipeline_run_name)  # prev. task failed
+            or bool(self.data.plugins_errors)  # this task failed
+        )
+        return failed, cancelled
 
     def throw_canceled_build_exception(self, *args, **kwargs):
         self.data.build_canceled = True
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -635,7 +635,10 @@ class DockerBuildWorkflow(object):
 
         :return: Tuple of (failed, cancelled). Note that a cancelled build also counts as failed.
         """
-        cancelled = self.osbs.build_has_any_cancelled_tasks(self.pipeline_run_name)
+        cancelled = (
+            self.osbs.build_has_any_cancelled_tasks(self.pipeline_run_name)  # prev. task cancelled
+            or self.data.build_canceled  # this task cancelled
+        )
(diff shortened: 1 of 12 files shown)


atomic_reactor/inner.py
```diff
@@ -626,12 +626,16 @@ def parent_images_to_str(self):
...
+    def osbs(self) -> OSBS:
+        return get_openshift_session(self.conf, self.namespace)
+
+    @functools.cached_property
+    def build_process_failed(self) -> bool:
         """
         Has any aspect of the build process failed?
         """
-        return self.data.plugin_failed
+        return self.osbs.build_has_any_failed_tasks(self.pipeline_run_name)
```
chmeliik, line 638: And another question, mainly for @tkdchen:

Is it OK that this will return False when the build is cancelled? It kinda looks like the cancel_reservation plugin would expect this to be True in that case

https://github.com/containerbuildsystem/atomic-reactor/blob/e1cb2b82527e5097009c4705474a5bc963f5826c/atomic_reactor/plugins/cancel_build_reservation.py#L52-L59

tkdchen: Why is False returned when build is cancelled?

atomic_reactor/inner.py
```diff
@@ -626,12 +626,16 @@ def parent_images_to_str(self):
     def image(self):
         return self.user_params['image_tag']
 
-    @property
-    def build_process_failed(self):
+    @functools.cached_property
+    def osbs(self) -> OSBS:
```
tkdchen, line 630: Nice for making this property.

atomic_reactor/inner.py
chmeliik: IMO 'failed' and 'cancelled' are different outcomes. A cancelled build is not a failed build. Should this return True for both failed and cancelled builds?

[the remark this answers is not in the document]
chmeliik: Did it that way to be more consistent with OSBS 1

chmeliik:
PTAL, should be ready (just not sure if `build_process_failed` should be True or False for cancelled builds)

atomic_reactor/inner.py
hjmodi: +1

atomic_reactor/inner.py
rcerven: we want to always call CGRefundBuild with correct state

failed build should call it with 'FAILED', canceled with 'CANCELED'

rcerven:
> PTAL, should be ready (just not sure if `build_process_failed` should be True or False for cancelled builds)

build_process_failed should return True even when build was cancelled but

IMHO, it should actually return tuple(failed or cancelled or plugin_errors, cancelled)

atomic_reactor/inner.py
```diff
@@ -626,12 +626,24 @@ def parent_images_to_str(self):
...
         Has any aspect of the build process failed?
         """
-        return self.data.plugin_failed
+        return (
+            self._any_previous_task_failed() or
+            # current task failed
+            bool(self.data.plugins_errors)
+        )
+
+    @functools.lru_cache
```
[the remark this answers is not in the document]
chmeliik: Ack, that's a fair point. The check shouldn't be _that_ slow anyway

atomic_reactor/inner.py
chmeliik: Updated with suggestion about returning (failed, cancelled) from the same method. Cancelled counts as failed (the method can't return (False, True)

atomic_reactor/inner.py
chmeliik: Thanks :slightly_smiling_face:

chmeliik:
Added handling for `data.build_canceled`, let me know if anyone is against that

mkosiarc:
Code looks good to me, I noticed that you set the osbs property in [Fix build failure and cancellation check](https://github.com/containerbuildsystem/atomic-reactor/pull/1932/commits/4467cd9aab777c4801336051412ce8752efb5c7f) commit, but associated changes are in [Re-use osbs property, don't create new instance](https://github.com/containerbuildsystem/atomic-reactor/pull/1932/commits/0baff4bccb37cb1c11179525f77b6853b1ff00d1) commit. Maybe having that in the same commit would be nice, but that is not that important.

chmeliik:
Moved the workflow.osbs property and the replacing of existing get_openshift_session() calls into separate commit

hjmodi (approved):

PR #1938, opened 2022-06-07
Output original Dockerfile when build fails

tkdchen:
* CLOUDBLD-10133

### DIFF
--- a/tekton/pipelines/binary-container.yaml
+++ b/tekton/pipelines/binary-container.yaml
@@ -287,3 +287,20 @@ spec:
           value: "$(context.pipelineRun.name)"
         - name: user-params
           value: '$(params.user-params)'
+    - name: binary-container-output-dockerfile
+      when:
+        - input: "$(tasks.status)"
+          operator: in
+          values:
+            - Failed
+      taskRef:
+        name: binary-container-output-dockerfile-0-1
+      workspaces:
+        - name: ws-build-dir
+          workspace: ws-container
+          subPath: build-dir
+      params:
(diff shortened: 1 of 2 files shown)


chmeliik:
> Unfortunately, the output of exit tasks will not be in the logs disappointed (at least not until CLOUDBLD-9857 is done, if it is even possible)

Ignore me, it won't be in the Brew **build** logs, but it will be in the Brew **task** logs (those are uploaded by koji-containerbuild) and that's good enough (failed tasks don't even produce builds)

tekton/pipelines/binary-container.yaml
```diff
@@ -287,3 +287,20 @@ spec:
           value: "$(context.pipelineRun.name)"
         - name: user-params
           value: '$(params.user-params)'
+    - name: binary-container-output-dockerfile
+      when:
+        - input: "$(tasks.status)"
+          operator: in
+          values:
+            - Failed
```
chmeliik, line 295: I'd say this doesn't need to be conditional, the original Dockerfile might be useful for successful builds too

(and if it's not conditional, I wonder if it still makes sense to have it as a separate task)

rcerven: yeah it should be consistent, for both successful & failed builds

and I'd just add it to some exit task, this doesn't even deserve it's own exit plugin (maybe store_metadata)

but also having it as separate task doesn't make sense

tkdchen: Have opened another PR #1942. This one will be closed.

PR #1939, opened 2022-06-08
Move sigterm handling to task level

mkosiarc:
CLOUDBLD-10115

### DIFF
--- a/atomic_reactor/plugins/add_filesystem.py
+++ b/atomic_reactor/plugins/add_filesystem.py
@@ -19,7 +19,7 @@ from atomic_reactor.dirs import BuildDir
 from atomic_reactor.constants import (DEFAULT_DOWNLOAD_BLOCK_SIZE, PLUGIN_ADD_FILESYSTEM_KEY,
                                       PLUGIN_RESOLVE_COMPOSES_KEY)
 from atomic_reactor.config import get_koji_session
-from atomic_reactor.plugin import Plugin, BuildCanceledException
+from atomic_reactor.plugin import Plugin, TaskCanceledException
 from atomic_reactor.utils.koji import TaskWatcher, stream_task_output
 from atomic_reactor.utils.yum import YumRepo
 from atomic_reactor.util import get_platforms, base_image_is_custom, map_to_user_params
@@ -286,8 +286,8 @@ class AddFilesystemPlugin(Plugin):
         task = TaskWatcher(self.session, task_id, self.poll_interval)
         try:
             task.wait()
-        except BuildCanceledException:
-            self.log.info("Build was canceled, canceling task %s", task_id)
+        except TaskCanceledException:
+            self.log.info("Tekton task was canceled, canceling brew task %s", task_id)
             try:
                 self.session.cancelTask(task_id)
                 self.log.info('task %s canceled', task_id)
--- a/atomic_reactor/tasks/common.py
+++ b/atomic_reactor/tasks/common.py
@@ -7,15 +7,18 @@ of the BSD license. See the LICENSE file for details.
 """
 
 import abc
+import signal
 from dataclasses import dataclass, fields
 from pathlib import Path
 from typing import Dict, Any, ClassVar, Generic, TypeVar
+from functools import cached_property
 
 from atomic_reactor import config
 from atomic_reactor import dirs
 from atomic_reactor import inner
 from atomic_reactor import source
 from atomic_reactor import util
+from atomic_reactor.plugin import TaskCanceledException
 
 
 @dataclass(frozen=True)
@@ -87,6 +90,8 @@ ParamsT = TypeVar("ParamsT", bound=TaskParams)
 class Task(abc.ABC, Generic[ParamsT]):
     """Task; the main execution unit in atomic-reactor."""
 
+    ignore_sigterm: ClassVar[bool] = False
+
     def __init__(self, params: ParamsT):
         """Initialize a Task."""
         self._params = params
@@ -101,9 +106,29 @@ class Task(abc.ABC, Generic[ParamsT]):
     def get_context_dir(self) -> dirs.ContextDir:
         return dirs.ContextDir(Path(self._params.context_dir))
 
-    def load_workflow_data(self) -> inner.ImageBuildWorkflowData:
+    @cached_property
+    def workflow_data(self) -> inner.ImageBuildWorkflowData:
         context_dir = self.get_context_dir()
         return inner.ImageBuildWorkflowData.load_from_dir(context_dir)
 
     def load_config(self) -> config.Configuration:
         return config.Configuration(self._params.config_file)
+
+    def throw_task_canceled_exception(self, *args, **kwargs):
+        self.workflow_data.task_canceled = True
+        raise TaskCanceledException("Tekton task was canceled")
+
+    def run(self, *args, **kwargs):
+        try:
+            if self.ignore_sigterm:
+                signal.signal(signal.SIGTERM, signal.SIG_IGN)
+            else:
+                signal.signal(signal.SIGTERM, self.throw_task_canceled_exception)
+
+            self.execute(*args, **kwargs)
+
+        finally:
+            signal.signal(signal.SIGTERM, signal.SIG_DFL)
+            # For whatever the reason a task fails, always write the workflow
+            # data into the data file.
+            self.workflow_data.save(self.get_context_dir())
--- a/tests/tasks/test_common.py
+++ b/tests/tasks/test_common.py
@@ -11,12 +11,14 @@ from typing import Optional, ClassVar
 
 from flexmock import flexmock
 import pytest
+import signal
 
 from atomic_reactor import dirs
 from atomic_reactor import inner
 from atomic_reactor import source
 from atomic_reactor import util
 from atomic_reactor.tasks import common
+
 
 TASK_ARGS = {
     "build_dir": "/build",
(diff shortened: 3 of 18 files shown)


atomic_reactor/tasks/common.py
```diff
@@ -90,6 +93,8 @@ class Task(abc.ABC, Generic[ParamsT]):
     def __init__(self, params: ParamsT):
         """Initialize a Task."""
         self._params = params
+        self.data = self.load_workflow_data()
```
chmeliik, line 96: Possibly a pro-tip for the future: fixing unit tests might have been less painful if this was a cached property. Currently, you need to mock context data just to initialize any Task. With a property, you'd only have to mock in the tests that actually use the data, which would probably be more or less the same tests that already do the mocking.

(No need to change it now)

atomic_reactor/tasks/common.py
```diff
@@ -107,3 +112,31 @@ def load_workflow_data(self) -> inner.ImageBuildWorkflowData:
...
         return config.Configuration(self._params.config_file)
+
+    def throw_canceled_build_exception(self):
+        self.data.build_canceled = True
+        raise BuildCanceledException("Build was canceled")
+
+    @classmethod
+    def run(cls, init_build_dirs: Optional[bool] = False):
+        try:
+            cls.fs_watcher.start()
```
chmeliik, line 123: I don't think this should be a class method, fs_watcher is an instance attribute (and execute() and others as well)

atomic_reactor/tasks/common.py
```diff
@@ -107,3 +112,31 @@ def load_workflow_data(self) -> inner.ImageBuildWorkflowData:
...
+
+            # non-plugin based tasks don't have the attribute
+            if hasattr(cls, 'keep_plugins_running') and cls.keep_plugins_running:
```
chmeliik, line 126: Maybe we could move this to Task level and rename? I can only think of silly names though, like `immortal` 😄

atomic_reactor/tasks/common.py
mkosiarc: hmm, would that not have detrimental effect when we want to actually update the property?

atomic_reactor/tasks/common.py
chmeliik: Well we're using it here :smile: 

I imagine that if we reworked the exit task to not be plugin based, we would still want to ignore SIGTERM in it

atomic_reactor/tasks/common.py
mkosiarc: I always welcome pro-tips, updated.

atomic_reactor/tasks/common.py
mkosiarc: so what about `keep_task_running` or `cancellable_task`

chmeliik: `ignore_sigterm` would be a good name if it's just for ignoring SIGTERM

Come to think of it, maybe it *should* be just for that, and `keep_plugins_running` should be a separate thing that decides only if plugin errors will be ignored

mkosiarc: added `ignore_sigterm` class variable, since it should do different thing than `keep_plugins_running`

atomic_reactor/tasks/common.py
```diff
@@ -87,9 +91,12 @@ def _drop_known_unset_args(cls, args: dict) -> dict:
 class Task(abc.ABC, Generic[ParamsT]):
     """Task; the main execution unit in atomic-reactor."""
 
+    ignore_sigterm: ClassVar[bool] = False
+
     def __init__(self, params: ParamsT):
         """Initialize a Task."""
         self._params = params
+        self.fs_watcher = FSWatcher()
```
mkosiarc, line 99: Hm, maybe not all tasks should have a FSWatcher() because the data from it are used only in store_metadata plugin to export it in annotations.

mkosiarc: @chmeliik @tkdchen what do you think?

chmeliik: Hmm, not gonna lie, I don't really know what `FSWatcher` is good for :smile: 

I guess it could be useful if we wanted to figure out how much disk space our builds are actually using, and for that, we would want to keep track of the usage across all tasks. Which would mean saving the results in context data somehow, and that sounds kinda complicated, so maybe it should be a separate ticket.

There might also be other options - e.g. just _log_ the usage after each task, which would make it easier to query in Splunk anyway. Or get rid of it entirely and rely on some OpenShift monitoring mechanism to get usage stats instead.

TL;DR, I think FSWatcher doesn't really need to do anything sensible at the moment and we should figure it out later

hjmodi (approved):

chmeliik (commented):
LGTM, just a minor nitpick for a log message

And another fairly important change here is saving context data in every Task, maybe that should be in the commit message as well

atomic_reactor/tasks/common.py
```diff
@@ -107,3 +112,26 @@ def load_workflow_data(self) -> inner.ImageBuildWorkflowData:
...
+
+        finally:
+            signal.signal(signal.SIGTERM, signal.SIG_DFL)
+            # For whatever the reason a task fails, always write the workflow
+            # data into the data file.
+            self.data.save(self.get_context_dir())
+
+    @cached_property
+    def data(self):
+        return self.load_workflow_data()
```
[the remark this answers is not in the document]
mkosiarc: well, it is already done, so..:D But yes, this needed to be done so the context could be saved at the last level, see line `self.data.save(self.get_context_dir())`

atomic_reactor/tasks/common.py
```diff
@@ -107,3 +112,26 @@ def load_workflow_data(self) -> inner.ImageBuildWorkflowData:
 
     def load_config(self) -> config.Configuration:
         return config.Configuration(self._params.config_file)
+
+    def throw_task_canceled_exception(self, *args, **kwargs):
+        self.data.task_canceled = True
+        raise TaskCanceledException("Task was canceled")
```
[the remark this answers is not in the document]
mkosiarc: It means a-r task, which could be any task that inherits from class Task. that is why keeping the name `BuildCanceledException` does not seem right to me.

atomic_reactor/plugins/add_filesystem.py
```diff
@@ -286,8 +286,8 @@ def run_image_task(self, image_build_conf):
         task = TaskWatcher(self.session, task_id, self.poll_interval)
         try:
             task.wait()
-        except BuildCanceledException:
-            self.log.info("Build was canceled, canceling task %s", task_id)
+        except TaskCanceledException:
+            self.log.info("Task was canceled, canceling task %s", task_id)
```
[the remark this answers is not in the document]
mkosiarc: What about "Prebuild task was cancelled, cancelling brew task"? I am okay with "Build was cancelled, cancelling brew task" as well.

atomic_reactor/tasks/common.py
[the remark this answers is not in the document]
chmeliik: Maybe we can just get rid of the `load_workflow_data()` method to avoid having one "wrong" way to load data? (@tkdchen is that what you meant?)

atomic_reactor/plugins/add_filesystem.py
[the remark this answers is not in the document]
chmeliik: "Prebuild task" works for me :+1:

atomic_reactor/tasks/common.py
```diff
@@ -107,3 +112,26 @@ def load_workflow_data(self) -> inner.ImageBuildWorkflowData:
...
+
+    def run(self, *args, **kwargs):
```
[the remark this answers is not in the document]
chmeliik: Hmm, maybe `init_build_dirs` never should have been a parameter in the first place, but that's probably not worth fixing at the moment

I'm fine with *args **kwargs

atomic_reactor/plugins/add_filesystem.py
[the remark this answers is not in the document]
rcerven: tekton task was canceled, canceling koji task

atomic_reactor/tasks/common.py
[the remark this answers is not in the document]
rcerven: 'tekton task was cancelled'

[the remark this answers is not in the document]
mkosiarc: updated

atomic_reactor/plugins/add_filesystem.py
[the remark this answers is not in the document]
mkosiarc: updated

atomic_reactor/tasks/common.py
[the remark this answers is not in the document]
tkdchen: I thought if we add another another layer (a cached property) to wrap `load_workflow_data`, why not just convert `load_workflow_data` to a cached property (whatever the name is).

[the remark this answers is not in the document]
mkosiarc: updated to have only one cached property handling this

tests/tasks/test_common.py
```diff
@@ -169,10 +168,47 @@ def test_load_workflow_data(self, tmp_path):
...
+
+    def workflow_data(self, params):
         expect_data = inner.ImageBuildWorkflowData()
         (flexmock(inner.ImageBuildWorkflowData)
          .should_receive("load_from_dir")
          .with_args(dirs.ContextDir)
          .and_return(expect_data))
 
-        assert self.ConcreteTask(params).load_workflow_data() == expect_data
+        assert self.ConcreteTask(params).workflow_data() == expect_data
```
chmeliik, line 182: If this is meant to be a unit test, I think it needs to start with `test_`

chmeliik (approved):
LGTM :+1:

PR #1944, opened 2022-06-17
New registries_cfg_path config

tkdchen:
* CLOUDBLD-9346

The auth.cfg_path configs under registry, pull_registries and
source_registry are moved outside and are replaced by the top level
config registries_cfg_path.

To keep the ability of indicating whether a registry requires
authentication, every kind of registry config has auth config in boolean
type. The final dockercfg_path is read from the registries_cfg_path only
when auth is True.

### DIFF
--- a/atomic_reactor/config.py
+++ b/atomic_reactor/config.py
@@ -7,6 +7,8 @@ of the BSD license. See the LICENSE file for details.
 """
 
 from copy import deepcopy
+from typing import Optional
+
 from atomic_reactor.utils.cachito import CachitoAPI
 from atomic_reactor.constants import REACTOR_CONFIG_ENV_NAME
 from atomic_reactor.util import (
@@ -159,6 +161,7 @@ class ReactorConfigKeys(object):
     CONTENT_VERSIONS_KEY = 'content_versions'
     REGISTRIES_ORGANIZATION_KEY = 'registries_organization'
     REGISTRY_KEY = 'registry'
+    REGISTRIES_CFG_PATH_KEY = 'registries_cfg_path'
     REMOTE_HOSTS_KEY = 'remote_hosts'
     YUM_PROXY_KEY = 'yum_proxy'
     SOURCE_REGISTRY_KEY = 'source_registry'
@@ -387,6 +390,10 @@ class Configuration(object):
         return self._get_value(ReactorConfigKeys.CONTENT_VERSIONS_KEY, fallback=[])
 
     @property
+    def yum_proxy(self):
+        return self._get_value(ReactorConfigKeys.YUM_PROXY_KEY, fallback=None)
+
+    @property
     def registries_organization(self):
         return self._get_value(ReactorConfigKeys.REGISTRIES_ORGANIZATION_KEY, fallback=None)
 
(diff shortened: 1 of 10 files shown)


atomic_reactor/config.py
```diff
@@ -402,27 +409,34 @@ def registry(self):
...
+        info = {
             'uri': RegistryURI(registry['url']),
             'insecure': registry.get('insecure', False),
-            'dockercfg_path': registry.get('auth', {}).get('cfg_path', None)
+            'dockercfg_path': None,
         }
+        if registry.get('auth', False):
+            if (cfg_path := self.registries_cfg_path) is None:
+                raise ValueError("Missing registries_cfg_path for registry authentication.")
+            info['dockercfg_path'] = cfg_path
```
chmeliik, line 433: What do you think about removing the `auth` param? If a registry is present in the dockercfg file, it uses authentication, otherwise it doesn't. Having another param could be confusing

tests/test_config.py
```diff
@@ -600,20 +666,21 @@ def test_get_methods(self, parse_from, method, tmpdir, caplog, monkeypatch):
         real_attr = getattr(conf, method)
 
         output = real_attr
+        reactor_config_map = yaml.safe_load(REACTOR_CONFIG_MAP)
 
         if method == 'registry':
             # map `registry` to `registries` in config so that neither use nor devs are confused
-            expected = yaml.safe_load(REACTOR_CONFIG_MAP)['registry']
+            expected = reactor_config_map['registry']
```
chmeliik, line 673: Just noticed the outdated comment, it can be removed

tests/test_config.py
```diff
@@ -53,8 +53,30 @@ class TestConfiguration(object):
...
+              url: https://container-registry.example.com/v2
+            registries_cfg_path: /var/run/secrets/atomic-reactor/v2-registry-dockercfg
+            """),
+            None,
+        ),
+        (
+            dedent("""\
+            registry:
+              url: https://container-registry.example.com/v2
+            registries_cfg_path: /var/run/secrets/atomic-reactor/v2-registry-dockercfg
```
chmeliik, line 79: Aren't both these cases duplicates now?

tests/test_util.py
```diff
@@ -415,42 +411,15 @@ def request_callback(request, all_headers=True):
...
+         ],
+         'registries_cfg_path': '/not/important',
      },
      {
          'uri': "https://some_registry.io",
          'insecure': False,
          'dockercfg_path': '/not/important'
      }),
-    # Have both source_registry and pull_registries, matches pull_registries
-    ('some_registry.io',
```
chmeliik, line 430: Why is this case not needed anymore?

chmeliik (commented):
LGTM, just a question about some test cases

tests/test_util.py
tkdchen: Good catch. Fixed both.

I think the reason should be I didn't have a good understand of `RegistrySession.create_from_config`.

tests/test_config.py
chmeliik: I meant both of these should be removed, the same test case already exists a few lines above this preview

atomic_reactor/schemas/config.json
```diff
@@ -456,23 +456,38 @@
         }
     },
     "registries_organization": {"$ref": "#/definitions/organization"},
+    "registries_cfg_path": {
```
hjmodi, line 459: now that we have only one registry I guess `registry_cgf_path` will be a better name.

hjmodi (approved):

hjmodi:
LGTM, just a minor comment

atomic_reactor/schemas/config.json
```diff
@@ -456,23 +456,38 @@
...
+      "description": "Path to directory containing .dockercfg for registry auth",
```
rcerven, line 460: registry => registries

rcerven:
base code LGTM

atomic_reactor/schemas/config.json
tkdchen: There is a discussion in the linked jira issue for this name.

It's not only for one registry. Multiple registries can be configured within that configuration file.

chmeliik (approved):
LGTM :+1:

PR #1955, opened 2022-06-29
check_and_set_platforms will always check disabled remote hosts,

rcerven:
and fail build if target or user platforms need it,
we still want to skip failure if remote hosts for platform aren't defined
at all

* CLOUDBLD-10421

### DIFF
--- a/atomic_reactor/plugins/check_and_set_platforms.py
+++ b/atomic_reactor/plugins/check_and_set_platforms.py
@@ -58,6 +58,7 @@ class CheckAndSetPlatformsPlugin(Plugin):
         run the plugin
         """
         user_platforms: Optional[List[str]] = self.workflow.user_params.get("platforms")
+        user_platforms_override = False
 
         if self.koji_target:
             koji_session = get_koji_session(self.workflow.conf)
@@ -68,21 +69,17 @@ class CheckAndSetPlatformsPlugin(Plugin):
             koji_build_conf = koji_session.getBuildConfig(build_tag, event=event_id)
             koji_platforms = koji_build_conf['arches']
             if not koji_platforms:
-                self.log.info("No platforms found in koji target")
-                return None
+                raise RuntimeError("No platforms found in koji target")
             platforms = koji_platforms.split()
             self.log.info("Koji platforms are %s", sorted(platforms))
 
             if is_scratch_build(self.workflow) or is_isolated_build(self.workflow):
                 override_platforms = set(user_platforms or [])
                 if override_platforms and override_platforms != set(platforms):
-                    sorted_platforms = sorted(override_platforms)
-                    self.log.info("Received user specified platforms %s", sorted_platforms)
+                    user_platforms_override = True
+                    platforms = sorted(override_platforms)
+                    self.log.info("Received user specified platforms %s", platforms)
                     self.log.info("Using them instead of koji platforms")
-                    # platforms from user params do not match platforms from koji target
-                    # that almost certainly means they were overridden and should be used
-                    self.workflow.build_dir.init_build_dirs(sorted_platforms, self.workflow.source)
-                    return sorted_platforms
         else:
             platforms = user_platforms
             self.log.info(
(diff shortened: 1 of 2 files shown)


atomic_reactor/plugins/check_and_set_platforms.py
```diff
@@ -115,11 +113,14 @@ def has_enabled_hosts(platform: str) -> bool:
                   ' {}'.format(defined_but_disabled)
             raise RuntimeError(msg)
 
-        final_platforms = self._limit_platforms(enabled_platforms)
-        self.log.info("platforms in limits : %s", final_platforms)
-        if not final_platforms:
-            self.log.error("platforms in limits are empty")
-            raise RuntimeError("No platforms to build for")
+        if user_platforms_override:
+            final_platforms = platforms
```
chmeliik, line 117: This allows users  to enable a platform that is not defined in the config, do we want that?

It worked like that before, but since we're already changing the logic, maybe we should fix it?

atomic_reactor/plugins/check_and_set_platforms.py
```diff
@@ -76,13 +77,10 @@ def run(self) -> Optional[List[str]]:
             if is_scratch_build(self.workflow) or is_isolated_build(self.workflow):
                 override_platforms = set(user_platforms or [])
                 if override_platforms and override_platforms != set(platforms):
-                    sorted_platforms = sorted(override_platforms)
-                    self.log.info("Received user specified platforms %s", sorted_platforms)
+                    user_platforms_override = True
```
chmeliik, line 80: Do we still want to consider it a user override only if it doesn't match the platforms from the koji target?

chmeliik: Might not be safe to change this though, it's user-facing behavior

rcerven: brew is ALWAYS passing platforms to osbs, so this is how we are able to distinguish if user overrode platforms

rcerven: non target case is corner case when disabled koji integration

atomic_reactor/plugins/check_and_set_platforms.py
chmeliik: I don't think it will be skipped, I think it will fail like this
https://github.com/containerbuildsystem/atomic-reactor/blob/20e8170fec6d3a0d21c0eec674d9a553b26ba4b9/atomic_reactor/utils/remote_host.py#L669-L671

https://github.com/containerbuildsystem/atomic-reactor/blob/20e8170fec6d3a0d21c0eec674d9a553b26ba4b9/atomic_reactor/tasks/binary_container_build.py#L130-L134

atomic_reactor/plugins/check_and_set_platforms.py
chmeliik: Hmm, but this branch is the one where target _is_ defined

rcerven: koji always provides platforms to osbs

override_platforms and override_platforms != set(platforms):

means that platforms changed, aka user specified them

if they are same as target, WE can't tell if they are from target OR if user specified ALL of them

atomic_reactor/plugins/check_and_set_platforms.py
rcerven: https://github.com/containerbuildsystem/atomic-reactor/blob/20e8170fec6d3a0d21c0eec674d9a553b26ba4b9/atomic_reactor/tasks/binary_container_build.py#L70

rcerven: also that "feature" is used by fedora

atomic_reactor/plugins/check_and_set_platforms.py
chmeliik: Ah, that makes sense :+1:

atomic_reactor/plugins/check_and_set_platforms.py
chmeliik: > https://github.com/containerbuildsystem/atomic-reactor/blob/20e8170fec6d3a0d21c0eec674d9a553b26ba4b9/atomic_reactor/tasks/binary_container_build.py#L70
> 
> if platform isn't enabled / defined it will just return

On that line, `enabled_platforms` is the return value of the check_and_set_platforms plugin. If this plugin returns a platform which is not defined in the config, the build will fail with "Failed to acquire a build slot ..."

chmeliik (approved):
LGTM :+1:

ben-alkov (approved):
LGTM

PR #1963, opened 2022-07-21
Ignore empty layers in image history

chmeliik:
CLOUDBLD-9968

When figuring out the sizes of layers corresponding to rootfs.diff_ids,
we use data from the image history. The diff_ids contain only non-empty
layers, history contains all layers including empty ones (created by
e.g. a LABEL or ENV instruction). Filter out empty layers from history
to get correct data.

To do that, we need to combine the .history attribute from the blob
config with the `docker history` output. The blob config history shows
which layers are empty but doesn't show layer sizes, docker history vice
versa. Note: Size == 0 does not necessarily mean the layer is empty (in
the sense of not creating a diff_id in the rootfs). See also
https://github.com/opencontainers/image-spec/blob/main/config.md

Previously, this was not an issue, because images built in OSBS 1 had
only non-empty layers in history. Those built in OSBS 2 also have empty
layers.

### DIFF
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -33,8 +33,6 @@ from atomic_reactor.source import get_source_instance_for, DummySource
 from atomic_reactor.constants import (
     CONTAINER_DEFAULT_BUILD_METHOD,
     DOCKER_STORAGE_TRANSPORT_NAME,
-    INSPECT_ROOTFS,
-    INSPECT_ROOTFS_LAYERS,
     PLUGIN_BUILD_ORCHESTRATE_KEY
 )
 from atomic_reactor.util import exception_message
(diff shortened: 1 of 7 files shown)


chmeliik:
Turns out we can't always ignore based on size = 0 :disappointed:

ben-alkov:
LGTM so far

chmeliik:
PTAL, should be ready now (assuming that my assumption about the tag_and_push plugin is correct -- that we always push v2s2 so this `else` branch can't actually happen (otherwise we wouldn't have the config blob we need) https://github.com/containerbuildsystem/atomic-reactor/blob/346de6126e2afcf5a729d9c37a8cd5123c894e79/atomic_reactor/plugins/post_tag_and_push.py#L238-L243

rcerven:
> PTAL, should be ready now (assuming that my assumption about the tag_and_push plugin is correct -- that we always push v2s2 so this `else` branch can't actually happen (otherwise we wouldn't have the config blob we need)
> 
> https://github.com/containerbuildsystem/atomic-reactor/blob/346de6126e2afcf5a729d9c37a8cd5123c894e79/atomic_reactor/plugins/post_tag_and_push.py#L238-L243

well there is still support for v2s1, but I am not sure if that is even working now (if forced to use only s1)

rcerven (approved):

ben-alkov (approved):

chmeliik:
Release notes:
* OSBS will correctly determine layer sizes for images with empty layers in docker history output

PR #1965, opened 2022-07-22
since we removed platforms from logs, koji-container can't find metadata

rcerven:
info for scratch build, so log platform:_metadata_ in log message

* CLOUDBLD-10586

### DIFF
--- a/atomic_reactor/plugins/koji_import.py
+++ b/atomic_reactor/plugins/koji_import.py
@@ -13,7 +13,6 @@ from itertools import chain
 import koji
 import os
 import time
-import logging
 from tempfile import NamedTemporaryFile
 from typing import Any, Dict, Iterator, List, Optional, Tuple, Iterable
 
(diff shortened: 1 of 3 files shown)


ben-alkov (approved):

chmeliik:
Do we still want to log it as `platform:_metadata_`? Maybe we should change it to something that makes more sense?

rcerven:
> Do we still want to log it as `platform:_metadata_`? Maybe we should change it to something that makes more sense?

well if we change it we have to update also koji-c, but if you want to change it, suggest something what makes more sense and is unique enough, I don't think that it matters that much as long it is unique

chmeliik (approved):
> > Do we still want to log it as `platform:_metadata_`? Maybe we should change it to something that makes more sense?
> 
> well if we change it we have to update also koji-c, but if you want to change it, suggest something what makes more sense and is unique enough, I don't think that it matters that much as long it is unique

Hmm, yeah that's true. Works for me

LGTM :+1:

PR #1970, opened 2022-08-03
build task will record error message for get_build_error to show

rcerven:
* CLOUDBLD-10627

### DIFF
--- a/atomic_reactor/cli/parser.py
+++ b/atomic_reactor/cli/parser.py
@@ -173,6 +173,12 @@ def _add_common_task_args(task_parser: argparse.ArgumentParser) -> None:
         required=True,
         help="PipelineRun name to reference current PipelineRun"
     )
+    task_parser.add_argument(
+        "--task-result",
+        metavar="FILE",
+        default=None,
+        help="file to write task result",
+    )
     # Two different ways to pass user params
     userparams_ex = task_parser.add_mutually_exclusive_group()
     userparams_ex.add_argument(
--- a/atomic_reactor/tasks/common.py
+++ b/atomic_reactor/tasks/common.py
@@ -8,9 +8,10 @@ of the BSD license. See the LICENSE file for details.
 
 import abc
 import signal
-from dataclasses import dataclass, fields
+import json
+from dataclasses import dataclass
 from pathlib import Path
-from typing import Dict, Any, ClassVar, Generic, TypeVar
+from typing import Dict, Any, ClassVar, Generic, TypeVar, Optional
 from functools import cached_property
 
 from atomic_reactor import config
@@ -19,6 +20,11 @@ from atomic_reactor import inner
 from atomic_reactor import source
 from atomic_reactor import util
 from atomic_reactor.plugin import TaskCanceledException
+
+
+def write_task_result(output_file, msg):
+    with open(output_file, 'w') as f:
+        f.write(msg)
 
 
 @dataclass(frozen=True)
@@ -33,6 +39,7 @@ class TaskParams:
     namespace: str
     pipeline_run_name: str
     user_params: Dict[str, Any]
+    task_result: Optional[str]
 
     # Note: do not give any attributes in this class default values, that would make dataclass
     #   inheritance difficult. If they should have defaults, define them in the CLI parser.
@@ -59,7 +66,6 @@ class TaskParams:
     @classmethod
     def from_cli_args(cls, args: dict):
         """Create a TaskParams instance from CLI arguments."""
-        args = cls._drop_known_unset_args(args)
         params_str = args.pop("user_params", None)
         params_file = args.pop("user_params_file", None)
 
@@ -71,17 +77,6 @@ class TaskParams:
             raise ValueError("Did not receive user params. User params are currently required.")
 
         return cls(**args, user_params=user_params)
-
-    @classmethod
-    def _drop_known_unset_args(cls, args: dict) -> dict:
-        # When an argument is not set on the CLI, argparse stores it as None. Drop those arguments
-        #   to avoid accidentally setting required attributes to None, make sure we instead get
-        #   a TypeError from __init__().
-        # Drop only arguments defined on this class (or a parent class), if an unknown argument
-        #   is received, we want a TypeError regardless of the value.
-        # The CLI should be responsible for not letting any of this happen, but let's double-check.
-        known_args = {f.name for f in fields(cls)}
-        return {k: v for k, v in args.items() if v is not None or k not in known_args}
 
 
 ParamsT = TypeVar("ParamsT", bound=TaskParams)
@@ -127,7 +122,14 @@ class Task(abc.ABC, Generic[ParamsT]):
             else:
                 signal.signal(signal.SIGTERM, self.throw_task_canceled_exception)
 
-            self.execute(*args, **kwargs)
+            result = self.execute(*args, **kwargs)
+            if self._params.task_result:
+                write_task_result(self._params.task_result, json.dumps(result))
+
+        except Exception as e:
+            if self._params.task_result:
+                write_task_result(self._params.task_result, repr(e))
+            raise e
 
         finally:
             signal.signal(signal.SIGTERM, signal.SIG_DFL)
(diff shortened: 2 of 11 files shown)


atomic_reactor/cli/parser.py
```diff
@@ -173,6 +173,12 @@ def _add_common_task_args(task_parser: argparse.ArgumentParser) -> None:
         required=True,
         help="PipelineRun name to reference current PipelineRun"
     )
+    task_parser.add_argument(
+        "--task-result",
+        metavar="FILE",
+        default=None,
+        help="file to write task result",
```
chmeliik, line 180: That doesn't sound quite right. We don't use it for task results but only for the error message. Which can't even be a task result, because failed tasks don't produce results.

How about something like `--write-error`?

atomic_reactor/tasks/common.py
```diff
@@ -33,6 +33,7 @@ class TaskParams:
     namespace: str
     pipeline_run_name: str
     user_params: Dict[str, Any]
+    task_result: str
```
chmeliik, line 36:
```suggestion
    task_result: Optional[str]
```

atomic_reactor/tasks/binary_container_build.py
```diff
@@ -119,23 +130,29 @@ def execute(self) -> None:
             image_size = podman_remote.get_image_size(dest_tag)
 
             if image_size > image_size_limit:
-                raise ExceedsImageSizeError(
+                exc = ExceedsImageSizeError(
                     'The size {} of image {} exceeds the limitation {} '
                     'configured in reactor config.'
                     .format(image_size, dest_tag, image_size_limit)
                 )
+                write_task_error(task_result, repr(exc))
```
chmeliik, line 138: Instead of handling this individually in 10 different places in the build task, could handle this in a single place in Task.run? https://github.com/containerbuildsystem/atomic-reactor/blob/e96d1b89b268731a294a9b7d6495de49f41178aa/atomic_reactor/tasks/common.py#L123-L135

Add an `except` block, check if the param for writing error message was set, write error.

atomic_reactor/tasks/common.py
```diff
@@ -59,6 +60,7 @@ def source(self) -> source.Source:
     @classmethod
     def from_cli_args(cls, args: dict):
         """Create a TaskParams instance from CLI arguments."""
+        task_result = args.pop("task_result", None)
         args = cls._drop_known_unset_args(args)
```
chmeliik, line 64: I see that `_drop_known_unset_args` will cause issues with `Optional` params. IMO we can remove the method, it's the CLI's responsibility to make sure required params will never be unset.

(and then `task_result` wouldn't need special handling here)

tekton/pipelines/binary-container.yaml
```diff
@@ -25,6 +25,14 @@ spec:
...
     - name: remote_sources
       value: $(tasks.binary-container-set-results.results.remote_sources)
+    - name: task_build_result_x86_64
+      value: $(tasks.binary-container-build-x86-64.results.task_result)
+    - name: task_build_result_s390x
+      value: $(tasks.binary-container-build-s390x.results.task_result)
+    - name: task_build_result_ppc64le
+      value: $(tasks.binary-container-build-ppc64le.results.task_result)
+    - name: task_build_result_aarch64
+      value: $(tasks.binary-container-build-aarch64.results.task_result)
```
chmeliik, line 35: Do we need these here? IMO they're misleading, failed tasks don't produce results so these can never have any value, right?

tekton/tasks/binary-container-build.yaml
```diff
@@ -28,6 +28,9 @@ spec:
     - name: ws-reactor-config-map
     - name: ws-remote-host-auth
 
+  results:
+    - name: task_result
```
chmeliik, line 32: Same here - but I guess this one might be needed, otherwise we don't get the thing we use to check the error message

Which remind me, could you explain -- in the commit message or in a comment in the YAML -- how this gets us the error message?  Cause it's not through task results but something else, right?

atomic_reactor/cli/parser.py
rcerven: sure

atomic_reactor/tasks/common.py
rcerven: that didn't work, as binary params has mandatory platforms

atomic_reactor/tasks/binary_container_build.py
rcerven: mmm I guess it should work

atomic_reactor/tasks/common.py
rcerven: well I can try to remove it, dunno if it won't break stuff

tekton/tasks/binary-container-build.yaml
rcerven: this is required for sure, because without results it won't set error msg, I can rename it to task_error or something tho

tekton/pipelines/binary-container.yaml
rcerven: this is for consistency, since we are using 'results' ........................ I'll rename it

rcerven:
after discussion with @chmeliik I'll leave it as task_result, because if task passes it can be used as result, if task fails it is used in error

atomic_reactor/tasks/common.py
chmeliik: It shouldn't, the CLI looks correct

rcerven: dropped

atomic_reactor/tasks/binary_container_build.py
rcerven: done

atomic_reactor/tasks/common.py
```diff
@@ -127,7 +121,14 @@ def run(self, *args, **kwargs):
             else:
                 signal.signal(signal.SIGTERM, self.throw_task_canceled_exception)
 
-            self.execute(*args, **kwargs)
+            result = self.execute(*args, **kwargs)
+            if self._params.task_result:
+                write_task_result(self._params.task_result, repr(result))
```
chmeliik, line 126: Will `repr` be useful? Maybe we could encode it as JSON instead?

atomic_reactor/tasks/common.py
```diff
@@ -127,7 +121,14 @@ def run(self, *args, **kwargs):
...
+
+        except Exception as e:
+            if self._params.task_result:
+                write_task_result(self._params.task_result, repr(e))
+            raise(e)
```
chmeliik, line 131: Do we have unit tests for writing the results? I couldn't find one

atomic_reactor/tasks/common.py
rcerven: done

atomic_reactor/tasks/common.py
rcerven: added

tests/tasks/test_common.py
```diff
@@ -225,3 +228,29 @@ def execute(self):
...
+        else:
+            task.run()
+
+        with open(params.task_result) as f:
+            results = f.read()
+
+        if fails:
+            results == "Exception('failed')"
+        else:
+            json.loads(results) == 'result'
```
chmeliik, line 256:
```suggestion
        if fails:
            assert results == "Exception('failed')"
        else:
            assert json.loads(results) == 'result'
```

atomic_reactor/tasks/common.py
```diff
@@ -127,7 +122,14 @@ def run(self, *args, **kwargs):
...
 
-            self.execute(*args, **kwargs)
+            result = self.execute(*args, **kwargs)
+            if self._params.task_result:
+                write_task_result(self._params.task_result, json.dumps(result))
+
+        except Exception as e:
+            if self._params.task_result:
+                write_task_result(self._params.task_result, repr(e))
+            raise(e)
```
chmeliik, line 132: style nitpick:
```suggestion
            raise e
```

mkosiarc (approved):

rcerven:
rn updated

PR #1972, opened 2022-08-04
Use the configured path for manifests instead of default

mkosiarc:
CLOUDBLD-10685

### DIFF
--- a/atomic_reactor/plugins/export_operator_manifests.py
+++ b/atomic_reactor/plugins/export_operator_manifests.py
@@ -100,7 +100,11 @@ class ExportOperatorManifestsPlugin(Plugin):
             raise ValueError(f'Operator manifests check in built image failed: {e}') from e
 
         repo_csv: OperatorCSV = OperatorManifest.from_directory(
-            os.path.join(self.workflow.build_dir.any_platform.path, MANIFESTS_DIR_NAME)).csv
+            os.path.join(
+                self.workflow.build_dir.any_platform.path,
+                self.workflow.source.config.operator_manifests["manifests_dir"],
+            )
+        ).csv
 
         if image_csv.checksum != repo_csv.checksum:
             image_csv_filename = os.path.basename(image_csv.path)
--- a/tests/plugins/test_export_operator_manifests.py
+++ b/tests/plugins/test_export_operator_manifests.py
@@ -125,6 +125,7 @@ def mock_env(workflow, has_appregistry_label=False, appregistry_label=False,
 
     env.workflow.build_dir.init_build_dirs(PLATFORMS, env.workflow.source)
     env.workflow.data.tag_conf.add_unique_image(TEST_IMAGE)
+    env.workflow.source.config.operator_manifests = {"manifests_dir": "manifests"}
 
     mock_oc_image_extract = functools.partial(extract_manifests_dir, empty=empty_archive,
                                               multiple_csv=multiple_csv, has_archive=has_archive,


atomic_reactor/plugins/export_operator_manifests.py
```diff
@@ -100,7 +100,11 @@ def _verify_csv(self, manifests_dir) -> None:
             raise ValueError(f'Operator manifests check in built image failed: {e}') from e
 
         repo_csv: OperatorCSV = OperatorManifest.from_directory(
-            os.path.join(self.workflow.build_dir.any_platform.path, MANIFESTS_DIR_NAME)).csv
+            os.path.join(
+                self.workflow.build_dir.any_platform.path,
+                self.workflow.source.config.operator_manifests["manifests_dir"],
+            )
+        ).csv
```
chmeliik, line 107: The pin_operator_digest plugin does additional validation on the configured path, which this one does not

https://github.com/containerbuildsystem/atomic-reactor/blob/e96d1b89b268731a294a9b7d6495de49f41178aa/atomic_reactor/plugins/pin_operator_digest.py#L316-L319

This one does run later, so the validation should have already happened, but are we confident there is no possible code path where pin_operator_digest doesn't do the validation and this plugin still runs?

Maybe BuildDir should have something like a `resolve_subpath` method where you pass a (supposed to be) relative path, it checks that it's really a subpath and returns the absolute path. We could use it to do this check in all the places that need it.

chmeliik: Related to the "no possible code path": what should this plugin do if `operator_manifests` are not in the config?

https://github.com/containerbuildsystem/atomic-reactor/blob/e96d1b89b268731a294a9b7d6495de49f41178aa/atomic_reactor/plugins/pin_operator_digest.py#L278-L281

(I guess it's fine to just fail, OSBS 1 would have failed as well)

mkosiarc: I don't think that the validation can be skipped as it is always called in the run method of the plugin in pin_operator_digest. It was implemented the same way in osbs1 - https://github.com/containerbuildsystem/atomic-reactor/blob/osbs_ocp3/atomic_reactor/plugins/post_export_operator_manifests.py#L148

I agree that your proposed idea makes sense in some places, but I feel like export_operator_manifests post plugin should do additional validation that already happened in prebuild plugins.

chmeliik: In OSBS 1, `Source.manifests_dir` was a property that did the validation https://github.com/containerbuildsystem/atomic-reactor/blob/e241459d10dbcd08ad4cc97bbf080aa4c0c985ac/atomic_reactor/source.py#L101

But it's not critical, like you said, if `pin_operator_digest` does run, then the validation does happen.

I think there is a case -- the one I mentioned above -- where pin_operator_digest will not run and export_operator_manifest will try to do this check anyway. But in that case, OSBS 2 will die with a KeyError (OSBS 1 would have died with a RuntimeError) so I guess that's OK

mkosiarc: I guess that is ok. We can also add check not to run the export plugin if pin_operator_digest did not run, what do you think?

rcerven: plugin is based on operator labels in dockerfile

no need to check if pin plugin ran

mkosiarc: soo..acks then ? :D

rcerven (approved):

chmeliik (approved):
LGTM :+1:

ben-alkov (approved):

PR #1973, opened 2022-08-09
check_and_set_platforms sets platforms_results with final platforms, for

rcerven:
koji-container build to consume

* CLOUDBLD-10661

### DIFF
--- a/atomic_reactor/cli/parser.py
+++ b/atomic_reactor/cli/parser.py
@@ -82,6 +82,8 @@ def parse_args(args: Optional[Sequence[str]] = None) -> dict:
         description="Execute binary container pre-build steps.",
     )
     binary_container_prebuild.set_defaults(func=task.binary_container_prebuild)
+    binary_container_prebuild.add_argument("--platforms-result", metavar="FILE", default=None,
+                                           help="file to write final platforms result")
 
     binary_container_build = tasks.add_parser(
         "binary-container-build",
--- a/atomic_reactor/inner.py
+++ b/atomic_reactor/inner.py
@@ -465,6 +465,7 @@ class DockerBuildWorkflow(object):
         plugins_conf: Optional[List[Dict[str, Any]]] = None,
         plugin_files: Optional[List[str]] = None,
         keep_plugins_running: bool = False,
+        platforms_result: Optional[str] = None,
     ):
         """
         :param context_dir: the directory passed to task --context-dir argument.
@@ -484,6 +485,7 @@ class DockerBuildWorkflow(object):
         :param client_version: osbs-client version used to render build json
         :param bool keep_plugins_running: keep plugins running even if error is
             raised from previous one. This is passed to ``PluginsRunner`` directly.
+        :param platforms_result: path to platform results for prebuild task
         """
         self.context_dir = context_dir
         self.build_dir = build_dir
@@ -492,6 +494,7 @@ class DockerBuildWorkflow(object):
         self.pipeline_run_name = pipeline_run_name
         self.source = source or DummySource(None, None)
         self.user_params = user_params or self._default_user_params.copy()
+        self.platforms_result = platforms_result
 
         self.keep_plugins_running = keep_plugins_running
         self.plugin_files = plugin_files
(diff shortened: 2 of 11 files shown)


rcerven:
related prs: https://github.com/containerbuildsystem/osbs-client/pull/1109  https://github.com/containerbuildsystem/koji-containerbuild/pull/247

atomic_reactor/cli/parser.py
```diff
@@ -179,6 +179,12 @@ def _add_common_task_args(task_parser: argparse.ArgumentParser) -> None:
         default=None,
         help="file to write task result",
     )
+    task_parser.add_argument(
+        "--platforms-result",
+        metavar="FILE",
+        default=None,
+        help="file to write final platforms result",
+    )
```
chmeliik, line 187: Could we use `--task-result` instead of adding a new param? (We would need to update BinaryPrebuildTask.execute):
```python
    def execute(self, init_build_dirs: Optional[bool] = False) -> Dict[str, Any]:
        super().execute(init_build_dirs)
        return {
            "final_platforms": util.get_platforms(self.workflow_data)
        }
```

rcerven: no !  task-result is also for future use, I don't want to use it randomly

rcerven: tekton is about new params .....

rcerven: it would be using it for something else, task result vs plugin result

and overloading results can create only additional logic when parsing them in the future

chmeliik: The prebuild task is still a task, why should it matter that it's implemented as plugins?

What do you mean additional parsing logic? Any `--task-result` of a successful task has to be JSON, so we'll be parsing JSON no matter what

rcerven: using it for different purposes, like in this case for very specific platforms result

chmeliik: Hmm, alright, I don't care _that_ much if it's a new param or not. But maybe the param should only be valid for the prebuild task? (In the parser, in a new PrebuildTaskParams class etc.). Since the param does nothing in any of the other tasks

rcerven: ok param is valid only for prebuild task

atomic_reactor/tasks/common.py
```diff
@@ -40,6 +40,7 @@ class TaskParams:
     pipeline_run_name: str
     user_params: Dict[str, Any]
     task_result: Optional[str]
+    platforms_result: Optional[str]
```
chmeliik, line 43: The param can't be in the common `TaskParams`, only the prebuild task has it

rcerven: it is optional here, so it doesn't matter, and it we don't have to create new subclass PreBuildTaskParams ......

chmeliik: I'm pretty sure you'll get `TypeError: ... missing 1 required positional argument: 'platforms_result'` in every task except prebuild, but I'll check

chmeliik: Indeed
```bash
$ atomic-reactor task --build-dir build/ --context-dir context/ --user-params-file user_params.json --namespace idk --pipeline-run-name idk clone
TypeError: BinaryBuildTaskParams.__init__() missing 1 required positional argument: 'platforms_result'
```

rcerven: fixed

atomic_reactor/inner.py
```diff
@@ -492,6 +493,7 @@ def __init__(
         self.pipeline_run_name = pipeline_run_name
         self.source = source or DummySource(None, None)
         self.user_params = user_params or self._default_user_params.copy()
+        self.platforms_result = platforms_result
```
lkolacek, line 496: Should we add a short dockstring also for `platforms_result` param?

rcerven: added

lkolacek (approved):

mkosiarc (approved):

rcerven:
rn updated

PR #1974, opened 2022-08-11
Don't follow symlinks when copying files

mkosiarc:
CLOUDBLD-10684

https://docs.python.org/3/library/shutil.html#shutil.copytree

### DIFF
--- a/tests/test_dirs.py
+++ b/tests/test_dirs.py
@@ -360,6 +360,41 @@ def test_rootbuilddir_for_all_platforms_copy_invalid_file_path(
         root.for_all_platforms_copy(creation_func)
 
 
+def test_rootbuilddir_for_all_platforms_copy_dir_with_broken_symlinks(build_dir, mock_source):
+    """
+    Test that broken symlinks inside a directory are not followed/resolved when
+    copying build_dir contents to platforms dir.
+    """
+
+    def create_dir_with_broken_symlinks_inside(build_dir: BuildDir) -> Iterable[Path]:
+        directory = build_dir.path.joinpath("dir")
+        directory.mkdir()
+
+        symlink_to_nonexistent_file = directory.joinpath("symlink.file")
+        symlink_to_nonexistent_file.symlink_to(build_dir.path / directory / "nonexistent.file")
+
+        circular_symlink_1 = directory.joinpath("circular-symlink_1.file")
+        circular_symlink_2 = directory.joinpath("circular-symlink_2.file")
+
+        circular_symlink_1.symlink_to(circular_symlink_2)
+        circular_symlink_2.symlink_to(circular_symlink_1)
+
+        return [directory]
+
+    root = RootBuildDir(build_dir)
+    platforms = ["x86_64", "ppc64le"]
+    root.init_build_dirs(platforms, mock_source)
+    root.for_all_platforms_copy(create_dir_with_broken_symlinks_inside)
+
+    for platform in platforms:
+        dir_path = root.path / platform / "dir"
+        assert {file.name for file in dir_path.iterdir()} == {
+            "symlink.file",
+            "circular-symlink_1.file",
+            "circular-symlink_2.file",
+        }
+
+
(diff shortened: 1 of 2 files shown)


chmeliik (approved):
LGTM :+1:

chmeliik:
Do we want a unit test for this?

lkolacek (approved):

mkosiarc:
Added unit test

tests/test_dirs.py
```diff
@@ -342,10 +343,6 @@ def create_file_build_dir_is_returned_as_created(build_dir: BuildDir) -> Iterabl
         create_file_return_nonexisting_relative_path,
         pytest.raises(FileNotFoundError, match="does not exist inside build directory"),
     ],
-    [
-        create_file_linked_to_outside_file,
-        pytest.raises(ValueError, match="File must be created inside"),
```
chmeliik, line 347: Hmm, maybe we should still check this? Don't follow broken symlinks, but do protect against potentially malicious ones?

mkosiarc: I am not sure what you mean, because the test is still checking the same thing, except now it does not fail.

chmeliik:
```
.
├── broken-link -> nonexistent-file
├── circular-link -> circular-link2
├── circular-link2 -> circular-link
└── malicious-link -> /etc/shadow
```
I think we should allow broken-link and circular-link but not malicious-link. The test was previously checking the malicious-link case, I don't think we had tests for the other two

mkosiarc: So, do you mean that:

1. that we should still raise error when symlink points to a path outside of build_dir
2. allow broken links, circular links, working links that point to files in build_dir (but still copy only the symlink)

chmeliik: Although we won't be checking malicious links in subdirectories anyway, so it doesn't matter that much.

Basically, my point was that we still want this line
```python
file_path = file_path.resolve()
```
But that's not that important. The more important point is that we still don't have a test case for broken symlinks :smile:

chmeliik: Definitely 2. (that's already solved, just needs the test), 1. matters less

chmeliik: Btw, it seems that creating a broken symlink can be done like this
```python
Path("broken-link").symlink_to("doesnt-exist")
```

chmeliik: Ahh, I forgot about the `exists` check. Broken symlinks won't work if we keep that check. 

But keep in mind that we only check files which were directly returned by the function passed to for_all_platforms. If the function returns a directory, we check the directory itself but not anything inside. And that's really the main thing we want to fix (for the Cachito use case).

It's probably OK to fail for broken symlinks returned directly

mkosiarc: there is no other way, because even if the exists check was removed, then it would fail on executing copy method. So I'll try to incorporate unit test that creates dir with broken symlinks inside.

mkosiarc: I believe that now it should be ok. Although it seems strange to me that we are treating files in a directory differently.

tests/test_dirs.py
```diff
@@ -360,6 +360,41 @@ def test_rootbuilddir_for_all_platforms_copy_invalid_file_path(
...
+
+        circular_symlink_1.symlink_to(circular_symlink_2)
+        circular_symlink_2.symlink_to(circular_symlink_1)
+
+        return [directory]
+
+    root = RootBuildDir(build_dir)
+    platforms = ["x86_64", "ppc64le"]
+    root.init_build_dirs(platforms, mock_source)
+    root.for_all_platforms_copy(create_dir_with_broken_symlinks_inside)
```
rcerven, line 387: why don't you just create all those files/dirs from the method after root.init_build_dirs ONCE, and then in the method just return build_dir.path.joinpath("dir") ????

all  you need is to copy directory from build_dir to all platforms

rcerven (approved):

PR #1979, opened 2022-08-25
don't cache registry client

rcerven:

### DIFF
--- a/tests/utils/test_imageutil.py
+++ b/tests/utils/test_imageutil.py
@@ -164,10 +164,13 @@ class TestImageUtil:
             .once()
             .and_return(registry_session)
         )
-        flexmock(util.RegistryClient).should_receive("__init__").with_args(registry_session).once()
+        (
+            flexmock(util.RegistryClient)
+            .should_receive("__init__")
+            .with_args(registry_session)
+            .once()
+        )
 
-        image_util._get_registry_client("registry.com")
-        # test caching (i.e. test that the create_from_config method is called only once)
         image_util._get_registry_client("registry.com")
 
     def test_extract_file_from_image_non_empty_dst_dir(self, tmpdir):
--- a/atomic_reactor/utils/imageutil.py
+++ b/atomic_reactor/utils/imageutil.py
@@ -108,7 +108,6 @@ class ImageUtil:
         client = self._get_registry_client(parsed_image.registry)
         return client.get_inspect_for_image(parsed_image, goarch)
 
-    @functools.lru_cache(maxsize=None)
     def _get_registry_client(self, registry: str) -> util.RegistryClient:
         session = util.RegistrySession.create_from_config(self._conf, registry)
         return util.RegistryClient(session)


chmeliik:
Assuming this is the solution, should we remove the caching just from imageutil or from all the other places that cache RegistryClients as well (check_base_image, koji_parent, pin_operator_digest, maybe others)?

tests/utils/test_imageutil.py
```diff
@@ -161,10 +161,15 @@ def test_get_registry_client(self):
...
-        flexmock(util.RegistryClient).should_receive("__init__").with_args(registry_session).once()
+        (
+            flexmock(util.RegistryClient)
+            .should_receive("__init__")
+            .with_args(registry_session)
+            .times(2)
+        )
 
         image_util._get_registry_client("registry.com")
         # test caching (i.e. test that the create_from_config method is called only once)
```
chmeliik, line 175: we probably don't want to "test caching" anymore

rcerven:
> Assuming this is the solution, should we remove the caching just from imageutil or from all the other places that cache RegistryClients as well (check_base_image, koji_parent, pin_operator_digest, maybe others)?

Other places should be fine, as they are caching static data afaik

chmeliik:
> > Assuming this is the solution, should we remove the caching just from imageutil or from all the other places that cache RegistryClients as well (check_base_image, koji_parent, pin_operator_digest, maybe others)?
> 
> Other places should be fine, as they are caching static data afaik

I meant the get_registry_client methods, I think they do exactly the same kind of caching that this one did

But they're only used in one plugin, so there isn't one hour delay between re-using the same client. Might be fine
https://github.com/containerbuildsystem/atomic-reactor/blob/36c63700ebb163b75b3209f2c0b49f97fbd7d95a/atomic_reactor/plugins/check_base_image.py#L205-L214

chmeliik:
Could you add to the commit message why we don't want the caching?

And the ticket ID if we have one

rcerven:
> > > Assuming this is the solution, should we remove the caching just from imageutil or from all the other places that cache RegistryClients as well (check_base_image, koji_parent, pin_operator_digest, maybe others)?
> > 
> > 
> > Other places should be fine, as they are caching static data afaik
> 
> I meant the get_registry_client methods, I think they do exactly the same kind of caching that this one did
> 
> But they're only used in the same plugin, so there isn't one hour delay between re-using the same client. Might be fine
> 
> https://github.com/containerbuildsystem/atomic-reactor/blob/36c63700ebb163b75b3209f2c0b49f97fbd7d95a/atomic_reactor/plugins/check_base_image.py#L205-L214

yeah within the same plugin it is fine, as those plugins don't run for a long time

rcerven:
> Could you add to the commit message why we don't want the caching?
> 
> And the ticket ID if we have one

updated commit msg

chmeliik (approved):
LGTM :+1:

mkosiarc (approved):

lkolacek (approved):

rcerven:
no rn, fix for unreleased change

PR #1980, opened 2022-08-26
fix logging

hjmodi:

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -1389,6 +1389,9 @@ class OSBSLogs(object):
             if task_platform in logfiles:
                 log_file = logfiles[task_platform]
             else:
+                logfiles['noarch'].write(bytearray(f'{task_platform} build has started. '
+                                                   + 'Check platform specific logs\n', 'utf-8'))
+
                 log_file = NamedTemporaryFile(
                     prefix=f'{pipeline_run_name}-{filename}-{task_platform}-',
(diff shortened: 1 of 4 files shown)


rcerven (approved):

rcerven:
don't forget to remove bumping osbs commit before merging

atomic_reactor/util.py
```diff
@@ -1389,6 +1389,9 @@ def get_log_files(self, osbs: OSBS, pipeline_run_name) -> List[Output]:
             if task_platform in logfiles:
                 log_file = logfiles[task_platform]
             else:
+                logfiles['noarch'].write(bytearray(f'{task_platform} build has started. '
+                                                   + 'Check platform specific logs\n', 'utf-8'))
```
chmeliik, line 1393: Shouldn't we write this only if `task_platform != 'noarch'`?

rcerven: mmmm right it should

hjmodi: yes, this will only occur for `!= 'noarch'` because `noarch` will already be in `logfiles`

rcerven:
but I don't think we even need this change, because that is used just for koji build logs which are imported

so nobody can be even watching osbs-build.log at this point

hjmodi:
agreed, closing it.

hjmodi:
Don't need it.

PR #1983, opened 2022-08-30
Cleanup on retry

hjmodi:

### DIFF
--- a/atomic_reactor/plugins/build_source_container.py
+++ b/atomic_reactor/plugins/build_source_container.py
@@ -37,9 +37,10 @@ class SourceContainerPlugin(Plugin):
         source_img = 'oci:{}'.format(image_output_dir)
         dest_img = 'docker-archive:{}'.format(output_path)
         cmd += [source_img, dest_img]
+        cleanup_cmd = ['rm', str(output_path)]
 
         try:
-            retries.run_cmd(cmd)
+            retries.run_cmd(cmd, cleanup_cmd)
         except subprocess.CalledProcessError as e:
             self.log.error("failed to save docker-archive :\n%s", e.output)
             raise
--- a/atomic_reactor/plugins/flatpak_create_oci.py
+++ b/atomic_reactor/plugins/flatpak_create_oci.py
@@ -85,9 +85,10 @@ class FlatpakCreateOciPlugin(Plugin):
 
         cmd = ['skopeo', 'copy', 'oci:{path}:{ref_name}'.format(**metadata), '--format=v2s2',
                'docker-archive:{}'.format(str(build_dir.exported_squashed_image))]
+        cleanup_cmd = ['rm', str(build_dir.exported_squashed_image)]
 
         try:
-            retries.run_cmd(cmd)
+            retries.run_cmd(cmd, cleanup_cmd)
         except subprocess.CalledProcessError as e:
             self.log.error("skopeo copy failed with output:\n%s", e.output)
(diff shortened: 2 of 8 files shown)


ben-alkov (approved):
LGTM

atomic_reactor/plugins/build_source_container.py
```diff
@@ -37,9 +37,10 @@ def export_image(self, image_output_dir: Path) -> Dict[str, Union[str, int]]:
         source_img = 'oci:{}'.format(image_output_dir)
         dest_img = 'docker-archive:{}'.format(output_path)
         cmd += [source_img, dest_img]
+        cleanup_cmd = ['rm', output_path]
 
         try:
-            retries.run_cmd(cmd)
+            retries.run_cmd(cmd, cleanup_cmd)
```
chmeliik, line 43: It would have been a bit more flexible to have any callable as the cleanup action

e.g. like this
```python
retries.run_cmd(cmd, cleanup=lambda: output_path.unlink(missing_ok=True)
```
But I don't mind doing the cleanup by calling a subprocess

atomic_reactor/utils/retries.py
```diff
@@ -109,6 +111,16 @@ def run_cmd(cmd: List[str]) -> bytes:
             e.stdout.decode(),
             e.stderr.decode(),
         )
+        if cleanup_cmd:
+            try:
+                subprocess.run(cleanup_cmd, check=True, capture_output=True)
+            except subprocess.CalledProcessError as c_e:
+                logger.warning(
+                    "Cleanup command: %s failed:\nSTDOUT:\n%s\nSTDERR:\n%s",
+                    cleanup_cmd[0],
```
chmeliik, line 120: We might want to log the full cleanup command here

Or log it before running it, like we do for the main command

atomic_reactor/plugins/flatpak_create_oci.py
```diff
@@ -85,9 +85,10 @@ def build_flatpak_image(self, source, build_dir: BuildDir) -> Dict[str, Any]:
 
         cmd = ['skopeo', 'copy', 'oci:{path}:{ref_name}'.format(**metadata), '--format=v2s2',
                'docker-archive:{}'.format(str(build_dir.exported_squashed_image))]
+        cleanup_cmd = ['rm', str(build_dir.exported_squashed_image)]
```
chmeliik, line 88: Hmm, just noticed that we convert the `Path` to string here but not above. Is it needed or not?

atomic_reactor/utils/retries.py
hjmodi: added logging

atomic_reactor/plugins/build_source_container.py
```diff
@@ -37,9 +37,10 @@ def export_image(self, image_output_dir: Path) -> Dict[str, Union[str, int]]:
         source_img = 'oci:{}'.format(image_output_dir)
         dest_img = 'docker-archive:{}'.format(output_path)
         cmd += [source_img, dest_img]
+        cleanup_cmd = ['rm', output_path]
```
chmeliik, line 40: The output_path is still a `Path`, no? The one that gets formatted into a string is dest_img

hjmodi: oh! That's right, we do need to convert it to string.

hjmodi: updated

PR #1984, opened 2022-08-31
tolerate failures when checking for available slots

hjmodi:
* CLOUDBLD-10930

### DIFF
--- a/atomic_reactor/utils/remote_host.py
+++ b/atomic_reactor/utils/remote_host.py
@@ -653,9 +653,7 @@ class RemoteHostsPool:
                 hostname=hostname, username=attr["username"], ssh_keyfile=attr["auth"],
                 slots=attr.get("slots", 1), socket_path=attr["socket_path"], slots_dir=slots_dir
             )
-            # Check whether host is operational before use it
-            if host.is_operational:
-                hosts.append(host)
+            hosts.append(host)
 
         return cls(hosts, platform)
 
@@ -672,7 +670,15 @@ class RemoteHostsPool:
 
         resources = []
         for host in self.hosts:
-            available_slots = host.available_slots()
+            available_slots = []
+            try:
+                if host.is_operational:
+                    available_slots = host.available_slots()
+            except Exception as ex:
+                # Specific exceptions should be handled in nested methods
+                logger.warning("%s: unable to get available slots: %s", host.hostname, ex)
+                continue
+
             if not available_slots:
                 logger.info("%s: no available slots", host.hostname)
                 continue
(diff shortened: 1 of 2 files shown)


atomic_reactor/utils/remote_host.py
```diff
@@ -653,9 +653,7 @@ def from_config(cls, config: dict, platform: str):
                 hostname=hostname, username=attr["username"], ssh_keyfile=attr["auth"],
                 slots=attr.get("slots", 1), socket_path=attr["socket_path"], slots_dir=slots_dir
             )
-            # Check whether host is operational before use it
-            if host.is_operational:
-                hosts.append(host)
```
chmeliik, line 658: We would still need to `mkdir` the slots dir (is_operational does that as a side effect)

atomic_reactor/utils/remote_host.py
```diff
@@ -672,7 +670,12 @@ def lock_resource(self, prid: str) -> Optional[LockedResource]:
 
         resources = []
         for host in self.hosts:
-            available_slots = host.available_slots()
+            try:
+                available_slots = host.available_slots()
+            except Exception as ex:
+                # Specific exceptions should be handled in nested methods
+                logger.warning("%s: unable to get available slots: %s", host.hostname, ex)
+                available_slots = []
```
chmeliik, line 678: Maybe `continue` straight away? Otherwise logs will say this
```
<host>: unable to get available slots: <error>
<host>: no available slots
```
^ "no available slots" is not true, we don't know that

atomic_reactor/utils/remote_host.py
rcerven: you can also change is_operational and make there few retries with small delay .....

hjmodi: I actually moved `is_operational` to call just before `available_slots` so that way we immediately know if a machine is operational or not right before checking the slots

atomic_reactor/utils/remote_host.py
hjmodi: I removed the log to avoid confusion

chmeliik: Don't we lose the info about hosts which really don't have any available slots?

atomic_reactor/utils/remote_host.py
chmeliik: Seems reasonable :+1:

atomic_reactor/utils/remote_host.py
hjmodi: fixed it. Logging both cases in different conditions

chmeliik (approved):
LGTM :+1:

mkosiarc (approved):

PR #1986, opened 2022-09-01
inject_yum_repos: skip platforms without repos

chmeliik:
CLOUDBLD-10966

Currently, the plugin updates the Dockerfile even for platforms that do
not have any repos. That causes the builds to fail.

Don't update the Dockerfile for platforms that don't have repos. Add a
warning message for such cases - it is strange to have repos only for
some platforms but not others.


atomic_reactor/plugins/inject_yum_repos.py
```diff
@@ -260,14 +267,14 @@ def run(self):
                 fetched_yum_repos[repourl] = yum_repo
 
         if not self.yum_repos:
```
mkosiarc, line 269: Do we need to keep this check here now even if we are returning in _inject_repo_files if there are no repos for the platform?

chmeliik: If we want to keep the `no repos for <platform>` message a warning then yes. Not having _any_ repos is normal

But I can remove the check and reduce the log level to info

mkosiarc: I don't mind then, you can keep it as it is.

mkosiarc (approved):

atomic_reactor/plugins/inject_yum_repos.py
rcerven: this makes sens, this is for when there aren't ANY repos at all, the warning is when there aren't any repos for specific arch

atomic_reactor/plugins/inject_yum_repos.py
```diff
@@ -174,14 +174,19 @@ def add_koji_repo(self):
         for platform in self.platforms:
             self.yum_repos[platform].append(yum_repo)
 
-    def _inject_into_repo_files(self, build_dir: BuildDir):
+    def _inject_repo_files(self, build_dir: BuildDir) -> None:
         """Inject repo files into a relative directory inside the build context"""
+        platform_repos = self.yum_repos[build_dir.platform]
+        if not platform_repos:
+            self.log.warning("no repos for %s, skipping this platform", build_dir.platform)
```
rcerven, line 181: can we change the message to :  

no repos for $platform, no repos to inject

chmeliik: Saying "no repos" twice seemed a bit redundant, changed to

no repos to inject for $platform platform

let me know what you think

rcerven: right sounds good

PR #1989, opened 2022-09-02
Fix how we decide whether to include koji repo

chmeliik:
CLOUDBLD-10966

The include_koji_repo param has to be set separately for each platform.

Fixes a new OSBS 2 bug as well as an old OSBS 1 bug.

---

How did this work in OSBS 1? (Read at your own risk.)

The param was calculated in the resolve_composes plugin (same as now)
and set for all platforms (yes, one value for all platforms).

However, the koji_repo plugin didn't look only at the include_koji_repo
param. It also checked if there were any yum_repourls in user params
(those were also set by the resolve_composes plugin).

    include_koji_repo or not repourls => include koji repo

During OSBS 2 implementation, we simplified (and broke) the logic to

    include_koji_repo => include koji repo

Why is that wrong? The resolve_composes logic for determining
include_koji_repo is roughly this

    not (any(noarch repourls) or any(compose which isn't a module))

Which is obviously wrong, composes are arch-specific. Consider this
situation:

- the build is for x86_64 and ppc64le
- there are no noarch repourls
- only x86_64 has composes

=> include_koji_repo will be False for both arches even though ppc64le
has no repos. This was less of a problem in OSBS 1 thanks to the more
confusing logic.

There *was* a still a bug though:

- x86_64 has regular composes and module composes
- ppc64le has only module composes

=> include_koji_repo will be False for both. ppc64le has some repos, so
the koji_repo plugin will not include anything when it should.

### DIFF
--- a/atomic_reactor/plugins/inject_yum_repos.py
+++ b/atomic_reactor/plugins/inject_yum_repos.py
@@ -12,6 +12,7 @@ import os
 import shutil
 from collections import defaultdict
 from io import StringIO
+from typing import Optional
 from urllib.parse import urlparse
 
 from atomic_reactor.config import get_koji_session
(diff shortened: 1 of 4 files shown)


chmeliik:
Missing unit tests for the corner cases

chmeliik:
> Missing unit tests for the corner cases

added

mkosiarc (approved):

rcerven (approved):

PR #1990, opened 2022-09-07
when using podman-remote paralel build tasks are writing and reading

rcerven:
podman connections to ~/.config/containers/containers.conf, and race between read/write can happen, so instead of using shared pvc for homedir use just emptyDir, home directory contains only few config files but has to be isolated

* CLOUDBLD-10958

- [n/a] Code coverage from testing does not decrease and new code is covered

### DIFF
--- a/tekton/pipelines/binary-container.yaml
+++ b/tekton/pipelines/binary-container.yaml
@@ -13,6 +13,7 @@ spec:
 
   workspaces:
     - name: ws-container
+    - name: ws-home-dir
     - name: ws-registries-secret
     - name: ws-koji-secret
     - name: ws-reactor-config-map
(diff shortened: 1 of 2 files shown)


mkosiarc:
The change itself looks good to me, but the commit message is weird here. I would prefer if the title was something distinct to see right away its meaning. Like "Use separate pvc/workspace for homedir" and the rest of the explanation can be below the commit message title.

rcerven:
> The change itself looks good to me, but the commit message is weird here. I would prefer if the title was something distinct to see right away its meaning. Like "Use separate pvc/workspace for homedir" and the rest of the explanation can be below the commit message title.

done

mkosiarc (approved):

ben-alkov (approved):
LGTM

rcerven:
rn updated

PR #1991, opened 2022-09-12
add build tools and rpms from build hosts to koji metadata, use build…

rcerven:
… hostname as buildroot_id

* CLOUDBLD-10926

### DIFF
--- a/atomic_reactor/plugins/gather_builds_metadata.py
+++ b/atomic_reactor/plugins/gather_builds_metadata.py
@@ -6,12 +6,15 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
 from typing import Any, Dict, Optional
+import json
 
 from atomic_reactor.plugin import Plugin
 from atomic_reactor.constants import PLUGIN_GATHER_BUILDS_METADATA_KEY
 from atomic_reactor.util import Output, is_scratch_build, get_platforms
 from atomic_reactor.utils.koji import get_buildroot, get_output, get_output_metadata
 from osbs.utils import ImageName
+from atomic_reactor.utils.remote_host import RemoteHost
+from atomic_reactor.utils.rpm import parse_rpm_output
 
 
 class GatherBuildsMetadataPlugin(Plugin):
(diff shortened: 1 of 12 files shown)


rcerven:
requires https://github.com/containerbuildsystem/osbs-client/pull/1113 to be merged

atomic_reactor/plugins/gather_builds_metadata.py
```diff
@@ -54,6 +57,56 @@ def _generate_build_log_output(self, platform: str, buildroot_id: int) -> Option
...
+
+        for hostname, attr in platform_config.items():
+            if build_host != hostname:
+                continue
+
+            remote_host = RemoteHost(
+                hostname=hostname, username=attr["username"], ssh_keyfile=attr["auth"],
+                slots=attr.get("slots", 1), socket_path=attr["socket_path"], slots_dir=slots_dir
+            )
+            break
```
mkosiarc, line 97: this whole for loop seems redundant to me. Can't we simply do something like this?
```
if build_host in platform_config: # maybe even this check is redundant
    attr = platform_config[build_host]
    remote_host = RemoteHost(
                hostname=build_host, username=attr["username"], ssh_keyfile=attr["auth"],
                slots=attr.get("slots", 1), socket_path=attr["socket_path"], slots_dir=slots_dir
            )

```
This way we avoid the for loop, ugly continue and break commands.

atomic_reactor/plugins/gather_builds_metadata.py
```diff
@@ -54,6 +57,56 @@ def _generate_build_log_output(self, platform: str, buildroot_id: int) -> Option
...
+
+        for task_name in task_results:
+            if not (task_name.startswith('binary-container-build') and task_platform in task_name):
+                continue
+
+            if 'task_result' not in task_results[task_name]:
+                raise RuntimeError(f"task_results is missing from: {task_name}")
+
+            build_host = json.loads(task_results[task_name]['task_result'])
+            break
```
mkosiarc, line 73: even here we have a for loop, continue and break statements when looking for one element, it does not seem readable or pythonic. I would prefer something like:
```
task_name = next(filter(lambda task_name: task_name.startswith('binary-container-build'), task_results))
...rest of the code
```
It seems more readable, less lines and it has straightforward flow, so you don't have to watch the continue, break statements.

atomic_reactor/plugins/gather_builds_metadata.py
rcerven: ok

atomic_reactor/plugins/gather_builds_metadata.py
rcerven: ok

ben-alkov (approved):
LGTM

mkosiarc (approved):

rcerven:
rn updated

PR #1995, opened 2022-09-22
report which specific task was sucessful

rcerven:
* CLOUDBLD-10933


mkosiarc:
generally looks good to me. But it seems to me that the class itself should have the information about its name instead of passing the name to the class in the params. Seems more reasonable, but if there is no simple way to do that then I'll ack this PR.

ben-alkov (approved):

rcerven:
> generally looks good to me. But it seems to me that the class itself should have the information about its name instead of passing the name to the class in the params. Seems more reasonable, but if there is no simple way to do that then I'll ack this PR.

fixed

mkosiarc (approved):

rcerven:
change in unreleased feature no rn update

PR #1999, opened 2022-10-04
edit .dockerignore file and allow Dockerfile* and *.json for osbs

rcerven:
* CLOUDBLD-11150

### DIFF
--- a/atomic_reactor/plugins/add_dockerfile.py
+++ b/atomic_reactor/plugins/add_dockerfile.py
@@ -44,6 +44,7 @@ from osbs.utils import Labels
 from atomic_reactor.constants import DOCKERFILE_FILENAME, IMAGE_BUILD_INFO_DIR
 from atomic_reactor.dirs import BuildDir
 from atomic_reactor.plugin import Plugin
+from atomic_reactor.util import allow_path_in_dockerignore
 
 
 class AddDockerfilePlugin(Plugin):
@@ -98,10 +99,12 @@ class AddDockerfilePlugin(Plugin):
         if self.use_final_dockerfile:
             # when using final dockerfile, we should use DOCKERFILE_FILENAME
             add_line = f'ADD {DOCKERFILE_FILENAME} {self.df_path}\n'
+            allow_path_in_dockerignore(build_dir.path, DOCKERFILE_FILENAME)
         else:
             # otherwise we should copy current snapshot and use the copied version
             shutil.copy2(build_dir.dockerfile_path, build_dir.path / self.df_name)
             add_line = f'ADD {self.df_name} {self.df_path}\n'
+            allow_path_in_dockerignore(build_dir.path, self.df_name)
 
         # put it before last instruction
         lines.insert(-1, add_line)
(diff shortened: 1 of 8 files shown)


ben-alkov (approved):

atomic_reactor/plugins/add_dockerfile.py
```diff
@@ -110,6 +110,20 @@ def add_dockerfile(self, build_dir: BuildDir) -> None:
...
         self.log.info("added %s for the %s build", self.df_path, build_dir.platform)
 
+    def _modify_dockerignore(self, build_dir: BuildDir) -> None:
+        dockerignore = build_dir.path / DOCKERIGNORE
+
+        if not dockerignore.exists():
+            return
+
+        with open(dockerignore, "a") as f:
+            f.write('\n!/Dockerfile*\n')
```
MartinBasti, line 120: Shouldn't we say explicitly which Dockerfile and json ?

rcerven: I don't think it matters that much, since we are allowing them only in /

either way, even if we would allow just specific ones, AND they would do COPY /*   they would copy those anyway

so if they are relying on this for such copy command, they will be already copying our files anyway

MartinBasti: is it hard to use exacts filenames OSBS, or why we have to do just a semi-fix that may result in another issue in future when somebody will need to avoid copying some json files?

mkosiarc: I agree that we should explicitly add files that are being ADDed in the Dockerfile. Also, I am not sure that we should limit this to .json files and Dockerfiles

rcerven: ok then, I'll change it

mkosiarc (approved):

atomic_reactor/util.py
```diff
@@ -1252,13 +1252,13 @@ def validate_with_schema(data: dict, schema: str, package: str = "atomic_reactor
...
 
-def allow_repo_dir_in_dockerignore(build_path):
+def allow_path_in_dockerignore(build_path, allow_path):
     docker_ignore = os.path.join(str(build_path), DOCKERIGNORE)
 
     if os.path.isfile(docker_ignore):
         with open(docker_ignore, "a") as f:
-            f.write("\n!%s\n" % RELATIVE_REPOS_PATH)
-        logger.debug("Allowing %s in %s", RELATIVE_REPOS_PATH, DOCKERIGNORE)
+            f.write("\n!%s\n" % allow_path)
```
MartinBasti, line 1260: didn't we agreed on new style format string or F-strings?
```suggestion
            f.write(f"\n!{allow_path}\n")
```

MartinBasti:
Can you please also describe better in commit message why?

atomic_reactor/util.py
rcerven: ok

rcerven:
> Can you please also describe better in commit message why?

done

ben-alkov (approved):

MartinBasti (approved):

rcerven:
rn updated

PR #2002, opened 2022-10-05
Import koji metadata from file

mkosiarc:
Upload koji metadata file and use its filename in CGImport call This allows handling files containing control characters

CLOUDBLD-11191

### DIFF
--- a/atomic_reactor/plugins/koji_import.py
+++ b/atomic_reactor/plugins/koji_import.py
@@ -57,6 +57,7 @@ from atomic_reactor.constants import (
     KOJI_SUBTYPE_OP_APPREGISTRY,
     KOJI_SUBTYPE_OP_BUNDLE,
     KOJI_SOURCE_ENGINE,
+    KOJI_METADATA_FILENAME,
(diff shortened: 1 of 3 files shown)


ben-alkov (approved):

atomic_reactor/plugins/koji_import.py
```diff
@@ -532,12 +533,11 @@ def run(self):
         if build_id is not None and build_token is not None:
             koji_metadata['build']['build_id'] = build_id
 
+        self.upload_metadata(koji_metadata, server_dir)
         try:
-            if build_token:
```
MartinBasti, line 536: is it safe to remove this if ?

mkosiarc: If not present, it will have value None. and that is the default value for token in CGImport method anyway.

mkosiarc: So I would say yes.

MartinBasti (approved):
:rocket:

PR #2012, opened 2022-10-17
Create parent dirs when injecting config files

tichavskym:
CLOUDBLD-11265

Cachito's RubyGems package manager creates config files in .bundle/config, expecting the parent directory to already exist. Atomic reactor should support that.

Signed-off-by: Milan Tichavský <mtichavs@redhat.com>

### DIFF
--- a/tests/plugins/test_resolve_remote_source.py
+++ b/tests/plugins/test_resolve_remote_source.py
@@ -218,6 +218,11 @@ CACHITO_CONFIG_FILES = [
         "type": CFG_TYPE_B64,
         "content": base64.b64encode(b"gomod requests don't actually have configs").decode(),
     },
+    {
+        "path": "app/to-be-created-dir/config",
+        "type": CFG_TYPE_B64,
+        "content": base64.b64encode(b"Config file contents").decode()
+    }
 ]
 
 # The response from SECOND_CACHITO_REQUEST_CONFIG_URL
@@ -591,6 +596,7 @@ def test_resolve_remote_source(workflow, scratch, dr_strs, dependency_replacemen
                 "cachito.env": cachito_env_content,
                 "app/README.txt": "Content of remote-source.tar.gz",
                 "app/some-config.txt": "gomod requests don't actually have configs",
+                "app/to-be-created-dir/config": "Config file contents",
             },
         )
     )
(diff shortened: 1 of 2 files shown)


tests/plugins/test_resolve_remote_source.py
```diff
@@ -967,3 +967,27 @@ def test_generate_cachito_config_files_unknown_type(workflow):
...
+            "path": ".bundle/foo",
+            "type": CFG_TYPE_B64,
+            "content": base64.b64encode(b"contents")
+        }
+    ]
+
+    plugin.generate_cachito_config_files(dest_dir, cfg_files)
+
+
+def test_generate_cachito_config_parent_dir_already_exists(workflow):
```
chmeliik, line 987: It might be best to add another config file to the test constant instead: https://github.com/containerbuildsystem/atomic-reactor/blob/e54f9d295656f7e09c73d33614391f2451e60af9/tests/plugins/test_resolve_remote_source.py#L214-L221

And update the places where `check_injected_files()` is used to check the file contents

chmeliik:
Nitpick: the commit message says `./bundle`, which AFAIK should be `.bundle` or `./.bundle`

tests/plugins/test_resolve_remote_source.py
```diff
@@ -844,6 +850,7 @@ def test_allow_multiple_remote_sources(workflow, allow_multiple_remote_sources):
                     "gomod/cachito.env": first_cachito_env,
                     "gomod/app/README.txt": "Content of remote-source-gomod.tar.gz",
                     "gomod/app/some-config.txt": "gomod requests don't actually have configs",
+                    "gomod/app/.bundle/config": "RubyGems config contents",
```
tichavskym, line 853: @chmeliik Not sure if I completely understood, but I quite changed it, just here it doesn't make completely sense, because it seems to me that this test is using it in a way that it's gonna be part of gomod package.

chmeliik: Hmm, true. It doesn't really matter, the main point is that the parent directory doesn't exist, but if you want to make it less confusing you could change the path/content to something less ruby-specific

And I think we don't need the `test_generate_cachito_config` test anymore, since config files are tested as part of the larger tests

tichavskym: Yeah, I made it less ruby specific, it makes much sense now and double-checked that `test_generate_cachito_config` can be removed (tests failed when I uncommented my changes).

chmeliik (approved):
LGTM :+1:

PR #2016, opened 2022-10-19
allow limiting memory for binary builds on remote hosts

rcerven:

### DIFF
--- a/atomic_reactor/tasks/binary_container_build.py
+++ b/atomic_reactor/tasks/binary_container_build.py
@@ -21,6 +21,7 @@ from osbs.utils import ImageName
 from atomic_reactor import dirs
 from atomic_reactor import util
 from atomic_reactor.constants import REMOTE_HOST_MAX_RETRIES, REMOTE_HOST_RETRY_INTERVAL
+
 from atomic_reactor.tasks.common import Task, TaskParams
 from atomic_reactor.utils import retries
 from atomic_reactor.utils import remote_host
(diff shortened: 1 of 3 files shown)


atomic_reactor/constants.py
```diff
@@ -211,3 +211,4 @@
 }
 
 KOJI_METADATA_FILENAME = "metadata.json"
+DEFAULT_BINARY_MEMORY_LIMIT = "1k"
```
MartinBasti, line 214: Shouldn't be default no limit?

mkosiarc (approved):

mkosiarc:
We should probably also update osbs-docs

rcerven:
> We should probably also update osbs-docs

but what to update there? option is described in config.json (schema) and value is optional

atomic_reactor/tasks/binary_container_build.py
```diff
@@ -225,6 +227,7 @@ def build_container(
         build_args: Dict[str, str],
         dest_tag: ImageName,
         flatpak: bool,
+        memory_limit: str,
```
MartinBasti, line 230: Shouldn't be this `Optional[str]`?

rcerven: well I am always passing it, so it isn't optional

mkosiarc:
> > We should probably also update osbs-docs
> 
> but what to update there? option is described in config.json (schema) and value is optional

If we consider schema as reasonable docs then nothing needs to be done I guess.

ben-alkov (approved):

atomic_reactor/tasks/binary_container_build.py
MartinBasti: No you are not, this can be `None`:
```python
memory_limit=config.remote_hosts.get("memory_limit"),
```

rcerven: ok fixed

MartinBasti (approved):

rcerven:
rn updated

PR #2019, opened 2022-10-31
don't store remote_sources in annotation because if there is too many

rcerven:
sources it will hit task results limit

* CLOUDBLD-11401


chmeliik:
They are also used the store-results task and then in osbs-client and koji-c :disappointed:

rcerven:
> They are also used the store-results task and then in osbs-client and koji-c disappointed

store-results task is using workflow.json and reading repositories & koji-id to set results, nothing is using 'remote_sources'

PR #2028, opened 2022-11-14
raise a specific error when image is not available in registry

hjmodi:
Otherwise an UNAUTHORIZED error is raised that
causes confusion for the user.

* CLOUDBLD-7086

### DIFF
--- a/atomic_reactor/util.py
+++ b/atomic_reactor/util.py
@@ -666,6 +666,10 @@ class RegistryClient(object):
         except (HTTPError, RetryError) as ex:
             if ex.response is None:
                 raise
+            if ex.response.status_code == requests.codes.unauthorized:
+                logger.warning('Requested parent/base image "%s" not found', image.to_str())
+                raise RuntimeError('Unable to fetch image: "{}"'
+                                   ' (Does the image exist?)'.format(image.to_str())) from ex
             if ex.response.status_code == requests.codes.not_found:
                 saved_not_found = ex
             # If the registry has a v2 manifest that can't be converted into a v1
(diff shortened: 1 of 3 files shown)


atomic_reactor/util.py
```diff
@@ -685,8 +689,6 @@ def get_manifest(
                 logger.debug("skipping version %s due to status code %s",
                              version, ex.response.status_code)
                 return None, saved_not_found
-            else:
```
MartinBasti, line 688: is it safe to remove it?

rcerven: we shouldn't remove this just in case

atomic_reactor/util.py
```diff
@@ -666,6 +666,10 @@ def get_manifest(
         except (HTTPError, RetryError) as ex:
             if ex.response is None:
                 raise
+            if ex.response.status_code == requests.codes.unauthorized:
+                logger.warning('Requested parent/base image "%s" not found', image.to_str())
+                raise RuntimeError('Unable to fetch image: "{}"'
```
rcerven, line 671: I don't think that it should be raising here, it should act same as not_found IMHO

atomic_reactor/util.py
hjmodi: remnant of trial and error, added it back

atomic_reactor/util.py
hjmodi: I am not sure that's ideal because the `saved_not_found` exception is only examined during digests, i.e. https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/util.py#L720 during other calls it is simply ignored(https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/util.py#L757, https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/util.py#L784).

it's quite confusing TBH, that's why I chose to do this in this function only rather than relying on other functions I don't fully understand

rcerven: yeah I know that saved_not_found is later ignored, my idea was to propagate that up and not ignore that, but I guess it would need more refactoring

MartinBasti (approved):

rcerven (approved):

PR #2036, opened 2022-12-16
retry commands on remote hosts also on connection and timeout errors

rcerven:
* CLOUDBLD-11291

### DIFF
--- a/atomic_reactor/utils/remote_host.py
+++ b/atomic_reactor/utils/remote_host.py
@@ -23,9 +23,11 @@ from atomic_reactor.utils.rpm import rpm_qf_args
 SSH_COMMAND_TIMEOUT = 30
 SLOTS_RELATIVE_PATH = "osbs_slots"
 RETRY_ON_SSH_EXCEPTIONS = (paramiko.ssh_exception.NoValidConnectionsError,
-                           paramiko.ssh_exception.SSHException)
-BACKOFF_FACTOR = 3
-MAX_RETRIES = 3
+                           paramiko.ssh_exception.SSHException, ConnectionError, TimeoutError)
+# wait time is calculated for backoff.expo: factor * 2 ** n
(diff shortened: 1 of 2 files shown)


ben-alkov (approved):

atomic_reactor/utils/remote_host.py
```diff
@@ -23,9 +23,9 @@
 SSH_COMMAND_TIMEOUT = 30
 SLOTS_RELATIVE_PATH = "osbs_slots"
 RETRY_ON_SSH_EXCEPTIONS = (paramiko.ssh_exception.NoValidConnectionsError,
```
mkosiarc, line 25: I would maybe rename this, since it is not only SSH exceptions

atomic_reactor/utils/remote_host.py
```diff
@@ -23,9 +23,9 @@
...
-                           paramiko.ssh_exception.SSHException)
+                           paramiko.ssh_exception.SSHException, ConnectionError, TimeoutError)
 BACKOFF_FACTOR = 3
-MAX_RETRIES = 3
+MAX_RETRIES = 30
```
mkosiarc, line 28: How did you decide on 30? The formula for the backoff seems to be
```
factor * 2 ** n
```
where the factor is currently 3 and n is the number of the try, thus with the current settings this would continue for a **very** long time until the build timeouted. Is that what we want?

atomic_reactor/utils/remote_host.py
rcerven: but it is only for SSHRetrySession

atomic_reactor/utils/remote_host.py
rcerven: changed

mkosiarc (approved):

rcerven:
rn updated

PR #2062, opened 2023-04-17
split sbom (icm.json) to arch specific files

rcerven:
* STONEBLD-1158


atomic_reactor/plugins/generate_sbom.py
```diff
@@ -181,16 +183,26 @@ def get_sbom_json(self, sbom_url: str) -> Dict[str, Any]:
...
 
         archives = self.koji_session.listArchives(build['build_id'], type=KOJI_BTYPE_ICM)
         sbom_path = self.pathinfo.typedir(build, btype=KOJI_BTYPE_ICM)
         for archive in archives:
-            if archive['filename'] == ICM_JSON_FILENAME:
-                sbom_url = os.path.join(sbom_path, ICM_JSON_FILENAME)
-                break
-        return sbom_url
+            for platform in self.all_platforms:
+                arch_icm_filename = ICM_JSON_FILENAME.format(platform)
```
mkosiarc, line 193: this nested for loop does not seem right, we should not regenerate this for every archive
I prefer something like:
```
...
icm_filenames_platforms = {ICM_JSON_FILENAME.format(platform): platform for platform in self.all_platforms}
for archive in archives:
    if archive['filename'] in icm_filenames_arches:
        platform = icm_filenames_arches[archive['filename']]
        sbom_urls[platform] = os.path.join(sbom_path, archive['filename'])
...   
```
This way we could avoid nested loop and a break statement. What do you think?

rcerven: changed

atomic_reactor/plugins/generate_sbom.py
```diff
@@ -203,47 +214,68 @@ def get_parent_images_nvr(self) -> List[Optional[str]]:
...
+            sbom_urls = self.get_sbom_urls_from_build(build)
+
+            if sbom_urls:
+                for platform in self.all_platforms:
+                    parent_image_sbom_json = self.get_sbom_json(sbom_urls[platform])
+                    parent_components[platform] = parent_image_sbom_json['components']
+
+                    # add reasons from parent images
+                    if 'incompleteness_reasons' in parent_image_sbom_json:
+                        for reason in parent_image_sbom_json['incompleteness_reasons']:
```
mkosiarc, line 236: Maybe
```
for reason in parent_image_sbom_json.get('incompleteness_reasons', {}):
```
would improve the readability, as we could omit the condition and one nested level

mkosiarc:
LGTM with one minor comment. Seems like the pipeline is broken, even in master. I'll try to see what's going on.

atomic_reactor/plugins/generate_sbom.py
rcerven: changed

mkosiarc (approved):

ben-alkov (approved):

rcerven:
rn updated

PR #2108, opened 2024-08-16
fix(source-containers): all cachito metadata

MartinBasti:
Exporting all cachito metadata caused failure in source contianer builds as it only expected 2 archives.

STONEBLD-2655

### DIFF
--- a/tests/plugins/test_fetch_sources.py
+++ b/tests/plugins/test_fetch_sources.py
@@ -40,17 +40,22 @@ KOJI_UPLOAD_TEST_WORKDIR = 'temp_workdir'
 ALL_ARCHIVE_NAMES = ['remote-source-first.json', 'remote-source-first.tar.gz',
                      'remote-source-second.json', 'remote-source-second.tar.gz']
 RS_TYPEINFO = [{'name': 'first', 'url': 'first_url',
+                # compat with older builds, only 2 archives
                 'archives': ['remote-source-first.json', 'remote-source-first.tar.gz']},
                {'name': 'second', 'url': 'second_url',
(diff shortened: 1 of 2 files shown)


tests/plugins/test_fetch_sources.py
```diff
@@ -47,10 +47,10 @@
...
                        {'name': 'second', 'url': 'second_url',
                         'archives': ['remote-source-second.bad', 'remote-source-second.tar.gz']}]
-RS_TYPEINFO_NO_2 = [{'name': 'first', 'url': 'first_url',
-                     'archives': ['remote-source-first.tar.gz']},
-                    {'name': 'second', 'url': 'second_url',
-                     'archives': ['remote-source-second.tar.gz']}]
+RS_TYPEINFO_NO_TAR_GZ = [{'name': 'first', 'url': 'first_url',
+                         'archives': ['remote-source-first.json']},
+                         {'name': 'second', 'url': 'second_url',
+                         'archives': ['remote-source-second.json']}]
```
chmeliik, line 53: Would a test case with `.config.json` and `.env.json` be worth it?

chmeliik (approved):
LGTM, extra testcase optional

tests/plugins/test_fetch_sources.py
MartinBasti: right, I can add it there to check if they are safely ignored

mkosiarc (approved):

tests/plugins/test_fetch_sources.py
MartinBasti: added

chmeliik (approved):

PR #2113, opened 2024-08-20
cachi2: generate remote-source.json

MartinBasti:
Cachi2 doesn't provide remote-source.json but for backward compatibility OSBS must provide it.

STONEBLD-2585

### DIFF
--- a/atomic_reactor/utils/cachi2.py
+++ b/atomic_reactor/utils/cachi2.py
@@ -8,7 +8,9 @@
 Utils to help to integrate with cachi2 CLI tool
 """
 
-from typing import Any, Dict
+from typing import Any, Callable, Dict, Optional, Tuple
+
+from packageurl import PackageURL
 
 
 def remote_source_to_cachi2(remote_source: Dict[str, Any]) -> Dict[str, Any]:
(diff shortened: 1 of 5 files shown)


atomic_reactor/utils/cachi2.py
```diff
@@ -68,3 +68,49 @@ def convert_SBOM_to_ICM(sbom: Dict[str, Any]) -> Dict[str, Any]:
...
+    as some tools are depending on it, we have to generate also request.json from cachi2
+    """
+    def gen_dependency(sbom_dep: Dict[str, Any]) -> Dict[str, str]:
+        """Generate a single dependency from SBOM"""
+        # we need to detect type from purl, this is just heuristics,
+        # we cannot reliably construct type from purl
+        heuristic_type = "unknown"
+        purl_type_map = {
+            "pkg:golang/": "go-package",
+            "pkg:npm/": "npm",
```
chmeliik, line 89: You could also detect `gomod` by looking at the `type` qualifier in the purls
* `pkg:golang/github.com/cachito-testing/cachito-gomod-with-deps@v0.0.0-20201001160613-4c65d49cae6b?type=module`
* `pkg:golang/github.com/cachito-testing/cachito-gomod-with-deps@v0.0.0-20201001160613-4c65d49cae6b?type=package`

atomic_reactor/utils/cachi2.py
```diff
@@ -68,3 +68,49 @@ def convert_SBOM_to_ICM(sbom: Dict[str, Any]) -> Dict[str, Any]:
...
+        for purl_prefix, request_type in purl_type_map.items():
+            if purl.startswith(purl_prefix):
+                heuristic_type = request_type
+                break
+
+        return {
+            "name": sbom_dep["name"],
+            # "replaces": None, #  TBD: this is go only, but will be always none as we don't support replacements in cachi2; should it be included?
+            "type": heuristic_type,
+            "version": sbom_dep.get("version"),
```
chmeliik, line 104: For a bit more backward compatibility, could fall back to extracting the `download_url` or `vcs_url` qualifiers from the purl (Cachito used to report the URL as the version for non-registry dependencies)

For even more compatibility, another fallback would be to extract the `#path` from the purl and report is as `./path` for Go dependencies and as `file:path` for anything else

atomic_reactor/utils/cachi2.py
```diff
@@ -68,3 +68,49 @@ def convert_SBOM_to_ICM(sbom: Dict[str, Any]) -> Dict[str, Any]:
...
+        }
+        purl = sbom_dep["purl"]
+        for purl_prefix, request_type in purl_type_map.items():
+            if purl.startswith(purl_prefix):
+                heuristic_type = request_type
+                break
+
+        return {
+            "name": sbom_dep["name"],
+            # "replaces": None, #  TBD: this is go only, but will be always none as we don't support replacements in cachi2; should it be included?
```
chmeliik, line 102: It is optional per the API spec https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/web/static/api_v1.yaml#L897-L900

But I think Cachito always included it, so we can keep it to be safe

atomic_reactor/utils/cachi2.py
```diff
@@ -68,3 +68,49 @@ def convert_SBOM_to_ICM(sbom: Dict[str, Any]) -> Dict[str, Any]:
...
+        }
+
+    res = {
+        "dependencies": [gen_dependency(dep) for dep in remote_source_sbom["components"]],
+        "pkg_managers": remote_source.get("pkg_managers", []),
+        "ref": remote_source["ref"],
+        "repo": remote_source["repo"],
+        "environment_variables": {key: val["value"] for key, val in remote_source_env_json.items()},
+        "flags": remote_source.get("flags", []),
+        "packages": [],  # TBD: this will be always empty, should we include it?
```
chmeliik, line 114: Same here, as far as I can tell from the API spec it's not required (in fact I did not find any required properties) but previously it would have always been there

atomic_reactor/utils/cachi2.py
chmeliik: And now that I think about it, those shouldn't really be fallbacks but more like "fallfronts" - Cachito would prefer reporting a URL or path as the version even when a version number was available.

atomic_reactor/utils/cachi2.py
```diff
@@ -68,3 +70,74 @@ def convert_SBOM_to_ICM(sbom: Dict[str, Any]) -> Dict[str, Any]:
...
+        sbom_dep.get("version")
+        # returns None if version cannot be determined
+    )
+
+    return {
+        "name": sbom_dep["name"],
+        "replaces": None,  # it's always None, replacements aren't supported by cachi2
+        "type": heuristic_type,
+        "version": version,
+        # "dev": True  # TODO
```
MartinBasti, line 117: @chmeliik can we somehow determine this?

chmeliik: For npm packages (and maybe yarn in the future) yes, from the [cdx:npm:package:bundled](https://github.com/containerbuildsystem/cachi2/blob/ae89bbab309468d3db4bc7ab3c9f88356c46b4d0/cachi2/core/models/property_semantics.py#L80) property

For others no

atomic_reactor/utils/cachi2.py
MartinBasti: updated

atomic_reactor/utils/cachi2.py
MartinBasti: updated

atomic_reactor/utils/cachi2.py
MartinBasti: updated

atomic_reactor/utils/cachi2.py
MartinBasti: ack

atomic_reactor/utils/cachi2.py
MartinBasti: added for NPM

mkosiarc (approved):

mkosiarc:
LGTM, in ideal world we could use more abstraction by exporting some functionality to individual helper functions so they could be read and understood better (for example, the type matching or version finding). But I am not expecting any change here.

mkosiarc:
I appreciate the comments, without them I would be completely lost.

PR #2121, opened 2024-10-18
Cachi2 task

MartinBasti:


atomic_reactor/plugins/cachi2_postprocess.py
```diff
@@ -0,0 +1,243 @@
...
+                raise RuntimeError(
+                    f"Conflicting path {dest_dir.relative_to(build_dir.path)} already exists "
+                    "in the dist-git repository"
+                )
+
+            dest_dir.mkdir(parents=True)
+            created_dirs.append(dest_dir)
+
+            # copy app and deps generated by cachito into build_dir
+            # TODO: reflink?
```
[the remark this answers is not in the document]
MartinBasti: added as commit

atomic_reactor/plugins/cachi2_init.py
```diff
@@ -0,0 +1,154 @@
...
+
+        return processed_remote_sources
+
+    def clone_remote_source(self, repo: str, target_dir: Path, commit: str):
+        self.log.debug("Cloning %s at %s into %s", repo, commit, target_dir)
+        clone_git_repo(
+            repo,
+            target_dir,
+            commit
+        )
```
chmeliik, line 129: I think you're already aware, but this will also need git-submodule handling eventually

tekton/tasks/binary-container-cachi2.yaml
```diff
@@ -0,0 +1,53 @@
...
+        requests:
+          memory: 512Mi
+          cpu: 250m
+        limits:
+          memory: 1Gi
+          cpu: 395m
+      script: |
+        set -x
+        atomic-reactor -v task \
+          --user-params="$(params.user-params)" \
```
chmeliik, line 53: Is this how we pass user params in other tasks? It would surprise me if this didn't break, it's JSON and contains double-quotes (and `$(params)` substitution doesn't escape any special characters)

It would be safer to pass it as an environment variable

chmeliik (commented):
Looks reasonable :+1:

atomic_reactor/cli/task.py
```diff
@@ -65,6 +66,16 @@ def binary_container_cachito(task_args: dict):
...
 
 
+def binary_container_cachi2_init(task_args: dict):
+    """Run binary container Cachi2 init step.
+
+    :param task_args: CLI arguments for a binary-container-cachi2-init task
+    """
+    params = TaskParams.from_cli_args(task_args)
+    task = BinaryCachi2InitTask(params)
+    return task.run(init_build_dirs=True)
```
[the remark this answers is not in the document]
MartinBasti: without this, it fails horribly that source wasn't initialized

atomic_reactor/plugins/cachi2_init.py
MartinBasti: yes, there is followup story to support git-submodule "pkg-manager"

tekton/tasks/binary-container-cachi2.yaml
MartinBasti: I tested this bash magic locally and works fine, it works fine also in integration tests; :magic_wand:

MartinBasti: updated to env var

chmeliik (approved):
LGTM, just do not merge the DO NOT MERGE commit :smile:

MartinBasti:
Rebased and removed DO NOT MERGE commit

PR #2127, opened 2024-11-06
Cachi2: update postbuild plugins

MartinBasti:
Updating post build plugins to support cachi2 task

### DIFF
--- a/atomic_reactor/plugins/koji_import.py
+++ b/atomic_reactor/plugins/koji_import.py
@@ -36,6 +36,7 @@ from atomic_reactor.utils.koji import (
 from atomic_reactor.plugins.fetch_sources import PLUGIN_FETCH_SOURCES_KEY
 
 from atomic_reactor.constants import (
+    PLUGIN_CACHI2_POSTPROCESS,
     PLUGIN_EXPORT_OPERATOR_MANIFESTS_KEY,
     PLUGIN_KOJI_IMPORT_PLUGIN_KEY,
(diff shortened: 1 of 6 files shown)


atomic_reactor/plugins/koji_import.py
```diff
@@ -303,41 +304,76 @@ def set_pnc_build_metadata(self, extra):
...
-            extra.setdefault("typeinfo", {}).update(remote_source_typeinfo)
+
+        remote_source_typeinfo = {
+            KOJI_BTYPE_REMOTE_SOURCES: remote_sources_typeinfo_metadata,
+        }
+        extra.setdefault("typeinfo", {}).update(remote_source_typeinfo)
+
+    def set_remote_sources_metadata_cachi2(self, remote_source_result, extra):
+        remote_sources_typeinfo_metadata = []
+        if self.workflow.conf.allow_multiple_remote_sources:
```
chmeliik, line 341: It seems the `else` to this `if` is missing

MartinBasti: it seems but it's not, else is only to export Cachito URL, but we don't have that in cachi2. Dunno if I should export some other metadata in this case.

Even better would be to just allow multiple sources permanently, it would reduce lot of code

chmeliik: Ah, right

chmeliik (approved):
LGTM :+1:

mkosiarc (approved):

PR #2137, opened 2024-12-05
cachi2: allow only relative paths

MartinBasti:
For security reasons, only relative paths within cloned remote source can be specified by users

### DIFF
--- a/atomic_reactor/utils/cachi2.py
+++ b/atomic_reactor/utils/cachi2.py
@@ -9,8 +9,52 @@ Utils to help to integrate with cachi2 CLI tool
 """
 
 from typing import Any, Callable, Dict, Optional, Tuple, List
+from pathlib import Path
+import os.path
 
 from packageurl import PackageURL
+
+
+def validate_paths(repo_path: Path, remote_sources_packages: dict) -> None:
+    """Paths must be relative and within cloned repo"""
+    def is_path_ok(path_string):
+        path = Path(path_string)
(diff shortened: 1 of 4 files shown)


atomic_reactor/utils/cachi2.py
```diff
@@ -9,10 +9,54 @@
...
 
+def _validate_paths(remote_sources_packages: dict) -> None:
+    """Paths must be relative and within cloned repo"""
+    def is_path_ok(path_string):
+        path = Path(path_string)
+        if path.is_absolute():
+            return False
+
+        fake_root = Path("/fake_root")
+        full_path = (fake_root/path).resolve()
```
chmeliik, line 26: A path may be OK within a fake root but not within a real root (if it's a symlink in the real root)

MartinBasti: but I can rework it to use real cloned data, not a big issue to be sure that validation is complete

chmeliik: Hmm, ok

chmeliik (approved):

atomic_reactor/utils/cachi2.py
chmeliik: But yeah, if it's not a big change to run on the real repo, that would be nicer

MartinBasti:
Updated to check symlinks, PTAL

mkosiarc (approved):

PR #2140, opened 2024-12-19
Add to the icm schema a field to indicate if the icm should be ignored

ralphbean:
This enables a path to deprecation.

Today, scanners depend on this file primarilly for the list of content_sets.

In the future, with konflux-ci, we want third-party scanners to refer to the dnf database for a more precise mapping of which rpms come from which content_sets.

The idea here is that image providers can still publish an icm file, but if the from_dnf_hint is set to true, then scanners are supposed to ignore the icm file and trust the dnf database instead. Scanners that haven't been ported to the new method will continue to work in the current imprecise way until they adapt to scanning the dnf database.

### DIFF
--- a/atomic_reactor/schemas/content_manifest.json
+++ b/atomic_reactor/schemas/content_manifest.json
@@ -35,6 +35,11 @@
       "description": "Pulp content sets available during the current image build",
       "type": "array",
       "items": { "$ref": "#/definitions/content_set" }
+    },
+    "from_dnf_hint": {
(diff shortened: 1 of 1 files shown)


MartinBasti (approved):

MartinBasti:
https://github.com/containerbuildsystem/atomic-reactor/pull/2142 fix for CI failure

mkosiarc (approved):

MartinBasti:
I'm merge it, only coveralls report failed

PR #2144, opened 2025-01-17
feat(cachi2): add support for git-submodules

MartinBasti:
for git-submodules pkg manager, OSBS msut hadnle all the work, cachi2 doesn't manipuilate git.

Submodules has to be cloned and metadata exported into SBOM and request.json

### DIFF
--- a/atomic_reactor/plugins/cachi2_init.py
+++ b/atomic_reactor/plugins/cachi2_init.py
@@ -10,6 +10,7 @@ from collections import Counter
 from typing import Any, Optional, List, Dict
 from pathlib import Path
 
+import git
 from osbs.utils import clone_git_repo
 
 from atomic_reactor.constants import (
@@ -28,6 +29,8 @@ from atomic_reactor.util import map_to_user_params
 from atomic_reactor.utils.cachi2 import (
     remote_source_to_cachi2, clone_only, validate_paths,
     normalize_gomod_pkg_manager, enforce_sandbox,
+    has_git_submodule_manager, update_submodules,
+    get_submodules_sbom_components, get_submodules_request_json_deps,
 )
 
 
@@ -135,6 +138,15 @@ class Cachi2InitPlugin(Plugin):
                 remote_source_data["ref"]
             )
 
+            if has_git_submodule_manager(remote_source_data):
+                update_submodules(source_path_app)
+                repo = git.Repo(str(source_path_app))
+                git_submodules = {
+                    "sbom_components": get_submodules_sbom_components(repo),
+                    "request_json_dependencies": get_submodules_request_json_deps(repo)
+                }
+                remote_source["git_submodules"] = git_submodules
+
             remove_unsafe_symlinks = False
             flags = remote_source_data.get("flags", [])
             if "remove-unsafe-symlinks" in flags:
@@ -144,6 +156,7 @@ class Cachi2InitPlugin(Plugin):
                 source_path_app,
                 remove_unsafe_symlinks,
             )
+
             validate_paths(source_path_app, remote_source_data.get("packages", {}))
 
             if clone_only(remote_source_data):
--- a/atomic_reactor/plugins/cachi2_postprocess.py
+++ b/atomic_reactor/plugins/cachi2_postprocess.py
@@ -20,6 +20,7 @@ import reflink
 from atomic_reactor.constants import (
     CACHITO_ENV_ARG_ALIAS,
     CACHITO_ENV_FILENAME,
+    CACHI2_BUILD_DIR,
     PLUGIN_CACHI2_INIT,
     PLUGIN_CACHI2_POSTPROCESS,
     REMOTE_SOURCE_DIR,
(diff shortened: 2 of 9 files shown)


atomic_reactor/plugins/cachi2_postprocess.py
```diff
@@ -110,13 +111,38 @@ def run(self) -> Optional[List[Dict[str, Any]]]:
...
+            git_submodules = remote_source.get('git_submodules')
+            if not git_submodules:
+                continue
+
+            all_sboms.extend(git_submodules['sbom'])
+
+        if not all_sboms:
+            return
+
+        # TODO: deduplicate sboms?
```
MartinBasti, line 135: I'm not sure if there should be a deduplication for global sbom in case of that the same submodule is used within multiple remote sources.

Maybe play safe and don't deduplicate, WDYT?

MartinBasti:
I tried to cover everything by unitests, however changes in init plugin would require too much git mocking and I would end up with testing mocking rather than testing something useful. It has integration test, that should be sufficient.

atomic_reactor/plugins/cachi2_postprocess.py
chmeliik: Cachito did use to de-duplicate https://github.com/containerbuildsystem/cachito/blob/2e4795a7687b70a468e4f686ef05988994af1a7b/cachito/web/api_v1.py#L931-L939

But CycloneDX doesn't technically require uniqueness if you're not using [bom-ref](https://cyclonedx.org/docs/1.5/json/#components_items_bom-ref) attributes (which we're not), so it's not really necessary

atomic_reactor/utils/cachi2.py
```diff
@@ -287,4 +292,89 @@ def clone_only(remote_source: Dict[str, Any]) -> bool:
...
+
+def has_git_submodule_manager(remote_source: Dict[str, Any]) -> bool:
+    """Returns true when for specific remote source git-submodule manager is requested"""
+    pkg_managers = remote_source.get("pkg_managers") or []
+    return 'git-submodule' in pkg_managers
+
+
+def update_submodules(repopath: Path):
+    """Update submodules in the given repo"""
+    cmd = ["git", "submodule", "update", "--init", "--filter=blob:none"]
```
chmeliik, line 310: What is the `--filter=blob:none` based on?

atomic_reactor/plugins/cachi2_init.py
```diff
@@ -135,6 +138,15 @@ def process_remote_sources(self) -> List[Dict[str, Any]]:
                 remote_source_data["ref"]
             )
 
+            if has_git_submodule_manager(remote_source_data):
+                update_submodules(source_path_app)
+                repo = git.Repo(str(source_path_app))
+                git_submodules = {
+                    "sbom": get_submodules_sbom(repo),
+                    "request_json": get_submodules_request_json_deps(repo)
```
chmeliik, line 146: Naming nitpick: these two are just the `.components` part of the SBOM and the `.dependencies` part of the request JSON. But the code that uses these is simple enough that the confusion is not too large, so renaming is optional

atomic_reactor/utils/cachi2.py
MartinBasti: I took it from Cachito to do exactly same behavior, I don't know

atomic_reactor/plugins/cachi2_init.py
MartinBasti: I updated more parts, trying to be more consistent with mentioning components

atomic_reactor/utils/cachi2.py
chmeliik: Ah yup, found it https://github.com/containerbuildsystem/cachito/blob/2e4795a7687b70a468e4f686ef05988994af1a7b/cachito/workers/scm.py#L297

chmeliik (approved):

atomic_reactor/utils/cachi2.py
```diff
@@ -287,4 +292,89 @@ def clone_only(remote_source: Dict[str, Any]) -> bool:
...
+    # only git-submodule
+    if pkg_managers is not None and pkg_managers == ['git-submodule']:
+        return True
+
     return False
+
+
+def has_git_submodule_manager(remote_source: Dict[str, Any]) -> bool:
+    """Returns true when for specific remote source git-submodule manager is requested"""
+    pkg_managers = remote_source.get("pkg_managers") or []
```
ben-alkov, line 304: Why not

```suggestion
    pkg_managers = remote_source.get("pkg_managers", [])
```

Not required, just to satisfy my curiosity...

MartinBasti: in case that somebody specify `pkg_managers:`  as empty in yaml a.k.a `null`, it would explode later. Probably jsonschema should catch it, but defensive programming :D

ben-alkov (approved):
LGTM

atomic_reactor/utils/cachi2.py
ben-alkov: Isn't that what the second arg to `get` is for? Maybe I'm misreading it...

PR #2148, opened 2025-02-03
cachi2: remote sources include-git-dir flag

MartinBasti:
When include-git-dir flag is used, keep .git directory present

### DIFF
--- a/tekton/tasks/binary-container-cachi2.yaml
+++ b/tekton/tasks/binary-container-cachi2.yaml
@@ -78,6 +78,7 @@ spec:
         set -eux
         CACHI2_DIR="$(workspaces.ws-build-dir.path)/_cachi2_remote_sources"
         CACHI2_PKG_OPT_PATH="cachi2_pkg_options.json"
+        CACHI2_INCLUDE_GIT_DIR_FILE="cachi2_include_git_dir.txt"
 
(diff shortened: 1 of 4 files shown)


ben-alkov (approved):
LGTM

tekton/tasks/binary-container-cachi2.yaml
```diff
@@ -113,7 +114,11 @@ spec:
           # we can rely on presence of this file
           SBOMS+=("${REMOTE_SOURCE_PATH}/bom.json")
 
-          rm -fr app/.git/   # remove git directory
+          if [ -f "${CACHI2_INCLUDE_GIT_DIR_FILE}" ]; then
```
MartinBasti, line 117: something here is wrong, file exists, but this condition wasn't evaluated as true

however it is true in my local testing, magic

MartinBasti: because I'm stupid that's why, I haven't updated reference to this task for testing PR

MartinBasti: it's ready, it's was harder to tests it than write it

chmeliik (approved):
LGTM, just need to drop the second commit

PR #2161, opened 2025-03-07
Skip unneeded platform builds

MartinBasti:

### DIFF
--- a/tekton/pipelines/binary-container.yaml
+++ b/tekton/pipelines/binary-container.yaml
@@ -243,6 +243,10 @@ spec:
         - name: platform
           value: x86_64
       timeout: "0"
+      when:
+        - input: "x86_64"
+          operator: in
+          values: ["$(tasks.binary-container-init.results.platforms_array[*])"]
 
     - name: binary-container-build-s390x
       runAfter:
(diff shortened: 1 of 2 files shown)


ben-alkov (approved):

tekton/pipelines/binary-container.yaml
```diff
@@ -78,7 +78,14 @@ spec:
       runAfter:
         - clone
       taskRef:
-        name: binary-container-init-0-1
+        resolver: git
```
rcerven, line 81: this seems like some testing change which should be reverted

MartinBasti: yes, it's from commit `DO NOT MERGE`. I'll remove it before merging

tekton/tasks/binary-container-init.yaml
```diff
@@ -50,6 +52,7 @@ spec:
           cpu: 395m
       script: |
         set -x
+        set -e
```
eskultety, line 55: nitpick: this one is technically an unrelated change that should go to a separate commit. Also, can you put it on the same line as the setting above instead of running the builtin twice?

eskultety (commented):
Trivial nitpick. ACK to the PR (I'll hit the approve once commit nr. 2 is dropped).

tekton/tasks/binary-container-init.yaml
MartinBasti: it's related change, now task has 2 commands, thus it wouldn't work as expected because only last return status is being processed by tekton. Thus with 2 commands, I need to stop execution and return error at first occurrence.

Merged into one line

ben-alkov (approved):

PR #2164, opened 2025-03-12
Hermeto: generate cyclonedx SBOM

MartinBasti:
OSBS supports only cyclonedx, make sure to explicitly require cyclonedx format in case that default format changes


eskultety (approved):
Okay, but changing defaults like these is not nice and we shouldn't do that in the upstream project (changing tekton tasks is a different story) unless we have very strong motivation for SPDX become the only format, especially since cachi2 can't generate SPDX in a way other than converting from CycloneDX.

ACK, trivial.

ben-alkov (approved):

PR #2183, opened 2026-04-20
Fix for UBI10 RPMDB location

obriencj:
Updates the rpmqa plugin such that it searches both the original /var/lib/rpm location, and if that fails attempts a fallback list of dirs, currently just /usr/lib/sysimage/rpm

If nothing is found, it still raises a NothingExtractedError as before, unless it is a base_from_scratch image in which case it preserves original behavior and returns an empty list

### DIFF
--- a/atomic_reactor/plugins/rpmqa.py
+++ b/atomic_reactor/plugins/rpmqa.py
@@ -18,7 +18,12 @@ from atomic_reactor.utils.imageutil import NothingExtractedError
 from atomic_reactor.utils.rpm import parse_rpm_output
 from atomic_reactor.utils.rpm import rpm_qf_args
 
+# Where we look for the RPMDB in an image by default
 RPMDB_PATH = '/var/lib/rpm'
+
+# Additional locations if the default path above fails
+RPMDB_PATH_FALLBACKS = ['/usr/lib/sysimage/rpm', ]
+
 RPMDB_DIR_NAME = 'rpm'
 
 __all__ = ('RPMqaPlugin',)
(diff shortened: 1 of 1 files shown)


atomic_reactor/plugins/rpmqa.py
```diff
@@ -53,13 +58,24 @@ def run(self):
...
+            for rpmdb_origin in [RPMDB_PATH, *RPMDB_PATH_FALLBACKS]:
+                try:
+                    self.workflow.imageutil.extract_file_from_image(image, rpmdb_origin, rpmdb_dir)
+                except NothingExtractedError:
+                    if self.workflow.data.dockerfile_images.base_from_scratch:
+                        self.log.info("scratch image doesn't contain or has empty rpmdb %s", rpmdb_origin)
+                    else:
+                        self.log.info("image doesn't contain or has empty rpmdb %s", rpmdb_origin)
+                else:
+                    self.log.info("found rpmdb %s" % rpmdb_origin)
```
MartinBasti:
```suggestion
                    self.log.info("found rpmdb %s", rpmdb_origin)
```

MartinBasti:
also Flake8 should be improved, ignore mypy and tekton linters

atomic_reactor/plugins/rpmqa.py
```diff
@@ -65,11 +65,11 @@ def gather_output(self, build_dir: BuildDir) -> List[RpmComponent]:
                     self.workflow.imageutil.extract_file_from_image(image, rpmdb_origin, rpmdb_dir)
                 except NothingExtractedError:
                     if self.workflow.data.dockerfile_images.base_from_scratch:
-                        self.log.info("scratch image doesn't contain or has empty rpmdb %s", rpmdb_origin)
+                        self.log.info(f"scratch image doesn't contain rpmdb at {rpmdb_origin}")
```
MartinBasti: f-strings shouldn't be used in logger. To prevent unneeded string interpolation, use log.info("text %s", var, ...)

MartinBasti:
Please squash those fixup commits into one commit

MartinBasti (approved):

mkosiarc (approved):

